PR #2367, opened 2021-02-23
EES-1920 - add .env.example for UI tests

luke-h1:
This PR: 
 - Alters the root `.gitignore` in order to allow the `.env.example ` to be added in ~tests/robot-tests
 - Adds a `.env.example` to `~tests/robot-tests`


tests/robot-tests/.env.example
```diff
@@ -0,0 +1,9 @@
+PUBLIC_URL=PUBLIC_URL
```
ntsim, line 1: Would be a good idea to put the default values in and leave out the passwords

.gitignore
```diff
@@ -32,7 +32,11 @@ junit.xml
 .pabotsuitenames
 /tests/robot-tests/backup-data
 /tests/robot-tests/webdriver
-/tests/robot-tests/.env.*
+/tests/robot-tests/.env
```
ntsim, line 35: If we're just looking to preserve the `.env.example`, then the following might be better:
```
/tests/robot-tests/.env.*
!/tests/robot-tests/.env.example
```

This prevents accidentally checking in envs for pre-prod and test for example.

ntsim (changes requested):

.gitignore
luke-h1: Oh that's pretty cool. Didn't know you could do that.

ntsim (approved):

PR #2369, opened 2021-02-23
EES-1929 - Fix pre-release UI test

luke-h1:


tests/robot-tests/tests/admin_and_public/bau/prerelease_visibility.robot
```diff
@@ -142,7 +142,6 @@ Go to Table Tool page
 Check scheduled release isn't visible
     [Tags]  HappyPath
     environment variable should be set  TEST_THEME_NAME
-    user opens details dropdown    %{TEST_THEME_NAME}
     user checks page does not contain  UI test topic %{RUN_IDENTIFIER}
```
ntsim, line 145: This could also give false positives when running the tests in parallel.

This would be a better assertion:

```
user checks page does not contain  ${PUBLICATION_NAME}
```

luke-h1: Ah yeah, my bad

ntsim (approved):

PR #2371, opened 2021-02-23
Ees 1882 mat file slow import

duncan-at-hiveit:
This PR:
- Replaces the use of Owned Entities in the Location class to greatly simplify the SQL generated by EF Core 3 when fetching Locations from the DB, which was causing the slowdown of the MAT file import.
- Replaces the separate load of Locations in ObservationService with a standard Include, as the SQL is now sane!
- Adds in a utility class `LoggerTimingExtensions` which is now unused in this PR, but I would like to retain for future use to test timings of code blocks when performing future optimisation work (example usage below) 

It looks like EF Core 5 will make the SQL around Owned Entities better, so the IObservationalUnit hierarchy has been retained in this PR in order to switch back to the use of Owned Entities after upgrade.

It's also worth noting that the lookup of Locations (in `ImporterLocationServer.Lookup()`) can be greatly improved further if it's possible to rewrite the null check pattern from:

```
predicateBuilder = predicateBuilder
                .And(location => location.LocalAuthorityDistrict_Code == (localAuthorityDistrict != null ? localAuthorityDistrict.Code : null)
                                 && location.LocalAuthorityDistrict_Name == (localAuthorityDistrict != null ? localAuthorityDistrict.Name : null));
```
to 
```
if (localAuthorityDistrict != null) {
       predicateBuilder = predicateBuilder
                .And(location => location.LocalAuthorityDistrict_Code == localAuthorityDistrict.Code
                                 && location.LocalAuthorityDistrict_Name == localAuthorityDistrict.Name);
}
```

but in the scope of this PR, this is behaviour change that could potentially be damaging.

`LoggerTimingExtensions` class is unused in this PR but logs out indented timing logs for code chunks, like:

```
[processor] [2021-02-23T21:20:08.578Z] Starting to time complete Stage 2
[processor] [2021-02-23T21:20:24.227Z]     Starting to time look up a location or create a new one
[processor] [2021-02-23T21:20:24.258Z]     Took 30.7239 milliseconds to look up a location or create a new one
[processor] [2021-02-23T21:20:24.259Z]     Starting to time look up a location or create a new one
[processor] [2021-02-23T21:20:24.259Z]          Took 0.095 milliseconds to create new location
[processor] [2021-02-23T21:20:24.259Z]     Took 0.1569 milliseconds to look up a location or create a new one
.......................
.......................
.......................
[processor] [2021-02-23T21:20:08.578Z] Took 2037.123 milliseconds to complete Stage 2
```

I'd like to retain this in the codebase for future use if possible.


src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
```diff
@@ -185,7 +185,7 @@ public SubjectMeta GetMeta(DataTable table, Subject subject, StatisticsDbContext
             await InsertObservations(context, observations);
         }
 
-        public static GeographicLevel GetGeographicLevel(IReadOnlyList<string> line, List<string> headers)
+        public GeographicLevel GetGeographicLevel(IReadOnlyList<string> line, List<string> headers)
```
benoutram, line 188: Just wondering if  these methods turning from static into instance methods is a hangover of adding logging and then removing it, or something like that?

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/Interfaces/IImporterService.cs
```diff
@@ -26,5 +29,11 @@ public interface IImporterService
             DataRowCollection rows,
             SubjectMeta subjectMeta,
             StatisticsDbContext context);
+
+        GeographicLevel GetGeographicLevel(IReadOnlyList<string> line, List<string> headers);
+        
+	TimeIdentifier GetTimeIdentifier(IReadOnlyList<string> line, List<string> headers);
```
benoutram, line 35: Looks like some tab characters here.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ValidatorService.cs
```diff
@@ -262,9 +266,9 @@ public class ValidatorService : IValidatorService
                     var rowValues = CsvUtil.GetRowValues(row);
                     var colValues = CsvUtil.GetColumnValues(cols);
 
-                    ImporterService.GetGeographicLevel(rowValues, colValues);
-                    ImporterService.GetTimeIdentifier(rowValues, colValues);
-                    ImporterService.GetYear(rowValues, colValues);
+                    _importerService.GetGeographicLevel(rowValues, colValues);
```
benoutram, line 269: See previous comment about static methods.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/host.json
```diff
@@ -4,6 +4,7 @@
   "logging": {
     "logLevel": {
       "default": "Information",
+      "GovUk.Education": "Information",
```
benoutram, line 7: Just wondering, does this make any difference to what is currently logged?

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ObservationService.cs
```diff
@@ -125,6 +125,7 @@ public IEnumerable<Observation> FindObservations(ObservationQueryContext query)
                     .Observation
                     .AsNoTracking()
                     .Include(o => o.FilterItems)
+                    .Include(o => o.Location)
```
benoutram, line 128: Just wondering if you also need the same line `.Include(o => o.Location)`  in method FindObservations(SubjectMetaQueryContext query) below on line 143 because that's what actually uses `ObservationPredicateBuilder` in this code.

benoutram (approved):
Approving, but with a question about whether there's a missing `.Include`, in `ObservationService#FindObservations(SubjectMetaQueryContext query)` if that's required?

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
duncan-at-hiveit: It was exactly that!  I figured I didn't see the harm in leaving them as instance methods

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ObservationService.cs
duncan-at-hiveit: I'm thinking probably not.  The reason this one is here is that I am replacing the "manual setting of Locations" code that was in this method beforehand due to the funky SQL being produced.  The other method has no such Location setting code, so I'd assume it doesn't need the Locations retrieving in order to work.

PR #2380, opened 2021-02-26
EES-1935 Various UI changes

markhiveit:


src/explore-education-statistics-common/src/modules/table-tool/components/SubjectForm.tsx
```diff
@@ -115,7 +113,13 @@ const SubjectForm = ({
           <Form {...form} id={formId} showSubmitError>
             <FormFieldRadioGroup<SubjectFormValues>
               name="subjectId"
-              legend={legend}
+              legend={[
+                legend,
+                <div key="legend-description" className="govuk-inset-text">
```
ntsim, line 118: Should use the `hint` prop. Having it like this isn't super accessible as it doesn't provide an association between it and the legend.

It's also weird to be using inset text here as well.

src/explore-education-statistics-frontend/src/modules/find-statistics/PublicationReleasePage.tsx
```diff
@@ -319,7 +319,7 @@ const PublicationReleasePage: NextPage<Props> = ({ data }) => {
               );
             }}
           >
-            <AccordionSection heading="Download data">
+            <AccordionSection heading="Download data and files">
```
ntsim, line 322: This is going to break a bunch of UI tests. Make sure to fix these once this has been merged back into dev.

ntsim (changes requested):

ntsim (approved):
Bam :boom: nice!

PR #2395, opened 2021-03-03
EES-1713 Fixes for Admin text block images

markhiveit:


src/explore-education-statistics-common/src/modules/release/hooks/useReleaseImageAttributeTransformer.ts
```diff
@@ -16,8 +16,10 @@ export default function useReleaseImageAttributeTransformer(options: {
     ({ src, srcset, ...attributes }: Dictionary<string>) => {
       return {
         ...attributes,
-        src: `${rootUrl}${replaceReleaseIdPlaceholders(src, releaseId)}`,
-        srcset: `${rootUrl}${replaceReleaseIdPlaceholders(srcset, releaseId)}`,
+        src: src && `${rootUrl}${replaceReleaseIdPlaceholders(src, releaseId)}`,
```
benoutram, line 19: Is `src` ever null here?

src/explore-education-statistics-admin/src/pages/release/hooks/useReleaseImageUpload.ts
```diff
@@ -23,9 +23,7 @@ export default function useReleaseImageUpload(
     async (file, updateProgress) => {
       promise.current = releaseImageService
         .upload(releaseId, file, updateProgress)
-        .then(result =>
-          mapValues(result, insertReleaseIdPlaceholders),
-        ) as CancellablePromise<ImageUploadResult>;
+        .then() as CancellablePromise<ImageUploadResult>;
```
benoutram, line 26: How come we need this change?

src/explore-education-statistics-frontend/src/modules/find-statistics/components/PublicationSectionBlocks.tsx
```diff
@@ -3,25 +3,24 @@ import useGetReleaseFile from '@common/modules/release/hooks/useGetReleaseFile';
...
 import useReleaseImageAttributeTransformer from '@common/modules/release/hooks/useReleaseImageAttributeTransformer';
-import { Release } from '@common/services/publicationService';
 import { Block } from '@common/services/types/blocks';
 import ButtonLink from '@frontend/components/ButtonLink';
 import { logEvent } from '@frontend/services/googleAnalyticsService';
 import React from 'react';
 
 export interface PublicationSectionBlocksProps {
-  release: Release;
+  releaseId: string;
```
benoutram, line 12: I don't think this change is related to any problems found, and it might have been safer to keep it as Release so that it's known to definitely be the ReleaseId rather than just any other string? It might be useful for the future too if other fields of Release are needed.

src/explore-education-statistics-admin/src/pages/release/hooks/useReleaseImageUpload.ts
markhiveit: If this is left in, you get a broken image immediately after uploading it to a text block - if I remember correctly. So I moved the place where the `{releaseId} placeholders were add to when a block is saved, preventing the user from seeing a broken image.

src/explore-education-statistics-common/src/modules/release/hooks/useReleaseImageAttributeTransformer.ts
markhiveit: I'll try removing it.

src/explore-education-statistics-frontend/src/modules/find-statistics/components/PublicationSectionBlocks.tsx
markhiveit: I figured that since only the releaseId was being used that there was no reason to pass the whole release. Just did it I think when I was adding the `replace('/api', '')` stuff. I'll change it back

src/explore-education-statistics-common/src/modules/release/hooks/useReleaseImageAttributeTransformer.ts
benoutram: Cool 👍

src/explore-education-statistics-frontend/src/modules/find-statistics/components/PublicationSectionBlocks.tsx
benoutram: Ok, cool 🙂

markhiveit:
I tried moving the sanitizing/transforming of img attributes to FormEditor rather than do it in EditableContentBlock, but this led to a replace error occurring in the sanitize-html code. It seems to be because in CKEditor, after uploading an image, the first content it produces includes just an `` tag with no attributes, and that seems to trip up with the particular transform we have sanitizeHtml doing. I inspected the sanitize-html code a little, and error is occuring on the `>` of the `` tag. So, for the sake of time, rather than go down the rabbit hole on this, I'm going to leave it.

benoutram (approved):

PR #2401, opened 2021-03-05
EES-185 - Default data boundaries footnote added to charts 

HiveIT-Lewis:


src/explore-education-statistics-common/src/modules/charts/components/ChartRenderer.tsx
```diff
@@ -63,14 +63,29 @@ function ChartRenderer({
...
+    const boundaryFootnoteLabel = `The boundary data used in this map includes${meta.boundaryLevels
+      .map((value, index, array) => {
+        let separator = index === 0 ? ' ' : ', ';
+        if (array.length - 1 === index) {
+          separator = ' and ';
+        }
+        return `${separator}${value.label}`;
+      })
+      .join('')}`;
+    footnotes.push({ id: `map-footnote`, label: boundaryFootnoteLabel });
```
markhiveit, line 78: I don't think this is the right place to be adding a default footnote. My inclination is that this should be added in the backend.

Probably requires a quick discussion to determine what the best place to do this is.

markhiveit (changes requested):

markhiveit (approved):

PR #2404, opened 2021-03-05
EES-1914 Log exceptions before failing imports and handle null error

benoutram:
This PR makes a change to log exceptions before attempting to fail the imports which could in itself fail,
leaving us without any trace of the exception in Application Insights.

It also handles being called with a null errors array or a null error string.


benoutram:
I'm not 100% sure that this will fix the problems we've been having with the importer appearing to be 'stuck' as I couldn’t see how a null error message or a null errors array could be occurring in the first place but it seems sensible to make these changes to rule them out as causes.

luke-h1 (approved):

PR #2412, opened 2021-03-11
EES-1983 Add Created/CreatedBy fields to File and retain existing values when migrating data files

benoutram:
This PR adds new Created and CreatedBy fields to File for auditing/debugging as well as displaying the "Uploaded on" field of data files in the Admin.

It also has a manual endpoint `PATCH http://localhost:5021/api/files/private/data/migrate-created-fields` to retain values of existing data files which it gets from their blobs in the Storage container.

This needs to be run prior to the files migration which changes their paths by moving the files where the Created value on the blob is reset due to the move operation being a copy and delete.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
@@ -224,18 +224,19 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
                                             subjectId: subjectId,
                                             filename: dataFormFile.FileName.ToLower(),
                                             type: FileType.Data,
+                                            createdById: _userService.GetUserId(),
                                             replacingFile: replacingFile);
 
                                         var metaFile = await _releaseDataFileRepository.Create(
                                             releaseId: releaseId,
                                             subjectId: subjectId,
                                             filename: metaFormFile.FileName.ToLower(),
-                                            type: Metadata);
+                                            type: Metadata,
+                                            createdById: _userService.GetUserId());
 
                                         var dataInfo = GetDataFileMetaValues(
                                             name: validSubjectName,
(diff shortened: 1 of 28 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -263,9 +268,9 @@ await ids.Select(id => _releaseFileRepository.CheckFileExists(releaseId, id, Fil
                                             MetaFileId = metaFile.Id,
                                             MetaFileName = metaFile.Filename,
                                             Rows = blob.GetNumberOfRows(),
-                                            UserName = blob.GetUserName(),
+                                            UserName = dataFile.CreatedBy.Email,
```
duncan-at-hiveit, line 271: Is CreatedBy always guaranteed to be not null?  I'm just thinking of the case in the actual migration code where you're having to check if the user's email address from the metadata isn't null.  Also I see mention in other places where the `?` operator has been used after CreatedBy

benoutram: It's used with the `?` elsewhere in case there was no CreatedById set e.g. the blob didn't have a `userName` field (yes there are one or two like that!) or we didn't match the email in the `userName` property with a User in the db.

We can guarantee it's not null here because straight above this the file is created with the Id from the user service and then we do an explicit load of `CreatedBy` to ensure its loaded.

The only way for it to be null here I guess would be if the user wasn't authenticated somehow.

markhiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
duncan-at-hiveit: Cool ta!

PR #2413, opened 2021-03-11
EES-1816 Move Subject.Name to ReleaseFiles.Name

markhiveit:
Making sure there are no large changes required before updating all the tests.


src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/MigrateSubjectNameController.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        [HttpPatch("api/subjects/migrate-subject-names")]
+        public async Task<ActionResult<Unit>> MigrateSubjectNames()
+        {
+            return await _userService.CheckCanRunMigrations()
+                .OnSuccessVoid(async () =>
+                {
+                    await _contentDbContext.ReleaseFiles
+                        .Include(rf => rf.File)
+                        .Where(rf => rf.File.Type == FileType.Data)
```
benoutram, line 45: You could add a check here for ReleaseFiles with no Name to make sure you don't overwrite any that are created after the deploy but before this manual endpoint is run. It seems like it might be possible at the moment that after the deploy a new file upload begins and the subject gets created but now without a name, and then the manual migration is run where `rf.Name = subject.Name` will wipe out the name that's been set by the user.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -452,13 +449,19 @@ private async Task<DataFileInfo> GetFallbackDataFileInfo(Guid releaseId, File da
 
         private async Task<string> GetSubjectName(File file)
```
benoutram, line 450: There's one case

```
Name = dataFile.SubjectId.HasValue ? await GetSubjectName(dataFile) : blob.Name,
```

That can now just be `Name = await GetSubjectName(dataFile)`

and another case

```
Name = zipBlob.Name
```

that can be `Name = await GetSubjectName(dataFile)`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/FileRepository.cs
```diff
@@ -29,5 +30,17 @@ public async Task<File> Get(Guid id)
             return await _contentDbContext.Files
                 .SingleAsync(f => f.Id == id);
         }
+
+        public async Task<string> GetSubjectName(Guid releaseId, Guid subjectId)
```
benoutram, line 34: The `FileRepository` is dedicated to files and there's a separate repository already for `ReleaseFiles`. There's even a `ReleaseDataFileRepository` for the things more specific to data files.

I would add a method to `ReleaseDataFileRepository` like `GetBySubjectId(releaseId, subjectId)` and return the ReleaseFile which is then more useful than returning string, and consume that.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -452,13 +449,19 @@ private async Task<DataFileInfo> GetFallbackDataFileInfo(Guid releaseId, File da
...
-                var subject = await _statisticsDbContext.Subject.FindAsync(file.SubjectId.Value);
-                return subject.Name;
+                throw new ArgumentException("file.Type should equal FileType.Data");
             }
 
-            return "Unknown";
+            var releaseFile = await _contentDbContext
+                .ReleaseFiles
+                .Include(rf => rf.File)
+                .SingleOrDefaultAsync(rf =>
```
benoutram, line 460: This is going to blow up when the file is reused on an amendment because you aren't constraining it by Release Id here. Also once you've got the releaseId to this method you can just do something along the lines of  `(await _releaseFileRepository.Get(releaseId, file.Id)).Name ?? "Unknown" `

src/GovUk.Education.ExploreEducationStatistics.Data.Services/MetaGuidanceSubjectService.cs
```diff
@@ -139,11 +154,11 @@ private async Task<MetaGuidanceSubjectViewModel> BuildSubjectViewModel(ReleaseSu
                 Id = subject.Id,
                 Content = releaseSubject.MetaGuidance ?? "",
                 Filename = subject.Filename ?? "Unknown",
```
benoutram, line 156: Now we have the content database connection and you have added that as a dependency here and fetched the ReleaseFile, we don't need to access `subject.filename`, which means we no longer need to add the filename to Subject when creating a subject and synchronise them across Public/Private with DF.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/MetaGuidanceSubjectService.cs
```diff
@@ -131,6 +136,16 @@ private List<LabelValue> GetVariables(Guid subjectId)
         private async Task<MetaGuidanceSubjectViewModel> BuildSubjectViewModel(ReleaseSubject releaseSubject)
         {
             var subject = releaseSubject.Subject;
+
+            var releaseFileData = await _contentDbContext
+                .ReleaseFiles
+                .Include(rf => rf.File)
+                .SingleOrDefaultAsync(rf =>
```
benoutram, line 143: I think wherever you use `SingleOrDefaultAsync` should be `SingleAsync` because the code would blow up on the next line anyway on `.Name` if no row was found.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/FileRepository.cs
markhiveit: I saw this last night - because I noticed there were several FileRepository type things in ReleaseDataFilesService that could be removed and which could remove ReleaseDataFilesService's dependency on ContentDbContext, and at that point I noticed that this stuff should go in ReleaseFileRepository.

To remove ReleaseDataFileService's dependency on ContentDbContext - I'd need to also move GetAssociatedMetaFile, and create a new method which I called UnlinkReplacementFromOriginal. Worth doing?

benoutram: It sounds like you would be trying to shift more complex logic that belongs in the service into the repository which is typically for CRUD like database operations. Moving that kind of logic would likely make the repository less useful to other consumers especially if extra dependencies need to be added to it.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/MetaGuidanceSubjectService.cs
markhiveit: I've changed the bit you refer to, but I think I'll create a separate ticket to remove Filename from Subject

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/MetaGuidanceSubjectServiceTests.cs
```diff
@@ -205,47 +235,86 @@ public async Task Get_OrderedByName()
...
+                Subject = new Subject(),
                 MetaGuidance = "Subject 2 Meta Guidance"
             };
             var releaseSubject3 = new ReleaseSubject
             {
                 Release = release,
-                Subject = new Subject
-                {
-                    Filename = "file3.csv",
-                    Name = "Subject 3"
```
benoutram, line 231: I think we need a Tidy Up ticket after this PR to drop Subject name.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseDataFileServiceTest.cs
```diff
@@ -845,72 +845,63 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
-
             var blobStorageService = new Mock<IBlobStorageService>(MockBehavior.Strict);
             var dataImportService = new Mock<IDataImportService>(MockBehavior.Strict);
 
             await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContextId))
             await using (var statisticsDbContext = InMemoryStatisticsDbContext(statisticsDbContextId))
             {
                 blobStorageService
-                    .Setup(s => s.CheckBlobExists(PrivateReleaseFiles, dataFile.Path()))
+                    .Setup(s => s.CheckBlobExists(PrivateReleaseFiles, dataReleaseFile.File.Path()))
```
benoutram, line 897: You can use the shortcut `dataReleaseFile.Path()` here

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseDataFileServiceTest.cs
```diff
@@ -845,72 +845,63 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
                     .ReturnsAsync(true);
 
                 blobStorageService
-                    .Setup(s => s.GetBlob(PrivateReleaseFiles, dataFile.Path()))
+                    .Setup(s => s.GetBlob(PrivateReleaseFiles, dataReleaseFile.File.Path()))
```
benoutram, line 901: You can use `dataReleaseFile.Path()` here and all places below throughout this test class

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseDataFileServiceTest.cs
```diff
@@ -1275,50 +1252,49 @@ public async Task GetInfo_MatchingSourceZipBlob()
         {
             var release = new Release();
 
-            var zipFile = new File
+            var zipReleaseFile = new ReleaseFile
```
benoutram, line 1255: This is wrong as ReleaseFile links don't get created for files of type `DataZip` so leave the `zipFile` as a `File` here

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseDataFileServiceTest.cs
```diff
@@ -1275,50 +1252,49 @@ public async Task GetInfo_MatchingSourceZipBlob()
...
                 await contentDbContext.AddRangeAsync(
-                    new ReleaseFile
-                    {
-                        Release = release,
-                        File = dataFile
-                    },
-                    new ReleaseFile
-                    {
-                        Release = release,
-                        File = zipFile
```
benoutram, line 1307: This was actually wrong before just in case your wondering. Zip files don't have a ReleaseFile (and I'm not sure why they need to exist as a File at all!)

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseDataFileServiceTest.cs
```diff
@@ -1673,138 +1638,136 @@ await using (var statisticsDbContext = InMemoryStatisticsDbContext(statisticsDbC
...
+                Release = release1,
+                File = new File
+                {
+                    RootPath = Guid.NewGuid(),
+                    Filename = "ancillary-file.pdf",
+                    Type = Ancillary
+                }
+            };
 
+        var contentDbContextId = Guid.NewGuid().ToString();
```
benoutram, line 1743: Formatting looks out here

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseDataFileServiceTest.cs
```diff
@@ -2116,50 +2067,50 @@ public async Task ListAll_MatchingSourceZipBlob()
         {
             var release = new Release();
 
-            var zipFile = new File
+            var zipReleaseFile = new ReleaseFile
```
benoutram, line 2070: As mentioned before, DataZip won't have a ReleaseFile

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseDataFileServiceTest.cs
```diff
@@ -2432,7 +2383,7 @@ public async Task Upload_Replacing()
             var contentDbContextId = Guid.NewGuid().ToString();
             await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContextId))
             {
-                await contentDbContext .AddAsync(originalDataReleaseFile);
+                await contentDbContext.AddRangeAsync(originalDataReleaseFile);
```
benoutram, line 2386: You don't need this change

benoutram (changes requested):

benoutram:
I think there's still some cases where ReleaseFile.File.Path could be ReleaseFile.Path

benoutram (approved):

PR #2415, opened 2021-03-12
EES-1967 - Add another create data block button

HiveIT-Lewis:


src/explore-education-statistics-admin/src/pages/release/datablocks/ReleaseDataBlockEditPage.module.scss
```diff
@@ -0,0 +1,14 @@
+.dataBlockSelector {
```
markhiveit, line 1: I think it's possible that using existing components/styles will provide the styles required and/or we should be using the govuk-foo type classnames. I think Marv knows what the GDS toolkit provides and what we could potentially use here.

src/explore-education-statistics-common/src/components/form/FormSelect.tsx
```diff
@@ -56,7 +56,7 @@ const FormSelect = ({
   value,
 }: FormSelectProps) => {
   return (
-    <>
+    <div>
```
markhiveit, line 59: Doesn't seem any reason to turn this into a div? Can just leave it as `<>`

src/explore-education-statistics-admin/src/pages/release/datablocks/ReleaseDataBlockEditPage.tsx
```diff
@@ -105,12 +116,28 @@ const ReleaseDataBlockEditPage = ({
...
         <h2>{canUpdateRelease ? 'Edit data block' : 'View data block'}</h2>
 
-        <DataBlockSelector
-          canUpdate={canUpdateRelease}
-          publicationId={publicationId}
-          releaseId={releaseId}
-          dataBlockId={dataBlockId}
-        />
+        <p className="govuk-!-margin-bottom-1">
+          {canUpdateRelease
```
markhiveit, line 120: I might be wrong, but could `DataBlockSelector`'s label prop be used instead of having this separate p tag? So just have `label={canUpdateRelease}`?

Or even remove the label prop completely and just refer to canUpdate to display the text?

markhiveit (changes requested):

markhiveit:
Some frontend unit tests need fixing also

src/explore-education-statistics-admin/src/pages/release/datablocks/ReleaseDataBlockEditPage.tsx
HiveIT-Lewis: I needed this with a separate tag so that the dropdown, "or" and datablock button could all be aligned horizontally and centrally in a straight line. If it has a label it adds extra height to the div.

benoutram:
Seems to be working well

image

I'm just wondering why after creating a data block the user isn't just returned to the Data blocks page where they will now see the new data block and be able to create another already I.e. by hitting the Create Data Block button. That would seem a natural workflow to me rather than leaving them in the edit screen.

It's also possible now that they could make an edit, save, and use the dropdown to swap to another, edit, save, repeat.

By that time, _Create another_ loses a bit of it's meaning if you've edited, saved, change to another,  edited, save, change to another, edited save, etc, when your now several steps beyond creating the first data block.

lauraselby:
I don’t think we could just return them to the manage data blocks page as there isn’t a sensible point to do so? After they save the initial data block they may want to stay on the specific data block page to create a chart.
The original issue was that it wasn’t obvious that you’d need to navigate back to the manage data block page to create another and there wasn’t an obvious button to get back there. I’m happy with the create another button as it stands here, but maybe it should be ‘create **new** data block’ instead? It could also work if the button was just changed to something like ‘back to manage data blocks’ page if that makes you more comfortable.

mrmarvino:
I think we should keep the label as it is -> 'Create another data block' 

We can definitive feedback on what people expect when we do the next round of admin user testing

src/explore-education-statistics-admin/src/pages/release/datablocks/ReleaseDataBlockEditPage.tsx
```diff
@@ -105,12 +116,33 @@ const ReleaseDataBlockEditPage = ({
...
+              ? 'Select a data block to edit'
+              : 'Select a data block to view'
+          }
         />
+        <div className="dfe-flex dfe-align-items--center govuk-!-margin-top-1">
+          <DataBlockSelector
+            canUpdate={canUpdateRelease}
+            publicationId={publicationId}
+            releaseId={releaseId}
+            label={false}
```
markhiveit, line 132: This is the only instance where `DataBlockSelector` is used. So I don't think you need to add this extra `label` prop to `DataBlockSelector`.

The simplest solution would be to remove the label prop from DataBlockSelector and remove it from the `FormSelect` in `DataBlockSelector`. Alternatively, you could move most (if not all) these changes you've made to `ReleaseDataBlockEditPage` into `DataBlockSelector`.

In my mind, that makes sense for the label - so remove the label from the `FormSelect` in `DataBlockSelector` and move `FormLabel` into `DataBlockSelector`. You might still want to keep the `Create another data block` button in `ReleaseDataBlockEditPage`. 

But regardless happy to go along with whatever you think is best. I'm not sure if it's likely that the `DataBlockSelector` component will get used elsewhere and what labels are likely to be appropriate if it ever is used elsewhere. If it's only ever going to be used here, I'd consider even moving the above `h2` into it.

markhiveit (changes requested):

markhiveit (approved):

PR #2436, opened 2021-03-19
EES-1985 Users can only edit draft methodologies

markhiveit:


luke-h1 (approved):

src/explore-education-statistics-admin/src/pages/methodology/edit-methodology/content/MethodologyContentPage.tsx
```diff
@@ -22,11 +22,14 @@ import {
 const MethodologyContentPageInternal = () => {
   const { methodology, canUpdateMethodology } = useMethodologyState();
```
benoutram, line 23: Isn't `canUpdateMethodology` now unused and if so won't need the call on 109 to fetch it?

If that's true though it then this leaves the backend permission unused so we should be removing that too.

However it doesn't make sense to remove any of this if methodology should have separate permissions, only if we are simplifying things to have one permission for updating all content, and I'm not sure we've decided that yet.

PR #2451, opened 2021-04-01
EES-2022  Refactor importer & publisher test suites to use async / await

luke-h1:


luke-h1:
Noticed some problems so have marked as draft

useful-scripts/importer-tests/index.ts
```diff
@@ -0,0 +1,203 @@
...
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${JWT_TOKEN}`,
+      },
+    });
+    console.timeEnd('Create-release-time');
+    const releaseId = res.data.id;
+    console.log(
+      `Release URL: ${API_URL}/publication/${publicationId}/release/${releaseId}/data`,
+    );
+    await FileOps(releaseId);
```
markhiveit, line 72: I'd `return releaseId;` here and call `FileOps()` after `createRelease()` is called

useful-scripts/importer-tests/index.ts
```diff
@@ -0,0 +1,203 @@
...
+  } catch (e) {
+    console.error(e);
+  }
+};
+
+const FileOps = async (releaseId: string) => {
```
markhiveit, line 78: I'd rename `FileOps` to `fileOps` so it named consistently with other functions. I'd also consider renaming it to something more meaningful, like `extractZip`.

useful-scripts/importer-tests/index.ts
```diff
@@ -0,0 +1,203 @@
...
+      url: `${API_URL}/api/release/${releaseId}/data/${subjectId}/import/status`,
+      headers: {
+        Authorization: `Bearer ${JWT_TOKEN}`,
+        'Content-Type': 'application/json',
+      },
+    });
+    if (res.data.status !== 'COMPLETE') {
+      console.log(res.data);
+      await Sleep(300);
+      await getSubjectProgress(releaseId, subjectId);
```
markhiveit, line 157: Rather than making getSubjectProgress recursive, I'd create a loop at the bottom of the file so that it looks something like this

```
var pubId = createPub();
var releaseId = createRelease();
FileOps(releaseId);
ZipDirectory(blah);
var subId = addSubject(releaseId);
var subComplete = false;
while(!subComplete) { subComplete = getSubjectProgress(releaseId, subId); }
```

So getSubjectProgress can just return the progress itself.

useful-scripts/importer-tests/index.ts
```diff
@@ -0,0 +1,203 @@
...
+  );
+
+  // rename the meta file...
+  const metaGlob = await globby('./test-files/*.meta.csv');
+  const metaGlob2 = metaGlob[0];
+
+  fs.rename(`${metaGlob2}`, `./test-files/testfile.meta.${randomId}.csv`, e =>
+    e ? console.error(e) : '',
+  );
+  await ZipDirectory('./test-files', `./zip-files/clean-test-zip-${v4()}.zip`);
```
markhiveit, line 106: I'd also pull this out.

useful-scripts/publisher-tests/index.ts
```diff
@@ -0,0 +1,340 @@
...
+import FormData from 'form-data';
+import Sleep from './utils/Sleep';
+import { ReleaseDataProps, SubjectArrayT } from 'types';
+
+// disable insecure warnings
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
+dotenv.config();
+const { JWT_TOKEN, API_URL, TOPIC_ID } = process.env;
+
+const createPublication = async () => {
```
markhiveit, line 20: This code is mostly duplicated from the importer tests? I'd pull out common functionality from the files and then importing them into both publish and importer tests. You could put both the importer-tests and publisher-tests in the same load test directory.

useful-scripts/publisher-tests/index.ts
```diff
@@ -0,0 +1,340 @@
+/* eslint-disable */
```
markhiveit, line 1: Many of the comments I've left for the importer-tests file also apply to this file.

markhiveit (changes requested):

useful-scripts/importer-publisher-tests/index.ts
```diff
@@ -0,0 +1,365 @@
...
+  if (!fs.existsSync('./test-files')) {
+    fs.mkdirSync('./test-files');
+  }
+
+  if (!fs.existsSync('./zip-files')) {
+    fs.mkdirSync('./zip-files');
+  }
+
+  const staleZipFiles = await globby('*.zip');
+  if (staleZipFiles[1]) {
```
markhiveit, line 317: This checks that if there is a second zip file it removes it? But if globby found three zip files, then it would only remove one of them I believe? To prevent this problem you could put the zips the script creates into a separate directory so you can delete everything in that directory instead.

I think rather than rely on globby as much, it might be worth considering passing the zip file name/path to the script as an argument.

useful-scripts/importer-publisher-tests/index.ts
```diff
@@ -0,0 +1,365 @@
...
+
+  fs.rename(`${csvGlob2}`, `./test-files/testfile-${randomId}.csv`, e =>
+    e ? console.error(e) : '',
+  );
+
+  // rename the meta file...
+  const metaGlob = await globby('./test-files/*.meta.csv');
+  const metaGlob2 = metaGlob[0];
+
+  fs.rename(`${metaGlob2}`, `./test-files/testfile.meta.${randomId}.csv`, e =>
```
markhiveit, line 104: I think it would be better to rename it `testfile-${randomId}.meta.csv` - to keep it consistent with how the data csv is named.

markhiveit (changes requested):

useful-scripts/importer-publisher-tests/utils/ZipDirectory.ts
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable */
+
+import archiver from 'archiver';
+import fs from 'fs';
+
+/**
+ *
```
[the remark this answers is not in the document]
luke-h1: the eslint disable ?

useful-scripts/importer-publisher-tests/index.ts
luke-h1: There should never be an instance where there are 3 zip files in the project directory. Before sending the zip file to the API (with renamed data & meta files) it will move it into the root of the project (in order for the compiled index.js to pick it up). When you re run the test again the array returned from globby will look something like: 
```
[
'archive.zip', 
'clean-test-zip-<random_id>.zip'
]
```

So the test can safely assume that what is in staleZipFiles[1] is a stale zip file

useful-scripts/importer-publisher-tests/multiple-import-publication.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+# Usage: 
+# ./multiple-import-publication.sh -n <num_of_imports> <importer or publisher>
+# -n = number of subjects to upload 
+# last option = test the importer or the publisher
+
+# examples: 
+# ./multiple-importer-publication.sh -n 5 importer 
+# ./multiple-importer-publication.sh -n 2 Spublisher
```
markhiveit, line 10: Typo here `Spublisher`

markhiveit (approved):

PR #2457, opened 2021-04-01
EES-1815 Retrieve file names from releaseFile.Name where possible

markhiveit:
Now that we have releaseFile.Name, we don't need to store subject names
in blob file metadata and can retrieve subject names from releaseFile.Name instead.

This currently is only happening with data files, not ancillary files.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Extensions/FileExtensionTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Extensions/FileExtensionTests.cs
@@ -9,6 +9,14 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Model.Tests.Extensi
 {
     public class FileExtensionTests
     {
+        private readonly File _file = new File
+        {
+            Id = Guid.NewGuid(),
+            RootPath = Guid.NewGuid(),
+            Filename = "ancillary.pdf",
+            Type = Ancillary
+        };
+
         [Fact]
         public void BatchesPath()
         {
@@ -169,6 +177,19 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Model.Tests.Extensi
                     Assert.Throws<ArgumentOutOfRangeException>(() => file.PublicPath(release));
                 }
             });
+        }
+
+        [Fact]
+        public void ToFileInfoNotFound()
+        {
+            var result = _file.ToFileInfoNotFound();
+
+            Assert.Equal(_file.Id, result.Id);
+            Assert.Equal("pdf", result.Extension);
+            Assert.Equal("ancillary.pdf", result.FileName);
+            Assert.Equal("Unknown", result.Name);
+            Assert.Equal("0.00 B", result.Size);
+            Assert.Equal(Ancillary, result.Type);
         }
     }
 }
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Extensions/FileExtensionTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Extensions/FileExtensionTests.cs
@@ -14,12 +14,16 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Model.Tests.Extensi
         private const string PublicationSlug = "publication-slug";
         private const string ReleaseSlug = "release-slug";
 
-        private readonly File _file = new File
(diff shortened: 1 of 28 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Model/Extensions/FileExtensions.cs
```diff
@@ -52,17 +52,38 @@ public static string PublicPath(this File file, Guid releaseId)
             return $"{FilesPath(releaseId, file.Type)}{file.Id}";
         }
 
-        public static FileInfo ToFileInfo(this File file, BlobInfo blobInfo)
+        public static FileInfo ToFileInfo(this ReleaseFile rf, BlobInfo blobInfo)
         {
+            var name = "Unknown";
```
benoutram, line 57: It should never be "Unknown", but see suggestion below

benoutram (changes requested):

amyb-hiveit (approved):

benoutram:
There's a TODO in FileExtensionTests.cs that's been missed, but I think it probably needs to remain if we still rely on the Name field on blobs for Ancillary files? Maybe that comment should be updated to explain why we still set the name on FileInfo from blob properties.

image

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Extensions/FileExtensions.cs
```diff
@@ -52,42 +52,39 @@ public static string PublicPath(this File file, Guid releaseId)
             return $"{FilesPath(releaseId, file.Type)}{file.Id}";
         }
 
-        public static FileInfo ToFileInfo(this File file, BlobInfo blobInfo)
+        public static FileInfo ToFileInfo(this ReleaseFile releaseFile, BlobInfo blobInfo)
```
benoutram, line 55: I think this should be in ReleaseFileExtensions now which already exists.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Extensions/FileExtensions.cs
```diff
@@ -52,42 +52,39 @@ public static string PublicPath(this File file, Guid releaseId)
...
+                Name = releaseFile.Name ?? 
+                       (blobInfo.Name.IsNullOrEmpty() ? releaseFile.File.Filename : blobInfo.Name),
                 Size = blobInfo.Size,
-                Type = file.Type
+                Type = releaseFile.File.Type
             };
         }
 
-        public static FileInfo ToFileInfoNotFound(this File file)
+        public static FileInfo ToFileInfo(this MethodologyFile methodologyFile, BlobInfo blobInfo)
```
benoutram, line 68: There's a MethodologyFileExtensions which this can go in.

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Extensions/FileExtensionTests.cs
```diff
@@ -14,12 +14,16 @@ public class FileExtensionTests
         private const string PublicationSlug = "publication-slug";
         private const string ReleaseSlug = "release-slug";
 
-        private readonly File _file = new File
+        private readonly ReleaseFile _releaseFile = new ReleaseFile
```
benoutram, line 17: Everything for ReleaseFile to do with testing GetFileInfo can go in ReleaseFileExtensionTests

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Extensions/FileExtensionTests.cs
```diff
@@ -225,60 +228,25 @@ public void ToFileInfo_BlobWithNoNameKey()
...
 
-            Assert.Equal(_file.Id, result.Id);
+            Assert.Equal(_releaseFile.FileId, result.Id);
             Assert.Equal("pdf", result.Extension);
             Assert.Equal("ancillary.pdf", result.FileName);
             Assert.Equal("ancillary.pdf", result.Name);
-            Assert.Equal(_file.Path(), result.Path);
             Assert.Equal("400 B", result.Size);
             Assert.Equal(Ancillary, result.Type);
         }
```
benoutram, line 238: I don't think anythings been added to test `ToFileInfo(this MethodologyFile)`. That should be added to MethodologyFileExtensionTests

benoutram (approved):

PR #2470, opened 2021-04-12
EES-2040 Reordering Key Statistics

amyb-hiveit:
Adds functionality to allow Key Statistics to be reordered.
ks1
ks2

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/content/components/KeyStatistics.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/content/components/KeyStatistics.tsx
@@ -2,10 +2,14 @@ import EditableKeyStat from '@admin/components/editable/EditableKeyStat';
 import KeyStatSelectForm from '@admin/pages/release/content/components/KeyStatSelectForm';
 import useReleaseContentActions from '@admin/pages/release/content/contexts/useReleaseContentActions';
 import { EditableRelease } from '@admin/services/releaseContentService';
+import BlockDraggable from '@admin/components/editable/BlockDraggable';
+import BlockDroppable from '@admin/components/editable/BlockDroppable';
 import Button from '@common/components/Button';
 import WarningMessage from '@common/components/WarningMessage';
 import { KeyStatContainer } from '@common/modules/find-statistics/components/KeyStat';
+import useToggle from '@common/hooks/useToggle';
 import React, { useCallback, useState } from 'react';
+import { DragDropContext, DropResult } from 'react-beautiful-dnd';
 
 export interface KeyStatisticsProps {
   release: EditableRelease;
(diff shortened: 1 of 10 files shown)


src/explore-education-statistics-admin/src/pages/release/content/components/KeyStatistics.tsx
```diff
@@ -29,20 +101,23 @@ const KeyStatistics = ({ release, isEditing }: KeyStatisticsProps) => {
...
+            <AddKeyStatistics release={release} />
+            {keyStatisticsBlocks.length > 1 && <ReorderKeyStatisticsButton />}
+          </ButtonGroup>
         </>
       )}
-      <KeyStatContainer>
-        {release.keyStatisticsSection.content
-          .filter(block => block.type === 'DataBlock')
-          .map(block => (
+      {!isReordering && (
```
ntsim, line 110: Is it possible to combine the reordering and non-reordering states into a single declaration? 

Ideally it'd be good to have something like `EditableAccordion` where we only declare things once and just toggle the reordering state via props. Too much duplication like this can easily lead to bugs in the future.

src/explore-education-statistics-admin/src/pages/release/content/components/KeyStatistics.tsx
```diff
@@ -18,6 +26,70 @@ const KeyStatistics = ({ release, isEditing }: KeyStatisticsProps) => {
...
+    );
+  }, [release]);
+
+  const [isReordering, toggleReordering] = useToggle(false);
+
+  const { updateSectionBlockOrder } = useReleaseContentActions();
+
+  const ReorderKeyStatisticsButton = () => {
+    return !isReordering ? (
+      <Button variant="secondary" onClick={() => toggleReordering.on()}>
```
ntsim, line 49: Can make this a little more succint by:

```tsx
onClick={toggleReordering.on}
```

src/explore-education-statistics-admin/src/pages/release/content/components/KeyStatistics.tsx
amyb-hiveit: Good call. I've combined them, it was a little bit fiddly because of the flexbox so I've moved the styles to a common file to avoid having to duplicate them and not break the other places where the KeyStatColumn is used.

markhiveit (approved):

src/explore-education-statistics-common/src/modules/find-statistics/components/KeyStat.tsx
```diff
@@ -22,7 +23,7 @@ interface KeyStatColumnProps {
 
 export const KeyStatColumn = ({ children, testId }: KeyStatColumnProps) => {
   return (
-    <div className={styles.column} data-testid={testId}>
+    <div className='dfe-keyStatistics-column' data-testid={testId}>
```
ntsim, line 26: Rather than put this in `_content.scss`, couldn't we it in from `KeyStatistics.module.scss` as a `className` prop?

amyb-hiveit: Yep, don't know why I didn't do that!

ntsim: :smile: looks good now!

ntsim (approved):

PR #2474, opened 2021-04-14
Block content removed from content blocks

HiveIT-Lewis:


src/explore-education-statistics-admin/src/pages/methodology/edit-methodology/content/components/MethodologyEditableBlock.tsx
```diff
@@ -53,7 +53,7 @@ const MethodologyEditableBlock = ({
         <EditableContentBlock
           editable={editable && !isBrowser('IE')}
           id={blockId}
-          label="Block content"
+          label=""
```
amyb-hiveit, line 56: Could the label be made optional on EditableContentBlock instead of sending and empty string?

amyb-hiveit (changes requested):
Giving it an empty string means it renders an empty span which isn't ideal. What do you think to either:

- making the label optional, or 
- pass an option so can be visually hidden?

src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseEditableBlock.tsx
```diff
@@ -71,6 +71,7 @@ const ReleaseEditableBlock = ({
           editable={editable && !isBrowser('IE')}
           id={blockId}
           label=""
+          hideLabel
```
amyb-hiveit, line 74: Good to add this prop :+1: 
Can you put the label text back as this will still be rendering an empty span if you give it an empty string and the label might be useful for accessibility?

markhiveit (approved):

PR #2475, opened 2021-04-14
EES-1858 Show loader when click featured table

amyb-hiveit:
Show a loader during the short delay when a featured table is clicked.
loader

### DIFF
--- a/src/explore-education-statistics-common/src/modules/table-tool/components/SubjectStep.tsx
+++ b/src/explore-education-statistics-common/src/modules/table-tool/components/SubjectStep.tsx
@@ -20,7 +20,7 @@ const subjectTabIds = {
 
 interface Props {
   highlights?: TableHighlight[];
-  loadingFastTrack: boolean;
+  loadingFastTrack?: boolean;
   subjects: Subject[];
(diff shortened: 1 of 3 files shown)


markhiveit (changes requested):
I think we can also add this to PreReleaseTableToolPage's TableToolWizard.

Just FYI - shouldn't need to add it the TableToolWizard on ReleaseTableToolPage (for data block creation in Admin) as featured tables aren't shown there.

amyb-hiveit:
> I think we can also add this to PreReleaseTableToolPage's TableToolWizard.
> 
> Just FYI - shouldn't need to add it the TableToolWizard on ReleaseTableToolPage (for data block creation in Admin) as featured tables aren't shown there.

The PreReleaseTableToolPage already has a loader that shows when you click on a featured table, it shows a loader instead of the whole form instead of just in the relevant section as I've done on TableToolPage.

I had a quick look to see if it'd be easy to change PreReleaseTableToolPage to behave the same way as TableToolPage but the way the pages are set up is a bit different so think it'd need some work.

markhiveit (approved):

PR #2478, opened 2021-04-15
EES-2060 Filter out legend datasets when saving datablocks where the dataset is now orphaned.

benoutram:
We were previously filtering out datasets from the chart major axis configuration where the dataset was orphaned, but not filtering legend items where their dataset might also be orphaned.

This was causing a bug rendering the charts if a user alters a datablock query of a datablock containing a chart and removes a filter or indicator that's used by a legend item.

This PR alters the function used to filter datasets to be a predicate which can be used to filter both datasets from the chart major axis configuration and also the chart legend items.


markhiveit (approved):

src/explore-education-statistics-common/src/modules/charts/util/__tests__/isOrphanedDataSet.test.ts
```diff
@@ -0,0 +1,231 @@
...
+        level: 'localAuthority',
+      },
+    };
+
+    expect(isOrphanedDataSet(testDataSet, testFullTable.subjectMeta)).toBe(
+      false,
+    );
+  });
+
+  test('is true for a data set where the location does not match on `value`', () => {
```
luke-h1, line 118: Did you mean to put `${value}` here instead of `value`  as string surrounded by backticks ?

src/explore-education-statistics-common/src/modules/charts/util/__tests__/isOrphanedDataSet.test.ts
```diff
@@ -0,0 +1,231 @@
...
+      },
+    };
+
+    expect(isOrphanedDataSet(testDataSet, testFullTable.subjectMeta)).toBe(
+      true,
+    );
+  });
+
+  test('is true for a data set where the location does not match on `level`', () => {
+    const testDataSet: DataSet = {
```
luke-h1, line 134: Just wondering again if you meant to put `${level}` instead of level with backticks ? If you didn't then mean to I think it's safe to remove the backticks

benoutram: There's no interpolation intended here, it's just referring to the names of the `value` and `level` fields by name. I've removed the backticks to avoid confusion. Thanks.

luke-h1 (approved):

PR #2479, opened 2021-04-15
Ees 2069

luke-h1:


luke-h1:
I think this has bumped other dependencies when fixing the security vulnerability. To be on the safe side I'm going to revert the security dependency fix

PR #2488, opened 2021-04-19
EES-2052 Ancillary file names should be accessed from ReleaseFiles

markhiveit:


src/GovUk.Education.ExploreEducationStatistics.Content.Model/Extensions/ReleaseFileExtensions.cs
```diff
@@ -26,8 +26,9 @@ public static FileInfo ToFileInfo(this ReleaseFile releaseFile, BlobInfo blobInf
             {
                 Id = releaseFile.FileId,
                 FileName = releaseFile.File.Filename,
-                Name = releaseFile.Name ??
-                       (blobInfo.Name.IsNullOrEmpty() ? releaseFile.File.Filename : blobInfo.Name),
+                Name = releaseFile.Name.IsNullOrEmpty()
```
benoutram, line 29: I think I would change the migration to set `null` if no blob property exists, which then makes checking for inconsistencies a bit more straightforward, and just make this `Name = releaseFile.Name ?? releaseFile.File.Filename`. At the moment we are saying it's ok for some files to have an empty string as the name, which shouldn't be allowed.

benoutram (changes requested):

benoutram (approved):

PR #2497, opened 2021-04-21
EES-1512 Add ability to rename Data and Ancillary files

markhiveit:
On the backend, this PR adds two endpoints, one to rename ReleaseFiles.Name for Data or Ancillary files. The other gets a single Ancillary file's FileInfo.

On the frontend, this adds two pages for editing data file and ancillary file details, and links to those pages.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseFileServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseFileServiceTests.cs
@@ -1498,6 +1498,39 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
         }
 
         [Fact]
+        public async Task UpdateName()
+        {
+            var releaseFile = new ReleaseFile
+            {
+                Release = new Release(),
+                Name = "Test PDF File",
+                File = new File()
+            };
+
+            var contentDbContextId = Guid.NewGuid().ToString();
+            await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContextId))
+            {
+                await contentDbContext.AddAsync(releaseFile);
+                await contentDbContext.SaveChangesAsync();
+            }
+
+            await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContextId))
+            {
+                var service = SetupReleaseFileService(contentDbContext: contentDbContext);
+
+                var result = await service.UpdateName(releaseFile.ReleaseId, releaseFile.FileId, "New file title");
+
+                Assert.True(result.IsRight);
+                Assert.IsType<Unit>(result.Right);
+
+                var updatedReleaseFile = await contentDbContext.ReleaseFiles.FirstAsync(rf =>
+                    rf.ReleaseId == releaseFile.ReleaseId
+                    && rf.FileId == releaseFile.FileId);
+                Assert.Equal("New file title", updatedReleaseFile.Name);
+            }
+        }
+
+        [Fact]
         public async Task UploadAncillary()
         {
             const string filename = "ancillary.pdf";
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseFileServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseFileServiceTests.cs
@@ -14,6 +14,7 @@ using GovUk.Education.ExploreEducationStatistics.Common.Utils;
 using GovUk.Education.ExploreEducationStatistics.Content.Model;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Extensions;
+using Microsoft.AspNetCore.Mvc;
 using Microsoft.EntityFrameworkCore;
 using Moq;
 using Xunit;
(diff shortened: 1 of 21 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseFileService.cs
```diff
@@ -152,6 +170,18 @@ public class ReleaseFileService : IReleaseFileService
                 });
         }
 
+        public Task<Either<ActionResult, Unit>> UpdateName(Guid releaseId, Guid fileId, string name)
+        {
+            return _persistenceHelper
+                .CheckEntityExists<Release>(releaseId)
+                .OnSuccess(_userService.CheckCanUpdateRelease)
+                .OnSuccess(async () =>
```
benoutram, line 178: `OnSuccessVoid` exists for this so that you don't have to return `Unit`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseFileService.cs
```diff
@@ -152,6 +170,18 @@ public class ReleaseFileService : IReleaseFileService
                 });
         }
 
+        public Task<Either<ActionResult, Unit>> UpdateName(Guid releaseId, Guid fileId, string name)
```
benoutram, line 173: You should add a test for this to `ReleaseFileServiceTests`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseFileService.cs
```diff
@@ -131,6 +132,23 @@ public class ReleaseFileService : IReleaseFileService
                 });
         }
 
+        public async Task<Either<ActionResult, FileInfo>> GetFile(Guid releaseId, Guid fileId)
```
benoutram, line 135: You should add a tests for this to `ReleaseFileServiceTests` such as both file and blob exist, file not exists, blob not exists.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleaseFileController.cs
```diff
@@ -60,6 +69,14 @@ public async Task<ActionResult<IEnumerable<FileInfo>>> GetAncillaryFiles(Guid re
                 .HandleFailuresOrOk();
         }
 
+        [HttpGet("release/{releaseId}/ancillary/{fileId}")]
+        public async Task<ActionResult<FileInfo>> GetFile(Guid releaseId, Guid fileId)
```
benoutram, line 73: I think there's a bit of a naming problem here although not sure what to suggest as the solution.

The route here is `/ancillary/` implying that it can only get ancillary files but the name is `GetFile` and the method is generic so sure enough, supplying the id of any file will return it.

The equivalent for getting this file 'info' for data files includes the term `FileInfo` like `GetDataFileInfo` in  the controller and `GetInfo` in the service.

I see that you can't use `/file/` in the route because it would conflict with the `Stream` method.

Maybe we need to think about the naming at a wider level.

markhiveit: Talked to Ben about this - we decided to rename the existing `release/{releaseId}/file/{id}` endpoint to stream files to `release/{releaseId}/file/{id}/download`, allowing the new endpoint to use the original.

src/explore-education-statistics-admin/src/pages/release/data/ReleaseAncillaryFilePage.tsx
```diff
@@ -0,0 +1,88 @@
+import {
+  ReleaseAncillaryFileRouteParams,
+  ReleaseRouteParams,
+  releaseDataAncillaryRoute,
+  releaseAncillaryFileRoute,
```
luke-h1, line 5: Don't think this is being used anywhere in this file ?

src/explore-education-statistics-admin/src/pages/release/data/ReleaseAncillaryFilePage.tsx
```diff
@@ -0,0 +1,88 @@
...
+                    releaseDataAncillaryRoute.path,
+                    {
+                      publicationId,
+                      releaseId,
+                    },
+                  ),
+                );
+              }}
+            >
+              {form => {
```
luke-h1, line 72: Don't think you need the return here. I think you should just be able to do: 
``` 
{form => (
                  <Form {...form} id="edit-data-file-form">
                    <FormFieldTextInput id="title" label="Title" name="title" />
                    <Button type="submit">Save changes</Button>
                  </Form>
)}

File: src/explore-education-statistics-admin/src/pages/release/data/ReleaseDataFilePage.tsx [hunk 7]
```diff
@@ -105,149 +44,39 @@ const ReleaseDataFilePage = ({
...
+                  .then();
+                history.push(
+                  generatePath<ReleaseRouteParams>(releaseDataRoute.path, {
+                    publicationId,
+                    releaseId,
+                  }),
+                );
+              }}
+            >
+              {form => {
```
luke-h1, line 68: Similar to the other comment I think you don't need the extra return here

luke-h1 (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseFileServiceTests.cs
```diff
@@ -1171,6 +1172,125 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+            await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContextId))
+            {
+                await contentDbContext.AddAsync(releaseFile);
+                await contentDbContext.SaveChangesAsync();
+            }
+
+            var blobStorageService = new Mock<IBlobStorageService>(MockBehavior.Strict);
+            var blob = new BlobInfo(
+                path: null,
+                size: "93",
```
benoutram, line 1200: Doesn't really mean anything but I think due to the way it's set it will always have some sort of size suffix on it like "800 B" or "93 Kb"

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseFileServiceTests.cs
```diff
@@ -1171,6 +1172,125 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+        public async Task GetFile_NoReleaseFile()
+        {
+            var blobStorageService = new Mock<IBlobStorageService>(MockBehavior.Strict);
+
+            await using (var contentDbContext = InMemoryApplicationDbContext(Guid.NewGuid().ToString()))
+            {
+                var service = SetupReleaseFileService(contentDbContext: contentDbContext,
+                    blobStorageService: blobStorageService.Object);
+
+                var result = await service.GetFile(Guid.Empty, Guid.Empty);
```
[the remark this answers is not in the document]
markhiveit: Discussed on Slack - switching them to Guid.NewGuid()'s

benoutram (approved):

PR #2500, opened 2021-04-23
EES-2087 - invalid data blocks

duncan-at-hiveit:
This PR:
- flags a data block as invalid if a whole new filter is introduced in a replacement subject
- marks scenarios as fixable / unfixable depending upon whether or not the user can edit and fix a data block with the current screens available, and only presents the "Edit Data Block" in the Replacement screen if it's fixable
- fixes a bug in the Replacement screen whereby deleting the last invalid items would not then show the "Confirm replacement" button

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReplacementServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReplacementServiceTests.cs
@@ -1511,8 +1511,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                 Label = "Default group - not changing",
                 FilterItems = new List<FilterItem>
                 {
-                    replacementDefaultFilterItem,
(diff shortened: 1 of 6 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReplacementServiceTests.cs
```diff
@@ -873,6 +874,861 @@ await using (var statisticsDbContext = InMemoryStatisticsDbContext(statisticsDbC
...
+                }
+            };
+
+            var replacementDefaultFilterGroup = new FilterGroup
+            {
+                Label = "Default group - not changing",
+                FilterItems = new List<FilterItem>
+                {
+                    replacementDefaultFilterItem,
+                    replacementNewlyIntroducedFiltersFilterItem
```
benoutram, line 1515: Should this be here @duncan-at-hiveit as I think you might have filter items belonging to two different filter groups. See also line 1524.

benoutram (approved):

PR #2504, opened 2021-04-27
EES-2103 - Fix UI tests

luke-h1:


tests/robot-tests/tests/admin_and_public/bau/footnotes.robot
```diff
@@ -29,6 +29,7 @@ Create new publication and release via API
 
 Upload subject
     [Tags]  HappyPath
+    user waits for page to finish loading
```
markhiveit, line 32: It's best not to resort to this keyword if you don't need to and instead wait on something that should appear on the page.

The `user waits for page to finish loading` keyword got created because of a particular scenario where there was no element on the page I could wait on - I think it was something like a radio button that I had waited on already, but it still hadn't loaded the react javascript around it. So I was forced to add a sleep.

All that said I notice this keyword has got used all over the tests now - either way better to use something else if possible.

tests/robot-tests/tests/admin_and_public/bau/prerelease_visibility.robot
```diff
@@ -27,7 +27,8 @@ Verify release summary
 Upload subject
     [Tags]  HappyPath
     user clicks link  Data and files
-    user waits until h2 is visible  Add data file to release
+    user waits for page to finish loading
```
markhiveit, line 30: This shouldn't be necessary here, as we have a wait immediately after it.

tests/robot-tests/tests/general_public/release_page_absence.robot
```diff
@@ -110,7 +110,7 @@ Validate headlines -- Summary tab content
 
 Validate Key Statistics data block -- Charts tab
     [Tags]  HappyPath
-    user scrolls to element  css:#releaseHeadlines-chart 
+    user scrolls down  400
```
markhiveit, line 113: I wouldn't think this would be better than scrolling to a specific element. Why change this?

markhiveit (changes requested):

tests/robot-tests/tests/general_public/release_page_absence.robot
luke-h1: For some reason when Selenium attempted to scroll to that element it caused the below error on the Azure pipeline: 
```
ElementNotInteractableException: Message: element not interactable: [object HTMLElement] has no size and location
```
Using the `user scrolls down` keyword seems to prevent this

tests/robot-tests/tests/admin_and_public/bau/footnotes.robot
luke-h1: Without this keyword the pipeline seemed to consistently fail on: `Element 'xpath://*[contains(@class,"govuk-accordion__section-button") and text()="UI test subject"]' did not appear in 30 seconds.`

tests/robot-tests/tests/admin_and_public/bau/prerelease_visibility.robot
luke-h1: Without this keyword the pipeline seemed to consistently fail on: Element 'xpath://*[contains(@class,"govuk-accordion__section-button") and text()="UI test subject"]' did not appear in 30 seconds.

tests/robot-tests/tests/general_public/release_page_absence.robot
markhiveit: I've investigated this. It was failing on the `user scrolls to element` because `releaseHeadlines-chart` element wasn't visible because the Table tab was currently selected. So I've updated this to `user scrolls to element  css:#releaseHeadlines-charts-tab`

tests/robot-tests/tests/admin_and_public/bau/prerelease_visibility.robot
```diff
@@ -27,7 +27,8 @@ Verify release summary
 Upload subject
     [Tags]  HappyPath
     user clicks link  Data and files
-    user waits until h2 is visible  Add data file to release
+    user waits until h2 is visible  Add data file to release  90
```
markhiveit, line 30: I'm going to remove the 90 wait here too. The new wait below should do the trick.

Adding extra waits should be a tactic of last resort as it lengthens the overall run time, particularly when there are failures.

tests/robot-tests/tests/admin_and_public/bau/publish_release_and_amend_2.robot
```diff
@@ -373,10 +373,13 @@ Edit methodology
     user clicks link  manage methodologies
     user clicks element  //*[@id="approved-methodologies-tab"]
     user clicks link  ${METHODOLOGY_NAME}
+    user clicks link  Sign off
```
markhiveit, line 376: With these new additions to this test case, I've split this into three separate test cases instead - just so if there is a failure when changing the methodology status to Draft or Approved it's a bit clearer when looking at the test logs

tests/robot-tests/tests/admin_and_public/bau/prerelease_visibility.robot
```diff
@@ -28,13 +28,15 @@ Upload subject
...
+    user waits until page contains element  id:dataFileUploadForm-subjectTitle
     user enters text into element  id:dataFileUploadForm-subjectTitle   UI test subject
     user chooses file   id:dataFileUploadForm-dataFile       ${FILES_DIR}upload-file-test.csv
     user chooses file   id:dataFileUploadForm-metadataFile   ${FILES_DIR}upload-file-test.meta.csv
     user clicks button  Upload data files
 
     user waits until h2 is visible  Uploaded data files
-    user waits until page contains accordion section   UI test subject
+    user waits until element contains  id:dataFileUploadForm-subjectTitle   ${EMPTY}
+    user waits until page contains accordion section   UI test subject  60
```
markhiveit, line 39: I investigated the failure on this line - it was quite hard to pin down. Looking at the screenshot of the failure from the pipeline, it was a "Sorry, there is a problem with the service" page. I tried to mess with importing subjects manually to see if I could recreate the error, but couldn't, so I assume it was happening due to the speed of the UI tests, and so I've not submitted a bug for the issue.

When trying to pin down what was happening, I tried adding they "capture large screenshot" keyword to see at what point the error was occurring, but then predictably the test case started passing.

I speculatively tried waiting on a few things - eventually trying waiting on the Subject name field being empty - and the test case now seems to pass. So it is solved but still a bit of a mystery exactly what happened here.

luke-h1: hmm that's interesting. I didn't realise screenshots were available to view on the pipeline. That'll make debugging a lot easier in the future.

luke-h1:
Currently the UI tests are passing after pulling the latest changes and fixing a few other failures. I've had to use the `Sleep` keyword in order to make a few flaky UI tests pass. This isn't ideal as it extends the run time of the UI test run. Before I request another review I'm going to debug and see if there is a better way to fix these intermittent failures.

PR #2508, opened 2021-04-27
EES-1236 remove validation preventing chart preview update

amyb-hiveit:
Removes validation that prevents the chart preview from updating unless the required fields (title and alt) are filled.

I couldn't think of a reason the validation needed to there for the onChange event and have checked it hasn't changed how the validation works when the form is submitted or the required fields changed. Let me know if there's something I'm missing that requires it.


markhiveit (approved):

amyb-hiveit:
@luke-h1 when you test this please check that it doesn't accidentally allow you to submit any invalid chart configuration, I don't think it does, but the validation might have been there for a reason I've missed. Thanks!

PR #2510, opened 2021-04-28
EES-2059 increase cookie banner cookie expiry

amyb-hiveit:
Increases the banner seen cookie length to 1 year, this is still shorter than the analytics one but 10 years seemed to long and this makes it consistent with the gov.uk equivalent (see https://www.gov.uk/help/cookie-details).

Note for testing: this will also update the info on https://explore-education-statistics.service.gov.uk/cookies/details


markhiveit (approved):

markhiveit:
Could shorten the expiry for ees_disable_google_analytic cookies so that it is also a year - but don't think its a big deal so approved.

PR #2515, opened 2021-05-05
Ees 2101 allow users to explore older release data

duncan-at-hiveit:
This PR:
- adds Release ID and Release Slug as addition arguments to the Table Tool page when accessed via a Fast Track link, allowing that particular Release's Subjects to be explored
- corrects "View the release for this data" link to take you back to the specific Release that they were from, rather than the latest release
- adds a message stating "This is the latest data" if the user has created a table based on the latest data
- adds a message stating "This is NOT the latest data" if the user has created a table based on the older Release data
- adds in some high-level tests for Table Tool page

Previously, the user could navigate to the Table Tool page on the public site by finding a data block and choosing "Explore this data".  When they were presented with the prepopulated Table Tool page and the resultant table, if they then attempted to change some of the Table Tool criteria of previous steps e.g. the Filter selection, the service would throw a 500.

This is because the ability to change Table Tool criteria was always assuming that the Subject being explored was from the latest Release.

This PR changes this by providing the Release ID of the Subject being explored (or rather, the Release ID that is stored in the saved Fast Track that the prepopulated Table Tool page is driven from).

We also now capture from the Fast Track call if this is data from the latest Release, and if not, present the user with a way to navigate to the latest Release to explore the data there.

image

image


src/GovUk.Education.ExploreEducationStatistics.Data.Model/Services/AbstractRepository.cs
```diff
@@ -5,9 +5,12 @@
 using System.Linq.Expressions;
 using System.Threading.Tasks;
 using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Common.Model;
 using GovUk.Education.ExploreEducationStatistics.Data.Model.Database;
 using GovUk.Education.ExploreEducationStatistics.Data.Model.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Data.Model.Services.Interfaces;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Azure.Documents.SystemFunctions;
```
benoutram, line 13: I think this might be a mistake because I can't see what would be using it in the changes below

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Either.cs
```diff
@@ -174,12 +173,27 @@ public static class EitherTaskExtensions
...
 
-        public static async Task<Either<Tl, T>> OnSuccess<Tl, Tr, T>(this Task<Either<Tl, Tr>> task, Func<Tr, Task<T>> func)
+        /**
+         * Allows 2 OnSuccess(...) calls to be chained and the next OnSuccess() to receive a Tuple containing both of
+         * the results.  If either OnSuccess(...) fails, the entire result fails and additionally, if the first result
+         * fails, the second OnSuccess(...) will not be called.
+         *
+         * When C# allows better destructuring, we will be able to destructure the resulting Tuple much better. 
+         */
+        public static async Task<Either<TFailure, Tuple<TSuccess1, TSuccess2>>> OnSuccessCombineWith<TFailure, TSuccess1, TSuccess2>(this Task<Either<TFailure, TSuccess1>> task, Func<TSuccess1, Task<Either<TFailure, TSuccess2>>> func)
```
benoutram, line 188: Cool, there's probably lots of places we could start using this! 🥇

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/AssertExtensions.cs
```diff
@@ -0,0 +1,13 @@
+using Xunit;
+using Xunit.Sdk;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Tests.Extensions
+{
+    public static class AssertExtensions
+    {
+        public static T AssertFail<T>(string message)
```
benoutram, line 8: I think it would be good to add a comment to say how this is intended to be used.

src/explore-education-statistics-common/src/services/permalinkService.ts
```diff
@@ -37,7 +37,7 @@ export default {
     releaseId?: string,
   ): Promise<Permalink> {
     if (releaseId) {
-      return dataApi.post(`/release/${releaseId}/permalink`, query);
+      return dataApi.post(`/permalink/release/${releaseId}`, query);
```
benoutram, line 40: Just out of interest was there something broken here?

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Services/AbstractRepository.cs
duncan-at-hiveit: Agreed, will clean up thanks!

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Either.cs
duncan-at-hiveit: Yeah I reckon so - it's a shame we can't currently write

OnSuccess((thing1, thing2) => ...DoSomeWork())

but I'm sure that C# will get that destructuring syntax in its lambdas soon enough, fingers crossed!

src/explore-education-statistics-common/src/services/permalinkService.ts
duncan-at-hiveit: Yeah, if the releaseId argument was supplied, it was trying to call a non-existent endpoint.  Because all of the PermalinkController methods in the Data API projects are prefixed with `/api/permalink`.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/AssertExtensions.cs
duncan-at-hiveit: Added a comment

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Services/AbstractRepository.cs
duncan-at-hiveit: Cleaned up

benoutram (approved):

PR #2520, opened 2021-05-10
EES-1629 remove manage contacts page

amyb-hiveit:
Removes the Platform Administration Manage Contacts page as it's no longer required.


benoutram:
Raised the question with Cam and Laura on whether MethodologySummaryForm still needs the code to be able to get the list of all contacts because if not there's quite a bit more cleanup that can be done I think.

PR #2532, opened 2021-05-18
EES-2190 display chart indicator units, warning for mixed units

amyb-hiveit:
- EES-2277 Show indicator units from the meta data, if no override set
- EES-2278 Leave unit blank if £ as unable to change position of the indicator (issue raised with recharts library for this)
- EES-2281 If data sets with different units are added to the same chart, leave the units blank and show a warning message on the data sets tab
- EES-2279 Fixes saving a blank override unit 
- EES-2280 Fixes displaying the override unit in the form

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartDataSetsConfiguration.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartDataSetsConfiguration.tsx
@@ -13,6 +13,7 @@ import { LocationFilter } from '@common/modules/table-tool/types/filters';
 import { FullTableMeta } from '@common/modules/table-tool/types/fullTable';
 import { Dictionary } from '@common/types';
 import Yup from '@common/validation/yup';
+import WarningMessage from '@common/components/WarningMessage';
 import { Formik } from 'formik';
 import difference from 'lodash/difference';
 import mapValues from 'lodash/mapValues';
@@ -55,6 +56,19 @@ const ChartDataSetsConfiguration = ({
     [meta.locations],
   );
 
+  const hasMixedUnits = useMemo(() => {
+    const units: string[] = [];
+    dataSets.forEach(dataSet => {
+      const foundIndicator = meta.indicators.find(
+        indicator => indicator.value === dataSet.indicator,
+      );
+      if (foundIndicator) {
+        units.push(foundIndicator.unit);
+      }
+    });
+    return !units.every(unit => unit === units[0]);
+  }, [dataSets, meta.indicators]);
+
   useEffect(() => {
     updateForm({
       formKey: 'dataSets',
@@ -64,6 +78,11 @@ const ChartDataSetsConfiguration = ({
 
   return (
     <>
+      {hasMixedUnits && (
+        <WarningMessage>
+          Selected data sets have different indicator units.
+        </WarningMessage>
+      )}
       <Formik<FormValues>
         initialValues={{
           filters: mapValues(meta.filters, filterGroup =>
--- a/src/explore-education-statistics-common/src/modules/charts/util/getUnit.ts
+++ b/src/explore-education-statistics-common/src/modules/charts/util/getUnit.ts
@@ -1,0 +1,23 @@
+import { DataSetCategoryConfig } from '@common/modules/charts/util/getDataSetCategoryConfigs';
+
+const getUnit = (
+  configUnit: string | undefined,
+  dataSetCategoryConfigs: DataSetCategoryConfig[],
+) => {
+  if (configUnit) {
+    return configUnit;
+  }
+
+  const units = dataSetCategoryConfigs.map(config => {
+    return config.dataSet.indicator.unit;
+  });
+
+  // Check if all units are the same, leave blank if not.
+  // Don't display £ as unable to position them before the value. (EES-2278)
+  if (units.every(unit => unit === units[0])) {
+    return units[0] === '£' ? '' : units[0];
+  }
+
+  return '';
+};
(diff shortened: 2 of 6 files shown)


benoutram (dismissed):

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartDataSetsConfiguration.tsx
```diff
@@ -55,6 +56,19 @@ const ChartDataSetsConfiguration = ({
...
+    const units: string[] = [];
+    dataSets.forEach(dataSet => {
+      const foundIndicator = meta.indicators.find(
+        indicator => indicator.value === dataSet.indicator,
+      );
+      if (foundIndicator) {
+        units.push(foundIndicator.unit);
+      }
+    });
+    return !units.every(unit => unit === units[0]);
```
duncan-at-hiveit, line 69: Are we sure at this point that we've got at least one unit available in our `units` array at this point?  If not, this'll get an index out of bounds error, so maybe worth doubling up with a length check?

src/explore-education-statistics-common/src/modules/charts/util/getUnit.ts
```diff
@@ -0,0 +1,23 @@
+import { DataSetCategoryConfig } from '@common/modules/charts/util/getDataSetCategoryConfigs';
+
+const getUnit = (
+  configUnit: string | undefined,
```
duncan-at-hiveit, line 4: Instead of writing `configUnit: string | undefined` in Typescript, you can write `configUnit?: string` - the question mark shows it's possibly undefined.  We predominantly use this syntax for possibly-undefined properties elsewhere in the codebase I believe.

src/explore-education-statistics-common/src/modules/charts/util/getUnit.ts
```diff
@@ -0,0 +1,23 @@
...
+    return configUnit;
+  }
+
+  const units = dataSetCategoryConfigs.map(config => {
+    return config.dataSet.indicator.unit;
+  });
+
+  // Check if all units are the same, leave blank if not.
+  // Don't display £ as unable to position them before the value. (EES-2278)
+  if (units.every(unit => unit === units[0])) {
```
duncan-at-hiveit, line 17: Could instead of writing:

```
if (units.every(unit => unit === units[0])) {
    return units[0] === '£' ? '' : units[0];
}

return '';
```

write instead:

```
return units[0] === '£' || !units.every(unit => unit === units[0])) ? '' : units[0];
```

I guess at this point we're happy that our `dataSetCategoryConfigs` array has at least one element?

src/explore-education-statistics-common/src/modules/charts/util/getUnit.ts
```diff
@@ -0,0 +1,23 @@
+import { DataSetCategoryConfig } from '@common/modules/charts/util/getDataSetCategoryConfigs';
+
+const getUnit = (
```
duncan-at-hiveit, line 3: Would possibly consider calling this method `getUnitOrDefault` just to be explicit in the fact that it is falling back to a default value if the first value doesn't exist?

src/explore-education-statistics-common/src/modules/charts/util/getUnit.ts
benoutram: I've made this change. I had to reorder the params to do so.

src/explore-education-statistics-common/src/modules/charts/util/getUnit.ts
benoutram: I've made this change now - `getUnit` is now `getUnitOrDefault`

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartDataSetsConfiguration.tsx
benoutram: If `units` is empty the condition is not evaluated. The docs say  `Calling this method on an empty array will return true for any condition`. I tested this with an empty array and hasMixedUnits returns `false` as expected.

src/explore-education-statistics-common/src/modules/charts/util/getUnit.ts
benoutram: I've made this change now. 

This would have returned undefined before if `dataSetCategoryConfigs` was  empty. This is what was previously happening:

* If `dataSetCategoryConfigs` is empty then:
* `units` is empty
*  so `units[0]` is undefined
*  so `units[0] === '£'` evaluates to false
* so !units.every(unit => unit === units[0]) evaluates to false (since the condition is ignored, `every` returns true)
* so `units[0]` is returned which is undefined.

Now added an addition check `units.length === 0` to return `''` rather than undefined.

benoutram (approved):

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartDataSetsConfiguration.tsx
duncan-at-hiveit: Haha ahh yes of course!  Thanks :)

src/explore-education-statistics-common/src/modules/charts/util/getUnit.ts
duncan-at-hiveit: Oh sorry then @benoutram , that was probably a bad steer from me then.  I'll set it back - I was mixing up method parameters with object properties.

benoutram (approved):

PR #2536, opened 2021-05-19
Ees 2247 display old release subjects and in create tables path

duncan-at-hiveit:
This PR:
- corrects an error whereby navigating to an old release Fast Track would be displaying "Subject: None" in the "Choose Subject" step
- corrects an error whereby navigating to an old release Fast Track's "Choose Subject" step would not provide any Subjects to choose from
- changes the "Create tables" button when clicked from an old Release's Fast Track page to take the user to the Table Tool with that Release's Subjects available, rather than those from the latest Release for the same Publication

As part of this work, a few related properties around `TableToolPage.tsx`, `TableToolWizard.tsx` and `TableToolWizardFinalStep.tsx` have been combined together as they are always supplied together.  They have been combined into a single `selectedPublication` property, which represents either the Publication that the user has manually chosen in the Table Tool, or the one saved in a Fast Track, and always contains data about the `selectedRelease` (the latest for the chosen Publication or the specific owner of the Fast Track) and the chosen Publication's `latestRelease`.

By having these 2 properties (`selectedRelease` and `latestRelease`) consistently available, it helps to simplify the navigation code to determine where actions like "Create table" or "View the release for this data" should take the user at any given time.

It also helps to simplify which optional props should be provided to the TableTool components.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Api/Controllers/PublicationController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Api/Controllers/PublicationController.cs
@@ -7,7 +7,7 @@ using Microsoft.AspNetCore.Mvc;
 
 namespace GovUk.Education.ExploreEducationStatistics.Data.Api.Controllers
 {
-    [Route("api/publications")]
+    [Route("api")]
     [ApiController]
     public class PublicationController : ControllerBase
     {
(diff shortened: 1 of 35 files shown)


src/GovUk.Education.ExploreEducationStatistics.Data.Api/Controllers/PublicationController.cs
```diff
@@ -18,10 +18,20 @@ public PublicationController(IPublicationService publicationService)
...
+        [HttpGet("publications/{publicationId}")]
+        public async Task<ActionResult<SubjectsAndHighlightsViewModel>> GetLatestPublicationSubjectsAndHighlights(Guid publicationId)
         {
-            return await _publicationService.GetPublication(id).HandleFailuresOrOk();
+            return await _publicationService
+                .GetLatestPublicationSubjectsAndHighlights(publicationId)
+                .HandleFailuresOrOk();
+        }
+
+        [HttpGet("releases/{releaseId}")]
```
benoutram, line 29: This naming within the PublicationController doesn't seem super nice to me as it's not really a Release that's returns, nor a Publication, but I suppose this is the Data API rather than the Content API so a Release can mean something slightly different here.

PR #2542, opened 2021-05-25
Ees 2106 oicd server

duncan-at-hiveit:
This PR:
- adds in an optional Identity Provider, Keycloak, that can be run locally using Docker
- does not yet contain instructions on use in the README, as I will tackle these alongside EES-2107

This PR adds the option of starting up the project with the Keycloak IDP in a Docker container, and containing our usual set of users.

This PR also amends the registration code (in ExternalLogin.cshtml.cs) to allow newly logging in users to be registered on the platform as per usual (by the various Identity Framework AspNet* table entries being recorded) but with the ability now to either have a new set of "internal" (EES) User and Role records created (in the Users and UserReleaseRoles tables), or to link to *existing* Users and UserReleaseRoles records if some already exist in the `content` db with a matching email address to the newly registered User.

This means that we can preload the `content` db with a set of Users and UserReleaseRoles records but *without* the corresponding records in the AspNet* tables just yet, and have the ability to log in with those people and have instant access to the correct Role and Release Role assignments.  The reason we can't just do all this with UserInvites and UserReleaseInvites is that we still need those Users entries in the database so we can refer to them from seed data (wherever `CreatedById` is used, for instance). 

Instructions on how to set up and run the project with Keycloak will be provided after tackling EES-2107, and so for now, local development environments will continue to use Active Directory by default.  However, if wanting to test it out in the meantime:

```
# clear out Identity user records in local db and replace with invites (this'll be done already when we create the new baseline files but for now we can do it manually to our existing local development databases for testing)

DELETE FROM AspNetUserLogins;
DELETE FROM AspNetUserTokens;
DELETE FROM AspNetUserRoles;
DELETE FROM AspNetUsers;
DELETE FROM UserInvites;

INSERT INTO UserInvites
VALUES
('ees-bau1@education.gov.uk', 0, CURRENT_TIMESTAMP, null, (SELECT Id From AspNetRoles WHERE Name = 'BAU User')),
('ees-bau2@education.gov.uk', 0, CURRENT_TIMESTAMP, null, (SELECT Id From AspNetRoles WHERE Name = 'BAU User')),
('ees-analyst1@education.gov.uk', 0, CURRENT_TIMESTAMP, null, (SELECT Id From AspNetRoles WHERE Name = 'Analyst')),
('ees-analyst2@education.gov.uk', 0, CURRENT_TIMESTAMP, null, (SELECT Id From AspNetRoles WHERE Name = 'Analyst')),
('ees-analyst3@education.gov.uk', 0, CURRENT_TIMESTAMP, null, (SELECT Id From AspNetRoles WHERE Name = 'Analyst')),
('ees-prerelease1@education.gov.uk', 0, CURRENT_TIMESTAMP, null, (SELECT Id From AspNetRoles WHERE Name = 'Prerelease User')),
('ees-prerelease2@education.gov.uk', 0, CURRENT_TIMESTAMP, null, (SELECT Id From AspNetRoles WHERE Name = 'Prerelease User'));
```
and
```
# start up Keycloak 
cd useful-scripts
./run.js idp
```
and
```
# start up admin, targeting Keycloak
export IdpProviderConfiguration=Keycloak && ./run.js admin
```
and log in (suggest clearing out local storage and cookies or going Incognito first) with credentials like:
```
ees-bau1@education.gov.uk / password
or
bau1 / password
```

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Linq;
 using System.Security.Claims;
 using System.Threading.Tasks;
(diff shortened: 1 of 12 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
```diff
@@ -104,21 +104,34 @@ public async Task<IActionResult> OnGetCallbackAsync(string returnUrl = null, str
             {
                 var nameClaim = info.Principal.FindFirstValue(JwtClaimTypes.Name);
 
-                if (nameClaim.IndexOf(' ') > 0)
+                if (nameClaim == null)
```
benoutram, line 107: In future we could look at splitting this up a bit, maybe in another service, as `OnGetCallbackAsync` has reached 140+ lines. 🙂

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
```diff
@@ -130,74 +143,93 @@ public async Task<IActionResult> OnGetCallbackAsync(string returnUrl = null, str
...
+                inviteToSystem.Accepted = true;
+                _usersAndRolesDbContext.UserInvites.Update(inviteToSystem);
+
+                // See if we have an "internal" User record in existence yet that has a matching email address to the
+                // new user logging in.  If we do, create this new AspNetUser record with a matching one-to-one id.
+                // Otherwise, later on we'll create a new "internal" Users record with an id matching this AspNetUser's 
+                // id, continuing to establish the one-to-one relationship.
+                var existingInternalUser = await _contentDbContext
+                    .Users
+                    .SingleOrDefaultAsync(u => u.Email.ToLower() == email.ToLower());
```
benoutram, line 162: This string comparison if it's executed on the database, which it should be, won't be case sensitive, but probably a good idea to leave the `ToLower` here anyway.

PR #2546, opened 2021-05-26
EES-1126 show correct time period table headers in reorder list

amyb-hiveit:
When terms were used for time periods in the table tool extra terms were shown in the reorder table headers list, this PR filters the unnecessary ones out.

### DIFF
--- a/src/explore-education-statistics-common/src/modules/table-tool/utils/__tests__/getDefaultTableHeadersConfig.test.ts
+++ b/src/explore-education-statistics-common/src/modules/table-tool/utils/__tests__/getDefaultTableHeadersConfig.test.ts
@@ -152,74 +152,79 @@ describe('getDefaultTableHeadersConfig', () => {
   });
 
   test('returns correct config using a larger number of time periods than indicators', () => {
-    const testTableDataTimePeriods = JSON.parse(JSON.stringify(testTableData));
-    testTableDataTimePeriods.subjectMeta.indicators = [
-      {
-        value: 'overall-absence-sessions',
-        label: 'Number of overall absence sessions',
-        unit: '',
-        name: 'sess_overall',
-        decimalPlaces: 2,
+    const testTableDataTimePeriods = {
+      ...testTableData,
+      subjectMeta: {
+        ...testTableData.subjectMeta,
+        indicators: [
+          {
+            value: 'overall-absence-sessions',
+            label: 'Number of overall absence sessions',
+            unit: '',
+            name: 'sess_overall',
+            decimalPlaces: 2,
+          },
+          {
+            value: 'authorised-absence-sessions',
+            label: 'Number of authorised absence sessions',
+            unit: '',
+            name: 'sess_authorised',
+            decimalPlaces: 2,
+          },
+          {
+            value: 'persistent-absentees',
+            label: 'Number of persistent absentees',
+            unit: '',
+            name: 'enrolments_pa_10_exact',
+            decimalPlaces: 2,
+          },
+        ],
+        timePeriodRange: [
+          { label: '2014/15', code: 'AY', year: 2014 },
+          { label: '2015/16', code: 'AY', year: 2015 },
+          { label: '2016/17', code: 'AY', year: 2016 },
+          { label: '2017/18', code: 'AY', year: 2017 },
+          { label: '2018/19', code: 'AY', year: 2018 },
+        ],
       },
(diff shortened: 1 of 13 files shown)


src/explore-education-statistics-common/src/modules/table-tool/utils/__tests__/getDefaultTableHeadersConfig.test.ts
```diff
@@ -380,74 +325,59 @@ describe('getDefaultTableHeadersConfig', () => {
...
-                  },
-                ],
-              },
-            },
-          },
-        },
-        footnotes: [],
-        geoJsonAvailable: false,
-        indicators: [
+    const testTableDataSiblingless = JSON.parse(JSON.stringify(testTableData));
```
duncan-at-hiveit, line 328: I think for cases like this, can you achieve the same behaviour without the JSON-jugging by using object spreading like:

```
const testTableDataSiblingless = {
    ...testTableData,
    subjectMeta.filters.Characteristic.options: {
      EthnicGroupMajor: {
        blah blah
      }
    }
};
```

?

This should create the deep copy that you're after whilst also letting you override the bits that you want to have specifically different.

src/explore-education-statistics-common/src/modules/table-tool/utils/__tests__/mapTableHeadersConfig.test.ts
```diff
@@ -331,6 +346,76 @@ describe('mapTableHeadersConfig', () => {
     ]);
   });
 
+  test('correctly maps term time periods to only return those in the table result', () => {
+    const testHeaderConfigWithTerms = JSON.parse(
```
duncan-at-hiveit, line 350: Again as an example of the object spreading here, the same can be achieved with:

```
    const testHeaderConfigWithTerms = {
        ....testHeaderConfig,
        columns: [
            {
                 value: '2017_T1',
                 type: 'TimePeriod',
            },
            ........... big long array here .........., 
        ],
     };
```

src/explore-education-statistics-common/src/modules/table-tool/utils/getDefaultTableHeadersConfig.ts
```diff
@@ -90,19 +94,34 @@ function getFixedTimePeriodAndIndicatorTableHeadersConfig(
...
+    filters,
+    locations,
+    timePeriodRange,
+  } = fullTable.subjectMeta;
+  const { results } = fullTable;
+
+  // When terms are selected the time period range can include ones not displayed in the table,
+  // filter these out to ensure the reorder headings list is correct
+  const filteredTimePeriodRange = timePeriodRange.filter(period =>
+    results.find(result => result.timePeriod === period.value) ? period : null,
```
duncan-at-hiveit, line 110: I think you shouldn't need the `? period : null` part, as `filter()` wants a boolean returned and `results.find(...)` is already doing that for you here.

So I think it can be rewritten:

```
const filteredTimePeriodRange = timePeriodRange.filter(period =>
    results.find(result => result.timePeriod === period.value));
```

src/explore-education-statistics-common/src/modules/table-tool/utils/mapTableHeadersConfig.ts
```diff
@@ -12,32 +12,44 @@ import {
...
   tableHeaders: UnmappedTableHeadersConfig,
-  meta: FullTableMeta,
+  table: FullTable,
 ): TableHeadersConfig {
+  const { subjectMeta, results } = table;
   const mapToFilters = (headers: TableHeader[]): Filter[] => {
-    return headers.map(header => {
+    // When terms are selected the time period range can include ones not displayed in the table,
+    // filter these out to ensure the reorder headings list is correct
+    const filteredHeaders = headers.filter(header => {
```
duncan-at-hiveit, line 21: Again, I think this can be slightly simplified to:

```
    const filteredHeaders = headers.filter(header => {
      if (header.type !== 'TimePeriod') {
           return true;
      }
      return results.find(result => result.timePeriod === header.value)
    });
```

or even:

```
    const filteredHeaders = headers.filter(header => 
        header.type !== 'TimePeriod') || 
        results.find(result => result.timePeriod === header.value));
```

to be super-succinct!  But maybe harder to follow!

duncan-at-hiveit (approved):

PR #2547, opened 2021-05-26
EES-2312 remove edit pre-release access list button if dont have permission

amyb-hiveit:


src/explore-education-statistics-admin/src/pages/release/pre-release/components/__tests__/PublicPreReleaseAccessForm.test.tsx
```diff
@@ -156,4 +159,22 @@ describe('PublicPreReleaseAccessForm', () => {
...
+      <TestConfigContextProvider>
+        <PublicPreReleaseAccessForm
+          canUpdateRelease={false}
+          preReleaseAccessList={testAccessList}
+          onSubmit={noop}
+        />
+      </TestConfigContextProvider>,
+    );
+    expect(
+      screen.getByText(
```
duncan-at-hiveit, line 174: Nice one!  This should probably be `queryByText` rather than `getByText` as `getByText` will blow up if it can't find the given text, whereas we really want `expect` to be doing the blowing up :)  `queryByText` is more forgiving and just returns null, allowing `expect(...).toBeInTheDocument()` to do its work.

markhiveit (approved):

PR #2548, opened 2021-05-26
Ees 2107 bootstrap bau users

duncan-at-hiveit:
This PR:
- adds a mechanism for auto-inviting a set of BAU Users (normally to a blank database but not strictly necessary) if specified
- updates the README with details of using the Keycloak Identity Provider

The use of this bootstrapping user mechanism along with using the Keycloak IdP is documented and aimed towards getting the service set up for people quickly, particularly against a blank `content` database.  The 2 environment variables:
```
IdpProviderConfiguration=Keycloak
BootstrapUsersConfiguration=KeycloakBootstrapUsers
```
will allow admin to automatically add invites for "bau1" and "bau2", and redirect to Keycloak for authentication ("bau1" / password and "bau2" / password to log in).  These users will then be able to use the service immediately and bootstrap the rest of the data for the service, including inviting other users from within the Keycloak IdP user list.


README.md
```diff
@@ -59,25 +59,30 @@ You will need the following groups of dependencies to run the project successful
    - [.NET Core v3.1](https://dotnet.microsoft.com/download/dotnet-core/3.1)
    - [Azure Functions Core Tools v3+](https://github.com/Azure/azure-functions-core-tools)
    
-2. To emulate azure storage you will require either:
+2. To emulate azure storage you will require:
    - [Azure Storage Emulator](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-emulator)
 
-  To run Azure Storage Emulator, you'll need to install LocalDB via SQL Express. You can find the installer for Windows x64 [here](https://download.microsoft.com/download/3/9/F/39F968FA-DEBB-4960-8F9E-0E7BB3035959/SQLEXPR_x64_ENU.exe).
+  To run Azure Storage Emulator, you'll need to install LocalDB via SQL Express. You can find the installer for Windows
```
benoutram, line 65: You don't necessarily need SQL Express as you can initialise Storage Emulator to use the SQL Server. I do this and have a separate `AzureStorageEmulatorDb510` schema along side `master`, `content`, `statistics` etc. The instructions are here https://docs.microsoft.com/en-us/azure/storage/common/storage-use-emulator#start-and-initialize-the-storage-emulator

README.md
```diff
@@ -59,25 +59,30 @@ You will need the following groups of dependencies to run the project successful
...
+   Alternatively you could create your own Storage Account on Azure and amend your storage connection strings to point 
+   to this.
 
 3. To run the databases, you can use either:
 
    - [SQL Server 2017+](https://www.microsoft.com/en-gb/sql-server/sql-server-downloads)
    - [Docker and Docker Compose](https://docs.docker.com/)
 
+The SQL Server Developer Edition is free and more fully-featured than SQL Express. Note that SQL Express is still 
+required for its LocalDB support for the Azure Storage Emulator.
```
benoutram, line 85: We could recommend initialising storage explorer with SQL server rather than SQL Express as mentioned above, to avoid people needing to install that too.

README.md
```diff
@@ -88,21 +93,66 @@ for development, specifically the table storage component which is not supported
...
+```
+And additionally if wanting to set up a set of Keycloak users automatically in the service in order to start using 
+the service against an empty set of databases, set the following environment variable:
+```
+BootstrapUsersConfiguration=KeycloakBootstrapUsers
+```
+The effect of setting these 2 environment variables together will allow authentication of users with Keycloak and those
+users specified within the 
+`src\GovUk.Education.ExploreEducationStatistics.Admin\appsettings.KeycloakBootstrapUsers.json` will automatically be 
+able to be users to access the system as "BAU Users" - effectively able to bootstrap the rest of the data needed to use
```
benoutram, line 139: I think "bootstrap the rest of the data needed to use the service" is a bit confusing here because it made me think that it somehow grants permission for me as a BAU user logging in to be able to bootstrap some extra application data like Releases. Maybe we could we list what gets bootstrapped as a result of this? It seems like it's going to be the Content database Users and Invites run on Startup. Also "...will automatically be able to be users to access the system as..." doesn't sound right.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
```diff
@@ -391,813 +306,6 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
-                                        LineStyle = ChartLineStyle.solid
-
-                                    }
-                                }
-                            }
-                        }
-                    }
-                },
-                // absence generic data blocks used in content
-                new DataBlock
```
benoutram, line 860: Just letting you know the reason why we retained this seed data in this file here in case you didn't realise this as it's not very obvious.

We retained a bare minimum set of  Release content data for all the entity types being persisted as JSON e.g. Data block query, data block table config, data block chart config, methodology.

This was so that when creating an EF core migration we had the chance to spot if some of that JSON data was mutating in the generated migration and could then do something about it on a wider scale to migrate data in the wild if necessary e.g. by introducing a SQL update or a custom migration endpoint to transform the JSON for _all_ rows in the Prod database.

Without tracking any data used by these JSON fields here, when changing the models there was no indication in the migration that any data in the wild might need updating. Since the fields were of type String before and String afterwards, they remain as VARCHAR in the database and nothing appeared to have changed.

Updates probably aren't dangerous where fields are added or removed but might be where they are renamed or their type is altered.

Admittedly this was a poor way of testing for unsuspecting problematic changes as not everyone is going to attempt to generate a migration with dotnet tools after every change, it's not commented to explain the reason for it existing, and it's done nothing but confuse everyone who has looked at this and thought it's still a hangover from go-live!

benoutram (approved):

PR #2552, opened 2021-05-27
EES-177 Display Uploaded by and Date uploaded for ancillary files

markhiveit:


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseFileRepository.cs
```diff
@@ -102,7 +102,8 @@ public async Task<ReleaseFile> Get(Guid releaseId, Guid fileId)
         public async Task<List<ReleaseFile>> GetByFileType(Guid releaseId, params FileType[] types)
         {
             return await _contentDbContext.ReleaseFiles
-                .Include(f => f.File)
+                .Include(rf => rf.File)
+                .ThenInclude(f => f.CreatedBy)
```
benoutram, line 106: This include for the CreatedBy user is only added here on `GetByFileType` but there are other methods returning ReleaseFiles like `Get` where it's not been added. It's not immediately obvious to the reader why `CreatedBy` is being included here, since this is a generic repository. There's nothing stopping the `ToFileInfo` extension being called on a ReleaseFile that's been retrieved using `Get` which will then have a blank email, not because it was a legacy file that didn't have a `CreatedBy` but because of a bug where we forgot to include it. It might be better to use this explicit load of `CreatedBy` like we do elsewhere for data files instead in every place before creating `FileInfo` to ensure it's always been loaded.

```
await _contentDbContext.Entry(releaseFile)
    .Reference(rf => rf.CreatedBy)
    .LoadAsync();
```

src/GovUk.Education.ExploreEducationStatistics.Common/Model/FileInfo.cs
```diff
@@ -12,6 +12,8 @@ public class FileInfo
         public string FileName { get; set; }
         public string Name { get; set; }
         public string Size { get; set; }
+        public string UserName { get; set; }
```
benoutram, line 15: This class is used as a view model despite not containing the name `ViewModel`.
It lives in the `Common` package because it's being used in several places including the Public by the Publisher and the Content API.

With this change these fields might be persisted to the cached public JSON content and possibly returned in the public content API JSON response too, depending on the serialization config and whether its set to include null fields in JSON or omit them.

We might want a public file view model that doesn't contain the `UserName` fields and probably the `Created` field too ruling out the possibility of us ever exposing them.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Extensions/ReleaseFileExtensions.cs
```diff
@@ -28,7 +28,9 @@ public static FileInfo ToFileInfo(this ReleaseFile releaseFile, BlobInfo blobInf
                 FileName = releaseFile.File.Filename,
                 Name = releaseFile.Name ?? releaseFile.File.Filename,
                 Size = blobInfo.Size,
-                Type = releaseFile.File.Type
+                Type = releaseFile.File.Type,
```
benoutram, line 31: There's an existing test for this which should get updated.

src/GovUk.Education.ExploreEducationStatistics.Common/Model/FileInfo.cs
```diff
@@ -12,6 +12,8 @@ public class FileInfo
...
+        public DateTime? Created { get; set; }
```
benoutram, line 16: Have you considered the need to set these fields in other places that FileInfo is created, for example in `FileExtension#ToFileInfoNotFound` should we be setting some default/blank values? How about places like where we return image files, methodology image files, and chart files that all use `FileInfo`? This might be another case for introducing other view models here, or we might just not care.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleaseFileController.cs
```diff
@@ -90,7 +91,7 @@ public async Task<ActionResult<FileInfo>> UpdateChartFile(Guid releaseId, Guid i
         [HttpPost("release/{releaseId}/ancillary")]
         [RequestSizeLimit(int.MaxValue)]
         [RequestFormLimits(ValueLengthLimit = int.MaxValue, MultipartBodyLengthLimit = int.MaxValue)]
-        public async Task<ActionResult<FileInfo>> UploadAncillary(Guid releaseId,
+        public async Task<ActionResult<AdminFileInfo>> UploadAncillary(Guid releaseId,
```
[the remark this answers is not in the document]
markhiveit: I've created `AncillaryFileInfo` - I did this rather than create an `AncillaryFileViewModel` because the class for the file inherits from `FileInfo` on both the frontend and the backend.

I'm going to create a separate ticket to investigate the UpdateChartFile and UploadChart returning FileInfo rather than an InfographicFileInfo or void. A quick looks suggests it can just return void, but I don't want to get caught up in that with this PR. EES-2354

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseFileService.cs
```diff
@@ -250,5 +252,44 @@ public class ReleaseFileService : IReleaseFileService
...
+                FileName = releaseFile.File.Filename,
+                Name = releaseFile.Name ?? releaseFile.File.Filename,
+                Size = blobInfo.Size,
+                Type = releaseFile.File.Type,
+                UserName = releaseFile.File.CreatedBy?.Email ?? "",
+                Created = releaseFile.File.Created
+            };
+        }
+
+        private async Task<AdminFileInfo> ToAdminFileInfoNotFound(Content.Model.File file)
```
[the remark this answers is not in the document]
markhiveit: I've added comments to remove *FileInfoNotFound methods when EES-2343 is completed.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ManageContent/ManageContentPageServiceTests.cs
```diff
@@ -116,23 +116,27 @@ public async Task GetManageContentPageViewModel()
...
-                new FileInfo
+                new AdminFileInfo
                 {
                     Id = dataFileId,
                     FileName = "data.csv",
                     Name = "Subject File",
                     Size = "20 Kb",
-                    Type = FileType.Data
+                    Type = FileType.Data,
+                    Created = DateTime.UtcNow,
```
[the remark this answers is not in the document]
markhiveit: Now that ListAll returns a FileInfo list, I've switched the test back to how it was originally.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleaseFileController.cs
```diff
@@ -70,31 +70,31 @@ public async Task<ActionResult> Stream(Guid releaseId, Guid id)
         }
 
         [HttpGet("release/{releaseId}/ancillary")]
-        public async Task<ActionResult<IEnumerable<FileInfo>>> GetAncillaryFiles(Guid releaseId)
+        public async Task<ActionResult<IEnumerable<AncillaryFileInfo>>> GetReleaseAncillaryFilesInfo(Guid releaseId)
```
benoutram, line 73: I think I would have just kept this as GetAncillaryFiles throughout the controller, service and tests. Seems a bit verbose.

benoutram (dismissed):

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleaseFileController.cs
markhiveit: I'll change it before merging

benoutram (approved):

PR #2553, opened 2021-05-28
EES-2137 - fixed remaining failing UI tests after front-end changes

duncan-at-hiveit:
This PR:
- fixes currently-failing UI tests
- adds arguments to run_tests.py to rerun failed tests and suites and combine results with original run
- adds arguments to run_tests.py to specify own implicit and explicit wait times for Robot runs
- adds the generation of an HTML dump to accompany screenshots when failures occur.  From these it is easier to find out the cause of failures if down to difference in HTML structure.

An example usage of running with the extra options:

`pipenv run python3 run_tests.py -i robot -b chrome -e local -v --rerun-failed-suites --timeout 10 --implicit-wait 2` 

In this particular example, this is re-running only the suites that failed in the previous run, and then merging any new successes with the original run results and creating a new report.html.

### DIFF
--- a/src/explore-education-statistics-frontend/README.md
+++ b/src/explore-education-statistics-frontend/README.md
@@ -55,6 +55,10 @@ Your app is ready to be deployed!
 
 See the section about [deployment](https://facebook.github.io/create-react-app/docs/deployment) for more information.
 
+### `npm run start:prod`
+
+After building the app for production, this runs the production-ready bundle. This command can be helpful if you are experiencing flakiness or performance issues when running UI / unit tests locally, as pages will load much quicker.
+
 ### `npm run eject`
 
 **Note: this is a one-way operation. Once you `eject`, you can’t go back!**
--- a/tests/robot-tests/run_tests.py
+++ b/tests/robot-tests/run_tests.py
@@ -21,6 +21,7 @@ import requests
 from dotenv import load_dotenv
 from pabot.pabot import main as pabot_run_cli
 from robot import run_cli as robot_run_cli
+from robot import rebot_cli as robot_rebot_cli
 
 import scripts.keyword_profile as kp
 from tests.libs.setup_auth_variables import setup_auth_variables
@@ -90,6 +91,22 @@ parser.add_argument("--disable-teardown",
                     dest="disable_teardown",
                     help="disable tearing down of any test data after completion",
                     action='store_true')
+parser.add_argument("--rerun-failed-tests",
+                    dest="rerun_failed_tests",
+                    action='store_true',
+                    help="rerun individual failed tests and merge results into original run results")
+parser.add_argument("--rerun-failed-suites",
+                    dest="rerun_failed_suites",
+                    action='store_true',
+                    help="rerun failed test suites and merge results into original run results")
+parser.add_argument("--timeout",
+                    default="30",
+                    dest="timeout",
+                    help="default robot timeout in seconds (default is 30)")
+parser.add_argument("--implicit-wait",
+                    default="5",
+                    dest="implicit_wait",
+                    help="default robot implicit wait in seconds (default is 5)")
 
 """
 NOTE(mark): The admin and analyst passwords to access the admin app are stored in the CI pipeline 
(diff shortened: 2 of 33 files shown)


src/explore-education-statistics-frontend/README.md
```diff
@@ -55,6 +55,8 @@ Your app is ready to be deployed!
 
 See the section about [deployment](https://facebook.github.io/create-react-app/docs/deployment) for more information.
 
+Locally you can also run `npm run start:prod` to run the production-ready bundle.
```
luke-h1, line 58: It'd be best to make this stand out more and provide some more info on how the team typically uses this command (aside from deployment). So you could do something such as: 
``` 
### `npm run start:prod`
Builds the app for production. This command can also be helpful if you are experiencing flakiness when running tests locally

File: tests/robot-tests/run_tests.py [hunk 2]
```diff
@@ -90,6 +91,22 @@
                     dest="disable_teardown",
                     help="disable tearing down of any test data after completion",
                     action='store_true')
+parser.add_argument("--rerun-failed-tests",
+                    dest="rerun_failed_tests",
+                    action='store_true',
+                    help="rerun individual failed tests and merge results into original run results")
+parser.add_argument("--rerun-failed-suites",
```
luke-h1, line 98: Loving these changes! 🚀

tests/robot-tests/tests/admin_and_public/bau/prerelease_visibility.robot
```diff
@@ -54,7 +54,8 @@ Go to 'Sign Off' page
     user clicks link  Sign off
     user waits for page to finish loading
     user waits until page contains testid  public-release-url
-    ${PUBLIC_RELEASE_LINK}=  Get Text  xpath://*[@data-testid="public-release-url"]
+    ${PUBLIC_RELEASE_LINK}=  Get Value  xpath://*[@data-testid="public-release-url"]
+    check that variable is not empty  PUBLIC_RELEASE_LINK   ${PUBLIC_RELEASE_LINK}
```
luke-h1, line 58: Nice! That'll be a handy keyword for the future

tests/robot-tests/tests/admin_and_public/bau/publish_release_and_amend_2.robot
```diff
@@ -721,7 +719,7 @@ Generate table
 
 Wait until the table is generated
     [Tags]  HappyPath
-    user waits until page contains button  Generate permanent link
+    user clicks button  Share your table
```
luke-h1, line 722: You'll probably want to what until the button is rendered (dev can sometimes be slow depending on the amount of people using it). So you'll want to add the below line above `user clicks button Share your table` 

`user waits until page contains button. Share your table  60`

tests/robot-tests/tests/libs/admin-common.robot
```diff
@@ -29,7 +29,7 @@ user signs in as analyst1
 
     user waits until h1 is visible  Dashboard
     user waits until page contains title caption  Welcome Analyst1
-    user waits until page contains element   id:publicationsReleases-themeTopic-themeId   180
+    user waits until page contains element   css:#publicationsReleases-themeTopic-themeId,[data-testid='no-permission-to-access-releases']   180
```
luke-h1, line 32: Don't think the comma here is necessary?

luke-h1 (changes requested):

tests/robot-tests/tests/libs/admin-common.robot
markhiveit: I think this is necessary. The selector is either looking for either `#publicationsReleases-themeTopic-themeId` or `[data-testid='no-permission-to-access-releases']`. The first appears if analyst1 does have permission view/edit releases - the second appears is analyst1 has no permission to view/edit any releases. This covers both cases.

tests/robot-tests/tests/admin_and_public/bau/publish_release_and_amend_2.robot
markhiveit: `user clicks button` contains `user gets button element` which in turn calls `user waits until parent contains element` - so an additional wait shouldn't be necessary.

tests/robot-tests/tests/libs/admin-common.robot
luke-h1: ah that's pretty cool. Didn't realise that was possible

luke-h1:
UI test changes look good. However, it looks like some unit tests are failing. I think the snapshots will need updating

markhiveit (approved):

PR #2566, opened 2021-06-04
Ees 2203 make methodology amendment auth handler

duncan-at-hiveit:
This PR:
- adds a new `MakeAmendmentOfSpecificMethodologyAuthorizationHandler` class with associated permission checks
- adds new fields to `Methodology` to check if a given Methodology is publicly accessible
- adds a new stub `MethodologyAmendmentService` with permissions wired in and tested, but currently not wired into DI
- corrects various tests currently using the `PolicyCheckBuilder` that were not awaiting the results of their tests, and therefore always passing
- performs some slight `PolicyCheckBuilder` method renames to better reflect what they're doing
- creates a new `PolicyCheckBuilder.SetupResourceCheckToFailWithMatcher()` method that takes an argument-matching expression rather than an instance of an object, to match successfully against objects retrieved from the InMemory...DbContexts
- does some slight refactoring of the `MethodologyContentService`, `MethodologyContentServicePermissionTests` and `UpdateSpecificMetholodgyAuthorizationHandler` classes to push a Methodology Status check into the handler itself rather than the Service

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandlersTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandlersTests.cs
@@ -1,19 +1,41 @@
 ﻿using System.Threading.Tasks;
 using GovUk.Education.ExploreEducationStatistics.Admin.Security.AuthorizationHandlers;
+using GovUk.Education.ExploreEducationStatistics.Content.Model;
 using Xunit;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Security.SecurityClaimTypes;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Tests.Security.AuthorizationHandlers.Utils.AuthorizationHandlersTestUtil;
+using static GovUk.Education.ExploreEducationStatistics.Content.Model.MethodologyStatus;
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Security.AuthorizationHandlers
 {
     public class UpdateSpecificMethodologyAuthorizationHandlersTests
     {
         [Fact]
-        public async Task UpdateAllSpecificMethodologies()
+        public async Task UpdateAllSpecificMethodologies_DraftMethodology()
         {
-            await AssertHandlerSucceedsWithCorrectClaims<UpdateSpecificMethodologyRequirement>(
+            var draftMethodology = new Methodology
+            {
+                Status = Draft
+            };
+            
+            await AssertHandlerSucceedsWithCorrectClaims<Methodology, UpdateSpecificMethodologyRequirement>(
                 new UpdateSpecificMethodologyAuthorizationHandler(),
+                draftMethodology,
                 UpdateAllMethodologies
+            );
+        }
+
+        [Fact]
+        public async Task UpdateAllSpecificMethodologies_ApprovedMethodologyFails()
+        {
+            var approvedMethodology = new Methodology
+            {
+                Status = Approved
+            };
+            
+            await AssertHandlerSucceedsWithCorrectClaims<Methodology, UpdateSpecificMethodologyRequirement>(
+                new UpdateSpecificMethodologyAuthorizationHandler(),
+                approvedMethodology
             );
         }
     }
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandlersTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandlersTests.cs
@@ -1,41 +1,19 @@
-﻿using System.Threading.Tasks;
+﻿﻿using System.Threading.Tasks;
 using GovUk.Education.ExploreEducationStatistics.Admin.Security.AuthorizationHandlers;
-using GovUk.Education.ExploreEducationStatistics.Content.Model;
 using Xunit;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Security.SecurityClaimTypes;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Tests.Security.AuthorizationHandlers.Utils.AuthorizationHandlersTestUtil;
-using static GovUk.Education.ExploreEducationStatistics.Content.Model.MethodologyStatus;
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Security.AuthorizationHandlers
 {
     public class UpdateSpecificMethodologyAuthorizationHandlersTests
     {
         [Fact]
-        public async Task UpdateAllSpecificMethodologies_DraftMethodology()
+        public async Task UpdateAllSpecificMethodologies()
         {
-            var draftMethodology = new Methodology
-            {
-                Status = Draft
-            };
-            
-            await AssertHandlerSucceedsWithCorrectClaims<Methodology, UpdateSpecificMethodologyRequirement>(
+            await AssertHandlerSucceedsWithCorrectClaims<UpdateSpecificMethodologyRequirement>(
                 new UpdateSpecificMethodologyAuthorizationHandler(),
-                draftMethodology,
                 UpdateAllMethodologies
-            );
-        }
-
-        [Fact]
-        public async Task UpdateAllSpecificMethodologies_ApprovedMethodologyFails()
-        {
-            var approvedMethodology = new Methodology
-            {
-                Status = Approved
-            };
-            
-            await AssertHandlerSucceedsWithCorrectClaims<Methodology, UpdateSpecificMethodologyRequirement>(
-                new UpdateSpecificMethodologyAuthorizationHandler(),
-                approvedMethodology
             );
         }
     }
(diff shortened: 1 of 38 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/MakeAmendmentOfSpecificMethodologyAuthorizationHandlers.cs
```diff
@@ -0,0 +1,67 @@
...
+            // publicly-accessible Methodology.
+            if (SecurityUtils.HasClaim(context.User, MakeAmendmentsOfAllMethodologies))
+            {
+                context.Succeed(requirement);
+                return;
+            }
+
+            // If the user has a Publication Owner role on a Publication that uses this Methodology, they can create 
+            // an Amendment of this Methodology.
+            if (methodology.Publications.IsNullOrEmpty())
```
benoutram, line 48: I think it's worth bring this back up in discussion because so far we've said the amending ability will be limited to a user that has the Owner role on the Publication which the methodology belongs to, rather than Owner role on any Publication using the Methodology.

If we based it on any publication using the methodology then there will be a loophole where any publication owner can in theory get access to amend any methodology just by adopting it.

The same is true for the updating permission which is limited to users who can update the latest release of the publication which the methodology belongs to. The publication owner and their release team would be able to update any methodology that the publication owner decides to adopt.

Instead, the publication owner and release team will either ask the relevant team to make the change or request to become publication owner on the publication / contributor on that publications latest release.

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandlers.cs
```diff
@@ -7,10 +11,19 @@ public class UpdateSpecificMethodologyRequirement : IAuthorizationRequirement
...
-        HasClaimAuthorizationHandler<UpdateSpecificMethodologyRequirement>
+        AuthorizationHandler<UpdateSpecificMethodologyRequirement, Methodology>
     {
-        public UpdateSpecificMethodologyAuthorizationHandler() : base(SecurityClaimTypes.UpdateAllMethodologies)
+        protected override Task HandleRequirementAsync(
+            AuthorizationHandlerContext context,
+            UpdateSpecificMethodologyRequirement requirement,
+            Methodology methodology)
         {
+            if (methodology.Status == Draft && HasClaim(context.User, SecurityClaimTypes.UpdateAllMethodologies))
```
benoutram, line 21: Just wondering if you have tested unapproving a methodology as it looks like this might stop working now. Until we have amendments, analysts need to be able to unapprove them.

In future the approver of the latest release of the publication which the methodology belongs to should be able to unapprove the methodology as long as it's not publicallly accessible yet, so maybe leave a TODO for this here?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandlersTests.cs
```diff
@@ -1,20 +1,42 @@
...
                 new UpdateSpecificMethodologyAuthorizationHandler(),
+                draftMethodology,
                 UpdateAllMethodologies
             );
         }
+
+        [Fact]
+        public async Task UpdateAllSpecificMethodologies_ApprovedMethodologyFails()
+        {
+            var approvedMethodology = new Methodology
```
benoutram, line 31: See comment on the Authorization handler about this check being used to update a methodology which is also used by the call to unapprove a methodology. Due to its RESTful design there's only one update method which serves multiple purposes including unapproving. At least while we don't have amendments unapproving will still be necessary.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Common/Services/CollectionUtils.cs
```diff
@@ -0,0 +1,12 @@
+using System.Collections.Generic;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Services
+{
+    public static class CollectionUtils
+    {
+        public static List<T> AsList<T>(params T[] objects)
```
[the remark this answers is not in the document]
duncan-at-hiveit: The reason I did this was for brevity more than anything else, so rather than:

```
var publications = new List<Publication> {
    publication1,
    publication2
};
```

or 

```
var publications = (new Publication[] { publication1, publication2 }).ToList();
```

you can just write:

```
var publications = AsList(publication1, publication2);
```

Just thought it could help make test code a little more concise?

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/MakeAmendmentOfSpecificMethodologyAuthorizationHandlers.cs
duncan-at-hiveit: Yup that sounds good.  I guess we don't have the model changes in yet to support this, so I could just put a TODO against this with the ticket ref of the model change ticket?

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandlers.cs
duncan-at-hiveit: Will test the unapproving and add the TODO

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandlersTests.cs
duncan-at-hiveit: Cool yeah will retest around this, and no doubt there will need to be a fix put in place.  Now it's a bit more obvious why this status check wasn't in the handler in the first place.

Not sure what your thoughts are on this RESTful approach?  I could imagine you could spend quite a bit of time validating that the updates being made were valid combinations - for instance, should you be able to change a bunch of content and approve in the same call?  Could be argued both ways.  Personally although it's more boilerplate, I do prefer the explicit individual calls, especially if doing something important like updating the status, and I def prefer the idea of an individual handler for approval workflow stuff.

src/GovUk.Education.ExploreEducationStatistics.Common/Services/CollectionUtils.cs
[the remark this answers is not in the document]
benoutram: That makes sense, I was mistaking it for an extension method. Thanks 🙂

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandlersTests.cs
benoutram: I prefer the explicit calls too, as the update methods for methodology and releases are having to do quite a lot. Despite the way we have designed the UI's into separate tabs, currently we still need to worry about the update/validation/security of fields on other tabs that are never getting modified which seems pointless and can lead to bugs like this.

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/MakeAmendmentOfSpecificMethodologyAuthorizationHandlers.cs
benoutram: I think from the SOW4 branch we do have that change so you could either target SOW4 or wait until this is merged via dev back into it and keep this ticket open to make further changes at that point in time?

duncan-at-hiveit: Yup, will do the latter then, thanks!

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandlers.cs
duncan-at-hiveit: Rolled back behaviour and added a TODO

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandlersTests.cs
duncan-at-hiveit: Rolled back behaviour and added a TODO

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/MakeAmendmentOfSpecificMethodologyAuthorizationHandlers.cs
duncan-at-hiveit: Added a TODO to tackle after merge with SOW4 branch

benoutram (approved):

PR #2567, opened 2021-06-07
EES-2030 Add Created date to Admin Data Block Summary table

markhiveit:


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ContentServiceTests.cs
```diff
@@ -149,6 +149,12 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
                 Assert.Equal(release.Content[0].ContentSection.Content[1].Id, contentBlocks[1].Id);
                 Assert.Equal(release.Content[1].ContentSection.Content[0].Id, contentBlocks[2].Id);
                 Assert.Equal(release.Content[1].ContentSection.Content[1].Id, contentBlocks[3].Id);
+
+                // ContentBlocks shouldn't return Created date
+                Assert.Null(contentBlocks[0].Created);
```
benoutram, line 154: It seems like they should return created date as they are content blocks of type `HtmlBlock` rather than `HtmlBlockViewModel` and the change to `ContentService.cs` line 579 sets the created date on new blocks getting created. I think the created dates are null here because you didn't update the setup above to set created dates like they would have been created via the service initially.

benoutram (approved):

PR #2572, opened 2021-06-08
EES-2352 Remove Methodology from Publication after previous migration to associate them with Publication via MethodologyParent.

benoutram:
Methodology has a parent currently called `MethodologyParent`.

Publications can be linked to many `MethodologyParent`s through the `PublicationMethodologies` many-to-many relationship, allowing a Publication to adopt potentially many methodologies in future, including 'owning' one of its own.

This PR removes the `Methodology` and `MethodologyId` fields from `Publication` which is now possible after the previous migration in https://github.com/dfe-analytical-services/explore-education-statistics/pull/2560 to associate them with `Publication` via new parent `MethodologyParent` linked in `PublicationMethodologies`.

It adds many placeholder `TODO` comments for future work to determine the latest related versions of methodologies and for a change to publishing where the Content API will be responsible for returning methodologies from the database rather than from a storage account.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyRepositoryTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyRepositoryTests.cs
@@ -4,220 +4,37 @@ using GovUk.Education.ExploreEducationStatistics.Admin.Services.Methodologies;
 using GovUk.Education.ExploreEducationStatistics.Content.Model;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
 using Xunit;
+using static GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.DbUtils;
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.Methodologies
 {
     public class MethodologyRepositoryTests
     {
         [Fact]
-        public async Task GetMethodologiesForUser_ReleaseRoleLead()
+        public async Task GetLatestMethodologiesByRelease()
         {
-            var userId = Guid.NewGuid();
-            var methodology = new Methodology
-            {
-                Title = "Test methodology",
-            };
-            var publication = new Publication
-            {
-                Methodology = methodology
-            };
-            var release = new Release
-            {
-                Publication = publication
-            };
-            var userReleaseRole = new UserReleaseRole
-            {
-                UserId = userId,
-                Release = release,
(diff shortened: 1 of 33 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyRepositoryTests.cs
```diff
@@ -4,220 +4,37 @@
...
-                var methodologyRepository = BuildMethodologyRepository(contentDbContext);
-                var methodologies = await methodologyRepository.GetMethodologiesForUser(userId);
+                var result = await service.GetLatestMethodologiesByRelease(release.Id);
 
-                Assert.Equal(2, methodologies.Count);
-                Assert.Contains(methodologies, m => m.Id == methodology1.Id);
-                Assert.Contains(methodologies, m => m.Id == methodology2.Id);
-                Assert.DoesNotContain(methodologies, m => m.Id == methodology3.Id);
-                Assert.DoesNotContain(methodologies, m => m.Id == methodology4.Id);
+                // TODO SOW4 EES-2379 Test getting latest methodologies for a Release for approval checklist
```
duncan-at-hiveit, line 32: I'd be tempted to leave the correct assertion in this test (and any other test that has had to change temporarily) as it was (if it was correct still) and move this TODO to a [Skip] annotation on the test method.

Saying that it's not a biggie.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseChecklistServiceTests.cs
```diff
@@ -87,10 +92,20 @@ await using (var context = ContentDbUtils.InMemoryContentDbContext(contextId))
...
+            var methodologyRepository = new Mock<IMethodologyRepository>(MockBehavior.Strict);
+            var releaseDataFileRepository = new Mock<IReleaseDataFileRepository>(MockBehavior.Strict);
+
             await using (var context = ContentDbUtils.InMemoryContentDbContext(contextId))
             {
-                var releaseDataFileRepository = new Mock<IReleaseDataFileRepository>();
-
+                methodologyRepository
+                    .Setup(mock => mock.GetLatestMethodologiesByRelease(release.Id))
+                    .ReturnsAsync(new List<Methodology>
```
duncan-at-hiveit, line 104: Ooh just think, when you can write `.ReturnsAsync(AsList(methodology));` instead :)

duncan-at-hiveit (approved):

PR #2573, opened 2021-06-08
EES-51 published and scheduled tags

amyb-hiveit:
Updates the release tags on the admin dashboard:
- if a release is approved and complete show green 'Published' tag
- if a release is approved and scheduled show blue 'Scheduled' tag (this includes on the Scheduled releases tab)

This PR should also fix the issue raised in https://dfedigital.atlassian.net/browse/EES-791 of the loading spinner always showing. The lazy loader seems unnecessary here as it's not used in other places the release status is shown, and wasn't working, so I've removed it.

I couldn't recreate the other issue on EES-791 where not all the scheduled releases showed and there was a chunk error in the console, so it may have been fixed since this ticket was raised.

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/admin-dashboard/components/ReleaseSummary.tsx
+++ b/src/explore-education-statistics-admin/src/pages/admin-dashboard/components/ReleaseSummary.tsx
@@ -6,7 +6,6 @@ import {
 import { Release } from '@admin/services/releaseService';
 import Details from '@common/components/Details';
 import FormattedDate from '@common/components/FormattedDate';
-import LoadingSpinner from '@common/components/LoadingSpinner';
 import SummaryList from '@common/components/SummaryList';
 import SummaryListItem from '@common/components/SummaryListItem';
 import Tag from '@common/components/Tag';
@@ -16,7 +15,6 @@ import {
   isValidPartialDate,
 } from '@common/utils/date/partialDate';
 import React, { ReactNode } from 'react';
-import LazyLoad from 'react-lazyload';
 
 interface Props {
   release: Release;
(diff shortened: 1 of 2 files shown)


luke-h1 (approved):

src/explore-education-statistics-admin/src/pages/admin-dashboard/components/ReleaseSummary.tsx
```diff
@@ -40,21 +38,17 @@ const ReleaseSummary = ({
...
         <TagGroup className="govuk-!-margin-left-2">
-          <Tag>{getReleaseStatusLabel(release.status)}</Tag>
-
-          {release.amendment && <Tag>Amendment</Tag>}
-
+          {release.status !== 'Approved' && (
+            <Tag>{getReleaseStatusLabel(release.status)}</Tag>
+          )}
           {release.status === 'Approved' && (
-            <LazyLoad
```
markhiveit, line 48: With the lazy loading is removed, I think we're now requesting the status of every approved release on the dashboard as soon as the dashboard loads? I think the lazy loading was added to prevent this from happening, so we should keep it.

markhiveit (changes requested):

src/explore-education-statistics-admin/src/pages/admin-dashboard/components/ReleaseSummary.tsx
markhiveit: Seems it is requesting the status for every approved release under the topic currently selected - not sure if we should consider that an issue or not.

amyb-hiveit: @markhiveit I've restored the lazy loading now. Had a look at stopping the endpoint being called twice, think it'll mean a bit of refactoring of the components so probably out of scope for this PR.

markhiveit (approved):

PR #2574, opened 2021-06-09
EES-2203 - mering in EES-2203 work from dev

duncan-at-hiveit:
This PR:
- merges the dev branch into the SOW4 branch, bringing with it the Auth Handler work for EES-2203


src/GovUk.Education.ExploreEducationStatistics.Content.Model/MethodologyParent.cs
```diff
@@ -1,9 +1,12 @@
...
+﻿﻿using System;
+using System.Collections.Generic;
 
 namespace GovUk.Education.ExploreEducationStatistics.Content.Model
 {
     public class MethodologyParent
     {
         public Guid Id { get; set; }
+        
+        public List<PublicationMethodology> Publications { get; set; }
```
benoutram, line 10: I think we should change ContentDbContext to configure this in this side of the relationship with something like `.WithMany(parent => parent.Publications)`

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Model/MethodologyParent.cs
duncan-at-hiveit: Agreed!

PR #2575, opened 2021-06-09
EES-2201 methodology info and tags

amyb-hiveit:
Show methodology info, tags and view & amend buttons on the admin dashboard.

I've put in a fake permissions check for that'll need to be updated later and have made some assumptions about param names so they may also need ot be changed.


benoutram (approved):
Tried this out and it looks good with the different statuses:

image

image

PR #2579, opened 2021-06-09
Ees 2220 support for new create methodology process

duncan-at-hiveit:


src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ConfigurationController.cs
```diff
@@ -19,7 +19,7 @@ public ConfigurationController(IConfiguration configuration)
         [HttpGet("api/config")]
         public ActionResult<Dictionary<string, string>> GetConfig()
         {
-            return Ok(BuildConfigurationValues());
+            return new ActionResult<Dictionary<string, string>>(BuildConfigurationValues());
```
benoutram, line 22: I think you can just `return BuildConfigurationValues();` because of the `ActionResult<T>` implicit operator that's built-in and will do it for you.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyRepository.cs
```diff
@@ -21,6 +21,29 @@ public class MethodologyRepository : IMethodologyRepository
             _methodologyParentRepository = methodologyParentRepository;
         }
 
+        public async Task<Methodology> CreateMethodologyForPublication(Guid publicationId)
+        {
+            var methodology = (await _contentDbContext.Methodologies.AddAsync(new Methodology
+            {
+                Slug = publicationId.ToString(),
```
benoutram, line 28: Is this temporary until we get the slug from publication title or alternative title? Also as discussed I think we should move slug to MethodologyParent and migrate the existing slugs but could do that as a separate ticket / PR. This would benefit EES-2375 which needs to look up by slug.

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ConfigurationController.cs
duncan-at-hiveit: Hey thanks @benoutram - apparently not, as it thinks you're attempting to use the protected (non-generic) ActionResult constructor.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyRepository.cs
duncan-at-hiveit: Yes, temporary :D  I intend to migrate to MethodologyParent as part of my next ticket.

PR #2584, opened 2021-06-14
EES-2361 fix reselecting the same highlight in table tool

amyb-hiveit:
Currently in the table tool if you select a table highlight, then go back to step 1, select the same publication then the same highlight you get stuck on step 2 instead of showing the highlight.

This fixes it by forcing the Wizard to update the current step in this circumstance. Which feels a bit hacky... so any other suggestions are welcome.


amyb-hiveit:
Closed in favour of https://github.com/dfe-analytical-services/explore-education-statistics/pull/2589

PR #2597, opened 2021-06-16
Ees 2411 - Various UI test failure fixes

luke-h1:
This PR fixes a variety of UI tests that are currently failing in CI due to timeout-related issues. In addition to this, an argument has been added to the run tests bash script to allow pabot to make use of more than 2 processes. This offers a nice performance boost when tests are running in CI

### DIFF
--- a/tests/robot-tests/tests/admin/bau/create_data_block_with_chart.robot
+++ b/tests/robot-tests/tests/admin/bau/create_data_block_with_chart.robot
@@ -391,29 +391,29 @@ Validate line chart embeds correctly
     user checks chart tooltip item contains  ${datablock}  1  Admission Numbers (Nailsea Youngwood): 4,198
 
 Configure basic vertical bar chart
-    [Tags]  HappyPath
+    [Tags]  HappyPath  NotAgainstDev
     user goes to url  ${DATABLOCK_URL}
 
-    user waits until h2 is visible  ${DATABLOCK_NAME}
+    user waits until h2 is visible  ${DATABLOCK_NAME}  120
     user waits until page does not contain loading spinner
 
     user clicks link  Chart
     user configures basic chart   Vertical bar  500  800
 
 Change vertical bar chart legend
-    [Tags]  HappyPath
+    [Tags]  HappyPath  NotAgainstDev
     user clicks link  Legend
-    user waits until h3 is visible  Legend
+    user waits until h3 is visible  Legend  60
 
     user counts legend form item rows  1
-    user checks element value should be  id:chartLegendConfigurationForm-items-0-label  Admission Numbers (Nailsea Youngwood)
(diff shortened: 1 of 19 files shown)


tests/robot-tests/tests/admin_and_public/bau/release_and_publication_role_permissions.robot
```diff
@@ -202,11 +204,15 @@ Remove publication owner access
     [Tags]  HappyPath
     user waits until element is enabled  css:[name="selectedPublicationId"]
     user scrolls to element  css:[name="selectedPublicationId"]
-    user clicks element  testid:remove-publication-role-${PUBLICATION_NAME}
-    user waits until page does not contain loading spinner
+    # NOTE: The below wait is to prevent a transient failure that occurs on the UI test pipeline due to the DOM not being fully rendered which
```
duncan-at-hiveit, line 207: Can this not be achieved with:

```
user waits until testid is visible       remove-publication-role-${PUBLICATION_NAME}
```

rather than the wait?

luke-h1: Afraid not. I tried the above and various other waits but it seems that the only way to stop CI failures is to add  `Sleep`

duncan-at-hiveit: No worries!

duncan-at-hiveit (approved):

tests/robot-tests/tests/admin/bau/create_data_block_with_chart.robot
```diff
@@ -391,29 +391,29 @@ Validate line chart embeds correctly
     user checks chart tooltip item contains  ${datablock}  1  Admission Numbers (Nailsea Youngwood): 4,198
 
 Configure basic vertical bar chart
-    [Tags]  HappyPath
+    [Tags]  HappyPath  NotAgainstDev
     user goes to url  ${DATABLOCK_URL}
 
-    user waits until h2 is visible  ${DATABLOCK_NAME}
+    user waits until h2 is visible  ${DATABLOCK_NAME}  120
```
duncan-at-hiveit, line 397: I'd recommend as a follow-up piece of work that we assign these wait times to parameters that are configurable e.g. 120 would be ${LONG_WAIT}, 60 would be ${MEDIUM_WAIT} etc.

My driver for this is that if I'm running locally I don't wanna wait 2 minutes to be alerted of a failure, otherwise that'd slow down my own test writing.

duncan-at-hiveit (approved):

tests/robot-tests/tests/admin/bau/create_data_block_with_chart.robot
luke-h1: yeah, that's a good point. Will create a task for this

PR #2599, opened 2021-06-17
EES-1277 show confirm modal when edit publication

amyb-hiveit:
Screenshot (8)

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/publication/components/PublicationForm.tsx
+++ b/src/explore-education-statistics-admin/src/pages/publication/components/PublicationForm.tsx
@@ -15,9 +15,10 @@ import useFormSubmit from '@common/hooks/useFormSubmit';
 import { OmitStrict } from '@common/types';
 import { mapFieldErrors } from '@common/validation/serverValidations';
 import Yup from '@common/validation/yup';
+import ModalConfirm from '@common/components/ModalConfirm';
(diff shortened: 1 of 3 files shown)


src/explore-education-statistics-admin/src/pages/publication/components/PublicationForm.tsx
```diff
@@ -201,140 +208,169 @@ const PublicationForm = ({
...
-            <Button type="submit">Save publication</Button>
-            {cancelButton}
-          </ButtonGroup>
-        </Form>
+            <ButtonGroup>
+              <Button
+                type="submit"
+                onClick={e => {
+                  e.preventDefault();
+                  if (confirmOnSubmit && form.isValid) {
```
markhiveit, line 346: I think if confirmOnSubmit is true and form.isValid is false, then the form will submit?

markhiveit (changes requested):

src/explore-education-statistics-admin/src/pages/publication/components/PublicationForm.tsx
amyb-hiveit: Yes, but then it'll show the validation errors and not save the changes. I figured it was better to show the validation errors before the confirm modal as it'd be a bit annoying otherwise.

markhiveit (approved):

PR #2601, opened 2021-06-17
EES-1256 prevent unnecessary chart re-renders

amyb-hiveit:
I'm not sure about this one... the code I've removed here was forcing charts to re-render when any of their configuration props changed, for example editing the title. For infographics this meant that the image was being fetched on each re-render. For other chart types it causes a brief flicker as the chart disappeared and reappeared.

I've tested changing config options for charts, maps and infographics, including resizing them, and they all seem to update correctly.

But clearly it was added for a reason so it may have been decided at the time that the flickering & fetches were the lesser evil.

Thoughts and opinions welcome 😃


markhiveit (approved):
I think merge this and make sure it gets tested thoroughly?

markhiveit:
Waiting on a local test

luke-h1:
Tested & seems fine

luke-h1 (approved):

PR #2602, opened 2021-06-18
EES-2375 EES-2182 EES-283 Content updates to support multiple methodologies and multiple versions 

benoutram:
This PR makes changes to the way we publish and view methodologies. 

Previously we cached a single Methodology with the Publication. This cached document was fetched along with the Release when serving up a Release as a combination of the two by the Content API.

Now a Publication can have multiple methodologies each of which can have multiple versions. Those methodologies have titles which can vary per version. This means we need to work out the latest version of all the related methodologies.

The content API now serves the Publication and Release from the cached documents as before, but also fetches with the latest versions of each of the related methodologies of the Publication. This includes their titles so that each link can display it's own title.

This PR also makes the same changes to the Admin Release content page due to the way that the common `Publication` view model now has a methodologies array.

### Other changes

- Drop the unused Summary field from Methodology
- Remove unused methods and view models


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyRepositoryTests.cs
```diff
@@ -148,6 +158,51 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+                        Owner = true
+                    });
+                await contentDbContext.SaveChangesAsync();
+            }
+
+            var methodologyParentRepository = new Mock<IMethodologyParentRepository>(MockBehavior.Strict);
+
+            await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContextId))
+            {
+                contentDbContext.Attach(methodology);
```
duncan-at-hiveit, line 191: Ooh, what's this in aid of?

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Services/MethodologyServiceTests.cs
```diff
@@ -0,0 +1,346 @@
...
+                await contentDbContext.SaveChangesAsync();
+            }
+
+            await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId))
+            {
+                var service = SetupMethodologyServiceTests(contentDbContext: contentDbContext);
+
+                var result = await service.Get(slug);
+
+                Assert.True(result.IsRight);
```
duncan-at-hiveit, line 108: You can also use:

```
var methodology = result.AssertRight();
```

as a short-hand if you want?  Hot off the press!

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Controllers/MethodologyController.cs
```diff
@@ -9,15 +11,18 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Api.Controllers
...
         {
+            _methodologyService = methodologyService;
         }
 
         [HttpGet("methodologies/{slug}")]
         public async Task<ActionResult<MethodologyViewModel>> Get(string slug)
         {
-            // TODO SOW4 EES-2375 Return methodology from content database
-            return new NotFoundResult();
+            return await _methodologyService.Get(slug)
```
duncan-at-hiveit, line 24: Tempted to say rename `Get()` to something that reflects that it's getting the latest version of that Methodology that uses that slug, but not a biggie

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Services/MethodologyService.cs
```diff
@@ -0,0 +1,69 @@
...
+            _mapper = mapper;
+            _methodologyRepository = methodologyRepository;
+        }
+
+        public async Task<Either<ActionResult, MethodologyViewModel>> Get(string slug)
+        {
+            // TODO SOW4 EES-2375 lookup the MethodologyParent by slug when slug is moved to the parent
+            // For now, this does a lookup on the parent via any Methodology with the slug
+            return await _persistenceHelper
+                .CheckEntityExists<Methodology>(
```
duncan-at-hiveit, line 39: Nice use of the existence check and hydration!

src/GovUk.Education.ExploreEducationStatistics.Content.Model/MethodologyParent.cs
```diff
@@ -10,18 +12,56 @@ public class MethodologyParent
...
+        public bool IsLatestPublishedVersionOfMethodology(Guid methodologyVersionId)
+        {
+            if (Versions == null)
+            {
+                throw new ArgumentException("Methodology must be hydrated with Versions to test the latest published version");
+            }
+
+            var version = Versions.SingleOrDefault(mv => mv.Id == methodologyVersionId);
+
+            return version is {PubliclyAccessible: true}
```
duncan-at-hiveit, line 62: Ooh what's this syntax?   `version is {PubliclyAccessible: true}`

Is there a reason why not to do `version.PubliclyAccessible` instead?

src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseHelpAndSupportSection.tsx
```diff
@@ -30,28 +30,32 @@ const ReleaseHelpAndSupportSection = ({
...
-                (isEditing ? (
-                  <a>{`${publication.title}: methodology`}</a>
-                ) : (
-                  <Link to={publication.externalMethodology.url}>
-                    {`${publication.title}: methodology`}
-                  </Link>
-                ))}{' '}
-              guidance.
-            </p>
+          {publication.methodologies.length > 0 ||
```
duncan-at-hiveit, line 33: Can just be `publication.methodologies.length` I think - a length of zero is falsy

benoutram: Thanks I've made this change now

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Controllers/MethodologyController.cs
benoutram: Thanks I've renamed this to GetLatestMethodologyBySlug now

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Services/MethodologyServiceTests.cs
benoutram: Thanks I saw that incoming change and was meaning to do this but forgot! I've changed this test class to use this now 🙂 .

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyRepositoryTests.cs
benoutram: The test uses a mock for the `MethodologyParentRepository`.
The `methodology` (parent) is returned by that `MethodologyParentRepository` mock when the service calls `_methodologyParentRepository.GetByPublication(publication.Id)`

The service then goes on to lazily load `methodology.Versions` from the database.

So this test sets up the versions so they can be loaded, but the database needs to be aware of `methodology` otherwise you get an error like 

> Navigation property 'Versions' on entity of type 'MethodologyParent' cannot be loaded because the entity is not being tracked.

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Services/MethodologyService.cs
benoutram: Thanks 🙂

src/GovUk.Education.ExploreEducationStatistics.Content.Model/MethodologyParent.cs
benoutram: We do want to do `version.PubliclyAccessible` but we also want to check for version being null too as the `id` passed to this method could theoretically be none of the versions and completely unrelated so we should return false if that's the case.

So this `is` syntax with the pattern expression is null-safe and is an alternative of 

`version != null && version.PubliclyAccessible`

I didn't know about it but the IDE suggested it as soon as I wrote it so I accepted it 🙂.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/is

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyRepositoryTests.cs
duncan-at-hiveit: cool nice one!

src/GovUk.Education.ExploreEducationStatistics.Content.Model/MethodologyParent.cs
duncan-at-hiveit: Nice one!  I guess is this also the equivalent of version?.PubliclyAccessible   ?

PR #2603, opened 2021-06-18
EES-1890 Change Content DB to use vCores

markhiveit:


infrastructure/templates/template.json
```diff
@@ -2021,11 +2033,22 @@
           "dependsOn": [
             "[variables('sqlserverName')]"
           ],
+          "sku": {
```
duncan-at-hiveit, line 2036: I take it that these are the same settings as we use on the stats dbs?

markhiveit: yeah - i've just changed the max and min number of vCores - according to the azure docs 1 vCore ~= 100 DTUs, so I've set the min on prod to be 2 vCores and max 5, and then min 1 and max 1 on all other environments for now.

When we do the deploy i'll keep a close eye on it so we can bump it manually if it ends up being too slow

luke-h1 (approved):

PR #2611, opened 2021-06-22
EES-2413 toggle set page view box

amyb-hiveit:
Open:
Screenshot (15)

 Closed:
Screenshot (17)

### DIFF
--- a/src/explore-education-statistics-admin/src/components/editable/EditablePageModeToggle.module.scss
+++ b/src/explore-education-statistics-admin/src/components/editable/EditablePageModeToggle.module.scss
@@ -1,12 +1,13 @@
 @import '~govuk-frontend/govuk/base';
-/* stylelint-disable max-nesting-depth */
+
 .toggle {
   background: $govuk-brand-colour;
   bottom: 0;
   left: 0;
+  max-width: 390px;
   padding: 6px;
   position: fixed;
-  width: 390px;
+  width: 100%;
   z-index: 10000;
 
   @media print {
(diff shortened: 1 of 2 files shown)


src/explore-education-statistics-admin/src/components/editable/EditablePageModeToggle.module.scss
```diff
@@ -1,11 +1,12 @@
 @import '~govuk-frontend/govuk/base';
-
+/* stylelint-disable max-nesting-depth */
```
ntsim, line 2: Shouldn't need to nest as we can hoist things like `open` up to the top level. As we're in a CSS module, there aren't any potential scoping issues.

src/explore-education-statistics-admin/src/components/editable/EditablePageModeToggle.module.scss
```diff
@@ -1,11 +1,12 @@
...
 .toggle {
   background: $govuk-brand-colour;
   bottom: 0;
   left: 0;
-  padding: 0 6px 6px;
+  padding: 6px;
   position: fixed;
+  width: 390px;
```
ntsim, line 9: Does this need a specific width? If it's for the button, would flex work instead?

ntsim (changes requested):

src/explore-education-statistics-admin/src/components/editable/EditablePageModeToggle.module.scss
amyb-hiveit: It's a bit awkward because it needs some specificity to override the govuk styles, but I've rejigged it so the linter is happy.

src/explore-education-statistics-admin/src/components/editable/EditablePageModeToggle.module.scss
amyb-hiveit: I think it does, otherwise the button changes size when the content is hidden. I've tweaked it to be more responsive though

ntsim (approved):

PR #2613, opened 2021-06-23
EES-2423 Remove unused InternalReleaseNote column from Releases table

markhiveit:
Holding back merging of this until EES-1526 is successfully deployed to production.


benoutram (approved):
We are waiting until the previous internal note migration has been deployed before merging this PR so we have the chance to fix anything with that before the data is dropped. It's why we decided to do this as a separate change.

PR #2614, opened 2021-06-23
EES-2429 Correctly order TimeIdentifiers

markhiveit:


benoutram (commented):
There's something not quite right here because as we discussed, the unit tests being added here pass without any code change, so either it isn't fixing the bug that's been reported or the tests aren't enough to demonstrate the problem and prevent it from being reintroduced.

benoutram:
So I've taken another look at this and it's the overload `GetTimePeriodRange(IQueryable<Observation> observations)` that seems to have the bug in it, which is fixed by your change but weirdly I can't reproduce the problem before the change by unit testing that method.

Anyway, since you made changes to GetTimePeriodRange(IQueryable<Observation> observations)  as well I think you should add some unit tests for it.

benoutram (approved):

PR #2615, opened 2021-06-23
EES-2378 Reinstate support for the public View All Methodologies page and support multiple Methodologies

benoutram:
This PR reinstates the back end support for the public View All Methodologies view produced from the Content API.

It also alters the public view to support a Publication having multiple methodologies.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Api/Services/MethodologyService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Api/Services/MethodologyService.cs
@@ -9,6 +9,7 @@ using GovUk.Education.ExploreEducationStatistics.Content.Api.Services.Interfaces
 using GovUk.Education.ExploreEducationStatistics.Content.Api.ViewModels;
 using GovUk.Education.ExploreEducationStatistics.Content.Model;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
+using GovUk.Education.ExploreEducationStatistics.Content.Model.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Repository.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Publisher.Model.ViewModels;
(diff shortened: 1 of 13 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Api/Services/MethodologyService.cs
```diff
@@ -65,5 +69,105 @@ public class MethodologyService : IMethodologyService
...
+            // TODO SOW4 Potentially remove this check on Releases in future
+            var hasReleases = publication.Releases.Any(release => release.IsLatestPublishedVersionOfRelease());
+
+            if (hasReleases)
+            {
+                // TODO SOW4 There should be no LatestPublishedVersion of a Methodology returned if the Publication has no published releases
+                // removing the need for the outer check on hasReleases
+                var hasMethodologies = publication.Methodologies.Any(pm =>
+                    pm.MethodologyParent.LatestPublishedVersion() != null);
+                return hasMethodologies;
```
duncan-at-hiveit, line 155: Why does having published methodologies indicate that a Publication is published?  Does this mean that if it had lots of published Releases but still the Publication didn't have a Methodology attached, the Publication would be considered not published?

benoutram: Yes that's correct, just in the context of producing this Publication tree of methodologies, we say a Publication isn't published so that its not included in the result if it has no Methodologies.

The `IsXPublished` methods are filtering out nodes which don't have any leaf (methodology) nodes so no blank Themes/Topics/Publications are listed on the Methodologies page.

I believe the front end is also doing this kind of filtering although it shouldn't be required because there should never be a Theme/Topic/Publication with no methodologies returned by this service.

I think this method name made sense when producing the Publications tree for Releases, but not here so I'll rename it to something like `IsPublicationNodeIncluded`

duncan-at-hiveit (approved):

PR #2617, opened 2021-06-23
EES-2448 Only create ReleaseStatus row if ApprovalStatus changes

markhiveit:
The same release update endpoint is being used on three different pages: to change a release's ApprovalStatus, to add/edit PreRelease access list content, and to alter a release's summary info. 

By only adding a new ReleaseStatus row when the ApprovalStatus changes, we ensure that we only add a new ReleaseStatus entry when the user is on the Sign off page. This prevents the endpoint erroring if it is used to add/edit an access list or edit a release's summary info.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
@@ -733,14 +733,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                 Assert.Equal(TimeIdentifier.March, saved.TimePeriodCoverage);
                 Assert.Equal("New access list", saved.PreReleaseAccessList);
 
-                Assert.Single(saved.ReleaseStatuses);
-                var savedStatus = saved.ReleaseStatuses[0];
-                Assert.Equal(ReleaseApprovalStatus.Draft, savedStatus.ApprovalStatus);
-                Assert.Equal("Test internal note", savedStatus.InternalReleaseNote);
-                Assert.NotNull(savedStatus.Created);
-                Assert.InRange(DateTime.UtcNow
-                    .Subtract(savedStatus.Created.Value).Milliseconds, 0, 1500);
-                Assert.Equal(_userId, savedStatus.CreatedById);
+                // No ReleaseStatus created if the ApprovalStatus hasn't changed
+                Assert.Empty(saved.ReleaseStatuses);
             }
 
             MockUtils.VerifyAllMocks(contentService, releaseFileService);
@@ -1191,7 +1185,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                             ReleaseName = "2035",
                             TimePeriodCoverage = TimeIdentifier.March,
                             PreReleaseAccessList = "New access list",
-                            ApprovalStatus = ReleaseApprovalStatus.Draft,
+                            ApprovalStatus = ReleaseApprovalStatus.HigherLevelReview,
                             LatestInternalReleaseNote = "Internal note"
 
                         }
@@ -1226,7 +1220,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
 
                 Assert.Single(saved.ReleaseStatuses);
                 var savedStatus = saved.ReleaseStatuses[0];
-                Assert.Equal(ReleaseApprovalStatus.Draft, savedStatus.ApprovalStatus);
+                Assert.Equal(ReleaseApprovalStatus.HigherLevelReview, savedStatus.ApprovalStatus);
                 Assert.Equal("Internal note", savedStatus.InternalReleaseNote);
                 Assert.NotNull(savedStatus.Created);
                 Assert.InRange(DateTime.UtcNow
@@ -1340,7 +1334,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                             ReleaseName = "2035",
                             TimePeriodCoverage = TimeIdentifier.March,
                             PreReleaseAccessList = "New access list",
-                            ApprovalStatus = ReleaseApprovalStatus.Draft,
+                            ApprovalStatus = ReleaseApprovalStatus.HigherLevelReview,
                             LatestInternalReleaseNote = "Test internal note"
                         }
                     );
@@ -1381,7 +1375,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
 
                 Assert.Single(saved.ReleaseStatuses);
                 var savedStatus = saved.ReleaseStatuses[0];
-                Assert.Equal(ReleaseApprovalStatus.Draft, savedStatus.ApprovalStatus);
+                Assert.Equal(ReleaseApprovalStatus.HigherLevelReview, savedStatus.ApprovalStatus);
                 Assert.Equal("Test internal note", savedStatus.InternalReleaseNote);
                 Assert.NotNull(savedStatus.Created);
                 Assert.InRange(DateTime.UtcNow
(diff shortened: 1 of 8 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseService.cs
```diff
@@ -325,6 +325,7 @@ public class ReleaseService : IReleaseService
 
                     var releaseStatus = new ReleaseStatus
                     {
+                        Id = Guid.NewGuid(),
```
benoutram, line 328: How come you set an id here now?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseService.cs
```diff
@@ -337,7 +338,12 @@ public class ReleaseService : IReleaseService
                         .OnSuccess(async () =>
                         {
                             _context.Releases.Update(release);
-                            await _context.AddAsync(releaseStatus);
+
+                            if (oldStatus != request.ApprovalStatus)
```
benoutram, line 342: I think an approver can change the release note when a Release is already scheduled.  I think this change might stop this or force them to unapprove first before reapproving with a different note so worth checking it.

I.e, they won't be able to do this, and the change would be silently ignored I think? 

image

markhiveit: We discussed this issue on standup and it was decided that a new endpoint would be created to update the release from the sign off page. See EES-2448 ticket on Jira.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
```diff
@@ -1088,6 +1082,278 @@ await using (var context = InMemoryApplicationDbContext(contextId))
...
+                        {
+                            PublishMethod = PublishMethod.Scheduled,
+                            PublishScheduled = "2051-06-30",
+                            NextReleaseDate = nextReleaseDateEdited,
+                            ApprovalStatus = ReleaseApprovalStatus.Draft,
+                            LatestInternalReleaseNote = "Test internal note"
+                        }
+                    );
+
+                Assert.True(result.IsRight);
```
duncan-at-hiveit, line 1141: You can use:

```
var updatedRelease = result.AssertRight();
```

instead for brevity.

duncan-at-hiveit: Then you also don't need to use `result.Right.` all the time below either

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
```diff
@@ -1088,6 +1082,278 @@ await using (var context = InMemoryApplicationDbContext(contextId))
...
+                Assert.Equal(new DateTime(2051, 6, 29, 23, 0, 0, DateTimeKind.Utc),
+                    saved.PublishScheduled);
+                Assert.Equal(nextReleaseDateEdited, saved.NextReleaseDate);
+                Assert.Equal(adHocReleaseType.Id, saved.TypeId);
+                Assert.Equal("2030-march", saved.Slug);
+                Assert.Equal("2030", saved.ReleaseName);
+                Assert.Equal(TimeIdentifier.March, saved.TimePeriodCoverage);
+                Assert.Equal("Access list", saved.PreReleaseAccessList);
+
+                // No ReleaseStatus created if the ApprovalStatus hasn't changed
```
duncan-at-hiveit, line 1169: Think the comment is incorrect?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
```diff
@@ -1088,6 +1082,278 @@ await using (var context = InMemoryApplicationDbContext(contextId))
...
+                },
+                Publication = new Publication
+                {
+                    Title = "Old publication",
+                },
+                ReleaseName = "2030",
+                Slug = "2030",
+                Published = DateTime.Now,
+                PublishScheduled = DateTime.UtcNow,
+                //Version = 0,
```
duncan-at-hiveit, line 1264: Remove this?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
```diff
@@ -1088,6 +1082,278 @@ await using (var context = InMemoryApplicationDbContext(contextId))
...
+                var result = await releaseService
+                    .UpdateReleaseStatus(
+                        release.Id,
+                        new ReleaseStatusUpdateViewModel
+                        {
+                            ApprovalStatus = ReleaseApprovalStatus.Draft,
+                        }
+                    );
+
+                Assert.True(result.IsLeft);
```
duncan-at-hiveit, line 1350: In the same way we have `var xxx = result.AssertRight()`, it'd probably be nice to have an `result.AssertValidationProblems(prob1, prob2, prob3)` that does both the Left check, and also does 1 or more `AssertValidationProblem` checks as well.

Not suggesting as part of this PR though!

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
```diff
@@ -1191,7 +1457,7 @@ await using (var context = InMemoryApplicationDbContext(contextId))
                             ReleaseName = "2035",
                             TimePeriodCoverage = TimeIdentifier.March,
                             PreReleaseAccessList = "New access list",
-                            ApprovalStatus = ReleaseApprovalStatus.Draft,
+                            ApprovalStatus = ReleaseApprovalStatus.HigherLevelReview,
```
duncan-at-hiveit, line 1460: Why did these need to change out of interest?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseService.cs
```diff
@@ -358,6 +364,72 @@ public class ReleaseService : IReleaseService
...
+                            if (oldStatus == ReleaseApprovalStatus.Approved || request.ApprovalStatus == ReleaseApprovalStatus.Approved)
+                            {
+                                await _publishingService.ReleaseChanged(
+                                    releaseId,
+                                    releaseStatus.Id,
+                                    request.PublishMethod == PublishMethod.Immediate
+                                );
+                            }
+
+                            _context.Update(release); // TODO: EES-2450 Remove?
```
duncan-at-hiveit, line 425: Do we need these TODOs?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
markhiveit: I think this is only on the SOW branch at the moment?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
markhiveit: To prevent the bug when trying to create an prerelease access list from occurring, the UpdateRelease method still needs to only add a ReleaseStatus if the ApprovalStatus has changed - even though nothing using that endpoint will change the ApprovalStatus.

Once UpdateRelease is updated as part of the fixup ticket after this PR is merged, ReleaseUpdateViewModel will no longer take an ApprovalStatus and so this will no longer be an issue. So, to finally answer your question, I needed to do this to make the test pass. :)

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseService.cs
markhiveit: I've removed them

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
duncan-at-hiveit: Oh sorry!

duncan-at-hiveit (approved):

PR #2619, opened 2021-06-24
Ees 2452 - Add robot framework formatter

luke-h1:


.husky/pre-commit
```diff
@@ -2,3 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 ./node_modules/.bin/lint-staged
+robotidy --configure DiscardEmptySections:allow_only_comments=True --spacecount 2 tests/robot-tests/
```
ntsim, line 5: Would be better to add this  `package.json` under a `*.robot` wildcard pattern, otherwise it will be ran on the entire directory for every single commit.

Think this would do the trick:

```json
"*.robot": ["robotidy --configure DiscardEmptySections:allow_only_comments=True --spacecount 2"]
```

ntsim (changes requested):

luke-h1:
Putting do not merge on this PR for now as this PR will create a lot of merge conflicts on the SOW4 branch.

PR #2620, opened 2021-06-24
EES-2385 Publish Methodology image files prior to a Methodology being made publicly accessible

benoutram:
This PR is an update for recent changes to Methodologies to make sure that the Methodology image files are published at the correct time prior to a Methodology being made publicly accessible.

We publish methodology files when a methodology is approved that's going to be immediately accessible, or delay publishing them until a Release is published that's going to mean a Methodology related to that Release should then be accessible.

This change depends on implementing the `PublishingStrategy` field of  `Methodology` in the database and set all existing Methodologies to use the `Immediately` strategy.

### Other Changes

- `PublishMethodologyFunction` is renamed `PublishMethodologyFilesFunction` with similar renaming for the storage queue, the message class and all methods and tests related to it. 
- Some additional tests are added to `PublishingServiceTests`.
- The `manage methodologies` link is removed from the Admin UI.


src/GovUk.Education.ExploreEducationStatistics.Publisher.Model/PublisherQueues.cs
```diff
@@ -4,7 +4,7 @@ public static class PublisherQueues
     {
         public const string GenerateReleaseContentQueue = "generate-release-content";
         public const string PublishAllContentQueue = "publish-all-content";
-        public const string PublishMethodologyQueue = "publish-methodology";
+        public const string PublishMethodologyFilesQueue = "publish-methodology-files";
```
duncan-at-hiveit, line 7: Do we just need to make a note at deploy time to ensure there's nothing in the original queue?  And to remove the old queue afterwards?

duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Publisher.Model/PublisherQueues.cs
benoutram: Yes good point, I've started an SOW4 Milestone 1 deploy checklist which I will share. I've added this to it. Thanks.

PR #2621, opened 2021-06-24
Ees 2222 fix UI tests on dev before tests on 2222 branch

duncan-at-hiveit:
This PR:
- fixes up the rest of the latest UI test failures on the dev branch (and marks a few as Failing until later work)
- adds additional rerun failed tests support by adding a Prerebot Model Modifier that marks a test as PASSing if it passed on either the original run or the rerun, and merges that result into the rerun's report file

image

### DIFF
--- a/tests/robot-tests/tests/general_public/table_tool_absence_by_characteristic.robot
+++ b/tests/robot-tests/tests/general_public/table_tool_absence_by_characteristic.robot
@@ -240,15 +240,15 @@ Validate rows after reordering
 
 User generates a permanent link
     [Tags]   HappyPath
-    user waits until page contains button  Share your table 
-    user clicks button  Share your table
+    user waits until page contains button  Generate shareable link 
+    user clicks button  Generate shareable link
     user waits until page contains element  xpath://a[text()="View share link"]  60
     user checks generated permalink is valid
 
 User validates permanent link works correctly
     [Tags]   HappyPath
     user clicks link   View share link
-    select window    NEW
+    switch window   locator=NEW
     user waits until h1 is visible  'Absence by characteristic' from 'Pupil absence in schools in England'
 
 User validates permalink table
--- a/tests/robot-tests/tests/general_public/table_tool_absence_by_characteristic.robot
+++ b/tests/robot-tests/tests/general_public/table_tool_absence_by_characteristic.robot
@@ -248,7 +248,7 @@ User generates a permanent link
 User validates permanent link works correctly
     [Tags]   HappyPath
     user clicks link   View share link
-    switch window   locator=NEW
+    user selects newly opened window
     user waits until h1 is visible  'Absence by characteristic' from 'Pupil absence in schools in England'
 
 User validates permalink table
--- a/tests/robot-tests/tests/general_public/table_tool_exclusions_by_geographic_level.robot
+++ b/tests/robot-tests/tests/general_public/table_tool_exclusions_by_geographic_level.robot
@@ -101,22 +101,21 @@ Validate Bury Number of fixed period exclusions row
 
 User generates a permanent link
     [Tags]   HappyPath
-    user waits until page contains button  Share your table  60
-    user clicks button    Share your table
+    user waits until page contains button  Generate shareable link  60
+    user clicks button    Generate shareable link
     user waits until page contains testid  permalink-generated-url
     user checks generated permalink is valid
 
 User validates permanent link works correctly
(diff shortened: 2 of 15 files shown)


tests/robot-tests/tests/libs/admin-common.robot
```diff
@@ -53,6 +53,7 @@ user signs out
 
 user selects theme and topic from admin dashboard
     [Arguments]  ${theme}  ${topic}
+    user goes to url  %{ADMIN_URL}
```
markhiveit, line 56: I think in most instances this keyword is used, it is assumed that the user is already on the admin dashboard. So I think this is going to be redundant in most cases. I think I'd prefer that the test suite itself navigates the user to the dashboard.

tests/robot-tests/tests/general_public/table_tool_exclusions_by_geographic_level.robot
```diff
@@ -101,22 +101,21 @@ Validate Bury Number of fixed period exclusions row
...
+    user waits until page contains button  Generate shareable link  60
+    user clicks button    Generate shareable link
     user waits until page contains testid  permalink-generated-url
     user checks generated permalink is valid
 
 User validates permanent link works correctly
     [Tags]   HappyPath
     user clicks link   View share link
-    select window    NEW
+    switch window   locator=NEW
```
markhiveit, line 112: Can use the new `user selects newly opened window` keyword you've created here.

tests/robot-tests/tests/general_public/table_tool_absence_by_characteristic.robot
```diff
@@ -240,15 +240,15 @@ Validate rows after reordering
...
+    user waits until page contains button  Generate shareable link 
+    user clicks button  Generate shareable link
     user waits until page contains element  xpath://a[text()="View share link"]  60
     user checks generated permalink is valid
 
 User validates permanent link works correctly
     [Tags]   HappyPath
     user clicks link   View share link
-    select window    NEW
+    switch window   locator=NEW
```
markhiveit, line 251: Can use the new `user selects newly opened window` keyword you've created here.

luke-h1 (dismissed):

tests/robot-tests/tests/libs/utilities.py
```diff
@@ -283,4 +293,4 @@ def user_clicks_element_if_exists(selector):
 
 def user_chooses_what_to_do_on_step_failure(selector):
```
markhiveit, line 294: I know this isn't a change made in your PR, but it seems like this keyword isn't being used and can be removed?

markhiveit (changes requested):

tests/robot-tests/tests/libs/admin-common.robot
duncan-at-hiveit: I'd prefer to keep it because in quite a few cases it stops cascade failures when one step fails which in turn stops the user navigating to the dashboard, and then causes multiple failures off the back of it that could be avoided.

How about I make it conditional so that if you're already on the dashboard, it doesn't navigate there?

tests/robot-tests/tests/general_public/table_tool_exclusions_by_geographic_level.robot
duncan-at-hiveit: Noice

tests/robot-tests/tests/general_public/table_tool_absence_by_characteristic.robot
duncan-at-hiveit: Noice thanks

tests/robot-tests/tests/libs/utilities.py
duncan-at-hiveit: Cool I'll have a scan around and sanity check, ta.

duncan-at-hiveit: Yup agree I can't see it being used.  Begone...

tests/robot-tests/tests/libs/admin-common.robot
markhiveit: > I'd prefer to keep it because in quite a few cases it stops cascade failures

Cool - not that about that kind of thing before - makes sense. :)

That sounds good. To be honest, if you leave it I don't think its a big deal.

markhiveit (approved):

PR #2625, opened 2021-06-25
EES-2450 Change UpdateRelease into UpdateReleaseInfo

markhiveit:
After EES-2448, the ReleaseUpdate method still allowed the changing of the ApprovalStatus of a release even though the frontend didn't use this functionality, and if it were to be used to publish a release it would lead to bugs because the backend now only expects releases to be published with the ReleaseStatusUpdate method.

This PR fixes this issue by changing ReleaseUpdate to ReleaseUpdateInfo. This method is now only used on two pages - when editing a release's details (i.e. timeperiodcoverage, year, release type) and when adding or editing a prerelease access list.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleasesController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleasesController.cs
@@ -173,12 +173,12 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api
                 .HandleFailuresOrOk();
         }
 
-        [HttpPut("releases/{releaseId}/status")]
-        public async Task<ActionResult<ReleaseViewModel>> UpdateReleaseStatus(ReleaseStatusUpdateViewModel request,
+        [HttpPost("releases/{releaseId}/status")]
+        public async Task<ActionResult<ReleaseViewModel>> CreateReleaseStatus(ReleaseStatusCreateViewModel request,
             Guid releaseId)
         {
             return await _releaseService
-                .UpdateReleaseStatus(releaseId, request)
+                .CreateReleaseStatus(releaseId, request)
                 .HandleFailuresOrOk();
         }
 
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleasesController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleasesController.cs
@@ -164,12 +164,12 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api
                 .HandleFailuresOrOk();
         }
 
-        [HttpPut("releases/{releaseId}/info")]
-        public async Task<ActionResult<ReleaseViewModel>> UpdateReleaseInfo(ReleaseInfoUpdateViewModel request,
+        [HttpPut("releases/{releaseId}")]
+        public async Task<ActionResult<ReleaseViewModel>> UpdateRelease(ReleaseUpdateViewModel request,
             Guid releaseId)
         {
             return await _releaseService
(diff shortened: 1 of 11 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseService.cs
```diff
@@ -284,82 +284,24 @@ public class ReleaseService : IReleaseService
                 .OnSuccess(_mapper.Map<ReleasePublicationStatusViewModel>);
         }
 
-        public async Task<Either<ActionResult, ReleaseViewModel>> UpdateRelease(
-            Guid releaseId, ReleaseUpdateViewModel request)
+        public async Task<Either<ActionResult, ReleaseViewModel>> UpdateReleaseInfo(
```
ntsim, line 287: Why change this? Would suggest keeping it as `UpdateRelease` as this is still concerned with updating, well, a release.

Would also revert all of the other name changes accordingly.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleasesController.cs
```diff
@@ -164,12 +164,12 @@ public async Task<ActionResult<List<ReleaseStatusViewModel>>> GetReleaseStatuses
                 .HandleFailuresOrOk();
         }
 
-        [HttpPut("releases/{releaseId}")]
-        public async Task<ActionResult<ReleaseViewModel>> UpdateRelease(ReleaseUpdateViewModel request,
+        [HttpPut("releases/{releaseId}/info")]
```
ntsim, line 167: Wouldn't change the endpoint, would keep the semantics of this the same i.e. it updates a release.

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseService.cs
markhiveit: It isn't concerned with changing all aspects of a release, though - `UpdateRelease` gives the impression that it is capable of updating any aspect of a release which is no longer true. 

I changed the name to differentiate this update method from UpdateReleaseStatus - which is concerned with ApprovalStatus / publishing. The `UpdateReleaseInfo` method now only changes a few things on a release that aren't dependent on anything else. My idea is to demarcate the difference between "status" and "info" of a release, even if only conceptually.

I'd prefer to keep the new name - but its not a hill I'm going to die on so if you'd still prefer me to revert the name change I will.

ntsim: I'd say this creates too much specialisation in the method name when we should be using the DTO (`ReleaseUpdateViewModel`) to describe what can be updated on the release. This is the main benefit of using a static type system :wink: 
 
It could also be a little confusing to introduce the concept of `ReleaseInfo` as it is not backed by an actual database model. In reality, we're changing the `Release`, so it still just makes sense to call this `UpdateRelease`. Comparatively, `UpdateReleaseStatus` makes sense for its use-case and from a domain perspective (there is a `ReleaseStatus` in the database).

A wider issue with this naming is that it has been propagated through the code in a viral fashion and now distorts the REST API semantics. For example, it doesn't really make sense to be able to `POST /releases`, but then have to update the release via `PUT /releases/{id}/info`. A correct REST API implementation should absolutely allow consumers to `PUT /releases/{id}`. 

Similarly to the database, we also don't have an `info` resource that can be created/updated/deleted. This means there is no opportunity to re-use this endpoint for other purposes, and we are simply changing the API for no benefit. Note that if this was a well-versioned REST API, this would be a breaking change that would require a major version bump and would probably be an unacceptable change.

I'd suggest that we should revert this :slightly_smiling_face:

ntsim (approved):

PR #2628, opened 2021-06-28
EES-2222 - fixed logic around the check for ScheduledForPublishingWithPublishedRelease

duncan-at-hiveit:
This PR:
- swaps logic around the ScheduledForPublishingWithPublishedRelease check in Methodology to first check the publishing strategy, to avoid an early exception being thrown

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Model/Methodology.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Model/Methodology.cs
@@ -73,12 +73,17 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Model
         {
             get
             {
+                if (PublishingStrategy != MethodologyPublishingStrategy.WithRelease)
+                {
+                    return false;
+                }
+                
                 if (ScheduledWithReleaseId != null && ScheduledWithRelease == null)
                 {
                     throw new InvalidOperationException("ScheduledWithRelease field not included in Methodology");
                 }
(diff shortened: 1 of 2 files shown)


benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Methodology.cs
```diff
@@ -73,12 +73,17 @@ public bool ScheduledForPublishingWithPublishedRelease
...
+                }
+                
                 if (ScheduledWithReleaseId != null && ScheduledWithRelease == null)
                 {
                     throw new InvalidOperationException("ScheduledWithRelease field not included in Methodology");
                 }
-                return PublishingStrategy == MethodologyPublishingStrategy.WithRelease
-                       && ScheduledWithRelease.Live;
+                
+                return ScheduledWithRelease.Live;
```
benoutram, line 86: Is there a potential null reference exception here if `PublishingStrategy` is `WithRelease` but the ScheduledWithReleaseId isn't set? Looks like it would fall through to evaluating `ScheduledWithRelease.Live` with a null `ScheduledWithRelease`.

I don't think this matters though as `ScheduledWithReleaseId` should always be set with that strategy.

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Methodology.cs
duncan-at-hiveit: Agreed :)

PR #2631, opened 2021-06-29
EES-2427 Make Location Name columns case sensitive

markhiveit:
RscRegion has no associated Name, so we also make that column case sensitive


markhiveit:
Going to check for case sensitive matches in code instead.

PR #2635, opened 2021-06-29
EES-2215 updated methodology summary form

amyb-hiveit:
Adds a new version of the methodology summary form with an option to use the publication title or set an alternative one.

This is hidden under a flag so the current form will still work. Switch `canUpdateTitle` to true to see it in action.

It will need updating once the back end is ready but can be merged safely before then. 

Screenshot (22)

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/methodology/components/MethodologySummaryForm.tsx
+++ b/src/explore-education-statistics-admin/src/pages/methodology/components/MethodologySummaryForm.tsx
@@ -3,6 +3,7 @@ import ButtonGroup from '@common/components/ButtonGroup';
 import ButtonText from '@common/components/ButtonText';
 import Form from '@common/components/form/Form';
 import FormFieldTextInput from '@common/components/form/FormFieldTextInput';
+import FormFieldRadioGroup from '@common/components/form/FormFieldRadioGroup';
 import useFormSubmit from '@common/hooks/useFormSubmit';
 import { mapFieldErrors } from '@common/validation/serverValidations';
 import Yup from '@common/validation/yup';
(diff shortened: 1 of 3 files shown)


src/explore-education-statistics-admin/src/pages/methodology/components/MethodologySummaryForm.tsx
```diff
@@ -11,6 +12,7 @@ import React from 'react';
 
 export interface MethodologySummaryFormValues {
   title: string;
+  titleType?: 'publication' | 'alternative';
```
benoutram, line 15: I'm just wondering if this is needed because every field on the form needs to be backed by a form value? (Sorry, lack of knowledge on the FE here 🙂 ). Could it be switched on a field like `alternativeTitle` having a value present or not or is that a bad idea?

On the BE I imagine we wouldn't have a field for `titleType` but store an optional field like `alternativeTitle` which can be set when updating a Methodology and if that's got a value then the title is overridden with it rather than the publication title, when returning a Methodology view model response.

Do you think that idea fits with this change?

amyb-hiveit: I've had another look at this and don't think there's anything else to be done currently as need to maintain the existing version for the form. 

`titleType` is needed for the form radio buttons, but wouldn't be submitted to the api. On MethodologySummaryEditPage the initial value for it will need to be changed once the BE is in place, so something like `titleType: methodology.alternativeTitle ? 'alternative' : 'publication', ` and on the form we'd replace `title` with `alternativeTitle`.
The BasicMethodology type will also need to be updated then.

benoutram (approved):

PR #2636, opened 2021-06-29
EES-2427 Make comparison of location names case sensitive when importing

markhiveit:
This ensures all comparisons of Location *_Name's are case-sensitive compared. This also includes RscRegion_Code as the code for RscRegions doubles as the name.

Another thing to note about this PR is that the additional unit test currently passes before this change was made. This is because the in-memory DB in the version of EF we're using doesn't support case-insensitive comparisons: https://github.com/dotnet/efcore/issues/20770

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Processor.Tests/Services/ImporterLocationServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Processor.Tests/Services/ImporterLocationServiceTests.cs
@@ -12,30 +12,83 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Processor.Tests.Servic
     public class ImporterLocationServiceTests
     {
         [Fact]
-        public async Task Find_EnglishDevolvedAreaCaseSensitiveLocationName()
+        public async Task Find_PreExistingLocation()
         {
-            var country = new Country("1", "england");
-            var lowerCase = new EnglishDevolvedArea("2", "casesensitive");
-            var upperCase = new EnglishDevolvedArea("1", "CASESENSITIVE");
+            var location = new Location
+            {
+                Country = new Country("1", "England"),
+                LocalAuthority = new LocalAuthority("2", "3", "Bedford")
+            };
+
+            var service = BuildImporterLocationService();
+            
             var statisticsDbContextId = Guid.NewGuid().ToString();
             await using (var statisticsDbContext = InMemoryStatisticsDbContext(statisticsDbContextId))
             {
-                var service = BuildImporterLocationService();
+                statisticsDbContext.Add(location);
+                await statisticsDbContext.SaveChangesAsync();
+            }
+
+            await using (var statisticsDbContext = InMemoryStatisticsDbContext(statisticsDbContextId))
+            {
+                var result1 = service.Find(statisticsDbContext, location.Country,
+                    localAuthority: location.LocalAuthority);
+                Assert.NotNull(result1);
+                Assert.Equal("Bedford", result1.LocalAuthority_Name);
+                await statisticsDbContext.SaveChangesAsync();
+            }
+
+            await using (var statisticsDbContext = InMemoryStatisticsDbContext(statisticsDbContextId))
+            {
+                var statDbLocations = statisticsDbContext.Location.ToList();
+                Assert.Single(statDbLocations);
+                Assert.Equal("Bedford", statDbLocations[0].LocalAuthority_Name);
+            }
+        }
+
+        [Fact]
+        public async Task Find_CaseSensitiveLocationName()
+        {
+            var country = new Country("1", "england");
+            var lowerCase = new EnglishDevolvedArea("2", "casesensitive");
+            var upperCase = new EnglishDevolvedArea("2", "CASESENSITIVE");
+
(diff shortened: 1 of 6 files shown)


src/GovUk.Education.ExploreEducationStatistics.Data.Processor.Tests/Services/ImporterLocationServiceTests.cs
```diff
@@ -0,0 +1,46 @@
...
+using static GovUk.Education.ExploreEducationStatistics.Data.Model.Database.StatisticsDbUtils;
+namespace GovUk.Education.ExploreEducationStatistics.Data.Processor.Tests.Services
+{
+    public class ImporterLocationServiceTests
+    {
+        [Fact]
+        public async Task Find_EnglishDevolvedAreaCaseSensitiveLocationName()
+        {
+            var country = new Country("1", "england");
+            var lowerCase = new EnglishDevolvedArea("2", "casesensitive");
```
benoutram, line 18: Both of these need to have the same code otherwise you aren't testing case sensitivity at all since they are two different locations by their code and name combination regardless of case. At the moment, one has code 1, the other code 2.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor.Tests/Services/ImporterLocationServiceTests.cs
```diff
@@ -0,0 +1,46 @@
...
+                Assert.NotNull(result2);
+                Assert.Equal("CASESENSITIVE", result2.EnglishDevolvedArea_Name);
+                await statisticsDbContext.SaveChangesAsync();
+                var statDbLocations = statisticsDbContext.Location.ToList();
+                Assert.Equal(2, statDbLocations.Count);
+                Assert.Equal("casesensitive", statDbLocations[0].EnglishDevolvedArea_Name);
+                Assert.Equal("CASESENSITIVE", statDbLocations[1].EnglishDevolvedArea_Name);
+            }
+        }
+        private static ImporterLocationService BuildImporterLocationService()
```
benoutram, line 39: What's happened to all the spacing in this file? 🙂

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterLocationService.cs
```diff
@@ -268,8 +268,36 @@ public class ImporterLocationService : BaseImporterService
...
                                  && location.PlanningArea_Name == (planningArea != null ? planningArea.Name : null));
-            
-            return context.Location.AsNoTracking().FirstOrDefault(predicateBuilder);
+
+            // Can return multiple locations because SQL comparison is case insensitive
+            var locations = context.Location.AsNoTracking()
+                .Where(predicateBuilder)
+                .ToList();
+
+            foreach(var location in locations)
```
benoutram, line 277: You can use LINQ syntax here, replacing the whole `foreach(...) { if(...) { return location } return null }`
with `return locations.FirstOrDefault(...);`

src/GovUk.Education.ExploreEducationStatistics.Data.Processor.Tests/Services/ImporterLocationServiceTests.cs
```diff
@@ -0,0 +1,46 @@
...
+using GovUk.Education.ExploreEducationStatistics.Data.Processor.Services;
+using Moq;
+using Xunit;
+using static GovUk.Education.ExploreEducationStatistics.Data.Model.Database.StatisticsDbUtils;
+namespace GovUk.Education.ExploreEducationStatistics.Data.Processor.Tests.Services
+{
+    public class ImporterLocationServiceTests
+    {
+        [Fact]
+        public async Task Find_EnglishDevolvedAreaCaseSensitiveLocationName()
```
benoutram, line 15: Not sure if it's necessary but just pointing out this this test doesn't test this method with any pre-existing location data, only data which is added and saved through it not already existing. Might be a good idea to add an alternative to this with pre-existing data which should be found and returned without needing to be saved.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor.Tests/Services/ImporterLocationServiceTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            var country = new Country("1", "england");
+            var lowerCase = new EnglishDevolvedArea("2", "casesensitive");
+            var upperCase = new EnglishDevolvedArea("1", "CASESENSITIVE");
+            var statisticsDbContextId = Guid.NewGuid().ToString();
+            await using (var statisticsDbContext = InMemoryStatisticsDbContext(statisticsDbContextId))
+            {
+                var service = BuildImporterLocationService();
+                var result1 = service.Find(statisticsDbContext, country,
```
benoutram, line 24: I think it would be worth sticking in a link to https://github.com/dotnet/efcore/issues/6153 in a comment to say this test is not actually testing what it should be due to InMemory database being case sensitive rather than case insensitive like SqlServer.

benoutram (changes requested):

benoutram (approved):

PR #2639, opened 2021-06-30
EES-2401 Also update Statistics DB publication when updating Content DB publication

markhiveit:


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -148,12 +148,29 @@ public class PublicationService : IPublicationService
 
                     return (await ValidatePublicationSlugUniqueForUpdate(publication.Id, updatedPublication.Slug)).Map(_ =>
                     {
+                        var statisticsDbPublication = _statisticsDbContext.Publication.Find(publication.Id);
```
benoutram, line 151: We do all the saves in the block below, so below you could just `statisticsDbPublication.Slug = publication.Slug;` before that save, rather than fetching and saving twice, and remove this bit of code.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -148,12 +148,29 @@ public class PublicationService : IPublicationService
...
                 })
                 .OnSuccess(async publication =>
-                {
+                { 
+                    var statisticsDbPublication = await _statisticsDbContext.Publication.FindAsync(publication.Id);
+                    if (statisticsDbPublication != null)
+                    {
+                        _statisticsDbContext.Update(statisticsDbPublication);
+                        statisticsDbPublication.Title = updatedPublication.Title;
+                        statisticsDbPublication.TopicId = updatedPublication.TopicId;
```
benoutram, line 170: Is this topic guaranteed to exist at the point? What happens if no subjects got imported for it yet?

markhiveit:
See comment on EES-2401 ticket.

PR #2643, opened 2021-07-01
EES-2374 data and files accordion changes

amyb-hiveit:
Changes to the data and files accordion for SOW5.

These changes are on the public frontend and in the admin (release content and pre-release access), with some differences:
- download all files link and create table button in frontend only
- link to view pre-release access list in admin only

Screenshot (27)


src/explore-education-statistics-common/src/modules/release/components/__tests__/ReleaseDataAndFiles.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+      },
+      {
+        id: 'f2',
+        name: 'data file',
+        type: 'Data',
+      },
+      {
+        id: 'f3',
+        name: 'ancillary file',
+        type: 'Ancillary',
```
ntsim, line 24: Could use more unique names i.e. 'Test data file' and 'Test ancillary file'. The risk of using less unique names is that text assertions could pick these pieces of text by accident if, for example, the guide text is changed in the future.

src/explore-education-statistics-common/src/modules/release/components/__tests__/ReleaseDataAndFiles.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+    },
+    slug: 'r-slug',
+  };
+
+  test('renders the data and files accordion', () => {
+    render(
+      <ReleaseDataAndFiles
+        release={testRelease as Release}
+        renderDownloadLink={file => <a href="/">{file.name}</a>}
+        renderMetaGuidanceLink={<span>mock meta guidance link</span>}
```
ntsim, line 40: Let's use actual anchor elements for both of these to be consistent and make the test assertion stricter (see below).

src/explore-education-statistics-common/src/modules/release/components/__tests__/ReleaseDataAndFiles.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+    expect(screen.getByText('Explore data and files')).toBeInTheDocument();
+
+    expect(screen.getByText('All files')).toBeInTheDocument();
+    expect(screen.getByText('data file')).toBeInTheDocument();
+
+    expect(screen.getByText('Open data')).toBeInTheDocument();
+    expect(screen.getByText('mock meta guidance link')).toBeInTheDocument();
+
+    expect(screen.getByText('Other files')).toBeInTheDocument();
+    expect(screen.getByText('ancillary file')).toBeInTheDocument();
```
ntsim, line 53: For assertion on links, we could use:

```ts
expect(screen.getByRole('link', { name: 'The link text' })).toBeInTheDocument();
```

This is preferable as it's a little stricter and tells us a little bit more about what we're asserting on other than text :slightly_smiling_face:

src/explore-education-statistics-common/src/modules/release/components/__tests__/ReleaseDataAndFiles.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+        renderPreReleaseAccessLink={<span>mock pre-release access link</span>}
+        renderDownloadLink={file => <a href="/">{file.name}</a>}
+        renderMetaGuidanceLink={<span>mock meta guidance link</span>}
+      />,
+    );
+
+    expect(screen.getByText('Pre-release access list')).toBeInTheDocument();
+    expect(
+      screen.getByText('mock pre-release access link'),
+    ).toBeInTheDocument();
```
ntsim, line 83: Similar to above, could use a link assertion

src/explore-education-statistics-common/src/modules/release/components/ReleaseDataAndFiles.tsx
```diff
@@ -0,0 +1,117 @@
...
+  onSectionOpenEvent?: (accordionSection: {
+    id: string;
+    title: string;
+  }) => void;
+  release: Release;
+  renderDownloadLink: (file: FileInfo) => ReactNode;
+  renderMetaGuidanceLink: ReactNode;
+  renderPreReleaseAccessLink?: ReactNode;
+}
+const ReleaseDataAndFiles = ({
```
ntsim, line 19: Might be better to use more explicit naming like `ReleaseDataAndFilesAccordion` we could reasonably expect there to be other components around data and files.

src/explore-education-statistics-common/src/modules/release/components/ReleaseDataAndFiles.tsx
```diff
@@ -0,0 +1,117 @@
+import Accordion from '@common/components/Accordion';
+import AccordionSection from '@common/components/AccordionSection';
+import Details from '@common/components/Details';
+import { Release } from '@common/services/publicationService';
+import { FileInfo } from '@common/services/types/file';
+import React, { ReactNode } from 'react';
+
+interface Props {
+  renderCreateTablesButton?: ReactNode;
+  onSectionOpenEvent?: (accordionSection: {
```
ntsim, line 10: Could make this more succinct with: `onSectionOpen`. It's semantically an event handler based on its `on*` prefix so the 'event' part is redundant

src/explore-education-statistics-frontend/src/modules/find-statistics/PublicationReleasePage.tsx
```diff
@@ -328,76 +330,65 @@ const PublicationReleasePage: NextPage<Props> = ({ release }) => {
...
+            });
+          }}
+          release={release}
+          renderCreateTablesButton={
+            <CreateTablesButton
+              className="govuk-!-width-full"
+              release={release}
+            />
+          }
+          renderDownloadLink={(file: FileInfo) => {
```
ntsim, line 348: Could remove param type here as it's declared in the prop type

ntsim (changes requested):

src/explore-education-statistics-common/src/modules/release/components/ReleaseDataAndFiles.tsx
```diff
@@ -0,0 +1,117 @@
...
+const ReleaseDataAndFiles = ({
+  onSectionOpenEvent,
+  release,
+  renderCreateTablesButton,
+  renderDownloadLink,
+  renderMetaGuidanceLink,
+  renderPreReleaseAccessLink,
+}: Props) => {
+  const dataFiles = release.downloadFiles.filter(
+    file => file.name === 'All files' || file.type !== 'Ancillary',
```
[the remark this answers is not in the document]
amyb-hiveit: Showing these here is temporary, they will be on the data catalogue page once that exists. Currently all the download files are shown, this PR separates out the All and Ancillary files, so this is the rest.

src/explore-education-statistics-common/src/modules/release/components/ReleaseDataAndFiles.tsx
```diff
@@ -0,0 +1,117 @@
...
+  );
+
+  const ancillaryFiles = release.downloadFiles.filter(
+    file => file.type === 'Ancillary' && file.name !== 'All files',
+  );
+
+  return (
+    <div className="dfe-download-section">
```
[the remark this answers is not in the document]
amyb-hiveit: Strange how? Any suggestions for improving it?

src/explore-education-statistics-common/src/modules/release/components/ReleaseDataAndFiles.tsx
```diff
@@ -0,0 +1,117 @@
...
+interface Props {
+  renderCreateTablesButton?: ReactNode;
+  onSectionOpenEvent?: (accordionSection: {
+    id: string;
+    title: string;
+  }) => void;
+  release: Release;
+  renderDownloadLink: (file: FileInfo) => ReactNode;
+  renderMetaGuidanceLink: ReactNode;
+  renderPreReleaseAccessLink?: ReactNode;
```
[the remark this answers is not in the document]
amyb-hiveit: I've changed it here, but if there's a rule around props order I think it should be covered by the linter and handled by something like https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-sort-props.md rather than be a manual thing.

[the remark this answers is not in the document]
ntsim: Great idea :smile: we should put that in with another PR

src/explore-education-statistics-common/src/modules/release/components/ReleaseDataAndFilesAccordion.module.scss
```diff
@@ -0,0 +1,26 @@
+@import '~govuk-frontend/govuk/base';
+
+.container {
+  margin-bottom: govuk-spacing(9);
+  margin-top: -1 * (govuk-spacing(6));
```
ntsim, line 5: Think we simplify this to:

```css
margin-top: -(govuk-spacing(6));
```

src/explore-education-statistics-common/src/modules/release/components/ReleaseDataAndFilesAccordion.module.scss
```diff
@@ -0,0 +1,26 @@
...
+    :global(.govuk-accordion__section-header) {
+      border-top: 0;
+    }
+  }
+
+  :global(.dfe-flex) {
+    @include govuk-media-query($until: tablet) {
+      display: block;
+    }
+  }
```
ntsim, line 21: Rather than overriding `dfe-flex` like this, would suggest creating a responsive variant of it in `_flex.scss` and using that instead e.g.

```scss
.dfe-flex--tablet {
   @include govuk-media-query($until: tablet) {
      display: flex;
   }
}
```

src/explore-education-statistics-common/src/modules/release/components/ReleaseDataAndFilesAccordion.tsx
```diff
@@ -0,0 +1,115 @@
...
+            <div className="dfe-flex dfe-justify-content--space-between dfe-align-items--center">
+              <div>
+                <h3>Create your own tables</h3>
+                <p>
+                  Explore our range of data and build your own tables from it.
+                </p>
+              </div>
+              <p className="govuk-!-width-one-quarter dfe-flex-shrink--0">
+                {renderCreateTablesButton}
+              </p>
```
ntsim, line 73: A few little things here:

- This is probably better as a `div` than a `p`
- Don't think we need the `dfe-flex-shrink--0`

src/explore-education-statistics-common/src/modules/release/components/ReleaseDataAndFilesAccordion.module.scss
```diff
@@ -0,0 +1,26 @@
...
+
+  :global(.govuk-button) {
+    width: 100%;
+  }
```
ntsim, line 25: Don't need this as we pass the 'Create tables' button through as a render prop and this button already has `govuk-!-width-full` set

src/explore-education-statistics-common/src/modules/release/components/__tests__/ReleaseDataAndFilesAccordion.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+        release={testRelease as Release}
+        renderDownloadLink={file => <a href="/">{file.name}</a>}
+        renderMetaGuidanceLink={<a href="#">mock meta guidance link</a>}
+      />,
+    );
+
+    expect(screen.getByText('Explore data and files')).toBeInTheDocument();
+
+    expect(screen.getByText('All files')).toBeInTheDocument();
+    expect(screen.getByText('Test data file')).toBeInTheDocument();
```
ntsim, line 47: Could assert that this is a link?

src/explore-education-statistics-common/src/modules/release/components/__tests__/ReleaseDataAndFilesAccordion.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+    expect(screen.getByText('Open data')).toBeInTheDocument();
+    expect(
+      screen.getByRole('link', { name: 'mock meta guidance link' }),
+    ).toBeInTheDocument();
+
+    expect(screen.getByText('Other files')).toBeInTheDocument();
+    expect(screen.getByText('Test ancillary file')).toBeInTheDocument();
```
ntsim, line 55: Could assert that this is a link?

ntsim (changes requested):

src/explore-education-statistics-common/src/modules/release/components/ReleaseDataAndFiles.tsx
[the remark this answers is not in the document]
ntsim: Added some new suggestions to the PR!

src/explore-education-statistics-common/src/modules/release/components/ReleaseDataAndFiles.tsx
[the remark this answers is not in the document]
ntsim: Ah right I see! Maybe just rename the variable to `otherFiles` for now anyway?

src/explore-education-statistics-common/src/modules/release/components/ReleaseDataAndFilesAccordion.module.scss
amyb-hiveit: I've done it differently in the scss module instead. Code style wise I don't really understand how the mix of utility and module styles is meant to work, is it better to add a new utility class instead?

ntsim: Think it depends! I would say that the utility styles are mostly there for really simple cases where you only need a few styles and it would just be unnecessary to create a new module class e.g. centring stuff with flex.

The main problem with loads of little module classes doing the same thing is that they gradually bloat the final build with more and more CSS, hence it's a little more efficient to have the utils in these cases.

I wouldn't say there's a very clear cut-off point between deciding to put it in a CSS module or using utils, but I'd suggest that if you're sticking 3/4 or more utilities on something, refactor it into a module class. Even more so if it needs some media queries. We especially want to avoid applying media queries on top of util classes, like the previous code!

I think doing this as a module class is fine though, I was originally going to suggest this! :slightly_smiling_face: There is reasonably a use-case for responsive variants of utils as well, but I guess we might want to think about that a little more before jumping into it wholesale.

ntsim (approved):

PR #2645, opened 2021-07-01
Ees 2221 permissions issue with create methodology

duncan-at-hiveit:
This PR:
- fixes EES-2478 - a bug whereby analysts could still create methodologies on their publications.  This was due to a retained Claim against the Analyst role
- splits a large roles and permissions Robot test suite into a smaller end-to-end test suite that tests the happy path with various roles, and smaller test suites that test each individual publication and release role in turn.  These are quite basic at the moment but the groundwork's there to start adding more test cases.  It also tests that the analysts can no longer see the Create Methodology controls!
- adds a new bootstrapping test suite in for generating the seed data necessary for testing releases in various approval states with individual combinations of release roles.  This is excluded from running by default, but can be run against a target environment to generate the data needed to support the permissions ui tests.

Along with this PR, I will be supplying a new seed data zip file which contains the new data that was generated by the bootstrapping test.  Robot tests run against this data will succeed!

Tests Log

Dashboard - Expl


benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/CreateMethodologyForSpecificPublicationAuthorizationHandlers.cs
```diff
@@ -4,7 +4,6 @@
 using GovUk.Education.ExploreEducationStatistics.Common.Services.Security;
 using GovUk.Education.ExploreEducationStatistics.Content.Model;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
-using IdentityServer4.Extensions;
```
benoutram, line 7: Just for reference I have a change on my current branch to remove all other usages of `NullOrEmpty` in this package and I replace them with our own version of it.

duncan-at-hiveit: OK!

tests/robot-tests/tests/bootstrap_data/release_and_publication_role_permissions_data.robot
```diff
@@ -0,0 +1,64 @@
...
+Force Tags  BootstrapData  Local  Dev
+
+Suite Setup       user signs in as bau1
+Suite Teardown    user closes the browser
+
+*** Test Cases ***
+Create test theme and topic
+    Import bootstrap data roles and permissions variables
+    ${THEME_ID}=  user creates theme via api  ${THEME_NAME}
+    ${TOPIC_ID}=  user creates topic via api  ${TOPIC_NAME}  ${THEME_ID}
```
markhiveit, line 24: Is it necessary to create another theme and topic here instead of use the existing `%{TEST_THEME_NAME}` and `%{TEST_TOPIC_NAME}`?

I think these new themes and topics will quickly accumulate on Dev. The `run_tests.py` script assumes that all the stuff created by the tests is inside the test topic it creates, and it deletes that after the run is complete. And it keeps everything contained within "Test theme", so even if stuff doesn't get cleaned up it all stays contained in there.

The only other test suite that uses the `user creates theme/topic via api` keywords cleans up after itself using `user deletes theme via api` in the suite teardown.

tests/robot-tests/tests/libs/admin-common.robot
```diff
@@ -494,34 +539,49 @@ user changes methodology status to Approved
...
-    user waits until page does not contain loading spinner
\ No newline at end of file
+    user waits until page does not contain loading spinner
+
+user removes publication owner access from analyst 
+    [Arguments]  ${PUBLICATION_NAME}  ${ANALYST_EMAIL}=ees-analyst1@education.gov.uk
+    user goes to manage user  ${ANALYST_EMAIL}
+    user scrolls to element  css:[name="selectedPublicationId"]
+    # NOTE: The below wait is to prevent a transient failure that occurs on the UI test pipeline due to the DOM not being fully rendered which
+    # causes issues with getting the 'selectedPublicationId' selector (staleElementException)
```
markhiveit, line 565: I feel and know your pain.

markhiveit (changes requested):

tests/robot-tests/tests/admin/analyst/role_ui_permissions/release_and_publication_role_end_to_end_permissions.robot
```diff
@@ -0,0 +1,176 @@
...
+    user waits until page does not contain loading spinner
+    user enters text into element  id:metaGuidanceForm-content  Test meta guidance content
+    user waits until page contains accordion section  ${SUBJECT_NAME}
+    user enters text into meta guidance data file content editor  ${SUBJECT_NAME}
+    ...  meta guidance content
+    user clicks button  Save guidance
+
+Navigate to 'Footnotes' page
+    [Tags]  HappyPath
+    user waits for page to finish loading
```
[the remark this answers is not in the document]
duncan-at-hiveit: Yeah I basically lifted and shifted this from the original end-to-end test.  Pretty sure that the `user clicks link Footnotes` line directly below it does exactly what you're suggesting.  Where I've seen these micro-sleeps before, it seems as if they've been used to get over something a bit buggy or intermittent with Selenium and StaleElementExceptions.  Happy to try removing it, although I think that possibly it's worth doing this as part of a wider clean-up task?

tests/robot-tests/tests/bootstrap_data/release_and_publication_role_permissions_data.robot
duncan-at-hiveit: It is necessary in this particular case, as this will be a Theme and a Topic that will pre-exist (in seed data and on Dev) and will be readonly.  This is a manually-run (excluded by default) test that actually sets up the data necessary for the UI permissions tests.  The reason being that this is a very time-consuming process that would kill the Robot tests if needing to do it with every test run.

So on Drive, you'll see I've uploaded a new "ees-mssql-data-37.zip" file which contains our original seed data plus the Theme and Topic created here, along with 16 Releases in different combinations of user roles and approval statuses (as seen the the screenshot attached to the PR).

Once SOW4 is merged into the dev branch and the SOW4 code deployed to the Dev environment, we will run this test manually (pointed towards Dev) and this will create these Themes, Topics, Publications and Releases on Dev.  Then we won't have to run it again (but if we ever do need to, or if we need to prep any other environment for testing against, we can run this manually and it'll generate what we need).

duncan-at-hiveit: (and it's time consuming mostly due to the file uploads and the approval process needed for creating the Releases)

tests/robot-tests/tests/admin/analyst/role_ui_permissions/publication_owner_permissions.robot
```diff
@@ -0,0 +1,38 @@
+*** Settings ***
+Resource    ../../../libs/admin-common.robot
+Resource    ../../../libs/common.robot
+Resource    ../../../bootstrap_data/bootstrap_data_constants.robot
+Resource    role_ui_permissions_common.robot
```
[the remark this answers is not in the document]
duncan-at-hiveit: Yep, I can see advantages and disadvantages for any way, so happy to go with your preference.  Would that be mirrored structure in libs?  Or flat structure in libs?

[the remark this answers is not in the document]
markhiveit: I'd prefer to put them in libs

tests/robot-tests/tests/admin/analyst/role_ui_permissions/release_and_publication_role_end_to_end_permissions.robot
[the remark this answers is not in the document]
markhiveit: Yeah - I think wider clean-up task will do it. I created that keyword not intending to be used much and I have noticed it is all over the ui test! I've learned my lesson on that one - should have just put the sleep and comment in place in the test suite itself, like you have in this PR.

tests/robot-tests/tests/bootstrap_data/release_and_publication_role_permissions_data.robot
markhiveit: Right - you did say you were going to do this and I totally forgot 🤦

tests/robot-tests/tests/libs/admin-common.robot
duncan-at-hiveit: Haha - can't take credit for this one - was in a pre-existing set of steps that I moved here to reuse :)

tests/robot-tests/tests/admin_and_public/bau/publish_release_and_amend.robot
```diff
@@ -172,7 +172,6 @@ Add public prerelease access list
 
 Approve release
     [Tags]  HappyPath
-    user clicks link   Sign off
```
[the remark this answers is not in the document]
duncan-at-hiveit: I agree (and also I think it's hard to know what to do for the best).

We could do it with a naming convention but then you might get an explosion of keywords.

We could do it with conditional keywords that checks your current URL and decides whether or not to run the navigation step.

We could pass in named arguments like:

```
Approve release  ${navigate_to_publication}=${PUBLICATION}  ${navigate_to_release_summary}=${RELEASE}
```

My personal preference is that I ask a common step to perform an action and it work out best how to get there, thus taking a load of navigation boilerplate out of the test cases, but then you don't want the common steps being too clever either!

Maybe the common steps should have a way to work out if you're on the right page already and if not, have some basic way to navigate to it...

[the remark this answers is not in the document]
markhiveit: Discussed this on Slack - Duncan had the idea to create a new keyword that determines whether the user is on the right page, and if not does the navigation required. I'll create a ticket for this.

markhiveit (approved):

tests/robot-tests/tests/admin/analyst/role_ui_permissions/publication_owner_permissions.robot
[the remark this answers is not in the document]
duncan-at-hiveit: Done!

PR #2650, opened 2021-07-05
EES-2432 - misc test utils

luke-h1:
- [x] Get input from user (for upload subject to existing release) 
- [x] type errorHandler.ts correctly
- [x] Convert everything to use Axios 
- [x] Decide on whether to use commander lib or not 
- [x] fix upload subject & publish release 
- [x] Manual test 
- [x] cleanup debugging stuff
- [x] Add readme with instructions on how to get started

# This PR: 
* Creates a CLI tool that automates a few manual tasks that we do manually at the moment when trying to reproduce bugs
 
At the moment the following has been automated: 
* create a new release
* publish new release (with data files) 
* upload a subject to an existing release

## Notable changes: 
* Moves the importer-publisher test suite into this CLI utility and gets rid of the bash script associated with that. 
* replaces instances of `__dirname` to `process.cwd()`. This is because the various functions that upload data to the service are now looking at the root directory of this tool to grab test data. 
* Adds inquirer.js lib which is responsible for asking the user questions and getting input from the user  
* Adds `errorHandler.ts`. This function is responsible for catching various Axios related errors and letting the user know about them.  
* Improves general error handling


useful-scripts/misc-test-utils/src/modules/publication/publish/index.ts
```diff
@@ -7,21 +7,24 @@ import StreamZip from 'node-stream-zip';
...
+import { ReleaseData } from '../../../interfaces/ReleaseData';
+import errorHandler from '../../../utils/errorHandler';
+import chalk from 'chalk';
 
 // disable insecure warnings
 process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
-
-const { JWT_TOKEN, API_URL, TOPIC_ID } = process.env;
+const { ADMIN_URL, TOPIC_ID, JWT_TOKEN } = process.env;
+const cwd = process.cwd();
```
ntsim, line 20: I guess this is intended but is it a good idea to dump stuff into the current directory?

The main problem is that can potentially dump the files outside of the sub-project directory if you run the script outside of it. This is probably something we want to avoid? 

I'm guessing ideally we want the test results to go into `importer-publisher-tests/test-results`? If so, I would suggest setting up some constants file (perhaps in the sub-project root), which sets up variables for things like the intended test result output directory. We can then import this into the required files for use.

useful-scripts/misc-test-utils/src/modules/subject/upload/index.ts
```diff
@@ -0,0 +1,226 @@
...
+
+  // potential issue here with how windows paths are formatted in git bash
+  if (glob !== undefined) {
+    const oldPath =
+      process.cwd() +
+      '/zip-files/' +
+      fs.readdirSync(process.cwd() + '/zip-files')[0];
+
+    const newPath =
+      process.cwd() + '/' + fs.readdirSync(process.cwd() + '/zip-files')[0];
```
ntsim, line 68: To avoid path separator issues - use the `path` module :wink:. You can do stuff like:

```ts
const path = path.resolve(__dirname, 'somewhere', 'else') // absolute-path-to-current-dir/somewhere/else
```

Check out [the docs](https://nodejs.org/api/path.html#path_path_resolve_paths)

useful-scripts/misc-test-utils/src/utils/errorHandler.ts
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
+/* eslint-disable no-constant-condition */
```
ntsim, line 2: Don't think we need these

useful-scripts/misc-test-utils/src/utils/errorHandler.ts
```diff
@@ -0,0 +1,39 @@
...
+/* eslint-disable no-console */
+
+import { AxiosError } from 'axios';
+import chalk from 'chalk';
+
+const errorHandler = async (e: AxiosError) => {
+  switch (e!.response!.status) {
```
ntsim, line 9: Eh? Not sure why we're using non-null assertions here. We should be able to use optional chaining instead:

```ts
e.response?.status
```

useful-scripts/misc-test-utils/src/modules/subject/upload/index.ts
```diff
@@ -0,0 +1,226 @@
...
+        content: '<p>testing</p>',
+        subjects: subjArr,
+      },
+    });
+  } catch (e) {
+    errorHandler(e);
+  }
+};
+
+const UploadSingleSubject = async (releaseId: string) => {
```
ntsim, line 155: Functions should be camelCased i.e. `uploadSingleSubject`. The main exception are React components, which are PascalCased instead.

ntsim (changes requested):

useful-scripts/misc-test-utils/src/modules/release/create/index.ts
```diff
@@ -0,0 +1,75 @@
+/* eslint-disable */
```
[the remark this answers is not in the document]
luke-h1: I think I left this in accidentally when creating this initially. I've removed this line and sorted out the majority of linting errors.

markhiveit (changes requested):

useful-scripts/misc-test-utils/src/modules/publication/publish/index.ts
luke-h1: I don't think the test results directory is being used. I'll remove it as part of this PR. The whole cli will break if you attempt to start it outside the root of the `misc-test-utils` directory since it looks for a zipped data file in whatever the current working directory is (which we assume is `misc-test-utils`).

useful-scripts/misc-test-utils/src/modules/release/create/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  console.time('createRelease');
+  try {
+    const res = await axios({
+      method: 'POST',
+      url: `${ADMIN_URL}/api/publications/${publicationId}/releases`,
+      data: {
+        timePeriodCoverage: { value: 'AY' },
+        releaseName: 2222,
+        typeId: '1821abb8-68b0-431b-9770-0bea65d02ff0',
+        publicationId: 'c76c1bc4-0a01-4116-74cd-08d8eebbf0c1',
```
ntsim, line 53: What's this `publicationId` doing? There's already one in the URL?

ntsim (approved):

useful-scripts/misc-test-utils/src/modules/release/create/index.ts
luke-h1: I'm not too sure why I left this in. I think the `typeId` GUID being sent here is responsible for specifying what type of release to create (Ad hoc, National Statistics, etc.) so I don't think the `publicationId` is necessary and can be removed.

markhiveit (approved):

PR #2653, opened 2021-07-06
EES-2494 set editingMode default to preview to fix prerelease page

amyb-hiveit:


src/explore-education-statistics-admin/src/contexts/EditingContext.tsx
```diff
@@ -22,7 +22,7 @@ export interface EditingContextState {
 }
 
 export const EditingContext = createContext<EditingContextState>({
-  editingMode: 'edit',
+  editingMode: 'preview',
```
ntsim, line 25: We need to set this to `edit` for the release and methodology content pages

ntsim (approved):

PR #2656, opened 2021-07-06
EES-2496 Return latest Methodologies when getting PublicationViewModel.

benoutram:
This PR alters the Admin PublicationService to return the latest versions of Methodologies related to a Publication when building a PublicationViewModel.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
@@ -253,7 +253,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
             return Unit.Instance;
         }
 
-        public static IQueryable<Publication> HydratePublicationForPublicationViewModel(IQueryable<Publication> values)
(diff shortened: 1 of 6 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -184,7 +190,15 @@ public class PublicationService : IPublicationService
             return await _persistenceHelper
                 .CheckEntityExists<Publication>(publicationId, HydratePublicationForPublicationViewModel)
                 .OnSuccess(_userService.CheckCanViewPublication)
-                .OnSuccess(publication => _mapper.Map<PublicationViewModel>(publication));
+                .OnSuccess(async publication =>
+                {
+                    var publicationViewModel = _mapper.Map<PublicationViewModel>(publication);
+                    var methodologies = await _methodologyRepository.GetLatestByPublication(publicationId);
```
benoutram, line 196: I wasn't sure whether to do it like this where we defer to the repository to only fetch the latest methodologies for the publication, or keep the hydrate method like it was which loaded all methodologies/versions into the publications, and use automapper to only map the latest version like we do for Publication->MyPublicationViewModel.

duncan-at-hiveit (approved):

PR #2658, opened 2021-07-07
EES-2461 set y axis width for line and vertical bar charts

amyb-hiveit:
Prevents the axis values being cut off by determining the width based on the highest value, plus taking into account decimal places and units. The y-axis size field is now blank by default, if you  add a size it will override the calculated one.


src/explore-education-statistics-common/src/modules/charts/util/__tests__/getMinorAxisSize.test.ts
```diff
@@ -0,0 +1,71 @@
+import getMinorAxisSize from '@common/modules/charts/util/getMinorAxisSize';
+
+describe('getMinorAxisSize', () => {
+  const testDataSetCategories = [
```
ntsim, line 4: Would add explicit type as untyped variables can easily get out of sync with the reality - particularly in tests.

src/explore-education-statistics-common/src/modules/charts/util/getMinorAxisSize.ts
```diff
@@ -0,0 +1,38 @@
+import { DataSetCategory } from '@common/modules/charts/types/dataSet';
+import parseNumber from '@common/utils/number/parseNumber';
+
+interface Props {
+  dataSetCategories: DataSetCategory[];
+  minorAxisSize: number | undefined;
+  minorAxisDecimals: number | undefined;
```
ntsim, line 7: Rather than undefined, can mark these as optional:

```ts
minorAxisSize?: number;
minorAxisDecimals?: number
```
This way we don't have to explicitly set these as undefined in code e.g. 

```ts
const size = getMinorAxisSize({
  dataSetCategories: testDataSetCategories,
  minorAxisUnit: '',
});
```

instead of:

```ts
const size = getMinorAxisSize({
  dataSetCategories: testDataSetCategories,
  minorAxisSize: undefined,
  minorAxisDecimals: undefined,
  minorAxisUnit: '',
});
```

src/explore-education-statistics-common/src/modules/charts/components/LineChartBlock.tsx
```diff
@@ -104,17 +105,23 @@ const LineChartBlock = ({
...
     dataSetCategories,
     legend.items,
     meta,
   );
-
   const minorAxisDecimals = getMinorAxisDecimalPlaces(dataSetCategoryConfigs);
+  const minorAxisUnit = getUnitOrDefault(
+    axes.minor.unit,
+    dataSetCategoryConfigs,
+  );
```
ntsim, line 117: This function name is a little confusing. The name implies that it should try and find unit, and if it cannot, it should use a default. Looking at the code, this isn't the case though, as it just uses the first param (the 'default') if it's been set.

There's also something about the parameter order that doesn't sit right as it's allow undefined for the first value (typically optional/undefined properties should be set last in a function signature as these can be afforded default parameters).

I would suggest to improve the semantics of this, we refactor this to the following:

```ts
const minorAxisUnit = axes.minor.unit ?? getUnit(dataSetCategoryConfigs);
```

Where:

```ts
function getUnit(dataSetCategoryConfigs: DataSetCategoryConfig[]): string 
```

Note that I've included an explicit return type - this is good practice.

src/explore-education-statistics-common/src/modules/charts/util/getMinorAxisSize.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Get the highest value used on the axis and determine the axis width based on its length.
+ * 12 is a fairly arbitrary width here, more than needed for just the numbers to take into account commas etc.
+ */
+const getMinorAxisSize = ({
+  dataSetCategories,
+  minorAxisSize,
+  minorAxisDecimals = 0,
+  minorAxisUnit,
+}: Props) => {
```
ntsim, line 20: Add an explicit return type - this is good practice and can help prevent bugs from relying on type inference e.g. if we introduce a new conditional branch that returns a completely different type by accident, the compiler may not complain.

src/explore-education-statistics-common/src/modules/charts/util/getMinorAxisSize.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  const highestValueLength = dataSetCategories
+    .reduce((acc, dataSetCategory) => {
+      const value = Object.values(dataSetCategory.dataSets).reduce(
+        (acc2 = acc, dataSet) => (dataSet.value > acc2 ? dataSet.value : acc2),
+        0,
+      );
+      return value > acc ? value : acc;
+    }, 0)
```
ntsim, line 32: Would consider refactoring this to use Lodash's `maxBy` method as it may be a little more succinct

src/explore-education-statistics-common/src/modules/charts/util/getMinorAxisSize.ts
```diff
@@ -0,0 +1,38 @@
...
+    .toString().length;
+
+  return (highestValueLength + minorAxisUnit.length + minorAxisDecimals) * 12;
```
ntsim, line 35: Would extract 12 into a variable and then put the comment about it next to it so that they are more closely co-located :slightly_smiling_face:

src/explore-education-statistics-common/src/modules/charts/util/__tests__/getMinorAxisSize.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  test('it returns the axis size if set by the user', () => {
+    const size = getMinorAxisSize({
+      dataSetCategories: testDataSetCategories,
+      minorAxisSize: 55,
+      minorAxisDecimals: undefined,
+      minorAxisUnit: '',
+    });
+    expect(size).toBe(55);
+  });
+  test('it takes into account decimal places', () => {
```
ntsim, line 52: Nitpick - but would add line break to separate this case from the previous

src/explore-education-statistics-common/src/modules/charts/util/__tests__/getMinorAxisSize.test.ts
```diff
@@ -0,0 +1,71 @@
...
+      dataSets: {
+        key: {
+          dataSet: { indicator: '', filters: [] },
+          value: 6,
+        },
+      },
+      filter: { value: '', label: '', id: '' },
+    },
+  ];
+  test('it returns the size based on the length of the higest value', () => {
```
ntsim, line 33: Nitpick - but would add line break to separate this case from the test data

ntsim (changes requested):

src/explore-education-statistics-common/src/modules/charts/util/__tests__/getMinorAxisSize.test.ts
duncan-at-hiveit: Also typo in "higest" :)

src/explore-education-statistics-common/src/modules/charts/util/getMinorAxisSize.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  const highestValueLength = dataSetCategories
+    .reduce((acc, dataSetCategory) => {
+      const value = Object.values(dataSetCategory.dataSets).reduce(
+        (acc2 = acc, dataSet) => (dataSet.value > acc2 ? dataSet.value : acc2),
+        0,
+      );
+      return value > acc ? value : acc;
+    }, 0)
+    .toString().length;
```
duncan-at-hiveit, line 33: I don't know if this is a real-world scenario or not, but if we had 2 number values in our `dataSetCategories` like:

`9`

and

`3.5`

I guess this reduction could potentially identify `9` as the highest value and use that to calculate the length against (so 12 in this case), but `3.5` is actually the *longest* number string and so should ideally yield a length of 36.

Again though, I'm not sure if this would ever happen?

amyb-hiveit: Adding the minorAxisDecimals handles this.

src/explore-education-statistics-common/src/modules/charts/util/getMinorAxisSize.ts
amyb-hiveit: Could do it like this, but it's not any more succinct and I think the reduce is a bit clearer.
```
const highestValueLength = dataSetCategories
    .reduce((acc, dataSetCategory) => {
      const dataSetWithHighestValue = maxBy(
        Object.values(dataSetCategory.dataSets),
        'value',
      );
      return dataSetWithHighestValue && dataSetWithHighestValue.value > acc
        ? dataSetWithHighestValue.value
        : acc;
    }, 0)
    .toString().length;
```

PR #2663, opened 2021-07-08
EES-1812 multi select for table header re-ordering

amyb-hiveit:
Ctrl + click to select multiple headers for re-ordering. The implementation is largely based on https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/patterns/multi-drag.md

This causes failures in table_tool_absence_by_characteristic.robot, I had a quick look and think it might be a false positive that they pass on dev as the order of the gender headings is already what it expects after re-ordering. But I could well be wrong.

Screenshot (31)


luke-h1:
I've tested this locally and all seems ok. My only thoughts are to maybe add some sort of indicator to tell the user how many rows they are re-ordering. It also seems pretty hard to get the key-down events to work. Not sure if this is a focus issue that is interfering with the state being set.

ntsim (approved):

PR #2667, opened 2021-07-09
EES-2328 Import School/Provider data if its the only GeographicLevel

markhiveit:
This PR adds new GeographicLevels School and Provider. These new GeographicLevels are only imported if they are the sole GeographicLevel on a subject. Insitution and PlanningArea GeographicLevels are still never imported. All other GeographicLevels are always imported.

The GeographicLevels of a particular subject are now saved in the DataImport table. This is then used to determine whether that row should be imported in `ImporterService#AllowRowImport`. 

`ImporterService#HasSoloAllowedGeographicLevel` has been created to allow us to keep `FileImportService#CheckComplete` working, specifically that the number of Observations imported into the database matches what `ValidatorService#ValidateAndCountObservations` expected.

For future reference, the first commit of this PR should list all the places that need changing when adding a new GeographicLevel.

### DIFF
--- a/infrastructure/templates/datafactory/components/template.json
+++ b/infrastructure/templates/datafactory/components/template.json
@@ -1873,6 +1873,26 @@
                                     },
                                     {
                                         "source": {
+                                            "name": "Provider_Name",
+                                            "type": "String"
+                                        },
+                                        "sink": {
+                                            "name": "Provider_Name",
+                                            "type": "String"
+                                        }
+                                    },
+                                    {
+                                        "source": {
+                                            "name": "Provider_Code",
+                                            "type": "String"
+                                        },
+                                        "sink": {
+                                            "name": "Provider_Code",
+                                            "type": "String"
+                                        }
+                                    },
+                                    {
+                                        "source": {
                                             "name": "Region_Code",
                                             "type": "String"
                                         },
@@ -1898,6 +1918,26 @@
                                         },
                                         "sink": {
                                             "name": "RscRegion_Code",
+                                            "type": "String"
+                                        }
+                                    },
+                                    {
+                                        "source": {
+                                            "name": "School_Name",
+                                            "type": "String"
+                                        },
+                                        "sink": {
+                                            "name": "School_Name",
+                                            "type": "String"
+                                        }
+                                    },
+                                    {
+                                        "source": {
+                                            "name": "School_Code",
+                                            "type": "String"
+                                        },
+                                        "sink": {
+                                            "name": "School_Code",
                                             "type": "String"
                                         }
                                     },
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterLocationService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterLocationService.cs
@@ -18,7 +18,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Processor.Services
             _guidGenerator = guidGenerator;
         }
 
-        public Location Find(
+        public Location FindOrCreate(
             StatisticsDbContext context,
             Country country,
             EnglishDevolvedArea englishDevolvedArea = null,
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterLocationService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterLocationService.cs
@@ -30,8 +30,10 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Processor.Services
             Mat multiAcademyTrust = null,
             OpportunityArea opportunityArea = null,
             ParliamentaryConstituency parliamentaryConstituency = null,
+            Provider provider = null,
             Region region = null,
             RscRegion rscRegion = null,
+            School school = null,
             Sponsor sponsor = null,
             Ward ward = null,
             PlanningArea planningArea = null)
@@ -47,8 +49,10 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Processor.Services
                 multiAcademyTrust,
                 opportunityArea,
                 parliamentaryConstituency,
+                provider,
                 region,
                 rscRegion,
+                school,
                 sponsor,
                 ward,
                 planningArea);
@@ -70,8 +74,10 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Processor.Services
                 multiAcademyTrust,
                 opportunityArea,
                 parliamentaryConstituency,
+                provider,
                 region,
                 rscRegion,
+                school,
                 sponsor,
                 ward,
                 planningArea);
@@ -91,8 +97,10 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Processor.Services
             Mat multiAcademyTrust,
             OpportunityArea opportunityArea,
             ParliamentaryConstituency parliamentaryConstituency,
+            Provider provider,
             Region region,
             RscRegion rscRegion,
+            School school,
             Sponsor sponsor,
             Ward ward,
             PlanningArea planningArea)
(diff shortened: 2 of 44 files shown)


src/GovUk.Education.ExploreEducationStatistics.Data.Services/ObservationPredicateBuilder.cs
```diff
@@ -233,6 +251,16 @@ private static bool ObservationalUnitExists(LocationQuery query)
                 observation => query.RscRegion.Contains(observation.Location.RscRegion_Code));
         }
 
+        private static Expression<Func<Observation, bool>> SchoolPredicate(LocationQuery query)
+        {
+            var schoolOldCodes = query.School.Where(s => s.Length == 3).ToList();
```
benoutram, line 256: Are you sure the laestab is a 3 digit code. This looks like it might be based on the format for an old code.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterLocationService.cs
```diff
@@ -121,7 +131,12 @@ public class ImporterLocationService : BaseImporterService
             return string.Join(separator, observationalUnits
                 .Where(unit => unit != null)
                 .Select(unit =>
-                    $"{unit.GetType()}:{(unit is LocalAuthority la ? la.GetCodeOrOldCodeIfEmpty() : unit.Code)}:{unit.Name}"));
+                {
+                    var code = unit.Code;
+                    if (unit is LocalAuthority la) code = la.GetCodeOrOldCodeIfEmpty();
+                    else if (unit is School sch) code = sch.GetCodeOrOldCodeIfEmpty();
```
benoutram, line 137: I don't think code which is the urn is ever going to be null, so I don't think laestab will ever be used and we could just not import it/store it.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
```diff
@@ -100,6 +109,37 @@ public class ImporterService : IImporterService
...
+                    if (subjectGeographicLevels.Contains(allowedSoloLevel))
+                    {
+                        return true;
+                    }
+                }
+            }
+            return false;
+        }
+
+        public static bool AllowRowImport(HashSet<GeographicLevel> subjectGeographicLevels, GeographicLevel rowGeographicLevel)
```
benoutram, line 133: I think the methods `AllowRowImport` and `HasSoloAllowedGeographicLevel` and probably the fields `IgnoredGeographicLevels` and `AllowedSoloGeographicLevels` don't really belong in ImporterService becaused they are used by `ValidatorService` and `SplitFileService` as well as `ImporterService`. I think you should consider abstracting them out to a common util class, which could have class or method level comments explaining what all this is about and why we do it. It's not particularly obvious at the moment that we have this logic around ignoring school/provider in certain circumstances and I think that would be a good place to add the comments.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
```diff
@@ -100,6 +109,37 @@ public class ImporterService : IImporterService
             GeographicLevel.PlanningArea
         };
 
+        private static readonly List<GeographicLevel> AllowedSoloGeographicLevels = new List<GeographicLevel>
+        {
+            GeographicLevel.Provider,
+            GeographicLevel.School,
+        };
+
+        public static bool HasSoloAllowedGeographicLevel(HashSet<GeographicLevel> subjectGeographicLevels)
```
benoutram, line 118: I'm not sure that solo is the correct term since a file can have two levels, school **and** provider and still get imported.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
```diff
@@ -100,6 +109,37 @@ public class ImporterService : IImporterService
...
+                        return true;
+                    }
+                }
+            }
+            return false;
+        }
+
+        public static bool AllowRowImport(HashSet<GeographicLevel> subjectGeographicLevels, GeographicLevel rowGeographicLevel)
+        {
+            if (HasSoloAllowedGeographicLevel(subjectGeographicLevels))
```
benoutram, line 135: I think there's a potential optimisation that can be made here. Not sure if it would have a big impact on import time but it might be worth doing.

So to begin with, this is the same as

```
            return HasSoloAllowedGeographicLevel(subjectGeographicLevels) ||
                   !IgnoredGeographicLevels.Contains(rowGeographicLevel);
```

The result of `HasSoloAllowedGeographicLevel(subjectGeographicLevels)` never changes yet this `AllowRowImport` method is called for every csv row within loops in `CopyRows`, `ImportFiltersAndLocations`, `GetObservations`, so potentially millions of tiems. In each of those places, before entering the loop you could calculate the result of  `HasSoloAllowedGeographicLevel(subjectGeographicLevels)` and if true, you don't need to call `AllowRowImport` at all.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
markhiveit: It only allows the importing to school or provider data is there is one Geographic level: a file with school and provider data would have no data imported. I created a test subject for this instance.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
markhiveit: I've moved them into ImporterUtils and added comments to HasSoloAllowedGeographicLevel and AllowRowImport. Tried to balance the comment text between saying what they do while not have the comments easily go stale. Happy to change the comments to something else if you'd prefer.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Utils/ImporterUtils.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// If a subject contains a single geographic level that is contained in AllowedSoloGeographicLevels, then
+        /// we import every CSV row of that subject's data file.
+        /// </summary>
+        /// <remarks>
+        /// If a subject contains more than one geographic level or isn't included in AllowedSoloGeographicLevels,
+        /// we then defer to AllowRowImport to determine if a particular row should be imported.
+        /// </remarks>
+        public static bool HasSoloAllowedGeographicLevel(HashSet<GeographicLevel> subjectGeographicLevels)
+        {
+            if (subjectGeographicLevels.Count == 1)
```
benoutram, line 32: I think this whole method can be as `return subjectGeographicLevels.Count == 1 && AllowedSoloGeographicLevels.Contains(subjectGeographicLevels[0]))`

benoutram, line 32: I think this whole method can be simplified to `return subjectGeographicLevels.Count == 1 && AllowedSoloGeographicLevels.Contains(subjectGeographicLevels[0]))`

benoutram (approved):

benoutram (approved):

infrastructure/templates/datafactory/components/template.json
```diff
@@ -1871,6 +1871,26 @@
                                             "type": "String"
                                         }
                                     },
+                                    {
```
markhiveit, line 1874: This PR missed a couple of changes for Data Factory to work with the new GeographicLevels. So for anyone looking at this in the future, looking to see what they need to change to add / update a GeographicLevel, see the additional PR: https://github.com/dfe-analytical-services/explore-education-statistics/pull/2694

PR #2672, opened 2021-07-12
EES-2512 Reformat robot files

markhiveit:


package.json
```diff
@@ -63,7 +63,7 @@
       "prettier --write"
     ],
     "*.robot": [
-      "robotidy --config tests/robot-tests/robotidy.toml"
+      "python3 -m pipenv run robotidy --config tests/robot-tests/robotidy.toml tests/robot-tests/tests/"
```
luke-h1, line 66: You don't need to pass it a path here. This would be quite inefficient if you had to format the entire test directory even if you only changed 1 file.

luke-h1 (changes requested):

luke-h1 (approved):

PR #2673, opened 2021-07-12
Update Docker Compose setup and README

ntsim:
This PR removes the old (and unused) Docker containers for data and content APIs.


README.md
```diff
@@ -13,6 +13,7 @@ The project is primarily composed of two areas:
   - Depends on:
     - Content API
     - Data API
+    - Notifier
```
markhiveit, line 16: It is required to have Notifier running if you want to test subscription emails etc. locally

README.md
```diff
@@ -64,28 +68,20 @@ You will need the following groups of dependencies to run the project successful
...
-  AzureStorageEmulator.exe init /server .
-  ```
-  
-   You will currently need to use Windows (or a Windows VM with ports 10001, 10002 and 10003 exposed) to run the service 
-   due to a dependency on Azure Storage Emulator for development, specifically the table storage component which is not 
-   supported by [Azurite](https://github.com/Azure/Azurite). However this does not apply to every component of the 
-   service.
+   - [Azure Storage Emulator](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-emulator) -
+     installer for Windows x64 is [here](https://download.microsoft.com/download/3/9/F/39F968FA-DEBB-4960-8F9E-0E7BB3035959/SQLEXPR_x64_ENU.exe)
+   - [Docker and Docker Compose](https://docs.docker.com/)
```
markhiveit, line 75: This sounds like they're both the same option. I would reword it make it clear all the different options you have - azure storage emulator, docker/azurite, VM, etc. It could even be its own section maybe.

luke-h1 (approved):

markhiveit (changes requested):

README.md
ntsim: Yes, that's what I've said :smile: - see the Notifier change below

README.md
ntsim: Good point, will update the wording slightly

PR #2675, opened 2021-07-12
EES-2489 Split up large HydrateRelease query

markhiveit:


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/ManageContentPageService.cs
```diff
@@ -74,24 +79,57 @@ public class ManageContentPageService : IManageContentPageService
...
-                .Include(r => r.Publication)
-                .ThenInclude(publication => publication.Topic.Theme)
-                .Include(r => r.Type)
-                .Include(r => r.Content)
-                .ThenInclude(join => join.ContentSection)
-                .ThenInclude(section => section.Content)
-                .ThenInclude(content => content.Comments)
-                .ThenInclude(comment => comment.CreatedBy)
-                .Include(r => r.Updates);
+            await _contentDbContext.Entry(release)
```
ntsim, line 84: Just a general question - why is this more efficient than the alternative?

What is Entity Framework doing in both scenarios?

benoutram: I was just thinking about this too.

I read the Jira comment about not investigating the cause of the performance problem before applying this fix and I agree this is the most likely cause of the issue since this will be executed in a single query there's likely to be tons of redundant data in the result. Repeated data for every previous release, for every content section, content block, legacy release etc due to the joins. I imagine the database can respond quickly due to the indexes that exist but the delay will be due to a high level of I/O between database and Admin app, along with a high level of CPU on the Admin app to process the result.

I would have thought this solution here won't be the most optimal since some joins will be beneficial over individual selects, but this would need determining on a case by case basis and probably not worth the effort unless we find there's now a new performance degradation for other Releases.

It doesn't help here that we hide away the logic of what data is required within the Automapper config, and without that it would be easier to think about what related groups of entities make sense to fetch together.

For example, I think the following groups of queries would be more optimal than these individual selects.

Publication joined with theme, topic, contact.
Legacy releases
Previous releases (summary fields only)
Content sections joined with content blocks
Comments joined with user

markhiveit: Talked this through with Ben - I've updated the PR such that it doesn't get rid of all JOINs, but uses them where appropriate.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/ManageContentPageService.cs
```diff
@@ -74,24 +82,45 @@ public class ManageContentPageService : IManageContentPageService
...
-                .Include(r => r.Type)
-                .Include(r => r.Content)
-                .ThenInclude(join => join.ContentSection)
-                .ThenInclude(section => section.Content)
-                .ThenInclude(content => content.Comments)
-                .ThenInclude(comment => comment.CreatedBy)
-                .Include(r => r.Updates);
+            release.Updates = _contentDbContext.Update
+                .Where(u => u.ReleaseId == release.Id)
+                .ToList();
```
benoutram, line 89: There's a couple of places that are ToList here, rather than ToListAsync. I would make everything ToListAsync.

benoutram (approved):

PR #2676, opened 2021-07-12
EES-2513 - External Methodologies and EES Methodologies can coexist

duncan-at-hiveit:
# This PR
- refactors various bits of back and front end code that work under the assumption that External Methodologies and EES Methodologies are always mutually exclusive.

In actual fact, we need to allow Analysts to still link to externally hosted Methodologies whilst they are working on their Draft Methodologies.  There's also no reason why a Publication, which can now adopt several Methodologies, cannot use an enternal one and one published on EES, although it's not likely.

The "Edit" button for editing external methodology details has also been renamed "Edit externally hosted methodology" to avoid confusion, as there are now potentially more action buttons available on the Dashboard per Publication, which can be a little confusing.  There's perhaps better ways to present this, but might want a bit of a design overhaul.

# Notable changes
* Create Methodology and Manage External Methodologies auth handlers now don't enforce rules around whether or not the Publication is already linked to the other type of Methodology.  Tests updated to reflect this.
* The front end controls in `MethodologySummary.tsx` have been moved around a bit so they are no longer mutually exclusive but the user is able to just be presented with whichever buttons they're allowed to currently use.
* Various bits of the front end code rendered external and EES methodologies separately.  New code puts them in an array and treats them more uniformally where appropriate (not so much in the Next.js code where the routes are treated differently using the `as` attribute).
* Fixed a bug in `PublicationReleaseHelpAndSupportSection.tsx` where an invalid combination of `to` and `as` parameters were being supplied for external methodology routes. 

# Test results

All but 1 unrelated Robot test passing:

image

image


src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseContent.tsx
```diff
@@ -166,28 +190,15 @@ const ReleaseContent = () => {
                   View data and files
                 </a>
               </li>
-              {release.publication.methodologies.map(methodology => (
-                <li key={methodology.id}>
+              {allMethodologies.map(methodology => (
+                <li key={methodology.key} className="govuk-!-margin-bottom-9">
```
ntsim, line 194: This seems like quite a lot of bottom margin, why are we adding this?

benoutram (approved):
Cool, I can't see any reason to undo any of this in future either unless we are required to do so, just continue allowing external methodologies to co-exist. I also checked that the update of the external methodology is creating the publishing event so no problem there either.

src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseHelpAndSupportSection.tsx
```diff
@@ -30,31 +53,17 @@ const ReleaseHelpAndSupportSection = ({
...
-                <p key={methodology.id} className="govuk-!-margin-bottom-9">
+              {allMethodologies.map(methodology => (
+                <p key={methodology.key} className="govuk-!-margin-bottom-9">
                   {editingMode === 'edit' ? (
                     <a>{`${methodology.title}`}</a>
                   ) : (
-                    <Link to={`/methodology/${methodology.id}/summary`}>
-                      {methodology.title}
-                    </Link>
+                    <Link to={methodology.url}>{methodology.title}</Link>
```
ntsim, line 63: Looks like this should be list rather than a bunch of paragraphs. The bottom margin is also too generous.

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseContent.tsx
benoutram: I've tried this out now and we don't need any padding at all so I've removed it. Looked a bit strange before. 🙂

src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseHelpAndSupportSection.tsx
benoutram: I've done this now, thanks.

PR #2681, opened 2021-07-13
EES-2363 Add summary field to ReleaseFile and allow updating via endpoint

ntsim:
This PR introduces a new `Summary` field to `ReleaseFile` entities, which can then be updated via a new update endpoint (using `PATCH /release/{releaseId}/files/{fileId}`). This update endpoint supersedes the previous endpoint which only updated file names.

We have implemented this as a PATCH as it's not necessary for all files to have to update their name/title or summary. These fields are optional in the request, and will not unset the relevant entity's properties if set to null. 

If you wish to set these fields to empty values - use empty strings.

## Related changes

- Adds missing various checks to `ReleaseFileService.GetFile` to ensure that the user is allowed to view the release and the associated file.
- Consolidates `AncillaryFileInfo` into `FileInfo` as it only contains common properties that are not unique to ancillary files.
- Various cleanup across `ReleaseFileRepository` get/update methods.

## Other changes

- Renamed various `Name` fields to `Title` in upload/update requests. This is in preparation for further alignment of this field name across the frontend and backend (see [EES-2517](https://dfedigital.atlassian.net/browse/EES-2517)) for consistency.


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseFileServicePermissionTests.cs
```diff
@@ -132,9 +167,15 @@ await PolicyCheckBuilder<SecurityPolicies>()
...
                     {
                         var service = SetupReleaseFileService(userService: userService.Object);
-                        return service.UploadAncillary(releaseId: _release.Id,
-                            formFile: new Mock<IFormFile>().Object,
-                            name: "");
+                        return service.UploadAncillary(
+                            releaseId: _release.Id,
+                            upload: new ReleaseAncillaryFileUploadViewModel
+                            {
+                                File = Mock.Of<IFormFile>(),
```
duncan-at-hiveit, line 174: Didn't know of this short-hand before.  Worth mentioning in #development channel

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseFileServiceTests.cs
```diff
@@ -1467,23 +1472,56 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
             {
-                var service = SetupReleaseFileService(contentDbContext: contentDbContext,
+                var service = SetupReleaseFileService(
+                    contentDbContext: contentDbContext,
                     blobStorageService: blobStorageService.Object);
 
                 var result = await service.GetFile(Guid.NewGuid(), Guid.NewGuid());
+
+                Assert.True(result.IsLeft);
+                Assert.IsType<NotFoundResult>(result.Left);
```
duncan-at-hiveit, line 1493: Can now use `result.AssertNotFound()` which'll do these 2 checks in one.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseFileServiceTests.cs
```diff
@@ -1740,38 +1778,228 @@ await using (var contentDbContext = InMemoryApplicationDbContext())
...
+                    releaseFile.ReleaseId,
+                    releaseFile.FileId,
+                    new ReleaseFileUpdateViewModel
+                    {
+                        Title = "New file title",
+                        Summary = "New file summary"
+                    }
+                );
 
                 Assert.True(result.IsRight);
```
duncan-at-hiveit, line 1827: Doesn't really gain you anything here, but just so you're aware, you can also do:

`var unit = result.AssertRight()`

To both assert and extract the contents of the right in one go.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseFileServiceTests.cs
```diff
@@ -1824,7 +2052,15 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+                    release.Id,
+                    new ReleaseAncillaryFileUploadViewModel
+                    {
+                        Title = "Test name",
+                        Summary = "Test summary",
+                        File = formFile
+                    }
+                );
 
                 Assert.True(result.IsRight);
```
duncan-at-hiveit, line 2065: Now this'd be a better candidate for `var upload = result.AssertRight();`

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleasesController.cs
```diff
@@ -101,7 +101,7 @@ public async Task<ActionResult<IEnumerable<DataFileInfo>>> GetDataFileInfo(Guid
         [RequestFormLimits(ValueLengthLimit = int.MaxValue, MultipartBodyLengthLimit = int.MaxValue)]
         public async Task<ActionResult<DataFileInfo>> AddDataFilesAsync(Guid releaseId,
             [FromQuery(Name = "replacingFileId")] Guid? replacingFileId,
-            [FromQuery(Name = "name")] string subjectName,
+            [FromQuery(Name = "title")] string subjectName,
```
duncan-at-hiveit, line 104: Is there a reason we're doing this manually like this and not being handed a `AddDataFileViewModel` or similar, as we do in other places?  Like we currently use `ReleaseAncillaryFileUploadViewModel`?  I'm aware this is pre-existing code and not something you've added - just wondering if there was a reason...

src/GovUk.Education.ExploreEducationStatistics.Admin/Models/DataFileInfo.cs
```diff
@@ -9,17 +10,19 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Models
     public class DataFileInfo : FileInfo
     {
         [JsonConverter(typeof(EnumToEnumValueJsonConverter<FileType>))]
-        public new readonly FileType Type = FileType.Data;
+        public new const FileType Type = FileType.Data;
```
duncan-at-hiveit, line 13: Why the need for the `new` keyword here, I wonder?  Maybe you don't need it with `const`?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseFileRepository.cs
```diff
@@ -99,6 +102,11 @@ public async Task<ReleaseFile> Get(Guid releaseId, Guid fileId)
                     && releaseFile.FileId == fileId);
         }
 
+        public async Task<Either<ActionResult, ReleaseFile>> FindOrNotFound(Guid releaseId, Guid fileId)
```
duncan-at-hiveit, line 105: I think we already have this Find / FindOrNotFound pattern for free within AbstractRepository<T> - you could look to extend it and see if it gives you what you need.  It might give you more than you need too!

src/GovUk.Education.ExploreEducationStatistics.Content.Model/ReleaseFile.cs
```diff
@@ -8,15 +9,17 @@ public class ReleaseFile
         public Guid Id { get; set; }
 
         [JsonIgnore]
-        public Release Release { get; set; }
-        
+        public Release Release { get; set; } = null!;
```
duncan-at-hiveit, line 12: Why the need for the exclamation mark in this syntax?  If it's possible to set this to null, then should it be `public Release?` rather than `public Release`   ?

Then presumably you wouldn't have to tell the compiler to be forgiving?

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleasesController.cs
duncan-at-hiveit: Oh, I've answered my own question - query params.  Which changes my question to, is there any preference in using query params in POSTs over a request body?  Understandably we need the body wherever we upload Files, but just personal preference for me is to use a body all over the place.

Anyway, nowt to do with this PR, so forget it :)

ntsim: Yep, so I think this is just technical debt from when this was originally written. We really should be binding it to a model using the `[FromForm]` annotation like we've done in the ancillary file upload for this PR). 

I decided against doing this as part of this ticket as there's a bunch of things to clean up here :smile:

I've created this ticket for it: https://dfedigital.atlassian.net/browse/EES-2521

duncan-at-hiveit (dismissed):
Left a few comments and questions, but nothing that should hold this up until I'm back in on Friday, so approving now, and feel free to check out the comments.  Ta.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseFileRepository.cs
ntsim: I thought about this, but unfortunately it's only in the `Data.Services` project and is currently specific for the `StatisticsDbContext`. Thought it best to avoid re-purposing it here as it's used in a number of places.

It would be good if we could create a common base repository class for stuff like this that we could have in `Common`. I've created a ticket for it anyway: https://dfedigital.atlassian.net/browse/EES-2529

src/GovUk.Education.ExploreEducationStatistics.Content.Model/ReleaseFile.cs
ntsim: As we have enabled nullables, reference types need to be initialised upon construction, however, this really requires you to have explicit constructor methods that initialize everything correctly.  Unfortunately, this doesn't play well with Entity Framework as I _think_ it typically uses setters to construct everything, hence we need an empty constructor (hence fields are potentially null upon construction).

I believe a similar thing happens with Hibernate, so you typically need to have an empty constructor method.

So basically, a workaround for this is to use the null-forgiving operator to basically re-assure the C# compiler that everything is fine. By using it we say that our field will be initialized before we actually use it - by something like Entity Framework.

src/GovUk.Education.ExploreEducationStatistics.Admin/Models/DataFileInfo.cs
ntsim: Without `new`, we get this compiler warning:

> The keyword 'new' is required on 'Type' because it hides property 'FileType GovUk.Education.ExploreEducationStatistics.Common.Model.FileInfo.Type'

Given it's the same signature as the base class, you'd think it would be fine as it's not really hiding anything, but hey, I don't make the compiler rules :man_shrugging: 

I didn't really need to change it to `const` either, but I originally got the suggestion in another case that was similar to this. Seemed to make sense so thought I'd throw it in here too!

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseFileServiceTests.cs
ntsim: Done!

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseFileServiceTests.cs
ntsim: Done!

src/GovUk.Education.ExploreEducationStatistics.Content.Model/ReleaseFile.cs
duncan-at-hiveit: Good stuff, thanks!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseFileRepository.cs
duncan-at-hiveit: Cool ta

src/GovUk.Education.ExploreEducationStatistics.Admin/Models/DataFileInfo.cs
duncan-at-hiveit: Yeaqh I would've thought it would have been fine too, but meh!  Cool thanks.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleasesController.cs
duncan-at-hiveit: Yup understandable, I knew it was out of the remit of this PR, but thanks for creating the ticket!

duncan-at-hiveit (approved):

PR #2683, opened 2021-07-13
EES-2509 Optimise building All Methodologies view

benoutram:
This is an experimental PR attempting to optimise fetching and building the view model for the Public All Methodologies page.

The cause of the performance problem appears to be down to fetching the publication in  methods `MethodologyParentRepository#GetByPublication` and `MethodologyRepository#GetLatestPublishedByPublication` to check for its existence before proceeding.

We called these methods a number of times, one for each Publication each time we were attempting to check if the Publication has child Methodology nodes, and once to build the Methodologies for the Publication node view model.

While it's nice to check that the publication exists, in theory it should be impossible to reach this code with a publication id that hasn't first come from some data containing actual publications.

This PR also reorganises the way we build up the view model for the All Methodologies page.

It first selects from the database only the `Id` and `Title` data that's going to be required for the Themes/Topics/Publications rather than querying all of the fields. This is done by projecting from the source themes queryable into the new view model _before_ calling `ToList` which is the border in code between database and in-memory execution.

After building the view model and adding the methodologies it can then filter out leaf nodes with no methodologies at the end. This was previously done before building the view models for any nodes but resulted in additional overhead.

This PR is experimental because I would like to see how this performs in a real Azure environment with more data.

I think its quite likely that there will still be a delay and that we need to think of an alternative approach for caching the response.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Api/ViewModels/AllMethodologiesViewModels.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Api/ViewModels/AllMethodologiesViewModels.cs
@@ -1,0 +1,65 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using GovUk.Education.ExploreEducationStatistics.Publisher.Model.ViewModels;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Api.ViewModels
+{
+    public class AllMethodologiesThemeViewModel : IComparable<AllMethodologiesThemeViewModel>
+    {
+        public Guid Id { get; set; }
+        public string Title { get; set; }
+        public List<AllMethodologiesTopicViewModel> Topics { get; set; }
+
+        public void RemoveTopicNodesWithoutMethodologiesAndSort()
+        {
+            // Remove all publications without any methodologies
+            Topics.ForEach(topic => topic.RemovePublicationNodesWithoutMethodologiesAndSort());
+
+            // Remove all topics without any publications
+            Topics = Topics
+                .Where(topic => topic.Publications.Any())
+                .ToList();
+
+            Topics.Sort();
+        }
+
+        public int CompareTo(AllMethodologiesThemeViewModel other)
+        {
+            return other == null ? 1 : string.Compare(Title, other.Title, StringComparison.Ordinal);
+        }
+    }
+
+    public class AllMethodologiesTopicViewModel : IComparable<AllMethodologiesTopicViewModel>
+    {
+        public Guid Id { get; set; }
+        public string Title { get; set; }
+        public List<AllMethodologiesPublicationViewModel> Publications { get; set; }
+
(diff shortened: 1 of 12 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Api/ViewModels/AllMethodologiesViewModels.cs
```diff
@@ -0,0 +1,66 @@
...
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Api.ViewModels
+{
+    public class AllMethodologiesThemeViewModel : IComparable<AllMethodologiesThemeViewModel>
+    {
+        public Guid Id { get; set; }
+        public string Title { get; set; }
+        public List<AllMethodologiesTopicViewModel> Topics { get; set; }
+
+        public void RemoveTopicLeafNodesAndSort()
```
duncan-at-hiveit, line 14: Might consider just renaming this method slightly to indicate which leaf nodes we'd be removing e.g.

`RemoveTopicNodesWithoutMethodologiesAndSort`

Also below:

`RemovePublicationNodesWithoutMethodologiesAndSort`

The fact that they're leaf nodes didn't immediately occur to me that that meant they didn't have Methodologies under them.

src/GovUk.Education.ExploreEducationStatistics.Content.Api/ViewModels/AllMethodologiesViewModels.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
```
duncan-at-hiveit, line 15: Cool, do we have any existing tests covering this node removal?  If not, it's probably an important one to add coverage to, as this is the place where we stop potentially-not-yet-public Publications from being included in the tree.

src/GovUk.Education.ExploreEducationStatistics.Content.Api/ViewModels/AllMethodologiesViewModels.cs
benoutram: Thanks, I've renamed them now.

src/GovUk.Education.ExploreEducationStatistics.Content.Api/ViewModels/AllMethodologiesViewModels.cs
benoutram: Yes we do in the form of these three tests in MethodologyServiceTests

```
GetTree_ThemeWithoutTopicsIsNotIncluded
GetTree_ThemeWithoutPublicationsIsNotIncluded
GetTree_ThemeWithoutPublishedMethodologiesIsNotIncluded
```

but I will add some more direct coverage for these methods and update the PR. Thanks.

benoutram: More unit tests added.

duncan-at-hiveit (approved):

PR #2685, opened 2021-07-14
EES-2513 - avoid Stale Reference errors

duncan-at-hiveit:
This PR:
- moves the lookup of parents from a single initial lookup in `user_waits_until_parent_contains_element` and `user_waits_until_parent_does_not_contain_element` to a dynamic lookup within the `waiting._wait_until` loop, thus ensuring that the parent being used to find the child is always fresh
- refactors `get_child_element` and `get_child_elements` to reuse much of the code used within `user_waits_until_parent_contains_element` for easier maintenance.
- adds a warning within `get_child_element` if the lookup resulted in more than one element being found.  Previous behaviour was just to return the first element that the query matched without any warning.  This highlights some areas where we're currently "getting away with it".
- corrects a table xpath expression that was only succeeding previously due to an overzealous rewrite of the selector within `utilities.py`.

## Test results

Note that the screenshot says that there's 1 test failure, but this is actually just due to the fact that I changed a test name so that it was unique within its suite, and so the "failure" indicated here is just the originally named test that failed during the first run.  Its newly named test passed in the following run (otherwise there would have been 2 test failures indicated here, one under the old name, one under the new).

image

### DIFF
--- a/tests/robot-tests/tests/libs/utilities.py
+++ b/tests/robot-tests/tests/libs/utilities.py
@@ -292,6 +292,7 @@ def __normalise_child_locator(parent_locator: object, child_locator: str) -> str
     else:
         raise_assertion_error(f"Parent locator was neither a str or a WebElement - {parent_locator}")
 
+
 def __get_parent_webelement_from_locator(parent_locator: object, timeout: int = None, error: str = '') -> WebElement:
     if isinstance(parent_locator, str):
         sl.wait_until_page_contains_element(parent_locator, timeout=timeout, error=error)
--- a/tests/robot-tests/tests/libs/utilities.py
+++ b/tests/robot-tests/tests/libs/utilities.py
@@ -6,7 +6,6 @@ from logging import warning
 from SeleniumLibrary.utils import is_noney
 from robot.libraries.BuiltIn import BuiltIn
 from SeleniumLibrary import ElementFinder
-from SeleniumLibrary.errors import ElementNotFound
 from SeleniumLibrary.keywords.waiting import WaitingKeywords
 from selenium.webdriver.remote.webelement import WebElement
 import os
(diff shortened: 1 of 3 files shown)


tests/robot-tests/tests/libs/utilities.py
```diff
@@ -305,4 +277,27 @@ def remove_substring_from_right_of_string(string, substring):
...
+
+def __normalise_child_locator(parent_locator: object, child_locator: str) -> str:
+    if isinstance(parent_locator, str):
+        return child_locator
+    elif isinstance(parent_locator, WebElement):
+        # the below substitution is necessary if the parent is a Selenium WebElement in order to correctly find the 
+        # parent's descendants.  Without the preceding dot, the double forward slash breaks out of the parent container
+        # and returns the xpath query to the root of the DOM, leading to false positives or incorrectly found DOM 
+        # elements.
+        return child_locator.replace("xpath://", "xpath:.//")
```
markhiveit, line 291: I know this is an existing issue before your change, but I think it's possible to have a string selector without the `xpath:` prefix, as any locator without the prefix defaults to xpath. So I think we need to regex replace "^//" with ".//" too

luke-h1 (approved):

markhiveit (changes requested):

tests/robot-tests/tests/libs/utilities.py
duncan-at-hiveit: Cool, good catch.  Added the following substitution, which yields the following results:

```
>>> re.sub(r'^(xpath:)?//', ".//", "asdf")
'asdf'
>>> re.sub(r'^(xpath:)?//', ".//", "xpath:hello")
'xpath:hello'
>>> re.sub(r'^(xpath:)?//', ".//", "xpath://hello")
'.//hello'
>>> re.sub(r'^(xpath:)?//', ".//", "//hello")
'.//hello
```

markhiveit (approved):

PR #2692, opened 2021-07-16
EES-2536 Fix custom selectors failing for negation keywords

ntsim:
This fixes cases where we are using keywords like `user checks page does not contain`. These were primarily failing due to keywords that were waiting for elements that would never appear.

It also turns out that we need to return lists of elements for custom selectors instead of singular elements.

### DIFF
--- a/tests/robot-tests/tests/libs/common.robot
+++ b/tests/robot-tests/tests/libs/common.robot
@@ -28,17 +28,22 @@ do this on failure
     END
 
 custom testid locator strategy
-    [Arguments]  ${browser}  ${test_id}   ${tag}  ${constraints}
-    ${element}=  user gets testid element  ${test_id}
-    [Return]  ${element}
+    [Arguments]    ${browser}    ${label}    ${tag}    ${constraints}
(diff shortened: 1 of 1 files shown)


tests/robot-tests/tests/libs/common.robot
```diff
@@ -27,18 +27,24 @@ do this on failure
         set selenium timeout    3
     END
 
-custom testid locator strategy
```
markhiveit, line 30: I don't think this should be deleted? Since it's used on L44?

markhiveit (changes requested):

tests/robot-tests/tests/libs/common.robot
```diff
@@ -27,18 +27,24 @@ do this on failure
...
-    [Arguments]  ${browser}  ${test_id}   ${tag}  ${constraints}
-    ${element}=  user gets testid element  ${test_id}
-    [Return]  ${element}
+check that variable is not empty
```
markhiveit, line 30: This new keyword doesn't seem to be used either.

markhiveit (approved):

PR #2701, opened 2021-07-20
EES-2472 prevent chart ref line labels being cut off

amyb-hiveit:
Renders a custom label for reference lines that changes the text alignment if the label is at the start or end. This is for line charts only currently as I'm not sure it's necessary for bar charts and would look odd with wide bars

Before:
before

After:
after

### DIFF
--- a/src/explore-education-statistics-common/src/modules/charts/components/CustomReferenceLineLabel.tsx
+++ b/src/explore-education-statistics-common/src/modules/charts/components/CustomReferenceLineLabel.tsx
@@ -1,0 +1,44 @@
+import { ReferenceLine } from '@common/modules/charts/types/chart';
+import { ChartData } from '@common/modules/charts/types/dataSet';
+import React from 'react';
+import { ReferenceLineProps } from 'recharts';
+
+interface Props {
+  chartData: ChartData[];
+  referenceLine: ReferenceLine;
+  referenceLineProps: ReferenceLineProps;
+}
+
+const CustomReferenceLineLabel = ({
+  chartData,
+  referenceLine,
+  referenceLineProps,
+}: Props) => {
+  const getTextAnchor = () => {
+    if (referenceLine.position === chartData[0].name) {
+      return 'start';
+    }
+    if (referenceLine.position === chartData[chartData.length - 1].name) {
+      return 'end';
+    }
+    return 'middle';
+  };
+
+  const getY = () => {
+    const y = referenceLineProps.viewBox?.y ?? 0;
+    const height = referenceLineProps.viewBox?.height ?? 0;
+    return height / 2 + y;
+  };
+
+  return (
+    <text
+      x={referenceLineProps.viewBox?.x}
+      y={getY()}
+      textAnchor={getTextAnchor()}
+    >
+      <tspan>{referenceLine.label}</tspan>
+    </text>
+  );
+};
+
+export default CustomReferenceLineLabel;
--- a/src/explore-education-statistics-common/src/modules/charts/components/LineChartBlock.tsx
+++ b/src/explore-education-statistics-common/src/modules/charts/components/LineChartBlock.tsx
@@ -205,13 +205,13 @@ const LineChartBlock = ({
             <ReferenceLine
               key={`${referenceLine.position}_${referenceLine.label}`}
               x={referenceLine.position}
-              label={props =>
-                CustomReferenceLineLabel({
-                  chartData,
-                  referenceLine,
-                  referenceLineProps: props,
-                })
-              }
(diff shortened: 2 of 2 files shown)


src/explore-education-statistics-common/src/modules/charts/components/LineChartBlock.tsx
```diff
@@ -204,7 +205,13 @@ const LineChartBlock = ({
             <ReferenceLine
               key={`${referenceLine.position}_${referenceLine.label}`}
               x={referenceLine.position}
-              label={referenceLine.label}
+              label={props =>
+                CustomReferenceLineLabel({
```
ntsim, line 209: Should this be a JSX element?

src/explore-education-statistics-common/src/modules/charts/components/CustomReferenceLineLabel.tsx
```diff
@@ -0,0 +1,44 @@
+import { ReferenceLine } from '@common/modules/charts/types/chart';
+import { ChartData } from '@common/modules/charts/types/dataSet';
+import React from 'react';
+import { ReferenceLineProps } from 'recharts';
+
+interface Props {
+  chartData: ChartData[];
+  referenceLine: ReferenceLine;
+  referenceLineProps: ReferenceLineProps;
```
ntsim, line 9: Might be better to have `Props extend ReferenceLineProps` and then spread the props in?

src/explore-education-statistics-common/src/modules/charts/components/LineChartBlock.tsx
```diff
@@ -39,6 +39,7 @@ import {
   YAxis,
 } from 'recharts';
 import getDataSetCategoryConfigs from '@common/modules/charts/util/getDataSetCategoryConfigs';
+import CustomReferenceLineLabel from '@common/modules/charts/components/CustomReferenceLineLabel';
```
ntsim, line 42: Don't we need these for vertical/horizontal bar blocks too?

ntsim (changes requested):

src/explore-education-statistics-common/src/modules/charts/components/CustomReferenceLineLabel.tsx
amyb-hiveit: Why would that be better?

src/explore-education-statistics-common/src/modules/charts/components/LineChartBlock.tsx
amyb-hiveit: Yes it should!

src/explore-education-statistics-common/src/modules/charts/components/LineChartBlock.tsx
amyb-hiveit: This solution doesn't really work for them, it looks wrong with the width of the bars.  I've checked with Laura and she's happy to leave the bar charts as they are now, with the labels centre aligned, as the bars provide a bit of padding which makes the text less likely to overflow the chart and reference lines are used less on them.

src/explore-education-statistics-common/src/modules/charts/components/CustomReferenceLineLabel.tsx
ntsim: Generally, it's a good idea to try and keep props as flat as possible and avoid object/array props. The main reason is that if we want to use things like the `memo` HOC (for performance reasons), we need to define a custom memoization strategy e.g. to check over the contents of the object/array props (as well as the top-level props). This leads to more code and we can't just use the default by-reference memoization strategy.

Of course, it's not always practical to flatten all the props down. For example, some components may need to use all of the different parts of a really complex object(s). That said, I'd personally say it's a good idea to try and avoid passing in more props/data than a component actually needs :wink: 

In this case, I don't think it makes too much of a difference though. Happy to leave it as-is, but something to think about for other components :slightly_smiling_face:

ntsim (approved):

src/explore-education-statistics-common/src/modules/charts/components/LineChartBlock.tsx
ntsim: Cool, good enough for me!

src/explore-education-statistics-common/src/modules/charts/components/CustomReferenceLineLabel.tsx
amyb-hiveit: Thanks, will bear this in mind in future.

PR #2705, opened 2021-07-20
EES-2527 EES-2528 break long links in content and section index

amyb-hiveit:
Forces long links to break in dfe generated content (methodology and release content) and in the section index list on methodologies. Note that this solution doesn't work in IE 11, but the alternative is to use `break-all` which will cause words in links to break when they should wrap on to the next line, so I figured having it wrong in IE11 was the lesser evil. 

As a bonus I've trimmed the section index links as noticed there was white space creeping in on https://explore-education-statistics.service.gov.uk/methodology/longer-term-destinations-methodology

### DIFF
--- a/src/explore-education-statistics-common/src/components/ContentSectionIndex.module.scss
+++ b/src/explore-education-statistics-common/src/components/ContentSectionIndex.module.scss
@@ -1,3 +1,0 @@
-.link {
-  word-break: break-word;
-}
--- a/src/explore-education-statistics-common/src/components/ContentSectionIndex.module.scss
+++ b/src/explore-education-statistics-common/src/components/ContentSectionIndex.module.scss
@@ -1,0 +1,3 @@
+.link {
+  word-break: break-word;
+}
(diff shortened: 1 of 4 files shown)


src/explore-education-statistics-common/src/components/ContentSectionIndex.module.scss
```diff
@@ -0,0 +1,3 @@
+.link {
+  word-break: break-word;
```
ntsim, line 2: Maybe it might be better if we just do this for all link stlyes in `_typography.scss`?

amyb-hiveit: I'm a bit wary of making a global change in case there's unexpected consequences, but I'll try it locally and have a look around the site to see if it causes any problems.

amyb-hiveit: Seems fine so have changed it, good call :)

ntsim (approved):

PR #2706, opened 2021-07-21
EES-2533 Add official stats info

amyb-hiveit:
Shows Official Statistics info on releases.

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseHelpAndSupportSection.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseHelpAndSupportSection.tsx
@@ -70,12 +70,12 @@ const ReleaseHelpAndSupportSection = ({
             <p>No methodologies added.</p>
           )}
         </AccordionSection>
-        {release.type && release.type.title === ReleaseType.NationalStatistics && (
+        {release.type.title === ReleaseType.NationalStatistics && (
           <AccordionSection heading="National Statistics" headingTag="h3">
             <NationalStatisticsSection />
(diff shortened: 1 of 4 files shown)


src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseHelpAndSupportSection.tsx
```diff
@@ -74,6 +75,11 @@ const ReleaseHelpAndSupportSection = ({
             <NationalStatisticsSection />
           </AccordionSection>
         )}
+        {release.type && release.type.title === ReleaseType.OfficialStatistics && (
```
ntsim, line 78: Is `type` optional? Could make this more succinct with `release.type.title === ReleaseType.OfficialStatistics`. Think the same applies for the section above too.

If it is supposed to be optional, then I would suggest changing the corresponding field on `Release` optional, and then express this as `release.type?.title === ReleaseType.OfficialStatistics` (using optional chaining)

src/explore-education-statistics-common/src/modules/find-statistics/components/OfficialStatisticsSection.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+
+const OfficialStatisticsSection = () => (
+  <>
+    <p className="govuk-body">
```
ntsim, line 5: No need to put `govuk-body` on `p` - it's already applied globally already

ntsim (changes requested):

ntsim (approved):

PR #2712, opened 2021-07-22
EES-2505 fix adding chart data sets

amyb-hiveit:
My previous solution for adding 'All options' to filters for chart data sets (https://github.com/dfe-analytical-services/explore-education-statistics/pull/2591) was causing charts to not show any data if there were filters with only one option in the data set.

This re-does that work to take it into account.


ntsim:
Opening new PR against `master`

PR #2714, opened 2021-07-22
EES-2371 Add footnotes to data guidance endpoint and file

ntsim:
This PR adds footnotes to the data endpoint and the associated data guidance file (in the All files zip).

## Other changes

- Fixed indentation issue with `ol` elements in HTML to text conversion. This would previously lead to misalignment of the list item content if there were over 10 items in a list.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/HtmlToTextUtilsTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/HtmlToTextUtilsTests.cs
@@ -216,6 +216,34 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils
         }
 
         [Fact]
+        public async Task HtmlToText_OrderedList_OverTenItemsWithMultiline()
+        {
+            var text = await HtmlToTextUtils.HtmlToText(
+                @"
+                <ol>
+                    <li>List item 1</li>
+                    <li>List item 2</li>
+                    <li>List item 3</li>
+                    <li>List item 4</li>
+                    <li>List item 5</li>
+                    <li>List item 6</li>
+                    <li>List item 7</li>
+                    <li>List item 8</li>
+                    <li>
+                        <p>List item 9</p>
+                        <p>Over multiple lines</p>
+                    </li>
+                    <li>List item 10</li>
+                    <li>
+                        <p>List item 11</p>
+                        <p>Over multiple lines</p>
+                    </li>
+                </ol>");
+
+            Snapshot.Match(text);
+        }
+
+        [Fact]
         public async Task HtmlToText_UnorderedList_WithNestedText()
         {
             var text = await HtmlToTextUtils.HtmlToText(
@@ -271,6 +299,34 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils
                                 </ul>
                             </li>
                         </ul>
+                    </li>
+                </ul>");
+
+            Snapshot.Match(text);
+        }
+
+        [Fact]
+        public async Task HtmlToText_UnorderedList_OverTenItemsWithMultiline()
+        {
+            var text = await HtmlToTextUtils.HtmlToText(
+                @"
+                <ul>
+                    <li>List item 1</li>
+                    <li>List item 2</li>
+                    <li>List item 3</li>
+                    <li>List item 4</li>
+                    <li>List item 5</li>
+                    <li>List item 6</li>
+                    <li>List item 7</li>
+                    <li>List item 8</li>
+                    <li>
+                        <p>List item 9</p>
+                        <p>Over multiple lines</p>
+                    </li>
+                    <li>List item 10</li>
+                    <li>
+                        <p>List item 11</p>
+                        <p>Over multiple lines</p>
                     </li>
                 </ul>");
 
--- a/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/__snapshots__/HtmlToTextUtilsTests.HtmlToText_OrderedList_OverTenItemsWithMultiline.snap
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/__snapshots__/HtmlToTextUtilsTests.HtmlToText_OrderedList_OverTenItemsWithMultiline.snap
@@ -1,0 +1,15 @@
+﻿1. List item 1
+2. List item 2
+3. List item 3
+4. List item 4
+5. List item 5
+6. List item 6
+7. List item 7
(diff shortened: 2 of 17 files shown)


src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/HtmlToTextUtilsTests.cs
```diff
@@ -215,6 +215,34 @@ public async Task HtmlToText_OrderedList_HasLineAfter()
...
+                    <li>List item 6</li>
+                    <li>List item 7</li>
+                    <li>List item 8</li>
+                    <li>
+                        <p>List item 9</p>
+                        <p>Over multiple lines</p>
+                    </li>
+                    <li>List item 10</li>
+                    <li>
+                        <p>List item 10</p>
```
duncan-at-hiveit, line 238: Just a little typo here - should be 11 :)

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/__snapshots__/HtmlToTextUtilsTests.HtmlToText_OrderedList_OverTenItemsWithMultiline.snap
```diff
@@ -0,0 +1,15 @@
...
+4. List item 4
+5. List item 5
+6. List item 6
+7. List item 7
+8. List item 8
+9. List item 9
+
+   Over multiple lines
+10. List item 10
+11. List item 10
```
duncan-at-hiveit, line 13: Just a little typo here - should be 11 :)

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/Html/HtmlToTextConverter.cs
```diff
@@ -181,17 +181,21 @@ private void ParseListElement(IElement element)
...
             var listItems = element.ChildNodes
-                .Where(node => node is IElement { LocalName: "li" });
+                .Where(node => node is IElement { LocalName: "li" })
+                .ToList();
 
             ParseNodes(
                 listItems,
                 (item, index) =>
                 {
                     _builder.Append(isOrdered ? $"{index + 1}. " : "- ");
```
duncan-at-hiveit, line 194: Maybe using this string to determine the size of `indent` rather than having to calculate it based off `baseIndent` and `(index + 1).ToString().Length`.  Seeing as you already have this string constructed I guess you could just do something like:

```
var lineItem = isOrdered ? $"{index + 1}. " : "- ";
_builder.Append(lineItem);
var indent = lineItem.Length;
```

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/DataGuidanceFileWriter.cs
```diff
@@ -175,6 +175,43 @@ private static async Task WriteDataFiles(IList<MetaGuidanceSubjectViewModel> sub
...
+                        await file.WriteLineAsync();
+
+                        await footnotes
+                            .ForEachAsync(
+                                async (footnote, footnoteIndex) =>
+                                {
+                                    await file.WriteAsync($"{footnoteIndex + 1}. ");
+
+                                    var indexWidth = (footnoteIndex + 1).ToString().Length;
+                                    var indent = string.Empty.PadRight(indexWidth + 2);
```
duncan-at-hiveit, line 195: Same comment as per the `baseIndent` comment added previously here - seeing as you've already constructed the line item header value `$"{footnoteIndex + 1}. "` above, can't you use that same string to get the indent width rather than calculating it again?

src/GovUk.Education.ExploreEducationStatistics.Publisher/Startup.cs
```diff
@@ -107,8 +107,10 @@ public override void Configure(IFunctionsHostBuilder builder)
                 .AddScoped<IReleaseSubjectRepository, ReleaseSubjectRepository>(provider =>
                     new ReleaseSubjectRepository(
                         statisticsDbContext: provider.GetService<PublicStatisticsDbContext>(),
-                        footnoteRepository: provider.GetService<IFootnoteRepository>()))
+                        footnoteRepository: new FootnoteRepository(provider.GetService<PublicStatisticsDbContext>())
```
duncan-at-hiveit, line 110: I guess this is where you started to get frustrated with the public ad private db contexts then :)

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/MetaGuidanceSubjectViewModel.cs
```diff
@@ -8,11 +8,11 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Services.ViewModels
     public class MetaGuidanceSubjectViewModel
     {
         public Guid Id { get; set; }
-        public string Filename { get; set; }
-        public string Name { get; set; }
-        public string Content { get; set; }
-        public TimePeriodLabels TimePeriods { get; set; }
-        public List<string> GeographicLevels { get; set; }
-        public List<LabelValue> Variables { get; set; }
+        public string Filename { get; set; } = string.Empty;
```
duncan-at-hiveit, line 12: `string.Empty` is new to me - good to know

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/HtmlToTextUtilsTests.cs
ntsim: Good catch!

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/Html/HtmlToTextConverter.cs
ntsim: Good suggestion, I've changed it to this!

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/DataGuidanceFileWriter.cs
ntsim: Good suggestion, I've changed it to this!

src/GovUk.Education.ExploreEducationStatistics.Publisher/Startup.cs
ntsim: Indeed :smile:, but already been feeling that pain with the ticket for this. It's way too error prone!

duncan-at-hiveit (approved):

PR #2720, opened 2021-07-22
EES-2345 only show one date in table tool when start and end are the same

amyb-hiveit:


src/explore-education-statistics-common/src/modules/table-tool/components/TimePeriodForm.tsx
```diff
@@ -77,6 +77,16 @@ const TimePeriodForm = (props: Props & InjectedWizardProps) => {
...
   };
 
+  const getDisplayTimePeriod = (
+    startValue: string,
+    endValue: string,
+  ): string => {
+    if (startValue === endValue) {
+      return getOptionLabel(startValue);
+    }
+    return `${getOptionLabel(startValue)} to ${getOptionLabel(endValue)}`;
```
benoutram, line 87: I think this change might break the following UI test so worth double checking it 🙂 
`publish_release_and_amend_2.robot:192`
`user checks previous table tool step contains    4    Time period    2012/13 to 2012/13`

amyb-hiveit: Good spot, I checked the table tool ones but missed this. Fixed now.

ntsim (approved):

PR #2724, opened 2021-07-23
EES-2568 Rename `Data.Model.Services` classes to *Repository 

ntsim:
This PR:

- Renames all `Data.Model.Services` classes to suffix them with `Repository` instead of `Service`
- Renames the entire `Data.Model.Services` namespace to `Data.Model.Repository`

The motivation for this change is that these names are more consistent with what these classes are actually concerned with i.e. interacting with the database.


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
```diff
@@ -332,9 +332,9 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
 
             var dataBlockService = new Mock<IDataBlockService>(MockBehavior.Strict);
             var dataImportService = new Mock<IDataImportService>(MockBehavior.Strict);
-            var subjectService = new Mock<ISubjectService>(MockBehavior.Strict);
+            var subjectService = new Mock<ISubjectRepository>(MockBehavior.Strict);
```
benoutram, line 335: There's some cases throughout this class where the variables e.g. `subjectService` and `releaseSubjectService` haven't be renamed. I think it's been done consistently elsewhere in this PR so I think we should rename them here too?

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
ntsim: Good point! I looked through all of the other usages (or at least most of them) and renamed all the variables I could find. I'm sure people will be diligent and change any residual instances they encounter in the future :smile:

PR #2238, opened 2020-12-17
Ees 1719 add cancel button to bau uploads page 2

duncan-at-hiveit:
This PR:

- adds a "Cancel" button against ongoing data file import on the "Data files" tab of Releases (available currently to BAU Users only, and only during the "normal" import phases for a data file)
- links the presence of the "Cancel" button to permissions in the back end, relating it to user's Role and the current Import Status of the data file

This PR does NOT yet:

- add the ability for analysts to cancel files - once this is confirmed desired behaviour we can put it in, but the scope of EES-1553 is specifically for BAU Users only
- handle the visibility of the "Delete files" and "Replace files" buttons via permission rules from the back end, as the Cancel link is doing. It would be nice to accommodate these within the DataFilePermissions class so as to avoid duplicating permission rules in the front end and the back end.  In fact, the front end is cleverer than the back end in deciding when files can be deleted or replaced, so it would be nice to move that additional logic to the back end permission checks too.

PR #2728, opened 2021-07-26
EES-1757 try to prevent cancel upload button spamming

amyb-hiveit:
I haven't been able to reproduce this locally. The ticket was raised on the day the PR to add the button (https://github.com/dfe-analytical-services/explore-education-statistics/pull/2238) was merged so may well still be valid.

I've moved the code to set the state of the cancel button before it waits for the cancel import to start to try to make sure it always get set as soon as the modal confirmation is clicked.

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/data/components/ReleaseDataUploadsSection.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/data/components/ReleaseDataUploadsSection.tsx
@@ -420,13 +420,12 @@ const ReleaseDataUploadsSection = ({
           onCancel={() => setCancelDataFile(undefined)}
           onConfirm={async () => {
             try {
+              setCancelDataFile(undefined);
+              setFileCancelling(cancelDataFile, true);
               await releaseDataFileService.cancelImport(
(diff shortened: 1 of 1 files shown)


src/explore-education-statistics-admin/src/pages/release/data/components/ReleaseDataUploadsSection.tsx
```diff
@@ -420,13 +420,12 @@ const ReleaseDataUploadsSection = ({
           onCancel={() => setCancelDataFile(undefined)}
           onConfirm={async () => {
             try {
+              setCancelDataFile(undefined);
```
ntsim, line 423: Thinking it'd be a good idea to either:

- Move this to after the try/catch so that we always close the modal
- Show some error message if the cancellation fails

I'm kinda inclined to think we should at least show a message if the cancellation request failed for whatever reason.

We can leave the `setFileCancelling` up here though :slightly_smiling_face:

amyb-hiveit:
Closed in favour of https://github.com/dfe-analytical-services/explore-education-statistics/pull/2730

PR #2732, opened 2021-07-26
EES-2519 - order of deletion of Releases and Methodologies

duncan-at-hiveit:
This PR:
- Adds an order in which Releases and Methodologies are deleted under a Theme when the Theme and its Topics are deleted.  This is to prevent foreign key issues coming from Previous Version Id columns which reference entities of the same type.  This way, the latest version of Releases and Methodologies are deleted first, then their previous versions, then their previous etc etc, until finally the original versions are deleted.
- Adds the removal of Methodology Files and orphaned MethodologyParents from Blob Storage when deleting a Theme and its Topics, by way of using MethodologyService to delete Methodologies cleanly (with an extra "forceDelete" flag to override the default security checks).
- Adds per-environment configuration to allow or disallow the deleting of whole Themes.  By default, this is possible on all environments save for Production.
- Adds additional testing around the previously existing TopicService.Delete() and ThemeService.Delete() methods.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ThemeService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ThemeService.cs
@@ -7,6 +7,7 @@ using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.Security;
 using GovUk.Education.ExploreEducationStatistics.Admin.Validators;
 using GovUk.Education.ExploreEducationStatistics.Admin.ViewModels;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Model;
 using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces.Security;
 using GovUk.Education.ExploreEducationStatistics.Common.Utils;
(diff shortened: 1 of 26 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
```diff
@@ -400,7 +400,7 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
                 });
             }
 
-            if(env.IsDevelopment() || Configuration.GetValue<bool>("enableSwagger"))
```
markhiveit, line 403: I'm not 100%, but I think `env.IsDevelopment()` allows Swagger to work when you're running Admin locally.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ThemeService.cs
```diff
@@ -139,31 +144,36 @@ public class ThemeService : IThemeService
...
-                        // For now we only want to delete test themes as we
-                        // don't really have a mechanism to clean things up
-                        // properly across the entire application.
-                        // TODO: EES-1295 ability to completely delete releases
-                        if (!theme.Title.StartsWith("UI test theme"))
-                        {
-                            return;
-                        }
+                .OnSuccessDo(CheckCanDeleteTheme)
+                .OnSuccessVoid(async theme =>
```
markhiveit, line 148: Just for my own curiosity, I think in other places we do stuff like this `CheckCanDeleteTheme` would return a theme and would be contained in a `OnSuccess`. How come the `OnSuccessDo` here?

markhiveit (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
duncan-at-hiveit: Yup - but it's also set to true in appsettings.Development.json, so I decided this was a redundant check

markhiveit: I was sure Swagger wouldn't work locally without this, but tested it and it works fine.

This `env.IsDevelopment()` could also be removed from the same line for Content and Data too.

markhiveit (approved):

PR #2734, opened 2021-07-27
EES-1350 Add ability to resolve comments on Admin Content page

markhiveit:


src/GovUk.Education.ExploreEducationStatistics.Admin/Mappings/MappingProfiles.cs
```diff
@@ -205,7 +205,12 @@ public MappingProfiles()
 #pragma warning restore 612
                                 LastName = ""
                             }
-                            : comment.CreatedBy));
+                            : comment.CreatedBy))
+                .ForMember(dest => dest.ResolvedBy,
```
duncan-at-hiveit, line 209: I don't *think* you need this mapping.  

If `comment.ResolvedById` is null, then so will `comment.ResolvedBy` be.

If `comment.ResolvedById` is not null, then so will `comment.ResolvedBy` hopefully be (or it hasn't been hydrated, but I guess we're hoping that's not the case when using this mapper).

Either way, I think we can just let ResolvedBy get mapped straight over intot he same named field in the view model without the need for this.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ContentServiceTests.cs
```diff
@@ -169,17 +172,317 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+                        ContentSection = new ContentSection
+                        {
+                            Heading = "New section",
+                            Order = 1,
+                            Content = new List<ContentBlock>
+                            {
+                                new HtmlBlock
+                                {
+                                    Created = new DateTime(2001, 1, 1),
+                                    Comments = new List<Comment>
```
duncan-at-hiveit, line 193: We also have an `AsList` method now in CollectionUtils that can make test code a bit more succinct when statically imported, like:

```
Comments = AsList(new Comment {Content = "Existing comment"})
```

rather than

```
Comments = new List<Comment>
{
    new Comment {Content = "Existing comment"}
}
```

Just so you're aware for the future.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ContentServiceTests.cs
```diff
@@ -169,17 +172,317 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+
+        [Fact]
+        public async Task ResolveComment_True()
+        {
+            var comment = new Comment
+            {
+                Content = "Existing comment",
+                CreatedById = Guid.NewGuid(),
+            };
+            var release = new Release
```
duncan-at-hiveit, line 240: Another example of AsList.

```
    var release = new Release
    {
        Content = new List<ReleaseContentSection>
        {
            new ReleaseContentSection
            {
                ContentSection = new ContentSection
                {
                    Content = new List<ContentBlock>
                    {
                        new HtmlBlock
                        {
                            Comments = new List<Comment>
                            {
                                comment
                            }
                        }
                    }
                }
            }
        }
    };
```

can become:

```
    var release = new Release
    {
        Content = AsList(new ReleaseContentSection
        {
            ContentSection = new ContentSection
            {
                Content = AsList(new HtmlBlock
                {
                    Comments = AsList(comment)
                }
            }
        }
    };
```

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ContentServiceTests.cs
```diff
@@ -169,17 +172,317 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+
+                var comments = contentDbContext.Comment.ToList();
+                Assert.Equal(2, comments.Count);
+                Assert.Equal("Existing comment", comments[0].Content);
+                Assert.Equal("New comment", comments[1].Content);
+            }
+        }
+
+        [Fact]
+        public async Task ResolveComment_True()
```
duncan-at-hiveit, line 233: Might suggest that the method name `ResolveComment` might better be named `SetResolved` or similar, given that it can be used to both resolve and *un* resolve a comment.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ContentServiceTests.cs
```diff
@@ -169,17 +172,317 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+            {
+                var userId = Guid.NewGuid();
+                var service = SetupContentService(
+                    contentDbContext: contentDbContext,
+                    userId: userId);
+                var result = await service.ResolveComment(comment.Id, true);
+
+                var viewModel = result.AssertRight();
+                Assert.Equal("Existing comment", viewModel.Content);
+                Assert.NotNull(viewModel.Resolved);
```
duncan-at-hiveit, line 280: Might suggest you check dates with `InRange`, as per:

```
Assert.InRange(DateTime.UtcNow.Subtract(viewModel.Resolved).Milliseconds, 0, 1500);
```

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ContentServiceTests.cs
```diff
@@ -169,17 +172,317 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+                    {
+                        ContentSection = new ContentSection
+                        {
+                            Content = new List<ContentBlock>
+                            {
+                                new HtmlBlock
+                                {
+                                    Comments = new List<Comment>
+                                    {
+                                        comment
```
duncan-at-hiveit, line 312: Probably good for this test to start off with a Resolved date and ResolvedBy and ResolvedById values, to prove that the method will unset them.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ContentServiceTests.cs
```diff
@@ -169,17 +172,317 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+                var viewModel = result.AssertRight();
+                Assert.Equal("Existing comment updated", viewModel.Content);
+                Assert.Null(viewModel.Resolved);
+                Assert.Null(viewModel.ResolvedBy);
+
+                var comments = contentDbContext.Comment.ToList();
+                Assert.Single(comments);
+                Assert.Equal("Existing comment updated", comments[0].Content);
+                Assert.Equal(comment.CreatedById, comments[0].CreatedById);
+                Assert.Null(comments[0].ResolvedById);
```
duncan-at-hiveit, line 405: Might as well test the Comment.Updated date value here as well :)

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ManageContent/ContentController.cs
```diff
@@ -151,24 +151,33 @@ public async Task<ActionResult<List<DataBlock>>> GetAvailableDataBlocks(Guid rel
...
                 .HandleFailuresOrOk();
         }
 
         [HttpPut("comment/{commentId}")]
-        public async Task<ActionResult<CommentViewModel>> UpdateComment(Guid commentId, CommentSaveRequest saveRequest)
+        public async Task<ActionResult<CommentViewModel>> UpdateComment(Guid commentId, 
+            CommentSaveRequest saveRequest,
+            [FromQuery(Name = "resolved")] bool? resolved = null)
         {
+            if (resolved.HasValue)
```
duncan-at-hiveit, line 163: Is it possible to update other values whilst resolving?  Not sure what the front end looks like

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/ResolveSpecificCommentAuthorizationHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.EntityFrameworkCore;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Security.AuthorizationHandlers
+{
+    public class ResolveSpecificCommentRequirement : IAuthorizationRequirement
+    {
+    }
+
+    public class
```
duncan-at-hiveit, line 15: Maybe amend the formatting here from:

```
public class
        ResolveSpecificCommentAuthorizationHandler : AuthorizationHandler<ResolveSpecificCommentRequirement, Comment>
```

to

```
public class ResolveSpecificCommentAuthorizationHandler 
    : AuthorizationHandler<ResolveSpecificCommentRequirement, Comment>
```

Is that any better?  Meh...

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/ContentService.cs
```diff
@@ -598,12 +613,16 @@ private List<ContentSectionViewModel> OrderedContentSections(Release release)
...
-                    .ThenInclude(section => section.Content)
-                    .ThenInclude(content => content.Comments)
-                    .ThenInclude(comment => comment.CreatedBy)
-                ;
+                .Include(r => r.Content)
+                .ThenInclude(join => join.ContentSection)
+                .ThenInclude(section => section.Content)
+                .ThenInclude(content => content.Comments)
+                .ThenInclude(comment => comment.CreatedBy)
+                .Include(r => r.Content)
```
duncan-at-hiveit, line 621: Hehe it's annoying they make you have to muck around to include more than one field within a ThenInclude chain.  Don't think there's any way around this currently.  Some people indent their ThenIncludes to make it more obvious as to what's going on e.g.:

```
return releases
                .Include(r => r.Content)
                      .ThenInclude(join => join.ContentSection)
                      .ThenInclude(section => section.Content)
                      .ThenInclude(content => content.Comments)
                      .ThenInclude(comment => comment.CreatedBy)
                .Include(r => r.Content)
                      .ThenInclude(join => join.ContentSection)
                      .ThenInclude(section => section.Content)
                      .ThenInclude(content => content.Comments)
                      .ThenInclude(comment => comment.ResolvedBy);
```

but personal preference here I think.  I don't have any feelings either way, but I was at first confused as to why there were 2 Include chains going on.  Then the penny dropped!

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/CommentViewModel.cs
```diff
@@ -10,5 +10,7 @@ public class CommentViewModel
         public DateTime Created { get; set; }
         public User CreatedBy { get; set; }
         public DateTime? Updated { get; set; }
+        public DateTime? Resolved { get; set; }
+        public User ResolvedBy { get; set; }
```
duncan-at-hiveit, line 14: At some point we should switch to using UserViewModel to avoid accidentally leaking any info about users to the frontend

src/explore-education-statistics-admin/src/components/editable/Comments.tsx
```diff
@@ -97,14 +97,25 @@ const Comments = ({
       });
   };
 
+  const resolveComment = (commentId: string, resolved: boolean) => {
```
duncan-at-hiveit, line 100: Again, maybe rename to `setResolved` or `setCommentResolution` as this method can both resolve and unresolve comments

src/explore-education-statistics-admin/src/components/editable/Comments.tsx
```diff
@@ -97,14 +97,25 @@ const Comments = ({
...
+    const index = comments.findIndex(comment => comment.id === commentId);
+    releaseContentCommentService
```
duncan-at-hiveit, line 102: Why not make this function async i.e. `const resolveComment = async (.....`

Then you can do:

```
await releaseContentCommentService.updateContentSectionComment(comments[index], resolved);

const newComments = [...comments];
newComments[index] = savedComment;
onChange(blockId, newComments);
```

duncan-at-hiveit: Not a massive difference to be fair!

src/explore-education-statistics-admin/src/components/editable/Comments.tsx
```diff
@@ -186,24 +249,34 @@ const Comments = ({
...
+                              onClick={() => {
+                                removeComment(comment.id);
+                              }}
+                            >
+                              Delete
+                            </ButtonText>
+                          </>
+                        )}
 
                         <ButtonText
```
duncan-at-hiveit, line 274: Do we not need to show or hide this button based on permissions?

src/explore-education-statistics-admin/src/components/editable/Comments.tsx
```diff
@@ -133,21 +144,73 @@ const Comments = ({
...
+                    </p>
+                  )}
+                  <p className="govuk-!-margin-top-3">{comment.content}</p>
+                </Details>
+                <ButtonText
+                  onClick={() => {
+                    resolveComment(comment.id, false);
+                  }}
+                >
+                  Unresolve
```
duncan-at-hiveit, line 189: Do we not need to show and hide this button based on permissions?

src/explore-education-statistics-admin/src/components/editable/__tests__/Comments.test.tsx
```diff
@@ -69,11 +69,12 @@ describe('Comments', () => {
     expect(comments).toHaveLength(2);
 
     expect(comments[0]).toHaveTextContent('John Smith');
-    expect(comments[0]).toHaveTextContent('6 June 2020');
+    expect(comments[0]).toHaveTextContent('Created: 06/06/20');
```
duncan-at-hiveit, line 72: Might as well add the expected time to this check as well e.g. `expect(comments[0]).toHaveTextContent('Created: 06/06/20 12:00');`

src/explore-education-statistics-admin/src/components/editable/__tests__/Comments.test.tsx
```diff
@@ -69,11 +69,12 @@ describe('Comments', () => {
...
     expect(comments[0]).toHaveTextContent('Test comment 1');
 
     expect(comments[1]).toHaveTextContent('Jane Roberts');
-    expect(comments[1]).toHaveTextContent('Updated: 7 June 2020');
+    expect(comments[1]).toHaveTextContent('Created: 06/06/20');
```
duncan-at-hiveit, line 76: Same as above

src/explore-education-statistics-admin/src/components/editable/Comments.tsx
```diff
@@ -97,14 +97,25 @@ const Comments = ({
       });
   };
 
+  const resolveComment = (commentId: string, resolved: boolean) => {
+    const index = comments.findIndex(comment => comment.id === commentId);
+    releaseContentCommentService
+      .updateContentSectionComment(comments[index], resolved)
```
duncan-at-hiveit, line 103: Instead of mixing this in with the existing `updateContentSectionComment` method, why not spin off a separate `setCommentResolution` service method?  Then `updateContentSectionComment` doesn't need the `resolved: boolean | null` property.

Looking at the code paths, updating any other details on the comment and setting the Resolved value are mutually exclusive anyway.

src/explore-education-statistics-admin/src/services/types/content.ts
```diff
@@ -11,6 +11,13 @@ export interface Comment {
   };
   created: string;
   updated?: string;
+  resolved?: string;
+  resolvedBy?: {
```
duncan-at-hiveit, line 15: As we don't use the "id" or "email" field, you don't need to include them here.  We should I guess not really be returning them from the backend ideally if they're not going to be used, as they leak potentially sensitive information (see my other comment about switching to returning UserViewModels)

src/GovUk.Education.ExploreEducationStatistics.Admin/Mappings/MappingProfiles.cs
markhiveit: I removed the mapping and gave it a quick test, and it seems you're right 👍

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ContentServiceTests.cs
markhiveit: Cool - I've switched all instances of `new List` in ContentServiceTests to use `AsList`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/ContentService.cs
```diff
@@ -598,12 +613,16 @@ private List<ContentSectionViewModel> OrderedContentSections(Release release)
...
+                .ThenInclude(join => join.ContentSection)
+                .ThenInclude(section => section.Content)
+                .ThenInclude(content => content.Comments)
+                .ThenInclude(comment => comment.ResolvedBy);
```
ntsim, line 625: Eesh, you'd think EF would have a better way of doing this :grimacing:

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ContentServiceTests.cs
```diff
@@ -169,17 +172,317 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+                Assert.Null(viewModel.ResolvedBy);
+
+                var comments = contentDbContext.Comment.ToList();
+                Assert.Single(comments);
+                Assert.Equal("Existing comment", comments[0].Content);
+                Assert.Equal(comment.CreatedById, comments[0].CreatedById);
+                Assert.Null(comments[0].ResolvedById);
+            }
+        }
+
```
ntsim, line 330: Would suggest adding a test to check that the comment exists

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ContentServiceTests.cs
ntsim, line 225: Might be better to explicitly name this `ResolveComment_Resolve`?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ContentServiceTests.cs
```diff
@@ -169,17 +172,317 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+                var comments = contentDbContext.Comment.ToList();
+                Assert.Single(comments);
+                Assert.Equal("Existing comment", comments[0].Content);
+                Assert.Equal(comment.CreatedById, comments[0].CreatedById);
+                Assert.Equal(userId, comments[0].ResolvedById);
+            }
+        }
+
+        [Fact]
+        public async Task ResolveComment_False()
```
ntsim, line 278: Might be better to explicitly name this `ResolveComment_Unresolve`?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ContentServiceTests.cs
```diff
@@ -169,17 +172,317 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+                await contentDbContext.SaveChangesAsync();
+            }
+
+            await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContextId))
+            {
+                var userId = Guid.NewGuid();
+                var service = SetupContentService(
+                    contentDbContext: contentDbContext,
+                    userId: userId);
+                var result = await service.ResolveComment(comment.Id, true);
```
ntsim, line 263: Could make it a little clearer what's going on by using the named parameter i.e. `resolve: true`

Raw boolean arguments usually suck because they tell you very little about what they're doing

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ContentServiceTests.cs
```diff
@@ -169,17 +172,317 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+                await contentDbContext.SaveChangesAsync();
+            }
+
+            await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContextId))
+            {
+                var userId = Guid.NewGuid();
+                var service = SetupContentService(
+                    contentDbContext: contentDbContext,
+                    userId: userId);
+                var result = await service.ResolveComment(comment.Id, false);
```
ntsim, line 316: Could use the named parameter here to make this a little clearer

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/ResolveSpecificCommentAuthorizationHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+        ResolveSpecificCommentAuthorizationHandler : AuthorizationHandler<ResolveSpecificCommentRequirement, Comment>
```
ntsim, line 16: This looks a bit weird, could reformat it to:

```csharp
public class ResolveSpecificCommentAuthorizationHandler :
    AuthorizationHandler<ResolveSpecificCommentRequirement, Comment>
```

src/explore-education-statistics-admin/src/components/editable/Comments.tsx
markhiveit: Seems I've incorrectly assumed this was wrapped `canComment`, but it is not. But I tested this by creating a user that only had viewing permission, and they cannot see comments at all. And all users that can edit a release should also have permission to resolve comments. So it doesn't appear to be an issue, but I'll make the Unresolve button conditionally appear based on `canComment` such that it doesn't accidentally get revealed if permissions change in the future.

src/explore-education-statistics-admin/src/components/editable/Comments.tsx
```diff
@@ -133,21 +144,73 @@ const Comments = ({
...
+            const { createdBy, resolvedBy } = comment;
 
-            return (
+            return comment.resolved ? (
+              <li className="govuk-body-s">
+                <p className="govuk-!-margin-0">
+                  <strong>Comment resolved</strong>
+                </p>
+                <p className="govuk-!-margin-0">
+                  <FormattedDate format="dd/MM/yy HH:mm">
```
ntsim, line 155: Not a big fan of this date format as it's bit too dense on the eyes and is also ambiguous. Would suggest we use the old date format, but with shorter months: `d MMM yyyy HH:mm`.

Same applies for the other date/times here

src/explore-education-statistics-admin/src/components/editable/Comments.tsx
```diff
@@ -133,21 +144,75 @@ const Comments = ({
...
+                    </FormattedDate>
+                  </p>
+                  {comment.updated && (
+                    <p className="govuk-!-margin-0">
+                      {'Updated: '}
+                      <FormattedDate format="dd/MM/yy HH:mm">
+                        {comment.updated}
+                      </FormattedDate>
+                    </p>
+                  )}
```
ntsim, line 181: This is a little chaotic tbh. Think we could refactor a lot of this to use `dl/dt/dd` elements.

The markup would be more semantically correct and would read nicer. We'll probably need some extra styling in `Comments.module.scss`, but we can do away with all the `strong` elements and `govuk-!-margin-0`.

src/explore-education-statistics-admin/src/components/editable/Comments.tsx
```diff
@@ -133,21 +144,75 @@ const Comments = ({
...
+                    {comment.created}
                   </FormattedDate>
                 </p>
+                {comment.updated && (
+                  <p className="govuk-!-margin-0">
+                    {'Updated: '}
+                    <FormattedDate format="dd/MM/yy HH:mm">
+                      {comment.updated}
+                    </FormattedDate>
+                  </p>
```
ntsim, line 214: Similar to above - could refactor this to use `dl/dt/dd` elements

src/explore-education-statistics-admin/src/components/editable/Comments.tsx
```diff
@@ -186,24 +251,34 @@ const Comments = ({
...
+                        )}
 
                         <ButtonText
                           onClick={() => {
-                            removeComment(comment.id);
+                            resolveComment(comment.id, true);
                           }}
                         >
-                          Delete
+                          Resolve
```
ntsim, line 281: We could use some test cases for resolving/unresolving comments

src/explore-education-statistics-admin/src/services/releaseContentCommentService.ts
```diff
@@ -27,8 +27,15 @@ const releaseContentCommentService = {
...
-  updateContentSectionComment(comment: UpdateComment): Promise<Comment> {
-    return client.put(`/comment/${comment.id}`, comment);
+  updateContentSectionComment(
+    comment: UpdateComment,
+    resolved: boolean | null = null,
+  ): Promise<Comment> {
+    return client.put(
+      `/comment/${comment.id}`,
+      comment,
+      resolved != null ? { params: { resolved } } : {},
```
ntsim, line 37: Don't use query params. `resolved` can be part of the request body.

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ManageContent/ContentController.cs
markhiveit: You can't. The resolve button appears even if you aren't currently editing the comment.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/ContentService.cs
markhiveit: I was considering doing the whole switch to `_context.Entry(blah).References(b => b.Whatever)` to get around this, but it would have required a bit of work so I just left it at this. I'll indent the `ThenInclude`s

src/explore-education-statistics-admin/src/components/editable/Comments.tsx
markhiveit: The reason I went for this is because there is limited width for text in the comments - the longer date was wrapping which I think is more confusing than the shorter date. I was even considering getting rid of the time, but decided against that.

I'll run both by Laura and go with what she wants.

markhiveit: Laura says she prefers the first so I'm going to leave this.

ntsim: Cool, fair enough!

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/ResolveSpecificCommentAuthorizationHandler.cs
```diff
@@ -0,0 +1,63 @@
+﻿using System.Linq;
```
ntsim, line 1: Oh another thing - make sure to add `#nullable enable` to files you're touching (unless there is a good reason not to e.g. too many warnings or it's a database entity)

src/explore-education-statistics-admin/src/components/editable/Comments.tsx
markhiveit: Anyone who can edit the release should be able to resolve comments. So I believe this is correct as it is.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/CommentViewModel.cs
markhiveit: As discussed on standup, I'm leaving this as it is for now until we decide whether to split the resolved functionality into a separate endpoint.

src/explore-education-statistics-admin/src/components/editable/Comments.tsx
markhiveit: As discussed on standup, I'm leaving this as it is for now until we decide whether to split the resolved functionality into a separate endpoint.

src/explore-education-statistics-admin/src/services/types/content.ts
markhiveit: As discussed on standup, I'm leaving this as it is for now until we decide whether to split the resolved functionality into a separate endpoint.

ntsim (approved):

PR #2735, opened 2021-07-28
EES-2370 Add footnotes to meta data guidance document page

ntsim:
This PR adds footnotes to the (meta) data guidance document page. Whilst this change was fairly simple, we've also made a bunch of changes to the Robot tests.

## Test changes

- Combined the `publish_data` and `footnotes` suites into a single suite. This is now composed of significantly more tests around footnotes and includes testing of variables names/description and footnotes across the data guidance pages.
- Moved custom locators into `utilities.py` to enable them to work correctly with parent/child keywords written in Python. 
  Previously, these wouldn't work correctly as they wouldn't have access to the custom locator strategies. This was due to not using the framework (or global) instance of `ElementFinder`.
- Renamed keywords around `user checks list ...`  to the less confusing `user checks select ...` (as in `select` elements). This makes room for new keywords around checking actual list elements (as in `ul` / `ol`) and their associated items.


src/explore-education-statistics-common/src/services/releaseMetaGuidanceService.ts
```diff
@@ -15,6 +15,10 @@ export interface SubjectMetaGuidance {
     label: string;
     value: string;
   }[];
+  footnotes: {
```
duncan-at-hiveit, line 18: I know we've got various id / text pair types stashed around that you might or might not wanna make use of e.g. `IdLabelPair`.

tests/robot-tests/tests/libs/common.robot
```diff
@@ -502,7 +480,33 @@ user checks summary list contains
     ...    xpath:.//dl//dt[contains(text(), "${term}")]/following-sibling::dd[contains(., "${description}")]
     user waits until element is visible    ${element}    %{WAIT_MEDIUM}
 
-user selects from list by label
+user checks select contains x options
```
duncan-at-hiveit, line 483: Nice additions :)

tests/robot-tests/tests/libs/utilities.py
```diff
@@ -5,29 +5,64 @@
...
+import utilities_init
 import os
 import re
 
 sl = BuiltIn().get_library_instance('SeleniumLibrary')
-element_finder = ElementFinder(sl)
+element_finder = sl._element_finder
 waiting = WaitingKeywords(sl)
 
+# Should only initialise some parts once e.g. registration
```
duncan-at-hiveit, line 19: Wow, good work!

src/explore-education-statistics-admin/src/pages/release/footnotes/components/IndicatorDetails.tsx
```diff
@@ -50,7 +50,7 @@ const IndicatorDetails = ({
                   className="govuk-checkboxes--small"
                   id={`indicatorGroup-${indicatorGroupId}`}
                   name={`${valuePath}.indicatorGroups[${indicatorGroupId}].selected`}
-                  label={`${indicatorGroup.label} ${groupValue && '(All)' : ''}`}
+                  label={`${indicatorGroup.label}${groupValue && ' (All)' : ''}`}
```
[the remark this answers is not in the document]
ntsim: No idea :smile: Stu commented it out a while back. 

Tbh, I think a lot of the Footnotes code is in need of a refactor/cleanup, but just waiting for the right moment!

src/explore-education-statistics-common/src/services/releaseMetaGuidanceService.ts
ntsim: Probably gonna keep it as is in case we need to change it in the future. Tbh, `IdLabelPair` is such a simple type that it's not really gaining us anything by using it.

src/explore-education-statistics-common/src/modules/release/components/ReleaseMetaGuidanceDataFile.tsx
```diff
@@ -64,26 +64,43 @@ const ReleaseMetaGuidanceDataFile = ({ subject, renderContent }: Props) => {
...
+          <table data-testid="Variables">
+            <thead>
+              <tr>
+                <th scope="col">Variable name</th>
+                <th scope="col">Variable description</th>
               </tr>
+            </thead>
+            <tbody>
+              {variables.map(({ value, label }) => (
+                <tr key={value}>
```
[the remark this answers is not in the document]
ntsim: Yeh it's not guaranteed to be unique as it can be either a filter item or an indicator, so theoretically there could be instances where these are named the same. I'm not entirely sure if the various data file validators along the way check against this?

In this context, it doesn't really matter though as we don't change the ordering. To avoid any potential question marks around duplicate key warnings in the console, I'll swap it to use the index.

duncan-at-hiveit (approved):

PR #2736, opened 2021-07-28
EES-2481-2 - fix various UI test failures

luke-h1:


tests/robot-tests/tests/general_public/methodology_page_absence.robot
```diff
@@ -8,7 +8,7 @@ Force Tags          GeneralPublic    Local    Dev    Test    Preprod
 
 *** Test Cases ***
 Navigate to Pupil absence in schools in England methodology page
-    [Tags]    HappyPath
+    [Tags]    HappyPath  Failing  NotAgainstDev
```
markhiveit, line 11: Rather than adding these tags to every test case, you could add them to the `Force Tags` list at the top once.

Also, `Failing` stops the tests running against all environment, so the `NotAgainstDev` tag is redundant while the `Failing` tag is applied. `NotAgainstDev` is meant to be applied in situations where we don't want the test case to be run against the dev environment specifically.

tests/robot-tests/tests/general_public/glossary_page.robot
```diff
@@ -29,16 +29,18 @@ Search for Pupil referral unit
     [Tags]    HappyPath
     user verifies accordion is closed    P
 
-    user enters text into element    id:pageSearchForm-input    Pupil referral unit
+    user enters text into element    id:pageSearchForm-input    Permanent exclusion
```
markhiveit, line 32: Interested to know why it was failing with Pupil referral unit, as I don't believe the glossary page has changed, and I don't think we've had an issue with this before.

After clicking the search option I would expect to jump to that entry such that it is top of the page, but right now it seems like it jumps to the middle of the screen. Maybe that has changed?

markhiveit (changes requested):

tests/robot-tests/tests/general_public/glossary_page.robot
luke-h1: the pupil referral unit test data isn't present anymore. Assumed this had been removed as part of the new methodology work

luke-h1: looks like a wrong id was being selected. This has been fixed now

markhiveit (approved):

PR #2737, opened 2021-07-28
EES-2575 Upgrade to C# 9 without .NET 5

ntsim:
This was really easy to do as it just involves adding the [IsExternalInit](https://github.com/manuelroemer/IsExternalInit) package to get it working.

With this, we can get C# 9 features whilst avoiding general migration pain for upgrading to .NET 5. I found the main barrier to upgrading to .NET 5  was that Azure Functions didn't really work correctly. It appears that there will be proper support for .NET 6 in November, so it's worth waiting until then for us to do that upgrade.

All UI tests have been ran and there are no failures to speak of.

## Highlights

C# 9 gives us few really nice new features:

- Records - https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#record-types
- Init only setters - https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#init-only-setters
- Can omit explicit types for `new` declarations - https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#fit-and-finish-features
- Better pattern matching for switch expressions - https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#pattern-matching-enhancements
- Immutable copying with new `with` keyword - https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#nondestructive-mutation

## Recommendations

Based on the above features I would make the following suggestions to our code moving forward:

- Records should be pretty much used by default for all DTO-like types (view models, etc).
- Init setters should be used instead of standard setters, unless you need to be able to mutate a field later (hardly if ever). You can instead use the `with` keyword in most cases.
- Use the omitted type `new` declaration where possible - I don't think this should necessarily be a hard and fast rule though.

## Caveats

Note that **not all** features of C# 9 will work without the latest and great runtime that is included in .NET 5/6 e.g. covariant return types.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/GovUk.Education.ExploreEducationStatistics.Admin.Tests.csproj
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/GovUk.Education.ExploreEducationStatistics.Admin.Tests.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
-
+    <LangVersion>9.0</LangVersion>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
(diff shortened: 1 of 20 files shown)


duncan-at-hiveit:
Nicely documented, thanks!

duncan-at-hiveit (approved):

PR #2740, opened 2021-07-29
EES-XXXX - setting Created and CreatedById when creating new Methodol…

duncan-at-hiveit:
This PR:
- adds the created date and user id to newly-created Methodology versions.

As a follow-up to this, we should perform some cleanup to make the Created and CreatedById columns mandatory.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Repository/MethodologyRepositoryTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Repository/MethodologyRepositoryTests.cs
@@ -17,8 +17,6 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Model.Tests.Reposit
 {
     public class MethodologyRepositoryTests
     {
-        private static readonly Guid UserId = Guid.NewGuid();
-        
         [Fact]
         public async Task CreateMethodologyForPublication()
         {
@@ -27,6 +25,9 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Model.Tests.Reposit
                 Title = "The Publication Title",
                 Slug = "the-publication-slug"
             };
+
+            var userId = Guid.NewGuid();
+            var createdDate = DateTime.UtcNow.AddDays(-2);
 
             var contentDbContextId = Guid.NewGuid().ToString();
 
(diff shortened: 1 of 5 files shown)


luke-h1 (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Repository/MethodologyRepositoryTests.cs
```diff
@@ -26,6 +26,9 @@ public async Task CreateMethodologyForPublication()
                 Slug = "the-publication-slug"
             };
 
+            var userId = Guid.NewGuid();
+            var createdDate = DateTime.UtcNow.AddDays(-2);
```
luke-h1, line 30: out of curiosity why are we passing -2 here?

luke-h1 (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/MethodologyRepository.cs
```diff
@@ -44,7 +48,9 @@ public async Task<Methodology> CreateMethodologyForPublication(Guid publicationI
                             PublicationId = publicationId
                         }
                     }
-                }
+                },
+                Created = createdDate,
```
markhiveit, line 52: Since the createdDate and user are always going to be set to the current time / current user, I'd be inclined to not pass them in as arguments and just set them here - not sure if there is some advantage to passing them in as arguments?

markhiveit (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Repository/MethodologyRepositoryTests.cs
duncan-at-hiveit: Totally arbitrary - just anything that isn't exactly now :)

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/MethodologyRepository.cs
duncan-at-hiveit: It's just personal preference really - ever so slightly easier to test, and a bit more flexible in its use, that's all.  Happy either way

markhiveit (approved):

PR #2745, opened 2021-07-30
Ees 2572 - refactor CLI to use common services

luke-h1:
This PR splits functionality that each module in this CLI uses into service files. This is to avoid code repetition across the cli. It splits up functionality into 3 services: 
- `releaseService` 
- `commonService`
- `subjectService`

### DIFF
--- a/useful-scripts/misc-test-utils/src/services/commonService.ts
+++ b/useful-scripts/misc-test-utils/src/services/commonService.ts
@@ -1,0 +1,85 @@
+/* eslint-disable no-console */
+import chalk from 'chalk';
+import StreamZip from 'node-stream-zip';
+import globby from 'globby';
+import fs from 'fs';
+import rimraf from 'rimraf';
+import { projectRoot } from '../config';
+
+const cwd = projectRoot;
+
+const commonService = {
+  validateArchives: async () => {
+    const zipGlob = await globby(`${cwd}/*.zip`);
+    if (fs.existsSync(zipGlob[0]) && zipGlob[0].includes('archive.zip')) {
+      console.log(chalk.green('found zip file, continuing'));
+    } else {
+      throw new Error(
+        'No zip file named archive.zip in root dir! Exiting test with failures',
+      );
+    }
+    if (zipGlob[1]) {
+      rimraf(zipGlob[1], () => {
+        console.log(chalk.green('cleaned stale zip files'));
+      });
+    }
+  },
+
+  prepareDirectories: async () => {
+    if (fs.existsSync(`${cwd}/test-files`)) {
+      rimraf(`${cwd}/test-files/*`, () => {
+        console.log(chalk.green('cleaned test-files folder'));
+      });
+    }
+
+    if (fs.existsSync(`${cwd}/zip-files`)) {
+      rimraf(`${cwd}/zip-files/*`, () => {
+        console.log(chalk.green('cleaned zip-files folder'));
+      });
+    }
+
+    if (!fs.existsSync(`${cwd}/test-files`)) {
+      fs.mkdirSync(`${cwd}/test-files`);
+    }
+
+    if (!fs.existsSync(`${cwd}/zip-files`)) {
+      fs.mkdirSync(`${cwd}/zip-files`);
+    }
+  },
+  extractZip: async () => {
+    const zippedFile = await globby(`${cwd}/*.zip`);
+    const cleanZipFile = zippedFile[0];
+    const archive = fs.existsSync(zippedFile[0]);
+    if (archive) {
+      // eslint-disable-next-line new-cap
+      const zip = new StreamZip.async({ file: cleanZipFile });
+      const count = await zip.extract(null, `${cwd}/test-files`);
+      console.log(chalk.green(`Extracted ${count} entries to test-files`));
+      await zip.close();
+    }
+  },
+  renameFiles: async () => {
+    // define a random id to assign to the renamed data & meta file...
+    const randomId = Math.floor(Math.random() * 1000000);
+
+    // rename csv file..
+    const csvGlob1 = await globby(`${cwd}/test-files/*.csv`);
+
+    const csvGlob2 = csvGlob1[0];
+
+    fs.rename(`${csvGlob2}`, `${cwd}/test-files/testfile-${randomId}.csv`, e =>
+      e ? console.error(chalk.red(e)) : '',
+    );
+
+    // rename the meta file...
+    const metaGlob = await globby(`${cwd}/test-files/*.meta.csv`);
+    const metaGlob2 = metaGlob[0];
+
+    fs.rename(
+      `${metaGlob2}`,
+      `${cwd}/test-files/testfile-${randomId}.meta.csv`,
+      e => (e ? console.error(chalk.red(e)) : ''),
+    );
+  },
+};
+export default commonService;
--- a/useful-scripts/misc-test-utils/src/services/publicationService.ts
+++ b/useful-scripts/misc-test-utils/src/services/publicationService.ts
@@ -1,0 +1,35 @@
+/* eslint-disable no-console */
+import chalk from 'chalk';
+import { v4 } from 'uuid';
+import errorHandler from '../utils/errorHandler';
+import adminApi from '../utils/adminApi';
+
+const { TOPIC_ID } = process.env;
+
+const publicationService = {
+  // eslint-disable-next-line consistent-return
+  createPublication: async () => {
+    console.time('createPublication');
+    try {
+      const res = await adminApi.post('/api/publications', {
+        title: `importer-testing-${v4()}`,
+        topicId: TOPIC_ID,
+        contact: {
+          teamName: 'testing',
+          teamEmail: 'johndoe@gmail.com',
+          contactName: 'John Doe',
+          contactTelNo: '123456789',
+        },
+      });
+      console.timeEnd('createPublication');
+      console.log(
+        chalk.green(`Created publication. Status code ${res.status}`),
+      );
+      return res.data.id;
+    } catch (e) {
+      errorHandler(e);
+    }
+    return null;
+  },
+};
+export default publicationService;
--- a/useful-scripts/misc-test-utils/src/services/publicationService.ts
+++ b/useful-scripts/misc-test-utils/src/services/publicationService.ts
@@ -1,7 +1,6 @@
 /* eslint-disable no-console */
 import chalk from 'chalk';
 import { v4 } from 'uuid';
-import errorHandler from '../utils/errorHandler';
 import adminApi from '../utils/adminApi';
 
 const { TOPIC_ID } = process.env;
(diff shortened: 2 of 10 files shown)


useful-scripts/misc-test-utils/src/services/commonService.ts
```diff
@@ -0,0 +1,202 @@
...
+      fs.mkdirSync(`${cwd}/zip-files`);
+    }
+
+    if (zipGlob[1]) {
+      rimraf(zipGlob[1], () => {
+        console.log(chalk.green('cleaned stale zip files'));
+      });
+    }
+  },
+  createRelease: async (publicationId: string): Promise<string | null> => {
```
ntsim, line 54: Doesn't `releaseService.createRelease` already do this for us?

useful-scripts/misc-test-utils/src/services/commonService.ts
```diff
@@ -0,0 +1,202 @@
...
+    const metaGlob2 = metaGlob[0];
+
+    fs.rename(
+      `${metaGlob2}`,
+      `${cwd}/test-files/testfile-${randomId}.meta.csv`,
+      e => (e ? console.error(chalk.red(e)) : ''),
+    );
+  },
+
+  addMetaGuidance: async (
```
ntsim, line 125: Would suggest putting this in `releaseService` as meta guidance is a release-specific thing

useful-scripts/misc-test-utils/src/services/commonService.ts
```diff
@@ -0,0 +1,202 @@
...
+        ),
+      );
+      return releaseId;
+    } catch (e) {
+      errorHandler(e);
+    }
+    return null;
+  },
+
+  extractZip: async () => {
```
ntsim, line 89: Could make this method more generic so that it could handle other cases?

useful-scripts/misc-test-utils/src/services/commonService.ts
```diff
@@ -0,0 +1,202 @@
...
+    const archive = fs.existsSync(zippedFile[0]);
+    if (archive) {
+      // eslint-disable-next-line new-cap
+      const zip = new StreamZip.async({ file: cleanZipFile });
+      const count = await zip.extract(null, `${cwd}/test-files`);
+      console.log(chalk.green(`Extracted ${count} entries to test-files`));
+      await zip.close();
+    }
+  },
+  renameFiles: async () => {
```
ntsim, line 101: Could make this method more generic so that it could handle other cases?

useful-scripts/misc-test-utils/src/services/commonService.ts
```diff
@@ -0,0 +1,202 @@
...
+import errorHandler from '../utils/errorHandler';
+import { ReleaseData } from '../types/ReleaseData';
+import sleep from '../utils/sleep';
+
+const { ADMIN_URL, JWT_TOKEN } = process.env;
+
+const cwd = projectRoot;
+
+const commonService = {
+  validate: async () => {
```
ntsim, line 18: `validate` is a bit of a weird name as it's mostly checking that directories exist. Would suggest changing it to something else like `prepareDirectories`

useful-scripts/misc-test-utils/src/services/commonService.ts
```diff
@@ -0,0 +1,202 @@
...
+    const zipGlob = await globby(`${cwd}/*.zip`);
+    if (fs.existsSync(zipGlob[0]) && zipGlob[0].includes('archive.zip')) {
+      console.log(chalk.green('found zip file, continuing'));
+    } else {
+      throw new Error(
+        'No zip file named archive.zip in root dir! Exiting test with failures',
+      );
+    }
```
ntsim, line 26: This could go in a `validateArchive` or something method

useful-scripts/misc-test-utils/src/services/commonService.ts
```diff
@@ -0,0 +1,202 @@
...
+        },
+      });
+      return true;
+    } catch (e) {
+      errorHandler(e);
+    }
+    return false;
+  },
+
+  getFinalReleaseDetails: async (releaseId: string) => {
```
ntsim, line 149: Seems like this should also go in `releaseService`

useful-scripts/misc-test-utils/src/services/commonService.ts
```diff
@@ -0,0 +1,202 @@
...
+          id: releaseData.type.id,
+          title: releaseData.type.title,
+        },
+        contact: {
+          id: releaseData.contact.id,
+          teamName: releaseData.contact.teamName,
+          teamEmail: releaseData.contact.teamEmail,
+          contactName: releaseData.contact.contactName,
+          contactTelNo: releaseData.contact.contactTelNo,
+        },
```
ntsim, line 190: We could simplify this by spreading in the `releaseData` e.g.

```ts
return {
  ...releaseData,
  approvalStatus: 'Approved',
  // ... the things you're overriding it with 
};
```

I'm not entirely sure why we're overriding specific things though? Shouldn't we just return the release? In which case, this method could also just be named `getRelease`

useful-scripts/misc-test-utils/src/services/publicationService.ts
```diff
@@ -0,0 +1,104 @@
...
+            teamEmail: 'johndoe@gmail.com',
+            contactName: 'John Doe',
+            contactTelNo: '123456789',
+          },
+        },
+        headers: {
+          'Content-Type': 'application/json',
+          Authorization: `Bearer ${JWT_TOKEN}`,
+        },
+      });
```
ntsim, line 35: We can refactor all of the Axios invocations with an actual Axios instance that's pre-configured with the correct headers (so we don't have to repeat ourselves):

```ts
const adminApi = axios.create({
  baseURL: ADMIN_URL,
  headers: {
    'Content-Type': 'application/json',
    Authorization: `Bearer ${JWT_TOKEN}`,
  },
});

const publicationService = {
  createPublication: async () => {
    const res = adminApi.post(`/api/publications`, { .... });
    // etc
  },
};
```

This is basically what we do with all of our frontend services :wink:

useful-scripts/misc-test-utils/src/services/publicationService.ts
```diff
@@ -0,0 +1,104 @@
...
+          Authorization: `Bearer ${JWT_TOKEN}`,
+          'Content-Type': 'application/json',
+        },
+        data: obj,
+      });
+    } catch (e) {
+      errorHandler(e);
+    }
+  },
+  getPublicationProgress: async (releaseId: string) => {
```
ntsim, line 62: This seems like the wrong name - `getReleaseProgress` might be better?

Also - shouldn't this be in `releaseService`?

useful-scripts/misc-test-utils/src/services/publicationService.ts
```diff
@@ -0,0 +1,104 @@
...
+      console.log(
+        chalk.green(`Created publication. Status code ${res.status}`),
+      );
+      return res.data.id;
+    } catch (e) {
+      errorHandler(e);
+    }
+    return null;
+  },
+  publishRelease: async (obj: ReleaseData, releaseId: string) => {
```
ntsim, line 47: Shouldn't this be in `releaseService`?

useful-scripts/misc-test-utils/src/services/releaseService.ts
```diff
@@ -0,0 +1,43 @@
...
+          timePeriodCoverage: { value: 'AY' },
+          releaseName: 2222,
+          typeId: '1821abb8-68b0-431b-9770-0bea65d02ff0',
+          publicationId: 'c76c1bc4-0a01-4116-74cd-08d8eebbf0c1',
+          templateReleaseId: '',
+        },
+        headers: {
+          'Content-Type': 'application/json',
+          Authorization: `Bearer ${JWT_TOKEN}`,
+        },
```
ntsim, line 24: Similarly to `publicationService`, we can refactor all of these Axios invocations with a pre-configured Axios instance.

useful-scripts/misc-test-utils/src/services/subjectService.ts
```diff
@@ -0,0 +1,95 @@
...
+    try {
+      const res = await axios({
+        maxContentLength: Infinity,
+        maxBodyLength: Infinity,
+        method: 'POST',
+        url: `${ADMIN_URL}/api/release/${releaseId}/zip-data?title=importer-subject-${v4()}`,
+        data: form,
+        headers: {
+          ...form.getHeaders(),
+          Authorization: `Bearer ${JWT_TOKEN}`,
```
ntsim, line 37: Similarly to `publicationService`, we can refactor all of these Axios invocations with a pre-configured Axios instance.

ntsim (changes requested):

useful-scripts/misc-test-utils/src/services/publicationService.ts
luke-h1: cool. Initially, I took a look at the content API service but I didn't go forward with implementing an admin API instance as I wasn't sure what the param serializer was doing: 
```
export const contentApi = new Client(
  axios.create({
    baseURL: process.env.CONTENT_API_BASE_URL,
    paramsSerializer: commaSeparated,
  }),
);
```

ntsim: The param serializer is for query parameters. In the early days of the project, we tried to put the table tool query options inside of the URL query parameters. The table builder endpoint also used to be a GET.

This didn't work too well because we ended up with urls like `location[]=london&location[]=barnsley&...`, so we added the comma separated serializer so that we get query strings like `location=london,barnsley` instead (way more compact).

We didn't end up sticking with the GET endpoint though as it would lead to potentially huge query strings which could easily hit the URL character limit in browsers. We eventually turned it into a POST endpoint so that we could submit everything in the request body instead.

The comma separated serializer just stuck around after that really :smile:

I'd advise that you also use it in any of your own Axios instances too!

useful-scripts/misc-test-utils/src/services/commonService.ts
luke-h1: Going to leave this one for now. It'll be a good option to have but it'll require a fair bit of refactoring to handle this. I'll make a note of it so I can pick it up in the future

useful-scripts/misc-test-utils/src/services/publicationService.ts
```diff
@@ -0,0 +1,35 @@
...
+          contactTelNo: '123456789',
+        },
+      });
+      console.timeEnd('createPublication');
+      console.log(
+        chalk.green(`Created publication. Status code ${res.status}`),
+      );
+      return res.data.id;
+    } catch (e) {
+      errorHandler(e);
```
ntsim, line 30: Have you given anymore thought about wrapping the index with this try/catch error handler?

luke-h1: oops, I completely forgot about this. I'll move the try/catch stuff to `index.ts`

ntsim (approved):
Nice!

PR #2748, opened 2021-08-02
Ees 2547 tidy up tasks for 2156

duncan-at-hiveit:
This PR:
- removes the `allowAmendment` flag now that amendments are ready
- addresses leftover comments from original 2156 PR
- addresses leftover TODO to add Jest tests around removing a non-Amendment Methodology from a Publication

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/admin-dashboard/components/__tests__/MethodologySummary.test.tsx
+++ b/src/explore-education-statistics-admin/src/pages/admin-dashboard/components/__tests__/MethodologySummary.test.tsx
@@ -186,7 +186,7 @@ describe('MethodologySummary', () => {
       ).not.toBeInTheDocument();
 
       expect(
-        screen.queryByRole('button', {
+        screen.getByRole('button', {
           name: 'Link to an externally hosted methodology',
         }),
       ).toBeInTheDocument();
@@ -210,11 +210,11 @@ describe('MethodologySummary', () => {
       ).not.toBeInTheDocument();
 
       expect(
-        screen.queryByRole('button', { name: 'Create methodology' }),
+        screen.getByRole('button', { name: 'Create methodology' }),
       ).toBeInTheDocument();
 
       expect(
-        screen.queryByRole('button', {
+        screen.getByRole('button', {
           name: 'Link to an externally hosted methodology',
         }),
       ).toBeInTheDocument();
@@ -256,7 +256,7 @@ describe('MethodologySummary', () => {
         }),
       ).not.toBeInTheDocument();
 
-      expect(screen.queryByText('No methodologies added.')).toBeInTheDocument();
+      expect(screen.getByText('No methodologies added.')).toBeInTheDocument();
     });
 
     test('clicking the link external methodology button shows the form', async () => {
@@ -278,7 +278,7 @@ describe('MethodologySummary', () => {
 
       await waitFor(() => {
         expect(
-          screen.queryByText('Link to an externally hosted methodology', {
+          screen.getByText('Link to an externally hosted methodology', {
             selector: 'legend',
           }),
         ).toBeInTheDocument();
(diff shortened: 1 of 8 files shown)


benoutram (approved):

src/explore-education-statistics-admin/src/pages/admin-dashboard/components/__tests__/MethodologySummary.test.tsx
```diff
@@ -424,14 +434,90 @@ describe('MethodologySummary', () => {
...
+          <MethodologySummary
+            publication={testPublicationWithMethodologyCanRemove}
+            topicId={testTopicId}
+            onChangePublication={noop}
+          />
+        </MemoryRouter>,
+      );
+
+      expect(
+        screen.queryByText('Remove', { selector: 'button' }),
```
ntsim, line 450: General recommendation would be to use `getByRole('button', { name: 'Remove' });` for buttons.

The main advantage of using this is that the error output for `getByRole` explains what the accessible elements are (such as what buttons are actually on the page).

src/explore-education-statistics-admin/src/pages/admin-dashboard/components/__tests__/MethodologySummary.test.tsx
```diff
@@ -424,14 +434,90 @@ describe('MethodologySummary', () => {
...
+            topicId={testTopicId}
+            onChangePublication={noop}
+          />
+        </MemoryRouter>,
+      );
+
+      userEvent.click(screen.getByText('Remove', { selector: 'button' }));
+      await waitFor(() => {
+        expect(
+          screen.queryByText('Confirm you want to remove this methodology'),
```
ntsim, line 472: For things that actually exist on the page, would suggest using `getBy*` selectors rather than the equivalent `queryBy*`

This actually advised by the guy that actually wrote React Testing Library: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-query-variants-for-anything-except-checking-for-non-existence

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/admin-dashboard/components/__tests__/MethodologySummary.test.tsx
duncan-at-hiveit: Gotcha - have switched them all over, and also answered one of Amy's old mysteries as to why she could get one to work but not the other, being all down to visibility!  Thanks.

src/explore-education-statistics-admin/src/pages/admin-dashboard/components/__tests__/MethodologySummary.test.tsx
duncan-at-hiveit: Cool no worries - I figured that if it was within a waitFor() then it wouldn't really matter, as the `expect(...).toBeInTheDocument()` line would fail either way, one for if the `query` variant returned null to `expect` and one for if the `get` call failed because it blows up upon not finding anything.  I guess that the `get` variant's error message is more descriptive potentially?

Begs the question though, why write `expect(getByBlah(....)).toBeInTheDocument()` at all if the `getByBlah` explodes anyway?  Why not just write `getByBlah(...)` in isolation if it's going to explode anyway?  It's not like the `expect` it's getting passed to will be able to do anything if it throws an error.  I may well be missing an obvious thing here!

ntsim: Yep you're right there - the `getBy` variants are better for debugging purposes as they dump out the HTML.

Kent Dodds mentions it here: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-get-variants-as-assertions

Main benefit is just to communicate intent - i.e. that we're asserting something. I'd agree with him on this as just having free-floating `getBy` selectors feels a bit weird from a readability perspective.

ntsim (approved):

PR #2751, opened 2021-08-03
Add jest coverage option

amyb-hiveit:


src/explore-education-statistics-frontend/package.json
```diff
@@ -122,7 +123,7 @@
   ],
   "jest": {
     "collectCoverageFrom": [
-      "src/**/*.{js,jsx,ts,tsx}",
```
amyb-hiveit, line 125: Weirdly it hangs with js,jsx in, but only for this one.

ntsim (approved):
Nice!

PR #2733, opened 2021-07-27
EES-2265 methodology dependent release FE

amyb-hiveit:
Adds a 'When to publish' section to the methodology sign off form allow users to select a release to publish the methodology with.

I've copied and adapted the UI for setting a publish date on releases.

I've had to make some guesses & assumptions so will likely need a bit of tweaking when the BE is hooked up. 

This is hidden under the showWithRelease flag in the component, you'll need to flip it to true to see it. When the backend is ready this flag can be removed. With the flag set to false it should work as now so is safe to merge to dev.

Screenshot (41)

PR #2757, opened 2021-08-05
EES-2164 Allow approving methodology for publishing in combination with a Release

benoutram:
This PR:

- Adds a new Admin API endpoint `/api/methodology/{methodologyId}/unpublished-releases`  returning an ordered list of unpublished Releases for the owning or adopting publications using the methodology.
- Allows setting the chosen publishing strategy when approving a Release - Immediately or With a specifc release.
- Alters the Admin `UpdateMethodology` method to allow setting the chosen release id when approving with a specific release.
- Validates that the methodology can depend on a chosen release for publishing, i.e. checks that this release exists, that it's not already published, and that it's still linked via its publication to the methodology.
- Hooks up the FE work for this feature covered by @amyb-hiveit  in https://github.com/dfe-analytical-services/explore-education-statistics/pull/2733
- Renames the title on the Methodology Sign Off tab to Sign Off to make it consistent with release pages.
- Add a summary list to the read only view of the Methodology Sign off tab to contain the status, publishing strategy, and any release that's been chosen.
- Adds a new Jest test for MethodologyStatusPage which previously had no coverage.
- Adds a new Robot test suite `publish_methodology.robot` to test methodologies are publicly accessible when they should be.

A new endpoint for the list of unpublished methodologies was chosen rather than add this list to the existing  `Methodology SummaryViewModel`. This was to:
 - Avoid additional overhead in calculating this list which isn't always necessary - e.g. creating methodology, updating methodology, and making methodology amendments all return `MethodologySummaryViewModel`.
 - Make the unit test for this functionality clearly defined, avoiding changes to all of the existing tests for creating and updating methodologies which return `MethodologySummaryViewModel`.

image

image

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
@@ -22,6 +22,7 @@ using static GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.DbU
 using static GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.MapperUtils;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Validators.ValidationErrorMessages;
 using static GovUk.Education.ExploreEducationStatistics.Common.BlobContainers;
+using static GovUk.Education.ExploreEducationStatistics.Common.Model.TimeIdentifier;
 using static GovUk.Education.ExploreEducationStatistics.Common.Services.CollectionUtils;
 using static GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils.MockUtils;
 using static GovUk.Education.ExploreEducationStatistics.Content.Model.MethodologyPublishingStrategy;
@@ -898,8 +899,15 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.Method
         }
 
         [Fact]
-        public async Task UpdateMethodology_ApprovingUsingWithReleaseStrategy_NonLiveRelease()
+        public async Task UpdateMethodology_ApprovingUsingWithReleaseStrategy()
         {
+            var scheduledWithRelease = new Release
+            {
+                Id = Guid.NewGuid(),
+                TimePeriodCoverage = CalendarYear,
+                ReleaseName = "2021"
+            };
+
             var methodology = new Methodology
             {
                 PublishingStrategy = Immediately,
@@ -923,7 +931,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.Method
             {
                 LatestInternalReleaseNote = "Test approval",
                 PublishingStrategy = WithRelease,
-                // TODO SOW4 EES-2164 Add a ScheduledWithRelease here when implementing it
+                ScheduledWithReleaseId = scheduledWithRelease.Id,
                 Status = Approved,
                 Title = "Pupil absence statistics (updated): methodology"
             };
@@ -933,6 +941,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.Method
             await using (var context = InMemoryApplicationDbContext(contentDbContextId))
             {
                 await context.Methodologies.AddAsync(methodology);
+                await context.Releases.AddAsync(scheduledWithRelease);
                 await context.SaveChangesAsync();
             }
 
@@ -968,6 +977,10 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.Method
                 Assert.Null(viewModel.Published);
                 Assert.Equal(request.Status, viewModel.Status);
                 Assert.Equal(request.Title, viewModel.Title);
+
+                Assert.NotNull(viewModel.ScheduledWithRelease);
+                Assert.Equal(scheduledWithRelease.Id, viewModel.ScheduledWithRelease.Id);
+                Assert.Equal(scheduledWithRelease.Title, viewModel.ScheduledWithRelease.Title);
             }
 
             await using (var context = InMemoryApplicationDbContext(contentDbContextId))
@@ -980,6 +993,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.Method
                 Assert.Null(model.Published);
                 Assert.Equal(Approved, model.Status);
                 Assert.Equal(WithRelease, model.PublishingStrategy);
+                Assert.Equal(scheduledWithRelease.Id, model.ScheduledWithReleaseId);
                 Assert.Equal("pupil-absence-statistics-updated-methodology", model.Slug);
                 Assert.True(model.Updated.HasValue);
                 Assert.InRange(DateTime.UtcNow.Subtract(model.Updated.Value).Milliseconds, 0, 1500);
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
@@ -34,7 +34,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.Method
     public class MethodologyServiceTests
     {
         private static readonly Guid UserId = Guid.NewGuid();
-        
+
         [Fact]
         public async Task CreateMethodology()
         {
@@ -148,6 +148,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.Method
                 Assert.Equal(methodology.Id, viewModel.Id);
                 Assert.Equal(methodology.InternalReleaseNote, viewModel.InternalReleaseNote);
                 Assert.Equal(methodology.Published, viewModel.Published);
+                Assert.Equal(Immediately, methodology.PublishingStrategy);
                 Assert.Equal(methodology.Status, viewModel.Status);
                 Assert.Equal(methodology.Title, viewModel.Title);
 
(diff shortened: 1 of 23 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
```diff
@@ -980,9 +1081,261 @@ await using (var context = InMemoryApplicationDbContext(contentDbContextId))
...
+
+            var cacheService = new Mock<ICacheService>(Strict);
+            var contentService = new Mock<IMethodologyContentService>(Strict);
+            var imageService = new Mock<IMethodologyImageService>(Strict);
+            var methodologyRepository = new Mock<IMethodologyRepository>(Strict);
+            var publishingService = new Mock<IPublishingService>(Strict);
+
+            await using (var context = InMemoryApplicationDbContext(contentDbContextId))
+            {
+                var service = SetupMethodologyService(contentDbContext: context,
```
duncan-at-hiveit, line 1262: I'm tempted to say that rather than having to create all of these Mocks each time as Strict and then pass them into `SetupXxxxxService`, instead the default values that we construct in `SetupXxxxxService` should be strict.  As we're not setting up any calls on them in these cases, there's no need to have them available to pass into VerifyAllMocks()

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
```diff
@@ -161,6 +164,73 @@ await using (var context = InMemoryApplicationDbContext(contentDbContextId))
             }
         }
 
+        [Fact]
+        public async Task GetUnpublishedReleasesUsingMethodology_PublicationsHaveNoUnpublishedReleases()
```
duncan-at-hiveit, line 168: I was also expecting to see a test for when there *are* unpublished Releases, unless it's not necessary for a reason?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
```diff
@@ -980,9 +1081,261 @@ await using (var context = InMemoryApplicationDbContext(contentDbContextId))
...
+                    methodologyContentService: contentService.Object,
+                    methodologyImageService: imageService.Object,
+                    methodologyRepository: methodologyRepository.Object,
+                    publishingService: publishingService.Object);
+
+                var result = await service.UpdateMethodology(methodology.Id, request);
+                result.AssertNotFound();
+            }
+
+            VerifyAllMocks(cacheService, contentService, imageService, methodologyRepository, publishingService);
```
duncan-at-hiveit, line 1204: I'm always tempted to put this `VerifyAllMocks` directly under the `var result =` in new tests I write, because if the `result.AssertNotFound()` (or whatever) fails, you don't see whether or not it's down to a missing call on a Mock at that point.  Better to swap them round so that you can immediately see if the code wasn't using the Mocks correctly, and then that'll help to deduce what's going wrong, is my thinking.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
```diff
@@ -980,9 +1081,261 @@ await using (var context = InMemoryApplicationDbContext(contentDbContextId))
...
+            await using (var context = InMemoryApplicationDbContext(contentDbContextId))
+            {
+                var service = SetupMethodologyService(contentDbContext: context,
+                    cacheService: cacheService.Object,
+                    methodologyContentService: contentService.Object,
+                    methodologyImageService: imageService.Object,
+                    methodologyRepository: methodologyRepository.Object,
+                    publishingService: publishingService.Object);
+
+                var ex = await Assert.ThrowsAsync<ArgumentException>(() =>
```
duncan-at-hiveit, line 1269: In the general convention in our codebase, I'd expect this to be a Left result in the form of a  Bad Request, returning some ValidationErrorMessage.  Then we'd be checking here with `result.AssertBadRequest(MethodologyCannotDependOnPublishedRelease)`

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Methodologies/MethodologyController.cs
```diff
@@ -10,41 +13,46 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api.Metho
...
-        [HttpPost("api/publication/{publicationId}/methodology")]
+        [HttpPost("publication/{publicationId}/methodology")]
         public Task<ActionResult<MethodologySummaryViewModel>> CreateMethodology(Guid publicationId)
         {
             return _methodologyService
                 .CreateMethodology(publicationId)
                 .HandleFailuresOrOk();
         }
-        
-        [Produces("application/json")]
```
duncan-at-hiveit, line 37: I think these were here for Swagger to use for generating its doco.  Don't know if it's worth keeping them in or not though, as we certainly aren't consistent with it.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -200,6 +256,50 @@ private async Task DeleteMethodologyParentIfOrphaned(Methodology methodology)
...
+                return new NotFoundResult();
+            }
+
+            // Check that this release exists, that it's not already published, and that it's using the methodology
+            return await _persistenceHelper.CheckEntityExists<Release>(request.WithReleaseId.Value)
+                .OnSuccess(async release =>
+                {
+                    if (release.Live)
+                    {
+                        throw new ArgumentException("Methodology cannot depend on a published Release");
```
duncan-at-hiveit, line 279: Yeah, so here Id be tempted to return a BadResult, unless the thowing of the Exception does some DbContext-unwinding or something?

src/explore-education-statistics-admin/src/pages/methodology/edit-methodology/status/components/__tests__/MethodologyStatusPageTest.test.tsx
```diff
@@ -0,0 +1,209 @@
+import { render, screen, waitFor } from '@testing-library/react';
```
duncan-at-hiveit, line 1: Good work!

tests/robot-tests/tests/admin_and_public/bau/publish_methodology.robot
```diff
@@ -0,0 +1,100 @@
...
+    user navigates to methodologies page
+    user checks page does not contain    ${PUBLICATION_NAME}
+
+Verify that the methodology is not publicly accessible by URL without a published release
+    user goes to url    ${PUBLIC_METHODOLOGY_URL}
+    user waits until page contains    Page not found
+
+Alter the approval to publish the methodology with the release
+    user approves methodology for publication
+    ...    publication=${PUBLICATION_NAME}
```
duncan-at-hiveit, line 34: Ooh I'm going to have to start using named parameters like this

tests/robot-tests/tests/libs/admin-common.robot
```diff
@@ -542,18 +549,33 @@ user verifies release summary
...
 user changes methodology status to Approved
+    [Arguments]
+    ...    ${publishing_strategy}=Immediately
+    ...    ${with_release}=
+    ${is_publishing_strategy_with_release}=    Evaluate    '${publishing_strategy}' == 'WithRelease'
     user clicks button    Edit status
     user clicks element    id:methodologyStatusForm-status-Approved
     user enters text into element    id:methodologyStatusForm-latestInternalReleaseNote    Approved by UI tests
+    user clicks element    id:methodologyStatusForm-publishingStrategy-${publishing_strategy}
+    IF    '${publishing_strategy}' == 'WithRelease'
```
duncan-at-hiveit, line 560: Or can write `IF    ${is_publishing_strategy_with_release} is ${TRUE}`.  I'm thinking that was the original intention here :)

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
benoutram: That's a good idea, and it will remove quite a lot of unecessary setup. I've made this change now.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
benoutram: These tests did exist but I accidently lost them with a merge today! 😱 I've reinstated them now, thanks.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
benoutram: I saw you were doing this so thanks for explaining it. I've made this change now in combination with getting rid of those unnecessary mock declarations.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
benoutram: Not sure why I did it like this. I've swapped it over to an ActionResult so testing for a bad request now like you suggest. Thanks.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Methodologies/MethodologyController.cs
benoutram: I saw that we weren't adding them to newer stuff so decided to get rid of them here rather than spend time maintaining them. I will bring it up to discuss with the team.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
benoutram: I've changed this and also the one below now, thanks.

tests/robot-tests/tests/libs/admin-common.robot
benoutram: Ah that was the original intention, thanks. I just forgot to use it. 🙂

duncan-at-hiveit (approved):

PR #2760, opened 2021-08-05
EES-2170 EES-2301 contributors/publication owners to update methodologies

duncan-at-hiveit:
This PR:
- removes a Claim from the Analyst Role so they can no longer update any Methodology.
- splits the code path between updating a Methodology Status and updating Methodology details in `MethodologyService.UpdateMethodology()`.  This was to simplify our auth handlers and also reduce the amount of work that this service does or doesn't have to do depending on what is actually getting updated.
- adds lots more clauses to UpdateSpeificMethodologyAuthorizationHandler, for publication roles and release roles.

This PR has not yet:
- corrected 10 Robot tests that are failing due to changes in these update permissions (in `Release And Publication Role End To End Permissions`, `Create Methodology For Publication`, `Create Publication And Release` and `Methodology Page Absence`
- added any changes to `ApproveSpecificMethodologyAuthorizationHandler` or `MarkSpecificMethodologyAsDraftAuthorizationHandler`.


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -152,27 +152,62 @@ await _context.Entry(methodology.ScheduledWithRelease)
         public async Task<Either<ActionResult, MethodologySummaryViewModel>> UpdateMethodology(Guid id,
             MethodologyUpdateRequest request)
         {
-            return await _persistenceHelper.CheckEntityExists<Methodology>(id,
-                    HydrateMethodologyForManageMethodologySummaryViewModel)
-                .OnSuccess(methodology => CheckCanUpdateMethodologyStatus(methodology, request.Status))
-                .OnSuccess(_userService.CheckCanUpdateMethodology)
+            return await _persistenceHelper
+                .CheckEntityExists<Methodology>(id, HydrateMethodologyForManageMethodologySummaryViewModel)
+                .OnSuccess(methodology => methodology.Status != request.Status
```
benoutram, line 157: There's a problem here which prevents re-approving with  the same status but a different release note, or changing the publishing strategy from scheduled to with a specific release, or changing the specific release to a different release. At the moment it's only checking the status doesn't match the existing status.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -311,22 +334,22 @@ await _context.Entry(methodology.MethodologyParent)
 
             return status switch
             {
-                MethodologyStatus.Draft => _userService.CheckCanMarkMethodologyAsDraft(methodology),
-                MethodologyStatus.Approved => _userService.CheckCanApproveMethodology(methodology),
+                Draft => _userService.CheckCanMarkMethodologyAsDraft(methodology),
```
benoutram, line 337: I think there needs to be a change a few lines above this one to avoid skipping this check if the status is not changing. Since we now only call this method when going down the code  path of changing the status (as opposed to changing the details), we should always do this rather than return early.

This might need to remove the check on the ApproveSpecificMethodologyAuthorizationHandler which first checks that a methodology is not approved, but I still think it's better to do that and check they can approve it, rather than bypassing this check.

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandler.cs
```diff
@@ -13,33 +19,105 @@ public class UpdateSpecificMethodologyRequirement : IAuthorizationRequirement
...
+            _publicationRepository = publicationRepository;
         }
 
         protected override async Task HandleRequirementAsync(AuthorizationHandlerContext context,
             UpdateSpecificMethodologyRequirement requirement,
             Methodology methodology)
         {
+            // An Approved Methodology cannot be updated.  Instead, it should firstly be unapproved if permissions
+            // allow and then updated.
+            if (methodology.Approved)
```
benoutram, line 48: This breaks the list of unpublished releases used on the Sign Off tab which is returned by method GetUnpublishedReleasesUsingMethodology currently secured by this handler, when the methodology is already approved. If a user who can approve/unapprove goes back to the Sign Off tab they get an error.

The GetUnpublishedReleasesUsingMethodology  needs changing to be secured by those who can change status, rather than approve.

benoutram (changes requested):
In addition to the other comments, at least one suite of robot tests has failures because the 'Edit Methodology' button on the Publications and Releases dashboard is now only showing as 'View Methodology' once the release is approved, so it's not possible to go back and change the sign off. I.e this test fails: `Alter the approval to publish the methodology with the release`

A simple UI test change could be made to change the expected button text when viewing the methodology before navigating to the signoff page e.g. `view_button_text=View this methodology`. However I spoke with @lauraselby who said the button should remain as Edit Methodology for users that can still update it.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandlerTests.cs
```diff
@@ -51,42 +85,333 @@ public async Task NoClaimsAllowUpdatingPubliclyAccessibleMethodology()
...
+                            (user, methodology);
+
+                    await handler.HandleAsync(authContext);
+                    VerifyAllMocks(
+                        methodologyRepository,
+                        publicationRepository,
+                        userPublicationRoleRepository,
+                        userReleaseRoleRepository);
+
+                    // As the user has Publication Owner role on the owning Publication of this Methodology, they are
```
benoutram, line 300: This appears to only be testing release roles not the publication owner role

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
```diff
@@ -1122,28 +1062,29 @@ await using (var context = InMemoryApplicationDbContext(contentDbContextId))
                 Assert.Equal(Immediately, viewModel.PublishingStrategy);
                 Assert.Null(viewModel.ScheduledWithRelease);
                 Assert.Equal(request.Status, viewModel.Status);
-                Assert.Equal(request.Title, viewModel.Title);
+                Assert.Equal("Pupil absence statistics: methodology", viewModel.Title);
```
benoutram, line 1065: I think we could remove some of the assertions on checking details (title) when updating status and vice versa, checking status when updating title, for brevity.

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandler.cs
```diff
@@ -13,33 +19,105 @@ public class UpdateSpecificMethodologyRequirement : IAuthorizationRequirement
...
+                .Entry(methodology)
+                .Reference(m => m.MethodologyParent)
+                .LoadAsync();
+            
+            await _contentDbContext
+                .Entry(methodology.MethodologyParent)
+                .Collection(mp => mp.Publications)
+                .LoadAsync();
+
+            // A Publication Owner for the owning Publication of this Methodology can
```
benoutram, line 76: There's something missing in this comment! 🙂

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandlerTests.cs
benoutram: Edit: looks like this is just a copy/paste of the wrong comment in the code

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandler.cs
benoutram: I've removed it now in #2769

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
benoutram: Trimmed down some of the tests now in #2769

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandlerTests.cs
benoutram: Corrected this in #2769

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
benoutram: This is fixed by #2769

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
benoutram: This is fixed by #2769

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandler.cs
benoutram: This is fixed by #2769, by changing how `GetUnpublishedReleasesUsingMethodology ` is secured.

benoutram (approved):

PR #2762, opened 2021-08-07
EES-2548 Add AOP based caching 

ntsim:
# :warning: Depends on PR #2744 

This PR adds a new AOP based caching mechanism using compile-time injection with the [Aspect Injector](https://github.com/pamidur/aspect-injector) library. Caching is enabled via the processing of caching attributes (that derive from `CacheAttribute`) that annotate methods.

This is advantageous to as we don't need to directly call the caching services themselves. This has the nice benefit of allowing us to get rid of various caching service mocks across our tests.

We have written this in a pretty generic way so that we can facilitate the implementation of different kinds of caching mechanisms e.g. in-memory, Redis, database, etc.

Our initial implementing is with blob storage, augmenting the existing caching work carried out by Ben.

## Enabling caching

Cache attribute processing is not enabled by default as this can interfere with the behaviour of tests. Ideally, we should only enable it for specific tests that we are actually testing.

To do so, enable and configure any caching services we need in relevant application `Startup` classes:

```csharp
public class Startup
{
    public void Configure(IApplicationBuilder app,
        IWebHostEnvironment env,
        ILogger<Startup> logger)
    {
        CacheAspect.Enabled = true;
        BlobCacheAttribute.AddService("default", app.ApplicationServices.GetService<IBlobCacheService>());

        // do some more stuff...
    }
}
```

In the above example, we are simply configuring it with the application's default `IBlobCacheService`, but we could also add additional services if we want to select different caching services to use (e.g. to upload blobs to different storage accounts).

We have already set up the Content and Data APIs with the above configuration.

Note that it would be bad form to add further services after application startup. Unfortunately, we have to use static properties/methods on the various AOP classes to configure them. Ideally, we would use actual class instances registered with the container, but there is no way to do this with the Aspect Injector library.

## Using blob caching

1. Create a caching key that implements `IBlobCacheKey`. Current examples include `AllMethodologiesKey` or `DataBlockTableResultCacheKey`.

  If the key should be able to accept arguments, then these must be defined in the constructor so that we can inject them into the key.

  ```csharp
  public record MyCacheKey : IBlobCacheKey
  {
      public IBlobContainer Container => BlobContainers.PublicContent;

      public string Key { get; }

      public TestCacheKeyWithString(string key)
      {
          Key = key;
      }
  }
  ```

2. Annotate the method you need to cache like so:

  ```csharp
  [BlobCache(typeof(MyCacheKey))]
  public MyResult DoSomething(string key) 
  {
      // The returned MyResult will be cached.
  }
  ```

  We try to match any of the arguments for the annotated method with the cache key that been chosen (using `typeof(MyCacheKey)`).

  The method's returned value will be cached by serializing it into JSON and storing it in the blob container specified by the caching key's `Container` property. The `Key` property defines the specific place in the container that the blob should go.

  Upon the next invocation of the method, we'll check the cache to see if there is an existing item that we can return instead of running the underlying method again.

## Automatic unbox/boxing of 'result' types

To make caching as easy as possible, we automatically unbox/box specific 'result' types, specifically `Either`, `Task` and `ActionResult`. This allows us to work with the final value that we're actually interested (these other types are just wrappers).

For example, during the caching step, all of the below will serialize/unserialize `MyResult` into/out of blob storage (rather than the wrapping `Task`, `Either` or `ActionResult` itself).

```csharp
public MyResult DoSomething(string key);
public Task<MyResult> DoSomething(string key);
public Task<ActionResult<MyResult>> DoSomething(string key);
public Task<Either<ActionResult, MyResult>> DoSomething(string key);
```

This has been made possible through some fairly gratuitous reflection, so hold your breath if you ever read the relevant code...

It should be noted that we don't unbox other container types such as `List` or `Dictionary`. This would be pretty undesirable as we may want to cache lists and other types of collections.

## Writing other implementations

Further implementations can be developed simply by implementing the `CacheAttribute`, `ICacheService` and `ICacheKey` interfaces. Attribute processing should only really be done in `CacheAspect` so that we can avoid having to do it in each caching implementation.

We should ideally be able to write implementations using only a caching key to get/set a value to the cache.

We've written quite a number of test cases in `CacheAspectTests` that illustrate how attributes/methods/parameters are all processed and matched up.

## Relevant changes

- Updated these existing `ICacheService` interface so that it's more generic (previously it was tied blob storage due to receiving `IBlobContainer` parameters). `IBlobCacheService` now inherits `ICacheService`.
- Added various runtime type variants of service methods that previously required compile-time types e.g. `ICacheService.GetItem` and `IBlobStorageService.GetDeserializedJson`. This is specifically needed for most of this work as we don't know the compile-time types when the caching layer kicks in.


src/GovUk.Education.ExploreEducationStatistics.Common/Services/BlobCacheService.cs
```diff
@@ -11,97 +11,100 @@
...
+        public async Task<object?> GetItem(IBlobCacheKey cacheKey, Type targetType)
         {
+            var blobContainer = cacheKey.Container;
             var key = cacheKey.Key;
 
             // Attempt to read blob from the storage container
             try
             {
-                return await _blobStorageService.GetDeserializedJson<TEntity>(blobContainer, key);
+                return await _blobStorageService.GetDeserializedJson(cacheKey.Container, cacheKey.Key, targetType);
```
benoutram, line 107: I think one of the original limitations of this implementation which still exists here is we have no relation constraining the cache key and the connection which the BlobStorageService is using. For example the potential exists for the BlobCacheKey to use a private container name, but if the service is set up to use blob caching with a BlobStorageService wired with the PublicStorage connection string, the storage service will go ahead and create that container in the public account, rather than private as desired. So that could get messy, for example in the Admin if we try to use this for caching items in the private account, as well as needing to invalidate items in the public account.

Not sure if there's anything we can do here to prevent this.

I think overall getting rid of the divide would be best!

src/GovUk.Education.ExploreEducationStatistics.Common/Cache/Interfaces/IBlobCacheKey.cs
```diff
@@ -0,0 +1,8 @@
+#nullable enable
+namespace GovUk.Education.ExploreEducationStatistics.Common.Cache.Interfaces
+{
+    public interface IBlobCacheKey : ICacheKey
+    {
+        IBlobContainer Container { get; }
```
benoutram, line 6: I think it's a great change to add Container to they key!

src/GovUk.Education.ExploreEducationStatistics.Common/Services/Interfaces/ICacheService.cs
```diff
@@ -1,34 +1,19 @@
...
 using System;
 using System.Threading.Tasks;
 using GovUk.Education.ExploreEducationStatistics.Common.Cache.Interfaces;
-using GovUk.Education.ExploreEducationStatistics.Common.Model;
-using Microsoft.AspNetCore.Mvc;
 
 namespace GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces
 {
-    public interface ICacheService
+    public interface ICacheService<in TKey> where TKey : ICacheKey
```
benoutram, line 8: I think it's a great use of the interfaces to support other caching providers in the future, and in a way that we don't have redundant parameters like the blob container in the cache key 🙂

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Services/BlobCacheServiceTests.cs
```diff
@@ -0,0 +1,446 @@
+﻿#nullable enable
```
benoutram, line 1: Seems to be some files on this PR that could be moved rather than deleted and added, e.g.

`MockCacheServiceExtensions` -> `MockBlobCacheServiceExtensions`.
`BlobStorageCacheServiceTests` -> `BlobCacheServiceTests`.

Not sure if that's a limitation of git when moving and making changes in the same commit? Would make reviewing a bit easier if it was possible though.

src/GovUk.Education.ExploreEducationStatistics.Common/Cache/BlobCacheAttribute.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        public static void ClearServices()
+        {
+            Services.Clear();
+        }
+
+        public override async Task<object?> Get(ICacheKey cacheKey, Type returnType)
+        {
+            if (cacheKey is IBlobCacheKey key)
```
benoutram, line 44: I think this would be obvious if I switched to browsing with IDE on this branch, but why is this cacheKey a parameter and not the key that this is constructed with, which is already validated for being a type of IBlobCacheKey in the protected constructor of CacheAttribute?

src/GovUk.Education.ExploreEducationStatistics.Common/Cache/CacheAspect.cs
```diff
@@ -0,0 +1,378 @@
+#nullable enable
```
benoutram, line 1: Woah, this was obviously very difficult to implement, well done! 💯 🥇 
Nice work being able to box and unbox the different types.
I think rather than try and understand everything I'm going to accept this and go through understanding it further as I need to interact with it, as I already spent quite a lot of time on this PR.

Just one question here, this looks like it will cache an ActionResult that is a failure, that we might then not clear until we invalidate the cache key, whereas we might want to retry it if it's an intermittent failure. Is that correct?

src/GovUk.Education.ExploreEducationStatistics.Common/Cache/CacheAttribute.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// The base type of the cache key to use.
+        /// The <see cref="Key"/> must be assignable to this.
+        /// </summary>
+        protected virtual Type BaseKey => typeof(ICacheKey);
+
+        /// <summary>
+        /// Cache with highest priority is checked first.
+        /// 0 - Lowest, 255 - Highest.
+        /// </summary>
+        public byte Priority { get; init; } = 0;
```
benoutram, line 28: Cool so you would have more than one cache annotations, each with their own priority, and it would go through them in priority order looking for the first of the cache types to return a result? If none have a cache hit which one is set?

src/GovUk.Education.ExploreEducationStatistics.Common/Services/BlobCacheService.cs
benoutram: Edit: think I can see now how you have implemented service name attribute to be used as a way of targeting the correct service. 🙂

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/StringExtensionTests.cs
```diff
@@ -162,13 +162,6 @@ public void NullIfWhitespace_IsNullForWhiteSpaceString()
                 Assert.Null("".NullIfWhiteSpace());
                 Assert.Null(" ".NullIfWhiteSpace());
             }
-
-            [Fact]
-            public void NullStringCanAppendTrailingSlash()
-            {
-                string? input = null;
```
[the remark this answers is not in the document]
ntsim: Don't think so. We've enabled nullable on `StringExtensions` and changed the type of `AppendTrailingSlash` to a non-nullable string now, so shouldn't need to worry about null handling anymore in that method.

src/GovUk.Education.ExploreEducationStatistics.Common/Services/BlobCacheService.cs
ntsim: Yep indeed! You're right that it would be preferable to consolidate everything into a single account though to avoid any of the inherent public/private nastiness that we've seen over the last few weeks.

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Services/BlobCacheServiceTests.cs
ntsim: Seems to be a limitation of Git I'm afraid. I've tried doing the renaming in a separate commit, but it still thinks it's an entirely new file(s) in the diff :shrug: 

I've reverted it back to the original commit structure as there isn't any benefit from the separate commit

src/GovUk.Education.ExploreEducationStatistics.Common/Cache/BlobCacheAttribute.cs
ntsim: So the main thing is that we're inheriting from `CacheAttribute` that defines abstract `Get` and `Set` methods. As we want to override these, we can't then redefine them in `BlobCacheAttribute` because parameters can't be covariant without using generics.

You might be thinking this would simply work by adding a type parameter on `CacheAttribute`, but unfortunately nope! We can't do this because attributes can't have type parameters unfortunately :cry:

So the only option is to override the `CacheAttribute` methods and just do a type check once inside the method basically!

src/GovUk.Education.ExploreEducationStatistics.Common/Cache/CacheAspect.cs
ntsim: I don't think this should be the case. I'm guessing you probably know, but just in case - the `ActionResult` and `ActionResult<T>` types are really different kettles of fish and imply different things. Typically we would need to do `new ActionResult(something)` to get an `ActionResult<T>`, whereas `ActionResult` is typically created using derivatives such as `OkResult` or `NotFoundResult`.

`ActionResult` is like a 'void' type and should not have an internal value to return. Consequently, our unbox/boxing logic only applies for `ActionResult<T>` as this represents a boxed value T.

I think we prove that this works as expected in the `NoParams_ActionResult` and `NoParams_ActionResult_CacheMiss_NotFound` test cases in `CacheAspectTests`.

Of course, if you wanted to, you could try and cache a method like:

```csharp
[BlobCache(typeof(MyKey))]
public ActionResult DoSomething()
```

In this case, we would end up caching the `ActionResult` itself. A similar thing would happen if we used `Task` instead of `Task<T>` as well. However, it should be obvious why returning a void `ActionResult` or `Task` and expecting it to be cached doesn't really make sense - it would be like caching a `void` method.

We actually throw an error if the return type is void, but we currently don't do this for void `ActionResult` or `Task` because we can only assume the consumer actually wishes to serialize the `ActionResult` or `Task` objects into the cache for whatever strange reason.

src/GovUk.Education.ExploreEducationStatistics.Common/Cache/CacheAttribute.cs
ntsim: Yep indeed :smile: idea would be to potentially have multiple levels of cache if required (kinda doubt it, but the option's there).

When there is no cache hit, we set the real result in all of the caches!

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/StringExtensionTests.cs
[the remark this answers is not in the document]
benoutram: Cool 🙂  Thanks.

src/GovUk.Education.ExploreEducationStatistics.Common/Cache/CacheAspect.cs
benoutram: Thanks, I missed that but taken a closer look now. `CacheAspectTests` looks pretty exhaustive!

src/GovUk.Education.ExploreEducationStatistics.Common/Cache/BlobCacheAttribute.cs
benoutram: Thanks I see that now, and yes that's a pain! 🙂

benoutram (approved):

PR #2766, opened 2021-08-10
EES-2551 Publication owner's releases should appear in Admin Draft releases tab

markhiveit:


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseRepository.cs
```diff
@@ -49,12 +49,24 @@ public class ReleaseRepository : IReleaseRepository
...
+                .UserPublicationRoles
+                .Where(r => r.UserId == userId && r.Role == PublicationRole.Owner)
+                .Select(r => r.PublicationId)
+                .ToListAsync();
+            var userPublicationRoleReleaseIds = await _contentDbContext
+                .Releases
+                .Where(r => userPublicationIds.Contains(r.PublicationId))
+                .Select(r => r.Id)
+                .ToListAsync();
+            userReleaseIds.AddRange(userPublicationRoleReleaseIds);
```
ntsim, line 63: Am I being blind, or are there no tests for this?

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseRepository.cs
markhiveit: Yeah - my bad. For some reason I came in this morning and assumed I'd had already written all the tests necessary - I'll add test(s) for publication owners.

benoutram, line 64: Is there any chance that a release could appear twice in this list? E.g included because they have had contributor role on a release which hasn't been removed before receiving publication owner role on the publication?

markhiveit: I think there is, but I assumed it wasn't an issue - I'll investigate.

markhiveit: I've made the elements of userReleaseIds distinct. I don't think it affects the code as it currently works, but I can imagine future bugs where this causes and issue so this is probably for the best.

ntsim (approved):

PR #2769, opened 2021-08-10
EES-2170 EES-2301 Fix broken UI tests and address other comments of PR #2760

benoutram:
This PR targets the branch of #2760 in order to fix the broken UI tests on that branch.

* After the PR altered the handler for the Update Methodology permission, the 'Edit this methodology' and 'Edit this amendment' buttons were showing as 'View this methodology' and 'View this amendment' including for users with permission to Approve or Mark As Draft, causing test failures. After checking with @lauraselby that we still want them to display as 'Edit' for users with permissions to make this type of status change, this has been fixed. The FE is now aware of these additional permissions.
* Fetching the list of unpublished releases caused an error after going back to the Sign Off tab of an approved methodology. This was because it was secured by the Update Methodology permission in conjunction with the change to the hander for that permission. This is now secured by checking the user has permission to change the status instead.
* After the PR added split code paths between updating a Methodology Status and updating Methodology details (title) in `MethodologyService.UpdateMethodology`, some updates from the Sign Off tab were being ignored. Re-approving a Methodology by leaving the status as approved but changing the internal note, changing the publishing strategy, or changing the Release that the methodology should be published with were following the code path for updating the details (title only). This was fixed by adding comparisons into `MethodologyUpdateRequest` to check what fields are changing.

Also addresses other comments that I have left on #2760 ready to approve it.


src/explore-education-statistics-admin/src/pages/admin-dashboard/components/__tests__/MethodologySummary.test.tsx
```diff
@@ -345,6 +351,125 @@ describe('MethodologySummary', () => {
...
+              ],
+            }}
+            topicId={testTopicId}
+            onChangePublication={noop}
+          />
+        </MemoryRouter>,
+      );
+
+      expect(
+        screen.getByText('Edit this methodology', { selector: 'a' }),
```
ntsim, line 377: Would suggest using the `getByRole` selector instead. We can use it like this:

```tsx
screen.getByRole('link', { name: 'Edit this methodology' });
```

This is a nicer to debug with as it tells you what the accessible elements are on the page and also reads a bit nicer as well. These are the elements  you're normally concerned with interacting with as a user.

Would apply the same selector to the other test cases too

src/explore-education-statistics-admin/src/pages/admin-dashboard/components/MethodologySummary.tsx
```diff
@@ -96,118 +96,127 @@ const MethodologySummary = ({
...
+                          {canEdit
+                            ? 'Edit this amendment'
+                            : 'View this amendment'}
+                        </ButtonLink>
+                        <ButtonLink
+                          to={generatePath(methodologySummaryRoute.path, {
+                            publicationId,
+                            methodologyId: methodology.previousVersionId,
+                          })}
+                          className="govuk-button--secondary govuk-!-margin-left-4"
```
ntsim, line 160: Could use `variant="secondary"` prop instead of the `govuk-button--secondary` class

src/explore-education-statistics-admin/src/pages/admin-dashboard/components/MethodologySummary.tsx
```diff
@@ -96,118 +96,127 @@ const MethodologySummary = ({
...
+                <div className="govuk-grid-column-one-third dfe-align--right">
+                  {methodology.permissions.canDeleteMethodology && (
+                    <Button
+                      onClick={() =>
+                        setDeleteMethodologyDetails({
+                          methodologyId: methodology.id,
+                          amendment: methodology.amendment,
+                        })
+                      }
+                      className="govuk-button--warning"
```
ntsim, line 204: Could use the `variant="warning"` prop instead of this class

src/explore-education-statistics-admin/src/pages/admin-dashboard/components/MethodologySummary.tsx
```diff
@@ -96,118 +96,127 @@ const MethodologySummary = ({
...
-                {methodology.published ? (
-                  <FormattedDate>{methodology.published}</FormattedDate>
-                ) : (
-                  'Not yet published'
-                )}
-              </SummaryListItem>
-              {methodology.internalReleaseNote && (
-                <SummaryListItem term="Internal release note">
-                  {methodology.internalReleaseNote}
+        methodologies.map(methodology => {
```
ntsim, line 106: As we're not doing anything with the inverse of the `methodologies.length > 0`, we could simplify this to just:
```ts
methodolgies.map(methodology => ...
```

If there are no methodologies, nothing would be rendered anyway.

src/explore-education-statistics-admin/src/pages/admin-dashboard/components/__tests__/MethodologySummary.test.tsx
```diff
@@ -345,6 +351,125 @@ describe('MethodologySummary', () => {
       ).toBeInTheDocument();
     });
 
+    test('the edit button is shown when a user can approve the methodology', () => {
```
ntsim, line 354: Might want to say 'edit link' instead of button to be a little more specific

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -153,15 +154,20 @@ await _context.Entry(methodology.ScheduledWithRelease)
...
                 .OnSuccess(_ => GetSummary(id));
         }
 
         private Task<Either<ActionResult, Methodology>> UpdateMethodologyStatus(Methodology methodologyToUpdate,
             MethodologyUpdateRequest request)
         {
+            if (!request.IsStatusUpdateForMethodology(methodologyToUpdate))
+            {
+                // Status unchanged
+                return Task.FromResult(new Either<ActionResult, Methodology>(methodologyToUpdate));
```
ntsim, line 168: If we make this method async, couldn't we remove the `Task.FromResult`? 

Think we would need to await the `CheckCanUpdateMethodologyStatus` to avoid the compiler giving us a warning about not awaiting anything.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -197,6 +203,12 @@ await _context.Entry(methodology.ScheduledWithRelease)
         private Task<Either<ActionResult, Methodology>> UpdateMethodologyDetails(Methodology methodologyToUpdate,
             MethodologyUpdateRequest request)
         {
+            if (!request.IsDetailUpdateForMethodology(methodologyToUpdate))
+            {
+                // Details unchanged
+                return Task.FromResult(new Either<ActionResult, Methodology>(methodologyToUpdate));
```
ntsim, line 209: Similar to earlier, could async/await this method to avoid needing `Task.FromResult`.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Methodology/MethodologyUpdateRequest.cs
```diff
@@ -24,5 +24,18 @@ public class MethodologyUpdateRequest
...
+            return methodology.Status != Status
+                   || methodology.PublishingStrategy != PublishingStrategy
+                   || methodology.ScheduledWithReleaseId != WithReleaseId
+                   || methodology.InternalReleaseNote != LatestInternalReleaseNote;
+        }
+
+        public bool IsDetailUpdateForMethodology(Content.Model.Methodology methodology)
+        {
+            return methodology.Title != Title;
+        }
```
ntsim, line 39: Nice, I like this :+1:

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/admin-dashboard/components/MethodologySummary.tsx
benoutram: Thanks, great tip.

src/explore-education-statistics-admin/src/pages/admin-dashboard/components/MethodologySummary.tsx
benoutram: Cool, I've removed the length check now.

src/explore-education-statistics-admin/src/pages/admin-dashboard/components/__tests__/MethodologySummary.test.tsx
benoutram: Thanks, `edit button` renamed to `edit link`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
benoutram: Made this change now, thanks 🙂

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
benoutram: Ah yep, I've done this now 🙂

src/explore-education-statistics-admin/src/pages/admin-dashboard/components/__tests__/MethodologySummary.test.tsx
benoutram: Thanks, I see the benefit in the log! I have made these changes throughout now.

ntsim (approved):

PR #2770, opened 2021-08-10
EES-2544 use worker for csv download

amyb-hiveit:
download

- Moves processing the data for csv downloads to a worker so the main browser thread is not blocked.
- Adds a processing state that's shown while the data is processed, using the same style as the table tool next button.
- Does not change how ods downloads work as these are generally quicker and rely on access to the DOM.

This is used on:
- Admin:  table tool preview &  pre-release access
- Frontend: table tool final step, table tool fast track & permalink page

IE11 - importing corejs and regenerator-runtime make it work in IE11, there might be a better way to do it in the next config, but I couldn't get one to work.

Unit tests:
- I've added a generic worker mock so that jest can handle the worker. However I couldn't figure out getting this to work usefully to actually test the workers functionality.
- As a result, I've moved the test for downloading the csv file to the util test file (now split from the ods utils as it was getting a bit unmanageable all together) as I couldn't test it from within the worker onmessage event.
- Any ideas on how to test this better are welcome.


src/explore-education-statistics-common/src/hooks/useClickAway.ts
```diff
@@ -17,7 +17,7 @@ export default function useClickAway<
   events: (keyof DocumentEventMap)[] = ['click'],
 ) {
   useEffect(() => {
-    const handler: EventListenerOrEventListenerObject = event => {
+    const handler: EventListenerOrEventListenerObject = (event: Event) => {
```
amyb-hiveit, line 20: This isn't related to the worker change, but `tsc` started complaining about it 🤷

src/explore-education-statistics-common/src/modules/table-tool/components/workers/downloadTable.worker.ts
```diff
@@ -0,0 +1,15 @@
...
+import { getCsvData } from '@common/modules/table-tool/components/utils/downloadCsvUtils';
+
+/* eslint-disable no-restricted-globals */
+const ctx: Worker = self as never;
+
+/**
+ * Process the csv data in a new thread and post the result back to the main thread.
+ */
+ctx.addEventListener('message', event => {
+  const result = getCsvData(event.data.fullTable);
```
ntsim, line 12: Ah! Think we might have fallen into one of the main traps of using web workers. When passing data between to/from web workers, JS classes lose all of their inheritance/type information as it gets serialized/deserialized using a structured clone algorithm: https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm

This means, you're most likely just working with plain JS objects within the worker itself. Whilst this might not be an immediate problem - the main issue is that the types are completely wrong and could cause issues if we want to do things like `instanceof` checks or use the `id` getter. We might just have been fortunate enough that the code doesn't do any of that so far.

I think there's two main fixes. We either:

1. Change the type of the `fullTable` argument that is received by `getCsvData` to the correct version. This isn't super ideal as the type will lack information about what the filter types are for example.

2. Re-work all of the `mapFullTable` stuff so that we no longer use classes, but change everything to plain JS objects that attach their type information instead.

With option 2 - we would swap the classes for interfaces that would look something like:
```ts
interface Filter {
  value: string;
  label: string;
  group: string;
  id: string;
}

interface CategoryFilter extends Filter {
  isTotal: boolean;
  category: string;
  type: 'category';
}

interface LocationFilter extends Filter {
  level: boolean;
  geoJson?: GeoJsonFeature[];
  type: 'location';
}

// etc etc
```

If I could re-write table tool, option 2 is what I would go with :wink: It would make it pretty simple to move data to/from workers without needing any additional re-mapping to/from classes.

Understand that this might be a little daunting though, so fine with going with option 1 for now. If we start doing more web worker stuff in the future, I think we'll want to consider option 2 more seriously.

src/explore-education-statistics-common/src/modules/table-tool/components/DownloadTable.tsx
```diff
@@ -1,66 +1,17 @@
...
-  appendTitle(sheet, generateTableTitle(subjectMeta));
-  appendFootnotes(sheet, footnotes);
-
-  writeFile(workBook, `${fileName}.ods`, {
-    type: 'binary',
-    bookType: 'ods',
-  });
-};
+// eslint-disable-next-line import/no-webpack-loader-syntax
+import DownloadTableWorker from 'worker-loader?name=static/[hash].worker.js!@common/modules/table-tool/components/workers/downloadTable.worker';
```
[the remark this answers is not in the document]
amyb-hiveit: I tried worker-plugin first, but couldn't get it working. It's possible it was just due to the weirdness I had with my local after the webpack5 attempt, but not sure.

I had a look at Comlink but didn't really like that it 'hides the fact that you are working with workers.' I don't think workers are that incomprehensible that they need to be hidden. Maybe something to consider if we end up adding more workers and have problems.

The better solution would be to get the upgrade to Webpack 5 working then we wouldn't need a loader at all. Maybe something you could look at when you have some time? (I was just hitting my head against a brick wall with it).

src/explore-education-statistics-common/src/modules/table-tool/components/workers/downloadTable.worker.ts
amyb-hiveit: I wasn't aware of this, thanks! Unsurprisingly I don't really want to go down the re-working mapFullTable route right now :)
For option 1, I'm not sure what you mean by changing the argument to the correct version, can you expand on that a bit?

ntsim: No problem, suspected as much :laughing:

By having a correct version I mean we need some kind of type, maybe called `RawFullTable`, that is basically the same as `FullTable` (and `FullTableMeta`), but with the actual correct typings. It might look something like this:

```ts
// rawFullTable.ts

interface RawFilter {
  value: string;
  label: string;
  group?: string;
  id: string // Not entirely sure this would get serialized into the worker properly - please check
}

interface RawCategoryFilter extends RawFilter {
  isTotal: boolean;
  category: string;
}

// the other filter types...

export interface RawFullTableMeta {
  publicationName: string;
  subjectName: string;
  filters: Dictionary<{
    name: string;
    options: RawCategoryFilter[];
  }>;
}

export interface RawFullTable {
  title?: string;
  subjectMeta: RawFullTableMeta;
  results: TableDataResult[];
}
```

We would then use `RawFullTable` as the argument for `getCsvData` instead of `FullTable`.

I think we would want to preface this entire file by saying that we should only be using this in web worker contexts currently, and ideally we want to refactor `mapFullTable` to avoid classes. Could you ticket that up as well?

amyb-hiveit: Thanks. I'm doing this now and the id is not being serialised into the worker. The code doesn't use the id at all. so could just leave it like that (with a comment explaining), but it would be better to sort it out. I could clone the fulltable and loop over it to convert the ids to strings, or can you think of a better way?

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/DownloadTable.test.tsx
```diff
@@ -95,47 +95,6 @@ describe('DownloadTable', () => {
     ).toBeInTheDocument();
   });
 
-  test('downloads the csv file', async () => {
```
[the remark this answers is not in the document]
amyb-hiveit: I've had a try but not got anywhere using jsdom-worker today.

src/explore-education-statistics-common/src/modules/table-tool/components/workers/downloadTable.worker.ts
amyb-hiveit: Ok, I've left it without any more transforms and with comments about not using the id.

ntsim (approved):

PR #2771, opened 2021-08-10
EES-2273 add owning and other publications on methodology status page

amyb-hiveit:
methodology-status


benoutram (approved):
I could only see one robot UI keyword which checks any details on this status page, `user changes methodology status to Approved`.

I think we'll need some extra tests when we do the work for allowing adopting methodologies in the UI. It should be covered by EES-2154 and subtask EES-2286.

PR #2777, opened 2021-08-11
EES-2548 Fix intermittent caching test failures

ntsim:
This uses a combination of xUnit collections (so that the caching tests don't run in parallel), and a class fixture `CacheTestFixture` to enable/disable caching.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/BlobCacheAttributeTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/BlobCacheAttributeTests.cs
@@ -5,25 +5,25 @@ using GovUk.Education.ExploreEducationStatistics.Common.Cache;
 using GovUk.Education.ExploreEducationStatistics.Common.Cache.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces;
+using GovUk.Education.ExploreEducationStatistics.Common.Tests.Fixtures;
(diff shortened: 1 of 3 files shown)


src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/BlobCacheAttributeTests.cs
```diff
@@ -5,25 +5,25 @@
...
+using GovUk.Education.ExploreEducationStatistics.Common.Tests.Fixtures;
 using GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils;
 using Moq;
 using Xunit;
 
 namespace GovUk.Education.ExploreEducationStatistics.Common.Tests.Cache
 {
-    public class BlobCacheAttributeTests : IDisposable
+    [Collection(CacheTestFixture.CollectionName)]
+    public class BlobCacheAttributeTests : IClassFixture<CacheTestFixture>, IDisposable
```
benoutram, line 16: Not sure that you need `IDisposable` here or not, but it's inconsistent with the other class where you don't declare it.

benoutram (approved):
Seems like a neat solution to the problem!

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/BlobCacheAttributeTests.cs
ntsim: In this case we do - we have a `Dispose` method to do some cleanup after each test case :smile:

benoutram: Oops, I see that now! 🙂

PR #2782, opened 2021-08-12
EES-2499 - add custom env arg to run_tests.py

luke-h1:
This PR: 

- Allows you to pass in a custom `.env` file via the run_tests.py script.   

To use this new argument: 
`pipenv run python run_tests.py -f tests -i robot -b chrome -e <test_env> --custom-env <custom_env_name>`

### DIFF
--- a/tests/robot-tests/run_tests.py
+++ b/tests/robot-tests/run_tests.py
@@ -119,6 +119,10 @@ parser.add_argument("--fail-fast",
                     dest="fail_fast",
                     action='store_true',
                     help="stop test execution on failure")
+parser.add_argument("--custom-env",
+                    dest="custom_env",
+                    default=None,
+                    help="load a custom .env file (must be in ~/robot-tests directory)")
 
 """
 NOTE(mark): The admin and analyst passwords to access the admin app are stored in the CI pipeline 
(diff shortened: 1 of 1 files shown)


tests/robot-tests/run_tests.py
```diff
@@ -186,6 +190,14 @@
     robotArgs += ["--removekeywords", "name:operatingsystem.environment variable should be set"]
     robotArgs += ['--removekeywords',
                   'name:common.user goes to url']  # To hide basic auth credentials
+
+if args.custom_env: 
+    load_dotenv(args.custom_env)
+    assert os.getenv('PUBLIC_URL') is not None
+    assert os.getenv('ADMIN_URL') is not None
+    assert os.getenv('ADMIN_EMAIL') is not None
+    assert os.getenv('ADMIN_PASSWORD') is not None
```
markhiveit, line 199: I'd move these asserts outside of the if/else so that you don't need to have them twice.
```
if args.custom_env:
  load_dotenv...
else args.blah
  load_dotenv..
assert ...
```

markhiveit (changes requested):

tests/robot-tests/run_tests.py
```diff
@@ -119,6 +119,10 @@
                     dest="fail_fast",
                     action='store_true',
                     help="stop test execution on failure")
+parser.add_argument("--custom-env",
+                    dest="custom_env",
+                    default=None,
+                    help="load a custom .env file")
```
markhiveit, line 125: Could add a note here that it should be placed in the robot-tests directory. Might be overkill, but does clear up any potential confusion.

markhiveit (changes requested):

tests/robot-tests/run_tests.py
```diff
@@ -186,13 +190,17 @@
     robotArgs += ["--removekeywords", "name:operatingsystem.environment variable should be set"]
     robotArgs += ['--removekeywords',
                   'name:common.user goes to url']  # To hide basic auth credentials
+
+if args.custom_env: 
+    load_dotenv(args.custom_env)
+
 else:
     load_dotenv('.env.' + args.env)
```
markhiveit, line 198: Just realised that this should all be in an else after the `if args.ci:` above. Right now the script will try to load a .env file when running on the pipeline.
```
if args.ci:
  ....
else:
  if args.custom_env
    ...
  else
    ...
  assert
```

markhiveit (changes requested):

luke-h1:
Not sure why the build is showing as failing. Looks to be passing when viewed on Azure

markhiveit (approved):

PR #2784, opened 2021-08-12
EES-2426 Add new endpoint to dynamically download release files

ntsim:
This PR adds a new endpoint (`/api/releases/{releaseId}/files`), that allows users to select the files they wish to download into a zip file. Files are selected using the `fileIds` query parameter.

To provide a better user experience, we initiate the zip file download immediately before we have even downloaded any of the files from blob storage. As the files are downloaded, they get appended to the zip file's download. This means the user doesn't have to wait for all the files to be downloaded to the API before being served.

## Relevant changes

- Updated `ReleaseFileService` to use the `UserService.CheckCanViewRelease` method to check if the release has been published before permitting access to files.
- Handle request cancellations so that we prevent downloading of additional files from blob storage if the user decides to cancel the download.

## Other changes

- Added HTTP caching for the `ReleaseFileController` endpoints via the `Cache-Control` and `Last-Modified` headers. This has been set to have a duration of 300 seconds. We ideally want to prevent user's from excessively downloading files as this increases our bandwidth usage unnecessarily.
- Added `SeparatedQuery*` model binding classes that allow our APIs to now parse query strings where we use comma-separated strings for array-like query params e.g. `fileIds=guid1,guid2,guid3` can be matched to a `List<Guid>`. Values for additional query string keys of the same name will simply be appended to the entire list.


src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+            writer.Flush();
+        }
+
+        public static string ReadToEnd(this Stream stream)
+        {
+            using var reader = new StreamReader(stream, leaveOpen: true);
+            return reader.ReadToEnd();
+        }
+
+        public static string ComputeMd5Hash(this Stream stream)
```
benoutram, line 23: Just wondering where we use this? I think I've only seen creating the unique zip filename by String on this PR, not by stream.

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Either.cs
```diff
@@ -281,8 +281,23 @@ public static class EitherTaskExtensions
             return await func(firstResult.Right);
         }
 
-        public static async Task<Either<TFailure1, TFaillure2>> OnFailureDo<TFailure1, TFaillure2>(
-            this Task<Either<TFailure1, TFaillure2>> task,
+        public static async Task<Either<Unit, TSuccess>> OnFailureVoid<TFailure, TSuccess>(
```
benoutram, line 284: You could add a test to EitherTests for this?

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/MockBlobStorageServiceExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Threading;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Common.Model;
+using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces;
+using Moq;
+using Moq.Language.Flow;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Tests.Extensions
+{
+    public static class MockBlobStorageServiceExtensions
```
benoutram, line 12: Seems like a nice use of extensions to slim down the tests! 🙂

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Services/ReleaseFileService.cs
```diff
@@ -85,14 +169,9 @@ private async Task<FileStreamResult> DownloadToStream(BlobInfo blob)
...
         {
-            if (!await _blobStorageService.CheckBlobExists(PublicReleaseFiles, path))
-            {
-                return new NotFoundResult();
-            }
-
-            var blob = await _blobStorageService.GetBlob(PublicReleaseFiles, path);
+            var blob = await _blobStorageService.FindBlob(PublicReleaseFiles, path);
 
-            if (!blob.IsReleased())
```
benoutram, line 95: Cool, does this mean we can get rid of `IsReleased` and clean up Publisher to no longer be setting the release date time blob property, or still more cleanup to do?

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Services/ReleaseFileService.cs
```diff
@@ -48,31 +68,95 @@ public class ReleaseFileService : IReleaseFileService
...
+                await _blobStorageService.DownloadToStream(
+                    containerName: PublicReleaseFiles,
+                    path: releaseFile.PublicPath(),
+                    stream: entryStream,
+                    cancellationToken: cancellationToken
+                );
+            }
+        }
+
+        public async Task<Either<ActionResult, FileStreamResult>> StreamAllFilesZip(Guid releaseId)
```
benoutram, line 138: Just thinking, will we still need the All Files zip when this functionality is eventually released? Not having it as a database File ref has caused us some overhead in the code that could be cleaned up if we can agree it's no longer required.

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Services/ReleaseFileService.cs
```diff
@@ -48,31 +68,95 @@ public class ReleaseFileService : IReleaseFileService
...
+                }
+
+                var entry = archive.CreateEntry(releaseFile.File.ZipFileEntryName());
+                await using var entryStream = entry.Open();
+
+                await _blobStorageService.DownloadToStream(
+                    containerName: PublicReleaseFiles,
+                    path: releaseFile.PublicPath(),
+                    stream: entryStream,
+                    cancellationToken: cancellationToken
```
benoutram, line 133: Nice to implement cancellation on the `BlobStorageService`! 🙂

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Services/ReleaseFileService.cs
```diff
@@ -22,24 +26,40 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Api.Services
...
+            FileType.Ancillary,
+            FileType.Data,
+        };
+
+        private static readonly FileType[] ZipFileTypes = {
+            FileType.Ancillary,
+            FileType.Chart,
+            FileType.Data,
+            FileType.DataGuidance,
+            FileType.Image,
```
benoutram, line 39: Do we plan to allow the content images to be included? Don't think they are currently in the All Files zip.

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Services/ReleaseFileServiceTests.cs
```diff
@@ -17,14 +22,23 @@
...
+        public void Dispose()
+        {
+            // Cleanup any files that have been
+            // written to the filesystem.
+            _filePaths.ForEach(File.Delete);
+        }
+
         [Fact]
         public async Task Stream()
         {
```
benoutram, line 44: Think this test is now `StreamFile` not `Stream`

src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            using var writer = new StreamWriter(stream, leaveOpen: true);
+            writer.Write(text);
+            writer.Flush();
+        }
+
+        public static string ReadToEnd(this Stream stream)
+        {
+            using var reader = new StreamReader(stream, leaveOpen: true);
+            return reader.ReadToEnd();
```
benoutram, line 20: Cool, seems like a nice addition to test stream content is correct within unit tests! 🙂

src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/StreamExtensions.cs
ntsim: Ah good catch. I actually forgot to use this in the HTTP caching - it's now being used in `CacheWithETag` to prevent user's re-requesting the same file over and over (at least in the browser).

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Either.cs
ntsim: Done!

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Services/ReleaseFileService.cs
ntsim: Not quite. Currently `MethodologyImageService` still uses this. Think we'd need to sort out the securing of that endpoint first to use the content DB.

I've created an overall ticket for removing the `IsReleased` meta from the blob anyway: https://dfedigital.atlassian.net/browse/EES-2619

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Services/ReleaseFileService.cs
ntsim: Not fully fleshed out my plan for this - but I'm thinking about the following:

- Removing the 'All files' zip from the publishing phase completely
- Still have an endpoint to fetch the 'All files' zip. When requested for the first time, we zip this up in a similar manner to the `/files` endpoint.
- Once streamed to the user, we then upload the 'All files' zip to blob storage.
- On subsequent requests, we just re-fetch that blob, unless the release has been re-published, in which case we regenerate the blob (and delete the old one).

This would basically be a slight variation of our cache-aside strategy that we've done for the `BlobCache` stuff.

If this all sounds sensible to you, I'm planning on doing this all as part of this upcoming ticket: https://dfedigital.atlassian.net/browse/EES-2546

I think this approach has the advantage that we don't need to create a migration for existing 'All files' zips to add the data guidance text file to each one :smile:

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Services/ReleaseFileService.cs
ntsim: Ah good point, I think I was being silly and looked in the wrong place in the publisher when I was deciding what files would be downloadable! Will remove the images.

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Services/ReleaseFileServiceTests.cs
ntsim: Done!

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Services/ReleaseFileService.cs
benoutram: I forgot about the `MethodologyImageService`! Great, thanks for raising the ticket.

src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/StreamExtensions.cs
benoutram: Cool, just seen this update.

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Services/ReleaseFileService.cs
benoutram: Just checked the update, looks good. 🙂

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Services/ReleaseFileService.cs
benoutram: This sounds like a really good approach to cache the file and still only zip 'all' the files once. We get to clean stuff up and fix the problem with data guidance too. 🙂

benoutram (approved):

PR #2786, opened 2021-08-13
EES-2614 Unit tests for Publication#LatestRelease and #LatestPublishe…

markhiveit:
…dReleasse

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/PublicationTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/PublicationTests.cs
@@ -1,0 +1,200 @@
+using System;
+using System.Collections.Generic;
+using Xunit;
+using static GovUk.Education.ExploreEducationStatistics.Common.Model.TimeIdentifier;
+using static GovUk.Education.ExploreEducationStatistics.Content.Model.ReleaseApprovalStatus;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Model.Tests
+{
+    public class PublicationTests
+    {
+        [Fact]
+        public void LatestPublishedRelease_NewDraftAmendment()
+        {
+            var originalReleaseId = Guid.NewGuid();
+            var amendedRelease1Id = Guid.NewGuid();
+            var amendedRelease2Id = Guid.NewGuid();
+            var publication = new Publication
+            {
+                Releases = new List<Release>
+                {
+                    new()
+                    {
+                        Id = originalReleaseId,
+                        Version = 0,
+                        ReleaseName = "2000",
+                        TimePeriodCoverage = AcademicYear,
+                        Published = DateTime.UtcNow.AddDays(-2),
+                        PreviousVersion = null,
+                        ApprovalStatus = Approved,
+
+                    },
+                    new()
+                    {
+                        Id = amendedRelease1Id,
+                        Version = 1,
+                        ReleaseName = "2000",
+                        Published = DateTime.UtcNow.AddDays(-1),
+                        TimePeriodCoverage = AcademicYear,
+                        PreviousVersionId = originalReleaseId,
+                        ApprovalStatus = Approved,
+                    },
+                    new()
+                    {
+                        Id = amendedRelease2Id,
+                        Version = 2,
+                        ReleaseName = "2000",
+                        Published = null,
+                        TimePeriodCoverage = AcademicYear,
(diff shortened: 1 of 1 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/PublicationTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            {
+                Releases = new List<Release>
+                {
+                    new()
+                    {
+                        Id = release2000Id,
+                        Version = 0,
+                        ReleaseName = "2000",
+                        TimePeriodCoverage = TimeIdentifier.AcademicYear,
+                        Published = DateTime.UtcNow.AddDays(-2),
```
benoutram, line 74: Published value doesn't matter here as long as they are published, and setting up one with -2 and one with -1 implies you might be expecting it to influence the order in some way. I'd be tempted to just use `DateTime.UtcNow` as it's the Year and TimePeriod and presence of the Published value which are important.

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/PublicationTests.cs
```diff
@@ -0,0 +1,160 @@
...
+                        Published = null,
+                        TimePeriodCoverage = TimeIdentifier.AcademicYear,
+                        ApprovalStatus = ReleaseApprovalStatus.Draft,
+                    },
+                },
+            };
+
+            var result = publication.LatestPublishedRelease();
+            Assert.Equal(release2001Id, result.Id);
+        }
```
benoutram, line 101: I think you could also add tests for LatestPublished_NoPublishedRelease and LatestPublished_NoReleases where we'd expect to catch bugs if we accidently assume at least one Release exists.

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/PublicationTests.cs
```diff
@@ -0,0 +1,160 @@
...
+                        TimePeriodCoverage = TimeIdentifier.AcademicYear,
+                        ApprovalStatus = ReleaseApprovalStatus.Draft,
+                    },
+                },
+            };
+
+            var result = publication.LatestRelease();
+            Assert.Equal(release2001AmendmentId, result.Id);
+        }
+    }
```
benoutram, line 159: Like above I think you could add a `LatestRelease_NoRelease`. Has Publication.cs enabled nullable reference types yet? If so should be clearer to see that we can expect both these methods to return null.

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/PublicationTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            var publication = new Publication
+            {
+                Releases = new List<Release>
+                {
+                    new()
+                    {
+                        Id = release2000Id,
+                        Version = 0,
+                        ReleaseName = "2000",
+                        TimePeriodCoverage = TimeIdentifier.AcademicYear,
```
benoutram, line 119: You could use a static import for the TimeIdentifiers and ReleaseApprovalStatus

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/PublicationTests.cs
```diff
@@ -0,0 +1,160 @@
...
+                        Published = DateTime.UtcNow.AddDays(-1),
+                        ApprovalStatus = ReleaseApprovalStatus.Approved,
+                    },
+                    new()
+                    {
+                        Id = release2001Id,
+                        Version = 0,
+                        ReleaseName = "2001",
+                        Published = DateTime.UtcNow.AddDays(-1),
```
benoutram, line 139: Same comment about Published date/time as above. If they were all published on the same day we should expect the same result, but the data is setting up days -1 and -2 which made me think it's influencing the order in some way.

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/PublicationTests.cs
markhiveit: I've enabled nullable to Publication.cs and PublicationTests.cs. I had to make various changes to Publication.cs to make sure no changes appear if you make a migration - which I tested and a blank migration was created.

duncan-at-hiveit (approved):

PR #2790, opened 2021-08-17
EES-2236 Add data guidance file to dynamic zip files

ntsim:
This PR adds the `data-guidance.txt` file to dynamic zip files generated from `/api/releases/{releaseId}/files`.

As part of this, we have had to move most of the `DataGuidanceFileWriter` work into a new `Content.Services` package. We've also moved all of the existing services from `Content.Api` into here as well.

## Other changes

- Fixed zip file contents having incorrect file permissions on Unix (Mac/Linux) due to .NET not automatically adding these. Files would be a non-read/write state. We now set these manually with 644 permissions, although this shouldn't be necessary anymore in .NET 6.


src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ReleaseFileServiceTests.cs
```diff
@@ -374,11 +372,37 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
...
+
+            var dataGuidanceFileWriter = new Mock<IDataGuidanceFileWriter>(MockBehavior.Strict);
+
+            dataGuidanceFileWriter
+                .Setup(
+                    s => s.WriteToStream(
+                        It.IsAny<Stream>(),
+                        It.Is<Release>(r => r.Id == release.Id),
+                        It.Is<IEnumerable<Guid>>(
+                            ids => ids.All(id => subjectIds.Contains(id))
```
duncan-at-hiveit, line 388: A shame that there's not an `It` method that does this for us already.  Also a shame that we can't create extension methods on static classes or we could have created one for reuse!

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ReleaseFileServiceTests.cs
```diff
@@ -374,11 +372,37 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
...
+                        )
+                    )
+                )
+                .Returns<Stream, Release, IEnumerable<Guid>?>((stream, _, _) => Task.FromResult(stream))
```
duncan-at-hiveit, line 392: Or can you write 

`.ReturnsAsync<Stream, Release, IEnumerable<Guid>?>((stream, _, _) => stream)`

?

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ReleaseFileServiceTests.cs
```diff
@@ -374,11 +372,37 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
...
+                .Callback<Stream, Release, IEnumerable<Guid>?>(
```
duncan-at-hiveit, line 393: Cool, not seen `Callback` before

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ReleaseFileServiceTests.cs
```diff
@@ -392,15 +416,134 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
...
                 Assert.Equal("data/data.csv", zip.Entries[1].FullName);
                 Assert.Equal("Test data blob", zip.Entries[1].Open().ReadToEnd());
+
+                // Data guidance is generated if there is at least one data file
+                Assert.Equal("data-guidance/data-guidance.txt", zip.Entries[2].FullName);
+                Assert.Equal("Test data guidance blob", zip.Entries[2].Open().ReadToEnd());
             }
 
-            MockUtils.VerifyAllMocks(blobStorageService);
+            MockUtils.VerifyAllMocks(blobStorageService, dataGuidanceFileWriter);
```
duncan-at-hiveit, line 431: I was saying to Ben the other day, that I've started to put `VerifyAllMocks()` calls directly after the `var result = service.blah(...)` line and before the `result.AssertRight()`, the reason being that if `result` is actually a failure and that failure is down to a mock invocation failing to be called, you'll not know about it as the test'll fail at the `AssertRight` leaving you to have to debug the code or whatever to work out why.

Putting the Verify right after the service call will let you know if it's down to a mock failing to be called straight away.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ReleaseFileService.cs
```diff
@@ -121,7 +125,10 @@ public class ReleaseFileService : IReleaseFileService
                     continue;
                 }
 
-                var entry = archive.CreateEntry(releaseFile.File.ZipFileEntryName());
+                var entry = archive
+                    .CreateEntry(releaseFile.File.ZipFileEntryName())
+                    .SetUnixPermissions("664");
```
duncan-at-hiveit, line 130: What effect does this have on non-Unix OS's?

src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/Services/DataGuidanceFileServiceTests.cs
```diff
@@ -136,7 +175,10 @@ await using (var contentDbContext = InMemoryContentDbContext(contextId))
         [Fact]
         public async Task CreateDataGuidanceFile_UploadThrows()
         {
-            var release = new Release();
+            var release = new Release
+            {
+                MetaGuidance = "Test guidance"
```
duncan-at-hiveit, line 180: I can't see this text being checked for later on in the test.  Is it not important to test the actual output based on this text value, or are you happy that it's covered elsewhere?

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ReleaseFileServiceTests.cs
ntsim: Yeh this kinda sucked. Moq has its nice parts, but it's also got a number of rough edges for stuff like this

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ReleaseFileServiceTests.cs
ntsim: Nice try but doesn't quite work! This leads to an ambiguous invocation to one of the overloads of `ReturnsAsync`. The real number is this horror show :laughing:

```cs
.ReturnsAsync<Stream, Release, IEnumerable<Guid>?, IDataGuidanceFileWriter, Stream>((stream, _, _) => stream)
```

I'm sure you can understand why I opted for the former.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ReleaseFileService.cs
ntsim: None as far as I've seen! File permissions appear to be fine for Windows

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ReleaseFileServiceTests.cs
ntsim: Fair enough, Will move these to after the method invocation!

src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/Services/DataGuidanceFileServiceTests.cs
ntsim: Good point, I originally had a check in `DataGuidanceFileWriter` to ensure there was guidance, but I removed it when I moved it to `Content.Services` so that it could work with older releases that may not have `MetaGuidance` defined.

I think my intention was to put the check in `DataGuidanceFileService` but I must have forgotten! Thanks!

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Mappings/MappingProfiles.cs
```diff
@@ -16,15 +16,15 @@ public MappingProfiles()
             CreateContentBlockMap();
 
             CreateMap<ContentSection, ContentSectionViewModel>().ForMember(dest => dest.Content,
-                m => m.MapFrom(section => section.Content.OrderBy(contentBlock => contentBlock.Order)));
+                m => m.MapFrom(section => Enumerable.OrderBy<ContentBlock, int>(section.Content, contentBlock => contentBlock.Order)));
```
benoutram, line 19: What's the reason for moving away from the extension method for this and the next two changes in this file?

ntsim: No idea, I didn't actually write this. Must have been part of Rider's refactoring. Will revert it.

src/GovUk.Education.ExploreEducationStatistics.Publisher/GovUk.Education.ExploreEducationStatistics.Publisher.csproj
```diff
@@ -6,8 +6,8 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="AngleSharp" Version="1.0.0-alpha-844" />
-    <PackageReference Include="AutoMapper" Version="9.0.0" />
-    <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="7.0.0" />
+    <PackageReference Include="AutoMapper" Version="10.1.1" />
```
benoutram, line 9: Cool, maybe we could look at upgrading AutoMapper across the other projects as we touch them 🙂

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Mappings/MappingProfiles.cs
benoutram: Ok great! Thanks 🙂

benoutram (dismissed):

benoutram (approved):

PR #2794, opened 2021-08-18
EES-2587 Refactor LocationRepository#GetObservationalUnits

markhiveit:


markhiveit:
Maybe we want to merge this to the same branch that all the other EES-2587 subtasks go to? Although I don't think there is anything stopping this going to dev.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/LocationRepository.cs
```diff
@@ -16,15 +16,34 @@ public LocationRepository(StatisticsDbContext context) : base(context)
 
         public Dictionary<GeographicLevel, IEnumerable<IObservationalUnit>> GetObservationalUnits(Guid subjectId)
         {
-            var locations = GetLocationsGroupedByGeographicLevel(subjectId);
-            return GetObservationalUnits(locations);
+            var observations = _context.Observation
```
duncan-at-hiveit, line 19: In data files where there are, say, 250,000 rows, we want to avoid pulling out 250,000 records out of the database which could take some time for EF and some chunk of memory usage too.

Could you just confirm that the 2 usages of this query in the `GetObservationalUnits(IQueryable<Observation> observations)` method don't cause this list to be pulled back early before initiating both their subqueries?  I'm guessing they're fine to be honest:

```
var geographicLevels = observations
                .Select(o => o.GeographicLevel)
                .Distinct()
                .ToList();
```

and

```
var locations = observations
                            .Include(observation => observation.Location)
                            .Where(observation => observation.GeographicLevel == geographicLevel)
                            .Select(observation => observation.Location)
                            .Distinct()
                            .ToList();
```

I think that both of these will just grab what they need without triggering the full retrieval of 250,000 Observation instances.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/LocationRepository.cs
```diff
@@ -16,15 +16,34 @@ public LocationRepository(StatisticsDbContext context) : base(context)
...
+                .Where(o => o.SubjectId == subjectId);
+            return GetObservationalUnits(observations);
         }
 
         public Dictionary<GeographicLevel, IEnumerable<IObservationalUnit>> GetObservationalUnits(
             IQueryable<Observation> observations)
         {
-            var locations = GetLocationsGroupedByGeographicLevel(observations);
-            return GetObservationalUnits(locations);
+            var geographicLevels = observations
```
duncan-at-hiveit, line 27: Looks a great deal cleaner now :)  Nice one.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/LocationRepository.cs
markhiveit: Discussed this with Duncan on Slack. The original code is written as it is to avoid multiple calls to the Observation table. So I've changed the code to do this, but have submitted it in a separate PR: https://github.com/dfe-analytical-services/explore-education-statistics/pull/2797

PR #2799, opened 2021-08-19
EES-2632 Refactor LocationRepository#GetObservationalUnits

markhiveit:


markhiveit:
Just added another comment, if anyone is wondering what I've force pushed.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/LocationRepository.cs
```diff
@@ -16,15 +16,49 @@ public LocationRepository(StatisticsDbContext context) : base(context)
...
+                .Select(elem => elem.GeographicLevel)
+                .Distinct()
+                .ToList();
+
+            // NOTE: Get all locations in a single DB call
+            var locationIds = list
+                .Select(elem => elem.LocationId)
+                .Distinct()
+                .ToArray();
+            var locationsDict = Find(locationIds)
```
benoutram, line 43: Just wondering if we are any likely to hit any character limit on SQLServer now when executing this. Following the previous logic, looks like we queried the locations in batches by geographic level, but now without this batching.
Have you tried this with a large subject data file?

benoutram: Did a quick test and I think this will fail at somewhere between 40-50,000 unique location id's. I doubt we'd ever hit a subject file with this many locations.

markhiveit: Talked to Ben about this on Slack and agreed to create a ticket to address it: EES-2645.

benoutram (approved):

PR #2802, opened 2021-08-19
EES-2520 Merge data API's `/themes` endpoint with content API

ntsim:
This PR consolidates the data API's `/themes` endpoint with the content API's. We have implemented this by adding a query parameter `publicationFilter` option to the endpoint.

When a filter of `LatestData` is provided, we filter out any publications that do not have data in their latest release. This mirrors the previous behaviour in the data API endpoint.

## Related changes

- Renamed `PublicationService` to less misleading `ThemeService`. We will need `PublicationService` later for work in EES-2366.
- Added blob caching for publication tree returned from using `publicationFilter=LatestData`.
- Added the `System.Linq.Async` package in `Content.Services` to handle async enumerables (we don't need to handroll our own in `EnumerableExtensions` anymore).


src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/ThemeControllerTests.cs
```diff
@@ -19,10 +19,10 @@ public class ThemeControllerTests
         [Fact]
         public async Task GetThemes()
         {
-            var publicationService = new Mock<IPublicationService>(MockBehavior.Strict);
+            var publicationService = new Mock<IThemeService>(MockBehavior.Strict);
```
benoutram, line 22: I think we should rename this themeService now

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/ReleaseService.cs
```diff
@@ -73,9 +73,11 @@ public async Task<Release> Get(Guid id)
         public async Task<IEnumerable<Release>> List(IEnumerable<Guid> ids)
         {
             return await _contentDbContext.Releases
+                .AsQueryable()
```
benoutram, line 76: I'm not sure that I follow any of these changes to Publisher.Services.ReleaseService. Are they improving performance? Thanks.

benoutram:
Seems great to finally align both into one and leave the Data API responsible for statistics 🙂

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/ReleaseService.cs
ntsim: We need to add these in `Publisher` as it depends on `Content.Services`. As I've added the `System.Linq.Async` package, this causes ambiguity in the compiler over which extension methods to use as `DbSet` implements both `IAsyncEnumerable` and `IQueryable`.

Consequently, we need to add `AsQueryable` to tell the compiler we're working with `IQueryable` specifically, and not an `IAsyncEnumerable`.

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/ThemeControllerTests.cs
ntsim: Good catch! Wish the IDE refactoring would do this automatically!

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/ReleaseService.cs
benoutram: Thanks, I understand now! 🙂

benoutram (approved):

PR #2807, opened 2021-08-23
EES-2642 - update CI build pipeline to use latest LTS version of Ubuntu

luke-h1:


azure-pipelines.dfe.yml
```diff
@@ -103,7 +103,7 @@ jobs:
 
   - job: 'Admin'
     pool:
-      vmImage: 'ubuntu-16.04'
```
markhiveit, line 106: Is there no 20.04 image? May as well use the latest LTS version if its available.

markhiveit (changes requested):

azure-pipelines.dfe.yml
markhiveit: There is a 20.04 image - probably should use that: https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml

Unless there is some reason I'm unaware of?

markhiveit (approved):
This will change the virtual image used on the Build pipeline. If you haven't already, we can make the same change manually on the Release pipeline.

PR #2811, opened 2021-08-24
EES-2553 Publication owners should be able to manage legacy releases

markhiveit:
This PR allows publication owners to manage legacy releases.

Now that users manage legacy releases on the Manage publication page, besides adding permissions for publication owners, the permissions have been aligned with the rest of the manage publication page.

Since the HandleRequirementAsync override for the manage legacy releases authorization handlers takes a LegacyRelease object, not a Release or Publication, I've created a new method to test all different publication roles for that.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Security/AuthorizationHandlers/Utils/PublicationAuthorizationHandlersTestUtil.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Security/AuthorizationHandlers/Utils/PublicationAuthorizationHandlersTestUtil.cs
@@ -1,12 +1,15 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using GovUk.Education.ExploreEducationStatistics.Common.Services.Security;
 using GovUk.Education.ExploreEducationStatistics.Content.Model;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
 using Microsoft.AspNetCore.Authorization;
+using Xunit;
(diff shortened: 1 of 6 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/LegacyReleaseAuthorizationHandlers.cs
```diff
@@ -1,89 +1,143 @@
+using System.Threading.Tasks;
```
benoutram, line 1: I think you can add `#nullable enable` here with no extra effort

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Security/AuthorizationHandlers/Utils/PublicationAuthorizationHandlersTestUtil.cs
```diff
@@ -117,12 +120,83 @@ public static class PublicationAuthorizationHandlersTestUtil
             });
         }
         
+        public class PublicationHandlerTestScenario : HandlerTestScenario
```
benoutram, line 123: Might be worth looking at the way we check publication roles elsewhere, .e.g. in `CreateMethodologyForSpecificPublicationAuthorizationHandlerTests` where  we use a mock for the `PublicationRoleRepository`, and we can verify interactions with it and no need to set up any test data except setting up the behaviour of `IsUserPublicationOwner` which can only be true or false.

Just wondering if you spotted this but dismissed it before creating this approach?

markhiveit: Discussed this on slack. I am going to leave it as it is for now.

PR #2813, opened 2021-08-25
EES-2648 show correct error when no subject meta returned in table response

amyb-hiveit:
Fixes bug where an error is shown in the table tool when there is no table response. Instead the correct message  'No data available for the options selected. Please try again with different options.' is shown.


luke-h1 (approved):
Looks good to me. Tested locally and seems ok

duncan-at-hiveit (approved):

PR #2820, opened 2021-08-26
EES-2555 location type in map dropdown label

amyb-hiveit:
Updates the 'Select a location' dropdown label on maps to show the specific level of location, eg 'Select a Local Authority'. If more than one level is in the dataset then 'Select a location' is shown.

location

### DIFF
--- a/src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
+++ b/src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
@@ -30,6 +30,7 @@ import formatPretty, {
   defaultMaxDecimalPlaces,
 } from '@common/utils/number/formatPretty';
 import { roundDownToNearest } from '@common/utils/number/roundNearest';
+import { LocationFilter } from '@common/modules/table-tool/types/filters';
 import classNames from 'classnames';
 import { Feature, FeatureCollection, Geometry } from 'geojson';
 import { Layer, Path, PathOptions, Polyline } from 'leaflet';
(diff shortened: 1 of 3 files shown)


markhiveit (dismissed):

src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
```diff
@@ -575,7 +604,7 @@ export const MapBlockInternal = ({
             <FormSelect
               name="selectedLocation"
               id={`${id}-selectedLocation`}
-              label="2. Select a location"
+              label={`2. Select a ${locationType}`}
```
duncan-at-hiveit, line 607: We should probably switch between "a" and "an" depending on whether or not the locationType starts with a vowel or not

duncan-at-hiveit (changes requested):

src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
```diff
@@ -360,6 +362,85 @@ export const MapBlockInternal = ({
...
 
+  const locationType = useMemo(() => {
+    const locationLevelsMap: Dictionary<Dictionary<string>> = {
+      country: {
+        label: 'Country',
+        prefix: 'a',
+      },
+      englishDevolvedArea: {
+        label: 'English Devolved Area',
+        prefix: 'an',
```
ntsim, line 373: Not to say that we should - but maybe we could just drop the a/an completely. Seems to me like 'Select English Devolved Area' (for example) sounds fine.

ntsim (approved):

duncan-at-hiveit (approved):

PR #2826, opened 2021-08-27
EES-2417 Add ability to use the table heading as the chart title

markhiveit:
This PR allows users to set the chart title to default to using the table heading. Alternatively, users can enter a custom title for a data block's chart.

image

The backend now expects no chart title to be in a request when the chart title should default to the table heading. This can be seen in `ContentBlock.cs`.

But the backend does return the table heading as the chart title in response to a request for a data block. Because of this, the frontend works out whether the data block response is defaulting to the table heading by comparing the table heading with the chart title. If they're equal, it assumes the chart title is inheriting the table title.

An additional wrinkle was that when changing a data block's heading, because the old data block heading was returned as the chart title when saving the new data block heading - i.e. it was also setting the old data block heading as a custom chart title. To counter this, we check whether the old table heading and old chart title are the same, and if they are we omit the chart title when changing a table heading. This can be seen in `DataBlockPageTabs.tsx`.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataBlockServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataBlockServiceTests.cs
@@ -110,6 +110,51 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
         }
 
         [Fact]
+        public async Task Get_ChartWithoutTitleReturnsHeading()
+        {
+            var dataBlock = new DataBlock
+            {
+                Heading = "Test heading",
+                Charts = new List<IChart>
+                {
+                    new LineChart
+                    {
+                        // No title
+                        Height = 400,
+                        Width = 500,
+                    }
+                },
+            };
+
+            var contextId = Guid.NewGuid().ToString();
+            await using (var context = ContentDbUtils.InMemoryContentDbContext(contextId))
+            {
+                await context.AddAsync(
+                    new ReleaseContentBlock
+                    {
+                        Release = new Release(),
+                        ContentBlock = dataBlock,
+                    }
+                );
+                await context.SaveChangesAsync();
+            }
+
+            await using (var context = ContentDbUtils.InMemoryContentDbContext(contextId))
+            {
+                var service = BuildDataBlockService(context);
+                var result = await service.Get(dataBlock.Id);
+
+                var viewModel = result.AssertRight();
+
+                Assert.Equal(dataBlock.Heading, viewModel.Heading);
+                Assert.Equal(dataBlock.Charts, viewModel.Charts);
+
+                Assert.Single(viewModel.Charts);
+                Assert.Equal(dataBlock.Heading, viewModel.Charts[0].Title);
+            }
+        }
+
+        [Fact]
         public async Task Get_NotFound()
         {
             var contextId = Guid.NewGuid().ToString();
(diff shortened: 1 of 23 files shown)


src/explore-education-statistics-admin/src/pages/release/datablocks/components/DataBlockPageTabs.tsx
```diff
@@ -171,14 +171,18 @@ const DataBlockPageTabs = ({
         const majorAxis = draft[0]?.axes?.major;
         const legend = draft[0]?.legend;
 
-        // Remove data sets that are no longer applicable to a given table's subject meta.
+        // If old chart title is the same as the old table title, then the chart title is defaulting to
```
duncan-at-hiveit, line 174: Do we need a migration to look at existing Chart titles versus their table titles and unset them if they're the same.

We did this for Methodologies adopting their owning Publications' titles.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/DataBlockPageTabs.tsx
```diff
@@ -171,14 +171,18 @@ const DataBlockPageTabs = ({
...
+        // the table title, so don't inadvertently set the old chart/table title
+        if (dataBlock?.charts[0]?.title === dataBlock?.heading) {
+          draft[0].title = undefined;
```
duncan-at-hiveit, line 177: I got an exception at this point while testing on this branch. Mind taking a look into it plz?  This is after trying to create a Chart for a new data block (for a pre-existing Subject).

```
Unhandled Rejection (TypeError): Cannot set property 'title' of undefined

 174 | // If old chart title is the same as the old table title, then the chart title is defaulting to
  175 | // the table title, so don't inadvertently set the old chart/table title
  176 | if (dataBlock?.charts[0]?.title === dataBlock?.heading) {
> 177 |   draft[0].title = undefined;
      | ^  178 | }
  179 | 
  180 | // Remove data sets that are no longer applicable to a given table's subject meta.
```

duncan-at-hiveit (changes requested):

src/explore-education-statistics-admin/src/pages/release/datablocks/components/DataBlockPageTabs.tsx
markhiveit: We shouldn't need to. As far as the frontend is concerned, when getting a chart's title, it gets the same response when a chart title is null in the DB and when the chart title is the same as the table heading in the DB. When the frontend changes a data blocks details, it doesn't set a chart title if the chart title is the same.

So older charts that have a table heading and chart title that are the same both set in the DB will get coerced into a table heading and null chart title in the DB whenever someone next updates that data blocks in the frontend.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataBlockServiceTests.cs
```diff
@@ -706,22 +751,20 @@ await using (var context = ContentDbUtils.InMemoryContentDbContext(contextId))
                 var service = BuildDataBlockService(context);
                 var result = await service.Create(release.Id, createRequest);
 
-                Assert.True(result.IsRight);
+                var viewModel = result.AssertRight();
```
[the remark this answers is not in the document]
markhiveit: The vim skills are starting to make this kind of thing quick to do :)

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartConfiguration.tsx
```diff
@@ -195,13 +210,26 @@ const ChartConfiguration = ({
...
+              options={[
+                {
+                  label: 'Use table title',
+                  value: 'default',
+                },
+                {
+                  label: 'Set an alternative title',
+                  value: 'alternative',
+                  conditional: (
+                    <FormFieldTextInput<FormValues>
```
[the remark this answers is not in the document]
markhiveit: I missed it. I've readded it.

markhiveit:
I checked if any UI tests needed fixing - it seems not.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/ContentBlock.cs
```diff
@@ -98,7 +100,27 @@ public DataBlock()
...
+                        chart.Title = Heading;
+                    }
+                    return chart;
+                }).ToList();
+
+            }
+            set => ChartsInternal = value;
+        }
+
+        [JsonProperty("Charts")]
```
duncan-at-hiveit, line 121: So this gets included in JSON even though it's private?  I would have thought that the one you actually wanted output in JSON would be "Charts", as that's the one that's had the Titles filled in for the frontend's benefit?

markhiveit: I'll make the switch - so I think this means that in the database, if a chart inherits a table heading, the chart title won't be null anymore - I guess it doesn't really matter either way, but since the frontend expects the chart title to be explicitly provided regardless of whether it is inherited, I think saving it this way in the database also makes sense.

markhiveit: I'm having a change of heart with this. So after making this change, now when the frontend sends a null chart title (if the chart is inheriting the table heading), in C# memory, the chart title is null, but when this gets serialized into JSON, it goes through the getter and therefore the chart title isn't null but is explicitly set as the table heading.

This difference between what the chart title is in the C# programs memory vs. how it is stored in the DB might be a little confusing - so maybe its best if they're the same to make it simpler to understand.

Thoughts?

duncan-at-hiveit: I think the initial confusion for me was that I assumed that the JSON annotations were for the benefit of converting to JSON for the front end.  But it turns out it's for the benefit of storing it as JSON in the db instead, in a form that is consistent with the in-memory C# representation.  Am I right?

In which case, I think all we need is a note by this JSONProperty annotation to explain that it's for the purposes of storing this object "pure" in the database.  Any idea where we do this out of interest?

markhiveit: Discussed this with Duncan on Slack. I've added a comment that hopefully clarifies why it is done the way its done. Now, the saved chart JSON in the database will be the same as the chart title as it is saved in program memory.

For anyone who happens to read this in the future, the chart JSON is currently stored in the ContentBlocks table, under the DataBlock_Charts column - `Charts` is mapped to this in ContentDbContext.

duncan-at-hiveit (approved):

PR #2828, opened 2021-08-29
EES-2657 Add Content API endpoint to query a publication's releases

ntsim:
# :warning: Depends on #2827 

This PR adds a new Content API endpoint (`/publications/{publicationId}/releases`) that fetches a publication's published releases.

## Related changes

- Added `ViewPublicationAuthorizationHandler` to secure the new endpoint.
- Added new `TryReloadEntity` and `ReloadEntity` extension methods for `DbContext`. These are intended to try to ensure that entities passed to a service method are actually up-to-date and have not been detached from the context or have been deleted from the database. 
  
  This is a common issue whilst passing raw database entities between various service classes and it is recommended that we use when we can't 100% know the state of the entity before it is received by the service e.g. authorization handlers.

## Other changes

- Enabled nullable on `PartialDate` and fixed various issues with it potentially having invalid values if the year component has not been populated.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/DbContextExtensions.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/DbContextExtensions.cs
@@ -1,0 +1,69 @@
+#nullable enable
+using Microsoft.EntityFrameworkCore;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Extensions
+{
+    public static class DbContextExtensions
+    {
+        /// <summary>
+        /// Reload an entity into the context if it has been detached.
+        /// </summary>
+        /// <remarks>
+        /// If it has been deleted, or cannot be reloaded, we return null as the
+        /// entity would be unsafe to work with.
+        /// <para>
+        /// This is most useful in code where we receive an entity from some other
+        /// service (e.g. in authorization handlers). As we cannot be sure about the
+        /// tracking state of the entity, it is recommended to call this method to
+        /// ensure that it is not stale and has not been deleted.
+        /// </para>
+        /// </remarks>
+        public static T? ReloadEntity<T>(this DbContext context, T entity) where T : class
+        {
+            var entry = context.Entry(entity);
+
+            // The entity has been detached from the context
+            // and may be stale, so we need to reload it for
+            // the most up-to-date version.
+            if (entry.State == EntityState.Detached)
+            {
+                entry.Reload();
+            }
+
+            if (entry.State == EntityState.Deleted)
+            {
+                return null;
+            }
+
+            // If entry is still detached, we can only assume
+            // it does not exist in the database anymore and
+            // is now unsafe to work with (it should be null).
+            if (entry.State == EntityState.Detached)
+            {
+                return null;
+            }
+
+            return entity;
+        }
+
+        /// <summary>
+        /// Try to reload an entity, returning false if it cannot be reloaded.
+        /// </summary>
+        public static bool TryReloadEntity<T>(this DbContext context, T entity, out T reloadedEntity)
+            where T : class
+        {
+            reloadedEntity = entity;
+
+            var reloaded = context.ReloadEntity(entity);
+
+            if (reloaded is null)
+            {
+                return false;
+            }
+
+            reloadedEntity = reloaded;
+            return true;
+        }
+
+    }
+}
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/ReleaseTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/ReleaseTests.cs
@@ -44,7 +44,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Model.Tests
         [Fact]
         public void NextReleaseDate_Ok()
         {
-            var releaseDate = new PartialDate {Day = "01"};
+            var releaseDate = new PartialDate {Year = "2021", Month = "1"};
             var release = new Release
             {
                 NextReleaseDate = releaseDate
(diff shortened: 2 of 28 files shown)


src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/DbContextExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// If it has been deleted, or cannot be reloaded, we return null as the
+        /// entity would be unsafe to work with.
+        /// <para>
+        /// This is most useful in code where we receive an entity from some other
+        /// service (e.g. in authorization handlers). As we cannot be sure about the
+        /// tracking state of the entity, it is recommended to call this method to
+        /// ensure that it is not stale and has not been deleted.
+        /// </para>
+        /// </remarks>
+        public static T? ReloadEntity<T>(this DbContext context, T entity) where T : class
```
benoutram, line 21: Cool, I saw a similar need to introduce something like this after an entity loaded elsewhere with no tracking was then used in an auth handler which failed with an exception when trying to explicitly load related data. I reworked the auth handler and it was no longer an issue but this would have resolved it!

src/GovUk.Education.ExploreEducationStatistics.Publisher.Model/ViewModels/ReleaseTypeViewModel.cs
```diff
@@ -1,11 +1,13 @@
-﻿using System;
+﻿#nullable enable
+using System;
 
 namespace GovUk.Education.ExploreEducationStatistics.Publisher.Model.ViewModels
 {
-    public class ReleaseTypeViewModel
+    public record ReleaseTypeViewModel
```
benoutram, line 6: Just letting you know we discussed this yesterday as there's a need to introduce a new ReleaseType soon, and said that `ReleaseType` could be replaced with an enum of types, since there a very few and new ones are rarely introduced. So this could potentially be dropped in future to be replaced with a label for the type.

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/ReleaseTests.cs
```diff
@@ -44,7 +44,7 @@ public void Live_False_PublishedInFuture()
         [Fact]
         public void NextReleaseDate_Ok()
         {
-            var releaseDate = new PartialDate {Day = "01"};
+            var releaseDate = new PartialDate {Year = "2021", Day = "01"};
```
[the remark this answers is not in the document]
ntsim: Fixed - changed this to use a year and month

src/GovUk.Education.ExploreEducationStatistics.Content.Security/AuthorizationHandlers/ViewReleaseAuthorizationHandler.cs
```diff
@@ -26,17 +28,18 @@ public ViewReleaseAuthorizationHandler(ContentDbContext context)
...
-            if (hydratedRelease == null)
+            if (!_context.TryReloadEntity(release, out var loadedRelease))
             {
                 return;
             }
 
-            if (hydratedRelease.IsLatestPublishedVersionOfRelease())
+            await _context.Entry(loadedRelease)
+                .Reference(p => p.Publication)
+                .Query()
```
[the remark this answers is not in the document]
ntsim: Nice yeh, definitely looks like we could it there!

src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/DbContextExtensions.cs
ntsim: Yep :smile: I think it'd be good practice to basically put this in any authorization handler from now on, given we can't necessarily ensure the state of the entity before it arrives there.

src/GovUk.Education.ExploreEducationStatistics.Content.Security/AuthorizationHandlers/ViewPublicationAuthorizationHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
+using Microsoft.AspNetCore.Authorization;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Security.AuthorizationHandlers
+{
+    public class ViewPublicationRequirement : IAuthorizationRequirement
+    {
+    }
+
+    public class ViewPublicationAuthorizationHandler
```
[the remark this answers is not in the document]
ntsim: These authorization handlers are in the `Content.Security` package, so I'd say that this provides enough context to know they they're for use in the content API and not the admin.

Personally, that feels good enough, but if you want us to be a little bit more explicit, I can update the handler names to `ViewPublicPublicationAuthorizationHandler` (and similar for releases).

src/GovUk.Education.ExploreEducationStatistics.Publisher.Model/ViewModels/ReleaseTypeViewModel.cs
ntsim: Cool cool :+1: always thought it was a little bit unwieldy in its current form. Are we thinking of dropping the database table in favour of storing it as enum strings?

benoutram: Yes that's what we chatted about, as they change so rarely. This didn't have anything much to do with your PR but thought I'd mention it here when I saw it! 🙂

src/GovUk.Education.ExploreEducationStatistics.Content.Security/AuthorizationHandlers/ViewPublicationAuthorizationHandler.cs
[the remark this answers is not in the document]
benoutram: Ah yep, I think the shorter names are ok then as like you say they are already in the `Content.Security` package which should never be included by the Admin 🙂 .

benoutram (approved):

PR #2830, opened 2021-09-01
EES-2650-2 - disable `methodlogy_page_absence` UI test

luke-h1:
This PR: 
- disables the `methodlogy_page_absence` UI test from running against pre-prod & prod environments. This is due to the test data not being present in both environments.

### DIFF
--- a/tests/robot-tests/tests/general_public/methodology_page_absence.robot
+++ b/tests/robot-tests/tests/general_public/methodology_page_absence.robot
@@ -4,7 +4,7 @@ Resource            ../libs/public-common.robot
 Suite Setup         user opens the browser
 Suite Teardown      user closes the browser
 
-Force Tags          GeneralPublic    Local    Dev    Test    Preprod    NotAgainstDev
+Force Tags          GeneralPublic    Local    Test
 
 *** Test Cases ***
 Navigate to Pupil absence in schools in England methodology page


tests/robot-tests/tests/general_public/methodology_page_absence.robot
```diff
@@ -4,7 +4,7 @@ Resource            ../libs/public-common.robot
 Suite Setup         user opens the browser
 Suite Teardown      user closes the browser
 
-Force Tags          GeneralPublic    Local    Dev    Test    Preprod    NotAgainstDev
+Force Tags          GeneralPublic    Local    Dev    Test    Preprod    NotAgainstDev    NotAgainstPreProd    NotAgainstProd
```
markhiveit, line 7: Rather than adding `NotAgainstPreProd` tag, you can just remove the `Preprod` tag. The same goes for `NotAgainstDev` - can just remove the `Dev` tag.

I think the `NotAgainstProd` tag is redundant because it looks like nothing in this test suite is set to run against `Prod`, so that can also be removed.

The `NotAgainst` tags are designed to be used against specific test cases. If you don't want to run a particular test suite against an environment, you only need to not apply a tag for that environment - `Dev` / `Test` / etc.

So all you need here is
`Force Tags     GeneralPublic  Local  Test`

markhiveit: This is probably another ticket, but if this test suite is only being run locally and against test, it seems possible a problem with existing methodology pages could be missed. 

I'd create a ticket to create a methodology page as part of the bootstrapped data, add that data to Dev / Test / PreProd, then create a new test suite for it. `methodology_page_absence.robot` could then be removed.

markhiveit (changes requested):

markhiveit (approved):

PR #2832, opened 2021-09-01
EES-2659 - update statistics page UI test

luke-h1:


markhiveit (changes requested):

tests/robot-tests/tests/general_public/statistics_page.robot
```diff
@@ -4,18 +4,164 @@ Resource            ../libs/public-common.robot
...
+
+Open and validate "UK education and training statistics" accordion section (prod)
+    [Tags]    NotAgainstLocal    NotAgainstDev    NotAgainstPreProd
+    user opens accordion section    UK education and training statistics
+    user waits until accordion section contains text    UK education and training statistics
+    ...    UK education and training statistics
+
+# below onwards assertions targets local and dev environments
+
+Navigate to Find Statistics page (local, dev)
```
markhiveit, line 156: You could have this test case run against all environments, rather than repeating it twice.

My general thought for how this file might be is something like this:

```
Force Tags Prod
*** Test Cases ***
Navigate to Find stats page
[Tags]  Local Dev   # Prod tag is applied by the Force Tags so no need to add it here
....
Check bootstrapped data
[Tags] Local Dev NotAgainstProd
...
First test case to run against prod
# no tags required
...
```
etc.

I would personally veer towards having one test case that does a quick sanity check against Local / Dev - I think it could just check the bootstrapped data, which shouldn't change - if it is too thorough then changing test data will cause failures moving forward.

markhiveit (changes requested):

tests/robot-tests/tests/general_public/statistics_page.robot
```diff
@@ -4,20 +4,40 @@ Resource            ../libs/public-common.robot
 Suite Setup         user opens the browser
 Suite Teardown      user closes the browser
 
-Force Tags          GeneralPublic    Local    Dev    Test    Preprod
+Force Tags          GeneralPublic    Prod
+#    Local    Dev
```
markhiveit, line 8: This comment can be removed.

markhiveit (changes requested):

markhiveit (approved):

PR #2835, opened 2021-09-02
EES 2661 - update `methodologies_page.robot` UI test 

luke-h1:


tests/robot-tests/tests/general_public/methodologies_page.robot
```diff
@@ -30,6 +30,27 @@ Validate page contents
     ...    Pupil absence statistics: methodology
     ...    %{PUBLIC_URL}/methodology/pupil-absence-in-schools-in-england
 
+Navigate to /methodology page (prod)
```
markhiveit, line 33: There is no need to repeat this test case - you can have it run once against the environments you want, since this doesn't depend on test data and should be the same across environments.

tests/robot-tests/tests/general_public/methodologies_page.robot
```diff
@@ -1,13 +1,13 @@
 *** Settings ***
 Resource            ../libs/public-common.robot
-
-Force Tags          GeneralPublic    Local    Dev    Test    Preprod
+Force Tags          GeneralPublic    Local    Dev    Test    Preprod    Prod
```
markhiveit, line 3: Rather than applying keywords to run this against every environment here, I think I'd force tags to be prod only, and then use Local / Dev / Preprod / NotAgainstProd tags against specific test cases depending on where you want them to run. That should mean you have to use less tags against specific test cases.

tests/robot-tests/tests/general_public/methodologies_page.robot
```diff
@@ -30,6 +30,27 @@ Validate page contents
...
+    user waits until page contains accordion section    Early years
+    user waits until page contains accordion section    Finance and funding
+    user waits until page contains accordion section    Further education
+    user waits until page contains accordion section    Higher education
+    user waits until page contains accordion section    Pupils and schools
+    user waits until page contains accordion section    School and college outcomes and performance
+    user waits until page contains accordion section    Teachers and school workforce
+    user waits until page contains accordion section    UK education and training statistics
+
 Validate Related information section links exist
```
markhiveit, line 54: As this file stands, this is the only test case that will run against Dev - but I think this test case might still pass because `user opens browser` leaves the user on the home page, and I think these links can be found there too?

You could check you're on the right page with `user checks url contains`.

markhiveit (changes requested):

tests/robot-tests/tests/general_public/methodologies_page.robot
```diff
@@ -1,44 +1,40 @@
...
-Force Tags          GeneralPublic    Local    Dev    Test    Preprod    Prod
+Force Tags          GeneralPublic    Prod
 
 Suite Setup         user opens the browser
 Suite Teardown      user closes the browser
 
 *** Test Cases ***
 Navigate to /methodology page
-    [Tags]    NotAgainstDev    NotAgainstPreProd    NotAgainstProd
+    [Tags]    Local    Prod
```
markhiveit, line 10: The `Prod` tag here is redundant - so it can be removed on this test case and any other test case it appears on. 

As the `Prod` tag is in `Force Tags` it is applied to every single test case. So any tag included in `Force Tags` doesn't need to appear on any test case.

tests/robot-tests/tests/general_public/methodologies_page.robot
```diff
@@ -1,44 +1,40 @@
...
     user navigates to public methodologies page
     user waits until page contains
     ...    Browse to find out about the methodology behind specific education statistics and data and how and why they're collected and published.
     ...    10
 
 Validate page contents
-    [Tags]    NotAgainstDev    NotAgainstPreProd    NotAgainstProd
+    [Tags]    Local
```
markhiveit, line 17: I think you're expecting this test case to only run against `Local`, but it will run against `Prod` as well as the prod tag is applied to this test case as it is included in `Force Tags`.

So if the intention is to not run this test case against prod, you need to add a `NotAgainstProd` tag.

markhiveit (changes requested):

markhiveit (changes requested):

tests/robot-tests/tests/general_public/methodologies_page.robot
```diff
@@ -1,35 +1,57 @@
...
+    user waits until page contains accordion section    Early years
+    user waits until page contains accordion section    Finance and funding
+    user waits until page contains accordion section    Further education
+    user waits until page contains accordion section    Higher education
+    user waits until page contains accordion section    Pupils and schools
+    user waits until page contains accordion section    School and college outcomes and performance
+    user waits until page contains accordion section    Teachers and school workforce
+    user waits until page contains accordion section    UK education and training statistics
+
+Validate bootstraped methodologies exist
```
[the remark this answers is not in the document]
markhiveit: You still need to update the name of this test case, since it isn't checking a bootstrapped methodology.

I don't think we have any bootstrapped methodology at the moment, and since this keyword is looking for Pupils and schools in `Validate page contents` which runs locally, I think this test case can just be removed.

markhiveit (changes requested):

markhiveit (approved):

PR #2836, opened 2021-09-03
EES-2658 Add file info to Data API's subjects endpoints

ntsim:
This PR primarily adds `FileInfo` to various subject endpoints so that we can re-use this in the data catalogue page on Step 3 (where the user selects their desired data downloads).

## Related changes

- Updated the frontend with the appropriate changes to integrate with the new API updates.
- Split out subjects and featured tables (previously referenced as table highlights) into separate endpoints: `/subjects` and `/featured-tables`, for a publication or release. Previously, these were combined into a `SubjectsAndHighlightsViewModel` which wasn't particularly RESTful and also meant that we couldn't retrieve the subjects or featured tables/highlights independently of one another.

## Test changes

- As noted in Slack, we've made some additional changes to how we write tests in the `ReleaseDataBlockEditPage` suite. This is primarily needed to avoid nasty test runner lockup due to over-saturation of `waitFor` calls by extremely slow `getByRole` calls (further reading in this [GitHub issue](https://github.com/testing-library/dom-testing-library/issues/820)). We have now refactored this to use much faster `getByText` calls first, before asserting on `getByRole` when not in the `waitFor` block e.g:
  ```ts
  await waitFor(() => {
    expect(screen.getByRole('dialog')).toBeInTheDocument();
  });
  ```
  Is now:
  ```ts  
  await waitFor(() => {
    expect(screen.getByText('Delete data block')).toBeInTheDocument();
  });
  ```
  We should aim to do similar things in other suites which test entire page components as these will be most affected by the slowness of `getByRole`.


src/GovUk.Education.ExploreEducationStatistics.Data.Services/ReleaseService.cs
```diff
@@ -137,14 +128,64 @@ private async Task<List<TableHighlightViewModel>> GetHighlights(Guid releaseId,
...
+        private IQueryable<ReleaseFile> QueryReleaseDataFiles(Guid releaseId)
+        {
+            return _contentDbContext.ReleaseFiles
+                .Include(rf => rf.File)
+                .Where(
+                    rf => rf.ReleaseId == releaseId
+                          && rf.File.Type == FileType.Data
+                          // Exclude files that are replacements in progress
+                          && !rf.File.ReplacingId.HasValue
+                          && rf.File.SubjectId.HasValue
```
benoutram, line 168: It's a shame we have all file types as `File` and can't use inheritance nicely to have a DataFile that has a non-null SubjectId. I think we could do it now but will be limited by EF to use TPH where all the types are still mapped to the same table. SubjectId would still be nullable in the database.

Maybe when we upgrade to EF6 we should refactor File to use TPT table inheritance and change this, removing the need for some of the other stuff around it like unit tests to check it's filtering Files with null Subject Id's.

We could also think about whether two databases and the `Subject` terminology still makes sense or if Observations can be linked direct to DataFile 🙂

benoutram (approved):
I think I follow everything. Seems like a great change to get Subjects and Featured Tables split out, and clearly mark the technical debt around building the FileInfo.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ReleaseService.cs
ntsim: Yep agreed. Would definitely be nice to get rid of all the implicit rules and checks we currently have to do.

> We could also think about whether two databases and the Subject terminology still makes sense or if Observations can be linked direct to DataFile

Also a good idea! Would be good if we could finally get rid of all the release/subject related stuff, although not sure how feasible that'll end up being :smile:

PR #2837, opened 2021-09-03
EES-2284 Return MethodologySummaryViewModel  as result of getting the adoptable methodologies

benoutram:
This PR changes the response of the URL `GET /api/publication/{publicationId}/adoptable-methodologies` to return a list of `MethodologySummaryViewModel` resources rather than `TitleAndIdViewModel`.

This is so that future work on the UI to implement managing adoption of methodologies can show additional details about them such as their status, publication, published date, amendment flag, etc.

It also refactors the various methods of `MethodologyService` that now return a `MethodologySummaryViewModel` to use a common method to build the view model.

The unit tests are enhanced around checking the fields of a `MethodologySummaryViewModel` result to ensure the various fields of it are being mapped correctly.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
@@ -10,6 +10,7 @@ using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.Secur
 using GovUk.Education.ExploreEducationStatistics.Admin.ViewModels;
 using GovUk.Education.ExploreEducationStatistics.Admin.ViewModels.Methodology;
 using GovUk.Education.ExploreEducationStatistics.Common.Cache;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Model;
 using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces.Security;
@@ -19,7 +20,6 @@ using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Repository.Interfaces;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.EntityFrameworkCore;
-using Microsoft.EntityFrameworkCore.Query;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Validators.ValidationErrorMessages;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Validators.ValidationUtils;
 using static GovUk.Education.ExploreEducationStatistics.Common.Model.FileType;
@@ -103,14 +103,14 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services.Methodologie
                 .OnSuccess(() => _methodologyRepository
                     .CreateMethodologyForPublication(publicationId, _userService.GetUserId())
                 )
-                .OnSuccess(_mapper.Map<MethodologySummaryViewModel>);
+                .OnSuccess(BuildMethodologySummaryViewModel);
         }
 
         public async Task<Either<ActionResult, Unit>> DropMethodology(Guid publicationId, Guid methodologyId)
         {
             return await _persistenceHelper
                 .CheckEntityExists<PublicationMethodology>(q =>
-                    q.Where(link => link.PublicationId == publicationId 
+                    q.Where(link => link.PublicationId == publicationId
                                     && link.MethodologyParentId == methodologyId))
                 .OnSuccess(link => _userService.CheckCanDropMethodologyLink(link))
                 .OnSuccessVoid(async link =>
@@ -120,7 +120,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services.Methodologie
                 });
         }
 
-        public async Task<Either<ActionResult, List<TitleAndIdViewModel>>> GetAdoptableMethodologies(Guid publicationId)
+        public async Task<Either<ActionResult, List<MethodologySummaryViewModel>>> GetAdoptableMethodologies(
+            Guid publicationId)
         {
             return await _persistenceHelper
                 .CheckEntityExists<Publication>(publicationId)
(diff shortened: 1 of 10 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -120,61 +120,27 @@ public class MethodologyService : IMethodologyService
...
             return await _persistenceHelper
                 .CheckEntityExists<Publication>(publicationId)
                 .OnSuccess(_userService.CheckCanAdoptMethodologyForPublication)
                 .OnSuccess(async publication =>
                 {
                     var methodologies = await _methodologyParentRepository.GetUnrelatedToPublication(publication.Id);
-                    return methodologies
-                        .Select(m => new TitleAndIdViewModel(m.Id, m.OwningPublicationTitle))
-                        .ToList();
+                    var latestVersions = await methodologies.SelectAsync(methodology =>
```
ntsim, line 132: Ugh, this reminds me that we should probably add the `System.Linq.Async` package into admin at some point

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
benoutram: Ah yep, I actually started doing this as I added it to the `Content.Model` package to use it there and then started updating all the dependant projects using `Content.Model` which turned into a much bigger change and abandoned it at that point! I think it makes sense to do it as a separate piece of work.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -211,12 +177,62 @@ await _context.Entry(methodology.ScheduledWithRelease)
...
                 .OnSuccess(methodology => UpdateMethodologyStatus(methodology, request))
                 .OnSuccess(methodology => UpdateMethodologyDetails(methodology, request))
                 .OnSuccess(_ => GetSummary(id));
         }
 
+        private async Task<MethodologySummaryViewModel> BuildMethodologySummaryViewModel(Methodology methodology)
+        {
+            if (!_context.TryReloadEntity(methodology, out var loadedMethodology))
+            {
+                throw new InvalidOperationException($"Unable to reload entity.");
```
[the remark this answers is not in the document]
benoutram: Thanks, I've added it in now 🙂

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -211,12 +177,62 @@ await _context.Entry(methodology.ScheduledWithRelease)
...
+
+            if (loadedMethodology.ScheduledForPublishingWithRelease)
+            {
+                await _context.Entry(loadedMethodology)
+                    .Reference(m => m.ScheduledWithRelease)
+                    .LoadAsync();
+
+                await _context.Entry(loadedMethodology.ScheduledWithRelease)
+                    .Reference(r => r!.Publication)
+                    .LoadAsync();
```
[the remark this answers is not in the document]
benoutram: Thanks, I've moved it within that not-null check further down now 🙂 Theoretically it should be loaded every time (unless already loaded) as it's already within another block that's only true if the publishing strategy is 'WithRelease'.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/MethodologyRepository.cs
```diff
@@ -54,10 +54,21 @@ public async Task<Methodology> CreateMethodologyForPublication(Guid publicationI
...
         }
 
+        public async Task<Methodology> GetLatestByMethodologyParent(Guid methodologyParentId)
+        {
+            var methodology = await _contentDbContext.MethodologyParents
+                .Include(m => m.Versions)
+                .SingleAsync(mp => mp.Id == methodologyParentId);
+
+            var versions = methodology.Versions;
+            return versions.Single(v => versions.All(inner => inner.PreviousVersionId != v.Id));
```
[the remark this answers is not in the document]
benoutram: That is a very good point 😄. I've swapped it for this now, thanks.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -211,12 +177,62 @@ await _context.Entry(methodology.ScheduledWithRelease)
...
+                .CheckEntityExists<Methodology>(id, q =>
+                    q.Include(m => m.MethodologyParent))
                 .OnSuccess(methodology => UpdateMethodologyStatus(methodology, request))
                 .OnSuccess(methodology => UpdateMethodologyDetails(methodology, request))
                 .OnSuccess(_ => GetSummary(id));
         }
 
+        private async Task<MethodologySummaryViewModel> BuildMethodologySummaryViewModel(Methodology methodology)
+        {
+            if (!_context.TryReloadEntity(methodology, out var loadedMethodology))
```
duncan-at-hiveit, line 189: Hey @benoutram - presumably this is to reload a Methodology after an update or something in order to have the latest and greatest details?  Just need to understand a bit more why we've got this here, as I've not seen us doing this elsewhere?

benoutram: Hi @duncan-at-hiveit ,

I noticed that we've recently started adding code to explicitly load properties closer in terms of lines to where they are used in an attempt to make it a little bit clearer why they are being included, rather than eagerly loading them with a `.Include` in code that can sometimes be a few calls away from where they are needed.

However, if you try and explicitly load properties of an entity that is not attached to the database context then you get a nasty run-time error along the lines of "attempt was made to load navigation property on detached entity".

One example way of reproducing this would be to write an auth handler that explicitly loads a database property on the entity being tested in an auth context, but call that auth handler with an entity loaded from the database using `.AsNoTracking`. Since the entity is not attached to the database context the auth handler will fail at runtime. The `.AsNoTracking` might have been added to try and improve performance, but with unintended consequences. See `ViewPublicationAuthorizationHandler` and `ViewReleaseAuthorizationHandler` which are now using this reloading technique to ensure the entities are not detached to avoid this kind of scenario.

Within the method changing here to build the view model we need to make sure the Publications are hydrated within the Methodology as the view model uses them.

The Methodology passed into this private method sometimes comes from the persistence helper, but sometimes comes from the `MethodologyRepository` dependency of this service.

From the persistence helper, providing that `.AsNoTracking` hasn't been used, it will be fine, but we have less control over a Methodology returned by the `MethodologyRepository`.

I would expect the Methodology returned by `MethodologyRepository` to be fine too, but we see a problem in the tests where this becomes useful to add in.

In the tests for methods that use, e.g. `[Fact] public async Task GetAdoptableMethodologies()`  a Methodology is saved in the database, but the service is run with a fresh database context where the Methodology is returned by a mock. Since it's not attached to the database context, but is present in the database, the `TryReloadEntity` will load it and the Publications can be explicitly loaded.

Another way of solving this would be to use `context.Attach(methodology)` just prior to calling the service in the test but I thought this was a good use of the `TryReloadEntity` now that we have it.

Happy to remove this and use `context.Attach` in the unit tests if you think that would be better. 🙂

duncan-at-hiveit: Hey Ben - fine with it this way, but I'm personally not a fan of the way it looks, so having it scattered around the codebase, it might be a bit ugly as we use it more often.  I wonder if the next time we use it we could consider wrapping it in an extension method of our own that would do the check and throw a consistent, useful error message if it could not be reloaded, and all as a one-liner?

duncan-at-hiveit: Or something that returns an Either...

benoutram: I agree, I think we need a variation of `ReloadEntity` and `TryReloadEntity` that ensures it's loaded without returning an entity that can be null.

ntsim: `AssertEntityLoaded`?

benoutram: Just added this in @ntsim @duncan-at-hiveit 

```
        /// <summary>
        /// Ensure that an entity is not detached from the database context, reloading it if necessary before returning.
        /// Throws an exception if the entity cannot be reloaded.
        /// </summary>
        public static T AssertEntityLoaded<T>(this DbContext context, T entity) where T : class
        {
            var reloaded = context.ReloadEntity(entity);

            if (reloaded is null)
            {
                var displayString = context.GetDisplayString(entity);
                throw new InvalidOperationException($"Unable to reload {displayString}");
            }

            return reloaded;
        }
```

```
        private static string GetDisplayString(this DbContext context, object entity)
        {
            var entry = context.Entry(entity);
            var primaryKey = entry.Metadata.FindPrimaryKey();
            var typeName = entity.GetType().ShortDisplayName();
            var primaryKeyValues = primaryKey.Properties
                .Select(property => $"{property.Name}: '{property.PropertyInfo.GetValue(entity)}'")
                .JoinToString(", ");

            return $"{typeName} {{{primaryKeyValues}}}";
        }
```

which lets me replace the ugly code with:

`var loadedMethodology = _context.AssertEntityLoaded(methodology);`

ntsim: Very nice :grin:

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
ntsim: Yep I agree!

ntsim (approved):

PR #2841, opened 2021-09-06
EES-2662

duncan-at-hiveit:
This PR:
- fixes a bug whereby Locations with duplicate codes but different names were not being returned via the API to the table tool

The cause of the bug was a very innocent `Distinct()` call on a list of IObservationalUnits.  The real bug was that the Observational Units in questions implemented Equals() and hashcode methods that only operated on Code, and not the combination of Code and Name.

I've restructured the various Observational Units to extend a common base class that holds Code and Name and the equality members.  The various subclasses are now all much smaller.

I have also removed some JSON annotations on RscRegion, which after speaking with Ben, we believe to be redundant nowadays.  Still, some good testing with all sorts of Location types will be required.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Model/LocalAuthorityDistrict.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Model/LocalAuthorityDistrict.cs
@@ -1,41 +1,13 @@
 namespace GovUk.Education.ExploreEducationStatistics.Data.Model
 {
-    public class LocalAuthorityDistrict: IObservationalUnit
+    public class LocalAuthorityDistrict : ObservationalUnit
     {
-        public string Code { get; set; }
-        public string Name { get; set; }
-
-        private LocalAuthorityDistrict()
+        public LocalAuthorityDistrict()
         {
         }
-
-        public LocalAuthorityDistrict(string code, string name)
+        
+        public LocalAuthorityDistrict(string code, string name) : base(code, name)
         {
-            Code = code;
-            Name = name;
-        }
-
-        public static LocalAuthorityDistrict Empty()
-        {
-            return new LocalAuthorityDistrict(null, null);
-        }
-
-        protected bool Equals(LocalAuthorityDistrict other)
-        {
-            return string.Equals(Code, other.Code);
-        }
-
-        public override bool Equals(object obj)
-        {
-            if (ReferenceEquals(null, obj)) return false;
-            if (ReferenceEquals(this, obj)) return true;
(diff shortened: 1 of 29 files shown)


src/GovUk.Education.ExploreEducationStatistics.Data.Services/Mappings/DataServiceMappingProfiles.cs
```diff
@@ -18,7 +18,7 @@ public DataServiceMappingProfiles()
             CreateMap<Location, LocationViewModel>();
 
             AppDomain.CurrentDomain.GetAssemblies().SelectMany(GetTypesFromAssembly)
-                .Where(p => typeof(IObservationalUnit).IsAssignableFrom(p))
+                .Where(p => typeof(ObservationalUnit).IsAssignableFrom(p))
```
benoutram, line 21: It might be possible to use `.OfType<ObservationalUnit>()` here?

src/GovUk.Education.ExploreEducationStatistics.Data.Model/LocalAuthorityDistrict.cs
```diff
@@ -1,41 +1,18 @@
...
-
-        public LocalAuthorityDistrict(string code, string name)
+        
+        public LocalAuthorityDistrict(string code, string name) : base(code, name)
         {
-            Code = code;
-            Name = name;
         }
-
+
```
benoutram, line 12: This whitespace creeping in made me look into Rider settings for formatting during saving. I was sure Rider had a set of On-Save actions where we could turn on things like formatting but I can't see them now.

benoutram: Edit: I see a feature request, looks like it hasn't been implemented 😭

duncan-at-hiveit: I'm gob-smacked!  Do you manually format them as you go?

benoutram: Ha ha, I use a combination of keyboard shortcuts to reformat code and organise imports. I know these can both be automated in the commit options. I think I've tried these in the past and didn't get on with them. I'll try them out again.

benoutram:
All comments are resolved by a duplicate of this PR https://github.com/dfe-analytical-services/explore-education-statistics/pull/2842 which targets the master branch. Closing this one as we will eventually merge the hotfix commit back to the dev branch.

PR #2842, opened 2021-09-06
Ees 2662 locations with duplicate codes removed from table tool as hotfix

duncan-at-hiveit:
This PR:
- fixes a bug whereby Locations with duplicate codes but different names were not being returned via the API to the table tool

The cause of the bug was a very innocent Distinct() call on a list of IObservationalUnits. The real bug was that the Observational Units in questions implemented Equals() and hashcode methods that only operated on Code, and not the combination of Code and Name.

I've restructured the various Observational Units to extend a common base class that holds Code and Name and the equality members. The various subclasses are now all much smaller.

I have also removed some JSON annotations on RscRegion, which after speaking with Ben, we believe to be redundant nowadays. Still, some good testing with all sorts of Location types will be required.


benoutram (dismissed):
Approving the hotfix. Not sure if you want to follow up on either of the conversations  of https://github.com/dfe-analytical-services/explore-education-statistics/pull/2841 first before merging.

benoutram (approved):

PR #2843, opened 2021-09-07
EES-2620 groupby filter

amyb-hiveit:
Backend changes to save the groupby filter. Plus a couple of frontend fixes.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartAxisConfiguration.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartAxisConfiguration.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿#nullable enable
+using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.Serialization;
 using Newtonsoft.Json;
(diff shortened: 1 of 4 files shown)


markhiveit:
It looks like the groupByFilter dropdown still isn't selecting a previously saved filter - just defaulting to `All filters`. In the below screenshot, you can see in the response that groupByFilter characteristic is returned by the BE, but the frontend shows `All filters`.

image

Also, having saved the chart with groupByFilter characteristic, if I save the chart with `All filters` selected, the FE request still sends groupByFilter characteristic:

image

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartAxisConfiguration.cs
```diff
@@ -8,25 +9,27 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Model.Chart
 {
     public class ChartAxisConfiguration
     {
-        public string Name;
+        public string Name = null!;
```
benoutram, line 12: Just wondering if we should also add the [Required] attribute to indicate that these fields are mandatory. That would hopefully cause a bad request error if the chart axis configuration is requested without a name, sortBy, title, unit etc.

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartAxisConfiguration.cs
markhiveit: I am not sure which fields are or are not required here, so I've created another ticket for this rather than investigate this immediately.

PR #2844, opened 2021-09-07
EES-968 Notifying subscribers for amendment release publication now optional

markhiveit:
This PR adds a `Notify subscribers by email` checkbox when approving an amended release on the Admin Sign off page.

image

Previously, subscribers would be notified every time a release would be published. Now, analysts can choose whether subscribers are notified when publishing an amendment. Subscribers are still always notified when an original release is published.

Edit: 
We've additionally added a `NotifyOn` column to the Content DB ReleaseStatus table. This is to reduce the chance that we send a notification email twice.

It is possible for a ReleaseStatus to have `NotifySubscribers` set to true and the `NotifyOn` column to never get set. This happens if a user schedules a release for publication and then moves it back to draft. The ReleaseStatus associated with the scheduling of the release for publication will never have a notification sent with it.

The publisher always checks the most recent ReleaseStatus for a release when determining whether a notification should be sent. This means that if in the future the possibility of an additional ReleaseStatus being created between the scheduling of a release for publication and its actual publication, notification may be effected.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
@@ -909,6 +909,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                 Assert.Equal(new DateTime(2051, 6, 30, 0, 0, 0, DateTimeKind.Unspecified),
                     viewModel.PublishScheduled);
                 Assert.Equal(nextReleaseDateEdited, viewModel.NextReleaseDate);
+                Assert.False(viewModel.NotifySubscribers);
                 Assert.Equal(adHocReleaseType, viewModel.Type);
                 Assert.Equal("2030", viewModel.ReleaseName);
                 Assert.Equal(TimeIdentifier.March, viewModel.TimePeriodCoverage);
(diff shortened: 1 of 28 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Model/Release.cs
```diff
@@ -66,6 +66,19 @@ public string ReleaseName
 
         public List<ReleaseStatus> ReleaseStatuses { get; set; }
 
+        public bool NotifySubscribers
+        {
+            get
+            {
+                return ReleaseStatuses.Any(rs => rs.ApprovalStatus == ReleaseApprovalStatus.Approved)
```
benoutram, line 73: I'm not sure what this is trying to achieve but it seems like "If there's been any approved status, then return NotifySubscribers of the last approved status, otherwise false".

That doesn't seem correct because a draft Release that's previously been approved with NotifySubscribers as true will return true but NotifySubscribers on the latest ReleaseStatus will be false as we know the service which sets it will force it to be false.

Seems like you might be trying to return the historical status of NotifySubscribers so that the checkbox remains ticked if it's previously been ticked, but that's not clear at all.

I would add some comments to explain what this is doing and add tests to ReleaseTests for this.

If it's simply intended to return the current value of NotifySubscribers then you could rewrite it as:

- Throw error if ReleaseStatuses is not hydrated
- Get the last status which would be LastOrDefault ReleaseStatus when ordered by Created
- Return `last?.NotifySubscribers ?? false`

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
```diff
@@ -1179,6 +1180,282 @@ await using (var context = InMemoryApplicationDbContext(contextId))
...
+                MockUtils.VerifyAllMocks(contentService, releaseFileService);
+
+                var viewModel = result.AssertRight();
+
+                Assert.True(viewModel.NotifySubscribers);
+            }
+        }
+
+        [Fact]
+        public async Task CreateReleaseStatus_Draft_NotifySubscribers()
```
benoutram, line 1412: Feel like this one is already covered by `CreateReleaseStatus` which is testing the draft status update and you've added this assertion to that test already.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Functions/PublishReleaseContentFunction.cs
```diff
@@ -83,7 +89,15 @@ public class PublishReleaseContentFunction
 
                 await _contentService.DeletePreviousVersionsDownloadFiles(message.ReleaseId);
                 await _contentService.DeletePreviousVersionsContent(message.ReleaseId);
-                await _notificationsService.NotifySubscribers(message.ReleaseId);
+
+                var release = _contentDbContext.Releases
+                    .Include(r => r.ReleaseStatuses)
+                    .Single(r => r.Id == message.ReleaseId);
+                if (release.NotifySubscribers)
```
benoutram, line 96: I'd be tempted to do move this down within NotificationService and avoid the extra logic in the function classes with the additional database dependency. This will keep the function cleaner, remaining as a series of service calls.

We only send notifications for subscribers at a Publication level.

NotificationService works out the distinct Publications for the release id's passed to it, so this will change to work out the distinct Publications for Releases where notifications have been requested.

One other benefit of moving it there will be that you can also unit test it more easily.

It will also work exactly the same for `PublishStagedReleaseContentFunction.cs` where there are multiple release id's so no need for two different variations of it.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Functions/PublishStagedReleaseContentFunction.cs
```diff
@@ -103,7 +108,16 @@ public class PublishStagedReleaseContentFunction
 
                     await _contentService.DeletePreviousVersionsDownloadFiles(releaseIds);
                     await _contentService.DeletePreviousVersionsContent(releaseIds);
-                    await _notificationsService.NotifySubscribers(releaseIds);
+
+                    var releaseIdsToNotifySubscribersFor = _contentDbContext.Releases
```
benoutram, line 112: If you take on my other suggestion then this won't be necessary as NotifySubscribers will change to only send notifications if applicable for all the Publications.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Functions/PublishReleaseContentFunction.cs
markhiveit: I've moved the logic into NotificationService as you suggest. Do I also want to create unit tests for NotificationsService now?

Also I am tempted to rename NotificationsService#NotifySubscribers to something like `NotifySubscribersIfAppropriate` just so its clear this method isn't a guarantee that a notification will be sent.

benoutram: Yes I think now is probably a good time to add the unit tests in. I thought the same about the naming when writing this comment. Perhaps `NotifySubscribersIfApplicable`?

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Release.cs
markhiveit: I think I've got the logic wrong around when notifications should be sent. And it seems that the code as it stands has bug potential, as there is no record of whether a notification has been sent. So we've decided to add a `NotifiedOn' column. This tells as whether a notification has been sent and provides protection against sending a notification twice.

markhiveit:
```
I'm not sure what this is trying to achieve but it seems like "If there's been any approved status, then return NotifySubscribers of the last approved status, otherwise false".```

To address this comment, my original thought on checking whether the ApprovalStatus was approved was to ensure that no future bugs were possible where an additional ReleaseStatus row was created between a release being scheduled for publication and actually being published. This could effect whether subscribers are notified, and potentially prevent them from being so because the code checks for the latest ReleaseStatus.

File: src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/Services/NotificationsServiceTests.cs [hunk 5]
```diff
@@ -0,0 +1,222 @@
...
+            }
+        }
+
+        private static NotificationsService BuildNotificationsService(
+            ContentDbContext contentDbContext,
+            IStorageQueueService storageQueueService = null)
+        {
+            return new (
+                contentDbContext,
+                storageQueueService ?? Mock.Of<IStorageQueueService>());
```
benoutram, line 219: Add Strict here to the default mock and then the two tests `NotifyOnSet` and `None` can make use of it without needing to set up or verify anything since they both expect no interactions.

src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/Services/NotificationsServiceTests.cs
```diff
@@ -0,0 +1,222 @@
...
+                    PublicationQueue,
+                It.IsAny<List<PublicationNotificationMessage>>()), Times.Never);
+            }
+        }
+
+        [Fact]
+        public async Task NotifySubscribersIfApplicable_None()
+        {
+            var storageQueueService = new Mock<IStorageQueueService>(MockBehavior.Strict);
+            storageQueueService.Setup(mock => mock.AddMessages(
```
benoutram, line 190: I'm not sure what the purpose of this setup is because it should never be called if there's no release id's.

I would simplify this massively in combination with my other suggested to create the default queue service mock with `Strict` behaviour when building the service.

```
        [Fact]
        public async Task NotifySubscribersIfApplicable_NoReleaseIds()
        {
            await using (var contentDbContext = InMemoryContentDbContext())
            {
                var notificationsService = BuildNotificationsService(contentDbContext);

                await notificationsService.NotifySubscribersIfApplicable();

                // No interaction with the queue service is expected
                // since no releases are being published
            }
        }
```

src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/Services/NotificationsServiceTests.cs
```diff
@@ -0,0 +1,222 @@
...
+            await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId))
+            {
+                var notificationsService = BuildNotificationsService(
+                    contentDbContext, storageQueueService.Object);
+
+                await notificationsService.NotifySubscribersIfApplicable(release1.Id, release2.Id, release3.Id);
+
+                storageQueueService.Verify(mock => mock.AddMessages(
+                    PublicationQueue,
+                It.Is<List<PublicationNotificationMessage>>(list =>
```
benoutram, line 115: Not 100% sure about this, but I think the test will still pass without `It.Is<List<PublicationNotificationMessage>>(list =>
                    list.SequenceEqual(` due to the equality methods that have been added on `PublicationNotificationMessage`

I'd be interested to know if there's any benefit to using `It.Is<List...` here

src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/Services/NotificationsServiceTests.cs
```diff
@@ -0,0 +1,222 @@
...
+                },
+            };
+            var contentDbContextId = Guid.NewGuid().ToString();
+            await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId))
+            {
+                await contentDbContext.AddRangeAsync(release);
+                await contentDbContext.SaveChangesAsync();
+            }
+
+            var storageQueueService = new Mock<IStorageQueueService>(MockBehavior.Strict);
```
benoutram, line 167: Similar to my comment below on the `NotifySubscribersIfApplicable_None` you could strip out the `storageQueueService ` from this test. The test will fail if there's an interaction with the strict mock.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/Services/NotificationsServiceTests.cs
markhiveit: Seems I was confused about what was necessary to check that a method didn't get called - fixed.

src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/Services/NotificationsServiceTests.cs
```diff
@@ -0,0 +1,201 @@
...
+
+                // No interaction with storage queue service expected as NotifyOn is set
+                // in ReleaseStatus.
+            }
+        }
+
+        [Fact]
+        public async Task NotifySubscribersIfApplicable_NoReleaseIds()
+        {
+            var contentDbContextId = Guid.NewGuid().ToString();
```
benoutram, line 180: Minor one but you don't need an id here, a unique db name will be generated if the id is omitted.

benoutram (approved):

PR #2848, opened 2021-09-08
EES-2675 - create prod table tool tests

luke-h1:


tests/robot-tests/tests/general_public/prod_table_tool.robot
```diff
@@ -0,0 +1,131 @@
...
+    ...    //*[@data-testid="Expand Details Section Teachers and school workforce" and text()="Teachers and school workforce"]
+
+    user checks page contains element
+    ...    //*[@data-testid="Expand Details Section UK education and training statistics" and text()="UK education and training statistics"]
+
+Check page contains correct publications
+    # Children's social care
+    user clicks element    //*[@data-testid="Expand Details Section Children's social care"]
+
+    user checks page contains element    //*[text()="Children in need and child protection"]
```
markhiveit, line 49: You can just use `user checks page contains   Children in need and child protection` in cases when you're just checking the text is there.

tests/robot-tests/tests/general_public/prod_table_tool.robot
```diff
@@ -0,0 +1,131 @@
...
+
+    user checks page contains element
+    ...    //*[@data-testid="Expand Details Section Teachers and school workforce" and text()="Teachers and school workforce"]
+
+    user checks page contains element
+    ...    //*[@data-testid="Expand Details Section UK education and training statistics" and text()="UK education and training statistics"]
+
+Check page contains correct publications
+    # Children's social care
+    user clicks element    //*[@data-testid="Expand Details Section Children's social care"]
```
markhiveit, line 47: I think we can use  `user clicks element    testid:Expand Details Section Children's social care` instead of using xpath to look for `data-testid`s.

tests/robot-tests/tests/general_public/prod_table_tool.robot
```diff
@@ -0,0 +1,131 @@
...
+Suite Teardown      user closes the browser
+
+Force Tags          GeneralPublic    Prod
+
+*** Test Cases ***
+Go to 'data-tables' page
+    user navigates to data tables page on public frontend
+
+Check page contains correct themes
+    user checks page contains element
```
markhiveit, line 14: I'd create a new keyword in common for these cases in common.robot - `user checks page for details dropdown`. Should be able to use the same selector that `user opens details dropdown` uses - `xpath:.//details/summary[contains(., "${text}")]`

markhiveit: You could also use this new keyword in other places in this test suite instead of just searching for the text itself.

tests/robot-tests/tests/general_public/prod_table_tool.robot
```diff
@@ -0,0 +1,131 @@
...
+Suite Setup         user opens the browser
+Suite Teardown      user closes the browser
+
+Force Tags          GeneralPublic    Prod
+
+*** Test Cases ***
+Go to 'data-tables' page
+    user navigates to data tables page on public frontend
+
+Check page contains correct themes
```
markhiveit, line 13: Before this test case, I'd add a test case to count the number of themes using `user checks element count is x`. May need a little bit of care to make sure you're only selecting the details elements that are publications, rather than count all the themes and topics as well.

The best I can come up with is `xpath://fieldset[@id="publicationForm-publicationId"]/div/div/details`. It might be possible to get something better by adding a data-testid in the frontend code.

tests/robot-tests/tests/general_public/prod_table_tool.robot
```diff
@@ -0,0 +1,131 @@
...
+
+    # Finance & funding
+    user clicks element    //*[@data-testid="Expand Details Section Finance and funding"]
+
+    user checks page contains element    //*[text()="Local authority and school finance"]
+
+    user checks page contains element    //*[text()="Student loan forecasts"]
+
+    # Further education
+    user scrolls down    400
```
markhiveit, line 86: Is this necessary for the tests to pass? If not, then I'd remove it.

tests/robot-tests/tests/general_public/prod_table_tool.robot
```diff
@@ -0,0 +1,131 @@
...
+    user checks page contains element
+    ...    //*[@data-testid="Expand Details Section School and college outcomes and performance" and text()="School and college outcomes and performance"]
+
+    user checks page contains element
+    ...    //*[@data-testid="Expand Details Section Teachers and school workforce" and text()="Teachers and school workforce"]
+
+    user checks page contains element
+    ...    //*[@data-testid="Expand Details Section UK education and training statistics" and text()="UK education and training statistics"]
+
+Check page contains correct publications
```
markhiveit, line 45: I think I'd split this into multiple test cases - one for each theme.

tests/robot-tests/tests/general_public/prod_table_tool.robot
```diff
@@ -0,0 +1,131 @@
...
+    user clicks element    //*[@data-testid="Expand Details Section Early years"]
+    user checks page contains element    //*[text()="Childcare and early years"]
+    user checks page contains element    //*[text()="Early years foundation stage profile"]
+
+    # Finance & funding
+    user clicks element    //*[@data-testid="Expand Details Section Finance and funding"]
+
+    user checks page contains element    //*[text()="Local authority and school finance"]
+
+    user checks page contains element    //*[text()="Student loan forecasts"]
```
markhiveit, line 83: The spacing of these sections (i.e. if there is a blank line between keywords) is different for each theme - I'd make the spacing the same across every theme, just so its easier to read. 

I think my preference would be one space between clicking the element and checking for the first topic, and then no extra blank lines between keywords checking for other topics under that theme - as `Destination of pupils & students`. 

But it doesn't really matter how - making it consistent is the main thing.

tests/robot-tests/tests/general_public/prod_table_tool.robot
markhiveit: I'd also check that the number of publications is correct - which could be done by checking the number of radio buttons on the page.

tests/robot-tests/tests/general_public/prod_table_tool.robot
markhiveit: Also this keyword is incorrect, as it is actually checking the themes and topics - the publications are represented by radio buttons.

markhiveit (changes requested):

tests/robot-tests/tests/general_public/prod_table_tool.robot
markhiveit: You've not created a new keyword for the checks in this test case.

I envision the keywords in this test case looking like  `user checks page for details dropdown    Children's social care`. See above comments.

tests/robot-tests/tests/general_public/prod_table_tool.robot
```diff
@@ -0,0 +1,120 @@
...
+    ...    //*[@data-testid="Expand Details Section School and college outcomes and performance" and text()="School and college outcomes and performance"]
+
+    user checks page contains element
+    ...    //*[@data-testid="Expand Details Section Teachers and school workforce" and text()="Teachers and school workforce"]
+
+    user checks page contains element
+    ...    //*[@data-testid="Expand Details Section UK education and training statistics" and text()="UK education and training statistics"]
+
+check publications have correct number of releases
+    # There are 56 releases on prod
```
markhiveit, line 47: This should be `There are 56 publications that have a live release with a subject on prod`

markhiveit (changes requested):

tests/robot-tests/tests/general_public/prod_table_tool.robot
luke-h1: ah my bad, not sure how I missed this. I've created two new keywords: 
- `user expands detail section`: Responsible for opening theme headings 
- `user checks page for details dropdown`: Responsible for asserting whether a topic is visible

markhiveit (approved):

PR #2851, opened 2021-09-09
EES-2007 Finalize various links and redirects to data catalogue page 

ntsim:
## Depends on #2849 

This PR completes EES-2007 by:

- Adding various links to the 'Data catalogue' page across the frontend's home page and release content page.
- Redirecting the previous `/download-latest-data` URL to `/data-catalogue`
- Removing all of the previous code around download themes (used by `/download-latest-data`)

We have also updated the 'Explore data and files' accordion section to remove the data file download links:

image

The main motivation is to push users to use the data catalogue. As part of this, we have made the 'Download all data and files for this release' link more prominent by making it a right aligned button.

Note that the previous data file download links will still be displayed in the admin, as there is no data catalogue page:

image


src/GovUk.Education.ExploreEducationStatistics.Content.Services/ThemeService.cs
```diff
@@ -130,81 +128,6 @@ private static PublicationTreeNode BuildPublicationNode(Publication publication)
             };
         }
 
-        // TODO: EES-2365 Remove once 'Download latest data' page no longer exists
-        [BlobCache(typeof(PublicationDownloadsTreeCacheKey))]
```
benoutram, line 134: We no longer need to bother invalidating this cache item now in the Admin on the Manage Publication page, when editing Themes and Topics, and during both workflows of the publisher. I think cleaning all that up will then leave `PublicationDownloadsTreeCacheKey` unused so it can be deleted after that.

benoutram: Also we could make a note to delete `publication-downloads-tree.json` from the environments after the deploy so it doesn't cause any confusion later on.

src/explore-education-statistics-frontend/next.config.js
```diff
@@ -80,6 +80,15 @@ const nextConfig = {
     HOTJAR_ID: process.env.HOTJAR_ID,
     PUBLIC_URL: process.env.PUBLIC_URL,
   },
+  async redirects() {
+    return [
+      {
+        source: '/download-latest-data',
```
benoutram, line 86: Cool, will have to remember this if we want to do any more redirects! 🙂

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ThemeService.cs
ntsim: Nice catch :smile: removed!

benoutram (approved):

benoutram (dismissed):
Depends on https://github.com/dfe-analytical-services/explore-education-statistics/pull/2849

benoutram (approved):

PR #2853, opened 2021-09-09
EES-2676 Populate Glossary page from Content database

markhiveit:
This PR creates a new Content database table, GlossaryEntries. This table is now used to populate the glossary page.

We use the blob cache to cache calls to the new glossary endpoint. Currently, the only way to invalidate this cache is to manually delete `glossary.json` from azure storage. At the time of writing, the future plan is for users be able to update the glossary via the admin UI, and at that point the cache blob file will get deleted whenever a change is made, so the glossary will update instantaneously.


src/GovUk.Education.ExploreEducationStatistics.Content.Model/GlossaryEntries.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Model
+{
+    public class GlossaryEntries
+    {
+        public Guid Id { get; set; }
+
+        public string Title { get; set; }
+
+        public string Slug { get; set; }
```
benoutram, line 13: What is slug going to be used for? Html anchor?

src/GovUk.Education.ExploreEducationStatistics.Content.Services/GlossaryService.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            _contentDbContext = contentDbContext;
+        }
+
+        [BlobCache(typeof(GlossaryCacheKey))]
+        public async Task<List<GlossaryCategoryViewModel>> GetAllGlossaryEntries()
+        {
+            var categories = "ABCDEFGHIJKLMNOPQRSTUVWXYZ".ToCharArray();
+
+            return await categories
```
benoutram, line 28: I was surprised to see this because it's a db call for every letter of the alphabet

I would have approached it with something like this, but I realise now that you want empty view models for every letter of the alphabet. Also there might be some differences in performance. Mine will querying the entire table in one go which might not be such a great idea 😄 

            return _contentDbContext.GlossaryEntries
                .ToList()
                .GroupBy(e => e.Title[..1].ToUpper(), e => e)
                .Select(e => new GlossaryCategoryViewModel
                {
                    Heading = e.Key,
                    Entries = e.Select(inner => new GlossaryEntryViewModel
                    {
                        Title = inner.Title,
                        Slug = inner.Slug,
                        Body = inner.Body
                    }).OrderBy(inner => inner.Title).ToList()
                })
                .OrderBy(model => model.Heading).ToList();
```

File: src/GovUk.Education.ExploreEducationStatistics.Content.Model/GlossaryEntries.cs (hunk 1 above)
Comment (markhiveit) [reply src/GovUk.Education.ExploreEducationStatistics.Content.Model/GlossaryEntries.cs:13]: It is used for the id tag on the glossary page. We need to keep this for links to specific glossary entries.

File: src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20210908110527_EES2676CreateGlossaryEntriesTableAndAddEntries.cs [hunk 3]
```diff
@@ -0,0 +1,48 @@
...
+            migrationBuilder.CreateTable(
+                name: "GlossaryEntries",
+                columns: table => new
+                {
+                    Id = table.Column<Guid>(nullable: false),
+                    Title = table.Column<string>(nullable: false),
+                    Slug = table.Column<string>(nullable: false),
+                    Body = table.Column<string>(nullable: false),
+                    Created = table.Column<DateTime>(nullable: true),
+                    CreatedById = table.Column<Guid>(nullable: true)
```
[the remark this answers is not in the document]
markhiveit: I've set it to Bau1 - which uses the same Id across all environments. And I believe there isn't a way to log in as Bau1 on Prod, so we know on Prod this must have been set manually or by migration.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/GlossaryService.cs
markhiveit: I've updated it so it only makes one database call now, but still keeps a element for every letter of the alphabet.

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20210908110527_InitialGlossaryEntries.sql
```diff
@@ -0,0 +1,104 @@
+﻿-- Create initial glossary entries
+
+INSERT INTO dbo.GlossaryEntries
+(Id, Created, CreatedById, Title, Slug, Body)
+VALUES
+-- A
+(NEWID(), GETUTCDATE(), 'b99e8358-9a5e-4a3a-9288-6f94c7e1e3dd', 'Absence', 'absence', N'<p>When a pupil misses (or is absent from) at least 1 possible school session.</p><p>Counted in sessions, where 1 session is equivalent to half-a-day.</p><p>There are 4 types of absence:</p><ul><li><a href="/glossary#authorised-absence">authorised absence</a></li><li><a href="/glossary#overall-absence">overall absence</a></li><li><a href="/glossary#persistent-absence">persistent absence</a></li><li><a href="/glossary#unauthorised-absence">unauthorised absence</a></li></ul>'),
```
benoutram, line 7: Just letting you know you could have done something like 

```DECLARE @bau1UserId uniqueidentifier = 'b99e8358-9a5e-4a3a-9288-6f94c7e1e3dd';```

or

```DECLARE @bau1UserId uniqueidentifier = (SELECT Id FROM Users WHERE Email = 'ees-bau1@education.gov.uk');```

and used it like

```INSERT INTO dbo.GlossaryEntries (Id, Created, CreatedById, Title, Slug, Body) VALUES
(NEWID(), GETUTCDATE(), @bau1UserId, 'Absence' ...
```

to make it a bit clearer what it is, and in the second example make sure its the correct id.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/GlossaryService.cs
```diff
@@ -0,0 +1,57 @@
...
+                        .Add(new GlossaryEntryViewModel
+                    {
+                        Title = e.Title,
+                        Slug = e.Slug,
+                        Body = e.Body,
+                    });
+                });
+
+            return categories
+                .Select(category => categoryDictionary[category.ToString()])
```
benoutram, line 52: You've got a value for every letter so you don't need to do this mapping in a `Select` from each letter to dictionary value.

Just use the `Values` property of the dictionary which is a collection of all the values.

```
            return categoryDictionary.Values
                .OrderBy(c => c.Heading)
                .ToList();
```

src/GovUk.Education.ExploreEducationStatistics.Content.Services/GlossaryService.cs
```diff
@@ -0,0 +1,57 @@
...
+                {
+                    Heading = c.ToString(),
+                    Entries = new List<GlossaryEntryViewModel>(),
+                });
+
+            var allGlossaryEntries = _contentDbContext.GlossaryEntries
+                .ToList();
+
+            allGlossaryEntries
+                .OrderBy(e => e.Title)
```
benoutram, line 39: I think I would have done the ordering in the database query above but it probably makes little difference.

markhiveit: I did try that but it gave me an error so I switched to doing the ordering here - but I just had a second attempt at it - I needed to add an initial `ToAsyncEnumerable` to make it work. So I'm ordering on the database query now.

benoutram (approved):

PR #2859, opened 2021-09-14
EES-2235 change meta guidance to data guidance

amyb-hiveit:
Rename from ‘Metadata guidance document' to 'Data guidance’. 

Adds text and link to data catalogue to the data guidance page. Note that in the admin if the release is not published then only text, instead of the link, is shown.


tests/robot-tests/tests/admin/analyst/role_ui_permissions/release_and_publication_role_end_to_end_permissions.robot
```diff
@@ -47,7 +47,7 @@ Check publication owner can upload subject file
     user uploads subject    ${SUBJECT_NAME}    seven_filters.csv    seven_filters.meta.csv
 
 Check publication owner can add meta guidance to ${SUBJECT_NAME}
-    user clicks link    Metadata guidance
+    user clicks link    Data guidance
     user waits until page does not contain loading spinner
     user enters text into element    id:metaGuidanceForm-content    Test meta guidance content
```
ntsim, line 52: Would suggest we also change all of the various element ids across the admin too i.e.

- Form id for `ReleaseMetaGuidanceSection` (admin)
- `metaGuidance-dataFiles` in `ReleaseMetaGuidanceSection` (admin)
- `metaGuidance-content` in `ReleaseMetaGuidancePageContent` (common)

There might be others too, so would recommend just giving it another once over

src/explore-education-statistics-admin/src/pages/release/ReleaseMetaGuidancePage.tsx
```diff
@@ -48,11 +52,22 @@ const ReleaseMetaGuidancePage = ({
...
               metaGuidance={model.metaGuidance.content}
+              renderDataCatalogueLink={
+                model.release.published ? (
+                  <Link
+                    to={`${PublicAppUrl}/data-catalogue/${model.release.publicationSlug}/${model.release.slug}`}
+                  >
+                    data catalogue
+                  </Link>
+                ) : (
+                  'data catalogue'
```
ntsim, line 68: If we have nothing to link to, shouldn't `renderDataCatalogueLink` be set to undefined?

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/release/ReleaseMetaGuidancePage.tsx
amyb-hiveit: I was thinking that as in the admin it's kind of a preview of the page then it made more sense to show the text (and the sentence it's part of) as that gives a more accurate representation of what the public page will look like. Do we have a standard convention for this sort of thing?

src/explore-education-statistics-admin/src/pages/release/ReleaseMetaGuidancePage.tsx
```diff
@@ -48,11 +52,22 @@ const ReleaseMetaGuidancePage = ({
               caption={model.release.title}
             />
 
-            <h2>Metadata guidance document</h2>
+            <h2>Data guidance</h2>
 
             <ReleaseMetaGuidancePageContent
```
ntsim, line 57: We should also rename all of the `MetaGuidance*` components across the project as well

src/explore-education-statistics-admin/src/pages/release/ReleaseMetaGuidancePage.tsx
ntsim: Not sure, but so far we've generally omitted anything that can't actually be displayed in the admin. I asked Laura if we were planning on putting in an admin data catalogue page and she said no.

It might be worth asking her/Cam what they think about showing this text, even if there is no page that can be displayed!

PR #2864, opened 2021-09-15
EES-2695 - update RF to 4.1.1

luke-h1:
This PR: 
- updates RF to 4.1.1 
- updates peer dependencies of RF
- updates azure-core from 1.16.0 to 1.8.0

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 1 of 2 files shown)


Pipfile.lock
```diff
@@ -18,10 +18,10 @@
     "default": {
```
duncan-at-hiveit, line 18: Does Pipfile not also need committing?

duncan-at-hiveit (approved):

PR #2867, opened 2021-09-15
EES-2697 - prevent prod UI tests from throwing false failures

luke-h1:
This PR: 
- prevents the prod UI tests from throwing errors due to not finding an element in x amount of seconds. This is caused by the UI tests being run in a slow CI environment. Once the UI tests are able to run in a VMSS this issue should go away

### DIFF
--- a/tests/robot-tests/tests/general_public/prod_table_tool.robot
+++ b/tests/robot-tests/tests/general_public/prod_table_tool.robot
@@ -40,7 +40,7 @@ check publications have correct number of releases
     user checks element count is x    //*[@name="publicationId"]    56
 
 check 'Children's social care' contains correct topics
-    user expands details section    Children's social care
+    user opens details dropdown    Children's social care
     user checks page for details dropdown    Children in need and child protection
     user checks page for details dropdown    Children looked after
     user checks page for details dropdown    Children's social work workforce
@@ -49,40 +49,40 @@ check 'Children's social care' contains correct topics
     user checks page for details dropdown    Serious incident notifications
 
 check 'COVID-19' contains correct topics
-    user expands details section    COVID-19
+    user opens details dropdown    COVID-19
     user checks page for details dropdown    Attendance
     user checks page for details dropdown    Confirmed cases reported by Higher Education providers
     user checks page for details dropdown    Devices
     user checks page for details dropdown    Testing
 
 check 'Destination of pupils and students' contains correct topics
-    user expands details section    Destination of pupils and students
+    user opens details dropdown    Destination of pupils and students
     user checks page for details dropdown    Destinations of key stage 4 and 16-18 pupils
     user checks page for details dropdown    NEET and participation
 
 check 'Early years' contains correct topics
-    user expands details section    Early years
+    user opens details dropdown    Early years
     user checks page for details dropdown    Childcare and early years
     user checks page for details dropdown    Early years foundation stage profile
 
 check 'Finance and funding' contains correct topics
-    user expands details section    Finance and funding
+    user opens details dropdown    Finance and funding
     user checks page for details dropdown    Local authority and school finance
     user checks page for details dropdown    Student loan forecasts
 
 check 'Further education' contains correct topics
-    user expands details section    Further education
+    user opens details dropdown    Further education
     user checks page for details dropdown    Further education and skills
     user checks page for details dropdown    Further education: outcome-based success measures
 
 check 'Higher education' contains correct topics
-    user expands details section    Higher education
+    user opens details dropdown    Higher education
     user checks page for details dropdown    Higher education graduate employment and earnings
     user checks page for details dropdown    Participation measures in higher education
     user checks page for details dropdown    Widening participation in higher education
 
 check 'Pupils and schools' contains correct topics
-    user expands details section    Pupils and schools
+    user opens details dropdown    Pupils and schools
     user checks page for details dropdown    Academy transfers
     user checks page for details dropdown    Admission appeals
     user checks page for details dropdown    Exclusions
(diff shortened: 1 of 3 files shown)


tests/robot-tests/tests/libs/common.robot
```diff
@@ -773,13 +773,13 @@ user closes Set Page View box
     user waits until element is not visible    id:editingMode
 
 user checks page for details section
-    [Arguments]    ${heading}
-    user checks page contains element    testid:Expand Details Section ${heading}
```
markhiveit, line 777: In `common.robot`, `user checks page contains element` doesn't take a wait argument, so this keyword is currently going to always fail due to having the wrong number of arguments.

Same for `user checks page for details dropdown`.

tests/robot-tests/tests/general_public/prod_table_tool.robot
```diff
@@ -11,27 +11,27 @@ Go to 'data-tables' page
     user navigates to data tables page on public frontend
 
 Check page contains correct themes
-    user checks page for details section    Children's social care
```
markhiveit, line 14: Rather than wait for every single one of these, I think it's better to wait for something that should be on the page every time, like "Choose a publication" text or the "Next step" button. I usually find if you wait for something inside the same react component as the thing you want to check, then that's good enough.

You could still add a wait for the first details section. But if the first details section is rendered, then I don't think there should be a need to wait for the others - they'll either all be rendered or none.

tests/robot-tests/tests/general_public/prod_table_tool.robot
```diff
@@ -41,70 +41,70 @@ check publications have correct number of releases
 
 check 'Children's social care' contains correct topics
     user expands details section    Children's social care
```
markhiveit, line 43: There is already a keyword to open a details dropdown which I think can be used here: `user opens details dropdown`. This is preferable as it actually waits for the dropdown to expand, which means  you shouldn't have to add waits to the `user checks page for details dropdown` keywords after it. So I would remove all those waits.

It seems like `user expands details section` is only being used in this test suite, so after subsituting all instances with `user opens details dropdown`, I think you can remove `user expands details section` from `common.robot`.

markhiveit (changes requested):

tests/robot-tests/tests/general_public/prod_table_tool.robot
luke-h1: This is mainly to assert that those details sections exist on the page and not to determine whether the page has loaded

tests/robot-tests/tests/general_public/prod_table_tool.robot
luke-h1: Cool, didn't realize this was in common.robot. I've removed it

tests/robot-tests/tests/libs/common.robot
```diff
@@ -411,8 +411,8 @@ user checks page does not contain
     page should not contain    ${text}
 
 user checks page contains element
-    [Arguments]    ${element}
-    page should contain element    ${element}
+    [Arguments]    ${element}    ${wait}
+    page should contain element    ${element}    ${wait}=${timeout}
```
markhiveit, line 415: `page should contain element` doesn't take a timeout argument. See https://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#Page%20Should%20Contain%20Element

The seleniumlibrary keyword you need to wait is `wait until page contains element`: https://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#Wait%20Until%20Page%20Contains%20Element

We already have a keyword to wrap this in common.robot: `user waits until page contains element` So I'd remove the wait from `user checks page contains element`.

tests/robot-tests/tests/libs/common.robot
```diff
@@ -720,9 +720,9 @@ user checks nth breadcrumb contains
     user checks element should contain    css:[data-testid="breadcrumbs--list"] li:nth-child(${num})    ${text}
 
 user checks page contains other release
-    [Arguments]    ${other_release_title}
+    [Arguments]    ${other_release_title}    ${wait}=${timeout}
     user checks page contains element
-    ...    xpath://li[@data-testid="other-release-item"]/a[text()="${other_release_title}"]
+    ...    xpath://li[@data-testid="other-release-item"]/a[text()="${other_release_title}"]    ${wait}
```
markhiveit, line 725: If you want to add a wait here, you'll want to use the `user waits until page contains element` and change the keyword name too so it reflects the fact that it waits rather than instantly checks.

That said, I'm not sure any test suite is using a wait argument with `user checks page contains other release`? If not, then I'd just remove the wait argument.

tests/robot-tests/tests/libs/common.robot
```diff
@@ -773,13 +773,9 @@ user closes Set Page View box
...
 
 user checks page for details section
-    [Arguments]    ${heading}
-    user checks page contains element    testid:Expand Details Section ${heading}
-
-user expands details section
-    [Arguments]    ${heading}
-    user clicks element    testid:Expand Details Section ${heading}
+    [Arguments]    ${heading}    ${wait}=${timeout}
+    user checks page contains element    testid:Expand Details Section ${heading}    ${wait}
```
markhiveit, line 777: Same here as with `user checks page contains other release`.

tests/robot-tests/tests/libs/common.robot
```diff
@@ -773,13 +773,9 @@ user closes Set Page View box
...
 
 user checks page for details dropdown
-    [Arguments]    ${text}
-    user checks page contains element    xpath:.//details/summary[contains(., "${text}")]
+    [Arguments]    ${text}    ${wait}=${timeout}
+    user checks page contains element    xpath:.//details/summary[contains(., "${text}")]    ${wait}
```
markhiveit, line 781: Same here as with `user checks page contains other release`.

markhiveit (changes requested):

tests/robot-tests/tests/libs/common.robot
luke-h1: There were a couple of test suites using this keyword. I've changed the keyword to reflect the changes I've made to: `user waits until page contains other release`

tests/robot-tests/tests/libs/common.robot
```diff
@@ -773,13 +773,9 @@ user closes Set Page View box
     user waits until element is not visible    id:editingMode
 
 user checks page for details section
```
markhiveit, line 775: This keyword can now be renamed to `user waits until page contains details section`

Although if nothing is using this keyword then I think it can be removed?

tests/robot-tests/tests/libs/common.robot
```diff
@@ -773,13 +773,9 @@ user closes Set Page View box
...
-    [Arguments]    ${heading}
-    user checks page contains element    testid:Expand Details Section ${heading}
-
-user expands details section
-    [Arguments]    ${heading}
-    user clicks element    testid:Expand Details Section ${heading}
+    [Arguments]    ${heading}    ${wait}=${timeout}
+    user waits until page contains element    testid:Expand Details Section ${heading}    ${wait}
 
 user checks page for details dropdown
```
markhiveit, line 779: This keyword can also be renamed to `user waits until...`

But I also think this can be removed potentially?

markhiveit (changes requested):

markhiveit (approved):

PR #2880, opened 2021-09-20
EES-2715 Order TimeIdentifiers with C# app, not DB

markhiveit:


src/GovUk.Education.ExploreEducationStatistics.Data.Services/MetaGuidanceSubjectService.cs
```diff
@@ -95,6 +95,7 @@ public async Task<TimePeriodLabels> GetTimePeriods(Guid subjectId)
                 .Observation
                 .Where(observation => observation.SubjectId == subjectId)
                 .Select(observation => new {observation.Year, observation.TimeIdentifier})
+                .ToList()
```
benoutram, line 98: Seems fine but maybe `MetaGuidanceSubjectService` shouldn't get too hung up about this and just defer to `TimePeriodService` which already has a better version of this query in it, doing a distinct before returning from the database so that all the Year/TimeIdentifier combinations aren't pulled into memory before the ordering.

This list will be as long as the number of observations in the Subject when it doesn't need to be to get the first and last.

So you could consider moving this to `TimePeriodService` as something like `GetTimePeriodLabels` which would be similar to `GetTimePeriodRange` and reuse the private method `GetDistinctObservationTimePeriods` which doesn't have this bug.

After this we will only have this logic in one place.

It seems like that suffered from the same fate until EES-2429

src/GovUk.Education.ExploreEducationStatistics.Data.Services/TimePeriodService.cs
```diff
@@ -48,6 +50,26 @@ public IEnumerable<(int Year, TimeIdentifier TimeIdentifier)> GetTimePeriods(Gui
             return TimePeriodUtil.GetTimePeriodRange(start, end);
         }
 
+        public TimePeriodRangeLabels GetTimePeriodRangeLabels(Guid subjectId)
+        {
+            var observationsQuery = _context
```
duncan-at-hiveit, line 55: Bear in mind that there could be hundreds of thousands of these, and pulling them all out of the db might be slow.  Maybe try testing using an nd01_anon file, which hs around 200,000 rows?

duncan-at-hiveit: I do see though that this hooks into pre-existing code, but I think if there's an inefficiency here, it'll strike in the first line of `GetDistinctObservationTimePeriods()` that this code calls.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/TimePeriodService.cs
```diff
@@ -48,6 +50,26 @@ public IEnumerable<(int Year, TimeIdentifier TimeIdentifier)> GetTimePeriods(Gui
...
+            if (!orderedTimePeriods.Any())
+            {
+                return new TimePeriodRangeLabels();
+            }
+
+            var first = orderedTimePeriods.First();
+            var last = orderedTimePeriods.Last();
+
+            return new TimePeriodRangeLabels(
+                TimePeriodLabelFormatter.Format(first.Year, first.TimeIdentifier),
```
duncan-at-hiveit, line 69: Out of interest, why do we pass in Year as the first argument?  Don't some time periods operate as just weeks i.e. without any Year component?  I can see from your tests that the Weeks behave themselves - just curious about this argument.

markhiveit: I think all time identifiers are always associated with a year. So with weeks the time period range labels could be from / to something like "2020 Week 12", "2021 Week 2".

src/GovUk.Education.ExploreEducationStatistics.Data.Services/TimePeriodService.cs
duncan-at-hiveit: My bad, sorry!

duncan-at-hiveit (approved):

PR #2882, opened 2021-09-20
EES-2680 Refresh permissions after methodology status update

benoutram:
This PR makes a change to the methodology Sign off view to refresh the permissions after the form is submitted.

Depending on the new status and the status of releases that are using this methodology, the permissions can change which affect whether the 'Edit status' button should still be shown or not.

Example:

* If a user approves a methodology for publishing immediately and it's already used by a published release, then its publicly accessible immediately. The ‘Edit status’ button should not be visible since it’s publicly accessible and can't be modified except by creating an amendment.

* If a user approves a methodology for publishing immediately and it's _not_ used by any published release, then it's not yet made publicly accessible. We continue to show the 'Edit status' button so that it can still be unapproved, or the publishing strategy can be changed.

I was unsure how to test this change but a couple of additional steps are added in to the UI tests to at least make sure that users just arriving at the Sign off view can't edit the status if the methodology is already publicly accessible.

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/methodology/edit-methodology/status/MethodologyStatusPage.tsx
+++ b/src/explore-education-statistics-admin/src/pages/methodology/edit-methodology/status/MethodologyStatusPage.tsx
@@ -42,7 +42,7 @@ const MethodologyStatusPage = () => {
 
   const {
     value: permissions,
-    setState: setPermissions,
+    retry: refreshPermissions,
     isLoading,
   } = useAsyncRetry(async () => {
     const [canApprove, canMarkAsDraft] = await Promise.all([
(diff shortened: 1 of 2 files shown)


amyb-hiveit (dismissed):
LGTM 👍 
Re testing the only other thing I can think to do would be to check in the unit tests if the permissions are refreshed, but that might be overkill.

src/explore-education-statistics-admin/src/pages/methodology/edit-methodology/status/MethodologyStatusPage.tsx
```diff
@@ -76,6 +80,15 @@ const MethodologyStatusPage = () => {
...
     onMethodologyChange(nextSummary);
 
+    const [nextCanApprove, nextCanMarkAsDraft] = await Promise.all([
+      permissionService.canApproveMethodology(methodologyId),
+      permissionService.canMarkMethodologyAsDraft(methodologyId),
+    ]);
+
+    setPermissions({
+      value: { canApprove: nextCanApprove, canMarkAsDraft: nextCanMarkAsDraft },
+    });
```
ntsim, line 90: `useAsyncRetry` exposes a `retry` method that can be used instead of all this (you might want to alias it as `refreshPermission` or something)

benoutram: That's great to know thanks as I was looking through existing code for ideas on how to avoid that duplication! Just made that change. Hopefully I have interpreted this comment correctly 🙂

ntsim (approved):

PR #2886, opened 2021-09-21
EES-2718 Add TimePeriodService to Publisher Startup.cs

markhiveit:


benoutram (approved):
Seems fine. The UI tests would help check there's no other problems.

The reason why the Publisher needs this is because its functions depend on `Publisher.Services.PublishingService` which has this rather long dependency chain, eventually ending in the TimePeriodService which the project didn't include originally.

```
PublishReleaseFilesFunction / PublishStagedReleaseContentFunction both depend on:
Publisher.Services.PublishingService depends on:
Content.Services.DataGuidanceFileService depends on:
Content.Services.DataGuidanceFileWriter depends on:
Data.Services.MetaGuidanceSubjectService depends on:
Data.Services.TimePeriodServicePublishReleaseFilesFunction / PublishStagedReleaseContentFunction both depend on:
Publisher.Services.PublishingService depends on:
Content.Services.DataGuidanceFileService depends on:
Content.Services.DataGuidanceFileWriter depends on:
Data.Services.MetaGuidanceSubjectService depends on:
Data.Services.TimePeriodService
```

PR #2891, opened 2021-09-23
EES-2546 Replace existing 'All files' zip with dynamically generated version

ntsim:
This PR replaces the existing 'All files' zip that is currently generated at publishing time, with a dynamically generated version from the content API's new `/releases/{releaseId}/files` endpoint (also used for the data catalogue).

This approach to generating the 'All files' zip is is preferable for the following reasons:

- We can keep it up to date with any changes to the structure or its contents e.g. changes to the data guidance file inside. This fixes the main issue EES-2546.
- No longer need to maintain any code related to the generation of this file in the publisher. Everything happens in the content API.

To implement this, we have updated the `/files` endpoint to make `fileIds` optional in its query parameters. This means that it will download a zip file with all of the downloadable file types instead of filtering on specific files. To an end-user, this is basically the same as going to the old `/releases/{releaseId}/files/all` endpoint (this has now been removed).

## Caching

Dynamically generating the 'All files' zip is advantageous, but it can potentially be a bit slow because we potentially need to perform a lot of network I/O. Making many requests for this file could also potentially cause network congestion and grind the content API to a halt (this could be over-thinking  the problem however).

To try and mitigate any potential performance problems, we cache this dynamically generated 'All files' zip for **1 hour**, before the next request re-caches it.

It should be noted that if the user specifies specific `fileIds` (e.g. using the data catalogue page), then the generated zip file will not be cached. We have avoided caching these files as we would need to cache files for many different combinations of `fileIds`. This didn't feel particularly manageable or nice to do!

## Other notes

- Ideally, we would implement the caching of the 'All files' zip via caching annotations. Unfortunately, this doesn't appear to be super trivial due to needing to target an output stream. Consequently, it isn't worth doing just for this single use-case. We may revisit this in the future if there are other blobs that need to be cached in a similar way.


src/GovUk.Education.ExploreEducationStatistics.Content.Services/ReleaseFileService.cs
```diff
@@ -18,45 +19,40 @@
...
 
 namespace GovUk.Education.ExploreEducationStatistics.Content.Services
 {
     public class ReleaseFileService : IReleaseFileService
     {
-        private static readonly FileType[] DownloadableFileTypes = {
-            FileType.Ancillary,
-            FileType.Data,
-        };
+        /// How long the all files zip should be
```
benoutram, line 29: After a release is published, would of thought that the files should never change if nothing is editable without creating an amendment, so I think we could have got away with this either being much longer or unlimited? Nice to have this in  though so that the zip file can easily evolve over time like the addition of data guidance without us touching the files.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ReleaseFileService.cs
```diff
@@ -78,31 +74,109 @@ public class ReleaseFileService : IReleaseFileService
...
+            await using var multiWriteStream = new MultiWriteStream(outputStream, fileStream);
+
+            await DoZipFilesToStream(releaseFiles, release, multiWriteStream, cancellationToken);
+            await multiWriteStream.FlushAsync();
+
+            // Now cache the All files zip into blob storage
+            // so that we can quickly fetch it again.
+            fileStream.Position = 0;
+
+            await _blobStorageService.UploadStream(
```
benoutram, line 164: Cool, was wondering why we were keeping the path util for the zip file until I saw this 🙂

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/MultiWriteStream.cs
```diff
@@ -0,0 +1,94 @@
...
+namespace GovUk.Education.ExploreEducationStatistics.Common.Utils
+{
+    /// <summary>
+    /// A stream implementation that wraps and writes to multiple target streams.
+    /// <para>
+    /// It should be noted that this stream is only intended for writing,
+    /// and is not intended to be seekable or readable.
+    /// </para>
+    /// </summary>
+    public class MultiWriteStream : Stream
```
benoutram, line 17: I think I understand the need for this but just checking. More efficient to start writing the response while creating the zip file at the same time?

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Common/Model/FileType.cs
```diff
@@ -6,11 +6,13 @@ public enum FileType
...
         [EnumLabelValue("ancillary")]
         Ancillary,
+        [EnumLabelValue("zip")]
+        AllFilesZip,
         [EnumLabelValue("chart")]
         Chart,
         [EnumLabelValue("data")]
         Data,
-        [EnumLabelValue("zip")]
+        [EnumLabelValue("data-zip")]
```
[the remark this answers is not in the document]
ntsim: Yeh good point. We should probably get Cam to tell everyone not to have any zip files in progress on the day we deploy!

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ReleaseFileService.cs
ntsim: Yeh I did think about that, but decided against as I imagined DfE would want the file to be updated fairly soon after a deployment updating the zip file goes out. Otherwise, they might be a little confused why the update hasn't worked. Of course, there is still a (current) timeframe of up to an hour where the file might be stale, but feels acceptable. Could potentially bump it up even higher if DfE are aware of the tradeoff though?

The other option I thought about was to add a new metadata field called `Version` or something to each zip blob. This would just be some hard-coded string from `ReleaseFileService` that we could bump up whenever there's a change to the file. My man issue is it could be accidentally missed out whenever someone makes a change to the zip and ends up needing another deployment to fix.

What do you think?

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/MultiWriteStream.cs
ntsim: So the main issue with streams is that they're like a one directional 'pipe' (but not to be confused with `System.IO.PIpe`). You can plug one stream into another stream, but there isn't a nice out of the box way to plug one stream into multiple streams - hence we need something like `MultiWriteStream` that can do exactly that.

For our use case, we're using `MultiWriteStream` to simultaneously:

- Create a temporary zip file so that we can then upload this to blob storage for caching purposes
- Send the zip blob response immediately to the client so they can start downloading it

If we did both of these things serially, this would lead to the client having an unnecessary delay before they can start download the zip.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ReleaseFileService.cs
benoutram: Ah ok, it seems fine as it is then! I expect that any updates being deployed that can change the content of the zip file will be fairly infrequent, and 1 hour is a short-ish amount of time to wait if there is a change.

I think the version field is a good idea but seems overkill for now and like you say, we might forget to update it. 🙂

PR #2893, opened 2021-09-23
EES-2698 Fix UI tests local stale element exception

markhiveit:


benoutram (approved):

luke-h1:
It looks like this is causing the following failure:  
``` 
Element 'id:publicationsReleases-themeTopic-topicId' did not appear in 30 seconds. 
```

PR #2900, opened 2021-09-24
EES-274 EES-2690 EES-2683 Add ability to view glossary entries in situ instead of visiting glossary page

markhiveit:
This PR gives admin users the ability to create glossary links that allow users to view a glossary link within the same page they see a reference to it, rather than visiting a link to the glossary page itself.

image

image

This is an MVP implementation of this feature that we intend to flesh out in EES-2701. In this current iteration, if a user adds a link to a glossary entry to a text block, we assume they want to give users the ability to view this glossary entry on the same page. Users adding glossary links are themselves responsible for finding the glossary entry link. In the fully-featured version, we intend to offer users the ability to select what glossary entry they want.

Glossary links that will be rendered with an info "?" are identified by having a `data-glossary` attribute. This attribute is added when a glossary link is added to a text block via a CKEditor link decorator.

I have created two seperate identical endpoints in the Admin API and Content API. This is due to several React components like ContentHtml being used in both the admin and public frontend projects. To prevent from having to create two versions of these components for each project, the `getGlossaryEntry` endpoint call is passed in as a prop so that the correct service is called.

I didn't create all those commits enabling the feature around the service *just* to boost my commit count. When adding `getGlossaryEntry`s around the service, I wanted to signal what each `getGlossaryEntry` being passed in related to in the frontend. So each commit message indicators what that component relates to.


src/explore-education-statistics-admin/src/components/form/FormEditor.tsx
```diff
@@ -204,6 +204,17 @@ const FormEditor = ({
...
         contentToolbar: tableContentToolbar,
       },
+      link: {
+        decorators: {
+          addDataGlossaryAttributeToGlossaryLinks: {
+            mode: 'automatic',
+            callback: (url: string) =>
+              url.startsWith(process.env.PUBLIC_URL) &&
+              url.match(/\/glossary#[a-zA-Z-0-9-]+$/),
+            attributes: { 'data-glossary': '' },
```
markhiveit, line 214: Unfortunately, there doesn't seem to be a way to pass in the slug (or other identifier for a glossary entry) to the attribute using a link decorator. It seems we'll need to use a plugin to do that.

src/explore-education-statistics-common/src/components/ContentHtml.tsx
```diff
@@ -1,31 +1,104 @@
...
+                setGlossaryEntry({
+                  title: newGlossaryEntry.title,
+                  slug: newGlossaryEntry.slug,
+                  body: newGlossaryEntry.body,
+                  link: node.attribs.href,
+                });
+                setDisplayGlossaryEntry(true);
+              }}
+            >
+              ?
```
markhiveit, line 68: I know I need to add some accessibility stuff for this, but wasn't sure what. So I thought I'd just let you pull me up in the PR about it.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/GlossaryServiceTests.cs
```diff
@@ -70,21 +71,57 @@ await using (var context = InMemoryContentDbContext(contextId))
...
         public async Task GetAllGlossaryEntries_NoEntries()
+        {
+            await using var context = InMemoryContentDbContext();
+            var glossaryService = new GlossaryService(context);
+
+            var result = await glossaryService.GetAllGlossaryEntries();
+
+            Assert.Equal(26, result.Count);
+
+            result.ForEach(category => { Assert.Empty(category.Entries); });
```
duncan-at-hiveit, line 82: Do you need the inner curly braces here?

src/explore-education-statistics-admin/src/components/PreviewHtml.tsx
```diff
@@ -15,6 +16,7 @@ const PreviewHtml = ({ className, html, testId }: Props) => {
       html={html}
       className={classNames(styles.preview, className)}
       testId={testId}
+      getGlossaryEntry={glossaryService.getGlossaryEntry}
```
duncan-at-hiveit, line 19: Might wanna check naming conventions with Nick here?

src/explore-education-statistics-admin/src/components/form/FormEditor.tsx
```diff
@@ -204,6 +204,17 @@ const FormEditor = ({
       table: {
         contentToolbar: tableContentToolbar,
       },
+      link: {
+        decorators: {
+          addDataGlossaryAttributeToGlossaryLinks: {
+            mode: 'automatic',
+            callback: (url: string) =>
+              url.startsWith(process.env.PUBLIC_URL) &&
```
duncan-at-hiveit, line 212: Could this startsWith instead actually just be part of the regex?

src/explore-education-statistics-admin/src/components/form/FormEditor.tsx
duncan-at-hiveit: Ooh that's annoying, but yeah I checked it out too and I agree.

src/explore-education-statistics-common/src/components/ContentHtml.tsx
duncan-at-hiveit: Fantastic, I was about to question it, and see that it's already on your radar :)

Potentially an aria-label attribute to provide a description of the link, as there's nothing else describing what it does on the page currently.

src/explore-education-statistics-common/src/components/ContentHtml.tsx
```diff
@@ -1,31 +1,107 @@
...
 }
 
 const ContentHtml = ({
   className,
   html,
   sanitizeOptions,
   testId,
+  getGlossaryEntry,
 }: ContentHtmlProps) => {
+  const [displayGlossaryEntry, setDisplayGlossaryEntry] = useState(false);
```
duncan-at-hiveit, line 25: I wouldn't bother with the `displayGlossaryEntry` flag if you don't need it.  Instead, you can make `glossaryEntry` "nullable", by replacing:

```
const [glossaryEntry, setGlossaryEntry] = useState({
    title: 'Glossary entry not retrieved',
    slug: '',
    body: '',
    link: '',
  });
```

with:

```
const [glossaryEntry, setGlossaryEntry] = useState<GlossaryEntry>();
```

Then check for it being defined or not rather than checking the flag.

duncan-at-hiveit: Unless you're doing it this way explicitly for caching purposes or some other reason?

src/explore-education-statistics-common/src/components/ContentHtml.tsx
```diff
@@ -1,31 +1,107 @@
...
+        data-testid={testId}
+      >
+        {parsedContent}
+      </div>
+      {displayGlossaryEntry && (
+        <Modal
+          title={glossaryEntry.title}
+          onExit={() => setDisplayGlossaryEntry(false)}
+        >
+          {parseHtmlString(sanitizeHtml(glossaryEntry.body))}
```
duncan-at-hiveit, line 90: I noticed we were using `sanitiseOptions` previously in this file - do we need them here as well or are they not relevant?

src/explore-education-statistics-frontend/src/services/glossaryService.ts
```diff
@@ -1,20 +1,16 @@
...
+  GlossaryCategory,
+  GlossaryEntry,
+} from '@common/services/types/glossary';
 
 const glossaryService = {
   listGlossaryEntries(): Promise<GlossaryCategory[]> {
     return contentApi.get('/glossary', {});
   },
+  getGlossaryEntry(slug: string): Promise<GlossaryEntry> {
+    return contentApi.get(`glossary/${slug}`, {});
```
duncan-at-hiveit, line 12: Do you need the empty braces at the end there?

tests/robot-tests/tests/admin_and_public/bau/publish_content.robot
```diff
@@ -0,0 +1,93 @@
...
+    user clicks link    Content
+    user waits until page contains button    Add new section
+
+Add an accordion section to release content
+    user clicks button    Add new section
+    user changes accordion section title    1    Test section
+
+Add text block with link to absence glossary entry to accordion section
+    user adds text block to editable accordion section    Test section    css:#releaseMainContent
+    #user adds content to accordion section text block    Test section    1    Test. <a href=%{PUBLIC_URL}/glossary#absence>Absence</a>. Test.    css:#releaseMainContent
```
duncan-at-hiveit, line 33: Might as well remove the commented-out text unless you want to revisit later?  I'm guessing it's superceded by the following steps though now.

src/explore-education-statistics-common/src/components/ContentHtml.tsx
```diff
@@ -1,31 +1,107 @@
...
+    slug: '',
+    body: '',
+    link: '',
+  });
+
   const cleanHtml = useMemo(() => {
     return sanitizeHtml(html, sanitizeOptions);
   }, [html, sanitizeOptions]);
 
+  const parsedContent = parseHtmlString(cleanHtml, {
```
duncan-at-hiveit, line 37: Be good to get a Jest test in for this

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/GlossaryService.cs
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+        public async Task<Either<ActionResult, GlossaryEntryViewModel>> GetGlossaryEntry(string slug)
+        {
+            return await _persistenceHelper
+                .CheckEntityExists<GlossaryEntry>(query => query
+                    .Where(e => e.Slug == slug))
+                .OnSuccess(e => new GlossaryEntryViewModel
+                {
+                    Title = e.Title,
```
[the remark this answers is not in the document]
markhiveit: Talked to Duncan about this on Slack and we're going to leave this as it is. We do use automapper in other places in the code for situations like this, but we've also had a case or two where bugs have arisen or it's not easy to see how something has been copied and then having to hunt in the automapper config to find out what's going on. I think I'd prefer to pay the cost in terms of boilerplate to avoid those issues.

src/explore-education-statistics-admin/src/components/form/FormEditor.tsx
markhiveit: If so, I'm not sure how to do it.

src/explore-education-statistics-admin/src/components/form/FormEditor.tsx
markhiveit: This is unfortunately because I think this means that, when this feature developed further, we'll either have to continue to parse the URL to determine the glossary entry to be rendered, or we'll have to migrate existing links to include some identifier to the glossary entry into the `data-glossary` attribute.

src/explore-education-statistics-common/src/components/ContentHtml.tsx
markhiveit: I've added an aria-label. Wasn't sure how to phrase the label - so I'll leave this comment unresolved.

src/explore-education-statistics-frontend/src/services/glossaryService.ts
markhiveit: Also removed it from all other calls (including in the admin glossaryService).

src/explore-education-statistics-common/src/components/ContentHtml.tsx
markhiveit: Good spot - fixed.

src/explore-education-statistics-common/src/components/ContentHtml.tsx
```diff
@@ -1,31 +1,102 @@
...
+                event.preventDefault();
+                const glossaryEntrySlug = node.attribs.href.split('#')[1];
+                const newGlossaryEntry = await getGlossaryEntry(
+                  glossaryEntrySlug,
+                );
+                setGlossaryEntry({
+                  title: newGlossaryEntry.title,
+                  slug: newGlossaryEntry.slug,
+                  body: newGlossaryEntry.body,
+                });
```
ntsim, line 60: Seems like we could simplify this to:

```ts
setGlossaryEntry(newGlossaryEntry);
```

src/explore-education-statistics-common/src/components/ContentHtml.tsx
ntsim: Probably would suggest we do something like the following (I haven't included all the button props for brevity):

```tsx
<button>
  <span aria-hidden>?</span>
  <span className="govuk-visually-hidden">
    Show glossary term definition for X
  </span>
</button>
```

Where X is the actual definition.

src/explore-education-statistics-common/src/components/ContentHtml.module.scss
```diff
@@ -0,0 +1,25 @@
+@import '~govuk-frontend/govuk/base';
+
+.infoIcon {
+  background-color: $govuk-link-colour;
+  border-radius: 10px;
+  color: govuk-colour('white') !important;
```
ntsim, line 6: Any particular reason we're using `!important` quite often in here? If it's to do with the usage of links, we'll probably find that we can change this more easily by using a button element for this instead.

ntsim (changes requested):

src/explore-education-statistics-admin/src/components/form/FormEditor.tsx
```diff
@@ -204,6 +204,17 @@ const FormEditor = ({
       table: {
         contentToolbar: tableContentToolbar,
       },
+      link: {
+        decorators: {
+          addDataGlossaryAttributeToGlossaryLinks: {
+            mode: 'automatic',
+            callback: (url: string) =>
```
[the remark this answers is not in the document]
markhiveit: I've added automatic and manual link decorator interfaces - will need checking over as I'm not 100% that it's right.

src/explore-education-statistics-common/src/components/ContentHtml.tsx
```diff
@@ -1,31 +1,102 @@
...
-      dangerouslySetInnerHTML={{ __html: cleanHtml }}
-    />
+    <>
+      <div
+        className={classNames('dfe-content', className)}
+        data-testid={testId}
+      >
+        {parsedContent}
+      </div>
+      {glossaryEntry !== undefined && (
```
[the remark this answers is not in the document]
markhiveit: I tried changing it but ended up with this (and a couple of ts warnings):
image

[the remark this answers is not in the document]
markhiveit: ah - it wants to be `glossaryEntry` not `!glossaryEntry`.

src/explore-education-statistics-common/src/components/ContentHtml.module.scss
markhiveit: They were necessary to apply the styles when it was a link - but now I've changed it to a button that aren't necessary anymore, so I've removed them.

src/explore-education-statistics-common/src/components/ContentHtml.tsx
```diff
@@ -1,31 +1,102 @@
...
+      if (
+        getGlossaryEntry &&
+        node instanceof Element &&
+        node.name === 'a' &&
+        node.attribs &&
+        Object.keys(node.attribs).includes('data-glossary')
+      ) {
+        return (
+          <>
+            <a href={node.attribs.href}>{domToReact(node.children)}</a>{' '}
```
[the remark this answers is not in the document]
markhiveit: I've added useMounted - I've removed the link but underlined the text if the infoIconButton is present, hopefully so its clear what the info icon refers to.

I originally had the link the the glossary entry on the glossary page in the modal, but decided to remove it. I'll run it by Cam and see what he wants.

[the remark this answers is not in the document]
markhiveit: So ran it past Cam. We're going with no link in the modal, but he wanted the glossary text before the info icon to also open the modal if clicked, so I've made that change.

src/explore-education-statistics-common/src/components/__tests__/ContentHtml.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+    userEvent.click(glossaryInfoModalLink);
+
+    await waitFor(() => {
+      expect(getEntry).toHaveBeenCalled();
+    });
+
+    expect(screen.getByLabelText('Absence heading')).toBeInTheDocument();
+    expect(screen.getByText('Absence body')).toBeInTheDocument();
+    const closeLink = screen.getByRole('link', { name: 'Close' });
+    expect(closeLink).toBeInTheDocument();
```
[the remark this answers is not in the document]
markhiveit: Done. I've also added an additional test now that you can open the modal by clicking the text as well as the info icon.

src/explore-education-statistics-common/src/components/ContentHtml.tsx
```diff
@@ -1,31 +1,107 @@
...
+            <InfoIconButton
+              description={`Show glossary term definition for ${linkText}`}
+              testId={`glossary-info-icon-${glossaryEntrySlug}`}
+              onClick={async () => {
+                const newGlossaryEntry = await getGlossaryEntry(
+                  glossaryEntrySlug,
+                );
+                setGlossaryEntry(newGlossaryEntry);
+              }}
+            />
```
ntsim, line 69: Nice try but having two buttons next to each other is redundant and also going to be quite confusing for screenreaders as it would end up announcing the presence of two buttons.

We should simplify it to something like the following:

```jsx
<ButtonText testId={`glossary-entry-${glossaryEntrySlug}`} onClick={...}>
  {linkText}{' '}
  <InfoIcon description="(show glossary term definition)" />
</ButtonText>
```

Note the following key features:

- `InfoIconButton` should now simply be the appropriately styled `?` icon i.e `InfoIcon`. It should still be `aria-hidden`. The `description` prop is still used to provide the accessible text.
- There is only the single `ButtonText` with the `onClick` being used to show the modal.

If there is an ugly link underline that goes beneath both the link text and the icon, I would suggest we

src/explore-education-statistics-common/src/components/InfoIconButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+    <ButtonText
+      className={styles.infoIconButton}
+      testId={testId}
+      onClick={onClick}
+    >
+      <span aria-hidden>?</span>
+      <span className="govuk-visually-hidden">{description}</span>
+    </ButtonText>
+  );
+};
```
ntsim, line 22: As mentioned earlier, we should refactor this component a bit:

- `InfoIconButton` should be renamed to `InfoIcon` and should literally just be an icon (no button).
- We can remove all the other props - only need `description`.

ntsim (changes requested):

src/explore-education-statistics-common/src/components/ContentHtml.tsx
```diff
@@ -1,31 +1,107 @@
...
+        >
+          {parseHtmlString(sanitizeHtml(glossaryEntry.body, sanitizeOptions))}
+          <p>
+            <ButtonText
+              onClick={event => {
+                setGlossaryEntry(undefined);
+              }}
+            >
+              Close
+            </ButtonText>
```
ntsim, line 100: Would use an actual `Button` instead of `ButtonText` as it stands out more

src/explore-education-statistics-common/src/components/ContentHtml.tsx
```diff
@@ -1,31 +1,107 @@
...
+      >
+        {parsedContent}
+      </div>
+      {glossaryEntry && (
+        <Modal
+          title={glossaryEntry.title}
+          onExit={() => setGlossaryEntry(undefined)}
+        >
+          {parseHtmlString(sanitizeHtml(glossaryEntry.body, sanitizeOptions))}
+          <p>
```
ntsim, line 93: Shouldn't need a wrapping `p` for htis

src/explore-education-statistics-admin/src/types/ckeditor.ts
```diff
@@ -30,6 +30,9 @@ export interface EditorConfig {
   heading?: {
     options: HeadingOption[];
   };
+  link?: {
+    decorators: { [key: string]: LinkDecoratorAutomatic | LinkDecoratorManual };
```
ntsim, line 34: Not a big deal, but with all of these index types, you can refactor them to use `Dictionary` which is a little more succinct:

```ts
decorators: Dictionary<LinkDecoratorAutomatic | LinkDecoratorManual>;
attributes: Dictionary<string>;
styles?: Dictionary<string>;
```

ntsim (approved):

PR #2903, opened 2021-09-27
EES-2709 Add publication links to public methodology page

benoutram:
This PR adds publications links to the general public methodology page for publications that are using the methodology.

This will always include the owning publication of a methodology but can also include other publications which have adopted it.
Since this is a change to the general public site, only visible publications _with published releases_ can be shown.

A recent change has already added in a 'Related information' section to the page which can be seen here on the live site:
image

This change now adds links to the publications within this section.

Here is the same example but running locally where another publication with a published release has adopted the methodology:

image

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Services/MethodologyService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Services/MethodologyService.cs
@@ -10,6 +10,7 @@ using GovUk.Education.ExploreEducationStatistics.Common.Model;
 using GovUk.Education.ExploreEducationStatistics.Common.Utils;
 using GovUk.Education.ExploreEducationStatistics.Content.Model;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
+using GovUk.Education.ExploreEducationStatistics.Content.Model.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Repository.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Content.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Content.Services.ViewModels;
@@ -55,7 +56,11 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Services
                         .Collection(m => m.Notes)
                         .LoadAsync();
 
-                    return _mapper.Map<MethodologyViewModel>(latestPublishedVersion);
+                    var viewModel = _mapper.Map<MethodologyViewModel>(latestPublishedVersion);
+                    viewModel.Publications =
+                        await GetVisiblePublicationsForMethodology(latestPublishedVersion.MethodologyId);
+
+                    return viewModel;
                 });
         }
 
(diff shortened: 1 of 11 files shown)


src/explore-education-statistics-frontend/src/modules/methodologies/MethodologyPage.tsx
```diff
@@ -89,6 +89,34 @@ const MethodologyPage: NextPage<Props> = ({ data }) => {
         </div>
         <div className="govuk-grid-column-one-third">
           <RelatedInformation>
+            {data.publications && data.publications.length > 0 && (
```
luke-h1, line 92: We might want to use optional chaining here instead of the two `&&` to be more concise. Not sure if we do this in other places. In addition to this, should we be returning some generic text just in case we have no publications? i.e. 
```
{data?.publications?.length > 0  ? (
  // display the publications
) : ( 
<p>No publications available</p>
) 

}
```

benoutram: Thanks @luke-h1 just made a slight change to use optional chaining for the notes, publications and annexes properties. Never expecting `data` to be null. Also I didn't include any default text as this shouldn't occur and I think it's worse for a general public user to see `No publications available` which they don't understand than not see the Publications heading at all.

luke-h1 (approved):

tests/robot-tests/tests/admin_and_public/bau/publish_methodology.robot
```diff
@@ -93,6 +93,14 @@ Verify that the methodology is publicly accessible
     user waits until h1 is visible    ${PUBLICATION_NAME}
     user waits until page contains title caption    Methodology
 
+Verify that the methodology displays a link to the publication
+    ${relatedInformation}=    get webelement    css:[aria-labelledby="related-information"]
```
markhiveit, line 97: We've had a few intermittent StaleElementException failures in the tests and, discussing this with Dunc and Luke, I think they can happen in situations like this, where you're grabbing the element immediately after a page has loaded, because React sometimes rerenders the page - I think this happens with the Admin dashboard when loading the theme/topic dropdowns. Even though its inefficient, it might be better to reuse the selector each time.

Not saying you need to change anything here, but just something to bear in mind.

markhiveit (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Services/MethodologyService.cs
```diff
@@ -104,6 +109,23 @@ await themes.SelectMany(model => model.Topics)
                 .ToList();
         }
 
+        private async Task<List<PublicationSummaryViewModel>> GetVisiblePublicationsForMethodology(Guid methodologyId)
```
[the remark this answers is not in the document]
benoutram: Thanks, I originally called it `GetPublishedPublicationsForMethodololgy ` but renamed it to use `visible` when I saw that terminology being used in `Content.Services.Tests.ThemeServiceTests`. There we are doing a similar thing where we check themes/topics/publications are not visible when they have no releases.

I've taken on your suggestion and renamed it back to `GetPublishedPublicationsForMethodololgy`.

tests/robot-tests/tests/admin_and_public/bau/publish_methodology.robot
benoutram: Thanks, I took on your suggestion in case it reduces the chances of seeing StaleElementException

markhiveit (approved):

PR #2904, opened 2021-09-27
EES-2687 - renamed existing ReleaseStatusServices and dependencies

duncan-at-hiveit:
This PR:
- as a forerunner to EES-2419 (breaking out ReleaseStatusService from ReleaseService in Admin), this PR renames the existing ReleaseStatusService and dependencies in the Publisher to "ReleasePublishingStatusXXX" rather than "ReleaseStatusXXX".

image


markhiveit:
I've found a few instances of variables/fields that could be updated from things like `releaseStatusRepository` to `releasePublishingStatusRepository` or `ReleaseStatusRepository` to `ReleasePublishingStatusRepository.

- ReleaseStatusAuthorizationHandlersTests.cs
- ReleaseControllerTests.cs
- UpdateSpecificReleaseAuthorizationHandlerTests.cs

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/IReleasePublishingStatusService.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces
 {
-    public interface IReleaseStatusService
+    public interface IReleasePublishingStatusService
```
markhiveit, line 9: Does this service want to be renamed? I thought it was just the ReleaseStatusService in the Publisher that should be renamed?

markhiveit (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/IReleasePublishingStatusService.cs
duncan-at-hiveit: Yes, it's because it still refers to the Publisher's Release Statuses (if you have a gander at the internals of the class, you'll see all the Table Storage stuff) :) .

markhiveit (approved):

PR #2906, opened 2021-09-27
EES-2724 - modify test wait times to avoid false positive fails

luke-h1:


tests/robot-tests/tests/libs/common.robot
```diff
@@ -25,7 +25,7 @@ do this on failure
         prompt to continue
     END
     IF    ${prompt_to_continue_on_failure} == 0
-        set selenium timeout    3
```
markhiveit, line 28: You could just remove this IF END block completely, since you're not changing the implicit wait time.

tests/robot-tests/tests/libs/admin-common.robot
```diff
@@ -559,7 +559,7 @@ user approves release for immediate publication
     user clicks button    Update status
     user waits until h2 is visible    Sign off    %{WAIT_MEDIUM}
     user checks summary list contains    Current status    Approved
-    user waits for release process status to be    Complete    ${release_complete_wait}
+    user waits for release process status to be    Complete    %{RELEASE_COMPLETE_WAIT}
```
markhiveit, line 562: There is another lower-cased `${release_complete_wait}` in the `Variables` section of `common.robot` that can be capitalised.

tests/robot-tests/tests/libs/common.robot
```diff
@@ -14,7 +14,7 @@ ${headless}=                            1
 ${FILES_DIR}=                           ${EXECDIR}${/}tests${/}files${/}
 
 ${timeout}=                             30
-${implicit_wait}=                       3
```
markhiveit, line 17: I think this is going to lengthen the time to run the tests significantly if we see failures. 

It's been a while since I've dealt with explicit vs. implicit waits, but I believe this means that every time selenium is looking for an element, it will wait for 15 seconds.

If a keyword needs to wait for longer than a few seconds for an element to appear, I think in those instances we should use a "wait until" keyword. So I think the solution here isn't to increase the implicit wait, but to change any "checks" to "wait until" - meaning it'll wait ${timeout} seconds instead.

Having an implicit wait of 3 seconds is already a compromise - ideally it would be 0. Any test that checks for an element should (ideally) have "wait until"s before that check to ensure that it is already on the page. 

All that said, I'm not against merging this and seeing what happens - it's an easy change to reverse. I'd be interested in other's opinions on this.

markhiveit (changes requested):

tests/robot-tests/tests/libs/common.robot
luke-h1: Moreover, I think this has the added benefit of preventing the boy who cried wolf mindset we currently have with the UI tests. At the moment I don't think they are doing a good job of catching actual bugs. At the moment I think there are too many false positives to have 100% confidence in them

markhiveit: Discussed this (me, Luke, Duncan) on a call and we agreed to merge this. If it causes the UI test running time on environments to increase drastically, it'll be simple to reverse the implicit wait time.

markhiveit (approved):

PR #2910, opened 2021-09-30
EES-2745 Delete Data catalogue page cache tree when publishing a release

markhiveit:
We also refactor the deleting of all cache tree json files in this PR to prevent the problem arising again.

It looks like due to this refactoring, the PublishTaxonomy function could now potentially be removed entirely. If that's right, I'll create a separate ticket for this work rather than address it here.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Publisher/Functions/PublishTaxonomyFunction.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Publisher/Functions/PublishTaxonomyFunction.cs
@@ -1,7 +1,5 @@
-using System;
 using System.Threading.Tasks;
 using GovUk.Education.ExploreEducationStatistics.Publisher.Model;
-using GovUk.Education.ExploreEducationStatistics.Publisher.Models;
 using GovUk.Education.ExploreEducationStatistics.Publisher.Services.Interfaces;
 using Microsoft.Azure.WebJobs;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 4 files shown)


benoutram:
That isn't quite the case about removing the `PublishTaxonomy` function just yet as it's executed indirectly by the Admin app when the taxonomy changes, e.g. when a BAU manages themes or topics. A message goes onto the queue, which is dequeued by that function, and the content gets updated on the public site by deleting those cache items. In the past it might have had something to do with the publisher having the connection on the public stats db which needed updating too, but I think it's also about separation of concerns. I.e. in the past the publisher has been solely responsible for managing the content json, so admin didn't have to know nothing about that. (This separation broke down recently though when I added a line in Admin to delete the cache file of the 'All methodologies' page).

To remove that function from the publisher you would need to do a similar thing and call the code to remove all the cache files when managing themes and topics in the Admin. The code you just wrote though won't be in a service callable from the Admin as it stands.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Functions/PublishTaxonomyFunction.cs
```diff
@@ -31,9 +29,7 @@ public PublishTaxonomyFunction(IContentService contentService)
                 executionContext.FunctionName,
                 message.ToString());
 
-            var context = new PublishContext(DateTime.UtcNow, false);
-
-            await _contentService.UpdateTaxonomy(context);
+            await _contentService.DeleteAllBlobCacheItems();
```
benoutram, line 32: There's other stuff cached, so I think we should come up with a better name for this, since we are not deleting all blob cache items which is a bit misleading.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Publisher/Functions/PublishTaxonomyFunction.cs
markhiveit: After running it past a couple of people, gone with `DeleteCachedTaxonomyBlobs`.

benoutram (approved):

PR #2914, opened 2021-10-01
EES-2463 Show release statuses for release's previous versions

markhiveit:
This PR changes the release statuses shown on the sign off page, such that statuses associated with previous versions of an amended release will also now appear.

A new "Release version" column has also been added, so it's clear which version of the release a particular status is associated with.

We find all previous versions of a release by searching for release's with the same publication, time identifier, and year. This seemed preferable to having to iterate through each release by checking its PreviousVersionId until we hit the original release.


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseApprovalService.cs
```diff
@@ -62,19 +62,31 @@ public class ReleaseApprovalService : IReleaseApprovalService
...
                 .OnSuccess(_userService.CheckCanViewReleaseStatusHistory)
                 .OnSuccess(release =>
-                    release.ReleaseStatuses
+                {
+                    var allReleaseVersionIds = _context.Releases
+                        .Where(r => r.PublicationId == release.PublicationId
+                                    && r.ReleaseName == release.ReleaseName
+                                    && r.TimePeriodCoverage == release.TimePeriodCoverage)
+                        .Select(r => r.Id)
+                        .ToList();
```
benoutram, line 71: I see why it's preferable to do this rather than iterating over the Publications releases because we think the overhead of doing so will be high, but just wondering what the difference would be?

I'm not really sure about this because there's no relationship between versions of a Release except for `PreviousVersionId`, so I would say this just happens to work 'ok' on the basis that analysts have been unlikely to change `Year` (ReleaseName) or `TimePeriodCoverage` between amendments in the past and same for in the future, rather than being the safest way of querying them.

Doesn't a bug exist here where by changing the release Year or Time identifier on an amendment, you end up missing the history of other amendments plus the history of the original version?

Similarly a bug where if you amend a different release of the same publication and change both the Year and Time Identifier to match another release, you effectively merge the status history into that unrelated release?

The data in Prod  looks like altering Year and Time Identifier in this has never happened but there are examples in other environments.

Two types of examples:
* Where the Year/Time identifier has changed between amendments of releases related to each other by previous version.
* Unrelated amendments within the same publication that use the the same Year and Time identifier.

If analysts shouldn't be changing these fields on amendments then maybe we should prevent it to stop the possibility of missing or incorrect status history in the future, given that it looks like it hasn't occurred in Prod yet, but even then getting them by PreviousVersionId is the safe way.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseRepositoryTests.cs
```diff
@@ -393,6 +393,43 @@ await using (var statisticsDbContext = InMemoryStatisticsDbContext(statisticsDbC
...
+            {
+                var releaseRepository = BuildReleaseRepository(contentDbContext: contentDbContext);
+                var result = await releaseRepository.GetAllReleaseVersionIds(amendedRelease2);
+                Assert.Equal(3, result.Count);
+
+                Assert.Contains(originalRelease.Id, result);
+                Assert.Contains(amendedRelease1.Id, result);
+                Assert.Contains(amendedRelease2.Id, result);
+            }
+        }
```
benoutram, line 431: I think you could add an additional test here for a release with no previous version id to check it can return a single version successfully.

src/explore-education-statistics-admin/src/pages/release/ReleaseStatusPage.tsx
```diff
@@ -169,6 +170,8 @@ const ReleaseStatusPage = () => {
                       </td>
                       <td>{status.approvalStatus}</td>
                       <td>{status.internalReleaseNote}</td>
+                      <td>{`${status.releaseVersion + 1}`}</td>{' '}
+                      {/* +1 because version starts from 0 in DB */}
```
benoutram, line 174: Just wondering if there's anything in `ReleaseStatusPage.test.tsx` or UI tests that can be updated for this?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServiceTests.cs
```diff
@@ -1071,7 +1179,8 @@ await using (var context = InMemoryApplicationDbContext(contextId))
                 contentService ?? Mock.Of<IContentService>(),
                 preReleaseUserService ?? Mock.Of<IPreReleaseUserService>(),
                 releaseFileRepository ?? new ReleaseFileRepository(contentDbContext),
-                releaseFileService ?? Mock.Of<IReleaseFileService>());
+                releaseFileService ?? Mock.Of<IReleaseFileService>(),
+                releaseRepository ?? Mock.Of<IReleaseRepository>());
```
benoutram, line 1183: You could use strict mocks to highlight any missing setup in the tests or anything being called unexpectedly.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServiceTests.cs
```diff
@@ -991,38 +996,140 @@ public async Task GetReleaseStatuses()
...
+                var releaseRepository = new ReleaseRepository(contentDbContext,
+                    Mock.Of<StatisticsDbContext>(),
+                    Mock.Of<IMapper>());
+                var releaseService = BuildService(contentDbContext, releaseRepository: releaseRepository);
+                var result = await releaseService.GetReleaseStatuses(amendedRelease.Id);
+
+                var resultStatuses = result.AssertRight();
+                Assert.Equal(2, resultStatuses.Count);
+
+                // originalRelease's ReleaseStatus appears first due the ReleaseStatus's Created date
```
benoutram, line 1121: Might want most recent first like we show the most recent data first in a lot of other places I think?

benoutram (changes requested):

PR #2921, opened 2021-10-05
EES-409 auto select single table tool options

amyb-hiveit:
In the table tool if there is only one option automatically select it so the user can just click 'Next step'. This applies to:
 - subject
 - location 
 - indicators (only applies if there are no filters)

No change for time periods as part of this, it's covered in a separate ticket.


markhiveit (changes requested):

src/explore-education-statistics-common/src/modules/table-tool/components/FiltersForm.tsx
```diff
@@ -52,6 +52,14 @@ const FiltersForm = (props: Props & InjectedWizardProps) => {
   } = props;
 
   const initialFormValues = useMemo(() => {
+    const indicators = Object.entries(subjectMeta.indicators);
+
+    // Automatically select indicator when no filters and one indicator group with one option
```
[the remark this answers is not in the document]
markhiveit: I ran this past Cam and he agreed that the ticket is incorrect and we should select a single indicator regardless of how many filters - and he also wants single filter items to be selected too. The ticket has been updated with details.

src/explore-education-statistics-common/src/modules/table-tool/components/FiltersForm.tsx
```diff
@@ -51,31 +51,57 @@ const FiltersForm = (props: Props & InjectedWizardProps) => {
     isActive,
   } = props;
 
+  // Automatically select filter when one filter group with one option
+  const filterEntries = Object.entries(subjectMeta.filters);
```
ntsim, line 55: Seems like in most cases where we're using `Object.entries`, we're only really using the property values, so we could make most of this code a little nicer by using `Object.values` instead as we don't need to access the values in the entries using `[1]` (not very readable).

Would do this for the code in `initialFormValues` too.

src/explore-education-statistics-common/src/modules/table-tool/components/FormFieldCheckboxGroupsMenu.tsx
```diff
@@ -9,11 +9,12 @@ import FormCheckboxSelectionCount from './FormCheckboxSelectedCount';
 interface Props<FormValues>
   extends FormFieldCheckboxSearchSubGroupsProps<FormValues> {
   legend: string;
+  openGroup?: boolean;
```
ntsim, line 12: Would be a little bit more succinct to call this `open`

src/explore-education-statistics-common/src/modules/table-tool/components/FormFieldCheckboxMenu.tsx
```diff
@@ -11,11 +11,12 @@ import FormCheckboxSelectionCount from './FormCheckboxSelectedCount';
 interface Props<FormValues>
   extends FormFieldCheckboxSearchGroupProps<FormValues> {
   legend: string;
+  openGroup?: boolean;
```
ntsim, line 14: Could rename this `open`

src/explore-education-statistics-common/src/modules/table-tool/components/FiltersForm.tsx
```diff
@@ -51,31 +51,57 @@ const FiltersForm = (props: Props & InjectedWizardProps) => {
...
+  const autoSelectFilter =
+    filterEntries.length === 1 &&
+    Object.entries(filterEntries[0][1].options).length === 1 &&
```
ntsim, line 58: We probably want to avoid excessively using `Object.entries` like this as I'm pretty sure that it's not a super cheap method to use (probably because it has to allocate a lot of arrays for all the entry pairs). 

As mentioned earlier, we should use `Object.values` instead, but would suggest that we allocate some additional variables for these intermediary values too. Looks like we'd be able to use them within `initialFormValues` as well.

src/explore-education-statistics-common/src/modules/table-tool/components/FiltersForm.tsx
```diff
@@ -51,31 +51,57 @@ const FiltersForm = (props: Props & InjectedWizardProps) => {
...
+          initialValues.filters.includes(filterValue),
+        );
+      }
+
+      if (filter.options.Default) {
+        // Automatically select filter option when there is only one and the filter is default
+        return filter.options.Default.options.length === 1
+          ? [filter.options.Default.options[0].value]
+          : [];
+      }
```
ntsim, line 95: Seems like this would be redundant now given that we have `autoSelectFilter`?

src/explore-education-statistics-common/src/modules/table-tool/components/LocationFiltersForm.tsx
```diff
@@ -57,22 +57,32 @@ const LocationFiltersForm = (props: Props & InjectedWizardProps) => {
...
+    Object.entries(formOptions)[0][1].options.length === 1;
+
+  const initialFormValues = useMemo(() => {
+    return {
+      locations: mapValues(options, (levelOptions, level) => {
+        const initialLevel = autoSelectLocation
+          ? [levelOptions.options[0].value]
+          : initialValues[level] ?? [];
+        return initialLevel.filter(
+          initialId =>
```
ntsim, line 72: `locationId` would probably be more appropriate here

src/explore-education-statistics-common/src/modules/table-tool/components/LocationFiltersForm.tsx
```diff
@@ -57,22 +57,32 @@ const LocationFiltersForm = (props: Props & InjectedWizardProps) => {
...
+            levelOptions.options.find(({ value }) => value === initialId) !==
+            undefined,
```
ntsim, line 74: Could use the `some` method here - will do the same thing and returns a boolean

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/FiltersForm.test.tsx
```diff
@@ -376,6 +446,28 @@ describe('FiltersForm', () => {
...
+      within(
+        screen.getByRole('group', {
+          name: 'Indicators - 1 selected',
+        }),
+      ).getAllByRole('checkbox'),
+    ).toHaveLength(1);
+
+    expect(
+      screen.getByLabelText('Number of excluded sessions'),
+    ).toHaveAttribute('checked');
```
ntsim, line 468: Same as above - can use `toBeChecked` method

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/FiltersForm.test.tsx
```diff
@@ -246,17 +286,47 @@ describe('FiltersForm', () => {
...
+      within(
+        screen.getByRole('group', {
+          name: 'Characteristic',
+        }),
+      ).getAllByRole('checkbox'),
+    ).toHaveLength(1);
+
+    expect(
+      screen.getByLabelText('Ethnicity Major Black Total'),
+    ).toHaveAttribute('checked');
```
ntsim, line 327: Can use the `toBeChecked` method which reads a little nicer

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/FiltersForm.test.tsx
```diff
@@ -246,17 +286,47 @@ describe('FiltersForm', () => {
...
+    expect(
+      within(
+        screen.getByRole('group', {
+          name: 'Characteristic',
+        }),
+      ).getAllByRole('checkbox'),
+    ).toHaveLength(1);
+
+    expect(
+      screen.getByLabelText('Ethnicity Major Black Total'),
```
ntsim, line 326: Might want to check that this checkbox is specifically the single one within the filter group

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/FiltersForm.test.tsx
```diff
@@ -376,6 +446,28 @@ describe('FiltersForm', () => {
...
+    expect(
+      within(
+        screen.getByRole('group', {
+          name: 'Indicators - 1 selected',
+        }),
+      ).getAllByRole('checkbox'),
+    ).toHaveLength(1);
+
+    expect(
+      screen.getByLabelText('Number of excluded sessions'),
```
ntsim, line 467: Same as above - might want to check this is the same checkbox as the one within the indicator group

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/LocationFiltersForm.test.tsx
```diff
@@ -0,0 +1,298 @@
...
+    expect(
+      within(
+        screen.getByRole('group', {
+          name: 'Region',
+          hidden: true,
+        }),
+      ).getAllByRole('checkbox', {
+        hidden: true,
+      }),
+    ).toHaveLength(2);
```
ntsim, line 122: Would suggest adding some extra checks to see if the options themselves are correctly labelled and ordered for each group. We should be precise about which group each option is in by using `within` as well.

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/LocationFiltersForm.test.tsx
```diff
@@ -0,0 +1,298 @@
...
+        onSubmit={noop}
+        initialValues={{
+          LocalAuthority: ['local-authority-1', 'local-authority-3'],
+          Region: ['region-2'],
+        }}
+      />,
+    );
+
+    expect(screen.getByLabelText('Local authority 1')).toHaveAttribute(
+      'checked',
```
ntsim, line 203: For all of these we could use the `toBeChecked` method - same for any other place in this test suite too.

PR #2925, opened 2021-10-12
EES-2462 Allow bulk upload of PreRelease Invites

benoutram:
This PR alters the 'Invite new user by email' field of the Prerlease users tab in the Admin be a textarea so that a bulk list of email addresses can be invited at a time.

EES-2461

A confirmation modal is shown which the analyst has to accept before inviting the users. This prevents accidently needing to unpick a large number of entries if a mistake is made, and is also useful since email notifications are sent out immediately if the release is approved.

EES-2461-2

The confirmation modal has a note about when emails will be sent which changes if the release is approved.

EES-2461-3

All email addresses must be valid @education.gov.uk email addresses.

If any email addresses are already accepted or invited they will be shown in the modal confirmation as being ignored.

EES-2461-4

### Other changes:
* Secure method PreReleaseUserService#SendPreReleaseUserInviteEmails which previously has no checks.
* Add missing `PreReleaseUserServicePermissionTests` unit tests.
* Change all methods of `EmailService` and `EmailTemplateService` to return `ActionResult` in the event of an error or `Unit` in the event of success. In future me might want to store the notification message id returned by Gov.Uk Notify so that we can poll and display the delivery status of emails.
  * Change all the consumers of `EmailService` and `EmailTemplateService` to respect the success / failure results including where multiple emails are sent in a batch.
* Remove unused `CreatedBy` field of `UserInvite` which only has null or blank values for every invite. Prod database checked ✔️. This field also had no relationship with `User`.
  * This has been replaced by a new `CreatedById` field to `UserInvite` for recording the User that created the invite, correctly referenced to the application User and set with the value from `UserService.GetUserId`.
* Create a new `Role` enum, refactoring everywhere which looked up db Roles or relied on hardcoded constants such as the BAU * Role Guid or a role name like 'Prerelease User'.
* Add new `UserRepository` for convenience of finding Users by email.
* Refactor PreReleaseUserService methods to have a much simpler flow for inviting Prerelease users
  * Call to `CreateUserInvite` if we don't have an existing user.
  * Call either a `CreateUserReleaseInvite` or `CreateAcceptedUserReleaseInvite` depending on whether the user exists or not.
* Applied missing UTC `DateTime` conversion to the `UserReleaseInvite` and `UserInvite` `CreatedDate` fields.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PreReleaseUserServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PreReleaseUserServiceTests.cs
@@ -17,12 +17,14 @@ using GovUk.Education.ExploreEducationStatistics.Common.Utils;
 using GovUk.Education.ExploreEducationStatistics.Content.Model;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Configuration;
 using Moq;
 using Xunit;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.DbUtils;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Validators.ValidationErrorMessages;
+using static GovUk.Education.ExploreEducationStatistics.Common.Services.CollectionUtils;
 using static GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils.MockUtils;
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
@@ -211,9 +213,11 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                 var service = SetupPreReleaseUserService(context, usersAndRolesDbContext: userAndRolesDbContext);
                 var result = await service.GetPreReleaseUsersInvitePlan(
                     release.Id,
-                    @"test1@test.com
-                           not an email
-                           test2@test.com");
+                    ListOf(
+                        "test1@test.com",
+                        "not an email",
+                        "test2@test.com")
+                );
 
                 result.AssertBadRequest(InvalidEmailAddress);
             }
@@ -225,7 +229,10 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
             await using var context = InMemoryApplicationDbContext();
             await using var userAndRolesDbContext = InMemoryUserAndRolesDbContext();
             var service = SetupPreReleaseUserService(context, usersAndRolesDbContext: userAndRolesDbContext);
-            var result = await service.GetPreReleaseUsersInvitePlan(Guid.NewGuid(), "test@test.com");
+            var result = await service.GetPreReleaseUsersInvitePlan(
+                Guid.NewGuid(),
+                ListOf("test@test.com")
+            );
 
             result.AssertNotFound();
         }
@@ -268,8 +275,10 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                 var service = SetupPreReleaseUserService(context, usersAndRolesDbContext: userAndRolesDbContext);
                 var result = await service.GetPreReleaseUsersInvitePlan(
                     release.Id,
-                    @"invited.prerelease@test.com
-                           existing.prerelease.user@test.com");
+                    ListOf(
+                        "invited.prerelease@test.com",
+                        "existing.prerelease.user@test.com")
+                );
 
                 result.AssertBadRequest(NoInvitableEmails);
             }
@@ -353,14 +362,15 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
 
                 var result = await service.GetPreReleaseUsersInvitePlan(
                     release.Id,
-                    @"new.user.1@test.com
-                           new.user.2@test.com
-                           existing.user.1@test.com
-                           existing.user.2@test.com
-                           invited.prerelease.1@test.com
-                           invited.prerelease.2@test.com
-                           existing.prerelease.user.1@test.com
-                           existing.prerelease.user.2@test.com"
+                    ListOf(
+                        "new.user.1@test.com",
+                        "new.user.2@test.com",
+                        "existing.user.1@test.com",
+                        "existing.user.2@test.com",
+                        "invited.prerelease.1@test.com",
+                        "invited.prerelease.2@test.com",
+                        "existing.prerelease.user.1@test.com",
+                        "existing.prerelease.user.2@test.com")
                 );
 
                 var plan = result.AssertRight();
@@ -399,9 +409,11 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                 var service = SetupPreReleaseUserService(context, usersAndRolesDbContext: userAndRolesDbContext);
                 var result = await service.InvitePreReleaseUsers(
                     release.Id,
-                    @"test1@test.com
-                           not an email
-                           test2@test.com");
+                    ListOf(
+                        "test1@test.com",
+                        "not an email",
+                        "test2@test.com")
+                );
 
                 result.AssertBadRequest(InvalidEmailAddress);
             }
@@ -413,7 +425,10 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
             await using var context = InMemoryApplicationDbContext();
             await using var userAndRolesDbContext = InMemoryUserAndRolesDbContext();
             var service = SetupPreReleaseUserService(context, usersAndRolesDbContext: userAndRolesDbContext);
-            var result = await service.InvitePreReleaseUsers(Guid.NewGuid(), "test@test.com");
+            var result = await service.InvitePreReleaseUsers(
+                Guid.NewGuid(),
+                ListOf("test@test.com")
+            );
 
             result.AssertNotFound();
         }
@@ -456,8 +471,10 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                 var service = SetupPreReleaseUserService(context, usersAndRolesDbContext: userAndRolesDbContext);
                 var result = await service.InvitePreReleaseUsers(
                     release.Id,
-                    @"invited.prerelease@test.com
-                           existing.prerelease.user@test.com");
+                    ListOf(
+                        "invited.prerelease@test.com",
+                        "existing.prerelease.user@test.com")
+                );
 
                 result.AssertBadRequest(NoInvitableEmails);
             }
@@ -505,7 +522,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                 .Returns(Unit.Instance);
 
             var preReleaseService = new Mock<IPreReleaseService>(MockBehavior.Strict);
-            SetupGetPrereleaseWindowForRelease(preReleaseService, release);
+            SetupGetPrereleaseWindow(preReleaseService, release);
 
             await using (var context = InMemoryApplicationDbContext(contextId))
             await using (var userAndRolesDbContext = InMemoryUserAndRolesDbContext(contextId))
@@ -519,7 +536,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
 
                 var result = await service.InvitePreReleaseUsers(
                     release.Id,
-                    "test@test.com"
+                    ListOf("test@test.com")
                 );
 
                 emailService.Verify(
@@ -557,6 +574,77 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                 Assert.Equal(ReleaseRole.PrereleaseViewer, releaseInvite.Role);
                 Assert.True(releaseInvite.Accepted); // User already exists, so permission has been applied immediately
                 Assert.True(releaseInvite.EmailSent); // Email sent immediately for approved releases
+            }
+        }
+
+        [Fact]
+        public async Task InvitePreReleaseUsers_FailsSendingEmail_ExistingUser_ApprovedRelease()
+        {
+            var release = new Release
+            {
+                ReleaseName = "2020",
+                TimePeriodCoverage = TimeIdentifier.CalendarYear,
+                PublishScheduled = PublishedScheduledStartOfDay,
+                Publication = new Publication
+                {
+                    Title = "Test publication"
+                },
+                ApprovalStatus = ReleaseApprovalStatus.Approved,
+            };
+
+            var user = new User
+            {
+                Email = "test@test.com"
+            };
+
+            var contextId = Guid.NewGuid().ToString();
+
+            await using (var context = InMemoryApplicationDbContext(contextId))
+            {
+                await context.AddAsync(release);
+                await context.AddAsync(user);
+
+                await context.SaveChangesAsync();
+            }
+
+            var emailService = new Mock<IEmailService>(MockBehavior.Strict);
+
+            var expectedTemplateValues = GetExpectedPreReleaseTemplateValues(release, newUser: false);
+
+            emailService.Setup(mock => mock.SendEmail(
+                    "test@test.com",
+                    PreReleaseTemplateId,
+                    expectedTemplateValues
+                ))
+                .Returns(new BadRequestResult());
+
+            var preReleaseService = new Mock<IPreReleaseService>(MockBehavior.Strict);
+            SetupGetPrereleaseWindow(preReleaseService, release);
+
+            await using (var context = InMemoryApplicationDbContext(contextId))
+            await using (var userAndRolesDbContext = InMemoryUserAndRolesDbContext(contextId))
+            {
+                var service = SetupPreReleaseUserService(
+                    context,
+                    usersAndRolesDbContext: userAndRolesDbContext,
+                    preReleaseService: preReleaseService.Object,
+                    emailService: emailService.Object
+                );
+
+                var result = await service.InvitePreReleaseUsers(
+                    release.Id,
+                    ListOf("test@test.com")
+                );
+
+                VerifyAllMocks(emailService, preReleaseService);
+
+                result.AssertLeft();
+            }
+
+            await using (var context = InMemoryApplicationDbContext(contextId))
+            {
+                Assert.Empty(context.UserReleaseRoles); 
+                Assert.Empty(context.UserReleaseInvites);
             }
         }
 
@@ -600,7 +688,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
 
                 var result = await service.InvitePreReleaseUsers(
                     release.Id,
-                    "test@test.com"
+                    ListOf("test@test.com")
                 );
 
                 var preReleaseUsers = result.AssertRight();
@@ -665,7 +753,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                 .Returns(Unit.Instance);
 
             var preReleaseService = new Mock<IPreReleaseService>(MockBehavior.Strict);
-            SetupGetPrereleaseWindowForRelease(preReleaseService, release);
+            SetupGetPrereleaseWindow(preReleaseService, release);
 
             await using (var context = InMemoryApplicationDbContext(contextId))
             await using (var userAndRolesDbContext = InMemoryUserAndRolesDbContext(contextId))
@@ -679,7 +767,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
 
                 var result = await service.InvitePreReleaseUsers(
                     release.Id,
-                    "test@test.com"
+                    ListOf("test@test.com")
                 );
 
                 emailService.Verify(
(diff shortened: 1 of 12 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PreReleaseUserService.cs
```diff
@@ -231,130 +234,185 @@ public class PreReleaseUserService : IPreReleaseUserService
...
+                    {
+                        Email = email.ToLower(),
+                        ReleaseId = release.Id,
+                        Role = ReleaseRole.PrereleaseViewer,
+                        Accepted = true,
+                        EmailSent = sendEmail,
+                        Created = UtcNow,
+                        CreatedById = _userService.GetUserId()
+                    }
+                );
```
ntsim, line 337: Similar to above - shouldn't we wait for the result of `SendPreReleaseInviteEmail` before saving the invite into the database as sent?

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseInvitePlanModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+              : 'when the release is approved for publication.'}
+          </WarningMessage>
+          <div className={styles.invitesOverflow}>
+            <ul
+              className="govuk-list govuk-list--bullet govuk-!-margin-2"
+              data-testid="invitableList"
+            >
+              {invitePlan.invitable?.map(email => (
+                <li key={email}>
+                  <p>{email}</p>
```
ntsim, line 54: No need to wrap the email in a `p`, it just adds unnecessary spacing between the list items. `<li>{email}</li>` is fine.

Would do this for the other lists too.

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseInvitePlanModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+}
+
+const PreReleaseInvitePlanModel = ({
+  invitePlan,
+  isReleaseApproved,
+  onConfirm,
+  onCancel,
+  onExit,
+}: Props) => {
+  if (!invitePlan) {
```
ntsim, line 22: Is this actually optional? The prop type isn't optional, so it doesn't seem like we need this check?

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseInvitePlanModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+        <>
+          <WarningMessage>
+            Email notifications will be sent{' '}
+            {isReleaseApproved
+              ? 'immediately.'
+              : 'when the release is approved for publication.'}
+          </WarningMessage>
+          <div className={styles.invitesOverflow}>
+            <ul
+              className="govuk-list govuk-list--bullet govuk-!-margin-2"
```
ntsim, line 49: No need to add `govuk-list govuk-list--bullet` classes - these are automatically applied to `ul` globally. 

Would do this for the other lists too.

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseInvitePlanModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+        pre-release?
+      </p>
+
+      {invitePlan.invitable && invitePlan.invitable.length > 0 ? (
+        <>
+          <WarningMessage>
+            Email notifications will be sent{' '}
+            {isReleaseApproved
+              ? 'immediately.'
+              : 'when the release is approved for publication.'}
```
ntsim, line 45: This will split the text over multiple text nodes in the DOM, making it harder for tests to pick this up as a single line of text. Instead, we could use a template string like so:

```ts
{`Email notification will be sent ${isReleaseApproved 
  ? 'immediately.' 
  : 'when the release is approved for publication.'}`}
```

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseInvitePlanModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+          <h2 id="already-accepted-heading" className="govuk-heading-m">
+            Already accepted
+          </h2>
+          <p>
+            The following email addresses will be ignored as they are already
+            accepted to access the pre-release:
+          </p>
+          <div className={styles.invitesOverflow}>
+            <ul
+              aria-labelledby="already-accepted-heading"
```
ntsim, line 78: How come we're using `aria-labelledby` here? (same for the 'Already invited' list)

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseInvitePlanModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+          </div>
+        </>
+      ) : (
+        <WarningMessage>
+          There are no email addresses that can be invited to access the
+          pre-release.
+        </WarningMessage>
+      )}
+
+      {invitePlan.alreadyAccepted && invitePlan.alreadyAccepted?.length > 0 && (
```
ntsim, line 67: Don't think we need the optional chain (`?.`) here as we already check that `alreadyAccepted` exists

src/explore-education-statistics-admin/src/services/preReleaseUserService.ts
```diff
@@ -4,13 +4,27 @@ export interface PreReleaseUser {
   email: string;
 }
 
+export interface PreReleaseInvitePlan {
+  alreadyAccepted?: string[];
+  alreadyInvited?: string[];
+  invitable?: string[];
```
ntsim, line 10: Are these supposed to be optional? Seems like the backend API always returns empty lists (which would be the right thing to do)? 

If these aren't optional, then I would also adjust the confirmation modal checks as we could simplify checks like:
`invitePlan.alreadyInvited && invitePlan.alreadyInvited.length > 0` to simply ` invitePlan.alreadyInvited.length > 0`

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
```diff
@@ -14,30 +16,36 @@ import useToggle from '@common/hooks/useToggle';
...
 }
 
 const errorMappings = [
   mapFieldErrors<FormValues>({
-    target: 'email',
+    target: 'emails',
     messages: {
-      USER_ALREADY_EXISTS: 'User with this email already exists',
+      INVALID_EMAIL_ADDRESS:
+        'Please enter valid @education.gov.uk email addresses',
```
ntsim, line 30: Would probably re-word this message a little. For a start, we don't need to use 'please' as it implies there is a choice. There's an entire section advising on how to write error messages in the GOV.UK [Design System](https://design-system.service.gov.uk/components/error-message/#be-clear-and-concise) that's worth reading.

In this instance, I'd probably suggest something like 'Enter only @education.govuk email addresses'

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
```diff
@@ -14,30 +16,36 @@ import useToggle from '@common/hooks/useToggle';
...
+      NO_INVITABLE_EMAILS:
+        'All of the email addresses are already invited or accepted users of the pre-release',
```
ntsim, line 32: Wording is a little confusing - perhaps 'All of the email addresses have already been invited or accepted'

I've excluded the additional part about pre-release because it's a bit redundant in the context of where the form is.

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
```diff
@@ -51,18 +59,31 @@ const PreReleaseUserAccessForm = ({
...
+  }, errorMappings);
 
-    setUsers({
-      value: [...users, newUser],
-    });
+  const handleModalSubmit = useFormSubmit<FormValues>(
+    async (values, actions) => {
+      await preReleaseUserService
+        .inviteUsers(releaseId, values.emails)
+        .then(newUsers => {
```
ntsim, line 74: No need to use `then` as we're in an async/await block. Can be rewritten as:

```ts
const newUsers = await preReleaseUserService.inviteUsers(releaseId, values.emails);

setUsers({
  value: [...users, ...newUsers],
});

actions.resetForm();
```

src/explore-education-statistics-admin/src/services/preReleaseUserService.ts
```diff
@@ -4,13 +4,27 @@ export interface PreReleaseUser {
...
+}
+
 const preReleaseUserService = {
   getUsers(releaseId: string): Promise<PreReleaseUser[]> {
     return client.get(`/release/${releaseId}/prerelease-users`);
   },
-  inviteUser(releaseId: string, email: string): Promise<PreReleaseUser> {
+  getInvitePlan(
+    releaseId: string,
+    emails: string,
```
ntsim, line 19: As mentioned in the backend - would update this so that we send a list of strings rather than the string of email lines. Same with the `inviteUsers` method.

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
```diff
@@ -81,28 +102,43 @@ const PreReleaseUserAccessForm = ({
...
-            email: '',
+            emails: '',
           }}
           validationSchema={Yup.object<FormValues>({
-            email: Yup.string()
-              .required('Enter an email address')
-              .email('Enter a valid @education.gov.uk email address')
+            emails: Yup.string()
+              .trim()
+              .required('Please enter 1 or more email addresses')
```
ntsim, line 110: No need to use 'please'

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
```diff
@@ -81,28 +102,43 @@ const PreReleaseUserAccessForm = ({
...
+              .test({
+                name: 'number of lines',
+                message: `Please enter between 1 and ${inviteLimit} lines`,
```
ntsim, line 113: No need to use 'please'. Would also reword this to 'Enter between 1 and 50 lines of email addresses'

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
```diff
@@ -81,28 +102,43 @@ const PreReleaseUserAccessForm = ({
...
+                    const numOfLines = (value.match(/\n/g) || '').length + 1;
+                    return numOfLines <= inviteLimit;
+                  }
+                  return true;
+                },
+              })
               .test({
                 name: 'email format',
-                message: 'Enter a valid @education.gov.uk email address',
+                message: 'Please enter valid @education.gov.uk email addresses',
```
ntsim, line 124: No need to use 'please' - would suggest rewording this to 'Enter only @education.govuk email addresses' (or whatever the server-side error message is)

src/explore-education-statistics-admin/src/pages/release/pre-release/components/__tests__/PreReleaseUserAccessForm.test.tsx
```diff
@@ -97,78 +97,122 @@ describe('PreReleaseUserAccessForm', () => {
...
         ).toBeInTheDocument();
       });
 
+      const emailsTextarea = screen.getByLabelText('Invite new users by email');
+      // type values up to but not exceeding the limit of lines
       await userEvent.type(
-        screen.getByLabelText('Invite new user by email'),
-        'not a valid email',
+        emailsTextarea,
+        `test@education.gov.uk{enter}`.repeat(50),
```
ntsim, line 139: Very cool. Didn't know you could do this in the `user-event` library! :smile:

src/explore-education-statistics-admin/src/pages/release/pre-release/components/__tests__/PreReleaseUserAccessForm.test.tsx
```diff
@@ -206,58 +253,248 @@ describe('PreReleaseUserAccessForm', () => {
...
+
+      await waitFor(() => {
+        expect(preReleaseUserService.getInvitePlan).toHaveBeenCalledWith(
+          'release-1',
+          'test1@education.gov.uk\ntest2@education.gov.uk\ntest3@education.gov.uk',
+        );
+      });
+
+      await waitFor(() => {
+        expect(screen.getByRole('dialog')).toBeInTheDocument();
```
ntsim, line 313: Mentioned this to Amy before, but `getByRole` is a little problematic when used within `waitFor`. It's not a quick selector as it has to do a bunch of visibility checks that are really slow (currently).

This means that if the component is big enough, it can cause `waitFor` (this is essentially just a loop that retries the callback every x milliseconds) to get jammed up with callbacks that are still processing. You can often see tests that are written like this can be noticeably slower, or in extreme cases, be unable to finish within the default 10 second time limit!

Consequently, my current advice is to use faster selectors such as `getByText` or `getByLabel` within the `waitFor` and use `getByRole` outside.

In this specific case, we might do something do something like this:

```ts
await waitFor(() => {
  expect(screen.getByText('Confirm pre-release invitations')).toBeInTheDocument();
});

// Now we can use `getByRole`
const modal = within(screen.getByRole('dialog'));

// We might also check that it's a heading within the modal
expect(modal.getByRole('heading', { name: 'Confirm pre-release invitations' }))
  .toBeInTheDocument();
```
Would suggest going through these tests and checking for any other such cases and refactoring them accordingly.

tests/robot-tests/tests/admin/bau/prerelease.robot
```diff
@@ -155,18 +155,79 @@ Go to prerelease access page
...
+Invite users to the prerelease
+    ${emails}=    Catenate    SEPARATOR=\n
+    ...    simulate-delivered@notifications.service.gov.uk
+    ...    ees-analyst1@education.gov.uk
+    user enters text into element    css:textarea[name="emails"]    ${emails}
+    user clicks button    Invite new users
+    ${modal}=    user waits until modal is visible    Confirm pre-release invitations
+    user waits until element contains    ${modal}    Email notifications will be sent immediately
+
+    user waits until page contains element    css:[data-testid="invitableList"] li    limit=2
```
ntsim, line 181: Could use the `user checks list has x items` keyword for this (and other similar instances):

```robot
user checks list has x items  testid:invitableList  2
```

Note that we're using our custom `testid` selector, which is less verbose than the equivalent `css` or `xpath` selectors

tests/robot-tests/tests/admin/bau/prerelease.robot
```diff
@@ -155,18 +155,79 @@ Go to prerelease access page
...
+    user waits until element contains    css:[data-testid="invitableList"] li:nth-of-type(1) p
```
ntsim, line 182: Assuming we remove the wrapping `p` tag, we could use the `user checks list item contains` keyword for this and other similar instances:

```robot
user checks list item contains  testid:invitableList  1  simulate-delivered@notifications.service.gov.uk
```

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PreReleaseUserService.cs
```diff
@@ -103,61 +98,71 @@ public class PreReleaseUserService : IPreReleaseUserService
                 );
         }
 
-        public async Task<Either<ActionResult, PreReleaseUserViewModel>> AddPreReleaseUser(
+        public async Task<Either<ActionResult, PreReleaseInvitePlan>> GetPreReleaseUsersInvitePlan(
             Guid releaseId,
-            string email)
+            string emails)
```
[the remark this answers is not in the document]
benoutram: Cool, have changed the API to expect an array and construct that array in the browser instead.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PreReleaseUserServiceTests.cs
```diff
@@ -1185,102 +1562,97 @@ await using (var context = DbUtils.InMemoryApplicationDbContext(contextId))
...
+                {"prerelease time", "09:30"},
+                {"publish day", "Wednesday 09 September 2020"},
+                {"publish time", "09:30"}
+            };
+        }
 
-            configuration
-                .Setup(m => m.GetSection("NotifyPreReleaseTemplateId"))
-                .Returns(section.Object);
+        private static void SetupGetPrereleaseWindowForRelease(Mock<IPreReleaseService> preReleaseService,
```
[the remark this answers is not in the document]
benoutram: Thanks, have simplified this name now as suggested. 👍

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PreReleaseUserService.cs
```diff
@@ -413,10 +461,40 @@ private static string FormatTimeForEmail(TimeSpan timeSpan)
...
+    public class PreReleaseInvitePlan
+    {
+        public List<string> AlreadyInvited { get; } = new List<string>();
+
+        public List<string> AlreadyAccepted { get; } = new List<string>();
+
+        public List<string> Invitable { get; } = new List<string>();
     }
 
     public class PreReleaseUserViewModel
```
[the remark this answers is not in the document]
benoutram: Also moved it to `PreReleaseUserViewModels.cs`

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseInvitePlanModal.tsx
benoutram: Cool, was meaning to do this to get rid of the extra spacing. Thanks 🙂

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseInvitePlanModal.tsx
benoutram: Great tip, I was struggling with this in a test and was having difficulty preventing it getting reformatted over multiple lines. I've used your method of templating it now. Thanks.

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
benoutram: This is hilarious because I actually spent time reading articles about good error messages and being polite before writing these but totally missed the point about not using 'please'. I will remember to void errors in the imperative in future. Thanks.

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
benoutram: Abbreviated this one to your suggested text. Thanks.

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
benoutram: That's good to know. Thanks for the tip.

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
benoutram: Updated to `Enter between 1 and 50 lines of email addresses`. Thanks.

tests/robot-tests/tests/admin/bau/prerelease.robot
benoutram: Cool, I didn't know about our `testid` selector, I've made this change. Thanks 🙂

tests/robot-tests/tests/admin/bau/prerelease.robot
benoutram: Yep, I've made this change, thanks for the suggestion.

src/GovUk.Education.ExploreEducationStatistics.Admin/Models/Role.cs
```diff
@@ -0,0 +1,16 @@
+﻿#nullable enable
+using GovUk.Education.ExploreEducationStatistics.Common.Database;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Models
+{
+    // TODO EES-2462 Consider making this a class with Name and Id properties
```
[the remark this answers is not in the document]
benoutram: Where we want to compare by Id we get the Id by
`Role.PrereleaseUser.GetEnumValue()`

Where we want to compare by name we get the name by 
`Role.PrereleaseUser.GetEnumLabel()`

I just thought it might be nicer to be able to get them by `Role.PrereleaseUser.Id` and `Role.PrereleaseUser.Name`, but I didn't make a decision on this so left in this TODO.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PreReleaseUserServiceTests.cs
```diff
@@ -544,184 +647,493 @@ public async Task AddPreReleaseUser_InvitesNewUser_ApprovedRelease()
...
+                // Four new release invites should have been created
+                var releaseInvites = await context.UserReleaseInvites.AsQueryable().ToListAsync();
+                Assert.Equal(existingReleaseInvites.Count + 4, releaseInvites.Count);
+
+                Assert.True(releaseInvites.All(invite => invite.ReleaseId == release.Id));
+                Assert.True(releaseInvites.All(invite => invite.Role == ReleaseRole.PrereleaseViewer));
+                Assert.True(releaseInvites.All(invite =>
+                    !invite.EmailSent)); // Emails are not sent for unapproved releases
+
+                Assert.True(releaseInvites.Exists(invite => invite.Email == "new.user.1@test.com" && !invite.Accepted));
```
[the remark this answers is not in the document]
benoutram: Thanks I've replaced it with `Exists` now but see previous comment on why I did it like this.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserRepositoryTests.cs
```diff
@@ -0,0 +1,86 @@
...
+                await contentDbContext.SaveChangesAsync();
+            }
+
+            await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContextId))
+            {
+                var repository = SetupUserReleaseRoleRepository(contentDbContext);
+                var result = await repository.FindByEmail("TEST@TEST.COM");
+
+                // While this test could fail using the in-memory db if the comparison was accidentally case sensitive,
+                // it wouldn't fail using SqlServer which is using a case insensitive collation by default.
```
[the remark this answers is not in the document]
benoutram: Case insensitive collation `SQL_Latin1_General_CP1_CI_AS` is the default for SQL Server. I've checked Azure Portal which confirms this is what we have set. We would have had to set this when creating the database and I'm not sure how safe or easy it is to change it now. Setting a different collation when initialising the database schema such as this case sensitive one `SQL_Latin1_General_CP1_CS_AS` became possible in EF 5.0 using `modelBuilder.UseCollation("SQL_Latin1_General_CP1_CS_AS")`. Since EF 5.0 its also possible to set it on individual columns.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PreReleaseUserService.cs
```diff
@@ -231,130 +234,185 @@ public class PreReleaseUserService : IPreReleaseUserService
...
+                    {
+                        Email = email.ToLower(),
+                        ReleaseId = release.Id,
+                        Role = ReleaseRole.PrereleaseViewer,
+                        EmailSent = sendEmail,
+                        Created = UtcNow,
+                        CreatedById = _userService.GetUserId()
+                    }
+                );
+                await _context.SaveChangesAsync();
```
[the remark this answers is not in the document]
benoutram: Thanks, I've made this so that it never gets saved if the email is not sent. I also reordered things so that the UserInvite in the `UsersAndRolesDbContext` is not created either if this fails and added a unit test for this.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PreReleaseUserService.cs
benoutram: Good point, thanks. Altered this so that it doesn't get saved if the email fails to send.

PR #2928, opened 2021-10-13
EES-2687 - added GA download files event to new Data Catalogue page

duncan-at-hiveit:
This PR:
- adds a new file download GA event to the Data Catalogue page
- amends existing file download GA events to send Publication title and Release name separately, to bring in line with the format of the new event in the Data Catalogue page

### DIFF
--- a/src/explore-education-statistics-frontend/.env
+++ b/src/explore-education-statistics-frontend/.env
@@ -1,4 +1,4 @@
-CONTENT_API_BASE_URL=http://localhost:5010/api      # or http://ees.local:5010/api
-DATA_API_BASE_URL=http://localhost:5000/api         # or http://ees.local:5000/api
+CONTENT_API_BASE_URL=http://localhost:5010/api
+DATA_API_BASE_URL=http://localhost:5000/api
 NOTIFICATION_API_BASE_URL=http://localhost:7073/api
-GA_TRACKING_ID=                                     # Available from development GA account
+GA_TRACKING_ID=
--- a/src/explore-education-statistics-frontend/.env
+++ b/src/explore-education-statistics-frontend/.env
@@ -1,3 +1,4 @@
-CONTENT_API_BASE_URL=http://localhost:5010/api
-DATA_API_BASE_URL=http://localhost:5000/api
+CONTENT_API_BASE_URL=http://localhost:5010/api      # or http://ees.local:5010/api
+DATA_API_BASE_URL=http://localhost:5000/api         # or http://ees.local:5000/api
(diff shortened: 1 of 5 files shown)


src/explore-education-statistics-frontend/.env
```diff
@@ -1,3 +1,4 @@
 CONTENT_API_BASE_URL=http://localhost:5010/api
 DATA_API_BASE_URL=http://localhost:5000/api
 NOTIFICATION_API_BASE_URL=http://localhost:7073/api
+GA_TRACKING_ID=
```
benoutram, line 4: Presumably adding this to the `.env` file with no value makes no difference to the initialisation of GA because the empty string will be falsy in `if (process.env.GA_TRACKING_ID && ...`?

Just checking 🙂

src/explore-education-statistics-frontend/src/modules/data-catalogue/DataCataloguePage.tsx
```diff
@@ -112,11 +113,24 @@ const DataCataloguePage: NextPage<Props> = ({
...
       throw new Error('Release has not been selected');
     }
 
-    await downloadService.downloadFiles(state.query.release.id, files);
+    await downloadService.downloadFiles(release.id, files);
+
+    const availableFiles = state.subjects.map(subject => subject.file);
+    const selectedFilenames = files
+      .map(fileId => availableFiles.find(file => file.id === fileId)?.fileName)
+      .join(', ');
```
benoutram, line 127: I suppose there's the potential for something like `'file a, file b, , file d` if an `undefined` creeps in the result of the `find` but that should never happen?

src/explore-education-statistics-frontend/src/modules/data-catalogue/DataCataloguePage.tsx
```diff
@@ -112,11 +113,24 @@ const DataCataloguePage: NextPage<Props> = ({
...
+
+    logEvent({
+      category: 'Downloads',
+      action: 'Data Catalogue page selected files download',
+      label: `Publication: ${state.query.publication?.title}, Release: ${release.title} File: ${selectedFilenames}`,
```
benoutram, line 132: I guess we don't want to change the label value prefix for the filenames from `File: ` to `Files: ` here because it might upset filtering on historical data already in GA?

src/explore-education-statistics-frontend/.env
duncan-at-hiveit: Yup that's correct!  Nick also pointed out that we can add actual values to a .env.local file and keep out of Git

PR #2929, opened 2021-10-13
EES-2758 Add geometry data 'Counties and Unitary Authorities (May 20211) UK BUC'

benoutram:


src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20211012154859_GeometryData.sql
```diff
@@ -0,0 +1,219 @@
+SET IDENTITY_INSERT geometry ON;
```
duncan-at-hiveit, line 1: Out of interest, is this geometry table a predefined table structure specifically for these file types that we get from Geodata sites?  Or did you have to do any magic to get this into the right format?

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20211012154859_EES-2758_Add_LA_Boundary_Data.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace GovUk.Education.ExploreEducationStatistics.Data.Model.Migrations
+{
+    public partial class EES2758_Add_LA_Boundary_Data : Migration
+    {
+        private const string MigrationId = "20211012154859";
+
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql("SET IDENTITY_INSERT BoundaryLevel ON");
```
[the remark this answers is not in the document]
duncan-at-hiveit: Also, does just the presence of this data in the db automatically allow it to be selected by users when building maps?

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20211012154859_GeometryData.sql
duncan-at-hiveit: Ahh nice - that's the sort of magic that I didn't know we did until just now!

PR #2932, opened 2021-10-15
EES-2804 🐛Delete 'All methodologies' cache item so that methodologies are published immeditately

benoutram:
This PR configures the `BlobStorageService` used by the `BlobCacheService` of `MethodologyApprovalService` to use the `PublicStorage` connection to Azure Blob Storage so that we are deleting the 'All Methodologies' cache item from the public storage account not the private one.

It was previously configured to use `CoreStorage`.

Using the wrong connection caused a bug where the 'All Methodologies' page was not showing new methodology versions after they were published immediately because the cache file wasn't being deleted.

### Other changes
  * Add missing UTC DateTime conversion to `Published` and `Updated` fields of `Publication`. Although this makes no difference to the bugfix, I originally spotted this when investigating the bug and evaluating the conditions for when methodologies are made publicly accessible.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
@@ -162,7 +162,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Model.Database
             modelBuilder.Entity<MethodologyNote>()
                 .Property(n => n.Updated)
                 .HasConversion(
-                    v => v, 
+                    v => v,
                     v => v.HasValue ? DateTime.SpecifyKind(v.Value, DateTimeKind.Utc) : (DateTime?) null);
 
             modelBuilder.Entity<MethodologyNote>()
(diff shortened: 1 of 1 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyApprovalServicePermissionTests.cs
```diff
@@ -73,7 +73,7 @@ await PolicyCheckBuilder<SecurityPolicies>()
         private MethodologyApprovalService SetupService(
             ContentDbContext? contentDbContext = null,
             IPersistenceHelper<ContentDbContext>? persistenceHelper = null,
-            IBlobCacheService? blobCacheService = null,
+            IBlobCacheService? publicBlobCacheService = null,
```
duncan-at-hiveit, line 76: Good to have that named explicitly :+1:

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
```diff
@@ -289,7 +289,19 @@ public void ConfigureServices(IServiceCollection services)
             services.AddTransient<IMethodologyFileRepository, MethodologyFileRepository>();
             services.AddTransient<IMethodologyImageService, MethodologyImageService>();
             services.AddTransient<IMethodologyAmendmentService, MethodologyAmendmentService>();
-            services.AddTransient<IMethodologyApprovalService, MethodologyApprovalService>();
+            services.AddTransient<IMethodologyApprovalService, MethodologyApprovalService>(provider =>
```
duncan-at-hiveit, line 292: Oh momma - all to wire in a specific BlobStorageService :smile:  Be nice to see this go!

duncan-at-hiveit (approved):

PR #2933, opened 2021-10-15
EES-2810 - fix transient failure in `create_methodology_amendment.robot`

luke-h1:
This PR: 
- fixes a transient failure in the 'remove an image from content block' test case in the `create_methodology_amendment.robot` test. This was due to a click being intercepted because the modal for deleting a content section wasn't fully closed.

### DIFF
--- a/tests/robot-tests/tests/libs/admin/manage-content-common.robot
+++ b/tests/robot-tests/tests/libs/admin/manage-content-common.robot
@@ -216,6 +216,7 @@ user deletes editable accordion section
     user clicks button    Remove this section    ${section}
     user waits until modal is visible    Are you sure?
     user clicks button    Confirm
+    user waits until page does not contain button    Confirm
 
 get accordion section text block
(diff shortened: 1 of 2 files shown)


tests/robot-tests/tests/libs/admin/manage-content-common.robot
```diff
@@ -216,6 +216,7 @@ user deletes editable accordion section
     user clicks button    Remove this section    ${section}
     user waits until modal is visible    Are you sure?
     user clicks button    Confirm
+    user waits until page does not contain button    Confirm
```
duncan-at-hiveit, line 219: I'd be tempted to rename this `user waits until modal is not visible  Are you sure?` or alternatively just `user waits util modal is not visible`, as we only ever normally deal with one at a time.

Then it can be used across all tests that use the keyword `user waits until modal is visible` to do the corresponding closed test.

PR #2934, opened 2021-10-15
EES-2811 - improve methodology test assertions

luke-h1:


tests/robot-tests/tests/admin/bau/comments.robot
```diff
@@ -200,3 +200,6 @@ Switch to bau1 to approve release for immediate publication
     user waits until parent contains element    ${details}    xpath:.//a[text()="Edit this release"]
     user clicks link    Edit this release
     user approves original release for immediate publication
+
+Visit methodology on public side and check it is visible on 'all methodologies' page
+    user verifies methodology is visible on public methodology page    %{TEST_THEME_NAME}    %{TEST_TOPIC_NAME}    ${PUBLICATION_NAME}    /methodology/ui-tests-comments-%{RUN_IDENTIFIER}
```
benoutram, line 205: I'm not sure why this comments test really needs to have anything to do with setting up a methodology at all? Would be interested to know.

tests/robot-tests/tests/admin/bau/create_methodology_amendment.robot
```diff
@@ -93,6 +94,11 @@ Verify the editable content is as expected
 Approve the Methodology
     user approves methodology for publication    ${PUBLICATION_NAME}    ${PUBLICATION_NAME} - first methodology version
 
+Visit methodology on public side and check it is visible on 'all methodologies' page
+    user verifies methodology is visible on public methodology page    %{TEST_THEME_NAME}    %{TEST_TOPIC_NAME}
```
benoutram, line 98: I think we can refactor similar checks in `bau\publish_methodology.robot` to use this now.

tests/robot-tests/tests/libs/public-common.robot
```diff
@@ -43,3 +43,12 @@ user checks methodology note
...
     user waits until element contains    css:#methodologyNotes li:nth-of-type(${number}) time    ${displayDate}
     user waits until element contains    css:#methodologyNotes li:nth-of-type(${number}) p    ${content}
+
+user verifies methodology is visible on public methodology page
+    [Arguments]    ${THEME}    ${TOPIC}    ${METHODOLOGY_NAME}    ${METHODOLOGY_SLUG}
+    user navigates to public methodologies page
+    user waits until page contains accordion section    ${THEME}
+    ${accordion}=    user opens accordion section    ${THEME}
+    ${details}=    user opens details dropdown    ${TOPIC}    ${accordion}
+    user checks page contains element    xpath://*[a/text()="${METHODOLOGY_NAME}"]
```
benoutram, line 53: I think they are listed within their publication and this doesn't seem to check the publication name so they could be appearing in any publication. I think where we do this kind of check already in `bau\publish_methodology.robot` that we check the publication name as well as the methodology name.

tests/robot-tests/tests/admin/bau/comments.robot
luke-h1: Realistically we could get rid of the methodology assertions in this test. I put the methodology assertions in this test to mirror the type of content analysts create on production. Happy to remove or discuss more if you feel it's not necessary (would reduce test time if we removed it)

tests/robot-tests/tests/admin/bau/comments.robot
```diff
@@ -182,21 +135,3 @@ Resolve comments
...
     ${comment_2}=    user gets comment    Fixed the problem
     user clicks button    Resolve    ${comment_2}
-
-Move publication into 'Higher Review'
-    [Tags]    HappyPath
-    user clicks link    Sign off
-    user clicks button    Edit release status
-    user clicks element    id:releaseStatusForm-approvalStatus-HigherLevelReview
-
-Switch to bau1 to approve release for immediate publication
```
benoutram, line 192: Just wondering if the purpose of this was to make sure that it's possible to approve following the comment that's now been resolved?

benoutram:
> I've removed most of the changes made in this PR. We're already testing for the EES-2804 bug in publish_methodology.robot, so there is no need to add additional checks for this bug.

I agree.

Looking at `publish_methodology.robot`, line 79 `Verify that the methodology is visible on the public methodologies page with the expected URL` wouldn't reproduce it because the cached methodologies file would be deleted by publishing the release just prior to it. This is done by the Publisher function which was deleting the blob files successfully.

However line 196 `Verify that the amended methodology is visible on the public methodologies page` should reproduce it as the methodology being approved immediately will have the Admin bug deleting the blob file and there's no other releases being published where the file could get deleted. So I'd have expected this UI test to be failing in Dev on this one which I think it was.
```

PR #2936, opened 2021-10-18
EES-2762 explore data accordion updates

amyb-hiveit:
Various small changes to the release page, mainly around the 'explore data and files' accordion, on the public frontend and, where applicable, the admin

### DIFF
--- a/tests/robot-tests/tests/general_public/release_page_absence.robot
+++ b/tests/robot-tests/tests/general_public/release_page_absence.robot
@@ -63,7 +63,7 @@ Validate "About these statistics" -- "Last updated"
     user closes details dropdown    See all updates (2)
 
 Validate "Useful information"
-     user checks page contains element    link:Download all data
+    user checks page contains element    link:Download all data
     user checks page contains link with text and url    Pupil absence statistics: methodology
     ...    /methodology/pupil-absence-in-schools-in-england
 


tests/robot-tests/tests/general_public/release_page_absence.robot
```diff
@@ -63,6 +63,7 @@ Validate "About these statistics" -- "Last updated"
     user closes details dropdown    See all updates (2)
 
 Validate "Useful information"
+     user checks page contains element    link:Download all data
```
benoutram, line 66: I think it might be worth running the robot formatter on this file as it looks like the formatting is out here.

src/explore-education-statistics-frontend/src/modules/find-statistics/PublicationReleasePage.tsx
```diff
@@ -47,6 +47,12 @@ const PublicationReleasePage: NextPage<Props> = ({ release }) => {
   // have the updates in the correct order.
   const updates = orderBy(release.updates, 'on', 'desc');
 
+  const showAllFilesButton = release.downloadFiles.some(
+    file =>
+      file.type === 'Data' ||
+      (file.type === 'Ancillary' && file.name !== 'All files'),
```
benoutram, line 53: I'm just wondering why we do this and if we need to take into account that the 'All files' file type is now 'zip' rather than 'ancillary'. After the recent changes I'm not sure that it's possible to get an 'Ancillary' file with name 'All files'.

PR #2937, opened 2021-10-19
EES-2764 Add ability to create snapshots

ghost:
This PR creates a python script that allows the creation of JSON snapshots of the Find Statistics page, the Data Tables page, and the Data Catalogue page.

### DIFF
--- a/src/explore-education-statistics-common/src/modules/table-tool/components/PublicationForm.tsx
+++ b/src/explore-education-statistics-common/src/modules/table-tool/components/PublicationForm.tsx
@@ -152,7 +152,6 @@ const PublicationForm = (props: Props & InjectedWizardProps) => {
                           key={theme.id}
                           id={`${formId}-theme-${theme.id}`}
                           detailsId="theme"
-                          headingId="theme-heading"
                           open={
                             searchTerm !== '' ||
                             theme.topics.some(topic =>
@@ -169,7 +168,6 @@ const PublicationForm = (props: Props & InjectedWizardProps) => {
                               key={topic.id}
                               id={`${formId}-topic-${topic.id}`}
                               detailsId="topic"
-                              headingId="topic-heading"
                               open={
                                 searchTerm !== '' ||
                                 topic.publications.some(
--- a/src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
@@ -81,7 +81,7 @@ const FindStatisticsPage: NextPage<Props> = ({ themes = [] }) => {
                       key={topicId}
                       summary={topicTitle}
                       id={`topic-${topicId}`}
-                      headingId="topic-heading"
+                      detailsId="topic"
                     >
                       <PublicationList publications={publications} />
                     </Details>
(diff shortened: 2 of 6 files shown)


src/explore-education-statistics-common/src/modules/table-tool/components/PublicationForm.tsx
```diff
@@ -151,6 +151,8 @@ const PublicationForm = (props: Props & InjectedWizardProps) => {
                           summary={theme.title}
                           key={theme.id}
                           id={`${formId}-theme-${theme.id}`}
+                          detailsId="theme"
```
duncan-at-hiveit, line 154: Was there a reason why we'd want to explicitly pass these in, rather than just concatenating "-details", "-heading" etc to the `id` parameter that's already being passed in?

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
```diff
@@ -81,6 +81,7 @@ const FindStatisticsPage: NextPage<Props> = ({ themes = [] }) => {
                       key={topicId}
                       summary={topicTitle}
                       id={`topic-${topicId}`}
+                      headingId="topic-heading"
```
duncan-at-hiveit, line 84: Could we run the risk here of ending up with duplicate ids in the DOM?

src/explore-education-statistics-frontend/src/modules/find-statistics/components/PublicationList.tsx
```diff
@@ -11,7 +11,12 @@ function PublicationList({ publications }: Props) {
     <ul className="govuk-!-margin-top-0">
       {publications.map(({ id, legacyPublicationUrl, slug, title }) => (
         <li key={id} className="govuk-!-margin-bottom-0">
-          <h3 className="govuk-heading-s govuk-!-margin-bottom-0">{title}</h3>
+          <h3
+            id="publication-heading"
```
duncan-at-hiveit, line 15: Duplicate ids will appear in the DOM if this is in a loop - you'd need to include the `id` in there to make it unique.

tests/robot-tests/scripts/create_snapshots.py
```diff
@@ -0,0 +1,145 @@
+import os
```
duncan-at-hiveit, line 1: Would you be against putting a quick description of what this script does at the top here?  Or would you prefer to document it in another way?

tests/robot-tests/scripts/create_snapshots.py
```diff
@@ -0,0 +1,145 @@
...
+import json
+import argparse
+import requests
+from bs4 import BeautifulSoup
+
+
+def gets_parsed_html_from_page(url):
```
duncan-at-hiveit, line 8: I think that in Python if you've got methods that don't get called from outside the file, you can mark them up as "private" with a cheeky underscore prefix?  Might be worth doing here, so we know what the main methods are to call from outside this script?

src/explore-education-statistics-frontend/src/modules/find-statistics/components/PublicationList.tsx
markhiveit: I forgot that the convention for id attributes is that they should be unique. I've switched to using `data-testid` in this case.

duncan-at-hiveit: I think you'd find the same issue with data-testid - worth a Google, but I *think* they're supposed to be unique too.

PR #2953, opened 2021-10-27
EES-2834 Only BAU users should be able to update publication titles

markhiveit:
This PR changes permissions around updating publication titles such that now only BAU users will be able to do so.

To achieve this, `PublicationEditPage.tsx` now takes `MyPublicationViewModel` from the backend rather than `PublicationViewModel`, so that permission information can be passed to the page. The frontend uses this information to hide the publication title input if necessary.

image


ntsim (dismissed):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -119,6 +119,11 @@ public class PublicationService : IPublicationService
                         return await ValidateSelectedTopic(updatedPublication.TopicId);
                     }
 
+                    if (publication.Title != updatedPublication.Title)
```
benoutram, line 122: We are saying if the topic id is changing skip the check on the title changing (an so also skip the security check on updating the publication title).

I guess this is ok since no user  is likely to be updating the topic id and changing the title at the same time.
Also there's no user who can move a publication between topics who can't update a title as that's for BAU only.

I would have made this independent of the Topic check though, so always check the title changing and do that security check if necessary, even if the topic id has changed and the user can move the topic.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -119,6 +119,11 @@ public class PublicationService : IPublicationService
...
+                    {
+                        return await _userService.CheckCanUpdatePublicationTitle();
```
benoutram, line 124: Looks like `PublicationService` `UpdatePublication` is completely untested, so this would have been a good opportunity to introduce some tests for it although appreciate it's not an easy method to write tests for and this change would have no affect on the test had it existed since we usually use a `userService` mock that always returns true.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
markhiveit: I've moved the `Title !=` check into a separate `OnSuccessDo`. I missed that that check wouldn't be executed if the topic had changed.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/MyPublicationViewModel.cs
```diff
@@ -24,11 +24,14 @@ public class MyPublicationViewModel
 
         public Contact Contact { get; set; }
 
+        public List<LegacyReleaseViewModel> LegacyReleases { get; set; }
```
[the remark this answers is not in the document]
benoutram: Update: just realised the view models are built in PublicationRepository which does already have unit tests so it should be easier to update those tests.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
markhiveit: There were some unit tests for UpdatePublication, but I added a couple that use a `userService` mock.

tests/robot-tests/tests/admin/analyst/role_ui_permissions/publication_owner_permissions.robot
```diff
@@ -29,7 +29,25 @@ Check cannot edit content for published release
 Navigate back to admin dashboard for publication
     user navigates to publication on admin dashboard    ${PUBLICATION_NAME}    ${THEME_NAME}    ${TOPIC_NAME}
 
+Check Edit publication page inputs are correct
+    ${accordion}=    user gets accordion section content element    ${PUBLICATION_NAME}
+    user clicks link    Manage this publication    ${accordion}
+    user waits until page contains title    Manage publication
+
+    user waits until page contains element    id:publicationForm-themeId
```
duncan-at-hiveit, line 37: I think we might have a way of checking for the existence of formfields by their labels in the Robot tests, which might be more readable.  Not 100% sure though, and def not high priority!

tests/robot-tests/tests/admin_and_public/bau/publish_publication_update.robot
```diff
@@ -0,0 +1,71 @@
...
+    ...    Academic Year 2046/47 (not Live)
+    user approves release for immediate publication
+
+Navigate to publication on admin dashboard
+    ${accordion}=    user opens publication on the admin dashboard    ${PUBLICATION_NAME}
+    user clicks link    Manage this publication    ${accordion}
+    user waits until page contains title    Manage publication
+
+Update publication details
+    user enters text into element    id:publicationForm-title    ${PUBLICATION_NAME_UPDATED}
```
duncan-at-hiveit, line 32: Can use a keyword that accepts the formfield label here for better readability, I thiiiiink...

duncan-at-hiveit (approved):

duncan-at-hiveit (approved):

PR #2960, opened 2021-11-01
EES-152 Return validation error if table builder query can exceed maximum allowable table size

benoutram:
This PR returns a validation error if a table builder query exceeds the maximum number of allowed table cells.

This applies to both the admin site for analysts building data blocks and the public site for users using the table builder.

It calculates the maximum number of table cells that could be rendered for the result of a table query, based on
the indicator, location, time period, and any other filter parameters of the query.

Calculating based on attributes of the query criteria is preferential to counting the actual result
to avoid wasted overhead of executing a query that won't be rendered due to exceeding the cell count limit.

This works by making an assumption that data is provided for _all_ combinations of location, time
period and any other filter values to calculate the maximum possible count of table cells, agreed with @cjrace.

The number of ways of observing each indicator is the product of the counts of the items in each of
the sets of locations, time periods, and filters. This is multiplied by the count of indicators 
to get the maximum possible count of table cells.

Note that filter items have to be counted by filter since each filter adds an independent set of possibilities, so it's necessary to lookup the filters for the filter items in the query to count them by.

The limit is configurable in ARM template configuration and can be varied between environments as well as set differently for the local environment. Altering the limit requires an Infrastructure deploy and a restart of the Admin and Data API  app services.

image

### Determining size limit

Although the default limit has been specified at 25000 table cells it's intended that shortly after going live with this change the limit will be adjusted to a more suitable value. We must make sure that we don't block requests for any existing data block tables / charts / fast tracks.

A temporary controller has been written that returns a report for determining the maximum possible size of a table that could be generated by existing datablocks in the service.

`GET /api/data-blocks/query-size-report`

For each published data block this returns a result

```
    {
        "dataBlockId": "7eeb280f-6a37-4566-4645-08d93c022a16",
        "dataBlockName": "Test Data Block",
        "filters": {
            "da1843f5-5d52-4f91-9e81-f79f8fcca811": 7
        },
        "indicators": 3,
        "locations": 1,
        "timePeriods": 11,
        "maxTableCells": 231
    }
```

The greatest figure will be used to establish the lowest possible limit of the maximum table size that we can set so that we don't reject any requests for tables in existing data blocks or fast track entries.

Note that we *won't* execute all datablock queries here and count the results to get the data for this report.

A table result might be smaller than the possible table that could be generated since all combinations of time
period, location and filters may not be provided in the data.

If we were to execute all the datablock queries and set a maximum based on the largest table result,
several datablock requests may fail if they have a possible table size greater than the largest actual table.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Model.Tests/FilterItemRepositoryTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Model.Tests/FilterItemRepositoryTests.cs
@@ -1,0 +1,178 @@
+﻿#nullable enable
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using GovUk.Education.ExploreEducationStatistics.Data.Model.Database;
+using GovUk.Education.ExploreEducationStatistics.Data.Model.Repository;
+using Xunit;
+using static GovUk.Education.ExploreEducationStatistics.Common.Services.CollectionUtils;
+
+namespace GovUk.Education.ExploreEducationStatistics.Data.Model.Tests
+{
+    public class FilterItemRepositoryTests
+    {
+        [Fact]
+        public async Task CountFilterItemsByFilter()
+        {
+            var filterItemCharacteristicSchoolYear1 = new FilterItem();
+            var filterItemCharacteristicFsmEligible = new FilterItem();
+            var filterItemCharacteristicFsmNotEligible = new FilterItem();
+            var filterItemSchoolTypePrimary = new FilterItem();
+            var filterItemSchoolTypeSecondary = new FilterItem();
+
+            var filterCharacteristic = new Filter
+            {
+                FilterGroups = new List<FilterGroup>
+                {
+                    new()
+                    {
+                        FilterItems = new List<FilterItem>
+                        {
+                            filterItemCharacteristicSchoolYear1
+                        }
+                    },
+                    new()
+                    {
+                        FilterItems = new List<FilterItem>
+                        {
+                            filterItemCharacteristicFsmEligible,
+                            filterItemCharacteristicFsmNotEligible
+                        }
+                    }
+                }
+            };
+
+            var filterSchoolType = new Filter
+            {
+                FilterGroups = new List<FilterGroup>
+                {
+                    new()
+                    {
+                        FilterItems = new List<FilterItem>
+                        {
+                            filterItemSchoolTypePrimary,
+                            filterItemSchoolTypeSecondary
+                        }
+                    }
+                }
+            };
+
+            var statisticsDbContextId = Guid.NewGuid().ToString();
+
+            await using (var statisticsDbContext = StatisticsDbUtils.InMemoryStatisticsDbContext(statisticsDbContextId))
+            {
+                await statisticsDbContext.Filter.AddRangeAsync(filterCharacteristic, filterSchoolType);
+                await statisticsDbContext.SaveChangesAsync();
+            }
+
+            await using (var context = StatisticsDbUtils.InMemoryStatisticsDbContext(statisticsDbContextId))
+            {
+                var repository = BuildFilterItemRepository(context);
+                var filterItemIds = new List<Guid>
+                {
+                    filterItemCharacteristicSchoolYear1.Id,
+                    filterItemCharacteristicFsmEligible.Id,
+                    filterItemCharacteristicFsmNotEligible.Id,
+                    filterItemSchoolTypePrimary.Id,
+                    filterItemSchoolTypeSecondary.Id
+                };
+                var result = await repository.CountFilterItemsByFilter(filterItemIds);
+
+                // Result should contain the counts of filter items in both filters
+                Assert.Equal(2, result.Count);
+
+                // 3 of the filter items belong to the Characteristic filter 
+                Assert.Equal(3, result[filterCharacteristic.Id]);
+
+                // 2 of the filter items belong to the School Type filter
+                Assert.Equal(2, result[filterSchoolType.Id]);
+            }
+        }
+
+        [Fact]
+        public async Task CountFilterItemsByFilter_EmptyFilterItemsIsEmpty()
+        {
+            var filter = new Filter
+            {
+                FilterGroups = new List<FilterGroup>
+                {
+                    new()
+                    {
+                        FilterItems = new List<FilterItem>
+                        {
+                            new()
+                        }
+                    }
+                }
+            };
+
+            var statisticsDbContextId = Guid.NewGuid().ToString();
+
+            await using (var statisticsDbContext = StatisticsDbUtils.InMemoryStatisticsDbContext(statisticsDbContextId))
+            {
+                await statisticsDbContext.Filter.AddRangeAsync(filter);
+                await statisticsDbContext.SaveChangesAsync();
+            }
+
+            await using (var context = StatisticsDbUtils.InMemoryStatisticsDbContext(statisticsDbContextId))
+            {
+                var repository = BuildFilterItemRepository(context);
+                var result = await repository.CountFilterItemsByFilter(new List<Guid>());
+                Assert.Empty(result);
+            }
+        }
+
+        [Fact]
+        public async Task CountFilterItemsByFilter_FilterItemsNotFoundThrowsException()
+        {
+            var filterItem = new FilterItem();
+            var filter = new Filter
+            {
+                FilterGroups = new List<FilterGroup>
+                {
+                    new()
+                    {
+                        FilterItems = new List<FilterItem>
+                        {
+                            filterItem
+                        }
+                    }
+                }
+            };
+
+            var statisticsDbContextId = Guid.NewGuid().ToString();
+
+            await using (var statisticsDbContext = StatisticsDbUtils.InMemoryStatisticsDbContext(statisticsDbContextId))
+            {
+                await statisticsDbContext.Filter.AddRangeAsync(filter);
+                await statisticsDbContext.SaveChangesAsync();
+            }
+
+            await using (var context = StatisticsDbUtils.InMemoryStatisticsDbContext(statisticsDbContextId))
+            {
+                var repository = BuildFilterItemRepository(context);
+
+                var filterItemNotFound1 = Guid.NewGuid();
+                var filterItemNotFound2 = Guid.NewGuid();
+
+                var exception = await Assert.ThrowsAsync<ArgumentException>(async () =>
+                {
+                    await repository.CountFilterItemsByFilter(
+                        ListOf(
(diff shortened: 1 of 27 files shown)


src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/TableBuilderServiceTests.cs
```diff
@@ -259,6 +268,115 @@ await using (var statisticsDbContext = StatisticsDbUtils.InMemoryStatisticsDbCon
...
+                    subjectRepository: subjectRepository.Object,
+                    releaseRepository: releaseRepository.Object,
+                    options: options
+                );
+
+                var result = await service.Query(query);
+
+                result.AssertBadRequest(ValidationErrorMessages.QueryExceedsMaxAllowableTableSize);
+
+                MockUtils.VerifyAllMocks(filterItemRepository, subjectRepository, releaseRepository);
```
duncan-at-hiveit, line 376: I'd put this immediately after `var result = await service.Query(query);`

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/Utils/TableBuilderUtilsTests.cs
```diff
@@ -0,0 +1,162 @@
...
+
+        [Fact]
+        public void MaximumTableCellCount_NoFilters_MultipleLocations()
+        {
+            /*
+             * 2011  Bury  cell1
+             * 2011  Kent  cell2
+             * 2011  York  cell3
+             */
+            Assert.Equal(3, MaximumTableCellCount(1, 3, 1));
```
duncan-at-hiveit, line 36: Might be useful for the readability of these various tests to use named parameters, like:

```
        [Fact]
        public void MaximumTableCellCount_NoFilters_MultipleLocations()
        {
            /*
             * 2011  Bury  cell1
             * 2011  Kent  cell2
             * 2011  York  cell3
             */
            Assert.Equal(
                 3, 
                 MaximumTableCellCount(
                     indicatorCount: 1, 
                     locationCount: 3, 
                     timePeriodCount: 1
                 )
            );
        }
```

duncan-at-hiveit: Dodgy formatting optional :)

src/GovUk.Education.ExploreEducationStatistics.Data.Services/TableBuilderService.cs
```diff
@@ -100,6 +119,30 @@ public class TableBuilderService : ITableBuilderService
                 });
         }
 
+        private async Task<int> GetMaximumTableCellCount(ObservationQueryContext queryContext)
+        {
+            var filterItemIds = queryContext.Filters;
+            var countsOfFilterItemsByFilter = filterItemIds == null
+                ? new List<int>()
+                : (await _filterItemRepository.CountFilterItemsByFilter(filterItemIds))
+                .Select(pair =>
```
duncan-at-hiveit, line 128: Could just go for `.Select(pair => pair.Item2)` here, but were you just hoping to be more explicit?

src/GovUk.Education.ExploreEducationStatistics.Data.Services/TableBuilderService.cs
```diff
@@ -111,4 +154,11 @@ await _userService.MatchesPolicy(subject, CanViewSubjectData))
             return new ForbidResult();
         }
     }
+
+    public class TableBuilderOptions
+    {
+        public const string TableBuilder = "TableBuilder";
```
duncan-at-hiveit, line 160: What's the convention here?  Why do we need this?

src/GovUk.Education.ExploreEducationStatistics.Data.Services/Utils/TableBuilderUtils.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <param name="countsOfFilterItemsByFilter"></param>
+        /// <returns></returns>
+        public static int MaximumTableCellCount(
+            int countOfIndicators,
+            int countOfLocations,
+            int countOfTimePeriods,
+            IEnumerable<int>? countsOfFilterItemsByFilter = null)
+        {
+            var result = countOfIndicators * countOfLocations * countOfTimePeriods;
+            if (countsOfFilterItemsByFilter != null)
```
duncan-at-hiveit, line 51: You could probably do a Reduce here to sum up the total filter items and just multiply the `result` by the sum, rather than do the loop?

Something like:

```
  var multiplier = countsOfFilterItemsByFilter == null ? 1 : countsOfFilterItemsByFilter.Reduce(0, (acc, filterItems) => acc + filterItems);
  return multiplier * result;
```

tests/robot-tests/tests/general_public/table_tool_exclusions_by_geographic_level.robot
```diff
@@ -33,8 +33,71 @@ Select subject "Exclusions by geographic level"
...
+    user clicks category checkbox    School type    State-funded primary
+    user clicks category checkbox    School type    State-funded secondary
+    user clicks category checkbox    School type    Special
+
+User clicks Create table button
+    user clicks element    id:filtersForm-submit
+
+Validate the query could exceed the maximum allowable table size
+    user waits until element contains    id:filtersForm-indicators-error
+    ...    A table cannot be returned as the filters chosen can exceed the maximum allowable table size
```
duncan-at-hiveit, line 87: Great stuff :)

tests/robot-tests/tests/general_public/table_tool_exclusions_by_geographic_level.robot
```diff
@@ -33,8 +33,71 @@ Select subject "Exclusions by geographic level"
...
+
+Go back to Locations step
+    user clicks element    xpath://button[contains(text(), "Edit locations")]
```
luke-h1, line 90: We could use `user clicks button  Edit locations` to be a bit more succinct here

tests/robot-tests/tests/general_public/table_tool_exclusions_by_geographic_level.robot
```diff
@@ -33,8 +33,71 @@ Select subject "Exclusions by geographic level"
...
+    user waits until page contains element    xpath://h1[text()="Go back to previous step"]
+    user clicks element    xpath://button[text()="Confirm"]
```
luke-h1, line 92: Similar to the last comment we should be able to use the `user clicks button` kw

luke-h1 (changes requested):

tests/robot-tests/tests/general_public/table_tool_exclusions_by_geographic_level.robot
benoutram: Thanks I've done this now

tests/robot-tests/tests/general_public/table_tool_exclusions_by_geographic_level.robot
benoutram: Good suggestion, I've made this change now, thanks

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/Utils/TableBuilderUtilsTests.cs
benoutram: Good suggestion, I've made this change now, thanks.

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/TableBuilderServiceTests.cs
benoutram: Thanks I've done this now and changed all the other tests in this class to use the same pattern.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/TableBuilderService.cs
benoutram: It's just the section name expected in the configuration JOSN data.

I suppose it's  not really essential.

Instead of 

```
services.Configure<TableBuilderOptions>(Configuration.GetSection("TableBuilder"));
```

we do

```
services.Configure<TableBuilderOptions>(Configuration.GetSection(TableBuilderOptions.TableBuilder));
```

Since we do this in two places (Data.API and Admin) the constant kind of makes sense to tie them together.

See the following section of the documentation where I got this convention from:

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-5.0#bind-hierarchical-configuration

where it says

> An alternative approach when using the options pattern is to bind the Position section and add it to the dependency injection service container. In the following code, PositionOptions is added to the service container with Configure and bound to configuration

Their example is

```
services.Configure<PositionOptions>(Configuration.GetSection(PositionOptions.Position));
```

src/GovUk.Education.ExploreEducationStatistics.Data.Model.Tests/FilterItemRepositoryTests.cs
```diff
@@ -0,0 +1,178 @@
...
+            var filterItemSchoolTypePrimary = new FilterItem();
+            var filterItemSchoolTypeSecondary = new FilterItem();
+
+            var filterCharacteristic = new Filter
+            {
+                FilterGroups = new List<FilterGroup>
+                {
+                    new()
+                    {
+                        FilterItems = new List<FilterItem>
```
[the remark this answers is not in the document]
benoutram: Seem to need this here

src/GovUk.Education.ExploreEducationStatistics.Data.Services/TableBuilderService.cs
benoutram: Ah yep I was trying to be more explicit about 'count' rather than returning 'Item2'.

Ideally I want to deconstruct them in the parameter so  something like `.Select((_, count) => count)` but that's not possible. Note this would  actually compile but the `int` count parameter would be assigned the index of the selector! 🙂

src/GovUk.Education.ExploreEducationStatistics.Data.Services/Utils/TableBuilderUtils.cs
benoutram: Rider does suggest this as an option (actually it suggests `Aggregate` rather than `Reduce`) and I considered it but thought it's less readable. Also `result` isn't really the result until it's returned so I was struggling with the naming of it.

duncan-at-hiveit (approved):

luke-h1 (approved):

PR #2963, opened 2021-11-01
EES-2858 - Refactor requests to use connection pooling

luke-h1:
### This PR 

Refactors our current implementation of the requests library to use connection pooling. This offers several performance improvements to using the standard `requests.<verb>` method  (lower CPU + memory usage, reduced latency in multiple requests, etc.)

We do this via creating a new session in requests: 

```python 
requests.sessions.HTTPAdapter(
    pool_connections=50,
    pool_maxsize=50,
    max_retries=3
)
session = requests.Session()

# Example: 
def my_request:
    myData = session.get("https://www.google.com", flush=True)
    # the response we get here is identical to the response we would get from `requests.get`


```

Sessions allow us to reuse an already opened connection. Each connection is stored in a pool of connections. Reusing the connection to send out multiple requests offers multiple performance improvements

In addition to connection pooling, I've added `flush=True` to all requests. The flush parameter provides a way to **not** load the full content of a given response into memory as soon as the request is executed. This is important as we don't have to read the entire body of the HTTP response (which will increase memory usage) and can instead read it in chunks.

image

[snapshots.zip](https://github.com/dfe-analytical-services/explore-education-statistics/files/7454224/snapshots.zip)

### DIFF
--- a/tests/robot-tests/run_tests.py
+++ b/tests/robot-tests/run_tests.py
@@ -215,6 +215,13 @@ def admin_request(method, endpoint, body=None):
     if method == 'POST':
         assert body is not None, 'POST requests require a body'
 
+    requests.sessions.HTTPAdapter(
+        pool_connections=50,
+        pool_maxsize=50,
+        max_retries=3
+    )
+    session = requests.Session()
+
     # To prevent InsecureRequestWarning
     requests.packages.urllib3.disable_warnings()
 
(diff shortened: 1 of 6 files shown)


duncan-at-hiveit (approved):
Looks good to me!  :)  I guess in the PR description, where you mention "Flush" this is actually "Stream" in the final implementation?

luke-h1:
> Looks good to me! :) I guess in the PR description, where you mention "Flush" this is actually "Stream" in the final implementation?

Yep that's what I meant

PR #2969, opened 2021-11-04
EES-2863 - removed versioned entity comparator in favour of util method

duncan-at-hiveit:
This PR:
- removes the Comparator previously being used to order entities for deletion in favour of a util method that returns a new sorted list.  The original Comparator was breaking in certain scenarios, and not always returning the correct order.
- fixes a few UI test issues


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Models/VersionedEntityDeletionOrderUtilTests.cs
```diff
@@ -0,0 +1,75 @@
...
+            var entity4Version3Id = "entity-4-version-3";
+            var entity4Version6Id = "entity-4-version-6";
+
+            var versions = new List<IdAndPreviousVersionIdPair<string>>
+            {
+                new(entity1Version6Id, entity1Version5Id),
+                new(entity3Version2Id, entity3Version1Id),
+                null,
+                new(entity1Version2Id, entity1Version1Id),
+                new(entity4Version3Id, "entity-4-version-2"),
```
markhiveit, line 36: How come you're using string literals here?

markhiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Models/VersionedEntityDeletionOrderUtilTests.cs
duncan-at-hiveit: I wanted to convey that these entries didn't have any associated previous version in the list of returned entities.   More as an edge case rather than anything I'd expect to happen.

duncan-at-hiveit: Would you say that I should add a comment to explain it in the test?

markhiveit: I'm ok with leaving it - does make sense now you say it.

PR #2973, opened 2021-11-08
EES-2862 - delete theme & topic utility

luke-h1:
This PR:
- adds a utility to delete themes and topics from environments
- Adds esbuild as a dependency (esr has added this as a peer dependency)


luke-h1:
don't merge just yet, need to add some finishing touches

duncan-at-hiveit (approved):

duncan-at-hiveit (approved):
Very nice, very nice!  We could do with this being documented somewhere, or a little demo :)

luke-h1:
> Very nice, very nice! We could do with this being documented somewhere, or a little demo :)

Thanks, yeah that's a good idea. Not sure what format that would take though

PR #2977, opened 2021-11-10
    EES-2878 - append statistics to release template tag

luke-h1:


src/explore-education-statistics-common/src/services/publicationService.ts
```diff
@@ -79,11 +79,11 @@ export interface ReleaseNote {
 
 // eslint-disable-next-line no-shadow
 export enum ReleaseType {
-  AdHoc = 'Ad Hoc',
-  Experimental = 'Experimental',
+  AdHoc = 'Ad Hoc Statistics',
```
luke-h1, line 82: I would rename `AdHoc` & `Experimental` to conform with the naming convention here. i.e. `AdHocStatistics` & `ExperimentalStatistics`

src/explore-education-statistics-admin/src/pages/release/content/components/BasicReleaseSummary.tsx
```diff
@@ -60,11 +60,7 @@ const BasicReleaseSummary = ({ release }: Props) => {
               <Tag className="govuk-!-margin-right-3 govuk-!-margin-bottom-3">
                 {getReleaseApprovalStatusLabel(release.approvalStatus)}
               </Tag>
-              {release.type.title !== 'Transparency Data' ? (
-                <Tag>{release.type.title} statistics</Tag>
-              ) : (
-                <Tag>{release.type.title}</Tag>
-              )}
+              <Tag>{release.type.title}</Tag>
```
luke-h1: not sure if we want Management Information to have statistics appended to it but if we don't we will need to add a ternary op to check for this
```
{release.type.title !== 'Management Information' ? <Tag>{release.type.title} statistics</Tag> : <Tag>{release.type.title}</Tag>}
```

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20211111103109_EES2878UpdateReleaseTypeTitles.cs
```diff
@@ -0,0 +1,21 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Migrations.ContentMigrations
+{
+    public partial class EES2878UpdateReleaseTypeTitles : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql("UPDATE ReleaseTypes SET Title = 'Ad Hoc Statistics' WHERE Id = '1821abb8-68b0-431b-9770-0bea65d02ff0';");
```
benoutram, line 9: We've now got a mixture of the ReleaseTypes reference data being managed within ContentDbContext and outside of its control as I think two new release types have recently been introduced without updating the definition that already exists there. See `modelBuilder.Entity<ReleaseType>().HasData` where they are defined.

We should have either introduced the new types to that, and be changing these titles by updating that definition, or we should have got rid of it and managed the unwanted migration that it would have generated removing them, and then this PR would have been acceptable.

I think there's an issue with removing the definition altogether that if the migrations are baselined, the table will be empty if starting up with a fresh db. As the application code expects these types to exist I would be tempted to keep the definition in code.

benoutram (changes requested):

src/explore-education-statistics-common/src/services/publicationService.ts
```diff
@@ -79,11 +79,11 @@ export interface ReleaseNote {
 
 // eslint-disable-next-line no-shadow
 export enum ReleaseType {
-  AdHoc = 'Ad Hoc',
-  Experimental = 'Experimental',
+  AdHocStatistics = 'Ad Hoc Statistics',
```
benoutram, line 82: Searched the UI tests and it doesn't look like these changes would affect anything but just checking?

duncan-at-hiveit: Tests passed ok (other than known issues)

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20211111103109_EES2878UpdateReleaseTypeTitles.cs
markhiveit: I added the two new release types to ContentDbContext, and when I created a migration it tries to insert the new data. But this wouldn't work because those rows were already added in the previous migration. So I've manually added lines to delete the new release types first with a comment referencing the migration where they were created.

src/explore-education-statistics-common/src/services/publicationService.ts
markhiveit: I've made a few updates to the backend and frontend unit tests - not to fix tests but just to change test data from `Ad Hoc` to `Ad Hoc Statistics`

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20211111132702_EES2878UpdateReleaseTypeTitles.cs
```diff
@@ -0,0 +1,60 @@
...
+                value: "Ad Hoc Statistics");
+
+            // To remove previous entry created by previous migration - EES-1500
+            migrationBuilder.DeleteData(
+                table: "ReleaseTypes",
+                keyColumn: "Id",
+                keyValue: new Guid("15bd4f57-c837-4821-b308-7f4169cd9330"));
+
+            // To remove previous entry created by previous migration - EES-1500
+            migrationBuilder.DeleteData(
```
benoutram, line 24: We won't be able to delete these if they are already in use. 

Would be safer to turn them into updates, assuming that the previous PR already inserted them.

```
            migrationBuilder.UpdateData(
                table: "ReleaseTypes",
                keyColumn: "Id",
                keyValue: new Guid("15bd4f57-c837-4821-b308-7f4169cd9330"),
                column: "Title",
                value: "Management Information");
```

If the migrations are baselined they should all appear in the baselined migration as Inserts.

benoutram (approved):

PR #2978, opened 2021-11-10
EES-2879 - use node 14 across admin & public frontend

luke-h1:


infrastructure/templates/template.json
```diff
@@ -1200,7 +1200,7 @@
       ],
       "properties": {
         "APPINSIGHTS_INSTRUMENTATIONKEY": "[reference(resourceId('microsoft.insights/components/', variables('dataAppInsights')), '2020-02-02').InstrumentationKey]",
-        "WEBSITE_NODE_DEFAULT_VERSION": "6.9.1",
+        "WEBSITE_NODE_DEFAULT_VERSION": "14.0.0",
```
markhiveit, line 1203: We should set this to be 14.18.1 - same with other instances.

markhiveit (changes requested):

infrastructure/templates/template.json
```diff
@@ -1200,7 +1200,7 @@
       ],
       "properties": {
         "APPINSIGHTS_INSTRUMENTATIONKEY": "[reference(resourceId('microsoft.insights/components/', variables('dataAppInsights')), '2020-02-02').InstrumentationKey]",
-        "WEBSITE_NODE_DEFAULT_VERSION": "6.9.1",
+        "WEBSITE_NODE_DEFAULT_VERSION": "14.18.1",
```
markhiveit, line 1203: Seems 14.18.1 isn't on the Azure app servers (or whatever they are). The looking at the service's `api/diagnostics/runtime` it seems "14.16.0" is the latest it has, so we should switch to using that.

markhiveit (changes requested):

ntsim (approved):

markhiveit (approved):

PR #2996, opened 2021-11-15
EES-2893 Fix UI tests failing due to basic auth credentials in URL

ntsim:
This changes how the UI tests perform basic authentication against the public frontend. Previously, we were putting these into the `PUBLIC_URL` variable, but unfortunately this seems to now be incompatible with Next.js now that we've upgrade to v11.

To get around this, we now add the basic auth credentials into the request headers using the Chrome DevTools prototocol. This necessitates the use of two new environment variables: `PUBLIC_AUTH_USER` and `PUBLIC_AUTH_PASSWORD`.


luke-h1 (approved):
LGTM!

PR #3000, opened 2021-11-17
Ees 2832 add timeout cancellations explicit parameter

duncan-at-hiveit:
This PR:
- adds CancellationTokens to the Controller endpoints that provide users the ability to manually query for data tables, in order to listen for user-initiated cancellation of requests.  If the user (or browser) stops listening for a table query request's response, we can cancel the expensive operation.  The framework provides these CancellationTokens to Controller methods automatically.
- adds an [AddTimeoutCancellation] attribute that can be used in conjunction with the above CancellationToken (or standalone),. to add a timeout component to the request cancellation.  If the timeout expires, the CancellationToken passed to the method will be marked as cancelled.
- propagates the CancellationTokens from Controller methods down to expensive service calls in order to allow them to observe the cancellation requests and stop processing if necessary.  On a previous branch, I was doing this via AOP with a couple of extra attributes and an AsyncLocal<CancellationToken> so as not to riddle the code with optional CancellationToken parameters, but it seemed too naughty in the end :)
- adds an OperationCancelledExceptionFilter, to convert task cancellation exceptions into nice API validation responses.

All tests passing (barring a Cookies page one that has already been fixed elsewhere).
image

UPDATE: All tests passing (22/11/21)
image


src/explore-education-statistics-common/src/modules/table-tool/components/FiltersForm.tsx
```diff
@@ -115,20 +125,30 @@ const FiltersForm = (props: Props & InjectedWizardProps) => {
...
+      setTableQueryError(undefined);
       await onSubmit(values);
-      toggleTableSizeError.off();
       goToNextStep();
     } catch (error) {
       if (
-        isServerValidationError(error, 'QUERY_EXCEEDS_MAX_ALLOWABLE_TABLE_SIZE')
+        isServerValidationError(
+          error,
+          'QUERY_EXCEEDS_MAX_ALLOWABLE_TABLE_SIZE',
```
ntsim, line 135: Not super keen on `isServerValidationError` accepting an error message argument. `isServerValidationError` is a type guard meaning that's specifically meant to narrow down the type i.e. it allows us to assert that a value is a `AxiosError<ServerValidationErrorResponse>`.

If we allow it to accept error message arguments and return true/false depending on if these are present inside of it, then this skews the semantics as receiving a `AxiosError<ServerValidationErrorResponse>` should still return true even if the response doesn't contain the specific error we're looking for.

Would suggest that we refactor this so that `isServerValidationError` no longer takes error messages to check against, and instead we use a function like `hasServerValidationError` instead

src/explore-education-statistics-common/src/modules/table-tool/components/FiltersForm.tsx
```diff
@@ -115,20 +125,30 @@ const FiltersForm = (props: Props & InjectedWizardProps) => {
...
+        ) ||
+        isServerValidationError(error, 'REQUEST_CANCELLED')
       ) {
-        if (onTableSizeError) {
-          onTableSizeError(
+        const errorCode = getServerValidationError(
```
ntsim, line 139: Could refactor `getServerValidationError` to use a type parameter i.e.

```ts
export function getServerValidationError<T extends string = string>(...): T
```

Note that we're using a default type of `string` so that we don't need to explicitly set the type in every scenario i.e. we're happy receiving a plain string.

We can then use it like this (avoiding the `as`):

```ts
const errorCode = getServerValidationError<TableQueryErrorCode>(error);

if (errorCode) { ... }
```

Note that we should be using an `if` to be safe and check that the error message/code actually exists.

src/explore-education-statistics-common/src/validation/serverValidations.ts
```diff
@@ -163,13 +163,22 @@ export function isServerValidationError(
...
-    }
+    const existingCodes = Object.values(error.response?.data.errors).flat();
+    return existingCodes.every(existingCode =>
+      errorCodes.includes(existingCode),
+    );
   }
   return false;
 }
+
+export function getServerValidationError(error: Error): string | null {
```
ntsim, line 177: Would suggest the following changes to this method:

```ts
export function getServerValidationErrorMessage<T extends string = string>(
  error: AxiosError<ServerValidationErrorResponse>, 
  fieldName = ''
): T | undefined
```

Note the following:
- A type parameter `T` has been added to let us specify the return type for string literal (instead of using `as` to coerce the type)
- The error's type has been changed to `AxiosError<ServerValidationErrorResponse>`. The intention would be to use `isServerValidationError` beforehand to assert the type in calling code. Alternatively, we could use `error: unknown`, but it could be a bit wasteful as we then potentially have to run `isServerValidationError` multiple times.
- We specify the field name to look for the error message under (see below).
- We return undefined as this is a bit terser in combination with the comment below around looking up using a field name.

src/explore-education-statistics-common/src/validation/serverValidations.ts
```diff
@@ -163,13 +163,22 @@ export function isServerValidationError(
...
+  if (!isServerValidationError(error) || !error.response?.data) {
+    return null;
+  }
+
+  const errors = Object.values(error.response?.data.errors);
+  return errors.flat()[0];
```
ntsim, line 183: This doesn't seem very safe - for example, we could pass in a server validation error where there are field errors, but no server error. It's errors look like:

```ts
errors: {
  theField: ["Message"]
}
```

With the above example and the current logic, we'd end up returning `Message`, even though it's not correct.

We should probably allow this method to specify the field name that we're looking for an error message under so that we simply need to do this:

```ts
export function getServerValidationErrorMessage<T extends string = string>(
  error: AxiosError<ServerValidationErrorResponse>, 
  fieldName = ''
): T | undefined {
  return error.response.data.errors[fieldName]?.[0];
}
```

We default to a `fieldName = ''` as this is the key corresponding to global errors.

src/GovUk.Education.ExploreEducationStatistics.Common/Cancellation/AddTimeoutCancellationAttribute.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            if (!AddTimeoutCancellationAspect.Enabled)
+            {
+                return;
+            }
+            
+            if (_timeoutConfiguration == null)
+            {
+                throw new ArgumentException("TimeoutConfiguration section cannot be null when using the " +
+                                            "AddTimeoutCancellation attribute alongside a timeout configuration key");
```
ntsim, line 43: Can use `nameof` expression to make refactoring of the attribute name easier (see other comment)

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Controllers/TableBuilderControllerTests.cs
```diff
@@ -34,12 +38,88 @@ public async Task Query()
...
+
+            var controller = BuildTableBuilderController(tableBuilderService: tableBuilderService.Object);
+            var result = await controller.Query(_query);
+
+            Assert.IsType<TableBuilderResultViewModel>(result.Value);
+            Assert.Single(result.Value.Results);
+        }
+        
+        [Fact]
+        public async Task Query_CancellationToken()
```
ntsim, line 60: Is it actually worth testing this? Seems a bit redundant if we know that the attribute works correctly in its own tests. Same for `Query_ReleaseId_CancellationToken`

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Common/Cancellation/AddTimeoutCancellationAttribute.cs
```diff
@@ -0,0 +1,68 @@
...
+using AspectInjector.Broker;
+using Microsoft.Extensions.Configuration;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Cancellation
+{
+    /// <summary>
+    /// An attribute that, when used upon a method that accepts a
+    /// CancellationToken as its final parameter, provides a 
+    /// CancellationToken with the given timeout in milliseconds
+    /// to the method as its CancellationToken argument.
```
benoutram, line 12: Just wondering if there's any way with the 'Usage' argument of attributes that it could be constrained as an attribute of the `CancellationToken` rather than adding at a method level? I suppose at the moment it could be added to a method without a `CancellationToken` and it would not be triggered.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ObservationService.cs
```diff
@@ -160,6 +81,117 @@ public IEnumerable<Observation> FindObservations(SubjectMetaQueryContext query)
...
+                        rscRegionListParam,
+                        schoolsListParam,
+                        sponsorListParam,
+                        wardListParam,
+                        planningAreaListParam,
+                        filterItemListParam);
+
+                return await inner
+                    .Select(obs => obs.Id)
+                    .ToArrayAsync(cancellationToken);
```
benoutram, line 192: Interested to know if this will stop execution of the stored procedure on the database server or if we need to make any further changes there to make it respect a cancellation?

duncan-at-hiveit: It does!  Very hard to unit test that though without guaranteeing a real DB connection.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Startup.cs
```diff
@@ -179,9 +181,12 @@ public void ConfigureServices(IServiceCollection services)
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
         public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
         {
-            // Enable caching and register any caching services
+            // Enable caching and register any caching services.
             CacheAspect.Enabled = true;
             BlobCacheAttribute.AddService("default", app.ApplicationServices.GetService<IBlobCacheService>());
+            // Enable cancellation aspects and register request timeout configuration.
+            AddTimeoutCancellationAspect.Enabled = true;
+            AddTimeoutCancellationAttribute.SetTimeoutConfiguration(Configuration.GetSection("RequestTimeouts"));
```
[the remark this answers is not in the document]
duncan-at-hiveit: I'm only enabling the aspect here because that allows us to add the timeout behaviour to the public APIs, whereas the Admin doesn't use the timeouts (yet).

The filter is used to handle the user-initiated cancellations.  Although in theory the user should never see these responses, it does prevent a big stacktrace getting logged out!

infrastructure/parameters/dev.parameters.json
```diff
@@ -73,6 +73,9 @@
     },
     "maxDbSizeBytes": {
       "value": 322122547200
+    },
+    "requestTimeoutsTableBuilderQuery": {
+      "value": 300000
```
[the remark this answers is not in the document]
duncan-at-hiveit: I put the Dev and Local amounts high, because Dev is a slow environment and Local can be slow for individual responses  when using Pabot.  For the other environments, I wanted them to be more useful timeouts, for when something might be excessively long.  To be fair, maybe Prod, Test and PreProd should be the same.  I will put them all to 60 seconds and I'll run this timeout by Cam.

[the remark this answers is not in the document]
duncan-at-hiveit: I'm also going to put it in each environment's parameters file explicitly, even if it is the default I think.  Just makes it clearer how each environment is laid out at a glance.

src/GovUk.Education.ExploreEducationStatistics.Common/Cancellation/AddTimeoutCancellationAspect.cs
```diff
@@ -0,0 +1,84 @@
...
+            [Argument(Source.Metadata)] MethodBase method,
+            [Argument(Source.ReturnType)] Type returnType,
+            [Argument(Source.Triggers)] Attribute[] triggers)
+        {
+            if (method.GetParameters().Length == 0 
+                || (method.GetParameters()[^1].ParameterType != typeof(CancellationToken)
+                    && method.GetParameters()[^1].ParameterType != typeof(CancellationToken?)))
+            {
+                throw new ArgumentException($"Method {method.Name} annotated with the " +
+                                            $"AddTimeoutCancellation attribute must accept a CancellationToken as " +
```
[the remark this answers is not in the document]
duncan-at-hiveit: Will do, thanks

src/GovUk.Education.ExploreEducationStatistics.Common/Cancellation/AddTimeoutCancellationAttribute.cs
```diff
@@ -0,0 +1,68 @@
...
+    ///
+    /// If a CancellationToken has been passed to this method as an
+    /// argument already by the calling code, this attribute will
+    /// create a new linked CancellationToken with the one originally
+    /// passed in, thus allowing cancellation to be initiated by either
+    /// the calling code or the timeout itself. 
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method)]
+    [Injection(typeof(AddTimeoutCancellationAspect))]
+    public class AddTimeoutCancellationAttribute : Attribute
```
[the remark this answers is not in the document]
duncan-at-hiveit: Sounds good - updated

src/GovUk.Education.ExploreEducationStatistics.Common/Cancellation/AddTimeoutCancellationAspect.cs
```diff
@@ -0,0 +1,84 @@
...
+        public object Handle(
+            [Argument(Source.Target)] Func<object[], object> target,
+            [Argument(Source.Arguments)] object[] args,
+            [Argument(Source.Metadata)] MethodBase method,
+            [Argument(Source.ReturnType)] Type returnType,
+            [Argument(Source.Triggers)] Attribute[] triggers)
+        {
+            if (method.GetParameters().Length == 0 
+                || (method.GetParameters()[^1].ParameterType != typeof(CancellationToken)
+                    && method.GetParameters()[^1].ParameterType != typeof(CancellationToken?)))
```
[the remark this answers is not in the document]
duncan-at-hiveit: As discussed, I couldn't find any documentation or samples of this usage with Aspect Injector unfortunately.  I agree this would be great.

We decided that it'd be best not to constrain the CancellationToken to just be the final parameter of a method instead, which has been done.

src/explore-education-statistics-common/src/modules/table-tool/components/TableQueryError.tsx
```diff
@@ -26,13 +29,28 @@ const TableSizeError = ({
...
-          <p>Select different filters or download the subject data.</p>
+          {errorCode === 'QUERY_EXCEEDS_MAX_ALLOWABLE_TABLE_SIZE' && (
+            <p>Select different filters or download the subject data.</p>
+          )}
+          {errorCode === 'REQUEST_CANCELLED' && (
+            <p>
+              Select different filters, try again later or download the subject
+              data.
+            </p>
+          )}
```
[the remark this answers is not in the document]
duncan-at-hiveit: I was glad you suggested this cos I didn't like it either!  I had it in my head that you couldn't do a switch statement over a string literal's values - evidently not!

src/GovUk.Education.ExploreEducationStatistics.Common/Cancellation/AddTimeoutCancellationAttribute.cs
duncan-at-hiveit: Done ta

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/AssertExtensions.cs
```diff
@@ -9,7 +9,7 @@ public static class AssertExtensions
          * Calling this method causes a Test to fail with the given message.  The equivalent of `Assert.Fail()` in
          * other testing frameworks.
          */
-        public static void AssertFail(string message) 
+        public static XunitException AssertFail(string message)
```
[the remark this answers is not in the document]
duncan-at-hiveit: I didn't end up using it in this PR, but it allows you to write `throw AssertFail(...);` to break out of code early e.g.

```
if (somethingUnexpected) {
    throw AssertFail("Wrong");
}
// some more checking here
```

src/GovUk.Education.ExploreEducationStatistics.Common/Cancellation/AddTimeoutCancellationAttribute.cs
duncan-at-hiveit: Couldn't find it, but see answer to Nick's similar comment.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ObservationService.cs
benoutram: Great to hear 🙂

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Startup.cs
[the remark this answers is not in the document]
benoutram: Got it, thanks. I think it's worth dropping a note to @luke-h1 to mention that the timeout feature won't apply to the Admin just yet, for when it comes to testing.

infrastructure/parameters/dev.parameters.json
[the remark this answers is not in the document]
benoutram: Cool thanks for the explanation about allowing extra time in the Dev environment. 🙂 

I've not checked this but for existing parameters I don't think we've written each environment file with all the possible parameters included explicitly. It might be worth thinking about being consistent about it so we are either explicit about all parameters or include overridden parameters only.

src/explore-education-statistics-common/src/modules/table-tool/components/TableQueryError.tsx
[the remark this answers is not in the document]
duncan-at-hiveit: Done, thanks

src/explore-education-statistics-common/src/validation/serverValidations.ts
```diff
@@ -163,13 +163,22 @@ export function isServerValidationError(
 
 export function hasServerValidationError(
   error: Error,
-  errorMessage: string,
+  ...errorCodes: string[]
```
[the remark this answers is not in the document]
duncan-at-hiveit: Np - method replaced anyway, but I'd be interested to know why not to use spread params?  Always seemed a bit more flexible in use to me.

src/explore-education-statistics-common/src/modules/table-tool/components/FiltersForm.tsx
duncan-at-hiveit: Yup nice, nice.  Figured out the type guards that were in place, so that was a nice learning experience.  Good improvement.

src/explore-education-statistics-common/src/validation/serverValidations.ts
duncan-at-hiveit: Nice improvements, and done, thanks.

src/explore-education-statistics-common/src/validation/serverValidations.ts
duncan-at-hiveit: Cool, have done ta.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Startup.cs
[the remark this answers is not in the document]
duncan-at-hiveit: Will do, ta

src/explore-education-statistics-common/src/validation/serverValidations.ts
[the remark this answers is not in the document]
ntsim: I personally find they're generally less flexible as they have specific rules. They're particularly offending as they make it so that you can't just add new parameter to the end of the method args without going to effort to refactor all the existing callsites.

Of course, there are types of APIs that suit spread params, but don't think this kind of thing is particularly it.

Can understand using spread params more in languages where we have to do things like `new List()`, but in JS, there's very pretty much zero effort between doing the two (besides some square brackets):

```js
myFunction(['a', 'b', 'c']);
// vs
myFunction('a', 'b', 'c');
```

src/GovUk.Education.ExploreEducationStatistics.Common/Cancellation/CancellationTokenTimeoutAspect.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            var cancellationTokenParamPosition = method
+                .GetParameters()
+                .ToList()
+                .IndexOf(cancellationTokenParameters.Single());
+            
+            if (!Enabled)
+            {
+                return target(args);
+            }
```
ntsim, line 54: Not a biggie, but could do this right at the start of the method to avoid unnecessary work before this point

src/GovUk.Education.ExploreEducationStatistics.Common/Cancellation/CancellationTokenTimeoutAspect.cs
```diff
@@ -0,0 +1,89 @@
...
+                .Single();
+
+            var timeoutToken = new CancellationTokenSource(trigger.TimeoutMillis).Token;
+            var passedInToken = args[cancellationTokenParamPosition] as CancellationToken?;
+            var newArgs = args.ToArray();
+            newArgs[cancellationTokenParamPosition] = CombineTokens(passedInToken, timeoutToken);
+            return target(newArgs);
+        }
+
+        private static CancellationToken CombineTokens(params CancellationToken?[] tokens)
```
ntsim, line 67: Is it worth extracting this function into a utility function in common?

src/explore-education-statistics-common/src/validation/serverValidations.ts
```diff
@@ -132,44 +132,76 @@ export function convertServerFieldErrors<FormValues>(
...
+ *
+ * If any are included in this error object, this method will return true.
+ *
+ * @param error - the error object to check the type of.
+ * @param errorMessages - array of error messages, any of which can appear
+ * in this error object in order for this method to return true.
+ * @param fieldName - optional fieldName that, if omitted, will be treated
+ * as checking for global errors.
+ */
+export function isAnyServerValidationError<T extends string = string>(
```
ntsim, line 176: The name for this is a bit weird as it doesn't really move away from my last comment as this still looks like a type guard and contorts its semantics. A type guard should return true if the type is correct, but having additional logic that says it should return false in other situations is a bit misleading.

Would suggest renaming it to something like `hasErrorMessages`?

src/explore-education-statistics-common/src/validation/serverValidations.ts
```diff
@@ -132,44 +132,76 @@ export function convertServerFieldErrors<FormValues>(
...
+ * Asserts whether the given error object is an
+ * {@link AxiosError<ServerValidationErrorResponse<T>>}.
+ *
+ * See {@link isServerValidationError}
+ *
+ * Additionally, this method checks whether or not the error contains any of
+ * the error messages provided, using {@param fieldName} to determine field
+ * validation, or global validation if {@param fieldName} is omitted.
+ *
+ * If any are included in this error object, this method will return true.
```
ntsim, line 168: Rather than lead with telling about the type assertion, would suggest making the focus of this docblock about the method checking that the error response contains a specific message (see below's comment about changing the name to `hasErrorMessages` as well.

src/explore-education-statistics-common/src/validation/serverValidations.ts
```diff
@@ -132,44 +132,76 @@ export function convertServerFieldErrors<FormValues>(
...
+ *
+ * @param error - the error object to check the type of.
+ * @param errorMessages - array of error messages, any of which can appear
+ * in this error object in order for this method to return true.
+ * @param fieldName - optional fieldName that, if omitted, will be treated
+ * as checking for global errors.
+ */
+export function isAnyServerValidationError<T extends string = string>(
+  error: unknown,
+  errorMessages?: readonly T[],
```
ntsim, line 178: Could make this parameter required? In combination with renaming it to `hasErrorMessages`, this method signature would make more sense.

ntsim (changes requested):

src/explore-education-statistics-common/src/validation/serverValidations.ts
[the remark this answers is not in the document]
duncan-at-hiveit: Good stuff, thanks for explanation.

src/GovUk.Education.ExploreEducationStatistics.Common/Cancellation/CancellationTokenTimeoutAspect.cs
duncan-at-hiveit: I wanted to put it a little lower down so that we'd know straight off the bat if an attribute had been used in an invalid place is all, otherwise I would have put it right at the top as suggested.

src/explore-education-statistics-common/src/validation/serverValidations.ts
duncan-at-hiveit: Yup sounds fine to me!

src/explore-education-statistics-common/src/validation/serverValidations.ts
duncan-at-hiveit: Yup, removed the type checking altogether as suggested, in favour of checking type ahead of time with existing function.

src/explore-education-statistics-common/src/validation/serverValidations.ts
duncan-at-hiveit: Done

ntsim (approved):

PR #3001, opened 2021-11-18
EES-2844 Fix table headers alignments for row/col headers with large cross spans

ntsim:
This PR fixes table cell misalignment that was being caused by row/col headers not taking up the correct cross span. This was caused by a bug meaning that headers only had a max cross span of 2. We have now fixed this so that headers take up the correct cross spans and fill the table cells correctly.

Ironically, due to recent changes limiting the table tool's maximum number of filters, it's not really possible to generate the table that demonstrates the issue in the ticket anymore.

## Relevant changes

- Refactored most of `MultiHeaderTable` to construct it's expanded headers in a more imperative way. This is hopefully a little simpler to understand as there's less going on. It also allows us to take advantage of breadth/depth first search algorithms to try improve performance.

## Screenshots

image


amyb-hiveit (approved):
Not going to pretend I fully understand all of this (though the comments and tests help, thanks!), but think it looks good to merge.

PR #3002, opened 2021-11-18
EES-2896 - setup slack notifications for UI tests

luke-h1:
This PR: 
- enables test reports to be sent via slack to the #build channel 
- formats various python files to PEP8 standards

### DIFF
--- a/tests/robot-tests/tests/libs/slack.py
+++ b/tests/robot-tests/tests/libs/slack.py
@@ -1,0 +1,85 @@
+import os
+from slack_sdk import WebClient
+from slack_sdk.errors import SlackApiError
+import shutil
+import json
+import requests
+from bs4 import BeautifulSoup
+
+PATH = f'{os.getcwd()}{os.sep}test-results'
+
+
+def _generate_slack_attachments():
+    failed_tests = dict()
+    passed_tests = dict()
+
+    report = open(f'{PATH}{os.sep}output.xml', 'rb')
+    contents = report.read()
+
+    soup = BeautifulSoup(contents, 'xml')
+    test = soup.find('total').find('stat')
+
+    failed_tests = int(test['fail'])
+    passed_tests = int(test['pass'])
+
+    return [
+        {
+            "pretext": "All results",
+            "color": "danger" if failed_tests else "good",
+            "mrkdwn_in": [
+                "pretext"
+            ],
+            "fields": [
+                {
+                    "title": "Total test cases",
+                    "value": passed_tests + failed_tests
+                },
+                {
+                    "title": "Passed tests",
+                    "value": passed_tests
+                },
+                {
+                    "title": "Failed tests",
+                    "value": failed_tests if failed_tests else "0"
+                },
+                {
+                    "title": "Results",
+                    "value": "Failed" if failed_tests else "Passed"
+                },
+
+            ]
+        }
+    ]
+
+
+def send_slack_report():
+    attachments = _generate_slack_attachments()
+    data = {"attachments": attachments}
+
+    webhook_url = os.getenv('SLACK_TEST_REPORT_WEBHOOK_URL')
+    slack_bot_token = os.getenv('SLACK_BOT_TOKEN')
+
+    assert webhook_url, print("SLACK_TEST_REPORT_WEBHOOK_URL env variable needs to bet set")
+    assert slack_bot_token, print("SLACK_BOT_TOKEN env variable needs to bet set")
+
+    response = requests.post(
+        url=webhook_url,
+        data=json.dumps(data), headers={'Content-Type': 'application/json'}
+    )
+    assert response.status_code == 200, print(f"Response wasn't 200, it was {response}")
+
+    print("Sent UI test statistics to #build")
+
+    client = WebClient(token=slack_bot_token)
+    print('Sending UI test report to #build')
+
+    shutil.make_archive('UI-test-report', 'zip', PATH)
+    try:
+        client.files_upload(
+            channels='#build',
+            file='UI-test-report.zip',
+            title='test-report.zip',
+        )
+    except SlackApiError as e:
+        print(f'Error uploading test report: {e}')
+    os.remove('UI-test-report.zip')
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ src/explore-education-statistics-frontend/*.xml
 /tests/robot-tests/backup-data
 /tests/robot-tests/webdriver
 /tests/robot-tests/.env.*
+/tests/robot-tests/UI-test-report.zip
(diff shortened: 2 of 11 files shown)


ntsim (changes requested):

tests/robot-tests/tests/libs/slack.py
```diff
@@ -0,0 +1,96 @@
...
+
+    report = open(f'{PATH}{os.sep}output.xml', 'rb')
+    contents = report.read()
+
+    soup = BeautifulSoup(contents, 'xml')
+    test = soup.find('total').find('stat')
+
+    failed_tests = test['fail']
+    passed_tests = test['pass']
+    skipped_tests = test['skip']
```
ntsim, line 25: Very nice! Looked pretty easy in the end :smile:

tests/robot-tests/tests/libs/slack.py
```diff
@@ -0,0 +1,96 @@
...
+
+    return [
+        {
+            "pretext": "All results",
+            "color": "danger" if failed_tests or skipped_tests is not None or not "" else "good",
```
ntsim, line 30: Think we could simplify this to: `"danger" if failed_tests or skipped_tests else "good"`

Python has pretty similar boolean coercion rules to JS so things like empty strings, 0 and None are already considered as falsy.

Would suggest going through the other conditionals in this file and apply similar treatment

tests/robot-tests/tests/libs/slack.py
```diff
@@ -0,0 +1,96 @@
...
+
+    assert webhook_url is not None or not "", print("SLACK_TEST_REPORT_WEBHOOK_URL env variable needs to bet set")
+    assert slack_bot_token is not None or not "", print("SLACK_BOT_TOKEN env variable needs to bet set")
+
+    requests.sessions.HTTPAdapter(
+            pool_connections=50,
+            pool_maxsize=50,
+            max_retries=3
+        )
+    session = requests.Session()
```
ntsim, line 72: I'm guessing this is the connection pooling stuff? What benefit does it provide, given that we run this right at the end of the test run? 

My assumption would be that there'd be no benefit as we're only making a single request and the test runner is about to complete anyway. Wouldn't the work to set up the connection pool would just be thrown away?

tests/robot-tests/tests/libs/slack.py
```diff
@@ -0,0 +1,96 @@
...
+                    "title": "Total test cases",
+                    "value": int(skipped_tests) + int(passed_tests) + int(failed_tests)
+                },
+                {
+                    "title": "Passed tests",
+                    "value": int(passed_tests)
+                },                
+                {
+                    "title": "Failed tests",
+                    "value": int(failed_tests) + int(skipped_tests) if failed_tests or skipped_tests is not None or not "" else "0"
```
ntsim, line 45: Why are skipped tests considered failing? This seems like a false positive to me

ntsim (changes requested):

tests/robot-tests/tests/libs/slack.py
luke-h1: Yep haha, had a few issues with Beautiful soup initially. It seemed to not like local HTML files and wouldn't read the local `report.html` file so Mark suggested getting the info from XML which turned out to be pretty simple

tests/robot-tests/tests/libs/slack.py
luke-h1: Yep, that's a good point. I've refactored this to use `requests.post`

tests/robot-tests/tests/libs/slack.py
luke-h1: Ah yes, that's a good catch. I initially thought that tests get marked as skipped if we fail them fast (due to the logic in https://github.com/dfe-analytical-services/explore-education-statistics/blob/dev/tests/robot-tests/listeners/KeywordListener.py)

tests/robot-tests/tests/libs/slack.py
```diff
@@ -0,0 +1,86 @@
...
+
+def send_slack_report():
+    attachments = _generate_slack_attachments()
+    data = {"attachments": attachments}
+
+    webhook_url = os.getenv('SLACK_TEST_REPORT_WEBHOOK_URL')
+    slack_bot_token = os.getenv('SLACK_BOT_TOKEN')
+
+    assert webhook_url is not None or not "", print("SLACK_TEST_REPORT_WEBHOOK_URL env variable needs to bet set")
+    assert slack_bot_token is not None or not "", print("SLACK_BOT_TOKEN env variable needs to bet set")
```
ntsim, line 63: Can simplify these conditionals too

ntsim (approved):

tests/robot-tests/tests/libs/slack.py
```diff
@@ -0,0 +1,86 @@
...
+        {
+            "pretext": "All results",
+            "color": "danger" if failed_tests else "good",
+            "mrkdwn_in": [
+                "pretext"
+            ],
+            "fields": [
+                {
+                    "title": "Total test cases",
+                    "value": int(passed_tests) + int(failed_tests)
```
markhiveit, line 35: I'd do these int conversions where `failed_tests` and `passed_tests` are instantiated. So,
```
failed_tests = int(test['fail'])
passed_tests = int(test['pass'])
```

tests/robot-tests/tests/libs/slack.py
```diff
@@ -0,0 +1,86 @@
...
+
+    assert webhook_url, print("SLACK_TEST_REPORT_WEBHOOK_URL env variable needs to bet set")
+    assert slack_bot_token, print("SLACK_BOT_TOKEN env variable needs to bet set")
+
+    response = requests.post(
+        url=webhook_url,
+        data=json.dumps(data), headers={'Content-Type': 'application/json'}
+    )
+    assert response.status_code == 200, print(
+        "Response wasn't 200, it was: {}".format({response}))
```
markhiveit, line 70: It's not a big deal, but I'd switch using an f-string here to be consistent with how we do things elsewhere:
`f"Response wasn't 200, it was {response}"`

tests/robot-tests/tests/libs/slack.py
```diff
@@ -0,0 +1,86 @@
...
+
+    shutil.make_archive('UI-test-report', 'zip', PATH)
+    try:
+        client.files_upload(
+            channels='#build',
+            file='UI-test-report.zip',
+            title='test-report.zip',
+        )
+    except SlackApiError as e:
+        print('Error uploading test report: {}'.format(e))
```
markhiveit, line 85: I'd use an f-string here too

markhiveit (changes requested):

ntsim (approved):

PR #3009, opened 2021-11-26
EES-2773 Display location hierarchies in location step of table tool

ntsim:
This PR allow location hierarchies to now be displayed in the table tool on its location step.  As part of this, we have removed most of the legacy code from the backend API and made the following additional changes:

- Changed the configured local authority hierarchy from `Country > Region > Local authority` to `Region > Local authority` in-line with the described example in the ticket.
- Added a local authority district hierarchy of  `Region > Local authority district` (the same as LA's)
- Re-added de-duplication logic that was previously present for the previous non-hierarchical location meta. This was accidentally omitted and is required to prevent locations with the same label being duplicated in the frontend.
- Added ordering of regions by their code as this is the intended ordering according to Cam. It seems like we may not even need any reference file to order locations.

## Other changes

- Fixed `LocationAttributeNode` having a nullabe `Attribute` property. This didn't make a whole lot of sense and it appears like it was made nullable just to satisfy the compiler. We have now made this property non-nullable. Consequently, a bunch of potential null reference warnings can now be avoided.

## Screenshots

image


src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
```diff
@@ -335,16 +297,6 @@ private string GetTotalValue(Filter filter)
             return _filterItemRepository.GetTotal(filter)?.Id.ToString() ?? string.Empty;
         }
 
-        private static LabelValue MapLocationAttributeToLabelValue(ILocationAttribute unit)
```
benoutram, line 338: Seeing this get removed has just made me think we might be missing something similar in `BuildLocationAttributeViewModel` where we need to `GetCodeOrOldCodeIfEmpty` for local authorities.

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/SubjectMetaServiceTests.cs
```diff
@@ -509,12 +499,422 @@ await using (var statisticsDbContext = InMemoryStatisticsDbContext(contextId))
...
+            {
+                Release = release,
+                Subject = subject
+            };
+
+            var locations = new Dictionary<GeographicLevel, List<LocationAttributeNode>>
+            {
+                {
+                    GeographicLevel.LocalAuthority,
+                    // Region-LA hierarchy in the LA level data
```
benoutram, line 518: I think this comment is wrong as there is no hierarchy.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
```diff
@@ -306,18 +254,32 @@ private TimePeriodsMetaViewModel GetTimePeriods(IQueryable<Observation> observat
...
+                    Value = locationAttributeNode.Attribute.Code ?? string.Empty,
+                    Options = DeduplicateLocationViewModels(
+                            locationAttributeNode.Children.Select(BuildLocationAttributeViewModel)
+                        )
+                        .OrderBy(OrderLocationViewModel)
+                        .ToList()
                 };
         }
 
+        private static string OrderLocationViewModel(LocationAttributeViewModel option)
```
benoutram, line 273: Cool, thanks for sorting out the ordering! 🙂 I knew there was no ordering in the new stuff but didn't highlight it while it was turned off by default because we had upcoming tickets to deal with ordering. It will be cool if this satisfies all the ordering requirements.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/Meta/LocationsMetaViewModels.cs
```diff
@@ -19,21 +18,12 @@ public interface ILocationsMetaViewModel
...
         public string? Level { get; init; }
-        public List<LocationAttributeViewModel>? Options { get; init; }
-    }
 
-    /// <summary>
-    /// Legacy locations view model returned in Subject meta data - to be removed
-    /// </summary>
-    [ObsoleteAttribute("TODO EES-2902 - Remove with SOW8 after EES-2773", false)]
-    public class ObservationalUnitsMetaViewModel :
-        LegendOptionsMetaValueModel<IEnumerable<LabelValue>>, ILocationsMetaViewModel
```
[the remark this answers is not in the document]
ntsim: Ah good shout, I didn't notice it without the TODO. Removed!

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/Meta/SubjectMetaViewModel.cs
```diff
@@ -23,12 +23,6 @@ public interface ISubjectMetaViewModel
         public TimePeriodsMetaViewModel TimePeriod { get; set; }
     }
 
-    /// <summary>
-    /// Legacy SubjectMetaViewModel to be removed
-    /// </summary>
-    [ObsoleteAttribute("TODO EES-2902 - Remove with SOW8 after EES-2773", false)]
-    public record LegacySubjectMetaViewModel : AbstractSubjectMetaViewModel<ObservationalUnitsMetaViewModel>;
```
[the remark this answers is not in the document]
ntsim: Awesome, removed!

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/SubjectMetaServiceTests.cs
ntsim: Removed!

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
ntsim: Good shout, I've fixed this across both services now so that the returned `LocationAttributeViewModel.Value` fields are set to a new `ILocationAttribute.GetCodeOrFallback` method (which gets the old code in the case of local authorities).

benoutram (approved):

PR #3010, opened 2021-11-29
    EES-2930 - fix UI tests

luke-h1:


tests/robot-tests/tests/admin/bau/release_content.robot
```diff
@@ -79,35 +79,35 @@ Add Useful information related page link to release
     user clicks button    Create link
 
 Add secondary statistics
-    user waits until page does not contain element    ${SECONDARY_STATS_TABLE_TAB_SELECTOR}
+    user waits until page does not contain element    //*[@id="${SECONDARY_STATS_TABLE_TAB_SELECTOR}"]
```
markhiveit, line 82: I think it's worth changing the variable name here to something like `SECONDARY_STATS_TABLE_TAB_ID`.

markhiveit: I think you can just use `id:${SECONDARY_STATS_TABLE_TAB_ID}` here rather than using xpath. There are other places to make this change too.

tests/robot-tests/tests/admin/bau/release_content.robot
```diff
@@ -79,35 +79,35 @@ Add Useful information related page link to release
...
-    user checks select contains option    css:select[name="selectedDataBlock"]    Key Stats Data Block 1
-    user checks select contains option    css:select[name="selectedDataBlock"]    Key Stats Data Block 2
+    user waits until page contains element    //*[@name="selectedDataBlock"]    %{WAIT_MEDIUM}
+    user checks select contains x options    //*[@name="selectedDataBlock"]    5
+    user checks select contains option    //*[@name="selectedDataBlock"]    Select a data block
+    user checks select contains option    //*[@name="selectedDataBlock"]    Data Block 1
+    user checks select contains option    //*[@name="selectedDataBlock"]    Data Block 2
+    user checks select contains option    //*[@name="selectedDataBlock"]    Key Stats Data Block 1
+    user checks select contains option    //*[@name="selectedDataBlock"]    Key Stats Data Block 2
     user chooses and embeds data block    Data Block 1
```
markhiveit, line 91: With all of these rows, rather than using an xpath selector you should be able to do `name:selectedDataBlock`. The same change can be made elsewhere too.

markhiveit (changes requested):

markhiveit (approved):

PR #3012, opened 2021-11-30
Hotfix/ees 2924 speed up filteritem retrieval

duncan-at-hiveit:
This PR:
- optimises the case where a final set of Observations have been selected from a table query and we retrieve the FilterItems that are included in just the Observations that were retrieved.

Prior to this PR, we selected between an "optimal" query and a "suboptimal" query using a `listFilterItems` flag.  Three things were changed here:

1. The flag is removed in favour of 2 distinct methods with obvious names and descriptions, thereby hopefully making clearer what the intentions of the 2 different code paths are.
2. The "suboptimal" query is replaced with a much faster query which, in the case of the QL01 Subject file, cut down the Filter Item retrieval time from 8000ms to 5ms, thereby shaving about 40% of the time off the overall table query locally.
3. The 2 `FindObservations()` methods in `ObservationService` now more clearly indicate what they return - a List of in-memory Observations, and a Queryable set of Observations that have not yet been retrieved.

There are lots of areas of our code whereby we still accept only an IQueryable<Observation>, and I suspect that a good deal of these might be able to be optimised further by separating out the IList<Observation> cases from the IQueryable<Observation> cases.  For this PR though, we're just focusing on the main offender which was the retrieval of large numbers of FilterItems.

image


benoutram:
While doing the Location hierarchy work I saw that some of the interfaces are using IQueryable when they are only ever called with a list and could be cleaned up to require a List instead. I think we should do that soon to make things clearer and more reusable, avoiding calling AsQueryable on a list which I think can fail in certain cases like after `ToListAsync`.

It will then be easier to spot which cases need IQueryable and might make it easier to spot performance gains.

benoutram (approved):

benoutram:
I think existing changes on the Dev branch mean that merging master back to Dev won't be as straightforward as normal.

PR #3017, opened 2021-12-02
EES-2936 - add environment information to UI test slack message

luke-h1:


tests/robot-tests/tests/libs/slack.py
```diff
@@ -8,8 +8,26 @@
 
 PATH = f'{os.getcwd()}{os.sep}test-results'
 
+def _get_environment():
```
markhiveit, line 11: Rather than get the environment this way, you could pass it as an argument to `send_slack_report`, as we have `args.env` in `run_tests.py`

markhiveit (changes requested):

tests/robot-tests/tests/libs/slack.py
```diff
@@ -9,7 +9,8 @@
 PATH = f'{os.getcwd()}{os.sep}test-results'
 
 
-def _generate_slack_attachments():
+def _generate_slack_attachments(env: str):
+
     failed_tests = dict()
     passed_tests = dict()
```
markhiveit, line 15: Not part of this PR, but I noticed these two lines can be removed

markhiveit (changes requested):

tests/robot-tests/tests/libs/slack.py
```diff
@@ -9,15 +9,15 @@
...
-    report = open(f'{PATH}{os.sep}output.xml', 'rb')
-    contents = report.read()
-
-    soup = BeautifulSoup(contents, 'xml')
-    test = soup.find('total').find('stat')
+    with open(f'{PATH}{os.sep}output.xml', 'rb') as report:
+        contents = report.read()
+        soup = BeautifulSoup(contents, 'xml')
+        test = soup.find('total').find('stat')
+        report.close()
```
markhiveit, line 20: This `report.close()` can be removed. `.close()` is automatically called when the `with` scope is exited

tests/robot-tests/tests/libs/slack.py
```diff
@@ -9,15 +9,15 @@
...
 
-    report = open(f'{PATH}{os.sep}output.xml', 'rb')
-    contents = report.read()
-
-    soup = BeautifulSoup(contents, 'xml')
-    test = soup.find('total').find('stat')
+    with open(f'{PATH}{os.sep}output.xml', 'rb') as report:
+        contents = report.read()
+        soup = BeautifulSoup(contents, 'xml')
+        test = soup.find('total').find('stat')
```
markhiveit, line 19: These two lines can be moved outside the `with` scope. We only need the `report` file handler for the `.read()`.

markhiveit (changes requested):

markhiveit (approved):

PR #3018, opened 2021-12-02
EES-2939 - add lxml to python dependencies

luke-h1:


Pipfile
```diff
@@ -17,6 +17,7 @@ pytz = "~=2021.3"
 beautifulsoup4 = "~=4.10.0"
 robotframework-tidy = "~=1.4.0"
 slack-sdk = "~=3.11.2"
+lxml = "*"
```
markhiveit, line 20: This isn't particularly important in this instance, but so far we've explicitly stated version numbers for dependencies. This means that we can't get unexpectedly hit by a new version of a package breaking our code.

markhiveit (changes requested):

markhiveit (approved):

PR #3019, opened 2021-12-02
EES-2841 Improve importer failure messaging

markhiveit:
This work improves the import messaging if a subject import fails.

image

Firstly, we no longer add exception messages as errors, as they aren't
useful to users - for debugging these can still be found in AppInsights.

Secondly, we displays a generic message for all failed imports
suggesting analysts try running their data files through the data
screener.

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/data/components/ImporterStatus.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/data/components/ImporterStatus.tsx
@@ -7,6 +7,7 @@ import Details from '@common/components/Details';
 import LoadingSpinner from '@common/components/LoadingSpinner';
 import ProgressBar from '@common/components/ProgressBar';
 import Tag, { TagProps } from '@common/components/Tag';
+import WarningMessage from '@common/components/WarningMessage';
 import useInterval from '@common/hooks/useInterval';
 import useMounted from '@common/hooks/useMounted';
 import React, { useCallback, useEffect, useState } from 'react';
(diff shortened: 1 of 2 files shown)


src/explore-education-statistics-admin/src/pages/release/data/components/ImporterStatus.tsx
```diff
@@ -161,17 +161,28 @@ const ImporterStatus = ({
...
-        >
-          <ul className="govuk-!-margin-top-0">
-            {currentStatus.errors.map((error, index) => (
-              <li key={index.toString()}>{error}</li>
-            ))}
-          </ul>
-        </Details>
+      {currentStatus.status === 'FAILED' && (
+        <>
+          {currentStatus.errors && currentStatus.errors.length > 0 && (
```
luke-h1, line 166: We could use optional chaining here to be more succinct: 
```
currentStatus?.errors?.length > 0 &&
```

luke-h1 (changes requested):

src/explore-education-statistics-admin/src/pages/release/data/components/ImporterStatus.tsx
ntsim: Unfortunately, you can't actually do this as the compiler won't let you compare undefined with 0 :smile:

src/explore-education-statistics-admin/src/pages/release/data/components/ImporterStatus.tsx
```diff
@@ -161,17 +161,28 @@ const ImporterStatus = ({
...
+              summary="See errors"
+            >
+              <ul className="govuk-!-margin-top-0">
+                {currentStatus.errors.map((error, index) => (
+                  <li key={index.toString()}>{error}</li>
+                ))}
+              </ul>
+            </Details>
+          )}
+          <p className="govuk-inset-text">
```
ntsim, line 178: Could use the `WarningMessage` component to make it a little more obvious that they need to do something?

src/explore-education-statistics-admin/src/pages/release/data/components/ImporterStatus.tsx
luke-h1: ah that's true, good catch haha

src/explore-education-statistics-admin/src/pages/release/data/components/ImporterStatus.tsx
```diff
@@ -161,17 +161,28 @@ const ImporterStatus = ({
...
-        </Details>
+      {currentStatus.status === 'FAILED' && (
+        <>
+          {currentStatus.errors && currentStatus.errors.length > 0 && (
+            <Details
+              className="govuk-!-margin-top-1 govuk-!-margin-bottom-0"
+              summary="See errors"
+            >
+              <ul className="govuk-!-margin-top-0">
+                {currentStatus.errors.map((error, index) => (
```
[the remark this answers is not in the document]
markhiveit: There are error ids in the backend, but they're not passed to the frontend. So index is the best we have right now without passing error ids, which I think would be more work than it's worth. So I'm going to leave this.

luke-h1 (approved):

PR #3023, opened 2021-12-03
EES-2941 - handle blob exists exceptions

luke-h1:
This PR: 
- handles `ResourceExist ` exceptions related to blob files already existing in local storage emulator environments 
- Moves `create_emulator_release_files` into the tests directory (in order to call various methods in `run_tests.py`)
- Runs `create_emulator_release_files` as part of `run_tests.py`

### DIFF
--- a/tests/robot-tests/tests/libs/create_emulator_release_files.py
+++ b/tests/robot-tests/tests/libs/create_emulator_release_files.py
@@ -64,7 +64,11 @@ class ReleaseFilesGenerator(object):
 
         # Instantiate a new ContainerClient for the emulator
         self.blob_service_client = BlobServiceClient.from_connection_string(
-            "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://data-storage:10000/devstoreaccount1")
+            "DefaultEndpointsProtocol=http;" +
+            "AccountName=devstoreaccount1;" +
+            "AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;" +
+            "BlobEndpoint=http://data-storage:10000/devstoreaccount1"
+        )
 
         exclusions_release = Release("e7774a74-1f62-4b76-b9b5-84f14dac7278")
         pupil_absence_release = Release("4fa4fe8e-9a15-46bb-823f-49bf8e0cdec5")
@@ -161,20 +165,24 @@ class ReleaseFilesGenerator(object):
         try:
             container_client = self.blob_service_client.create_container(
                 "downloads")
-        except ResourceExistsError:
+        except ResourceExistsError as e:
+            print(e)
+
             container_client = self.blob_service_client.get_container_client(
                 "downloads")
 
-        self.upload_files(container_client,
-                          self.all_files_zip_files)
-        self.upload_files(container_client,
-                          self.data_files)
+            self.upload_files(container_client,
+                              self.all_files_zip_files)
+
+            self.upload_files(container_client,
+                              self.data_files)
 
     def create_private_release_files(self):
         try:
             container_client = self.blob_service_client.create_container(
                 "releases")
-        except ResourceExistsError:
+        except ResourceExistsError as e:
+            print(e)
             container_client = self.blob_service_client.get_container_client(
                 "releases")
 
@@ -183,13 +191,18 @@ class ReleaseFilesGenerator(object):
         self.upload_files(container_client,
                           self.metadata_files)
 
-    def upload_files(self, container_client, files):
-        data = b'abcd'*128
+    @staticmethod
+    def upload_files(container_client, files):
+        data = b'abcd' * 128
         for file in files:
             path = file.path()
+
             blob_client = container_client.get_blob_client(path)
-            blob_client.upload_blob(
-                data, blob_type="BlockBlob", metadata=file.metadata())
+            try:
+                blob_client.upload_blob(
+                    data, blob_type="BlockBlob", metadata=file.metadata())
+            except ResourceExistsError as e:
+                print(e)
 
 
 if __name__ == '__main__':
--- a/tests/robot-tests/run_tests.py
+++ b/tests/robot-tests/run_tests.py
@@ -8,24 +8,24 @@ Run 'python run_tests.py -h' to see argument options
 
 import argparse
 import cProfile
+import datetime
 import json
 import os
 import platform
 import pstats
 import shutil
-import datetime
 from pathlib import Path
 
 import pyderman
 import requests
 from dotenv import load_dotenv
 from pabot.pabot import main as pabot_run_cli
-from robot import run_cli as robot_run_cli
 from robot import rebot_cli as robot_rebot_cli
+from robot import run_cli as robot_run_cli
 import scripts.keyword_profile as kp
-from tests.libs.slack import send_slack_report
 from tests.libs.setup_auth_variables import setup_auth_variables
-
+from tests.libs.slack import send_slack_report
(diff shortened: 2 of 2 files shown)


tests/robot-tests/tests/libs/create_emulator_release_files.py
```diff
@@ -64,7 +64,9 @@ def __init__(self):
 
         # Instantiate a new ContainerClient for the emulator
         self.blob_service_client = BlobServiceClient.from_connection_string(
-            "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://data-storage:10000/devstoreaccount1")
+            "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey"
+            "=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint"
+            "=http://data-storage:10000/devstoreaccount1")
```
ntsim, line 69: Could format this to be a little nicer by pairing they keys up with the values ie:

```py
"DefaultEndpointsProtocol=http;"
"AccountName=devstoreaccount1;"
"AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;"
"BlobEndpoint=http://data-storage:10000/devstoreaccount1"
```

tests/robot-tests/tests/libs/create_emulator_release_files.py
```diff
@@ -178,21 +187,32 @@ def create_private_release_files(self):
...
+                pass
 
 
-if __name__ == '__main__':
-    generator = ReleaseFilesGenerator()
-    generator.create_public_release_files()
-    generator.create_private_release_files()
+generator = ReleaseFilesGenerator()
+generator.create_public_release_files()
+generator.create_private_release_files()
```
ntsim, line 218: Does this mean that just importing the file will run the generator methods? We should probably be executing these in `run_tests.py` instead

tests/robot-tests/tests/libs/create_emulator_release_files.py
```diff
@@ -178,21 +187,32 @@ def create_private_release_files(self):
...
+    @staticmethod
+    def upload_files(container_client, files):
+        data = b'abcd' * 128
         for file in files:
             path = file.path()
-            blob_client = container_client.get_blob_client(path)
-            blob_client.upload_blob(
-                data, blob_type="BlockBlob", metadata=file.metadata())
+            try:
+                blob_client = container_client.get_blob_client(path)
```
benoutram, line 208: Is it the `get_blob_client` or the `upload_blob` that raises the `ResourceExistsError`? Just wondering if `get_blob_client` needs to be within the `try`, if it's only `upload_blob`  that raises it.

tests/robot-tests/tests/libs/create_emulator_release_files.py
```diff
@@ -165,10 +167,17 @@ def create_public_release_files(self):
...
                 "downloads")
 
-        self.upload_files(container_client,
-                          self.all_files_zip_files)
-        self.upload_files(container_client,
-                          self.data_files)
+        try:
+            self.upload_files(container_client,
+                              self.all_files_zip_files)
+        except ResourceExistsError:
```
benoutram, line 173: Are you sure you need to wrap every call to `upload_files` with try - except `ResourceExistsError` which I see in several places here?

I thought that within `upload_files` you are ignoring the `ResourceExistsError` using

```
except ResourceExistsError:
    pass
```

Doesn't this mean it won't be re-raised?

tests/robot-tests/tests/libs/create_emulator_release_files.py
luke-h1: Looks like the storage emulator doesn't like it when we split the connection string up into separate strings: 

```
binascii.Error: Invalid base64-encoded string: number of data characters (85) cannot be 1 more than a multiple of 4
```

ntsim: Can't we just concatenate the strings using `+` then?

tests/robot-tests/tests/libs/create_emulator_release_files.py
benoutram: I'm a bit confused because this comment is resolved but I don't see any changes in the code or a reply here to explain. I think your safe to raise `blob_client = container_client.get_blob_client(path)` up one line in the code above the `try`.

tests/robot-tests/tests/libs/create_emulator_release_files.py
benoutram: I see the change you made but you've moved `except ResourceExistsError:` lower down which means now if the create_container fails because the container already exists it will skip uploading files. The container might exist but be empty in which case we still want to upload files.

tests/robot-tests/tests/libs/create_emulator_release_files.py
luke-h1: sorry about this, I read this PR comment the wrong way

tests/robot-tests/tests/libs/create_emulator_release_files.py
luke-h1: I've reverted the changes & I have readded the `pass ` in the various except blocks as we don't want to do anything when an exception is raised (i.e. a blob container already exists).

tests/robot-tests/tests/libs/create_emulator_release_files.py
```diff
@@ -162,37 +166,60 @@ def create_public_release_files(self):
...
             path = file.path()
-            blob_client = container_client.get_blob_client(path)
-            blob_client.upload_blob(
-                data, blob_type="BlockBlob", metadata=file.metadata())
 
-
-if __name__ == '__main__':
-    generator = ReleaseFilesGenerator()
-    generator.create_public_release_files()
-    generator.create_private_release_files()
```
markhiveit, line 198: I'd keep this - someone might want to execute this script to restore things just so they can use the service without running the UI tests.

tests/robot-tests/tests/libs/create_emulator_release_files.py
```diff
@@ -162,37 +166,60 @@ def create_public_release_files(self):
...
+            blob_client = container_client.get_blob_client(path)
+            try:
+                blob_client.upload_blob(
+                    data, blob_type="BlockBlob", metadata=file.metadata())
+            except ResourceExistsError:
+                pass
```
markhiveit, line 225: I'd print something here - if not the entire error, then just a string to indicate that a file hasn't been uploaded.

tests/robot-tests/tests/libs/create_emulator_release_files.py
```diff
@@ -162,37 +166,60 @@ def create_public_release_files(self):
...
+
+        except ResourceExistsError:
+            pass
+
+        try:
+            self.upload_files(container_client,
+                              self.data_files)
+
+        except ResourceExistsError:
+            pass
```
markhiveit, line 186: There is already a try / except in `upload_files` itself, so I don't think these try / except are necessary.

tests/robot-tests/tests/libs/create_emulator_release_files.py
```diff
@@ -162,37 +166,60 @@ def create_public_release_files(self):
             container_client = self.blob_service_client.create_container(
                 "downloads")
         except ResourceExistsError:
+            pass
+
```
markhiveit, line 170: I think this change can be reverted to what it was originally. It was trying to create the blob service, and if that fails, it was then getting it.

After your changes, it tries to create the blob service, and if it fails, it passes, which means we now don't have the container_client. 

I think maybe you've not realised that both the `pass` and `container_client = self.blob_service_client.get_container_client(` are inside the same `except` scope.

If you did have `container_client = self.blob_service_client.get_container_client(` outside the `except` scope, then you would potentially be getting `container_client` twice, when the original code would only ever get it once.

markhiveit (changes requested):

markhiveit (approved):

PR #3024, opened 2021-12-03
EES-2942 - setup python linting

luke-h1:


luke-h1:
Looks like the frontend build is failing due to a flaky unit test

package.json
```diff
@@ -66,6 +66,9 @@
     ],
     "*.robot": [
       "python3 -m pipenv run robotidy --config tests/robot-tests/robotidy.toml"
+    ],
+    "*.py": [
```
markhiveit, line 70: Do we also want `format:python` and `format:python:check` scripts, similar to what we have to robotidy?

package.json
```diff
@@ -66,6 +66,9 @@
...
+      "python3 -m pipenv run autopep8 --max-line-length 120 -r --in-place tests"
```
markhiveit, line 71: I think this line shouldn't have `tests` at the end - as this runs it against the entire tests directory, rather than the specific python files that have changed.

markhiveit (changes requested):

package.json
ntsim, line 71: Seems like we can use a `pyproject.toml` file to configure things like this instead. It'd also be good to have a script in `package.json` to run this too, so having the `pyproject.toml` would make the configurations consistent for these without needing to change the command in multiple places.

tests/robot-tests/backup-local.py
```diff
@@ -47,10 +47,13 @@
 for blob in generator:
     if '/' in blob.name:
         head, tail = os.path.split(blob.name)
-        os.makedirs(os.path.join(os.getcwd(), 'backup-data', 'content-cache', head), exist_ok=True)
-        block_blob_service.get_blob_to_path('cache', blob.name, os.path.join(os.getcwd(), 'backup-data', 'content-cache', head, tail))
+        os.makedirs(os.path.join(os.getcwd(), 'backup-data',
+                    'content-cache', head), exist_ok=True)
```
ntsim, line 51: Did you run the formatter without the `max-line-length` set to 120? Seems like we have all sorts of formatting changes where the lines were below 120 characters.

I'm finding that if I change lines like this to how it was previously, then running the formatter, no changes are made. I'm guessing most of these changes are actually unnecessary.

Would suggest we re-do the formatting and make sure that the max line length has been set correctly to avoid creating lots of potentially incorrect formatting changes.

ntsim (changes requested):

tests/robot-tests/backup-local.py
ntsim: It looks like this was resolved, but it seems like none of the incorrect formatting changes have been reverted

ntsim (changes requested):
Unnecessary formatting changes need to be reverted still

tests/robot-tests/backup-local.py
luke-h1: I think this should be sorted now

ntsim (approved):

luke-h1:
> Unnecessary formatting changes need to be reverted still

This has been addressed now. It seems autopep8 isn't deterministic and wasn't picking up the new `pyproject.toml` configuration after the first formatting pass

PR #3006, opened 2021-11-22
EES-2881 Return hierarchical Locations in table result Subject meta data

benoutram:
This PR builds on the work of https://github.com/dfe-analytical-services/explore-education-statistics/pull/2997 to apply location hierarchies to _table result_ subject meta data.

A new response format is defined for the locations field in the table result subject meta data response (compare examples below of responses with the feature toggle turned off and on).

This format is similar but slightly different to the new format implemented for subject meta data in https://github.com/dfe-analytical-services/explore-education-statistics/pull/2997.

## Outstanding work to maintain backwards compatibility with Permalinks

Work is still outstanding on how to maintain backwards compatibility for Locations with the hierarchical response type
  * Backwards compatibility is required with existing Permalinks unless we migrate them.
  * Backwards compatibility is also required with the current UI expecting locations in this form until it's updated in EES-2777.

This has been split out to EES-2917 and for now hierarchical locations are populated in a separate field 'LocationsHierarchical' when the feature is turned on.

## Turning on/off and configuring location hierarchies

This change is only visible when the location hierarchies feature is turned on. The feature toggle is intended to stay in place until the UI work for Location hierarchies is completed by EES-2773 and EES-2777.

See the instructions in https://github.com/dfe-analytical-services/explore-education-statistics/pull/2997 to turn this feature on and configure location hierarchies as this has not changed in this PR.

## Affected API calls

Admin App

* `POST /api/data/tablebuilder/release/{releaseId}`

Data.Api App

* `POST /api/tablebuilder`
* `POST /api/tablebuilder/release/{releaseId}`
* `GET /api/tablebuilder/release/{releaseId}/data-block/{dataBlockId}`
* `GET /api/fasttrack/{dataBlockId}`
* `GET /api/permalink/{permalinkId}`

## Example query

No changes to the way that tables are queried.

Here's an example table query POST body for local authorities Hackney and Lambeth. This query is used in the examples below except where one where it's altered to also include a region.

```
{
  "subjectId": "7bc6f12a-37c3-4106-1088-08d9ae228ece",
  "includeGeoJson": true,
  "locations": {
    "localAuthority": ["E09000012", "E09000022"]
  },
  "timePeriod": {
    "startYear": 2013,
    "startCode": "AY",
    "endYear": 2017,
    "endCode": "AY"
  },
  "filters": [],
  "indicators": []
}
```

## Behaviour with feature toggle OFF

[...]

PR #3025, opened 2021-12-03
EES-2917 Maintain backwards compatibility with Permalinks for changes to Table subject meta Locations

benoutram:
## TL;DR?  🏃 🏃‍♀️ 

- Adds a conversion during deserialization of the ‘flat’ Locations in ResultSubjectMetaViewModel of old Permalinks to the new LocationsHierarchical field.

- FE client now only has to be aware of `LocationsHierarchical`. The conversion maintains backwards compatibility with existing Permalinks even if the Locations field is dropped from the model which can happen as soon as the hierarchical locations feature is complete and made permanent.

- In future we could apply the transformation permanently by a migration. The field could also be renamed back to Locations. (EES-2943)

## Detailed version 🐌 

This PR builds upon https://github.com/dfe-analytical-services/explore-education-statistics/pull/3006 which added a new `LocationsHierarchical` field into table results subject meta alongside the 'flat' `Locations` field.

`LocationsHierarchical` is populated instead of `Locations` when the feature toggle is turned on. This is planned to remain the case while development is still in progress (see EES-2777) after which this will become the permanent behaviour.

The PR aims to stop the front end client needing to be aware that there are now two different location fields in Permalink data by only needing to access the new `LocationsHierarchical` field, and allows for the `Locations` field to be dropped from the model shortly when the feature to add hierarchical locations in Table Result Subject Metadata becomes permanent.

It adds backwards compatibility with existing Permalinks by using a new `ResultSubjectMetaViewModelJsonConverter` to transform `Locations` -> `LocationsHierarchical`.

It should be safe to drop the `Locations` field after the feature becomes permanent as the transformation has been developed to access `Locations` by JSONPath during deserialization so that nothing depends directly on the old field. Until it becomes permanent, the converter is only applied to serializer settings when the feature toggle is turned on. With the feature toggle turned off, `Locations` are populated as before.

Transformation doesn't infer any hierarchy so the location attributes will continue to be single level meaning there's no overall change to the data of Permalinks except for moving it between fields and alteration of the format.

Use of the converter class is intended to be temporary until the transformation is made permanent by migrating old Permalinks in blob storage, at which point the field could also be renamed back to Locations (EES-2943).

### Other changes

- Add unit tests for PermalinkService.
- Add a new `PermalinkContractResolver` to serializer settings when creating Permalinks which controls conditionally serializing the `Locations` and `LocationsHierarchical` fields dependent on the state of the feature toggle.
- Remove FeatureManagement dependency in favour of simpler `LocationOptions` property to control the table hierarchy feature toggle. The FeatureManagement dependency was added partly due to thinking that it would make it easier to turn the feature on/off in Azure Environments for testing, which it would, but it needs to be used in conjunction with Azure App Configuration which we don't use yet. See https://docs.microsoft.com/en-us/azure/azure-app-configuration/overview for more info.
- Remove the majority of unit tests from `PermalinkControllerTests` as they don't seem very useful.
- Remove the unused base PermalinkMigrationService from Data.Api. This was intended to be wrapped by another service to migrate Permalinks and has been used successfully in the past. However due to the way it works by reading all Permalinks into memory, transforming them and then uploading them, it's thought that this will be very likely to fail now that there are a large number of Permalinks for some very big table results. If we do permanently migrate Permalinks by transforming `Locations` -> `LocationsHierarchical` in future then we can restore some of this code with additional reliability, potentially by using blob index tags to filter permalinks which are candidates to be migrated. We wouldn't run a Permalink migration from an endpoint in Data.Api in future anyway.
- Refactor `LocationsOptions` out of `SubjectMetaService.cs` as it's also shared by `ResultSubjectMetaService`.

image


src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Services/PermalinkServiceTests.cs
```diff
@@ -0,0 +1,658 @@
...
+            Assert.Equal(legacyLocations[1].Label, localAuthorities[1].Label);
+            Assert.Equal(legacyLocations[1].Value, localAuthorities[1].Value);
+            Assert.Equal(legacyLocations[1].GeoJson, localAuthorities[1].GeoJson);
+
+            Assert.Equal(legacyLocations[2].Label, localAuthorities[2].Label);
+            Assert.Equal(legacyLocations[2].Value, localAuthorities[2].Value);
+            Assert.Equal(legacyLocations[2].GeoJson, localAuthorities[2].GeoJson);
+        }
+
+        [Fact]
```
duncan-at-hiveit, line 402: I guess we're never planning on running with the flag enabled for a time and then switching back to it being disabled in an emergency?  Otherwise we'd need a converter to also transform `LocationsHierarchical` back to `Locations` as well.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Controllers/PermalinkController.cs
```diff
@@ -7,7 +8,7 @@
 
 namespace GovUk.Education.ExploreEducationStatistics.Data.Api.Controllers
 {
-    [Route("api/[controller]")]
+    [Route("api/permalink")]
```
duncan-at-hiveit, line 11: Why this change out of interest?  Are we generally switching to this convention now?

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
```diff
@@ -96,4 +114,42 @@ private async Task<PermalinkViewModel> BuildViewModel(Permalink permalink)
...
+
+        protected override JsonProperty CreateProperty(MemberInfo member, MemberSerialization memberSerialization)
+        {
+            JsonProperty property = base.CreateProperty(member, memberSerialization);
+            if (property.DeclaringType == typeof(Permalink))
+            {
+                property.ShouldSerialize = property.PropertyName switch
+                {
+                    "Locations" => _ => !_tableResultLocationHierarchiesEnabled,
+                    "HierarchicalLocations" => _ => _tableResultLocationHierarchiesEnabled,
```
duncan-at-hiveit, line 135: Should this not be `LocationsHierarchical`?

src/GovUk.Education.ExploreEducationStatistics.Data.Services/Converters/ResultSubjectMetaViewModelJsonConverter.cs
```diff
@@ -0,0 +1,89 @@
...
+    /// have locations in field <see cref="ResultSubjectMetaViewModel.LocationsHierarchical">ResultSubjectMetaViewModel.LocationsHierarchical</see>.
+    ///
+    /// Until old Permalinks are migrated, the translation provided by this converter ensures that consumers
+    /// are aware of legacy locations when accessing <see cref="ResultSubjectMetaViewModel.LocationsHierarchical" />.
+    /// </summary>
+    public class ResultSubjectMetaViewModelJsonConverter : JsonConverter<ResultSubjectMetaViewModel>
+    {
+        public override bool CanWrite => false;
+
+        public override ResultSubjectMetaViewModel ReadJson(JsonReader reader,
```
duncan-at-hiveit, line 33: Why the need for this class AND the ContractResolver out of interest?

duncan-at-hiveit: Oh I see, the ContractResolver uses this if it decides that it needs to?

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
benoutram: Good spot! Corrected now, thanks 🙂 . So far I only did some testing with the feature turned off, I will do some testing with it turned on now.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Controllers/PermalinkController.cs
benoutram: I felt like it's safer to define the route string explicitly as the path mustn't change unless there's corresponding changes to the FE.

I think that the replacement of the `[controller]` token for the class name into the route is nice but is vulnerable to the class names changing during a refactor.

After this change we are down to 5 controllers  still using the token.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/Converters/ResultSubjectMetaViewModelJsonConverter.cs
benoutram: Yes, the ContractResolver is aware of the feature toggle.

It stops an empty Locations field getting serialized when turned on.
It stops an empty LocationsHierarchical field getting serialized when turned off.

It also applies the JsonConverter when turned on so that old Permalinks have their Locations deserialised as LocationsHierarchical.

I think once the feature toggle is removed from the codebase and it becomes permanent the ContractResolver can be removed and we will always want to apply the converter using an annotation on `TableResultViewModel` instead:

```
    [JsonConverter(typeof(ResultSubjectMetaViewModelJsonConverter))]
    public ResultSubjectMetaViewModel SubjectMeta { get; set; }
```

If we eventually migrate all permalinks then the JsonConverter can also be dropped.

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Services/PermalinkServiceTests.cs
benoutram: Good point thanks, but not really planning for that to happen.

We only plan to turn the feature on for local development while the FE ticket EES-2777 is in progress. 

The toggle exists to prevent a long running code branch while these changes are in progress and not designed to be toggled in real environments as the BE change it brings can't  work independently without the FE change EES-2777.

I don't think we plan to keep the toggle either once the FE changes are complete so we should never have a situation in a real environment where `LocationsHierarchical` need transforming back to `Locations`.

duncan-at-hiveit: Yup no worries - I figured so, thanks!

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Controllers/PermalinkController.cs
duncan-at-hiveit: Yeah I much prefer this way tbh.  If for no other reason, it's quick at a glance to see all code that refers to that route, both in the front end and back.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/Converters/ResultSubjectMetaViewModelJsonConverter.cs
duncan-at-hiveit: Great!  Thanks for the detailed reply btw - always appreciated :)

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
duncan-at-hiveit: Nice one.

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Converters/ResultSubjectMetaViewModelJsonConverterTests.cs
```diff
@@ -0,0 +1,125 @@
...
+
+namespace GovUk.Education.ExploreEducationStatistics.Data.Api.Tests.Converters
+{
+    public class ResultSubjectMetaViewModelJsonConverterTests
+    {
+        [Fact]
+        public void ReadJson_LegacyLocationsFieldIsTransformed()
+        {
+            var jsonText =
+                @"{""Locations"":[{""Level"":""localAuthority"",""GeoJson"":[{""properties"":{""code"":""E06000009""}}],""Label"":""Blackpool"",""Value"":""E06000009""},{""Level"":""localAuthority"",""GeoJson"":[{""properties"":{""code"":""E06000015""}}],""Label"":""Derby"",""Value"":""E06000015""},{""Level"":""localAuthority"",""GeoJson"":[{""properties"":{""code"":""E06000018""}}],""Label"":""Nottingham"",""Value"":""E06000018""}]}";
```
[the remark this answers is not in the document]
benoutram: Thanks, good suggestion, I've done this now.

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Converters/ResultSubjectMetaViewModelJsonConverterTests.cs
```diff
@@ -0,0 +1,125 @@
...
+            Assert.Equal("Nottingham", localAuthorities[2].Label);
+            Assert.Equal("E06000018", localAuthorities[2].Value);
+            Assert.NotNull(localAuthorities[2].GeoJson);
+        }
+
+        [Fact]
+        public void ReadJson_LocationsHierarchicalIsUntouched()
+        {
+            const string jsonText =
+                @"{""LocationsHierarchical"":{""country"":[{""GeoJson"":[{""properties"":{""code"":""E92000001""}}],""Label"":""England"",""Value"":""E92000001""}]}}";
```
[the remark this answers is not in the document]
benoutram: Thanks, I've done both now.

ntsim (approved):

PR #3026, opened 2021-12-06
EES-2922 Lazy load data blocks upon opening of their content section

ntsim:
This PR adds further lazy-loading of the data blocks so that they will not load until their relevant section in the release content has been opened. 

Previously, most (if not all) of the data block requests would be triggered as soon as the user scrolled over the release's main content accordion. This could lead to large spikes in traffic if none of the data blocks were cached for that release.

To allow us to do this enhanced lazy loading, we now use wrapping `Gate` components that prevent their child data block from loading until the accordion section has been opened at least once.

## Relevant changes

- Refactors `Gate` to use a `useEffect` instead of `useAsyncCallback`. We needed more fine-grained control of how the states toggle, particularly as we have introduced a new `passOnce` prop that controls if the `Gate` will allow its child to unmount if its condition reverts back to false later.

### DIFF
--- a/src/explore-education-statistics-common/src/components/Gate.tsx
+++ b/src/explore-education-statistics-common/src/components/Gate.tsx
@@ -1,61 +1,91 @@
-import useAsyncCallback, {
-  AsyncStateSetterParam,
-} from '@common/hooks/useAsyncCallback';
+import usePrevious from '@common/hooks/usePrevious';
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-import React, { ReactNode, useEffect } from 'react';
+import React, { ReactNode, useEffect, useState } from 'react';
 
 interface GateProps {
-  condition: boolean | (() => Promise<boolean>);
+  condition: boolean | (() => Promise<boolean>) | (() => boolean);
   children: ReactNode;
+  /**
+   * Fallback to render if the {@property condition}
+   * has not passed yet.
+   */
   fallback?: ReactNode | ((error?: unknown) => ReactNode);
+  /**
+   * Loading state to render if the {@property condition}
+   * is dependent on an asynchronous task.
+   */
   loading?: ReactNode;
+  /**
+   * If the {@property condition} passes once, then we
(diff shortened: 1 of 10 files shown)


src/explore-education-statistics-common/src/components/Gate.tsx
```diff
@@ -1,61 +1,91 @@
...
-    async () => (typeof condition === 'boolean' ? condition : condition()),
-    [condition],
-    () => {
-      let initialState: AsyncStateSetterParam<boolean> = {
-        isLoading: true,
-      };
+  const previousCondition = usePrevious(condition);
+
+  const [passed, setPassed] = useState(false);
+  const [isLoading, setLoading] = useState(false);
```
luke-h1, line 42: Just wondering if this should be named `setIsLoading` to follow the naming convention here?

ntsim: I typically use the `isX` convention just to convey that the variable/state is a boolean. Personally, I think doing this convention with the setter just makes the function sound weird, and it doesn't add a whole lot. We should just be focusing on the name's subject (the X, or 'loading' in this example).

Can change it if you feel particularly strongly about it, but it's not something I'd get others to change personally.

luke-h1: Ah, that's good to know. Yeah, I don't have a strong opinion about it (`setLoading` sounds better imo), was mainly wondering if we followed that type of naming convention elsewhere.

markhiveit (approved):

src/explore-education-statistics-common/src/components/Gate.tsx
```diff
@@ -1,61 +1,91 @@
...
+          setError(err);
+        } finally {
+          setLoading(false);
+        }
+      } else {
+        setPassed(task);
+      }
+    };
+
+    checkCondition();
```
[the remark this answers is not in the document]
ntsim: Yep - it's purely just so we can use async/await :smile:

PR #3028, opened 2021-12-06
Ees 2920 caching data block results in admin

duncan-at-hiveit:
This PR:
- Adds caching to Data Block results within Admin so that when requested from e.g. the Release Content page, they will cache and subsequent retrievals will be very fast.
- Adds cache invalidation via the BlobCacheService when a Data Block is updated or deleted.
- Adds return values to various methods called by `DataBlockService.DeleteDataBlock()` so that we can check for success or failure of these steps in the deletion process.  Previously, they were just Tasks, and failures within them could be swallowed.  
- Adds numerous unit tests to anywhere where we currently use the caching annotations, to ensure the caching is implemented correctly and that the entities that they cache can be successfully serialised and deserialised from JSON (as we have been stung twice by this due to missing setters / changing from concrete classes to interfaces).
- Introduces a new testing dependency allowing us to assert deep equality of objects, useful for testing that entities can be serialised and deserialised from JSON successfully without needing a million Equals and HashCode methods cluttering up the ViewModels (and especially around things like Dictionary which are tricky to assert equality off OOTB).

This PR also adds a new test fixture that sets up a BlobCacheService mock for convenience.  It also prevents the tests that use it from running in parallel, which can otherwise interfere with each other due to the CachingAspect.Enabled flag being static and the need for a single static Mock BlobCacheService being available for use.  Note that any other tests that would otherwise be affected by the CachingAspect.Enabled flag being changed also have to use this same [Collection(BlobCachingServiceTests)] annotation in order to avoid having their behaviour be unpredictable.  Oh so many unit tests...

In the future, it might be preferable to just bite the bullet and enable the CachingAspect.Enabled flag by default, and just put in the expectations for the caching behaviour into the tests that need it rather than switching this static flag on and off.  Given that these tests which want to ignore the caching already have to explicitly belong to the same collection as the tests which DO want to test the caching, this suggests that maybe we just keep it enabled by default.  Thoughts welcome!

This PR does not currently:
- Cache and retrieve the cached versions of Data Blocks when editing a Data Block from the Release "Data Blocks" tab.  This is because the TableBuilder component on this page is actually issuing a Table Builder query as a POST with no notion of it being a Data Block at that point.  It's not immediately clear how tricky it would be to re-engineer this to work more like public Fast Tracks whereby the Data Block Id is passed rather than a full-blown table query.

Probably best start by looking at the actual production code changes first, then skip to the tests for those specific pieces.

image


src/GovUk.Education.ExploreEducationStatistics.Common/BlobContainers.cs
```diff
@@ -10,6 +10,7 @@ public static class BlobContainers
     {
         public static readonly IBlobContainer PrivateReleaseFiles = new BlobContainer("releases");
         public static readonly IBlobContainer PublicReleaseFiles = new BlobContainer("downloads");
+        public static readonly IBlobContainer PrivateContent = new PrivateBlobContainer("cache");
         public static readonly IBlobContainer PublicContent = new BlobContainer("cache");
```
benoutram, line 14: I don't think it will make a difference but we could make `PublicContent` a type of `PublicBlobContainer` so it gets the name `public-cache` locally now that we have `private-cache`, similar to how we have `public-methodologies` and `private-methodologies`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Cache/CacheKeyService.cs
```diff
@@ -0,0 +1,35 @@
...
+        private readonly IPersistenceHelper<ContentDbContext> _contentPersistenceHelper;
+
+        public CacheKeyService(IPersistenceHelper<ContentDbContext> contentPersistenceHelper)
+        {
+            _contentPersistenceHelper = contentPersistenceHelper;
+        }
+        
+        public async Task<Either<ActionResult, DataBlockTableResultCacheKey>> CreateCacheKeyForDataBlock(Guid releaseId, Guid dataBlockId)
+        {
+            return await _contentPersistenceHelper
```
benoutram, line 26: Just wondering if you considered caching them by just data block id and not caring about the publication id or the release id to partition the cache files by publication / release? If so then I don't think this cache key service is required.

I think data blocks are cloned when creating amendments so we should be guaranteed to be caching / invalidating the right data with just a data block id I think.

I suppose partitioning them into folders by release does make identifying them easier and also makes it easier to clean them up if a release is deleted.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/Cache/ICacheKeyService.cs
```diff
@@ -0,0 +1,14 @@
+using System;
```
benoutram, line 1: I think we are generally adding `nullable enable` to all new classes and wherever else we can.

markhiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Statistics/TableBuilderControllerTests.cs
```diff
@@ -221,126 +292,94 @@ public async Task QueryForDataBlock_NoGeoJson()
...
+        public void TableBuilderResultViewModel_SerialiseAndDeserialise()
+        {
+            var converted = DeserializeObject<TableBuilderResultViewModel>(SerializeObject(_tableBuilderResults));
+            converted.AssertDeepEquals(_tableBuilderResults);
+        }
 
-            Assert.IsType<NotFoundResult>(result.Result);
+        private static DataBlockTableResultCacheKey IsMatchingDataBlockCacheKey(ReleaseContentBlock releaseContentBlock)
+        {
+            return It.Is<DataBlockTableResultCacheKey>(cacheKey => cacheKey.AssertDeepEquals(new DataBlockTableResultCacheKey(releaseContentBlock)));
```
benoutram, line 360: Just wondering if there's a disadvantage to implementing equality method on DataBlockTableResultCacheKey and using `Equals` instead? As there's no complex types equals should be sufficient I think.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ThemeService.cs
```diff
@@ -25,7 +23,6 @@ public ThemeService(ContentDbContext contentDbContext)
             _contentDbContext = contentDbContext;
         }
 
-        [BlobCache(typeof(PublicationTreeCacheKey))]
         public async Task<IList<ThemeTree<PublicationTreeNode>>> GetPublicationTree(
```
benoutram, line 26: Should we add something to theme service to delete private cached data blocks when deleting a theme/topic during teardown (could create new ticket for that)?

src/GovUk.Education.ExploreEducationStatistics.Common/BlobContainers.cs
duncan-at-hiveit: I think this would be good - just conscious that time is ticking on this expedite, and this would require a bit of manual migration as part of rollout.  I could look at doing this as part of the lower priority caching of Subject Meta in Admin, which is a follow-up ticket to this?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Cache/CacheKeyService.cs
duncan-at-hiveit: I figured that it might be best to stick with the general convention of the publication / release paths for now until we decided one way or the other whether to move them wholesale to just Guids.  It does make deleting a whole Amendment's-worth of cached content a little easier during cancellation I guess though (see latest commit!).

benoutram (commented):
After data replacement I think we might need to invalidate the cached data blocks? I think we keep the same data blocks but switch them to using attributes of the new subject?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/Cache/ICacheKeyService.cs
duncan-at-hiveit: Done, thanks!

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Controllers/FastTrackControllerTests.cs
```diff
@@ -1,86 +1,316 @@
...
-            var (controller, mocks) = BuildControllerAndMocks();
+            var (controller, _) = BuildControllerAndMocks();
             var result = await controller.Get("InvalidGuid");
             result.AssertNotFoundResult();
         }
-        
-        private (
+
+        [Fact]
+        public void FastTrackViewModel_SerialiseAndDeserialise()
```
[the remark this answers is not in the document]
duncan-at-hiveit: The reason behind this is that we've had 2 issues whereby caching has broken but we've not been aware of it until it was revealed with manual testing / checking the log messages.

One time was due to something having an `init` rather than a `set` setter on a property, meaning that it could be written wholly to blob storage, but that particular field could not be retrieved as it had to setter.

The second was due to it not knowing how to deserialise the Location hierarchy / non-hierarchy property of a view model because it only had an `ISubjectMetViewModel` interface to work with when deserialising.

I'd hope that these would help us with such cases in the future.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataBlockServicePermissionTests.cs
```diff
@@ -122,17 +124,21 @@ private Mock<IPersistenceHelper<ContentDbContext>> PersistenceHelperMock()
             IPersistenceHelper<ContentDbContext> persistenceHelper = null,
             IReleaseFileService releaseFileService = null,
             IReleaseContentBlockRepository releaseContentBlockRepository = null,
-            IUserService userService = null)
+            IUserService userService = null,
```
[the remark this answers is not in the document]
duncan-at-hiveit: Done, thanks!

src/GovUk.Education.ExploreEducationStatistics.Common/BlobContainers.cs
benoutram: I was just thinking it should only affect local development using the storage emulator as non-emulated environments will use "cache" for both public and private which exist in totally separate storage accounts.

With the current setup, locally emulated containers will both be in the same storage account (there's only one anyway) and will be named "private-cache" and "cache" but could be "private-cache" and "public-cache" just for consistency.

Fine to do it later, or even just test what I say here later is correct 🙂 .

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Statistics/TableBuilderController.cs
```diff
@@ -66,31 +68,37 @@ public class TableBuilderController : ControllerBase
             Guid dataBlockId,
             CancellationToken cancellationToken = default)
         {
-            return await _contentPersistenceHelper.CheckEntityExists<ReleaseContentBlock>(
+            return await _contentPersistenceHelper
+                .CheckEntityExists<ReleaseContentBlock>(
                     query => query
                         .Include(rcb => rcb.ContentBlock)
                         .Include(rcb => rcb.Release)
+                        .ThenInclude(release => release.Publication)
```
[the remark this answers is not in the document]
duncan-at-hiveit: Hey, this is included in order to build the cache key that the caching annotation needs, in order to convert from the ReleaseContentBlock to a DataBlockTableResultsCacheKey.

I did want to reuse `cacheKeyService` here, but the actual Release is still needed for the permission check, so it seemed like it was a bit of duplicated effort to firstly grab the ReleaseContentBlock in order to do the permission check, and then to get the ReleaseContentBlock *again* to build the cache key.

It's definitely not the end of the world to change it to use the cacheKeyService, but feels a bit unnecessary at the moment.  Not 100% happy either way though :)

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Cache/CacheKeyService.cs
benoutram: Cool, thought that might be the case, thanks. See my comment about tearing down data which it probably also helps with.

src/GovUk.Education.ExploreEducationStatistics.Common/ViewModels/GlossaryViewModels.cs
```diff
@@ -1,5 +1,7 @@
 ﻿#nullable enable
+using System;
 using System.Collections.Generic;
+using System.Linq;
```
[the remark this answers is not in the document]
duncan-at-hiveit: Done, thanks!

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Controllers/FastTrackControllerTests.cs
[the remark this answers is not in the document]
benoutram: Good to know thanks. This could be relevant to the way I'm testing some changes around Permalinks. 🙂

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Statistics/TableBuilderControllerTests.cs
benoutram: That sounds good to me, thanks. I was also going to say that with c#9 Record types (when we can use them in the Admin!), I think this particular example would be equal with the default value equality that Record types offer.

src/GovUk.Education.ExploreEducationStatistics.Common/BlobContainers.cs
duncan-at-hiveit: Oh I see, no that sounds like a good idea then!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReplacementService.cs
```diff
@@ -112,6 +121,8 @@ public class ReplacementService : IReplacementService
                         return ValidationActionResult(ReplacementMustBeValid);
                     }
 
+                    await replacementPlan.DataBlocks.ForEachAsync(plan =>
```
benoutram, line 124: It will be good to move away from our home baked extensions to the `System.Linq.Async` extensions but appreciate this is using `ForEachAsync` for consistency without changing everything!

benoutram, line 124: It will be good to move away from our home baked extensions to the System.Linq.Async extensions but appreciate this is using ForEachAsync for consistency without changing everything!

benoutram (dismissed):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Statistics/TableBuilderControllerTests.cs
duncan-at-hiveit: Nice one - yeah a shame we can't use them in admin yet :cry:

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ThemeService.cs
duncan-at-hiveit: Done now thanks

src/GovUk.Education.ExploreEducationStatistics.Common/BlobContainers.cs
duncan-at-hiveit: Done now thanks.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Statistics/TableBuilderController.cs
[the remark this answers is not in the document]
markhiveit: Happy for you to leave it - as you say its minor - didn't realise it was being used by `cacheKeyService`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
```diff
@@ -31,20 +34,26 @@ public class DataBlockService : IDataBlockService
...
+        private readonly ICacheKeyService _cacheKeyService;
+        
         public DataBlockService(ContentDbContext context,
             IPersistenceHelper<ContentDbContext> persistenceHelper,
             IReleaseFileService releaseFileService,
             IReleaseContentBlockRepository releaseContentBlockRepository,
             IUserService userService,
-            IMapper mapper)
+            IMapper mapper, 
+            IBlobCacheService cacheService,
```
benoutram, line 46: I think this might need wiring up with a dedicated `BlobStorageService` that has a connection key string for `CoreStorage`, and the same in other services that are invalidating private cache items using `BlobStorageService` like `TopicService`. I'm not 100% sure though as I can't see where it's currently getting candidates of IBlobCacheService to autowire in `Startup` at all right now.

benoutram: > 

Sorry, just seen that you added this in your PR

`services.AddTransient<IBlobCacheService, BlobCacheService>();`

and combined with this that already existed explains how it's working

`services.AddTransient(provider => GetBlobStorageService(provider, "CoreStorage"));`

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReplacementService.cs
duncan-at-hiveit: Yeah, happy to move wholesale to the Linq ones tbh.  I'll bear that in mind for future work.

PR #3046, opened 2021-12-09
EES-2890 - increase statistics DB max size

luke-h1:


infrastructure/parameters/prod.parameters.json
```diff
@@ -70,6 +70,9 @@
     },
     "tableBuilderMaxTableCellsAllowed": {
       "value": 1000000
+    },
+    "maxDbSizeBytes": {
+      "value": 350000000000
```
benoutram, line 75: In `dev.parameters.json` we've got `322122547200`  which will be 300Gb.

In the default for this parameter in `template.json` we've got `268435456000` which will be 250Gb.

Based on that I think we should set this to `375809638400` for 350Gb rather than `350000000000`.

benoutram (approved):

PR #3047, opened 2021-12-09
EES-2957 - increase tableBuilderMaxCellsAllowed for pre-prod

luke-h1:


benoutram (approved):

benoutram:
@luke-h1 before merging this I think you need to be aware that this will cause the UI test around the max table cells to fail if it were to run against the Pre-Prod environment.

PR #3048, opened 2021-12-09
EES-2401 Remove unused table types from statistics database

benoutram:
This PR removes unused table types from statistics database.

* PublicationType
* ThemeType
* TopicType

It also remove unused 'current' versions of stored procedures that have already been dropped.

Here's the [related change that removed the stored procedures](https://github.com/dfe-analytical-services/explore-education-statistics/blob/45d445d086fb8fa1b6f142ae611896afcc63e881/src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20210714123342_EES2401RemoveThemeTopicPublicationTables.cs ) which used these types.


markhiveit:
There are references to UpsertTheme / UpsertTopic / UpsertPublication in `useful-scripts/drop-all-statistics-db-objects.sql`. I'm not sure if that file is being used anymore - so not sure if we can just delete it or it needs updating.

benoutram:
> There are references to UpsertTheme / UpsertTopic / UpsertPublication / PublicationType / TopicType / ThemeType in `useful-scripts/drop-all-statistics-db-objects.sql`. I'm not sure if that file is being used anymore - so not sure if we can just delete it or it needs updating. I imagine if we decide to update it, there will be other things missing from the script, new tables etc.
> 
> Could keep the script and update it when we next need it?

Thanks, I already saw that but figured it's so out of date that there will be plenty of other things not right about it and not worth me changing it. I can remove them though, will just push another change.

markhiveit (approved):

PR #3052, opened 2021-12-10
EES-2911 Strip comments from amendments and published content

ntsim:
This PR adds filtering to remove any comment markup from release content during amendment and publishing.

Comments are typically saved by CKEditor in the following form:

```html
<p>
  <comment-start name="comment-id"></comment-start>
  Something something
  <comment-end name="comment-id"></comment-end>
</p>
```

There are also other variations with other marker elements such as `<commentplaceholder>` and `<resolvedcomment>` that we also need to strip out.

Whilst this format is unlikely to change, we use a fairly aggressive regex to strip out any potential variation of this that would be parseable HTML.

Additionally, we have also updated the fast track replacement to also use this regex based approach too.

## Performance

According to [various benchmarks on StackOverflow](https://stackoverflow.com/a/62919963), it seems like using a compiled regex based approach is pretty much as fast as we can make this kind of string replacement without having to write some custom replacer (requiring use of low-level primitives such as pointers). 

I'm currently hoping that we should see some speedup when creating amendments that have lots of content. We should revisit this if it's a major issue though.

## Other changes

- Refactored amendment logic into its own `ReleaseAmendmentExtensions` to avoid cluttering the `Release` model.
- Added various extension methods. Notably, `String.TrimIndent` makes it easier to write test strings where we want to have nice indentation, but also avoid failing comparisons due to incorrect amounts of whitespace.


src/GovUk.Education.ExploreEducationStatistics.Content.Model/Extensions/ReleaseAmendmentExtensions.cs
```diff
@@ -0,0 +1,144 @@
...
+{
+    public static class ReleaseAmendmentExtensions
+    {
+        private static readonly string FilterPattern = string.Join(
+            '|',
+            // Strips out anything that looks like a comment
+            @"<\s*comment-(start|end)[^>]*(>[^>]*<\/\s*comment-(start|end)[^>]*>|\/>)"
+        );
+
+        public static Release CreateAmendment(this Release release, DateTime createdDate, Guid createdByUserId)
```
benoutram, line 18: It's great to see this separated out to simplify `Release` 🙂 .

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Extensions/ReleaseAmendmentExtensions.cs
```diff
@@ -0,0 +1,144 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Model.Extensions
+{
+    public static class ReleaseAmendmentExtensions
+    {
+        private static readonly string FilterPattern = string.Join(
```
benoutram, line 12: I thought this was a bit strange and was wondering why we don't just call it `CommentFilterPattern` or have a comment further down like `// Strip comments` in `UpdateAmendmentContent`.

I see now with the `string.Join` how you have made this generic, expecting to filter more elements of the content in future. Quite nice I guess, if that happens!

src/GovUk.Education.ExploreEducationStatistics.Publisher.Model/ViewModels/CachedReleaseViewModel.cs
```diff
@@ -1,96 +1,54 @@
...
-            ContentSectionViewModel summarySection,
-            ContentSectionViewModel headlinesSection,
-            ContentSectionViewModel keyStatisticsSection,
-            ContentSectionViewModel keyStatisticsSecondarySection,
-            List<FileInfo> downloadFiles,
-            string dataGuidance,
-            string preReleaseAccessList,
-            List<LinkViewModel> relatedInformation,
-            DateTime? dataLastPublished)
+        public CachedReleaseViewModel(Guid id)
```
benoutram, line 10: Could there be any risk here of things that are now always initialised by default that were previously null changing the behaviour at all for any of these fields? The changes do all look ok to me and I suppose if the tests for newly published Releases are ok then we should be fine for older Releases when we come to regenerate all the cached content next time. Just conscious that a complete refresh might occur much later down the line.

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Extensions/ReleaseAmendmentExtensionsTests.cs
```diff
@@ -0,0 +1,1155 @@
...
+                    <p>Content block 4 http://localhost/fast-track/{amendmentDataBlock1.Id} http://localhost/fast-track/{amendmentDataBlock2.Id}</p>
+                    <p><a href=""http://localhost/fast-track/{amendmentDataBlock1.Id}"">link 1 text</a></p>
+                    <p><a href=""http://localhost/fast-track/{amendmentDataBlock2.Id}/"">link 2 text</a></p>
+                    ",
+                amendmentContentBlock4.Body
+            );
+        }
+
+        [Fact]
+        public void CreateAmendment_StripsComments()
```
benoutram, line 661: I see we end up doing similar testing here between creating release amendments and publishing release content. I don't mean in terms of the content filtering functionality being called, but in terms of checking that comments in different formats are being removed correctly. I suppose the solution would be to abstract out something responsible for stripping comments that can be tested on its own and then just check it's used in both places.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Extensions/ReleaseAmendmentExtensions.cs
```diff
@@ -0,0 +1,144 @@
...
+                        }
+                    }
+                );
+
+            var regex = new Regex(
+                string.Join('|', replacements.Keys.Append(FilterPattern)),
+                RegexOptions.Compiled
+            );
+
+            foreach (var contentBlock in context.ContentBlocks.Values)
```
benoutram, line 85: Will this have all the `HtmlBlocks` in it from the generic content sections? Does it have content blocks from the key stats, headlines, and summary sections to strip comments from them too?

I might be getting confused but just want to check as I think `Release.ContentBlock` only contains data blocks so wasn't sure if `context.ContentBlock` is the same.

Without running this and debugging it I'm not sure 🙁 .

ntsim: Yep it does. During the clone, we pass around the `CloneContext` which gets all of the content blocks appended to it. As mentioned in an earlier comment, it's a bit 'cheeky' to re-use it as it's intended for another purpose during the clone, however, it's definitely the easiest way to get all the right content block references that we need.

src/GovUk.Education.ExploreEducationStatistics.Publisher.Model/ViewModels/CachedReleaseViewModel.cs
ntsim: The primary fields like `Title`, `YearTitle`, `CoverageTitle`, `ReleaseName` and `Slug` should all be set when a release is made, so declaring their defaults as empty string is mostly just to satisfying the nullable constraints.

For more 'optional' fields like `DataGuidance` and `PreReleaseAccessList`, we actually define these as non-optional strings in the frontend and already account for the possibility that they are non-null, but empty.

I've swapped the `ContentSectionViewModel` fields (e.g. summary, headlines, etc) back to non-nullable as it turns out we actually create all of these when a release is created. I'd mistakenly assumed these were optional because admin users don't technically need to create any blocks in any of these sections.

benoutram: Cool, thanks for checking!

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Extensions/ReleaseAmendmentExtensions.cs
benoutram: Great, thanks for the explanation. I'm going to approve now.

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Extensions/ReleaseAmendmentExtensionsTests.cs
ntsim: Good point! I've refactored the comments regex into some utils and now just test that more extensively in separate tests.

The tests for the publishing content/amendment now just do some basic checks around this too to make sure it mostly works.

benoutram (approved):

PR #3053, opened 2021-12-10
Refactor admin page tests to avoid blocking `waitFor` failures

ntsim:
This PR refactors tests for various admin page tests to avoid `waitFor` calls failing intermittently.

As previously mentioned on Slack, we should avoid using `getByRole` within a `waitFor` as it can cause test processes to choke and fail (due to `getByRole` being slow) e.g:

```ts
await waitFor(() => {
  // Bad ❌
  expect(screen.getByRole('button', { name: 'Something' })).toBeInTheDocument();
});
```

Instead, we should use faster `getByText` or `getByTestId` selectors:

```ts
await waitFor(() => {
  // Good ✅
  expect(screen.getByText('Something')).toBeInTheDocument();
});

// Fine to use `getByRole` outside of the `waitFor`
expect(screen.getByRole('button', { name: 'Something' })).toBeInTheDocument();
```

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/data/components/ImporterStatus.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/data/components/ImporterStatus.tsx
@@ -64,7 +64,6 @@ const getImportStatusColour = (
     case 'FAILED':
     case 'CANCELLED':
       return 'red';
-      return 'red';
     default:
       return undefined;
   }
--- a/src/explore-education-statistics-admin/src/pages/release/data/components/__tests__/ReleaseDataUploadsSection.test.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/data/components/__tests__/ReleaseDataUploadsSection.test.tsx
@@ -143,59 +143,56 @@ describe('ReleaseDataUploadsSection', () => {
         'release-1',
       );
 
-      const sections = screen.getAllByTestId('accordionSection');
-
-      expect(sections).toHaveLength(2);
-
-      const section1 = within(sections[0]);
-
-      expect(
-        section1.getByRole('button', { name: 'Test data 1' }),
-      ).toBeInTheDocument();
-
-      expect(section1.getByTestId('Subject title')).toHaveTextContent(
-        'Test data 1',
-      );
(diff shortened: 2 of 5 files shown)


luke-h1 (changes requested):
I'm noticing some tests are still failing after running `test:ci` locally: 

```
PASS src/pages/release/datablocks/__tests__/ReleaseDataBlocksCreatePage.test.tsx (11.119 s)
A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --runInBand --detectOpenHandles to find leaks.

Summary of all failing tests
FAIL src/pages/release/datablocks/__tests__/ReleaseDataBlockEditPage.test.tsx (116.822 s)
  ● ReleaseDataBlockEditPage › clicking `Cancel` on modal hides it

    : Timeout - Async callback was not invoked within the 10000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 10000 ms timeout specified by jest.setTimeout.Error:

      300 |   });
      301 | 
    > 302 |   test('clicking `Cancel` on modal hides it', async () => {
          |   ^
      303 |     dataBlockService.getDeleteBlockPlan.mockResolvedValue({
      304 |       dependentDataBlocks: [
      305 |         {

      at new Spec (node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.test (src/pages/release/datablocks/__tests__/ReleaseDataBlockEditPage.test.tsx:302:3)
      at Object.describe (src/pages/release/datablocks/__tests__/ReleaseDataBlockEditPage.test.tsx:40:1)

  ● ReleaseDataBlockEditPage › read-only view › renders page elements correctly

    TestingLibraryElementError: Unable to find an accessible element with the role "tab"

    Here are the accessible roles:
```

ntsim:
@luke-h1 I wasn't seeing any failures using `test:ci` but I've made the additional optimisations in the data block page tests. Unfortunately, these kinds of test failures depend are environment dependent so we're not likely to get consistent test failures across the board.

I'd suggest we get this merged in if the pipeline tests pass.

luke-h1 (approved):

PR #3054, opened 2021-12-10
EES-2960 - investigate alerts for data factory failed jobs

luke-h1:
This PR: 
- is an initial attempt to try to get data factory alerts working for production environments


markhiveit (approved):
I'm inclined to think that the infra deploy is going to fail with this - but still worth a shot. 

If it does fail, if you haven't already, might be worth manually creating Data Factory alerts for failures on Failed pipeline runs and Failed trigger runs. 

If manually creating them works, adding those alerts into the ARM templates might be tough with the way we have Data Factory config in `infrastructure/templates/datafactory/template.json` - maybe the alerts could be added there?

PR #3055, opened 2021-12-13
EES-2940 Allow BAU users and Publication Owners to invite new Contributors to the service

markhiveit:
This PR allows BAU users and publication owners to invite analysts to be contributors on a release under their control.

image

When this work is merged, the `notifyContributorTemplateId` variable will need to be created on all environments.

This work also includes:
- refactoring of several methods related to invites in `PreReleaseUserService`
- changes to `RemoveAllUserContributorPermissionsForPublication` such that any associated `UserReleaseInvite`s, accepted or not, are removed. This is done to match code in `PreReleaseUserService`

I've not yet written the unit tests. To prevent having to rewrite them, I'm creating this draft PR for feedback so I don't have to redo the unit tests if a major change is required.


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
```diff
@@ -305,5 +405,77 @@ await _usersAndRolesDbContext.UserRoles
...
+            {
+                return ValidationActionResult(InvalidEmailAddress);
+            }
+
+            return Unit.Instance;
+        }
+
+        private async Task<Either<ActionResult, Unit>> ValidateUserDoesNotExist(string email)
+        {
+            if (_usersAndRolesDbContext.Users.Any(u => u.Email.ToLower() == email.ToLower()))
```
benoutram, line 421: Does it have to be the usersAndRolesDbContext user that you are looking for here. If not you can use `UserRepository#FindByEmail`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
```diff
@@ -305,5 +405,77 @@ await _usersAndRolesDbContext.UserRoles
...
+            var scheme = _httpContextAccessor.HttpContext.Request.Scheme;
+            var host = _httpContextAccessor.HttpContext.Request.Host;
+            var url = $"{scheme}://{host}/";
+
+            var releaseTitleBullets = await _contentDbContext.Releases
+                .AsAsyncEnumerable()
+                .Where(r => releaseIds.Contains(r.Id))
+                .Select(r => $"* {r.Title}")
+                .ToListAsync();
+            var releaseList = string.Join("\n", releaseTitleBullets);
```
benoutram, line 465: Can also use `releaseTitleBullets.JoinToString('\n')` here instead of `string.Join`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
```diff
@@ -305,5 +405,77 @@ await _usersAndRolesDbContext.UserRoles
...
+            if (releaseList.IsNullOrEmpty())
```
benoutram, line 466: Can/should this ever occur?

benoutram (commented):
Disclaimer: didn't look at any tests until they are finished 🙂

infrastructure/templates/template.json
```diff
@@ -1661,6 +1665,7 @@
         "NotifyPublicationRoleTemplateId": "[parameters('notifyPublicationRoleTemplateId')]",
         "NotifyReleaseRoleTemplateId": "[parameters('notifyReleaseRoleTemplateId')]",
         "NotifyPreReleaseTemplateId": "[parameters('notifyPreReleaseTemplateId')]",
+        "NotifyContributorTemplateId": "[parameters('notifyContributorTemplateId')]",
```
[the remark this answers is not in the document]
markhiveit: I've renamed the template to `User Contributor Invite` and moved it into the `Admin/Users` folder in the live service.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
```diff
@@ -262,6 +277,91 @@ await _usersAndRolesDbContext.UserInvites
                 });
         }
 
+        public async Task<Either<ActionResult, Unit>> InviteContributor(string email, Guid publicationId,
```
markhiveit, line 280: Discussed this method with Ben on Slack - I will pull out most of the logic from this method into something like "CreateForNewUser" and "CreateForExistingUser" to match the pattern used in PreReleaseUserService.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserInviteRepository.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            _usersAndRolesDbContext = usersAndRolesDbContext;
+        }
+
+        public async Task<UserInvite> CreateIfNoOtherUserInvite(string email, Role role, Guid createdById)
+        {
+            return await CreateIfNoOtherUserInvite(email, role.GetEnumValue(), createdById);
+        }
+
+        public async Task<UserInvite> CreateIfNoOtherUserInvite(string email, string roleId, Guid createdById)
```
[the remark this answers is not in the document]
markhiveit: This is code that was originally in PreReleaseUserService that I pulled into the UserInviteRepository. I've not investigated why it is this way. I've created a ticket to investigate: EES-2978

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserReleaseInviteRepository.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        public async Task<Either<ActionResult, Unit>> CreateManyIfNotExists(List<Guid> releaseIds, string email,
+            ReleaseRole releaseRole, bool emailSent, Guid createdById, bool accepted = false)
+        {
+            foreach (var releaseId in releaseIds)
+            {
+                if (!await UserHasInvite(releaseId, email, releaseRole))
+                {
+                    await _contentDbContext.AddAsync(
```
ntsim, line 53: Not sure if it makes a big difference performance wise, but we could use `AddRangeAsync` to add many entities at the same time. In combination with `ToAsyncEnumerable` (one of the correct cases to use it), we could refactor it to something like this:

```cs
await _contentDbContext.AddRangeAsync(
    releaseIds.ToAsyncEnumerable()
        .WhereAwait(releaseId => UserHasInvite(releaseId, email, releaseRole))
        .Select(releaseId => ...)
);
```

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserInviteRepositoryTests.cs
```diff
@@ -0,0 +1,116 @@
...
+
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
+{
+    public class UserInviteRepositoryTests
+    {
+        private static readonly Guid _createdById = Guid.NewGuid();
+
+        [Fact]
+        public async Task CreateIfNoOtherUserInvite_role()
```
ntsim, line 20: Nitpick, but the casing for these test names is inconsistent. The noun at the otherside of the underscore shouldn't be lowercased e.g. `CreateIfNoOtherUserInvite_Role`. 

Same for the other test cases

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/IUserReleaseInviteRepository.cs
```diff
@@ -0,0 +1,27 @@
...
+        Task<Either<ActionResult, Unit>> CreateManyIfNotExists(List<Guid> releaseIds, string email,
+            ReleaseRole releaseRole, bool emailSent, Guid createdById, bool accepted = false);
+
+        Task MarkEmailAsSent(UserReleaseInvite invite);
+
+        Task<bool> UserHasInvite(Guid releaseId, string email, ReleaseRole role);
+
+        Task<bool> UserHasInvites(List<Guid> releaseIds, string email, ReleaseRole role);
+
+        Task RemoveMany(List<Guid> releaseIds, string email, ReleaseRole role);
```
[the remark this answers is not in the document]
ntsim: Seems like `RemoveByPublication` would probably work better if the above is true!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/IUserReleaseInviteRepository.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces
+{
+    public interface IUserReleaseInviteRepository
+    {
+        Task<Either<ActionResult, Unit>> Create(Guid releaseId, string email, ReleaseRole releaseRole,
+            bool emailSent, Guid createdById, bool accepted = false);
+
+        Task<Either<ActionResult, Unit>> CreateManyIfNotExists(List<Guid> releaseIds, string email,
+            ReleaseRole releaseRole, bool emailSent, Guid createdById, bool accepted = false);
```
ntsim, line 17: Would format these so that they're nicely formatted as multi-line args. No need to have them crammed onto two lines, particularly as it makes seeing diffs in the lines a little harder later on.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserReleaseInviteRepository.cs
```diff
@@ -0,0 +1,120 @@
...
+            );
+
+            await _contentDbContext.SaveChangesAsync();
+
+            return Unit.Instance;
+        }
+
+        public async Task<Either<ActionResult, Unit>> CreateManyIfNotExists(List<Guid> releaseIds, string email,
+            ReleaseRole releaseRole, bool emailSent, Guid createdById, bool accepted = false)
+        {
```
ntsim, line 48: Would format these args across multiple lines - one per line.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserReleaseInviteRepository.cs
```diff
@@ -0,0 +1,120 @@
...
+        private readonly ContentDbContext _contentDbContext;
+
+        public UserReleaseInviteRepository(ContentDbContext contentDbContext)
+        {
+            _contentDbContext = contentDbContext;
+        }
+
+        public async Task<Either<ActionResult, Unit>> Create(Guid releaseId, string email,
+            ReleaseRole releaseRole, bool emailSent, Guid createdById, bool accepted = false)
+        {
```
ntsim, line 27: Would format these args across multiple lines - one per line.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserManagementServiceTests.cs
```diff
@@ -192,6 +204,335 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+                Assert.True(userReleaseInvites[0].Accepted);
+                Assert.True(userReleaseInvites[0].EmailSent);
+                Assert.InRange(DateTime.UtcNow.Subtract(userReleaseInvites[0].Created).Milliseconds, 0, 1500);
+            }
+        }
+
+        //[Fact]
+        //public async Task InviteContributor_UserAlreadyHasReleaseRoleInvites()
+        //{
+        //}
```
ntsim, line 440: Should this be filled out?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserManagementServiceTests.cs
```diff
@@ -192,6 +204,335 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+                Assert.True(userReleaseInvites[0].Accepted);
+                Assert.True(userReleaseInvites[0].EmailSent);
+                Assert.InRange(DateTime.UtcNow.Subtract(userReleaseInvites[0].Created).Milliseconds, 0, 1500);
+            }
+        }
+
+        //[Fact]
+        //public async Task InviteContributor_SendContributorInviteEmailFailure()
+        //{
+        //}
```
ntsim, line 534: Should this be filled out?

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
```diff
@@ -305,5 +405,77 @@ await _usersAndRolesDbContext.UserRoles
...
+                .LoadAsync();
+
+            var publicationReleaseIds = publication
+                .ListActiveReleases()
+                .Select(r => r.Id)
+                .ToList();
+
+            var intersectedReleaseIds = releaseIds.Intersect(publicationReleaseIds).ToList();
+
+            if (!intersectedReleaseIds.All(releaseIds.Contains)) // if lists don't contain same elements
```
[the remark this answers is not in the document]
markhiveit: Not sure how I made that so complicated...

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserReleaseInviteRepository.cs
```diff
@@ -0,0 +1,120 @@
...
+                    i.ReleaseId == releaseId
+                    && i.Email.ToLower().Equals(email.ToLower())
+                    && i.Role == role);
+        }
+
+        public async Task<bool> UserHasInvites(List<Guid> releaseIds, string email, ReleaseRole role)
+        {
+            var inviteReleaseIds = await _contentDbContext
+                .UserReleaseInvites
+                .AsAsyncEnumerable()
```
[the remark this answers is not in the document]
markhiveit: Changed - but I'm not sure why `AsQueryable` is right here?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/IUserInviteRepository.cs
```diff
@@ -0,0 +1,15 @@
...
+using GovUk.Education.ExploreEducationStatistics.Admin.Areas.Identity.Data.Models;
+using GovUk.Education.ExploreEducationStatistics.Admin.Models;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces
+{
+    public interface IUserInviteRepository
+    {
+        public Task<UserInvite> CreateIfNoOtherUserInvite(string email, Role role, Guid createdById);
+
+        public Task<UserInvite> CreateIfNoOtherUserInvite(string email, string roleId, Guid createdById);
```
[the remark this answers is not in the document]
markhiveit: I didn't realise that email was the primary key - I've made the change. 

Generally, I think I'd prefer for methods to be more verbose and accurate in these kinds of instance - the extra words seems like a very small cost to me compared to the added potential for bugs by being concise. If it were up to me, even in this instance I think I'd leave it - I don't see much benefit to being concise beyond aesthetics - and I'll happily trade less bugs for bad aesthetics.

Be interested in what the other devs think about this.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserReleaseInviteRepository.cs
markhiveit: I've made the change. I did a cursory test of the above change and it seems to work fine.

I had originally written this code using async stuff - not exactly as you've suggested - and I was getting strange errors where not all the invites were getting created. So I switched to using the foreach, and that solved it.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserInviteRepositoryTests.cs
markhiveit: I lowercased it to indicate it was referring to the argument variable, to distinguish which overloaded method it's using. I'll still change it.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserReleaseInviteRepository.cs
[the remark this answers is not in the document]
ntsim: Whenever accessing the database context, `AsQueryable` is the right thing to use as we want to create compiled LINQ database expressions.

`AsAsyncEnumerable` is very different. It's for situations where we want to do async things on an enumerable. We typically do this via **extension methods** (this is important) like `SelectAwait`, `WhereAwait`, etc. The enumerable is typically for something that has already been fetched and is already in memory.

Incorrectly using `AsAsyncEnumerable` here is that it implies we have access to the various async methods available, when in reality we do not. I'm fairly sure if we tried to call one of those async methods, we would get a nice LINQ error saying we can't (rightfully) compile the expression.

`DbSet` (which we define our entities with inside of the context) inherits both an `IQueryable` and `IAsyncEnumerable`. This means that it isn't possible for the compiler to know if we want to access the extension methods for `IQueryable` or `IAsyncEnumerable`, hence why we need to tell it explicitly using `AsQueryable` or `AsAsyncEnumerable` in the first place. It's up to us to declare this correctly depending on what context we're working in.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
markhiveit: It shouldn't occur. I decided to include this defensively. It could throw a validation error instead?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
markhiveit: I've made the change. Is there an advantage to doing it this way?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
markhiveit: This was something I refactored into its own method, so I'm not 100% sure, but `InviteUser` does handle creating creating a usersAndRolesDbContext `UserInvite`. I suspect it would work still with using FindByEmail, but I think then it's potentially relying on the two different tables being in sync? 

Not sure if contentDbContext Users and usersAndRolesDbContext Users are different? I assume these, in the DB, refer to Users and AspNetUsers respectively?

Making this change would mean adding another repository to `UserManagementService`, so I'll leave it for now. Happy to make the change if you still think it's preferable.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
```diff
@@ -305,5 +405,77 @@ await _usersAndRolesDbContext.UserRoles
                         .OnSuccess(() => _userRoleService.AddGlobalRole(userId, roleId));
                 });
         }
+
+        private async Task<Either<ActionResult, Unit>> ValidateEmail(string email)
+        {
+            if (string.IsNullOrWhiteSpace(email))
```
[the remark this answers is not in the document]
markhiveit: As with `ValidateUserDoesNotExist`, this is preexisting code that I've refactored. So I've left it the way it was originally.

Since the plan now is to change email validation in a different ticket, I'm inclined to leave this for now.

src/GovUk.Education.ExploreEducationStatistics.Admin/Validators/EmailValidator.cs
```diff
@@ -0,0 +1,32 @@
...
+using GovUk.Education.ExploreEducationStatistics.Common.Model;
+using Microsoft.AspNetCore.Mvc;
+using static GovUk.Education.ExploreEducationStatistics.Admin.Validators.ValidationErrorMessages;
+using static GovUk.Education.ExploreEducationStatistics.Admin.Validators.ValidationUtils;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Validators
+{
+    public static class EmailValidator
+    {
+        public static Either<ActionResult, List<string>> ValidateEmailAddresses(IEnumerable<string> input)
```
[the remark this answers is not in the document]
markhiveit: As discussed, I've created a separate ticket to address this: EES-2989

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ContributorInviteRequest.cs
```diff
@@ -0,0 +1,13 @@
+#nullable enable
+using System;
+using System.Collections.Generic;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.ViewModels
+{
+    public class ContributorInviteRequest
+    {
+        public string Email { get; set; }
```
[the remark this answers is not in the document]
markhiveit: Added `[Required]` to both and tested sending requests without the fields, and got appropriate errors:

image

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseInviteService.cs
```diff
@@ -0,0 +1,231 @@
...
+                .CheckEntityExists<Publication>(publicationId, query => query
+                    .Include(p => p.Releases))
+                .OnSuccessDo(
+                    async publication => await _userService.CheckCanUpdateReleaseRole(publication, Contributor))
+                .OnSuccessDo(() => ValidateEmail(email))
+                .OnSuccess(publication => ValidateReleaseIds(publication, releaseIds))
+                .OnSuccess(async publication =>
+                {
+                    email = email.Trim();
+                    releaseIds = releaseIds.Distinct().ToList();
```
markhiveit, line 74: I've added these two lines - I wanted to highlight them in case others think I should do this a different way.

I've trimmed the email string to avoid duplicate entries for the same email being possible due to whitespace.
I've made releaseIds distinct, because I realised that maybe issues could occur if the same releaseId was provided twice in a request. Not sure that it actually would, but this removes all possibility.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/IUserInviteRepository.cs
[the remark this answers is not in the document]
ntsim: You're right, there's not much in it in this instance so I'm not super bothered either way. I've not mentioned it in other places where it's probably better to be more explicit :wink:

Think it's worth elaborating on this a bit further to give you a more concrete rationale for being less explicit. There's a few things I like to think about in terms of naming things:

1. Direct context - does the name need to include things that can already be inferred from the class/module/scope?

 This is fairly simple, avoid repeating what the surrounding context is already telling us. In this example, we're in a `IUserInviteRepository`, so there's no point in a method being called `CreateUserInvite`. This is just redundant to declare as the class/interface name tells us that it interacts with a database (it's a repository), and it's specifically working with `UserInvites`.

2. Method signature - does the name tell us anything that the arguments and return type don't?

A method isn't just it's name, it's also its arguments and return type. A method like `FindById(Guid Id)` is unnecessarily verbose as we only offer one parameter - an ID. It's entirely redundant to include the 'ById' part.

Explicit naming is way more beneficial in dynamic languages where we don't know what the hell we're working with :grimacing: It doesn't mean we need to do in static languages where we have much richer type information.

3. Other methods - does the name conflict with any similar sounding methods in the current scope?

When we have methods that sound or do similar things, we may want to be more explicit. However, if the functions are basically the same, but the arguments are different, we usually just want to fallback on overloading to simplify the API.

When consuming any API, we typically want to avoid burdening the consumer with excessive methods as this creates a burden of choice. It'd be pretty overwhelming to try auto-complete your method and be greeted by tens of methods with slight variations of the same name. 

I usually see overloading as a way to funnel consumers towards using certain methods. If you've ever used any decent library API/SDK, many of them use overloading in this way.

4. General architecture and responsibilities - does the name work with how the class and method is supposed to be used in the bigger picture?

This circles back a bit towards 1, but it's worth just repeating that naming is impacted by the direct context. In our example, we're working with a repository class, so we know that a method like `Create` should create something in the database. However, we should also think about the repository's responsibility and what it _should_ be doing in the grand scheme of things. 

In general, with the service-repository pattern, we want to leave business logic in the service, whilst repositories should be purely concerned with database access. Unfortunately, it can be unclear where one responsibility ends and the other starts (hence why we're even talking about this). 

I'd probably argue that a repository should just plainly allow us to save something into the database without worrying too much about the surrounding rules. From the service's standpoint, it just wants to tell the repository to do something and get some indication of whether that succeeded or not (e.g. from what it returns or throws).

As I mentioned initially, it's a bit of an implementation detail if an email actually gets created in the database or not, especially if it's not creating one because it already exists. It's not important to know how we got to the result, rather that we got to it at all.

Of course, this is something we could probably debate about forever :sweat_smile: 

5. Expressiveness - does the name express the action succinctly and avoid visual clutter? Is it confusing to read?

Whilst being explicit has benefits, it has a clear impact on creating more visual clutter and making things harder to read. For example, if we were being super explicit, we might name our method something silly like `CreateUserInviteIfItDoesNotExistInSQLDatabase`. This is of course ridiculous, but illustrates the point that being more explicit doesn't directly convert into a better experience.

Personally, a good name is short and sweet. It should concisely express an idea, like any good writing. Any additional information in the name needs to have a good reason to exist, otherwise, it's just noise and should be pruned. As Marie Kondo says - if it doesn't spark joy, get rid of it.

Finally, it's worth mentioning that explicitness can just increase confusion. In our example, `CreateIfNotExists` can imply that there is a way to create multiple of the same invite in the database. This isn't the case, so it's a little confusing to imply that something like that is on the table. Whilst not a major deal, it prompts an idea that we don't really need to burden people with.

Anyway, writing this all out ended up being way longer than I anticipated :laughing: I'm probably going to write a blog post about this at some point!

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/UserInviteRequest.cs
```diff
@@ -1,3 +1,4 @@
+#nullable enable
```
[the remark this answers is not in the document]
markhiveit: I've added [Required] to both fields - same as ContributorRequestInvite.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserReleaseRoleRepository.cs
```diff
@@ -135,16 +179,26 @@ public async Task<bool> IsUserEditorOrApproverOnLatestRelease(Guid userId, Guid
                 EditorAndApproverRoles);
         }
 
-        public async Task<bool> UserHasRoleOnRelease(Guid userId, Guid releaseId, ReleaseRole role)
+        public async Task<UserReleaseRole?> GetReleaseRole(Guid userId, Guid releaseId, ReleaseRole role)
```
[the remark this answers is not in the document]
markhiveit: I've created restored `UserHasRoleOnRelease` as `HasUserReleaseRole`. I've kept one of the `GetUserReleaseRole` methods which is used in `CreateIfNotExists` - to save an extra db call.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PreReleaseUserService.cs
```diff
@@ -260,7 +268,8 @@ await validEmails
                                     release,
                                     invite.Email.ToLower(),
                                     isNewUser: user == null)
-                                .OnSuccessDo(async () => await MarkInviteEmailAsSent(invite));
+                                .OnSuccessDo(async () =>
+                                    await _userReleaseInviteRepository.MarkEmailAsSent(invite));
```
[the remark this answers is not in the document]
markhiveit: I believe I pulled this out expecting that I would be using it in ReleaseInviteService, but that ended up not being the case. I'm inclined to leave it now it's done, but can move it back if you'd prefer.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/IUserReleaseInviteRepository.cs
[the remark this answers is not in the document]
markhiveit: Changed it to be `RemoveByPublication`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
```diff
@@ -262,6 +277,91 @@ await _usersAndRolesDbContext.UserInvites
...
+                            && urr.Role == Contributor
+                            && urr.UserId == user.Id)
+                        .Select(urr => urr.ReleaseId)
+                        .ToList();
+
+                    var missingReleaseRoleReleaseIds = releaseIds
+                        .Where(releaseId => !existingReleaseRoleReleaseIds.Contains(releaseId))
+                        .ToList();
+
+                    if (missingReleaseRoleReleaseIds.IsNullOrEmpty())
```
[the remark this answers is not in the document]
markhiveit: I'm inclined to leave this, as the `missingReleaseRoleReleaseIds` variable is used after this if, and I feel the `IsNullOrEmpty` call is easier to understand.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserReleaseInviteRepository.cs
```diff
@@ -0,0 +1,120 @@
...
+                    ReleaseId = releaseId,
+                    Role = releaseRole,
+                    Accepted = accepted,
+                    EmailSent = emailSent,
+                    Created = DateTime.UtcNow,
+                    CreatedById = createdById,
+                }
+            );
+
+            await _contentDbContext.SaveChangesAsync();
```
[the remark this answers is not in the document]
markhiveit: I think this was addressed when I moved the sending email stuff to the top and stopped if it returned an error. But I'm not 100% so leaving this unresolved.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ContributorInviteRequestViewModel.cs
```diff
@@ -0,0 +1,14 @@
+#nullable enable
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.ViewModels
+{
+    public class ContributorInviteRequestViewModel
```
ntsim, line 8: Could simplify this to either `ContributorInviteRequest` or `ContributorInviteViewModel`. Wouldn't combine them both.

Can appreciate that our naming conventions around this are basically a mess, but this isn't really a view model because it's not going to be featured in responses. On the otherhand, it could be argued that we might consider everything at the HTTP boundaries (incoming or outgoing) to be a 'view model'. Either way, wouldn't make the situation even worse by combining the two conventions.

Not sure about how others feel about this @benoutram @duncan-at-hiveit, but feels like we need to get a handle on these naming conventions! Personally, I'd probably lean towards splitting out a `Requests` namespace that specifically contains requests.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/PreReleaseUserViewModels.cs
```diff
@@ -23,7 +23,7 @@ public class PreReleaseUserInvitePlan
         public List<string> Invitable { get; } = new List<string>();
     }
 
-    public class PreReleaseUserInviteRequest
+    public class PreReleaseUserInviteRequestViewModel
```
ntsim, line 26: Same as above - would call this either `PreReleaseUserInviteViewModel` or `PreReleaseUserInviteRequest`.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/UserInviteRequestViewModel.cs
```diff
@@ -0,0 +1,12 @@
+using System.ComponentModel.DataAnnotations;
+
+#nullable enable
+namespace GovUk.Education.ExploreEducationStatistics.Admin.ViewModels
+{
+    public class UserInviteRequestViewModel
```
ntsim, line 6: Same as above - would call this either `UserInviteViewModel` or `UserInviteRequest`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseInviteService.cs
```diff
@@ -0,0 +1,230 @@
...
+                email: email,
+                releaseRole: Contributor,
+                emailSent: true,
+                createdById: _userService.GetUserId(),
+                accepted: true);
+
+            return Unit.Instance;
+        }
+
+        private async Task<Either<ActionResult, Unit>> ValidateEmail(string email)
```
ntsim, line 170: Don't need this - we can just use the `[EmailAddress]` annotation on the request. This validation is also useless because it doesn't properly check that it's a valid email address.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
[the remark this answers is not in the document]
ntsim: Disagree - it's super easy to fix this by adding the `[EmailAddress]` annotation to the request. In fact, this validation is useless because it doesn't even actually validate the email properly!

ntsim (changes requested):

infrastructure/templates/template.json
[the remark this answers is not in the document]
benoutram: I've also created https://dfedigital.atlassian.net/browse/EES-2991 about moving the template id's to key vault.

Reviewing the DevOps configuration in the parameters of the Infrastructure Deploy Task, I don't see this new template id added yet (or at least not in Dev) so I think you still need to update all of the DevOps configuration.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/UserInviteRequestViewModel.cs
benoutram: I think this change has been in response to my comment that we now seem to have a convention of suffixing all the request models `ViewModel`. I'd been thinking `PreReleaseUserInviteViewModel` and `UserInviteViewModel` would be best.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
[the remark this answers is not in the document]
markhiveit: I was considering the time consuming thing to be understanding what the EmailAddress annotation does, whether it is adequate, and removing all the existing email validation across all our services.

I'll add it to the places you've suggested.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ContributorInviteRequestViewModel.cs
markhiveit: I changed `ContributorInviteRequest` to `ContributorInviteRequestViewModel` in response to another PR comment. I've left a comment about this in Slack and will change it when a consensus is reached.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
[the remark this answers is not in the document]
markhiveit: Incidentally, looking into it, this is the checks that the annotation does:
```
            if (value == null)
            {
                return true;
            }

            if (!(value is string valueAsString))
            {
                return false;
            }

            // only return true if there is only 1 '@' character
            // and it is neither the first nor the last character
            int index = valueAsString.IndexOf('@');
            
            return
                index > 0 &&
                index != valueAsString.Length - 1 &&
                index == valueAsString.LastIndexOf('@');
```

So I think we may want something more thorough and this. Searching online, it seems others tend towards some crazy regex.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
benoutram: In this example not really but in general extension methods can help remove some boiler plate code and complexity.
In this case there's little difference between `x.JoinToString('\n')` and the implementation of it, `string.Join('\n', x)`.

In a change I'm currently making to these extension methods I allow the collection to be of any type avoiding the need to 'ToString' all the items of the collection if they aren't already strings, so it's about to become slightly more useful.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ContributorInviteRequestViewModel.cs
markhiveit: Consensus was to use `BlahViewModel` moving forward. I'll make this change to all other instances.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/PreReleaseUserViewModels.cs
markhiveit: After Slack discussing, renaming to follow `BlahViewModel` convention

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/UserInviteRequestViewModel.cs
markhiveit: After Slack discussing, renaming to follow `BlahViewModel` convention

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
benoutram: It just confused me and immediately made me wonder what situation allows that but it turns out it's hypothetical.

I don't do this type of defensive coding myself because it allows problems causing scenarios like this elsewhere to go unseen.

An end user isn't necessarily going to mention to us that they are receiving emails about being granted no releases.

I can understand being defensive if this situation could arise because of old or bad data that isn't worth tidying up but this situation can only arise as a result of there being something wrong with this new bit of functionality added here allowing users to be invited with no releases which should never happen.

If something could blow up as a result of there being no Releases then I'd say it would better to do so, so that we find out about it.

I see in this particular case though that I don't think anything could blow up, the email would probably just contain a blank value.

If you want to be defensive about it and give us a better chance of spotting an issue then you could begin with something like

```
if (!releases.Any()) { throw new ArgumentException("List of releases cannot be empty", nameof(releases)); }
```

and then there should be no need for the check later on.

infrastructure/templates/template.json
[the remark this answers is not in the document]
markhiveit: My intention is to add the template id after this gets merged

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseInviteService.cs
benoutram: Agree about the release id's. Could it potentially be hoisted any higher in this method as we also check all the release id's belong to the publication?

For the email address, it's a shame the validator doesn't prevent against this. Looking at the code of it, I don't think it does. If this input can come from a user then I'd also check we are doing 'ToLower' with anywhere that it's being used. I think that's probably already the case with all the repository classes though.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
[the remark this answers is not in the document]
benoutram: I couldn't find where this is now. It might have been refactored with othe changes?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PreReleaseUserService.cs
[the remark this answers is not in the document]
benoutram: In general I think its overly specific and I would keep the repository cleaner by limiting it to CRUD operations on the resource that are providing some value and reusable abstraction over the equivalent of what has to be done with the database context alone.

In this case 'MarkEmailAsSent' sounds like business logic that should be in a service rather than a repository and it provides nothing that can't be achieved by updating the invite and calling SaveChanges on it. If we were to write repository methods for updating resources for every place we allow updating a single field then they would be huge.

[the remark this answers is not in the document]
markhiveit: I've moved it back

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
markhiveit: Makes sense. I've changed it. In the future I'll throw Exceptions for this kind of thing.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseInviteService.cs
markhiveit: I've moved the duplicate releaseIds check into ValidateReleaseIds, where it throws an exception if there are duplicates.

I checked and email.ToLower() is called everywhere except where the email is sent.

markhiveit: I've moved the releaseIds distinct check into ValidateReleaseIds method.

Discussed emails with Ben on Slack a little. Feels like we need some kind of central validate/clean email class that all email arguments get pushed through - and then all the various ad hoc email validation and cleaning / ToLower() can be removed. I think this can be done as part of the work for EES-2989

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ContributorInviteRequest.cs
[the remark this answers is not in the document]
benoutram: I think that's right and what I was expecting but the annoying thing is it leaves a warning that those fields are potentially uninitialised .

image

If we fix those warning by initialising them (i.e. with string.empty for the email and a new list for the Release id's) then the `[Required]` validation will never fire.

Alternatively if we fix those warning by making them nullable then every place using them has to handle them potentially being null, when in reality we know that because of the `[Required]` validation that null should never be a possibility.

I think I'll raise this with the other developers to try and reach a consensus on what we should do here.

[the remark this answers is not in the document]
benoutram: We just discussed this and decided that we will do a test omitting the fields in the API request after initialising them to default values i.e. with string.empty for the email and a new list for the Release id's, and double check whether the `[Required]` validation fires or not.

If it doesn't then we can initialise them with `null!` so the warnings disappear but the validation should still fire.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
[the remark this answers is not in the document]
benoutram: I think my point about simplifying could still be applied though, 

 `.Where(releaseId => !existingReleaseRoleReleaseIds.Contains(releaseId))` 

can be simplified to 

`.Except(existingReleaseRoleReleaseIds)`

Also the result of `ToList` can never be null, so `if (missingReleaseRoleReleaseIds.IsNullOrEmpty())` can be `if (!missingReleaseRoleReleaseIds.Any())

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ContributorInviteRequest.cs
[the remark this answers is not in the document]
markhiveit: Made changes to `ContributorInviteViewModel` and `UserInviteViewModel` as discussed in Slack

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
benoutram: You are right, they are different types of Users (Users vs. AspNetUsers)

Because this is used for inviting users to the system with system roles (e.g. to invite the BAU user) I think it's correct that this needs to check `_usersAndRolesDbContext`.

By the way though, I think you could use `AnyAsync` in this async method.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserReleaseInviteRepositoryTests.cs
```diff
@@ -0,0 +1,306 @@
...
+                    releaseRole: Contributor,
+                    emailSent: true,
+                    createdById: createdById);
+            }
+
+            await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContextId))
+            {
+                var userReleaseInvite = contentDbContext.UserReleaseInvites
+                    .AsQueryable()
+                    .SingleOrDefault();
```
benoutram, line 38: You could also `var userReleaseInvite = Assert.Single(contentDbContext.UserReleaseInvites);`

alternatively

```
await contentDbContext.UserReleaseInvites
                    .AsQueryable()
                    .SingleOrDefaultAsync();
```

for consistent use of async on database operations although I don't think it really matters.

I noticed it's not consistent throughout this class either, e.g. elsewhere there's a mixture of two different styles for enumerating all the table entries, e.g.

 `var remainingInvites = contentDbContext.UserReleaseInvites.ToList();`

vs.

```
                var userReleaseInvites = contentDbContext.UserReleaseInvites
                    .AsQueryable()
                    .ToList();
```

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserReleaseRoleRepositoryTests.cs
```diff
@@ -69,7 +68,96 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+                    .AsQueryable()
+                    .ToListAsync();
+                Assert.Single(userReleaseRoles);
+
+                Assert.NotEqual(Guid.Empty, userReleaseRoles[0].Id);
+                Assert.Equal(userReleaseRole.UserId, userReleaseRoles[0].UserId);
+                Assert.Equal(userReleaseRole.ReleaseId, userReleaseRoles[0].ReleaseId);
+                Assert.Equal(userReleaseRole.Role, userReleaseRoles[0].Role);
+                Assert.Equal(userReleaseRole.CreatedById, userReleaseRoles[0].CreatedById);
+                Assert.Equal(userReleaseRole.Created, userReleaseRoles[0].Created);
```
benoutram, line 155: It would be simpler to just `var userReleaseRole = Assert.Single(contentDbContext.UserReleaseRoles);` and then use `userReleaseRole` rather than `userReleaseRoles[0]`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseInviteService.cs
```diff
@@ -0,0 +1,225 @@
...
+            return Unit.Instance;
+        }
+
+        private async Task<Either<ActionResult, Publication>> ValidateReleaseIds(Publication publication,
+            List<Guid> releaseIds)
+        {
+            var distinctReleaseIds = releaseIds.Distinct().ToList();
+            if (distinctReleaseIds.Count != releaseIds.Count)
+            {
+                throw new ArgumentException("releaseIds should not contain duplicates");
```
benoutram, line 174: Suspect I'm guilty of not doing this but we can use `nameof` so that if the field name is refactored the exception message gets refactored as well. It's also possible to specify an additional param referring to the offending argument name. See the last param here:

`throw new ArgumentException($"{nameof(releaseIds)} should not contain duplicates", nameof(releaseIds));`

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/PreReleaseUserViewModels.cs
```diff
@@ -23,13 +23,13 @@ public class PreReleaseUserInvitePlan
         public List<string> Invitable { get; } = new List<string>();
     }
 
-    public class PreReleaseUserInviteRequest
+    public class PreReleaseUserInviteViewModel
     {
-        [Required] public List<string> Emails { get; set; } = new List<string>();
+        [EmailAddress] public List<string> Emails { get; set; } = new List<string>();
```
benoutram, line 28: Does this `[EmailAddress]` validation attribute work on a collection? Interesting!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
markhiveit: Changed the `Any` to `AnyAsync`.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserReleaseInviteRepositoryTests.cs
markhiveit: Changed all calls to be async and added `AsQueryable` when appropriate

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/PreReleaseUserViewModels.cs
markhiveit: Missed that. I've changed it to use `MinLength` - the previous `Required` would have done nothing since it was being initialised as an empty list.

benoutram (dismissed):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseInviteService.cs
```diff
@@ -0,0 +1,227 @@
...
+
+        private async Task<Either<ActionResult, Publication>> ValidateReleaseIds(Publication publication,
+            List<Guid> releaseIds)
+        {
+            var distinctReleaseIds = releaseIds.Distinct().ToList();
+            if (distinctReleaseIds.Count != releaseIds.Count)
+            {
+                throw new ArgumentException($"{nameof(releaseIds)} should not contain duplicates",
+                    nameof(releaseIds));
+            }
```
ntsim, line 177: Not saying we need to change this, but throwing an error here is a little unnecessary. 

As we can filter out any duplicates ourselves, there isn't really any harm if they provide duplicates and no further action needs to be taken by the function consumer.

Throwing exceptions should be reserved more for cases where the consumer has done something that would actually cause an issue e.g. another exception being thrown (null pointers, etc), the creation of invalid state, etc. This is a case where coding defensively is just making life harder for the consumer :wink:

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/IUserInviteRepository.cs
```diff
@@ -0,0 +1,15 @@
...
+using GovUk.Education.ExploreEducationStatistics.Admin.Areas.Identity.Data.Models;
+using GovUk.Education.ExploreEducationStatistics.Admin.Models;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces
+{
+    public interface IUserInviteRepository
+    {
+        public Task<UserInvite> Create(string email, Role role, Guid createdById);
+
+        public Task<UserInvite> Create(string email, string roleId, Guid createdById);
```
ntsim, line 13: Really minor thing, but we don't need to include `public` in interface definitions

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseInviteService.cs
```diff
@@ -0,0 +1,227 @@
...
+        {
+            if (releaseIds.IsNullOrEmpty())
+            {
+                throw new ArgumentException("List of releases cannot be empty");
+            }
+
+            var template = _configuration.GetValue<string>("NotifyContributorTemplateId");
+            var scheme = _httpContextAccessor.HttpContext.Request.Scheme;
+            var host = _httpContextAccessor.HttpContext.Request.Host;
+            var url = $"{scheme}://{host}/";
```
ntsim, line 208: This doesn't look right - we set an `AdminUri` variable in configuration that should be used instead. See `EmailTemplateService` for examples where we're doing just that.

Using the context accessor could be a problem as the URL will be controlled by the request itself. It's possible (although unlikely) that a user can directly access the API through a non-DNS URL and can consequently generate an incorrect URL.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseInviteService.cs
```diff
@@ -0,0 +1,227 @@
...
+            if (!releaseIds.All(publicationReleaseIds.Contains))
+            {
+                return ValidationActionResult(NotAllReleasesBelongToPublication);
+            }
+
+            return publication;
+        }
+
+        private async Task<Either<ActionResult, Unit>> SendContributorInviteEmail(
+            string publicationTitle, List<Guid> releaseIds, string email)
```
ntsim, line 198: Seems like we construct a bunch of emails already in `EmailTemplateService`, should this be in there?

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseInviteService.cs
markhiveit: When I realised this was a potential issue, I did at first just filter duplicates out. 

Ben raised a comment about this in this PR and that's why it is now as it is. Now the change is made, I think I probably prefer this, as I think I'd want it to blow up if something is sending duplicates releaseIds. But at the same time, it seems like a slim possibility that duplicate releaseIds will reveal a bug elsewhere in our code.

I'll just leave this as is. You'll have to argue with Ben if you want it changed :)

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/IUserInviteRepository.cs
markhiveit: Fixed. I think that's me copying and pasting from the class.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseInviteService.cs
markhiveit: I didn't know `EmailTemplateService` was a thing. I was following what `PreReleaseUserService#SendPreReleaseInviteEmail` does.

In the instance of PreRelease, `SendPreReleaseInviteEmail` uses `PreReleaseService`, and if it was moved to `EmailTemplateService` either:
- that call to `_preReleaseService` would have to be done in `PreReleaseUserService` and passed to `_emailTemplateService.SendPreReleaseInviteEmail` as an argument.
- we'd need to add `PreReleaseService` to `EmailTemplateService`.

This is similarly true of `SendContributorInviteEmail`, as it requires release titles for all releaseIds fetching from the ContentDb.

Given that, I'm inclined to think that `Send*Email` methods can live in the services that use them, if they're unlikely to be called from more than one service. I checked the `Send*Email` methods in `EmailTemplateService`, and they're each all only called from one service. So my inclination is to have each service responsible for sending its own emails rather than having an `EmailTemplateService`.

All that said, I'm not sure I consider this an issue worth the time of changing things - it's probably not even worth the time I've spent writing this out. So I'd personally just leave things as they are. But happy to change things if others feel strongly about it.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseInviteService.cs
benoutram: I might be confused, as it's not easy navigating all the comments, but I think I raised the point about hoisting the `distinct` higher up in the method where you originally had it so that duplicates weren't passed into this method which checks they all belong to the same publication.

I don't think it was my suggestion to raise duplicates as an exception although later on when I saw it I did leave a comment to say the `"releaseIds"` could be `nameOf(releaseIds)` to assist with any renaming as part of refactoring in future.

image

I think you might be confusing this with the comment I left about `releaseIds` being empty when building the list of release titles before sending the email. This isn't client facing and something should be caught earlier before we reach the point of sending an email to invite for an empty list of Releases. For that reason I had an issue with sending the email containing a string like 'There are no releases'. I think I would either do nothing or throw the exception as I mentioned here which is what you ended up doing. 🙂 

image

markhiveit: Ah - my bad. For some reason I thought there was only one ArgumentException, but there are two. Getting a bit lost in all these comments. I'll be more careful before attributing changes to others in the future.

src/explore-education-statistics-admin/src/pages/publication/components/PublicationInviteNewUsersTab.tsx
```diff
@@ -1,7 +1,114 @@
...
+    },
+  }),
+];
+
+export interface Props {
+  publication: BasicPublicationDetails;
+  releases: ReleaseSummary[];
+}
+
+const PublicationInviteNewUsersTab = ({ publication, releases }: Props) => {
```
ntsim, line 45: Seems like we don't have any tests for this tab?

src/explore-education-statistics-admin/src/pages/publication/components/ReleaseContributorPermissions.tsx
```diff
@@ -31,7 +31,7 @@ const ReleaseContributorPermissions = ({
               {contributors.map(contributor => (
                 <tr key={contributor.userId}>
                   <td className="govuk-!-width-one-half">
-                    {contributor.userDisplayName}
+                    {contributor.userDisplayName} ({contributor.userEmail})
```
ntsim, line 34: Would write this line as a single template string so that it's easier to find this text in tests

src/explore-education-statistics-admin/src/services/userService.ts
```diff
@@ -158,6 +163,19 @@ const userService: UsersService = {
   inviteUser(invite: UserInvite): Promise<boolean> {
     return client.post(`/user-management/invites`, invite);
   },
+  inviteContributor(
+    email: string,
+    publicationId: string,
+    releaseIds: string[],
+  ): Promise<boolean> {
+    return client.post(
+      `user-management/publications/${publicationId}/invites/contributor`,
```
ntsim, line 172: URLs should start with a slash to be consistent

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/publication/components/PublicationInviteNewUsersTab.tsx
markhiveit: Created `PublicationInviteNewUsersTab.test.tsx`

ntsim (approved):
Awesome, LGTM :+1:

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseInviteService.cs
benoutram: No problem! I find the interface a little bit difficult when it comes to navigating through all the comments as well. 🙂

PR #3062, opened 2021-12-15
EES-2500 - unify wait times

luke-h1:


tests/robot-tests/.env.example
```diff
@@ -12,4 +12,6 @@ RELEASE_COMPLETE_WAIT=120  # (int) Time to wait for a release to publish
 WAIT_MEDIUM=120  # (int) Variable used throughout tests
 WAIT_SMALL=45 # (int) Variable used throughout tests
 WAIT_LONG=180  # (int) Variable used throughout tests
+TIMEOUT=30 # (int) Variable used throughout tests
```
duncan-at-hiveit, line 15: Hey Luke, as per Nick's comment on Friday, can we potentially remove these from the run_tests.py script now, in favour of providing environment variables in the CLI instead if we need to do overrides?

I think the syntax is something like:

`TIMEOUT=60; WAIT_LONG=240; pipenv run...`

Or at least in bash it would be that - I think Nick was gonna give you an example?

tests/robot-tests/tests/libs/common.robot
```diff
@@ -1,5 +1,5 @@
 *** Settings ***
-Library     SeleniumLibrary    timeout=${timeout}    implicit_wait=${implicit_wait}    run_on_failure=do this on failure
+Library     SeleniumLibrary    timeout=${timeout}    implicit_wait=%{IMPLICIT_WAIT}    run_on_failure=do this on failure
```
markhiveit, line 2: I'd also change `timeout=${timeout}` to use `timeout=%{TIMEOUT}`

That said, I'm not sure this change is necessary. It's possible at the point these variables are used in this line, the `*** Variables ***` section will already be run, so `${timeout}` and `%{TIMEOUT}` will be the same. But to be safe I'd use the environment variable on this line i.e. `timeout=%{TIMEOUT}`

markhiveit (changes requested):

tests/robot-tests/tests/libs/common.robot
luke-h1: Good catch, I've changed this now to %{TIMEOUT}

tests/robot-tests/.env.example
luke-h1: Discussed after standup: The `.env` file will be the source of truth for these variables

luke-h1:
In addition to unifying wait times, I've changed the type hint of `setup_auth_variables` to `-> Union[str, str]`. My IDE was giving a warning about: 
```python
Tuple expression not allowed in type annotation
  Use Tuple[T1, ..., Tn] to indicate a tuple type or Union[T1, T2] to indicate a union typePylance
```

duncan-at-hiveit (approved):

PR #3066, opened 2021-12-16
EES-2982 Throw exception if GetFootnotes and ReleaseFootnotes disagree

markhiveit:


src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
```diff
@@ -160,6 +160,18 @@ public IEnumerable<Footnote> GetFootnotes(Guid releaseId, IEnumerable<Guid> subj
                                   (f.Subjects != null && f.Subjects.Any())
                     )
                     .ToList();
+
+            // Remove the below check as part of EES-2971
+            var releaseFootnotes = _context.ReleaseFootnote
```
benoutram, line 165: The second `.Where` in the above code filters out all footnotes to only include those that have `f.Filters`, `f.Indicators`, `f.FilterGroups`, `f.FilterItems`, or `f.Subjects`.

Any of those are only going to be present if they come from a Subject in the array of SubjectId's.

So this 'count' check is only going to be correct if the array of Subject id's contains all the subjects of the Release.

If any less, e.g. called with one Subject Id, it's only going to return Footnotes with some relevance to that subject, so the check will fail.

This would be the case with GetFootnotes(Guid releaseId, Guid subjectId) which in turn calls this method. I would expect this to always fail the check if there are footnotes for more than just this subject.

benoutram (changes requested):

benoutram (approved):

PR #3067, opened 2021-12-16
EES-2980 Fixed create_emulator_release_files.py

markhiveit:


tests/robot-tests/tests/libs/create_emulator_release_files.py
```diff
@@ -203,7 +199,9 @@ def upload_files(container_client, files):
                 blob_client.upload_blob(
                     data, blob_type="BlockBlob", metadata=file.metadata())
             except ResourceExistsError:
-                pass
+                pass  # file already exists, so no action required
+            except Exception as e:
+                print('Exception when uploading file: ', e)
```
ntsim, line 204: Maybe we might want to do a similar thing in other places where we create resources as well?

ntsim (approved):

benoutram (approved):

PR #3072, opened 2021-12-21
EES-2995 Create and use CanDeleteSpecificComment security policy

markhiveit:


src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/DeleteSpecificCommentAuthorizationHandler.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    public class DeleteSpecificCommentAuthorizationHandler
+        : AuthorizationHandler<DeleteSpecificCommentRequirement, Comment>
+    {
+        private readonly ContentDbContext _contentDbContext;
+        private readonly IReleasePublishingStatusRepository _releasePublishingStatusRepository;
+        private readonly IUserPublicationRoleRepository _userPublicationRoleRepository;
+        private readonly IUserReleaseRoleRepository _userReleaseRoleRepository;
+
+        public DeleteSpecificCommentAuthorizationHandler(ContentDbContext contentDbContext,
```
ntsim, line 24: Where are the tests?

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/DeleteSpecificCommentAuthorizationHandler.cs
markhiveit: For some reason I assumed they were already written and I just needed to change the permission test - my bad

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Security/AuthorizationHandlers/Utils/ReleaseAuthorizationHandlersTestUtil.cs
```diff
@@ -326,13 +327,81 @@ public static class ReleaseAuthorizationHandlersTestUtil
...
 
-            using (var context = DbUtils.InMemoryApplicationDbContext(contextId))
+            using (var context = InMemoryApplicationDbContext(contextId))
             {
                 var handler = handlerSupplier(context);
                 await AssertHandlerHandlesScenarioSuccessfully<TRequirement>(handler, scenario);
             }
         }
 
+        public static async Task AssertHandlerOnlySucceedsWithReleaseRole<TRequirement, TEntity>(
```
duncan-at-hiveit, line 337: I'd rename this to `AssertHandlerOnlySucceedsWithReleaseRoles` plural, as you're asseting that multiple release roles could succeed.

I'd rename `successfulRoles` as `rolesExpectedToSucceed` to be in line with other methods in this class.

Is there a bit of overlap with the existing `AssertReleaseHandlerSucceedsWithCorrectReleaseRoles` method?  What's the driver for this new method?

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/DeleteSpecificCommentAuthorizationHandler.cs
```diff
@@ -0,0 +1,64 @@
...
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.EntityFrameworkCore;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Security.AuthorizationHandlers
+{
+    public class DeleteSpecificCommentRequirement : IAuthorizationRequirement
+    {
+    }
+
+    public class DeleteSpecificCommentAuthorizationHandler
```
duncan-at-hiveit, line 16: As this is a new handler that needs to be supplied via DI, you'll need to register it in `StartupSecurityConfig.ConfigureResourceBasedAuthorization()` or you'll get a null pointer error when a Service tries to use it.

duncan-at-hiveit (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Security/AuthorizationHandlers/Utils/ReleaseAuthorizationHandlersTestUtil.cs
markhiveit: I've made the name changes you suggest.

To be perfectly honest, I copied `AssertHandlerOnlySucceedsWithPublicationRoles` to create `AssertHandlerOnlySucceedsWithReleaseRoles` primarily because I understand the code and was confident I could write the test quickly.

I believe the methods both add more flexibility as the callers decides what they add to the db, where I believe other methods create and add the release and AuthorizationHandler argument for you.

In the case for which I created `AssertHandlerOnlySucceedsWithPublicationRoles`, I couldn't get the existing util methods to do what I needed. If I remember correctly, the AuthorizationHandler accepted a tuple - see `UpdateReleaseRoleAuthorizationHandlerTests` - which existing methods would try to add to the db and fail.

It's possible I've not understood the existing util methods and perhaps they already provide this flexibility. If I've not understood how to use the other util methods and they do have that flexibility, then I'm happy to switch to use them.

Either way, I think it might be good idea to unify how we go about testing permissions. Right now it feels like throughout our unit tests, we're testing permissions in many different ways, when there should be one or two standard way that we use everywhere. And thinking about DX, writing tests around permissions has been painful for me and I think we could perhaps make it easier in the future.

ntsim: Just throwing my two pence in here, but it would be nice if we could rewrite a lot this stuff to a more flexible builder approach like the `PolicyCheckBuilder`.

I'd tend to agree with Mark that these permission test utils typically do way too many things making them hard to understand.

markhiveit: I think I've got it wrong in the above comment. I think I originally created `AssertHandlerOnlySucceedsWithPublicationRoles` to handle instances where you needed to provide a AuthorizationHandler an argument, for `LegacyReleaseAuthorizationHandlerTests`, and then adapted it to handle passing a tuple to an AuthorizationHander for `UpdateReleaseRoleAuthorizationHandlerTests`.

duncan-at-hiveit: No worries, and yes, agree that the move towards builders is the way to go.  As far as this PR goes, happy that the other points are addressed thanks.

duncan-at-hiveit (approved):

ntsim (approved):

PR #3073, opened 2021-12-21
EES-2953 Add GeographicLevel field to Location (Location Id's change 1/3)

benoutram:
As an example to show why this is being added, the following CSV data would currently result in two Locations being created if they don't already exist:
 
 **CSV data** 

```
geographic_level    country_code    country_name    region_code    region_name    la_code    la_name
====================================================================================================
Local authority     E92000001       England         E12000004      East Midlands  E06000015  Derby
Local authority     E92000001       England         E12000004      East Midlands  E06000018  Nottingham
```

These both have principal attributes `LocalAuthority`, i.e. Derby and Nottingham but this is not obvious from the Locations alone which look as follows:

**Locations table**

```
country_code    country_name    region_code    region_name    la_code    la_name
===================================================================================
E92000001       England         E12000004      East Midlands  E06000015  Derby
E92000001       England         E12000004      East Midlands  E06000018  Nottingham
```

This PR adds a new, `GeographicLevel` field to `Location`.

In future when populated it will allow for the following benefits:

### Building meta data

When building subject meta data and table result subject meta data we currently get the level of each Observations and extract the principal attributes from its Location that corresponds with the level.

A GeographicLevel field on Location will allow simplifying extracting the principal attributes that are relevant to a set of Locations which can now be determined from the Location itself, without first needing the Geographic Level from the Observation.

### Querying Observations by Location Id and remove join to Location

Making Location records unique by GeographicLevel will remove the possibility that multiple Observations with different GeographicLevels can reference the same Location which is currently allowed. E.g. imagine if this data was supplied:

 **CSV data** 

```
geographic_level            country_code    country_name    region_code    region_name    la_code    la_name    lad_code    lad_name 
========================================================================================================================================
Local authority             E92000001       England         E12000004      East Midlands  E06000015  Derby      E05000003  Derbyshire
Local authority district    E92000001       England         E12000004      East Midlands  E06000015  Derby      E05000003  Derbyshire
```

**Locations table**

```
country_code    country_name    region_code    region_name    la_code    la_name    lad_code    lad_name 
=========================================================================================================
E92000001       England         E12000004      East Midlands  E06000015  Derby      E05000003  Derbyshire
```

In this scenario, there will be two Observations, one at LA level and the other at LAD which will both reference the single Location.

This means we can't currently query these Observations by Location Id alone as both the Observations for the LA and the LAD share the same Location. Both Observations would be returned.

We currently work around this limitation by applying a restriction on the Observations by GeographicLevel depending on the type of attribute being queried. This query is a highly abbreviated version of what we actually generate in dymaic SQL and execute but it shows the query by Location code and restriction on the GeographicLevel of the Observation:

```
SELECT o.Id FROM Observations o JOIN Locations l ON o.LocationId = l.Id
WHERE l.LocalAuthority_Code IN ('E06000015') AND o.GeographicLevel = 'LA'
```

In future when we can guarantee a Location is only used for one GeographicLevel we can query by Location Id:

```
SELECT o.Id FROM Observations o WHERE o.LocationId = '27bbc958-bcf8-49dd-b791-66a7b86a0627'
```

This will remove the need for the join on the Locations table when querying Observations.

It is required by future work EES-2778 which will switch from querying by Location codes to id's.

### Applying Location criteria to Footnotes

In future when we want to apply Location criteria to Footnotes this change will allow linking Footnotes to a specific Location e.g. a Local Authority by finding the Location with the Local Authority GeographicLevel that has the desired principal attribute.

Currently this wouldn't be possible as there could be many Locations returned.

### Procedure to copy Observation.GeographicLevel -> Location.GeographicLevel

This PR adds a new temporary stored procedure which copies `Location.GeographicLevel = Observation.GeographicLevel` for Locations referenced by Observations.
  * Skips any orphaned Locations with a warning message.
  * First checks that all Observations referencing a Location have a unique GeographicLevel.
  * Skips any Locations without a unique GeographicLevel with a warning message.
  * Locations are not updated if they already have a GeographicLevel making this script re-runnable.
  * Needs to be run in both the `statistics` and `public_statistics` databases.

This is planned to be executed manually, beginning in the Local developer database and the Dev Azure environment.

Future work by EES-2778 will change the importer to make use of this new field when determining whether to create or reuse Locations, and set the new value. It will also make the switch to querying by Location Id.

It also adds a new stored procedure for conveniently deleting orphaned locations. It's not possible to determine any level for these without any linked Observations so it's important that they are removed.

At least for now, GeographicLevel will be maintained on Observations as it is still used for querying legacy data blocks with location codes.

Finally, problematic Locations used by Observations with more than one value of GeographicLevel can be identified early using

```
    SELECT L.LocationId
    FROM (
             SELECT DISTINCT LocationId, GeographicLevel
             FROM Observation
         ) L
    GROUP BY L.LocationId
    HAVING COUNT(GeographicLevel) > 1;
```

It will be necessary to resolve any such Observations/Locations in order to fully complete the copy as these are otherwise skipped. A quick look at the Prod environment suggests there is a single Subject containing two Observations referencing the same Location, one is LA and the other is LAD.


src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_DeleteOrphanedLocations.sql
```diff
@@ -0,0 +1,12 @@
+﻿/**
+  Script which deletes Locations not referenced by any Observations.
+  Orphaned Locations occur when all Observations referencing them are deleted.
+ */
+CREATE OR
```
duncan-at-hiveit, line 5: Do we need a `Current_Routine_UpdateLocationGeographicLevel.sql` file as well?

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20211207175748_EES2776_AddGeographicLevelToLocation.cs
```diff
@@ -7,7 +7,9 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Model.Migrations
     public partial class EES2776_AddGeographicLevelToLocation : Migration
     {
         private const string MigrationId = "20211207175748";
-        
+        private const string PreviousLocationTypeMigrationId = "20210719084148";
```
duncan-at-hiveit, line 10: If we keep with this pattern of updating Types and Routines, and reference an old one to rollback, we should probably make our `MigrationId` public, and reference the previous class's actual `MigrationId` variable rather than repeat the string here.  That way, we can jump between source more easily.

duncan-at-hiveit (dismissed):

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_DeleteOrphanedLocations.sql
benoutram: Thanks, I didn't bother creating this because I plan to drop it as soon as the next part is deployed which depends on GeographicLevel existing and being populated so I think it will be fairly short lived.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20211207175748_EES2776_AddGeographicLevelToLocation.cs
benoutram: Thanks, good idea! I suppose we can leave the current one `private` until a rollback needs to refer to it.
I have made this change, making the previous migration id constant public, ands then referenced it in these two places.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_DeleteOrphanedLocations.sql
duncan-at-hiveit: Cool, I figured as much!

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20211207175748_EES2776_AddGeographicLevelToLocation.cs
duncan-at-hiveit: Nice one

duncan-at-hiveit (approved):

PR #3074, opened 2021-12-22
EES-2788 Add support for querying Observations by Location Id's (Location Id's change 2/3)

benoutram:
Episode two in a trilogy of PR's that will allow switching the way we query Observations from codes to Id's.

This adds support for querying Observations by Location Id's both to the stored procedure that exists in the database and the alternate `ObservationPredicateBuilder` which produces an Observation query without Filter arguments.

Rather than complicating the existing `FilteredObservations` stored procedure, this renames the existing stored procedure `SelectObservationsByLocationCodes` and  adds a new stored procedure `SelectObservations` with a LocationIds argument in place of codes. The method for selecting by Location codes (and its equivalent EF code in `ObservationPredicateBuilder`) will need to be maintained while we still query legacy data blocks containing Location codes.

This change depends on the previous PR https://github.com/dfe-analytical-services/explore-education-statistics/pull/3073 which adds a null field GeographicLevel to Location and has a temporary stored procedure to copy GeographicLevel from Observation.

This PR changes the Importer to set Location.Geographic level for all new CSV rows imported. It also changes the way we treat Locations to be unique by GeographicLevel in addition to all the other fields. Due to the addition of GeographicLevel to the fields we use to lookup existing Locations from the database and In-Memory cache, all existing Locations must have GeographicLevel populated by the time it is deployed. For this reason we will need to handle the deployment process carefully:

1. Deploy https://github.com/dfe-analytical-services/explore-education-statistics/pull/3073 which adds new null GeographicLocation field and the new stored procedures needed for copying the value and deleting orphaned Locations.
2. Identify and fix any Locations used by Observations with more than one value of GeographicLevel with the query mentioned by https://github.com/dfe-analytical-services/explore-education-statistics/pull/3073. This will be a case of deduplicating the Locations, adding GeographicLevel  to distinguish them based on the Observations using them and linking them to the Observations accordingly. At the time of writing in Prod there is 1 affected Location used by 2 Observations which need to be deduplicated into 2 Locations. Do this in both databases.
3. Execute database stored procedure `DeleteOrphanedLocations` also added by https://github.com/dfe-analytical-services/explore-education-statistics/pull/3073 in both Statistics databases to remove any Locations that are unused.
4. Execute  database stored procedure `UpdateLocationGeographicLevel` to copy the values from Observation to Location tables. A simple count on Locations without GeographicLevel can be used to check any Locations without a value. This is rerunnable so can be used to 'catch-up'.
5. Stop the Importer function app at a suitable time when no imports are in progress.
6. Check no new Location rows have been missed. Re-run `UpdateLocationGeographicLevel` if necessary.
7. Deploy this change.
8. Restart the Importer function app.

**A new `LocationsIds` query parameter is added to `ObservationQueryContext` and `SubjectMetaQueryContext` but no other changes are made to query by Location Id, yet. The switchover to start querying by LocationId will be a futher PR raised for EES-2955 which will change codes to Id's in the subject meta data and swap table requests to querying by Location Id's.**

### Other changes

* Fix to the Importer which only treat Locations unique by LA (Code,Name) rather than (Code,Old Code, Name). The location of an imported row would previously match an existing Location if the LA Code and Name were the same without checking the old code as the cache key string omitted the Old Code field.


src/GovUk.Education.ExploreEducationStatistics.Data.Model/Location.cs
```diff
@@ -11,8 +11,7 @@ public class Location
     {
         public Guid Id { get; set; }
 
-        // TODO EES-2776 SOW8 Make this non nullable when values have been migrated
-        public GeographicLevel? GeographicLevel { get; set; }
+        public GeographicLevel GeographicLevel { get; set; }
```
duncan-at-hiveit, line 14: I guess in order for this to happen, we need to have decided how to deal with that case on Prod whereby 2 Observation rows reference the same Location row but for different GeographicLevels?

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/LocationRepository.cs
```diff
@@ -86,49 +62,36 @@ public LocationRepository(StatisticsDbContext context)
                 ? new Dictionary<GeographicLevel, List<Func<Location, ILocationAttribute>>>()
                 : MapLocationAttributeSelectors(hierarchies);
 
-            var locationsWithGeographicLevels = observations
+            var locationIds = observations
```
duncan-at-hiveit, line 65: Can this instead now be:

```
var locations = await observations
    .AsNoTracking()
    .Select(observation => observation.Location)
    .Distinct()
    .ToListAsync();
```

or 

```
var locations = await observations
    .AsNoTracking()
    .Include(observation => observation.Location)
    .Select(observation => observation.Location)
    .Distinct()
    .ToListAsync();
```

instead of firstly obtaining the ids and then selecting locations by those ids?  Don't know if one is faster than the other?

src/GovUk.Education.ExploreEducationStatistics.Data.Model/LocalAuthority.cs
```diff
@@ -12,6 +12,11 @@ public LocalAuthority(string? code, string? oldCode, string? name) : base(code,
             OldCode = oldCode;
         }
 
+        public string GetCacheKey()
+        {
+            return $"{GetType().Name}:{Code ?? string.Empty}:{OldCode ?? string.Empty}:{Name ?? string.Empty}";
```
duncan-at-hiveit, line 17: Why do we not use `GetCodeOrFallback()` here?  Are you explicitly wanting to add to the cache key the fact that the `Code` isn't present?

duncan-at-hiveit: If so, might be worth a comment here to explain why.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
```diff
@@ -250,16 +251,19 @@ int rowsPerBatch
...
-                    subjectMeta,
-                    ((batchNo - 1) * rowsPerBatch) + i++ + 2);
-
-                if (hasSoloAllowedGeographicLevel || AllowRowImport(o.GeographicLevel))
+                var rowValues = CsvUtil.GetRowValues(row).ToArray();
+                var geographicLevel = GetGeographicLevel(rowValues, colValues);
+
+                if (hasSoloAllowedGeographicLevel || AllowRowImport(geographicLevel))
                 {
+                    var o = ObservationFromCsv(
```
duncan-at-hiveit, line 259: Nice to only have to execute this when we need to now :)

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
```diff
@@ -281,7 +286,8 @@ int rowsPerBatch
             {
                 Id = observationId,
                 FilterItems = GetFilterItems(context, rowValues, colValues, subjectMeta.Filters, observationId),
-                GeographicLevel = GetGeographicLevel(rowValues, colValues),
+                // TODO should be able to drop GeographicLevel in future once the migration to Location is complete
```
duncan-at-hiveit: Is it worth bumping a ticket ref on this if we have one available?

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
```diff
@@ -505,6 +512,7 @@ private async Task InsertObservations(DbContext context, IEnumerable<Observation
                 observationsTable.Rows.Add(
                     o.Id,
                     o.SubjectId,
+                    // TODO should be able to drop GeographicLevel in future once the migration to Location is complete
```
duncan-at-hiveit: Ticket ref comment again

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/LocationViewModel.cs
```diff
@@ -1,23 +1,25 @@
...
-        public CodeNameViewModel OpportunityArea { get; set; }
-        public CodeNameViewModel ParliamentaryConstituency { get; set; }
-        public CodeNameViewModel Provider { get; set; }
-        public CodeNameViewModel Region { get; set; }
-        public CodeNameViewModel RscRegion { get; set; }
-        public CodeNameViewModel School { get; set; }
-        public CodeNameViewModel Sponsor { get; set; }
-        public CodeNameViewModel Ward { get; set; }
-        public CodeNameViewModel PlanningArea { get; set; }
+        public CodeNameViewModel? Country { get; init; }
```
duncan-at-hiveit, line 7: Just double-check that we're not caching this anywhere that might complain if it can't find setters when trying to deserialise :)

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/Meta/ObservationalUnitMetaViewModel.cs
```diff
@@ -11,7 +11,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Services.ViewModels.Me
     /// </summary>
     public record ObservationalUnitMetaViewModel : LabelValue
     {
-        [JsonConverter(typeof(StringEnumConverter), true)]
+        [JsonConverter(typeof(StringEnumConverter), typeof(CamelCaseNamingStrategy))]
```
duncan-at-hiveit, line 14: What's the camelcase strategy for?

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Location.cs
benoutram: I was planning to just sort it manually . Here's what the Observation rows and Location row currently look like.

(Fictitious values, abbreviated columns names, etc.)

**Observation:**

```
Id      SubjectId  LocationId   GeographicLevel
===============================================
obs1    sub1       loc1         LA
obs2    sub1       loc1         LAD
```

**Location:**

New column `Location.GeorgraphicLevel` will have been added by this point from #3073.

```
Id    GeographicLevel   Country_Code    Country_Name    Region_Code    Region_Name    La_Code    La_Name    Lad_Code   Lad_Name
=================================================================================================================================
loc1  <null>            E92000001       England         E12000004      East Midlands  E06000015  Derby      E05000003  Derbyshire
```

We can just fix it with something like this:

```
INSERT INTO LOCATION (Id,GeographicLevel,Country_Code,Country_Name,LocalAuthority_Code,LocalAuthority_Name,LocalAuthorityDistrict_Code,LocalAuthorityDistrict_Name)
VALUES('loc2','LAD','E92000001','England','E12000004','East Midlands','E06000015','Derby','E05000003','Derbyshire');
`UPDATE Observation SET LocationId = 'loc2' WHERE Id = 'obs2';`
`UPDATE Location SET GeographicLevel = 'LA' Where Id = 'loc1';`
```

Re-running the stored procedure to copy the values across from Observation to Location to 'catch-up' would also have the same effect as the last `UPDATE` statement.

Ideally we will do all this and get rid of the uniqueness problem before this PR is merged so that when it is the querying and location lookup in the Importer works reliably for every possible case.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/LocalAuthority.cs
benoutram: Yes, and because we need a separate code and old code token inside the cache key, rather than using `GetCodeOrFallback` to differentiate locations with the same code and name but different old code.

Imagine we had `$"{GetType().Name}:{GetCodeOrFallback()}:{Name ?? string.Empty}";` or `$"{GetType().Name}:{GetCodeOrOldCodeIfEmpty()}:{Name ?? string.Empty}";`

Two imported LA rows with the same Code, and Name but different old codes will hit the same cache entry as the old code will be ignored.

Unit test `ImporterLocationServiceTests.FindOrCreate_UniqueByLocalAuthorityOldCode` added by this PR now covers this case.

Prior to this PR this exists as a bug in Prod, see the current code on line 135:

https://github.com/dfe-analytical-services/explore-education-statistics/blob/1090a7b9438aeb6f6f69934097b40c8ef1a05b6c/src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterLocationService.cs#L135

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/Meta/ObservationalUnitMetaViewModel.cs
benoutram: Yes I think it's to meet the required case of the `GeographicLevel` field expected by the front end.

The reason for this change though is as follows:

The second parameter of the `JsonConverterAttribute` is the parameter list to use when constructing the converter, in this case a `StringEnumConverter`.

The constructor overload for `bool` is obsolete. So instead we are just updating it use a different constructor.
image

The result should be exactly the same behaviour. I made this change in one other place as well.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/LocalAuthority.cs
duncan-at-hiveit: Brill thanks.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/Meta/ObservationalUnitMetaViewModel.cs
duncan-at-hiveit: Nice one, thanks!

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/LocationViewModel.cs
duncan-at-hiveit: An example area would be Admin's TableBuilderController, with:

```
        [BlobCache(typeof(DataBlockTableResultCacheKey))]
        private async Task<Either<ActionResult, TableBuilderResultViewModel>> GetReleaseDataBlockResults(
```

duncan-at-hiveit: And I'm guessing that if this had been a problem, then it should have been highlighted in `TableBuilderControllerTests.TableBuilderResultViewModel_SerialiseAndDeserialise()`, so we should be good?

duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Location.cs
benoutram: This has been dealt with by EES-3010 now.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/LocalAuthority.cs
benoutram: Comment added in EES-3011.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
benoutram: I removed this TODO as plan to keep GeographicLevel on Observation for now but raised EES-3067 and EES-3068 and left them as comments in the appropriate place as part of EES-3011.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
benoutram: I removed this TODO as plan to keep GeographicLevel on Observation for now but raised EES-3067 and EES-3068 and left them as comments in the appropriate place as part of EES-3011.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/LocationViewModel.cs
benoutram: Did a check by altering the data and it seems to be fine. Thanks for pointing it out and checking!

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/LocationRepository.cs
benoutram: I checked both suggestions and they both translate to exactly the same SQL.

I've tidied this up a bit to make it a bit more readable:

```
SELECT DISTINCT
                l.Id,
                l.Country_Code,
                l.Country_Name,
                l.EnglishDevolvedArea_Code,
                l.EnglishDevolvedArea_Name,
                l.GeographicLevel,
                l.Institution_Code,
                l.Institution_Name,
                l.LocalAuthorityDistrict_Code,
                l.LocalAuthorityDistrict_Name,
                l.LocalAuthority_Code,
                l.LocalAuthority_Name,
                l.LocalAuthority_OldCode,
                l.LocalEnterprisePartnership_Code,
                l.LocalEnterprisePartnership_Name,
                l.MayoralCombinedAuthority_Code,
                l.MayoralCombinedAuthority_Name,
                l.MultiAcademyTrust_Code,
                l.MultiAcademyTrust_Name,
                l.OpportunityArea_Code,
                l.OpportunityArea_Name,
                l.ParliamentaryConstituency_Code,
                l.ParliamentaryConstituency_Name,
                l.PlanningArea_Code,
                l.PlanningArea_Name,
                l.Provider_Code,
                l.Provider_Name,
                l.Region_Code,
                l.Region_Name,
                l.RscRegion_Code,
                l.School_Code,
                l.School_Name,
                l.Sponsor_Code,
                l.Sponsor_Name,
                l.Ward_Code,
                l.Ward_Name
FROM Observation AS o
         INNER JOIN Location AS l ON o.LocationId = l.Id
WHERE o.SubjectId = '<Subject Id>'
```

I've gone for the option without the `.Include`.

I can't see exactly where in the docs the `.Select` overrides the need for the `.Include` but the tests in `LocationRepositoryTests` should fail if we needed it and they don't.

I checked the performance in my local database when compared with the old two stage approach and found this was slightly faster. With the Subject with the largest number of rows in Prod I didn't find any performance problem.

I could test this further but I think we agree that the JOIN is going to be faster especially where there are a large number of Locations.

PR #3081, opened 2021-12-22
EES-3006 - fix UI test startup issues

luke-h1:
This PR: 
- Fixes UI test startup issues related to `run_tests.py` referencing a CLI flag that no longer exists 
- Adds additional asserts to make sure the correct `.env` variables are set

### DIFF
--- a/tests/robot-tests/run_tests.py
+++ b/tests/robot-tests/run_tests.py
@@ -167,6 +167,9 @@ os.environ["PATH"] += os.pathsep + str(Path('webdriver').absolute())
 
 output_file = "rerun.xml" if args.rerun_failed_tests or args.rerun_failed_suites else "output.xml"
 
+assert os.getenv('TIMEOUT') is not None
+assert os.getenv('IMPLICIT_WAIT') is not None
+
 # Set robotArgs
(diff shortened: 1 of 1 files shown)


tests/robot-tests/run_tests.py
```diff
@@ -208,7 +208,15 @@
...
 assert os.getenv('ADMIN_PASSWORD') is not None
-
+assert os.getenv('PUBLIC_AUTH_USER') is not None
+assert os.getenv('PUBLIC_AUTH_PASSWORD') is not None
+assert os.getenv('RELEASE_COMPLETE_WAIT') is not None
+assert os.getenv('WAIT_MEDIUM') is not None
+assert os.getenv('WAIT_LONG') is not None
+assert os.getenv('WAIT_SMALL') is not None
+assert os.getenv('TIMEOUT') is not None
+assert os.getenv('IMPLICIT_WAIT') is not None
```
ntsim, line 218: These are used directly above, so it's too late to be asserting that they exist here. We should assert these _before_ they're used

ntsim (changes requested):

ntsim (approved):

PR #3082, opened 2021-12-22
EES-2777 Add location hierarchies to tables

ntsim:
### :warning: This will require the data blocks caches to be cleared in each environment. 

The cached data blocks are backwards incompatible and will cause the frontend to throw errors after this PR is merged.

---

This PR adds location hierarchies to tables generated by table tool.

image

We have primarily enabled the backend `TableResultLocationHierarchiesEnabled` option, meaning that the `LocationsHierarchical` field is populated instead of `Locations` in the subject meta of table results.

## Relevant changes

- Fixed location deduplication logic not being applied to table result locations from the API. This was causing duplicate locations to appear in the table tool.
- Orders locations in the table result subject meta so that regions are ordered by code. Non-region locations are ordered by their label/name. As part of this, we have now removed any frontend re-ordering of filters when constructing table headers. This means the initial table headers will be in the same order they appear in the subject meta.
- Location `level` fields in subject meta are now all camelCased to be consistent with their usual level keys. This is important to ensure we can correctly match locations.

## Other changes

- Cleaned up and refactored parts of the frontend location deduplication code to make it more compatible with location hierarchies. We've avoided refactoring this _too_ much as this will hopefully be removed as part of the location ID work.


src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/ResultSubjectMetaServiceTests.cs
```diff
@@ -1828,6 +2202,237 @@ await using (var statisticsDbContext = InMemoryStatisticsDbContext(statisticsDbC
...
+                Assert.Equal(_northWest.Code, laOption2.Value);
+                Assert.Equal("region", laOption2.Level);
+                Assert.Single(laOption2.Options!);
+
+                var laOption3 = localAuthorities[2];
+                Assert.NotNull(laOption3);
+                Assert.Equal(_eastMidlands.Name, laOption3.Label);
+                Assert.Equal(_eastMidlands.Code, laOption3.Value);
+                Assert.Equal("region", laOption3.Level);
+                Assert.Single(laOption2.Options!);
```
benoutram, line 2432: There's a mistake here

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/ResultSubjectMetaServiceTests.cs
```diff
@@ -1678,6 +1681,377 @@ await using (var statisticsDbContext = InMemoryStatisticsDbContext(statisticsDbC
...
+                        new(_derby),
+                        new(_derbyDupe),
+                        new(_nottingham)
+                    }
+                }
+            };
+
+            // Hierarchies are enabled, but none have been defined
+            // so locations should still all be flat.
+            var options = Options.Create(new LocationsOptions
```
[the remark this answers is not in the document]
ntsim: Good point, but yeh thought it'd be a little clearer to be explicit in this case. I've moved the comment directly next to the `TableResultLocationHierarchiesEnabled` so that it's more obvious we need to change it when we remove the flag later on.

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/ResultSubjectMetaServiceTests.cs
ntsim: Great spot, fixed! :smile:

Yeh there's a few things that I needed to do. The location de-duping stuff on the frontend was the most pain tbh. It'll be good when we can remove it all in EES-2783.

benoutram: That will be great when we can remove it all! 🤞 

I can't see this fix yet btw. 🙂

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/ResultSubjectMetaServiceTests.cs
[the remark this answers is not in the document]
benoutram: Great, I see the change now, thanks

benoutram (approved):

PR #3086, opened 2022-01-04
EES-3007 Resolve some backend warnings

markhiveit:
Along with resolving some backend warnings, this ticket also removes `#nullable enable` from `User.cs`. Currently, if you create a migration, it will try to change the user table. The solution to this doesn't seem trivial so for now I'll remove it.


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ThemeServicePermissionTests.cs
```diff
@@ -42,7 +42,7 @@ await using (var context = DbUtils.InMemoryApplicationDbContext(contextId))
                 await context.SaveChangesAsync();
             }
 
-            PolicyCheckBuilder()
+            await PolicyCheckBuilder()
```
duncan-at-hiveit, line 45: Nice one - these ones creep in from time to time - hopefully if we can eliminate all warnings from the codebase then we'll just be left with legitimate ones so we can spot these appearing.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ThemeServicePermissionTests.cs
```diff
@@ -225,6 +225,15 @@ public void DeleteTheme()
                 );
         }
 
+        private static Mock<IConfiguration> DefaultConfigurationMock()
```
duncan-at-hiveit, line 228: We actually already have a `MockUtils.CreateMockConfiguration()` method that you can use with no parameters.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseImageServicePermissionTests.cs
```diff
@@ -47,8 +47,10 @@ await PolicyCheckBuilder<SecurityPolicies>()
             IReleaseFileRepository releaseFileRepository = null,
             IUserService userService = null)
         {
+            contentDbContext ??= new Mock<ContentDbContext>().Object;
+
             return new ReleaseImageService(
-                contentDbContext ?? new Mock<ContentDbContext>().Object,
+                contentDbContext,
```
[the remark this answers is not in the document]
markhiveit: This was to remove a warning on line 57 about contentDbContext potentially being null: `releaseFileRepository ?? new ReleaseFileRepository(contentDbContext),`.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ThemeServicePermissionTests.cs
markhiveit: I get a failing test if I use this, as `CreateMockConfiguration` uses `MockBehavior.Strict`.

`Moq.MockException: IConfiguration.GetSection("enableThemeDeletion") invocation failed with mock behavior Strict.`

I think I may have tried this and switched to creating DefaultConfigurationMock instead as I think `enableThemeDeletion` isn't a `Tuple<string, string>` but a `Tuple<string, bool>` and I think that was causing me issues.

duncan-at-hiveit: Ahh fair enough!

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseImageServicePermissionTests.cs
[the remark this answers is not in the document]
duncan-at-hiveit: Ah gotcha.  

I'd just change 

`releaseFileRepository ?? new ReleaseFileRepository(contentDbContext)`

to be:

`releaseFileRepository ?? Mock.Of<IReleaseFileRepository>()`

instead, in line with the other optional parameters.  There's no need in this test for it to be a real Repo.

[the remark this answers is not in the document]
duncan-at-hiveit: It also gets away from re-assigning parameters (in this case contentDbContext), which is a habit that I'd definitely encourage wherever possible.  In immutable languages, this wouldn't be allowed :)

duncan-at-hiveit (dismissed):

src/GovUk.Education.ExploreEducationStatistics.Content.Model/GlossaryEntry.cs
```diff
@@ -8,16 +7,16 @@ public class GlossaryEntry
...
-        public string Body { get; set; }
+        public string Body { get; set; } = string.Empty;
 
-        public DateTime Created { get; set; }
+        public DateTime Created { get; set; } = new DateTime();
 
         public Guid CreatedById { get; set; }
 
-        public User CreatedBy { get; set; }
+        public User CreatedBy { get; set; } = new User();
```
benoutram, line 20: Pretty sure you'd never want this to be assigned to a new User. If you ended up saving a new GlossaryEntry like this you could potentially also end up creating a User as part of the same save.

I'd use `public User CreatedBy { get; set; } = null!;` instead.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/GlossaryEntry.cs
```diff
@@ -8,16 +7,16 @@ public class GlossaryEntry
...
+        public string Title { get; set; } = string.Empty;
 
-        public string Slug { get; set; }
+        public string Slug { get; set; } = string.Empty;
 
-        public string Body { get; set; }
+        public string Body { get; set; } = string.Empty;
 
-        public DateTime Created { get; set; }
+        public DateTime Created { get; set; } = new DateTime();
```
benoutram, line 16: Is that a warning without it?

We initialise all our times to `DateTime.UtcNow` and what you have wouldn't be a UTC time if somebody accidently forgets to replace it.

markhiveit: I cannot remember 100%, but it seems like I added this to remove a `Created.set is never used` warning - although I'm not sure if I agree with my past self that that was the way to do it.

I'll restore the line to how it was for now.

benoutram (approved):

PR #3087, opened 2022-01-04
EES-3065 Drop the temporary stored procedure which was used to set values of Location.GeographicLevel

benoutram:
This PR drops the `UpdateLocationGeographicLevel` stored procedure that was added temporarily by https://github.com/dfe-analytical-services/explore-education-statistics/pull/3073 and used to set values of Location.GeographicLevel.

This was required to be run before https://github.com/dfe-analytical-services/explore-education-statistics/pull/3074 is deployed which relies on the values being set.

Marking this PR as 'Do Not Merge' until the procedure has been run in all environments successfully.


src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20220104093335_EES3065_DropRoutineUpateLocationGeographicLevel.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace GovUk.Education.ExploreEducationStatistics.Data.Model.Migrations
+{
+    [ExcludeFromCodeCoverage]
+    public partial class EES3065_DropRoutineUpateLocationGeographicLevel : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            // Drop the temporary procedure which was used to set Location.GeographicLevel = Observation.GeographicLevel
+            // for Locations referenced by Observations.
+            migrationBuilder.Sql("DROP PROCEDURE dbo.UpdateLocationGeographicLevel");
```
duncan-at-hiveit, line 15: Do we need to keep the deleteorphanedlocations for the time being, or can that also be dropped?

I'm guessing that we should keep it as it's probably pretty costly to check for orphaned locations during a subject deletion - maybe we need a ticket to put in a scheduled job to run this procedure nightly, but not in a way that it would interfere with any ongoing imports?  Hmmm...

benoutram: I've kept `DeleteOrphanedLocations` for now as I think it's a handy procedure that we can at least run manually periodically.

We have some confidence now that it's safe to run after testing which we might not have straight away if we write a new delete statement to manually run in future.

I agree with what you are saying about the pitfalls of removing orphaned locations during subject deletion and it feels like its possible that it could cause problems if importing is taking place and the Importer has any of those Locations in its In-Memory cache which then become candidates to be linked to new Observations.

That's the same reason why I've not made a change up until now to automate running it on a scheduled basis but it probably would be ok to run, say weekly in the middle of the night? I've raised this as https://dfedigital.atlassian.net/browse/EES-3070 now.

Thanks.

duncan-at-hiveit: Yeah that sounds good.  I think in order to avoid interfering with Releases that are getting published at midnight, it'd have to be automated at a time where we're happy that no imports are likely to be ongoing.  Really we should have some sort of mutual exclusivity check going on that doesn't let the 2 processes operate in parallel I suppose.

duncan-at-hiveit (approved):

PR #3090, opened 2022-01-05
EES-3073 Constrain Node and NPM versions to 16

ntsim:
This PR adds Node and NPM version checking to avoid installation issues that can arise due to different major versions. We have pinned the current Node version at 16 (the latest LTS) and NPM at v8 (bundled with Node 16).

We have implemented this via the `check-node-version` package, allowing us to check this in any install scripts, including the `bootstrap` ones.

This is a better solution compared to specifying `engines` in `package.json` as we can more thoroughly enforce the version. The `engines` approach unfortunately has really basic flaws that allow users to easily circumvent the engine check (e.g. running `npm ci` instead of `npm i`).

## Other changes

- Added `.nvmrc` to make it more convenient for NVM users to switch to the correct version of Node.
- Upgraded Husky as it was previously incompatible with Node 16.


luke-h1:
Might want to bump the version number in the root README if we only intend to support 16+

ntsim:
Thanks! Had done, but hadn't pushed the updated README (its done now). 

I've not updated any of the infrastructure as I'm not very confident that Azure supports Node 16 currently. Didn't want to get stuck in infrastructure re-deploy hell trying to figure out if it does or not. 

I'm fairly sure the changes should still work with these new changes. We know that the frontend builds currently run in the production environments despite the `package-lock.json` changes, and I haven't made any changes that directly impact the running of the apps (I've only added the checks on installation steps).

luke-h1 (approved):

PR #3091, opened 2022-01-05
EES-3012 Add BAU endpoints to clear blob caches

ntsim:
This PR adds the following new endpoints:

- `DELETE /api/bau/private-cache` - clears the entire private cache used by admin
- `DELETE /api/bau/public-cache/trees` - clears specific tree JSON files from the public cache
- `DELETE /api/bau/public-cache/releases` - clears specific paths (e.g. `data-blocks`, `subject-meta`, etc) related to releases from the public cache

It should be noted that all of these endpoints only clear blobs where it is safe to do so i.e. where the cached object can be regenerated easily by a user visiting the relevant part of the app. This means that we don't delete blobs like the release content JSON as these are only generated by the publisher and cannot be regenerated without regenerating all of the content (this is an expensive operation).

Ideally, we would rework the remaining published content blobs so that they use the cache-aside strategy used in other places. This would allow us to safely delete any part of the blob caches without needing to regenerate all of the content.

## Relevant changes

- Updated the `BlobStorageService.DeleteBlobs` method to accept regexes for including and excluding blobs from the deletion. This is necessary for the kind of behaviour we need for this PR's endpoints.
- Add a new `ContainsOnly` validation attribute that can be used to validate that the validated field only contains specific allowed values.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/BauCacheController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/BauCacheController.cs
@@ -1,0 +1,113 @@
+#nullable enable
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using GovUk.Education.ExploreEducationStatistics.Admin.Models;
+using GovUk.Education.ExploreEducationStatistics.Common;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Common.Services;
+using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces;
+using GovUk.Education.ExploreEducationStatistics.Content.Services.Cache;
+using GovUk.Education.ExploreEducationStatistics.Content.Services.Requests;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using static GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces.IBlobStorageService;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api.Bau
+{
+    [Route("api/bau")]
+    [ApiController]
+    [Authorize(Roles = RoleNames.BauUser)]
+    public class BauCacheController : ControllerBase
+    {
+        private static readonly HashSet<string> AllowedPublicReleasePaths = new HashSet<string>
+        {
+            FileStoragePathUtils.PublicContentDataBlocksDirectory,
+            // TODO: EES-2865 Remove with other fast track code
+            FileStoragePathUtils.PublicContentFastTrackResultsDirectory,
+            FileStoragePathUtils.PublicContentSubjectMetaDirectory
+        };
+
+        private static readonly HashSet<string> AllowedPublicTreePaths = new HashSet<string>
+        {
+            PublicationTreeCacheKey.GetKey(PublicationTreeFilter.AnyData),
+            PublicationTreeCacheKey.GetKey(PublicationTreeFilter.LatestData),
+            PublicationTreeCacheKey.GetKey(null)
+        };
+
+        private readonly IBlobStorageService _privateBlobStorageService;
+        private readonly IBlobStorageService _publicBlobStorageService;
+
+        public BauCacheController(
+            IBlobStorageService privateBlobStorageService,
+            IBlobStorageService publicBlobStorageService)
+        {
+            _privateBlobStorageService = privateBlobStorageService;
+            _publicBlobStorageService = publicBlobStorageService;
+        }
+
+        [HttpDelete("private-cache")]
+        public async Task<ActionResult> ClearPrivateCache()
+        {
+            // Remove all cache entries as we currently don't really need
+            // any more granularity currently (we only cache data blocks).
+            await _privateBlobStorageService.DeleteBlobs(BlobContainers.PrivateContent);
+
+            return NoContent();
+        }
+
+        [HttpDelete("public-cache/trees")]
+        public async Task<ActionResult> ClearPublicCacheTrees(ClearCachePathsViewModel request)
+        {
+            var paths = request.Paths
+                .Where(AllowedPublicTreePaths.Contains)
+                .ToHashSet();
+
+            if (paths.Any())
+            {
+                await paths
+                    .ToAsyncEnumerable()
+                    .ForEachAwaitAsync(
+                        path =>
+                            _publicBlobStorageService.DeleteBlob(BlobContainers.PublicContent, path)
+                    );
+            }
+
+            return NoContent();
+        }
+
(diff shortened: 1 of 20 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/BauCacheController.cs
```diff
@@ -0,0 +1,116 @@
...
+                            _publicBlobStorageService.DeleteBlob(BlobContainers.PublicContent, path)
+                    );
+            }
+
+            return NoContent();
+        }
+
+        [HttpDelete("public-cache/releases")]
+        [Authorize(Roles = RoleNames.BauUser)]
+        public async Task<ActionResult> ClearPublicCacheReleases(ClearCachePathsViewModel request)
```
markhiveit, line 85: Not suggesting a change here, but I think I would have been inclined to create more endpoints - one for data blocks, fast-tracks, etc. - rather than having the user provide the provides the directory names. It's a minor point, but my thought is that if someone unfamiliar to these endpoints comes to use them, their intention will be to just "remove all cached data blocks", and they could simply ping the right endpoint. 

As it stands, if you want to deleted all cached data blocks, it's not starkly obvious which endpoint you want. They will have to dig more deeply into this controller to understand which endpoint they want and what string to provide to delete data blocks.

That said, I accept that if you wanted to delete lots of different things - tree files / data blocks / etc. - and you already know how these endpoints work, you're potentially pinging quite a few endpoints if it was done the way I'm suggesting.

src/GovUk.Education.ExploreEducationStatistics.Common/Services/FileStoragePathUtils.cs
```diff
@@ -7,6 +7,11 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Services
 {
     public static class FileStoragePathUtils
     {
+        public const string PublicContentDataBlocksDirectory = "data-blocks";
```
markhiveit, line 10: Now we have this, you can use it in PrivateContentDataBlockParentPath:
```
        public static string PrivateContentDataBlockParentPath(
            Guid publicationId,
            Guid releaseId)
        {
            return $"{PrivateContentReleaseParentPath(publicationId, releaseId)}/data-blocks";
        }
```

markhiveit (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
```diff
@@ -237,6 +238,7 @@ public void ConfigureServices(IServiceCollection services)
                     options.EnableEndpointRouting = false;
                     options.AllowEmptyInputInBodyModelBinding = true;
                 })
+                .AddControllersAsServices()
```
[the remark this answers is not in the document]
ntsim: Not sure that's true. .NET doesn't have a way to mark a class as a service, so it'd be pretty silly for it to mandate that any class that isn't registered with the container would fail startup (e.g. what about model-like classes).

Anyway, I originally added this because I needed to manually wire specific dependencies into the `BauCacheController`. However, I've just checked and it doesn't seem like adding this actually makes any difference to how the DI works. Have removed it now!

src/GovUk.Education.ExploreEducationStatistics.Common/Services/FileStoragePathUtils.cs
ntsim: Done!

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/BauCacheController.cs
```diff
@@ -0,0 +1,116 @@
...
+            // TODO: EES-2865 Remove with other fast track code
+            FileStoragePathUtils.PublicContentFastTrackResultsDirectory,
+            FileStoragePathUtils.PublicContentSubjectMetaDirectory
+        };
+
+        private static readonly HashSet<string> AllowedPublicTreePaths = new HashSet<string>
+        {
+            PublicationTreeCacheKey.GetKey(PublicationTreeFilter.AnyData),
+            PublicationTreeCacheKey.GetKey(PublicationTreeFilter.LatestData),
+            PublicationTreeCacheKey.GetKey(null)
```
benoutram, line 37: This doesn't look right to me
`PublicationTreeCacheKey.GetKey(null)`

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/BauCacheController.cs
```diff
@@ -0,0 +1,116 @@
...
+    {
+        private static readonly HashSet<string> AllowedPublicReleasePaths = new HashSet<string>
+        {
+            FileStoragePathUtils.PublicContentDataBlocksDirectory,
+            // TODO: EES-2865 Remove with other fast track code
+            FileStoragePathUtils.PublicContentFastTrackResultsDirectory,
+            FileStoragePathUtils.PublicContentSubjectMetaDirectory
+        };
+
+        private static readonly HashSet<string> AllowedPublicTreePaths = new HashSet<string>
```
benoutram, line 33: I think it's safe to add the Methodology tree cache key to this list as `methodology-tree.json` is cached by the new cache-aside strategy after its generation was moved away from the Publisher.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/BauCacheController.cs
benoutram: Ignore me, I see it's an explict way of requesting the key for the other variation `publication-tree.json`

markhiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/BauCacheController.cs
ntsim: Added :+1:

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/BauCacheController.cs
ntsim: To make this a little more palatable/clearer, it no longer passes in null. Instead, we now use a default arg from the key's constructor.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/BauCacheController.cs
ntsim: Decided against changing it, but to improve the UX around this endpoint, I've spent some time adding a new `ContainsOnly` attribute and added `MinLength(1)`. This should make it more obvious what values are allowed in the `paths` field.

The `ContainsOnly` attribute should be fairly useful for other use-cases as well hopefully.

benoutram (approved):

PR #3092, opened 2022-01-05
EES-2913 Add Pending invites to Manage team access page

markhiveit:
This PR adds pending contributor invites for releases to the Manage team access page.

image

This includes implementing the ability to "Cancel invite", which removes all release invites for that email address for the publication.


src/explore-education-statistics-admin/src/pages/publication/components/PublicationManageTeamAccessTab.tsx
```diff
@@ -43,14 +63,17 @@ const PublicationManageTeamAccessTab = ({ publication, release }: Props) => {
   return (
     <>
       <h2>Update access for release ({release.title})</h2>
-      {!contributors.length ? (
+      {!contributors.length && !pendingInviteEmails.length ? (
```
ntsim, line 66: It's usually a good idea to use numeric comparisons on `length` instead relying on a falsy check. 

We typically use `&&` to conditionally render fragments and when combined with falsy `length` checks, this has previously meant that random 0's would appear in the UI when the check is falsy.

As we're using a ternary, it's not such a big deal, but it's probably better to avoid that possibility in the future (maybe due to refactoring).

src/explore-education-statistics-admin/src/pages/publication/components/__tests__/ReleaseContributorPermissions.test.tsx
```diff
@@ -50,22 +60,28 @@ describe('ReleaseContributorPermissions', () => {
       expect(
         screen.getByTestId('releaseContributors-warning').textContent,
       ).toContain(
-        'There are currently no team members associated with this publication.',
+        'There are currently no team members or pending invites associated with this publication.',
       );
     });
```
ntsim, line 65: We probably want an additional assertion that the table does not render.

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleasePermissionService.cs
```diff
@@ -39,36 +39,49 @@ public class ReleasePermissionService : IReleasePermissionService
             _userService = userService;
         }
 
-        public async Task<Either<ActionResult, List<ContributorViewModel>>>
-            ListReleaseContributors(Guid releaseId)
+        public async Task<Either<ActionResult, ContributorsAndInvitesViewModel>>
+            ListReleaseContributorsAndContributorInvites(Guid releaseId)
```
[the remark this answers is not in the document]
markhiveit: For some reason my mind didn't twig to do two endpoints. My original stab at this feature involved getting invites as part of the original method, and then I refactored to get the contributors and contributor invites separately, but didn't go on to follow our convention and separating them out as entirely separate endpoints.

I am slowly learning how you'd like things done. But I you're probably going to need more patience with me. 😄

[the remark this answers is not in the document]
markhiveit: Done

src/explore-education-statistics-admin/src/pages/publication/components/__tests__/ReleaseContributorPermissions.test.tsx
```diff
@@ -134,12 +148,61 @@ describe('ReleaseContributorPermissions', () => {
...
+      <ReleaseContributorsPermissions
+        contributors={[]}
+        invites={testInvites}
+        onUserRemove={onUserRemove}
+        onUserInvitesRemove={onUserInvitesRemove}
+      />,
+    );
+
+    await waitFor(() => {
+      expect(screen.getByRole('table')).toBeInTheDocument();
```
luke-h1, line 176: I think we should be using `findByRole` here instead of wrapping this assertion in a `waitFor`. i.e. 
``` 
expect(screen.findByRole('table').toBeInTheDocument()
```

We should avoid using getByRole inside a waitFor as it can cause tests to fail due to getBByRole being slow

https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-waitfor-to-wait-for-elements-that-can-be-queried-with-find

ntsim: In general, I've generally found that the `find*` selectors rarely work properly so my recommendation is to still prefer using `waitFor`.

Also, mentioned this on Slack a while back, but we shouldn't be using `getByRole` (or its variants) within a `waitFor` anyway. It's an expensive assertion which can block the `waitFor` from finishing when used on larger components, potentially causing intermittent test failures.

The recommendation is to use exclusively use more performant selectors `getByText` or `getByTestId` in `waitFor` e.g.:

```ts
await waitFor(() => {
  expect(screen.getByText('something that should appear')).toBeInTheDocument();
});

// Now it's fine to use `getByRole` outside of `waitFor`
expect(screen.geyByRole('table')).toBeInTheDocument()
```

markhiveit: Done

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleasePermissionsController.cs
```diff
@@ -24,14 +25,23 @@ public class ReleasePermissionsController : ControllerBase
         }
 
         [HttpGet("releases/{releaseId}/contributors")]
-        public async Task<ActionResult<List<ContributorViewModel>>> ListReleaseContributors(
+        public async Task<ActionResult<List<ContributorViewModel>>> ListReleaseContributorsAndContributorInvites(
```
ntsim, line 28: This name looks wrong

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseInviteService.cs
```diff
@@ -78,6 +78,23 @@ public class ReleaseInviteService : IReleaseInviteService
                 });
         }
 
+        public async Task<Either<ActionResult, Unit>> RemoveByPublication(
```
ntsim, line 81: Doesn't seem like we've added any tests for this?

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseInviteService.cs
markhiveit: Missed this - my bad. Added some tests

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleasePermissionService.cs
```diff
@@ -45,33 +45,58 @@ public class ReleasePermissionService : IReleasePermissionService
...
+
+                    return await invites
+                        .ToAsyncEnumerable()
+                        .Where(i => accepted == null || i.Accepted == accepted)
+                        .Select(i => new ContributorInviteViewModel
+                        {
+                            Email = i.Email,
+                        })
+                        .OrderBy(model => model.Email)
+                        .ToListAsync();
```
ntsim, line 96: Minor thing, but this isn't doing any async work so there's no need to use `ToAsyncEnumerable` here and we could also termininate it with `ToList` without needing to await.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleasePermissionService.cs
```diff
@@ -98,7 +124,8 @@ public class ReleasePermissionService : IReleasePermissionService
                         .Distinct()
                         .ToListAsync();
 
-                    return users
+                    return await users
+                        .ToAsyncEnumerable()
```
ntsim, line 128: Same as above - we're not doing any async work here so there's no need to convert this to an async enumerable, and we can terminate with `ToList` without needing to await.

ntsim (approved):

PR #3093, opened 2022-01-06
Ees 2480 - scale set ARM template

luke-h1:


infrastructure/templates/template.json
```diff
@@ -3853,6 +3866,61 @@
...
+            "runOutputName": "[parameters('managedimagename')]",
+            "aritfactTags": {
+              "source": "azVmImageBuilder",
+              "releaseStatus": "production",
+              "type": "mdionly"
+            }
+          }
+        ]
+      }
+    },
```
ntsim, line 3923: All of these changes should go into their own template file. We want to create a new DevOps pipeline to try and build this image separately from the rest of the infrastructure, so we can't store all of this in the same template file as everything else.

Would suggest putting it in new directory i.e. `infrastructure/templates/test-runner/image-template.json`. If this image builds successfully, we might want to make deployment of the VMs independent of the building of the image by having them in separate template files.

ntsim (changes requested):

infrastructure/templates/template.json
```diff
@@ -564,9 +564,22 @@
...
+      "type": "string"
+    },
+    "svcLocation": {
+      "type": "string"
+    },
+    "managedImageResGroupId": {
+      "type": "string"
+    },
+    "managedImageName": {
+      "type": "string"
```
ntsim: Any reason we have any changes in this file? `template.json` and `image-template.json` should be completely independent of one another so that we can deploy one without the other.

It wouldn't be a good idea to be have to rebuild the test runner image every time we want to deploy the rest of our infrastructure as building an image isn't necessarily fast.

ntsim (changes requested):

infrastructure/templates/template.json
luke-h1: I think this must have been something leftover from when I force pushed during standup. Should be fixed now

infrastructure/templates/test-runner/image-template.json
```diff
@@ -0,0 +1,71 @@
...
+    "svcLocation": {
+      "type": "string"
+    },
+    "managedImageResGroupId": {
+      "type": "string"
+    },
+    "managedImageName": {
+      "type": "string"
+    }
+  },
```
ntsim, line 20: This template doesn't look right - we need to follow the same format as the main `template.json`:

```json5
{
  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
  "contentVersion": "1.0.0.0",
  "parameters": {
  	// your parameters
  },
  "variables": {
  	// your variables
  },
  "resources": [
    {
      "type": "Microsoft.VirtualMachineImages/imageTemplates",
      "name": "[parameters('imageTemplateName')]",
       // etc
    }
  ]
}
``

File: infrastructure/templates/test-runner/image-template.json [hunk 4]
```diff
@@ -0,0 +1,71 @@
...
+        "aritfactTags": {
+          "source": "azVmImageBuilder",
+          "releaseStatus": "production",
+          "type": "mdionly"
+        }
+      }
+    ]
+  },
+  "variables": {
+    "resourceid": "[concat(parameters('managedImageResGroupId'),parameters('managedImageName'))]"
```
ntsim, line 69: Would make the casing of this consistent with parameters i.e `resourceId` instead of `resourceid`

ntsim (changes requested):

infrastructure/templates/test-runner/image-template.json
ntsim: This hasn't been resolved. Also, it'd be good if the `variables` section could be at the top of the template rather than the bottom so it's more obvious where to find them.

infrastructure/templates/test-runner/image-template.json
```diff
@@ -0,0 +1,73 @@
...
+  "contentVersion": "1.0.0.0",
+  "name": "[parameters('imageTemplateName')]",
+  "type": "Microsoft.VirtualMachineImages/imageTemplates",
+  "apiVersion": "2020-02-14",
+  "location": "[parameters('svcLocation')]",
+  "dependsOn": [],
+  "variables": {
+    "resourceId": "[concat(parameters('managedImageResGroupId'),parameters('managedImageName'))]"
+  },
+  "parameters": {
```
ntsim, line 12: This still isn't right. See my example again:

```json5
{
  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
  "contentVersion": "1.0.0.0",
  "parameters": {
  	// your parameters
  },
  "variables": {
  	// your variables
  },
  "resources": [
    {
      "type": "Microsoft.VirtualMachineImages/imageTemplates",
      "name": "[parameters('imageTemplateName')]",
       // etc
    }
  ]
}
```

ntsim (changes requested):

infrastructure/templates/test-runner/image-template.json
```diff
@@ -0,0 +1,74 @@
...
+        "location": "[parameters('svcLocation')]",
+        "runOutputName": "[parameters('managedimagename')]",
+        "aritfactTags": {
+          "source": "azVmImageBuilder",
+          "releaseStatus": "production",
+          "type": "mdionly"
+        }
+      }
+    ]
+  }
```
ntsim, line 73: All of this needs to go with the image template resource object under `resources`

infrastructure/templates/test-runner/image-template.json
```diff
@@ -0,0 +1,74 @@
...
+      "type": "string"
+    }
+  },
+  "variables": {
+    "resourceId": "[concat(parameters('managedImageResGroupId'),parameters('managedImageName'))]"
+  },
+  "resources": [
+    {
+      "type": "Microsoft.VirtualMachineImages/imageTemplates",
+      "name": "[parameters('imageTemplateName')]"
```
ntsim, line 24: Seems like we're missing an `apiVersion` property. We should set this to `2021-10-01` according to the docs

ntsim (changes requested):

ntsim:
Hey @luke-h1 just noticed this was closed, any particular reason?

luke-h1:
> Hey @luke-h1 just noticed this was closed, any particular reason?

Hi, we've decided to stop investigating/putting effort into getting scale sets working. It seems that after the concurrency changes Mark made to data factory in development environments, the standard Azure pipelines seem fine for the UI tests (with processes set to 3). If we ever needed to increase the concurrency of the tests for faster CI builds, this would likely result in having to scale the Admin APIs and private DBs which, along with provisioning scale-sets, will increase the cost of the development environment quite a bit. In an ideal world, we would be able to run the tests at 8 - 12 processes (like we do when running tests locally) and be able to quickly iterate with any failures that happen. I think the main restriction with that happening is that the development environment simply won't be able to handle that amount of traffic without significant auto-scaling being put in place.

PR #3094, opened 2022-01-06
EES-3085 - cli validation

luke-h1:


useful-scripts/misc-test-utils/package.json
```diff
@@ -29,6 +29,7 @@
     "typescript": "^4.3.5"
   },
   "dependencies": {
+    "guid-typescript": "1.0.9",
```
ntsim, line 32: Would suggest using the `uuid` package (with `@types/uuid`) instead. It's way more popular and is actively maintained unlike the above package.

ntsim (changes requested):

useful-scripts/misc-test-utils/package.json
luke-h1: Looks like the `validate` function from UUID doesn't seem to play nice with Microsoft's GUIDs. I attempted to upload a subject using an existing release's ID (with validation in place from the UUID package) but it seems that the validate function throws an error. I initially thought that the UUID package generated UUIDs / GUIDs in a different way to how Microsoft defines GUIDs which is why I chose the `guid-typescript` package. 

```
 >> Not a valid GUID
```

useful-scripts/misc-test-utils/src/utils/guid.ts
```diff
@@ -0,0 +1,84 @@
+/* eslint-disable import/prefer-default-export */
+export class Guid {
```
ntsim, line 2: Wouldn't ignore this lint - we can make this a default export. We'd also want to the name the file `Guid.ts` as class files should typically be PascalCased.

useful-scripts/misc-test-utils/src/utils/guid.test.ts
```diff
@@ -0,0 +1,76 @@
+import { Guid } from './guid';
+
+describe('Guids', () => {
```
ntsim, line 3: We normally put the test files inside of `__tests__` directories adjacent to the associated file i.e.. `__tests__/Guid.test.ts`

ntsim (changes requested):

useful-scripts/misc-test-utils/package.json
```diff
@@ -42,5 +47,41 @@
...
+      "^.+\\.(t|j)sx?$": [
+        "@swc/jest",
+        {
+          "sourceMaps": true,
+          "jsc": {
+            "parser": {
+              "syntax": "typescript",
+              "tsx": true
+            },
+
```
ntsim, line 75: Unnecessary space here

useful-scripts/misc-test-utils/src/index.ts
```diff
@@ -14,6 +14,7 @@ import createPublication from './modules/publication/createPublication';
 import addReleaseTextContentBlock from './modules/release/addContentBlock';
 import themeService from './services/themeService';
 import topicService from './services/topicService';
+import Guid from './utils/Guid';
```
ntsim, line 17: Does this import actually work? The files still seem to be lowercase `guid.ts` rather than `Guid.ts`

luke-h1: hmm, that's strange (not seeing that). I've just ran `git rm -r --cached misc-test-utils` to try to get git to pick up the changes in file casing. Not sure if this has sorted it?

useful-scripts/misc-test-utils/src/utils/__tests__/guid.test.ts
```diff
@@ -0,0 +1,76 @@
+import Guid from '../Guid';
```
ntsim, line 1: The test file name also needs to be uppercased too

luke-h1: Cool, didn't realize the test file had to be uppercase too.

ntsim (approved):

PR #3097, opened 2022-01-06
EES-3004 - optimised matching on FilterItemIds in table results stored procedures

duncan-at-hiveit:
This PR:
- optimises the 2 existing "SelectObservations" stored procedures with a couple of changes.

This is in response to problems we're having on Prod with Subjects like QL01, where certain Filters can have thousands of potential FilterItem choices available.  In the case of QL01, around 10,000 choices exist for the Qualification filter for example when all Time Periods are chosen.

One change is to replace IN clauses that handle large numbers of values from table variables with IN clauses on a temporary table instead.  It was not possible to convert the list of FilterItemIds to a csv string of FilterItemIds using STRING_AGG and then use that string of ids in an IN clause, as is used widely in the `SelectObservationsByLocationCodes` stored proc for other lists of ids and codes, because the query engine cannot handle a SQL-stringed IN clause with several thousand possible values in it.  5,000 was OK, but 10,000 caused it to run out of resources.

Another change is the optimisation to the HAVING clause, whereby it now checks that the number of matching FilterItems per Observation row is equal to the number of Filters that that are present in the set of FilterItemIds without having to do several JOINs to other tables in the process.

Note that we might also want to make a similar change to remove the IN clause for Location ids in the new `SelectObservation` query, replacing it with IN clauses on a temporary table of LocationIds, but this'll come with testing when the Location Id support is all in.

The results of these changes as tested manually on pre-prod's public statistics db with a QL01 table tool query brought a 240+ seconds query down to 6-20 seconds.

Capturing before-and-after datablock results on pre-prod and then comparing them with the command `diff -I"^time for response:.*" -I "Not Found" -r results_preprod_before/ results_preprod_after/` shows that the results were the same, and the overall run time reduced from 2100 seconds to 1033 seconds to query 120+ datablocks when public-statistics was unscaled at 1 vCore.

Note also that:
* The original variant of this stored procedure had no index on the temp table and was using a JOIN on the temporary table.  Its run time was 1900 seconds.
* The original variant of this stored procedure plus the index on the temp table and the JOIN on the temporary table's run time was 1200 seconds.

Therefore all variants of the original were faster, but the best performing was with the index on the temp table, and the IN clause against the temp table rather than the JOIN. 

image


src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_SelectObservations.sql
```diff
@@ -13,24 +13,31 @@ AS
...
-                '    FilterItem fi' +
-                '    JOIN FilterGroup fg ON fi.FilterGroupId = fg.Id' +
-                '    JOIN Filter f ON fg.FilterId = f.Id' +
-                '    WHERE fi.Id IN (SELECT id FROM @filterItemIds)' +
-                ') ' +
-                'ORDER BY o.Id;'
+
+    IF (EXISTS(SELECT * FROM #FilterItemId))
+        SET @sql += N'GROUP BY o.Id ' +
+                     'HAVING COUNT(DISTINCT ofi.FilterId) = ' +
```
benoutram, line 37: Can this lead to incorrect results if the filter items in the query belong to less number of filters than the number of filters of the subject?

If so, because we only enforce one filter item per filter in the UI, I think we should either add a warning in the comments of the stored procedure, or change this back to work as before.

As discussed on Slack, it looks like the right hand side of the `HAVING` clause that we had previously can be worked out earlier in the procedure.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_SelectObservations.sql
```diff
@@ -13,24 +13,31 @@ AS
...
+
+    CREATE TABLE #FilterItemId(Id uniqueidentifier)
+    INSERT INTO #FilterItemId SELECT * FROM @filterItemIds
+
+    SET @sql = N'SELECT o.id ' +
+                'FROM Observation o '
+                
+    IF (EXISTS(SELECT * FROM #FilterItemId))
+        SET @sql += N'JOIN ObservationFilterItem ofi ON o.Id = ofi.ObservationId ' +
+                     'JOIN #FilterItemId filterItemId ON ofi.FilterItemId = filterItemId.id '
```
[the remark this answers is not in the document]
duncan-at-hiveit: As discussed on Slack, the temp table performs much better than the table variable.

In addition, with further testing it shows that adding an index to the temp table (which is an advantage of temp tables over table variables) provides further speed optimisations.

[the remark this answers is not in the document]
duncan-at-hiveit: Also in addition and as discussed over Slack, I also tested with both an IN clause with the temp table and a JOIN against the temp table, and the IN clause consistently performed better than the JOIN.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_SelectObservations.sql
duncan-at-hiveit: Have amended this so that we grab a unique Filter count upfront (the count being based on the unique Filters that appear in the set of FilterItemIds passed in.  Therefore if a Subject has 5 Filters but only the FilterItemIds of 3 of these Filters are passed to the stored proc, the unique Filter count will be 3.

When we do the GROUP BY and the HAVING clauses at the end, we're left with (for each Observation row) all of that row's ObservationFilterItem rows that matched any of the FilterItemIds passed in.  We then just have to get the count of the unique ids of these OFIs' FilterId columns, and check that they match with the unique Filter count.  If there's less, then that means that the Observation row didn't have at least one matching OFI for one of the Filters.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_SelectObservationsByLocationCodes.sql
```diff
@@ -177,60 +197,21 @@ DECLARE
...
-                'ORDER BY o.Id;'
+
+    IF (@filterItemsExist = 1)
+        SET @sql += N'GROUP BY o.Id ' +
+                     'HAVING COUNT(DISTINCT ofi.FilterId) = @uniqueFiltersCount '
+
+    SET @sql += N'ORDER BY o.Id;'
+
     SET @paramDefinition = N'@subjectId uniqueidentifier,
-                           @filterItemIds IdListGuidType READONLY,
```
benoutram, line 191: I think the wrong version of the code is in this 'Current' version file?

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_SelectObservationsByLocationCodes.sql
duncan-at-hiveit: I think they're correct - they're the same as the latest updating migrations

benoutram: Got it, saw these params being removed and thought it must be the other version! I see now that we don't need them in the param definition for the `sp_executesql`. 🙂

benoutram (approved):

PR #3099, opened 2022-01-07
Hotfix/ees 3094 correct missing await

duncan-at-hiveit:
This PR is a fix for an issue that teams were seeing around Footnotes after deleting Subjects.

The problem was caused by a Task not being awaited prior to the next unit of work being called.  Specifically in this instance, an `OnSuccess()` method was being used that called the function passed in and returned immediately, rather than recognizing it as a Task and awaiting it before returning.

To fix this, the problem piece of code now uses the correct pre-existing `OnSuccessVoid()` method that expects a non-generic Task to be passed in.

An additional safety measure is added by adding another Obsolete extension method that will signpost us to using the correct method in the future and avoid this pitfall.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Either.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Either.cs
@@ -179,7 +179,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Model
             return await func();
         }
         
-        [Obsolete("Use OnSuccessDo or OnSuccessVoid for chaining a Task with no generic type")]
+        [Obsolete("Use OnSuccessDo or OnSuccessVoid for chaining a non-generic Task")]
         public static async Task<Either<TFailure, Unit>> OnSuccess<TFailure, TSuccess1>(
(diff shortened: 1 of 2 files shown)


src/GovUk.Education.ExploreEducationStatistics.Common/Model/Either.cs
```diff
@@ -178,6 +178,19 @@ public static class EitherTaskExtensions
 
             return await func();
         }
+        
+        [Obsolete("Use OnSuccessDo or OnSuccessVoid for chaining a non-generic Task")]
+        public static async Task<Either<TFailure, Unit>> OnSuccess<TFailure, TSuccess1>(
```
ntsim, line 183: Would suggest we write some basic tests for this, although it's going to end up throwing warnings about using obsolete methods?

duncan-at-hiveit: Yeah it would do - I'd hope that we don't need to write any tests for it cos we'll never use it!

ntsim (approved):

PR #3104, opened 2022-01-10
EES-1683 Use different notify templates when publishing original and …

markhiveit:
…amended releases


markhiveit:
I updated the variables / override template parameters on the CI pipeline for Dev, but still need to do it for the other environments.

markhiveit:
I've added the variables to all environments on the CI now

infrastructure/templates/template.json
```diff
@@ -270,6 +270,20 @@
...
       }
     },
+    "originalReleasePublicationEmailTemplateId": {
+      "type": "securestring",
+      "defaultValue": "",
+      "metadata": {
+        "description": "Gov UK Notify service template Id for amended release notification"
+      }
+    },
+    "amendedReleasePublicationEmailTemplateId": {
```
benoutram, line 280: I think you are missing changes to `template.json` within the config for the Notifier Function app to make use of these parameters. Without that, nothing is going to use them.

See elsewhere for 
`"PublicationNotificationEmailTemplateId": "[parameters('publicationNotificationEmailTemplateId')]",`
 (which will need removing btw).

src/GovUk.Education.ExploreEducationStatistics.Notifier.Model/PublicationNotificationMessage.cs
```diff
@@ -1,16 +1,27 @@
...
-        public string Name { get; set; }
-        public string Slug { get; set; }
+        public string PublicationName { get; set; } = "";
+        public string PublicationSlug { get; set; } = "";
+
+        public string ReleaseName { get; set; } = "";
+        public string ReleaseSlug { get; set; } = "";
+
+        public bool Amendment { get; set; }
+        public string UpdateNote { get; set; } = "";
```
benoutram, line 16: Personal preference I guess but I think we use `string.Empty` in most places that I've seen for the default initialisation.

src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/Services/NotificationsServiceTests.cs
```diff
@@ -108,23 +134,31 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
...
                         {
-                            Name = release1.Publication.Title,
                             PublicationId = release1.Publication.Id,
-                            Slug = release1.Publication.Slug,
+                            PublicationName = release1.Publication.Title,
+                            PublicationSlug = release1.Publication.Slug,
+                            ReleaseName = release1.Title,
+                            ReleaseSlug = release1.Slug,
+                            Amendment = false,
+                            UpdateNote = "No update note provided.",
```
benoutram, line 151: `UpdateNote` values of `No update note provided.` vs. `latest update note` won't be tested here as it's not defined as a member in equals/hashcode of `PublicationNotificationMessage`.

src/GovUk.Education.ExploreEducationStatistics.Notifier.Model/PublicationNotificationMessage.cs
```diff
@@ -1,16 +1,27 @@
...
 
         protected bool Equals(PublicationNotificationMessage other)
         {
-            return Name == other.Name && PublicationId.Equals(other.PublicationId) && Slug == other.Slug;
+            return PublicationId.Equals(other.PublicationId)
```
benoutram, line 20: I'm assuming we are overriding Equals/Hashcode to facilitate easier comparison during unit testing?

I think you could make this a `record` type in which case you won't have to do this as two records objects are the same if they store the same values. Then you wouldn't need to worry about which fields are included. See other comments about `UpdateNote` being included here which means it's not being unit tested correctly in the current code.

The setters can also be made `init` as they should never be updated.

src/GovUk.Education.ExploreEducationStatistics.Notifier/PublicationNotifier.cs
```diff
@@ -82,9 +90,14 @@ public class PublicationNotifier
...
-                        {"publication_link", webApplicationBaseUrl + "find-statistics/" + publicationNotification.Slug},
-                        {"unsubscribe_link", baseUrl + entity.PartitionKey + "/unsubscribe/" + unsubscribeToken}
+                        { "publication_name", notificationMessage.PublicationName },
+                        { "release_name", notificationMessage.ReleaseName },
+                        {
+                            "release_link",
+                            $"{webApplicationBaseUrl}find-statistics/{notificationMessage.PublicationSlug}/{notificationMessage.ReleaseSlug}"
+                        },
+                        { "update_note", notificationMessage.UpdateNote },
+                        { "unsubscribe_link", baseUrl + entity.PartitionKey + "/unsubscribe/" + unsubscribeToken }
```
benoutram, line 100: Could use string interpolation here to be consistent with how we create the `release_link` variable.

src/GovUk.Education.ExploreEducationStatistics.Notifier/PublicationNotifier.cs
```diff
@@ -173,8 +186,8 @@ public class PublicationNotifier
 
                     var values = new Dictionary<string, dynamic>
                     {
-                        {"publication_name", title},
-                        {"verification_link", baseUrl + id + "/verify-subscription/" + activationCode}
+                        { "publication_name", title },
```
benoutram, line 189: As `title` has changed from `string` to `string?` is this not giving a warning that `dynamic` should be `dynamic?`?

src/GovUk.Education.ExploreEducationStatistics.Notifier/local.settings.json
```diff
@@ -6,6 +6,8 @@
     "NotifyApiKey": "change-me",
     "VerificationEmailTemplateId": "change-me",
     "PublicationNotificationEmailTemplateId": "change-me",
+    "OriginalReleasePublicationEmailTemplateId": "change-me",
+    "AmendedReleasePublicationEmailTemplateId": "change-me",
```
benoutram, line 10: I think these both replace `PublicationNotificationEmailTemplateId` which needs to be removed.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
```diff
@@ -57,22 +55,32 @@ public async Task NotifySubscribersIfApplicable(params Guid[] releaseIds)
             }
         }
 
-        private IEnumerable<Publication> GetPublications(IEnumerable<Guid> releaseIds)
+        private async Task<PublicationNotificationMessage> BuildPublicationNotificationMessage(Release release)
         {
-            return _context.Releases
-                .AsNoTracking()
```
benoutram, line 63: You could add`.AsNoTracking` into the database access statement in `NotifySubscribersIfPossible`, e.g. `_context.Releases.AsNoTracking`...

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
```diff
@@ -57,22 +55,32 @@ public async Task NotifySubscribersIfApplicable(params Guid[] releaseIds)
...
+
+            await _context.Entry(release)
+                .Collection(r => r.Updates)
+                .LoadAsync();
+
+            var latestUpdateNote = release.Updates
+                .OrderBy(u => u.Created)
+                .LastOrDefault();
+
+            var latestUpdateNoteReason =
```
benoutram, line 72: Can there ever be no update note provided?

It's just that this code potentially allows for a scenario where a subscriber is sent an email

```
The update note for this change is:
  * No update note provided
```

or even worse if the update note is the empty string

```
The update note for this change is:
  * 
```

If we ever allow no update note I think it would be better to make the `UpdateNote` field of the message optional, and handle the update note and the surrounding wording as optional content in Gov.Uk notify. See [Optional Content](https://www.notifications.service.gov.uk/using-notify/guidance/edit-and-format-messages#optional-content).

If we don't ever allow null then you will already know I'm not a fan of this defensive coding where it's not necessary and would use `Last` rather than `LastOrDefault` to cause an exception so we are alerted to investigate what is allowing no update note.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
```diff
@@ -30,17 +28,17 @@ public async Task NotifySubscribersIfApplicable(params Guid[] releaseIds)
         {
             var releasesToNotify = _context.Releases
                 .Include(r => r.ReleaseStatuses)
+                .Include(r => r.Updates) // used by BuildPublicationNotificationMessage
+                .Include(r => r.Publication) // used by BuildPublicationNotificationMessage
                 .Where(r => releaseIds.Contains(r.Id))
+                .Distinct()
```
benoutram, line 34: I see the `Distinct` has probably been moved from below in the method that use to be `GetPublications` but I'm not sure what it's doing here. I think each row is going to be distinct by release id and update id and release status id already so there shouldn't be any duplicates.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
```diff
@@ -30,17 +28,17 @@ public async Task NotifySubscribersIfApplicable(params Guid[] releaseIds)
         {
             var releasesToNotify = _context.Releases
                 .Include(r => r.ReleaseStatuses)
+                .Include(r => r.Updates) // used by BuildPublicationNotificationMessage
+                .Include(r => r.Publication) // used by BuildPublicationNotificationMessage
```
benoutram, line 32: I don't think this set of includes necessarily generates the most efficient query. It would be interesting to see what this is if there's a lot of status updates and update notes.

I think it will join release, updates, and publication and release statuses so it pulls back all the fields of the publication for every update / release status combination and if there's a lot of approvals and update notes that would be pulling back a lot of rows with a lot of duplicated data!

src/GovUk.Education.ExploreEducationStatistics.Notifier/PublicationNotifier.cs
```diff
@@ -48,32 +52,36 @@ public class PublicationNotifier
         [FunctionName("PublicationNotifier")]
         // ReSharper disable once UnusedMember.Global
         public void PublicationNotifierFunc(
-            [QueueTrigger(PublicationQueue)] PublicationNotificationMessage publicationNotification,
+            [QueueTrigger(PublicationQueue)] PublicationNotificationMessage notificationMessage,
```
benoutram, line 55: Prior to this change the code only sent emails at a publication level regardless of whether  multiple releases were published within the same publication on the same evening.

The email contained the link to the latest release of the Publication.

I see that this PR is an overall change to now send emails for each Release version being published  so I'm not sure about the whole naming of this function, queue and message type any longer.

It was a originally a `PublicationQueue` which received a `PublicationNotificationMessage ` when a Publication had any new releases but that's no longer the case, so the naming may no longer be appropriate.

What happens now if there are two original releases of a publication being published on the same evening?

As th

benoutram (changes requested):

infrastructure/templates/template.json
```diff
@@ -270,6 +270,20 @@
         "description": "Gov UK Notify service template Id for new release notification"
       }
     },
+    "originalReleasePublicationEmailTemplateId": {
```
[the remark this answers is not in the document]
markhiveit: For some reason I thought I should delete this after this PR is deployed to Prod - I'll delete it now

src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/Services/NotificationsServiceTests.cs
markhiveit: I've added `UpdateNote` in `GetHashCode()`

src/GovUk.Education.ExploreEducationStatistics.Notifier.Model/PublicationNotificationMessage.cs
markhiveit: Switched PublicationNotificationMessage from a class to a record

src/GovUk.Education.ExploreEducationStatistics.Notifier/PublicationNotifier.cs
markhiveit: There is a null check on line 140
```
            if (id == null || email == null || slug == null || title == null)
            {
                return new BadRequestObjectResult("Please pass a valid email & publication");
            }
```

So I think this is okay

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
markhiveit: I've removed the `Distinct`

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
markhiveit: I've added `AsNoTracking` to the DB call in `NotifySubscribersIfApplicable`

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
markhiveit: I figured I'd do these includes here and save them being done again in `BuildPublicationNotificationMessage`. I'll remove them and let `BuildPublicationNotificationMessage` handle it.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
markhiveit: After doing this, it seems this prevents the `_context.Entry(release)[...].LoadAsync()`s from working, so I've removed `AsNoTracking` again.

src/GovUk.Education.ExploreEducationStatistics.Notifier/PublicationNotifier.cs
benoutram: Ok I see now, thanks. Nothing to do with this PR but it seems like we don't need to use `HttpRequest` as the request type, we can use a custom type according to [Payload](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook-trigger?tabs=csharp#payload). Then we might be able to get some built in validation rather than having to declare these as nullable. Worth trying in future I guess.

src/GovUk.Education.ExploreEducationStatistics.Notifier/PublicationNotifier.cs
markhiveit: Discussed this with Ben on Slack. We agreed to reluctantly leave the name as it is, as despite the loss of clarity, there are issues with having to manually delete resources across environments as this is deployed, and having to ask for those resources to be deleted on Prod. Also, it is potentially easy to miss a reference to the resources being renamed that could introduce a bug.

Not sure what others think?

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
markhiveit: Discussed this with Ben and Cam on Slack. For now, I'm going to leave the `The update note for this change is:
  * No update note provided` text, as amendments should always have a new update note, so by changing this we'd be protecting against a possibility that should never happen.

I have changed the `LastOrDefault` to `Last`, but that DB query only runs if it is an amendment being published. I've also added comments to the code clarifying that the update_note is only used in the amendment email template, not in the original release email template. This means that if someone tries to publish an amendment 

There is a question on whether it is better for the service to blow up if it expects an update note and one isn't found, or if it should continue. Not sure what others think about this, both in this instance and across the codebase generally? 

On the one hand, we want to avoid having lots of bugs that we're unaware of, but at the same time we don't want the service to have production issues - perhaps is better to have a production issue sooner rather than later, and have bugs that have made loads of bad data etc. 

Also possible there should be some balance between "make it blow up" and "defensively code" depending on the situation. Maybe in a particular instance the cost of a production issue is so high that making it blow up is a bad idea - even if that risks having a bug that doesn't present itself?

src/GovUk.Education.ExploreEducationStatistics.Notifier/PublicationNotifier.cs
benoutram: As mentioned on the call today, having thought about it, there shouldn't be ARM template config to clean up or resources to delete in Azure if renaming the `PublicationNotifier` function because individual functions inside the function project have no specific config as far as I know. 

The new queue should create itself when first accessed. We can remove the unused queue once the change is deployed,(checking its empty first).

markhiveit: > Seems like the more appropriate name would be release-notifications (or similar - also note that I've not added queue into the name as that's just redundant).

I've renamed the queue as you suggest, but I've named the variable in `NotifierQueues` to `ReleaseNotificationQueue`. With people normally importing classes like `NotifierQueues` statically, having suffixing `Queue` seem worth it in this instance.

markhiveit: Done

src/GovUk.Education.ExploreEducationStatistics.Notifier.Model/NotifierQueues.cs
```diff
@@ -2,6 +2,6 @@
 {
     public static class NotifierQueues
     {
-        public const string PublicationQueue = "publication-queue";
+        public const string ReleaseNotificationQueue = "release-notificiations";
```
ntsim, line 5: Typo in the string

src/GovUk.Education.ExploreEducationStatistics.Notifier.Model/PublicationNotificationMessage.cs
```diff
@@ -1,29 +1,18 @@
+#nullable enable
 using System;
 
 namespace GovUk.Education.ExploreEducationStatistics.Notifier.Model
 {
-    public class PublicationNotificationMessage
+    public record PublicationNotificationMessage
```
ntsim, line 6: Shouldn't this be `ReleaseNotificationMessage`?

src/GovUk.Education.ExploreEducationStatistics.Notifier/ReleaseNotifier.cs
```diff
@@ -0,0 +1,88 @@
...
+{
+    // ReSharper disable once UnusedType.Global
+    public class ReleaseNotifier
+    {
+        private readonly ITokenService _tokenService;
+        private readonly IEmailService _emailService;
+        private readonly IStorageTableService _storageTableService;
+
+        public ReleaseNotifier(ITokenService tokenService, IEmailService emailService,
+            IStorageTableService storageTableService)
```
ntsim, line 23: Would keep method arguments one per line for neatness and ease of refactoring.

src/GovUk.Education.ExploreEducationStatistics.Notifier/ReleaseNotifier.cs
```diff
@@ -0,0 +1,88 @@
...
+                context.FunctionName);
+
+            var config = LoadAppSettings(context);
+
+            var emailTemplateId = notificationMessage.Amendment
+                ? config.GetValue<string>(AmendedReleasePublicationEmailTemplateIdName)
+                : config.GetValue<string>(OriginalReleasePublicationEmailTemplateIdName);
+
+            var baseUrl = config.GetValue<string>(BaseUrlName);
+            var webApplicationBaseUrl = config.GetValue<string>(WebApplicationBaseUrlName);
```
ntsim, line 47: We can't be certain this URL comes with a trailing slash, so it might be a good idea to use the `AppendTrailingSlash` extension method on the string.

src/GovUk.Education.ExploreEducationStatistics.Notifier/Utils/NotifierUtils.cs
```diff
@@ -0,0 +1,50 @@
...
+        public const string WebApplicationBaseUrlName = "WebApplicationBaseUrl";
+        public const string TokenSecretKeyName = "TokenSecretKey";
+        public const string PendingSubscriptionsTblName = "PendingSubscriptions";
+
+        public const string ConfirmationEmailTemplateIdName = "SubscriptionConfirmationEmailTemplateId";
+        public const string VerificationEmailTemplateIdName = "VerificationEmailTemplateId";
+        public const string OriginalReleasePublicationEmailTemplateIdName =
+            "OriginalReleasePublicationEmailTemplateId";
+        public const string AmendedReleasePublicationEmailTemplateIdName =
+            "AmendedReleasePublicationEmailTemplateId";
```
ntsim, line 26: Would suggest moving these into a separate util class called `ConfigKeys` (or similar) to make it clearer what these are actually referring to.

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Notifier.Model/NotifierQueues.cs
markhiveit: My bad - fixed.

src/GovUk.Education.ExploreEducationStatistics.Notifier/ReleaseNotifier.cs
markhiveit: I've added it here and in existing instances in `PublicationNotifier.cs`. 

`AppendTrailingSlash` exists in `Common`, which wasn't added to `Notifier`, so I've added it.

src/GovUk.Education.ExploreEducationStatistics.Notifier/local.settings.json
```diff
@@ -5,7 +5,8 @@
     "FUNCTIONS_WORKER_RUNTIME": "dotnet",
     "NotifyApiKey": "change-me",
     "VerificationEmailTemplateId": "change-me",
-    "PublicationNotificationEmailTemplateId": "change-me",
+    "OriginalReleasePublicationEmailTemplateId": "change-me",
+    "AmendedReleasePublicationEmailTemplateId": "change-me",
```
ntsim, line 9: These keys are a bit wordy.

Not saying we 100% need to change these, but it'd be a nice touch to shorten these down to something like:

- `ReleaseEmailTemplateId`
- `ReleaseAmendmentEmailTemplateId`

Including the word 'publication' is a bit redundant in the context of a release (a release by its definition always involves a publication). 

It's also a good idea to name things so that they sit close together when arranged alphabetically to make things easier to find. We can do this by keeping the less specific nouns first and adding progressively more specific nouns later (almost how you would arrange files in folders and similar to how we name components in the frontend).

ntsim (approved):

PR #1862, opened 2020-07-17
Bugfix/ees 1176 correct subject and footnote delete

shakes90:
This change ensures that subjects & footnotes which are part of the amendment being published but were not generated as part of the original version are not deleted on republish.

PR #3110, opened 2022-01-14
EES-3103 Update DropAndCreateRelease to not delete orphaned footnotes

markhiveit:
We are making this change due to an intermittent bug: The `pl_release_statistics_stage2` pipeline was failing on the `Copy_ReleaseFootnotes` stage, due to the rows added by the previous job `Copy_Footnotes` not being present. This was happening because sometimes the `pl_release_statistics_stage1` pipeline job `Copy_Release` would run between these two jobs, which calls the DropAndCreateRelease stored procedure, which would delete orphaned footnotes.

As part of this PR, we're removing the LinkedSubject and LinkedFootnote arguments to DropAndCreateRelease. Both these arrays would be fetched from the Statistics DB.

LinkedSubjects is an array of every subject that both had a ReleaseSubject entry for the release in question AND had a ReleaseSubject entry for another release. 

LinkedFootnotes is an array of every footnote that both had a ReleaseFootnote entry for the release in question AND had a ReleaseFootnote entry for another release.

The previous version of DropAndCreateRelease would not delete Subject table rows that were included in LinkedSubjects, and not delete Footnote table rows that were in LinkedFootnotes. 

I'm not sure why this was done: I speculate that this was to save having recopy Subjects and Footnotes to the PublicStatistics DB that were inevitably going to be copied from a future release in the Statistics DB.

Edit: Did some more investigation with Ben. This is when the arguments were introduced: https://dfedigital.atlassian.net/browse/EES-1176
https://github.com/dfe-analytical-services/explore-education-statistics/pull/1862

We believe we understand 1176 now. It seems the problem was that a subjects would only get copied across to public-stats db for the particular version of the release they were imported. So if a subject was imported for version 0 of a release, then if version 1 was republished, that subject would be deleted, but then not recopied across, and so we got the INSERT foreign key restraint error on Copy_ReleaseSubjects.


src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_DropAndCreateRelease.sql
```diff
@@ -10,19 +10,25 @@ BEGIN
...
 
+    DECLARE @FootnoteIds TABLE (Id UNIQUEIDENTIFIER PRIMARY KEY NOT NULL);
+    INSERT INTO @FootnoteIds (Id) SELECT RF.FootnoteId FROM dbo.ReleaseFootnote RF WHERE ReleaseId = (SELECT R.Id FROM @Release R);
+
+    DELETE FROM dbo.SubjectFootnote WHERE FootnoteId IN (SELECT F.Id FROM @FootnoteIds F);
+    DELETE FROM dbo.FilterFootnote WHERE FootnoteId IN (SELECT F.Id FROM @FootnoteIds F);
+    DELETE FROM dbo.FilterGroupFootnote WHERE FootnoteId IN (SELECT F.Id FROM @FootnoteIds F);
+    DELETE FROM dbo.FilterItemFootnote WHERE FootnoteId IN (SELECT F.Id FROM @FootnoteIds F);
+    DELETE FROM dbo.IndicatorFootnote WHERE FootnoteId IN (SELECT F.Id FROM @FootnoteIds F);
+    DELETE FROM dbo.SubjectFootnote WHERE FootnoteId IN (SELECT F.Id FROM @FootnoteIds F);
```
benoutram, line 21: There's a mistake here. This line is duplicated.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_DropAndCreateRelease.sql
```diff
@@ -10,19 +10,25 @@ BEGIN
...
     DELETE FROM dbo.ReleaseFootnote WHERE dbo.ReleaseFootnote.ReleaseId IN (SELECT R.Id FROM @Release R);
+
+    DELETE FROM dbo.Footnote WHERE Id IN (SELECT F.Id FROM @FootnoteIds F);
+
     DELETE FROM dbo.ReleaseSubject WHERE dbo.ReleaseSubject.ReleaseId IN (SELECT R.Id FROM @Release R);
     DELETE FROM dbo.Release WHERE dbo.Release.Id IN (SELECT R.Id FROM @Release R);
 
     DELETE FROM dbo.Subject WHERE dbo.Subject.Id NOT IN (SELECT id FROM @LinkedSubjectsList) AND NOT EXISTS (SELECT * FROM dbo.ReleaseSubject WHERE dbo.ReleaseSubject.SubjectId = dbo.Subject.Id) AND SoftDeleted != 1;
 
-    DELETE FROM dbo.Footnote WHERE dbo.Footnote.Id NOT IN (SELECT id FROM @LinkedFootnotesList) AND NOT EXISTS (SELECT * FROM dbo.ReleaseFootnote WHERE dbo.ReleaseFootnote.FootnoteId = dbo.Footnote.Id)
```
benoutram, line 19: I think we should drop the `@LinkedFootnotes` parameter now and all the associated data factory config associated with the tasks leading up to `Copy_Release` which work out the list.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_DropAndCreateRelease.sql
```diff
@@ -10,18 +10,28 @@ BEGIN
     IF LEN(ISNULL(@LinkedFootnotes, '')) > 0
         INSERT @LinkedFootnotesList SELECT value FROM STRING_SPLIT(@LinkedFootnotes,',');
 
+    DECLARE @FootnoteIds TABLE (Id UNIQUEIDENTIFIER PRIMARY KEY NOT NULL);
+    INSERT INTO @FootnoteIds (Id) SELECT RF.FootnoteId FROM dbo.ReleaseFootnote RF WHERE ReleaseId = (SELECT R.Id FROM @Release R);
+
+    DELETE FROM dbo.SubjectFootnote WHERE FootnoteId IN (SELECT F.Id FROM @FootnoteIds F);
```
benoutram, line 16: Take it you didn't like my suggestion of deleting with a join? 🙂

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_DropAndCreateRelease.sql
```diff
@@ -10,18 +10,28 @@ BEGIN
...
+    DELETE FROM dbo.Footnote WHERE Id IN (SELECT F.Id FROM @FootnoteIds F);
 
-    DELETE FROM dbo.Footnote WHERE dbo.Footnote.Id NOT IN (SELECT id FROM @LinkedFootnotesList) AND NOT EXISTS (SELECT * FROM dbo.ReleaseFootnote WHERE dbo.ReleaseFootnote.FootnoteId = dbo.Footnote.Id)
-                               AND NOT EXISTS (SELECT * FROM dbo.FilterFootnote WHERE dbo.FilterFootnote.FootnoteId = dbo.Footnote.Id)
-                               AND NOT EXISTS (SELECT * FROM dbo.FilterItemFootnote WHERE dbo.FilterItemFootnote.FootnoteId = dbo.Footnote.Id)
-                               AND NOT EXISTS (SELECT * FROM dbo.FilterGroupFootnote WHERE dbo.FilterGroupFootnote.FootnoteId = dbo.Footnote.Id)
-                               AND NOT EXISTS (SELECT * FROM dbo.IndicatorFootnote WHERE dbo.IndicatorFootnote.FootnoteId = dbo.Footnote.Id)
-                               AND NOT EXISTS (SELECT * FROM dbo.SubjectFootnote WHERE dbo.SubjectFootnote.FootnoteId = dbo.Footnote.Id);
+    DECLARE @SubjectIds TABLE (Id UNIQUEIDENTIFIER PRIMARY KEY NOT NULL);
+    INSERT INTO @SubjectIds (Id) SELECT RS.SubjectId FROM dbo.ReleaseSubject RS WHERE ReleaseId = (SELECT R.Id FROM @Release R);
```
benoutram, line 28: Same comment here about  dropping the `linkedSubjects` parameter and updating the data factory config.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_DropAndCreateRelease.sql
markhiveit: Just not seen that syntax before so I sidestepped it :)

I've switched to using a join now

infrastructure/templates/datafactory/components/template.json
```diff
@@ -2086,143 +2063,6 @@
...
-                            "retryIntervalInSeconds": 30,
-                            "secureOutput": false,
-                            "secureInput": false
-                        },
-                        "userProperties": [],
-                        "typeProperties": {
-                            "source": {
-                                "type": "AzureSqlSource",
-                                "sqlReaderQuery": {
-                                    "value": "SELECT FootnoteId FROM dbo.ReleaseFootnote RF WHERE RF.ReleaseId = CONVERT(UNIQUEIDENTIFIER, '@{pipeline().parameters.releaseId}')\nAND RF.FootnoteId IN (SELECT FootnoteId FROM dbo.ReleaseFootnote WHERE FootnoteId = RF.FootnoteId AND ReleaseId != CONVERT(UNIQUEIDENTIFIER, '@{pipeline().parameters.releaseId}'))\n",
```
benoutram, line 2178: I think it would be helpful in the PR description to explain what the purpose of these two 'linked' vars was (as they select from the `private-stats` db) and what they meant in terms of what was/wasn't being deleted by the old version from the `public-stats` db.

It will then be easier to review what the new version is doing to make sure we aren't missing something from before.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20220114144422_E3103UpdateDropAndCreateReleaseToNotDeleteOrphanedFootnotes.cs
```diff
@@ -0,0 +1,26 @@
...
+using Microsoft.EntityFrameworkCore.Migrations;
+using static GovUk.Education.ExploreEducationStatistics.Data.Model.Migrations.MigrationConstants;
+
+namespace GovUk.Education.ExploreEducationStatistics.Data.Model.Migrations
+{
+    [ExcludeFromCodeCoverage]
+    public partial class E3103UpdateDropAndCreateReleaseToNotDeleteOrphanedFootnotes : Migration
+    {
+        private const string PreviousMigrationId = "20210512112804";
+        private const string MigrationId = "20220114144422";
```
benoutram, line 12: See example of very recent stats migrations where we find the `PreviousMigrationId`  in its migration file, make it `public` and then reference it in this class. Reduces the chance of there being any mistake.

infrastructure/templates/datafactory/components/template.json
markhiveit: I've added further description to this PR. I am just guessing as to why these arguments were added - it's more than possible I'm missing their true reason for being added.

benoutram (approved):
As mentioned on Slack, still very suspicious about the original intentions of this code and the linked variables which we haven't been able to completely explain, so suggest lots of testing around this, in particular with existing a new data on amendments, and visibility of public data on the previous version of amended releases in the window between data getting copied by the data stage during publishing and the new version being made live at 9:30.

PR #3113, opened 2022-01-18
EES-3117 Add sensitive data filtering for Application Insights

ntsim:
This PR adds filtering to strip out sensitive information from telemetry being sent to Application Insights in both the backend and frontend.

The main intention for this is to be coupled with [EES-2823](https://dfedigital.atlassian.net/browse/EES-2823) where the websocket handshake relies on passing the user's access token via an `access_token` query parameter. Unfortunately, this is problematic as these query strings would get logged out to Application Insights, making it a security vulnerability.

We have implemented this as a fairly simplistic algorithm that detects if a query string parameter contains one of several redactable words (e.g. the word `email`, `code` or `token`). We try to account for potential false positives (e.g. `code` is in the word `encode`) by splitting up the key up into its constituent tokens i.e. `emailAddress1` becomes `email`, `address` and `1`. We can then check more precisely against each of these tokens to see if there was a match.

This is currently only aimed at filtering sensitive data from URL query strings, but could be extended in the future to other things as well.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Config/SensitiveDataTelemetryProcessor.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Config/SensitiveDataTelemetryProcessor.cs
@@ -1,0 +1,167 @@
+#nullable enable
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+using System.Web;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+using Microsoft.ApplicationInsights.Channel;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.ApplicationInsights.Extensibility;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Config
+{
+    /// <summary>
+    /// Tries to filter out any potentially sensitive data from telemetry
+    /// before it is sent to Application Insights e.g. query string parameters
+    /// containing things like access tokens or codes.
+    /// </summary>
+    public class SensitiveDataTelemetryProcessor : ITelemetryProcessor
+    {
+        private const string RedactedValue = "__redacted__";
+
+        private readonly HashSet<string> _redactedTokens = new()
+        {
+            "code",
+            "email",
+            "phone",
+            "token",
+            "passphrase",
+            "password",
+            "pass",
+            "user",
+            "name",
+            "username",
+            "firstname",
+            "lastname",
+            "fullname"
+        };
+
+        private readonly char[] _tokenDelimiters =
+        {
+            '_', '-', ':', '.', '|', ' ', '[', ']', '+'
+        };
+
+        private readonly Regex _casingRegex = new("([A-Z][a-z]*|[0-9])", RegexOptions.Compiled);
+
+        private ITelemetryProcessor Next { get; }
+
+        public SensitiveDataTelemetryProcessor(ITelemetryProcessor next)
+        {
+            Next = next;
+        }
+
+        public void Process(ITelemetry item)
+        {
+            switch (item)
+            {
+                case RequestTelemetry telemetry:
+                {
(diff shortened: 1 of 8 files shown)


src/GovUk.Education.ExploreEducationStatistics.Common/Config/SensitiveDataTelemetryProcessor.cs
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        private string? FilterUriQuery(string? uri)
+        {
+            if (uri.IsNullOrEmpty())
+            {
+                return uri;
+            }
+
+            var pathAndQuery = uri.Split('?', 2);
```
duncan-at-hiveit, line 107: Surprised there isn't a less manual way of getting a list of query params from a URI, but I've looked and can't see anything immediately obvious.

src/explore-education-statistics-common/src/services/applicationInsightsService.ts
```diff
@@ -38,3 +41,91 @@ export function initApplicationInsights(
...
+    default:
+      return true;
+  }
+
+  return true;
+}
+/* eslint-enable no-param-reassign */
+
+function filterUri(uri: string): string {
+  const [path, query] = uri.split('?', 2);
```
duncan-at-hiveit, line 97: No JS out-of-the-box one-liners to get the query params out of a URI?  Oh, or is the point that you still want to try to redact invalid URLs?

src/GovUk.Education.ExploreEducationStatistics.Common/Config/SensitiveDataTelemetryProcessor.cs
duncan-at-hiveit: Oh, or is the point that you still want to try to redact invalid URLs?

src/GovUk.Education.ExploreEducationStatistics.Common/Config/SensitiveDataTelemetryProcessor.cs
```diff
@@ -0,0 +1,160 @@
...
+    /// before it is sent to Application Insights e.g. query string parameters
+    /// containing things like access tokens or codes.
+    /// </summary>
+    public class SensitiveDataTelemetryProcessor : ITelemetryProcessor
+    {
+        private const string RedactedValue = "__redacted__";
+
+        private readonly HashSet<string> _redactedTokens = new()
+        {
+            "code",
```
duncan-at-hiveit, line 25: Was wondering if there was some OWASP list of common vulnerable tokens that was maintained?  Looks like a good starter for ten here.  Can't immediately think of any more to add off the top of my head.

duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Common/Config/SensitiveDataTelemetryProcessor.cs
ntsim: Yep that's why. Originally, I was doing some parsing of the URLs, but it would have potentially let through malformed URL structures. Dumbing the algorithm down _should_ make it a little more robust for these types of cases.

src/explore-education-statistics-common/src/services/applicationInsightsService.ts
ntsim: Same as above

src/GovUk.Education.ExploreEducationStatistics.Common/Config/SensitiveDataTelemetryProcessor.cs
ntsim: Doesn't seem to be, but I've added a few more around 'user' and 'name' like tokens. Some of these are a little overkill anyway. They mostly only help with cases where we've accidentally using query parameters in a non-secure way, but most of the time this should be caught in code review and be heavily discouraged (use request bodies instead).

PR #3117, opened 2022-01-19
EES-2880 Allow importing Institution and Planning Area geographic levels

benoutram:
This PR allows `Institution` and `PlanningArea`  geographic level data to be imported in statistical data.

Institution and Planning Area have previously been added as types of `Location` in the codebase and support already exists for querying them but they were configured in a list of levels deliberately ignored during import.

They now join `Provider` and `School` as levels which data will only be imported for if a file consists exclusively of data of that level.

If a file consists of a sole geographic level then we allow every row to be imported regardless of the level. This allows us to tidy up some of the code previously written to determine whether a row can be imported now that we no longer configure a set of levels which are always ignored.

If a file consists of multiple levels then any rows of levels `Institution`, `PlanningArea`, `Provider` and `School` will be ignored.

### Data file geography columns

```
geographic_level    Code column           Name column
============================================================
Institution         institution_id        institution_name
Planning Area       planning_area_code    planning_area_name
````

### Other changes

- Add Planning Area code/name to Location lookup query in `ImporterLocationService` which looks like it was previously missed.
- Added 'small' file containing only Planning Area level data.

### UI Test Result*

image

*Tested in combination with the fixes from https://github.com/dfe-analytical-services/explore-education-statistics/pull/3116

### CODE REVIEW

File: src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/SplitFileService.cs [hunk 1]
```diff
@@ -234,14 +215,13 @@ private static void CopyColumns(DataTable source, DataTable target)
         private static void CopyRows(DataTable target,
             IEnumerable<DataRow> rows,
             List<string> colValues,
-            HashSet<GeographicLevel> importGeographicLevels)
+            DataImport dataImport)
```
markhiveit, line 218: I know this change gives you access to `.IsSoleGeographicLevel()`, but then you're passing the whole `dataImport` object unnecessarily. 

My general rule of thumb up to now has to try and only pass what is required in arguments, but I've seen you and others make this kind of change in other places. Maybe I'm overestimating the cost of passing a large object between functions / or I am missing something?

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Utils/ImporterUtils.cs
```diff
@@ -1,51 +1,64 @@
...
         {
             GeographicLevel.Institution,
+            GeographicLevel.PlanningArea,
             GeographicLevel.Provider,
             GeographicLevel.School,
-            GeographicLevel.PlanningArea
         };
 
-        private static readonly List<GeographicLevel> AllowedSoloGeographicLevels = new List<GeographicLevel>
+        public static bool IsSoloGeographicLevel(this GeographicLevel geographicLevel)
```
markhiveit, line 24: I'd be tempted to rename this. I feel like there is potential confusion with the new `IsSoleGeographicLevel` function. Not sure what though. `CanImportGeographicLevelSolo`, `IsImportableSoloGeographicLevel`, `IsSoleImportableOnlyGeographicLevel`?

markhiveit (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterLocationService.cs
```diff
@@ -285,6 +284,10 @@ public class ImporterLocationService
                 .And(location => location.ParliamentaryConstituency_Code == (parliamentaryConstituency != null ? parliamentaryConstituency.Code : null)
                                  && location.ParliamentaryConstituency_Name == (parliamentaryConstituency != null ? parliamentaryConstituency.Name : null));
 
+            predicateBuilder = predicateBuilder
+                .And(location => location.PlanningArea_Code == (planningArea != null ? planningArea.Code : null) 
+                                 && location.PlanningArea_Name == (planningArea != null ? planningArea.Name : null));
+
```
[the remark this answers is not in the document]
benoutram: Yes looks like it was missed when Planning Areas were originally added.

Probably irrelevant to historical data like you say but worth noting that while the 'ignored' feature filtered out **rows** of those ignored levels, it didn't mean the columns associated with those levels weren't necessarily in use.

For example we could ignore 'Region' today to filter out all Regional rows from being imported but an LA or any other type of row that additionally specifies a Region attribute would still have the Region code/name stored against that location.

This bit of query criteria code is used to find a location unique by all location attributes rather than just its principal attribute so it theoretically could have been an issue. It seems unlikely that data for any other level would have used Planning Area as an additional attribute though so not worth investigating.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/ProcessorStatistics.cs
```diff
@@ -1,21 +1,22 @@
...
 
 namespace GovUk.Education.ExploreEducationStatistics.Data.Processor
 {
     public class ProcessorStatistics
     {
-        public int FilteredObservationCount { get; set; }
-        public int RowsPerBatch { get; set; }
-        public int NumBatches { get; set; }
-        public HashSet<GeographicLevel> GeographicLevels { get; set; }
+        public int TotalRows { get; }
```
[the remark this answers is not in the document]
benoutram: Have now renamed it to `ImportableRowCount` to reflect more accurately what it's a count of.
Name was originally `FilteredObservationCount` but I think 'rows' is more appropriate here to importing data.

The `rows` and `totalRows` thing is a mess but hopefully this will be addressed by a combination of EES-2006 and EES-1810.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Utils/ImporterUtils.cs
benoutram: Have removed `ImporterUtils` now in the most recent commits and created separate `SoloImportableLevels`.
Methods of `ImporterUtils` that were getting values based on cols and rows have now joined other similar methods in `CsvUtil`.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/SplitFileService.cs
benoutram: If you see the latest changes it should make more sense now where I've added it as a property of `DataImport`.

In relation to what you said, if we were passing the argument around between methods across different projects/libraries and that meant the receiving method being called had to be dependent on the same model, which in turn could pull in other unnecessary dependencies, then I can see the benefit of keeping it as separate arguments restricted to basic built-in data types.

In two other places (both in `ImporterService` rather than `SplitFileService`), the `DataImport` is being used as a parameter for 2 or 3 different fields that belong to it, one being this check on the geographic levels.

Rather than there being duplicated code in two places of `ImporterService` and here in `SplitFileService` to work out if the import is for a single level, the DataImport can provide a helper method `HasSoleGeographicLevel`.

The alternative would to get the set of levels from the `DataImport` and count if there's one, in 3 different place.

markhiveit (approved):

PR #3119, opened 2022-01-21
EES-3126 Replace ReleaseType database entity with enum.

benoutram:
This PR replaces the existing `ReleaseType` database entity with a new `ReleaseType` enum in both the back-end and the front-end.

@cjrace has confirmed there are no plans to make Release types user editable in the future which would warrant it staying as database reference data.

It should make the work of EES-3123 easier which is going to change the back-end 'themes' endpoint to include the Release type of the latest release of every publication.

There are a few benefits to removing `ReleaseType` and swapping to an enum:

* We avoid an API call `GET /api/meta/releasetypes` to get the value list when creating or editing a Release in the Admin application which is rather pointless when the values never change.
  * We also avoid all the service code and unit tests in both the front-end and back-end supporting this API call.
* We avoid having to create `ReleaseType`'s when setting up unit test data and save them in-memory database . This cuts out a chunk of existing unit test code around Releases and makes writing new tests easier in future.
  * Similarly in the front-end test code where the type is an object it is now a literal value.
* We avoid having to join on `ReleaseType` with a `.include()` in a few different places when hydrating a Release in order to know its type to produce a `ReleaseViewModel`.
* We avoid having to synchronise new release type entries in each of the environments with a database migration and the complication that configuring `ReleaseType` values as reference data in `ContentDbContext` gave us.

Backwards compatibility with `ReleaseTypeViewModel` in  `CachedReleaseViewModel` is introduced so that a content cache refresh isn't required at the same time that this change is deployed (with the thinking that it's best avoided right now as this change is already complicated by a database migration).

After this change is deployed we can replace `ReleaseTypeViewModel` with `ReleaseType`  in `CachedReleaseViewModel`. This will require a refresh of cached releases in the public content cache. A new ticket EES-3127 has been raised for this and it is mentioned in the code.

### UI Test Result

image


markhiveit:
There is some stuff in `useful-scripts/misc-test-utils/` that needs changing - I noticed that `releaseService.ts` contained the guid for an Ad hoc release.

markhiveit:
There is some stuff in `useful-scripts/misc-test-utils/` that needs changing - I noticed that `releaseService.ts` contained the guid for an Ad hoc release.

markhiveit:
I noticed there is a reference to the ReleaseType in `infrastructure/templates/datafactory/components/template.json - I believe this is for `Current_TableType_ReleaseType`. There is also a reference to `dbo.ReleaseType` in `Current_Routine_DropAndCreateRelease`. Pretty sure these both refer to the ReleaseType database object rather than the table, but I thought I'd mention it anyway in case I've read either wrong.

markhiveit:
I ran this locally, and existing releases were erroring for me on the frontend. A newly published release displayed correctly for me.

image

src/explore-education-statistics-admin/src/pages/release/components/ReleaseSummaryForm.tsx
```diff
@@ -105,7 +90,9 @@ const ReleaseSummaryForm = <
     timePeriodCoverageStartYear: Yup.string()
       .required('Enter a year')
       .length(4, 'Year must be exactly 4 characters'),
-    releaseTypeId: Yup.string().required('Choose a release type'),
+    releaseType: Yup.mixed<FormValues['releaseType']>()
+      .oneOf(Object.keys(releaseTypes) as ReleaseType[])
+      .required('Choose a release type'),
```
markhiveit, line 95: This validation message wasn't being shown for me, if I didn't select a release type on `ReleaseCreatePage`

src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseHelpAndSupportSection.tsx
```diff
@@ -72,25 +71,25 @@ const ReleaseHelpAndSupportSection = ({
             <p>No methodologies added.</p>
           )}
         </AccordionSection>
-        {release.type.title === ReleaseType.NationalStatistics && (
+        {release.type === 'NationalStatistics' && (
```
markhiveit, line 74: I was thinking that we should have a type safe way of doing this, rather than having 'NationalStatistics' as a string, but didn't see an easy way of doing it.

Maybe instead of `releaseTypes` being an object, it could be an array of strings - similar to `ImportStatus.tsx`'s `terminalImportStatuses` - might be a bad idea but just throwing it out there.

I also recognise that changing this basically means you have to redo a load of changes, so maybe not worth it regardless.

Would be nice if `release.type` was the same as the string values of `releaseTypes`, but I imagine a load of migration and cache refreshing is required to make that change...

markhiveit (changes requested):

benoutram:
> I noticed there is a reference to the ReleaseType in `infrastructure/templates/datafactory/components/template.json` - I believe this is for `Current_TableType_ReleaseType`. There is also a reference to `dbo.ReleaseType` in `Current_Routine_DropAndCreateRelease`. Pretty sure these both refer to the `ReleaseType` database object rather than the table, but I thought I'd mention it anyway in case I've read either wrong.

Thanks for checking but yes that's completely unrelated as it's the statistics database table type `ReleaseType` used as a parameter of the stored procedure. The content `ReleaseType` being remove here is not present in the statistics database so no chance of there being anything related to it in data factory.

benoutram:
> There is some stuff in `useful-scripts/misc-test-utils/` that needs changing - I noticed that `releaseService.ts` contained the guid for an Ad hoc release.

Good spot, have never used the CLI. I've updated it now and tested I can use it locally to create a Release. Thanks.

src/explore-education-statistics-admin/src/pages/release/components/ReleaseSummaryForm.tsx
benoutram: I was having trouble with this on Friday and it's the reason why there's a couple of failing front-end tests causing the build failure on this PR. I've not been able to sort it yet.

I made `releaseType` optional because I don't want to initialise it with a value but want to force a value to be chosen instead using validation. I suspect it's to do with that.

The docs on `oneOf` say
> Note that undefined does not fail this validator, even when undefined is not included in arrayOfValues. If you don't want undefined to be a valid value, you can use Schema.required.

... but `required` doesn't seem to be working here.

I'm looking for help on this one if anyone knows the solution? 🙂

markhiveit (approved):

PR #3122, opened 2022-01-25
EES-2470 - change table tool generated N/A to no data

luke-h1:


src/explore-education-statistics-common/src/modules/table-tool/components/utils/getCsvData.ts
```diff
@@ -65,13 +65,13 @@ export default function getCsvData(fullTable: WorkerFullTable): string[][] {
...
-          return 'n/a';
+          return 'no data';
         }
 
-        return matchingResult.measures[indicator.value] ?? 'n/a';
+        return matchingResult.measures[indicator.value] ?? 'no data';
       });
 
-      if (indicatorCells.every(cell => cell === 'n/a')) {
+      if (indicatorCells.every(cell => cell === 'no data')) {
```
ntsim, line 74: Let's extract the 'no data' out into a variable like `TimePeriodDataTable` so we only need to change it in one place in the future.

ntsim (changes requested):

ntsim (approved):

PR #3132, opened 2022-01-27
EES-3135 - cleanup slack test notifications

luke-h1:


tests/robot-tests/tests/libs/slack.py
```diff
@@ -39,21 +47,34 @@ def _generate_slack_attachments(env: str):
...
+            return True
+        else:
+            return False
+
+
 def send_slack_report(env: str):
+    should_send_report = _get_report_status()
+
     attachments = _generate_slack_attachments(env)
     data = {"attachments": attachments}
```
markhiveit, line 80: I noticed a couple of typos on line 84 and 85 but github doesn't let me comment on those lines, so I'm commenting here: "needs to bet set" -> "needs to be set"

tests/robot-tests/tests/libs/slack.py
```diff
@@ -71,16 +92,17 @@ def send_slack_report(env: str):
...
-            channels='#build',
-            file='UI-test-report.zip',
-            title='test-report.zip',
-        )
-    except SlackApiError as e:
-        print(f'Error uploading test report: {e}')
-    os.remove('UI-test-report.zip')
+    if should_send_report:
+        client = WebClient(token=slack_bot_token)
+        print('Sending UI test report to #build')
```
markhiveit, line 97: This is a minor gripe, but I'd keep the print statements consistent in whether we make them at the start or at the end. Currently, line 93 prints at the end, and this line prints at the start. 

I think you can skip this print and instead print on it's successful completion inside the try block below.

tests/robot-tests/tests/libs/slack.py
```diff
@@ -71,16 +92,17 @@ def send_slack_report(env: str):
...
+
+        shutil.make_archive('UI-test-report', 'zip', PATH)
```
markhiveit, line 99: I know it hasn't changed as part of this PR, but I'd rename the `PATH` variable to something like `PATH_TO_TEST_RESULTS`

markhiveit (changes requested):

tests/robot-tests/tests/libs/slack.py
luke-h1: Don't really agree with this. A bit too verbose for my liking. If we had multiple directories defined in this file then this would make sense.

markhiveit: I thought it was too ambiguious as it is - but I'm not precious about it - so feel free to leave it.

markhiveit (approved):

PR #3134, opened 2022-01-28
EES-3139 Add timestamp interfaces for entities

ntsim:
This PR adds timestamp interfaces that allow us to automatically update timestamp fields on entities automatically via events attached to Entity Framework's `ChangeTracker`.

We intend to use this as part the content locking work in [EES-2823](https://dfedigital.atlassian.net/browse/EES-2823), but this has been split out as it has cross-over with [EES-3138](https://dfedigital.atlassian.net/browse/EES-3138) to add consistent timestamp columns to entities.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Timestamps.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Timestamps.cs
@@ -1,0 +1,36 @@
+#nullable enable
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Model
+{
+    public interface ICreatedTimestamp<TDate> : ITimestampsInternal.ICreated
+    {
+        TDate Created { get; set; }
+    }
+
+    public interface IUpdatedTimestamp<TDate> : ITimestampsInternal.IUpdated
+    {
+        TDate Updated { get; set; }
+    }
+
+    public interface ISoftDeletedTimestamp<TDate> : ITimestampsInternal.ISoftDeleted
+    {
+        TDate SoftDeleted { get; set; }
+    }
+
+    public interface ICreatedUpdatedTimestamps<TCreated, TUpdated>
+        : ICreatedTimestamp<TCreated>, IUpdatedTimestamp<TUpdated>
+    {
+    }
+
+    /// <summary>
+    /// Marker interfaces for type guards without needing reflection.
+    /// Don't use this application level code (should only be
+    /// used in Entity Framework infrastructure code).
+    /// </summary>
+    public interface ITimestampsInternal
+    {
+        public interface ICreated {}
+        public interface IUpdated {}
+        public interface ISoftDeleted {}
+    }
+}
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Utils/DbContextUtils.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Utils/DbContextUtils.cs
@@ -1,0 +1,111 @@
+#nullable enable
+using System;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Common.Model;
+using Microsoft.EntityFrameworkCore;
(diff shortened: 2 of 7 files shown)


src/GovUk.Education.ExploreEducationStatistics.Common/Utils/DbContextUtils.cs
```diff
@@ -0,0 +1,54 @@
...
+            switch (e.Entry.State)
+            {
+                case EntityState.Added:
+                {
+                    if (e.Entry.Entity is ICreatedTimestamp entityWithCreated)
+                    {
+                        entityWithCreated.Created = DateTime.UtcNow;
+                    }
+
+                    if (e.Entry.Entity is IUpdatedTimestamp entityWithUpdated)
```
benoutram, line 22: Just wondering if there's a particular reason for setting the updated date here on entities that are only just being created?

Looking at the way the date is assigned to both I suppose these two dates could differ by a tiny fraction making it more difficult to report on entities that have been created but not updated yet.

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Timestamps.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public interface ICreatedTimestamp
+    {
+        DateTime? Created { get; set; }
+    }
+
+    public interface IUpdatedTimestamp
+    {
+        DateTime? Updated { get; set; }
+    }
```
benoutram, line 14: I suppose it's more flexible to keep these public but I'm struggling to think of a time when you would deliberately implement only one of these interfaces rather than `ICreatedUpdatedTimestamps` if you were adding these fields to either a new or existing type today.

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Timestamps.cs
```diff
@@ -0,0 +1,24 @@
+#nullable enable
+using System;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Model
+{
+    public interface ICreatedTimestamp
+    {
+        DateTime? Created { get; set; }
```
benoutram, line 8: I was wondering how you would solve the problem of some entities having a not-nullable Created field in the database.

For newly added types I think we would like this field to be not-null in the database.

If this interface is applied to any existing types with a `Created` field that is not optional does it cause a database migration converting the field from not-null to allow null?

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Timestamps.cs
ntsim: There might be cases where there is deliberately immutable data and there isn't any point in having both of these. `Subject` comes to mind as it's not supposed to be modifiable.

Additionally, having only the `ICreatedUpdatedTimestamps` means that we would have to create a new migration to be able to use the interface in instances where we don't already have both columns in place. This might be undesirable until we make a concerted effort to fix all of our tables in EES-3138.

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Timestamps.cs
ntsim: As discussed on Slack, this has been fixed by adding type parameters to the interfaces to let the consumers decide the type of their timestamp field. This is also more flexible as it allows them to use different types like `DateTimeOffset` if required.

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Timestamps.cs
benoutram: Cool, good examples and a good reason to keep them public then! 👍

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Timestamps.cs
benoutram: Cool, changes to allow different types of Date look good!

benoutram (approved):

PR #3135, opened 2022-01-31
EES-1477 prevent bugs when change table tool subject and location

amyb-hiveit:
Fixes two table tool bugs:
1. After changing location, can move past time period step without selecting a valid time period - fixed by checking that the selected time period is still available and and resetting the selection if not.

2. Can create table despite not selecting an indicator - fixed by resetting selections when the subject is changed.

## Related changes

n/a

## Screenshots

n/a

### DIFF
--- a/src/explore-education-statistics-common/src/modules/table-tool/components/TableToolWizard.tsx
+++ b/src/explore-education-statistics-common/src/modules/table-tool/components/TableToolWizard.tsx
@@ -172,8 +172,11 @@ const TableToolWizard = ({
 
     updateState(draft => {
       draft.subjectMeta = nextSubjectMeta;
-
       draft.query.subjectId = selectedSubjectId;
+      draft.query.indicators = [];
+      draft.query.filters = [];
+      draft.query.locations = {};
+      draft.query.timePeriod = undefined;
     });
   };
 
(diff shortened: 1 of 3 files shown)


src/explore-education-statistics-common/src/modules/table-tool/components/TableToolWizard.tsx
```diff
@@ -194,11 +197,27 @@ const TableToolWizard = ({
...
+      ) !== -1;
 
     updateState(draft => {
       draft.subjectMeta.timePeriod = nextSubjectMeta.timePeriod;
-
       draft.query.locations = locations;
+      draft.query.timePeriod =
+        hasStartTimePeriod && hasEndTimePeriod
+          ? state.query.timePeriod
+          : undefined;
```
ntsim, line 220: How about if only one of the time periods exists in the meta? Would suggest that we still pre-populate the start/end date if it exists

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/TableToolWizard.test.tsx
```diff
@@ -311,4 +320,148 @@ describe('TableToolWizard', () => {
...
+      hidden: true,
+    });
+    expect(filterCheckboxes[0]).not.toBeChecked();
+
+    const indicatorCheckboxes = within(
+      screen.getByRole('group', {
+        name: 'Indicators',
+      }),
+    ).getAllByRole('checkbox', {});
+    expect(indicatorCheckboxes[0]).not.toBeChecked();
```
ntsim, line 416: This is interesting. Seems like if we're clearing time periods, we should also put in some stuff to clear away any filters and indicators too when submitting the time period step?

ntsim (changes requested):

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/TableToolWizard.test.tsx
amyb-hiveit: Not sure what you mean, this test relates to resetting all the subsequent steps when you change the subject. It's only when the subject changes that this should happen, if you change the time period the filters and indicators should still be selected.

ntsim: I mean that we've put in some logic to only select time periods if they still exist in subject meta once it's been changed (when moving from locations -> time periods), however, it doesn't look like we've put in anything to remove any filter/indicator options when moving from time periods -> filters/indicators.

On the basis that we're filtering out time periods, it might make sense to also do this for filters/indicators too?

ntsim (commented):
Few days late, but I forgot to submit the review :grimacing:

PR #3136, opened 2022-01-31
EES-3141 fix broken permalink 

amyb-hiveit:
Added defensive code to prevent the permalink page breaking when the configuration for tableHeaders rows returned from getPermalink is an empty array. 

A jira ticket will be raised separately to investigate why the array is empty, once that's resolved we should be able to remove this check.

### DIFF
--- a/src/explore-education-statistics-common/src/modules/table-tool/components/TimePeriodDataTable.tsx
+++ b/src/explore-education-statistics-common/src/modules/table-tool/components/TimePeriodDataTable.tsx
@@ -112,6 +112,14 @@ const TimePeriodDataTableInternal = forwardRef<HTMLElement, Props>(
   ) {
     const { subjectMeta, results } = fullTable;
 
+    if (!tableHeadersConfig.rows.length) {
+      return (
+        <WarningMessage>
+          There was a problem rendering the table.
+        </WarningMessage>
+      );
+    }
+
     if (results.length === 0) {
       return (
         <WarningMessage>
@@ -296,7 +304,7 @@ const TimePeriodDataTableInternal = forwardRef<HTMLElement, Props>(
 
     // Only need to use first row's columnFilters
     // as they are the same for every row.
-    const columnHeaders = (filteredCartesian[0] ?? []).reduce<Header[]>(
+    const columnHeaders = filteredCartesian[0].reduce<Header[]>(
       (acc, column) => {
         const filters = optimizeFilters(column.columnFilters, [
           ...tableHeadersConfig.columnGroups,
--- a/src/explore-education-statistics-common/src/modules/table-tool/components/TimePeriodDataTable.tsx
+++ b/src/explore-education-statistics-common/src/modules/table-tool/components/TimePeriodDataTable.tsx
@@ -112,14 +112,6 @@ const TimePeriodDataTableInternal = forwardRef<HTMLElement, Props>(
   ) {
     const { subjectMeta, results } = fullTable;
 
-    if (!tableHeadersConfig.rows.length) {
-      return (
-        <WarningMessage>
-          There was a problem rendering the table.
-        </WarningMessage>
-      );
-    }
-
     if (results.length === 0) {
       return (
         <WarningMessage>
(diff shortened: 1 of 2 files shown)


src/explore-education-statistics-common/src/modules/table-tool/components/TimePeriodDataTable.tsx
```diff
@@ -296,17 +296,16 @@ const TimePeriodDataTableInternal = forwardRef<HTMLElement, Props>(
...
+    const columnHeaders = filteredCartesian.length
+      ? filteredCartesian[0].reduce<Header[]>((acc, column) => {
+          const filters = optimizeFilters(column.columnFilters, [
+            ...tableHeadersConfig.columnGroups,
+            tableHeadersConfig.columns,
+          ]);
+
+          return addFilters(acc, filters);
+        }, [])
+      : [];
```
ntsim, line 308: Would suggest writing this to avoid the extra ternary:

```ts
const columnHeaders = (filteredCartesian[0] ?? []).reduce(...);
```

Isn't this going to end up with no column headers and a broken table anyway?

amyb-hiveit: You get the page rendering without the table. 
Capture

ntsim: Hmmm, not sure it's a good idea to have a caption for a completely empty table. Would suggest that we should render an error message to say that there was a problem rendering the table. 

Interestingly, this is what should already have been happening as `TimePeriodDataTable` wraps `TimePeriodDataTableInternal` with an error boundary. Why was this throwing a page error?

amyb-hiveit: Good point, I've changed it to show a warning message instead.

I don't think error boundaries work with server rendered pages (this component is used in the admin as well where presumably the boundary would work).

src/explore-education-statistics-common/src/modules/table-tool/components/TimePeriodDataTable.tsx
```diff
@@ -112,6 +112,14 @@ const TimePeriodDataTableInternal = forwardRef<HTMLElement, Props>(
   ) {
     const { subjectMeta, results } = fullTable;
 
+    if (!tableHeadersConfig.rows.length) {
+      return (
+        <WarningMessage>
+          There was a problem rendering the table.
+        </WarningMessage>
```
ntsim, line 119: Would suggest adding better monitoring whenever a broken table might render. Doing this via Google Analytics would probably be the best place to start (we could add Application Insights tracking later if we think it might be useful). We could create a re-usable component like this:

```tsx
const LogGAEvent = (event: AnalyticsEvent) => {
  useEffect(() => {
    googleAnalyticsService.logEvent(event);
    // Probably a good idea to add in some
    // other stuff to stop logging on every render (you can figure that out)
  }, [event]);

  return null;
}
```

Might want to check with Cam about what the best categorisation/labelling for this might be.

We could then use this component in any of the places where we render `WarningMessage` in `TimePeriodDataTable` (this includes the error boundary) like:

```tsx
<>
  <LogGAEvent ...theEvent />
  <WarningMessage>
    There was a problem rendering the table.
  </WarningMessage>
</>
```

ntsim (changes requested):

src/explore-education-statistics-common/src/modules/table-tool/components/TimePeriodDataTable.tsx
ntsim: Ah yes, that really sucks on Next's part! An alternative then might be to wrap it all with a try/catch in `TimePeriodDataTableInternal`, although I'm not 100% sure if this will work or not.

ntsim (dismissed):

amyb-hiveit:
@ntsim I've wrapped the innards of it in a try catch and removed the error boundary. The error is sent to GA as an exception. Have also moved a couple of functions up out of the component as they didn't need to be in it and it makes it clearer to read.
Re-review when you have a min please!

ntsim (approved):
Looks good :+1:

PR #3148, opened 2022-02-03
EES-3124 Display Publications grouped by their type on the public Find Statistics page

benoutram:
⚠️This requires the tree JSON files to be cleared from the public cache in each environment using `DELETE /api/bau/public-cache/trees` .

This PR builds upon the work of https://github.com/dfe-analytical-services/explore-education-statistics/pull/3119 and https://github.com/dfe-analytical-services/explore-education-statistics/pull/3137 and is the front-end part to the feature EES-3038 which inserts a new section into the hierarchy of Topics/Themes/Publications on the public Find Statistics page to group Publication's by their latest release type.

We introduce a new Publication type based on the set of Release types. Publications are grouped by their type and displayed under a new 'Publication type' heading.

In the back-end we map from the latest Release type to Publication type as follows.
- 'National statistics' and 'Official statistics' are mapped to publication type 'National and official statistics'.
- Publications with no Releases which exist because they have a legacy URL have a type 'Not yet on this service'.

```
Latest release type        Publication type label
===========================================================
National statistics        National and official statistics
Official statistics        National and official statistics
Experimental statistics    Experimental statistics
Ad hoc statistics          Ad hoc statistics
Management information     Management information
None                       Not yet on this service
```

Publication types are ordered within a topic according to the reference order 'National and official statistics' > 'Experimental statistics' > 'Ad hoc statistics' > 'Management information' > 'Not yet on this service'.

An example of an issue this fixes is where one off Ad hoc statistics releases in a publication like 'Free school meals: Autumn term'  in 'Pupils and schools' -> 'School and pupil numbers' currently appears above the 'Schools, pupils and their characteristics' publication which is a series of regular National statistics releases and should take higher precedence than the ad hoc publication.

## Related changes

- Update the Find Statistics page snapshot generated by the Python `create_snapshot.py` script to match the new hierarchy.
- Fixes the capitalisation of Release types heading on the 'Create new release' page from `Release Types` to `Release types`.
- Add a Robot UI test to check the Release type options on the 'Create new release' page.
- Add a Robot UI test to amend the Release type as part of amending a Release.
- Change Robot UI tests to check publications appear under the correct publication type heading on the 'Find Statistics' page.
- Fixes the capitalisation of Release type labels so that the second word of each type is now lower cased e.g. `National statistics` rather than `National Statistics`.
- Fixes the `id` attribute of `h3` in every Publication on the Find Statistics page which currently has an erroneous brace in it e.g. `<h3 id="publication-heading-657b1484-0369-4a0e-873a-367b79a48c35}"`

## Screenshots

image


tests/robot-tests/scripts/create_snapshots.py
```diff
@@ -34,21 +34,31 @@ def create_find_statistics_snapshot(public_url) -> str:
...
+            for publication_type_html in publication_types:
+                publication_type_key = publication_type_html['id'].replace('publication-type-heading-', '')
+                publication_type = {
+                    'publication_type_heading': publication_type_html.string,
+                    'publications': [],
                 }
 
-                topic['publications'].append(publication)
+                publications = topic_html.select(
+                    f'ul[data-testid="publications-list-{publication_type_key}"] > li') or []
```
markhiveit, line 53: I was expecting you to use `publication_type_html` here - then potentially you can avoid all the `publication_type_key` stuff. Maybe it doesn't work?

ntsim: Personally, I'm not against wrapping stuff with a div with `data-testid` where necessary. If it makes it easier, go ahead, but of course, avoid it if there are better alternatives :smile:

I'd also say that general preference is towards `data-testid` instead of `id` as we can avoid potential screen reader issues from having multiple elements with the same id. It does make sense to use `id` for headings as it's fairly common to see this on other websites (e.g. in documentation) so that you can easily navigate back to a section by putting the id in the URL hash.

src/explore-education-statistics-frontend/src/modules/find-statistics/components/PublicationList.tsx
```diff
@@ -4,19 +4,20 @@ import React from 'react';
...
-    <ul className="govuk-!-margin-top-0">
+    <ul className="govuk-!-margin-top-0" data-testid={testId}>
       {publications.map(({ id, legacyPublicationUrl, slug, title }) => (
         <li key={id} className="govuk-!-margin-bottom-0">
-          <h3
-            id={`publication-heading-${id}}`}
-            className="govuk-heading-s govuk-!-margin-bottom-0"
+          <h4
+            id={`publication-heading-${id}`}
+            className="govuk-heading-s govuk-!-font-weight-regular govuk-!-margin-bottom-0"
```
ntsim, line 17: Is it a good idea to make this heading non-bold? Kinda feels to me like it blending into the rest of the text. 

Were you worried it might be indistinguishable from the publication type heading? In which case, it might better to make that a bit bigger like so:

image

ntsim (changes requested):

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
```diff
@@ -83,7 +143,22 @@ const FindStatisticsPage: NextPage<Props> = ({ themes = [] }) => {
...
+                            >
+                              {publicationTypes[type]}
+                            </h3>
+                            <PublicationList
+                              publications={group}
+                              testId={`publications-list-${type}`}
+                            />
+                          </Fragment>
+                        ),
+                      )}
```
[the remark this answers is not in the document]
benoutram: Pushed this down into `PublicationList` now, thanks.

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
```diff
@@ -1,21 +1,81 @@
...
+    ...releaseTypesExcludingNationalAndOfficalStatistics
+  } = releaseTypes;
+
+  const publicationTypes = {
+    ...releaseTypesExcludingNationalAndOfficalStatistics,
+    NationalAndOfficial: 'National and official statistics',
+    Legacy: 'Not yet on this service',
+  } as const;
+
+  type PublicationType = keyof typeof publicationTypes;
```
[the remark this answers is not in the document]
benoutram: Pushed this into `publicationType.ts` now, thanks. I wasn't sure if these filenames should be plural or singular? See existing `releaseType.ts`.

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
```diff
@@ -1,21 +1,81 @@
...
+
+  const releaseTypeToPublicationType = (
+    releaseType?: ReleaseType,
+  ): PublicationType => {
+    switch (releaseType) {
+      case 'NationalStatistics':
+      case 'OfficialStatistics':
+        return 'NationalAndOfficial';
+      case undefined:
```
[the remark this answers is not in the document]
benoutram: Pushed mapping to back-end after creating a new `PublicationType` enum which simplifies things. Publication tree view model now has a `Type` field rather than `LatestReleaseType`.

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
```diff
@@ -1,21 +1,81 @@
...
+    }
+  };
+
+  const publicationTypeReferenceOrder: PublicationType[] = [
+    'NationalAndOfficial',
+    'ExperimentalStatistics',
+    'AdHocStatistics',
+    'ManagementInformation',
+    'Legacy',
+  ];
```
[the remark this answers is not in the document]
benoutram: This is moved to `PublicationList` now. As it's only the order for this component I've left it there for now.

src/explore-education-statistics-frontend/src/modules/find-statistics/__tests__/FindStatisticsPage.test.tsx
```diff
@@ -0,0 +1,293 @@
...
+    expect(
+      screen.getByText('No data currently published.', { selector: 'div' }),
+    ).toBeInTheDocument();
+  });
+
+  test('renders the accordion correctly with multiple themes, topics and publications', async () => {
+    const { container } = render(<FindStatisticsPage themes={testThemes} />);
+
+    const contentAccordion = screen.getByTestId('accordion');
+    expect(contentAccordion.innerHTML).toMatchSnapshot();
```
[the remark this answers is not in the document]
benoutram: Thanks I did spot that it was a bit unwieldy and I've removed it now. Good to know where they can still be useful for.

src/explore-education-statistics-frontend/src/modules/find-statistics/components/PublicationList.tsx
benoutram: Yes, I went through a few iterations, one where they were both the same size and bold and they couldn't be distinguished like you say.

The suggestion to make the Publication title not bold came from @cjrace.

I've changed it how you suggested and will re-issue the example screenshot to @cjrace and @lauraselby, thanks.

tests/robot-tests/scripts/create_snapshots.py
benoutram: Thanks, for some reason I thought it would be bad to that. It certainly makes things a lot easier here in the snapshot script where the code now follows the same convention, and also in the Jest tests.

src/explore-education-statistics-frontend/src/modules/find-statistics/components/PublicationList.tsx
benoutram: Update: Spoke to @lauraselby who in turn has checked with @cjrace and they prefer the original so I have changed it back.

src/explore-education-statistics-frontend/src/modules/find-statistics/components/PublicationList.tsx
```diff
@@ -1,58 +1,100 @@
...
+  'ManagementInformation',
+  'Legacy',
+];
+
+const groupPublicationsByType = (publications: PublicationSummary[]) =>
+  Object.entries(
+    publications.reduce((acc, publication) => {
+      (acc[publication.type] ??= []).push(publication);
+      return acc;
+    }, {} as { [key in PublicationType]?: PublicationSummary[] }),
```
ntsim, line 26: Minor thing really, but realized this can be simplified using Lodash's `groupBy`:

```ts
import groupBy from 'lodash/groupBy';

Object.entries(groupBy(publications, publication => publication.type))
```

Side note whilst we're here - always import the specific Lodash function:

```ts
import { groupBy } from 'lodash'; ❌
import groupBy from 'lodash/groupBy'; ✅
```

The former imports the entirety of Lodash and would increase the final bundle size considerably.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ThemeService.cs
```diff
@@ -110,17 +110,17 @@ private async Task<bool> HasAnyDataFiles(Release release)
         private static PublicationTreeNode BuildPublicationNode(Publication publication)
         {
             var latestRelease = publication.LatestPublishedRelease();
-            var legacyPublicationUrlIgnored = latestRelease != null;
+            var type = latestRelease == null ? PublicationType.Legacy : GetPublicationType(latestRelease.Type);
```
ntsim, line 113: Couldn't we bundle this all up into `GetPublicationType`'s switch expression:

```cs
private static PublicationType GetPublicationType(ReleaseType? releaseType)
{
   return releaseType switch
   {
     ....
     null => PublicationType.Legacy,
   };
}
```

src/explore-education-statistics-frontend/src/modules/find-statistics/components/PublicationList.tsx
benoutram: I originally used Lodash but switched to the native reduce thinking it would assist getting the typed keys rather than strings. I found this didn't work out but decided to stick with the native reduce rather than importing the Lodash function.

I've switched to using Lodash now to simplify things, thanks.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ThemeService.cs
benoutram: I wasn't sure which way to go with this and opted for doing it outside prior to calling this method but you've convinced me to do it within the `switch` now. 😄  Made the parameter optional in the latest commit.

ntsim (approved):

PR #3149, opened 2022-02-03
EES-2779 group LAs in map dropdown

amyb-hiveit:
- Groups Local Authorities by region in the Map block location dropdown
- If other location levels are also in the locations then they are grouped by level
- If there are no LAs in the locations then the list is not grouped and remains as it is currently.
- Also changed the 'None selected' option to be added as a placeholder on the FormSelect instead of being added to the options array.

## Screenshots

groups

### DIFF
--- a/src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
+++ b/src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
@@ -36,7 +36,6 @@ import classNames from 'classnames';
 import { Feature, FeatureCollection, Geometry } from 'geojson';
 import { Layer, Path, PathOptions, Polyline } from 'leaflet';
 import clamp from 'lodash/clamp';
-import { groupBy } from 'lodash';
 import keyBy from 'lodash/keyBy';
 import orderBy from 'lodash/orderBy';
 import times from 'lodash/times';
@@ -352,11 +351,15 @@ export const MapBlockInternal = ({
     );
   }, [dataSetCategoryConfigs]);
 
-  const shouldGroupLocationOptions = dataSetCategories.some(
-    element => element.filter.level === 'localAuthority',
-  );
+  const shouldGroupLocationOptions = useMemo(() => {
+    return dataSetCategories.some(
+      element =>
+        element.filter.level === 'localAuthority' ||
+        element.filter.level === 'localAuthorityDistrict',
+    );
+  }, [dataSetCategories]);
 
-  // If no LAs don't group the locations.
+  // If there are no LAs or LADs don't group the locations.
   const locationOptions = useMemo(() => {
     if (shouldGroupLocationOptions) {
       return undefined;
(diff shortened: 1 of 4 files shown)


src/explore-education-statistics-common/src/modules/charts/components/__tests__/MapBlockInternal.test.tsx
```diff
@@ -15,7 +17,7 @@ import produce from 'immer';
 import React from 'react';
 import userEvent from '@testing-library/user-event';
 
-describe('MapBlock', () => {
+describe('MapBlockInternal', () => {
```
amyb-hiveit, line 20: I've moved the tests to be on MapBlockInternal as that's where the code actually is and I've seen an intermittent failure on the first test where it wasn't rendering the map. Probably a long shot but maybe the tests have problems with the dynamic import of MapBlockInternal so have bypassed that.

amyb-hiveit, line 20: I've moved the tests to be on MapBlockInternal as that's where the code actually is and I've seen an intermittent failure on the first test where it wasn't rendering the map. Probably a long shot but maybe the tests have problems with the dynamic import of MapBlockInternal so have bypassed that.

src/explore-education-statistics-common/src/modules/charts/components/__tests__/MapBlockInternal.test.tsx
```diff
@@ -334,28 +318,89 @@ describe('MapBlock', () => {
...
           screen.getByLabelText('2. Select a location'),
         ).toBeInTheDocument();
       });
+      const select = screen.getByLabelText('2. Select a location');
+
+      expect(select.children[0]).toHaveTextContent('None selected');
+
+      const groups = within(select).getAllByRole('group');
+      expect(groups).toHaveLength(4);
+
```
ntsim, line 385: We should probably do some checks to see if the group labels are correct (and not just the options themselves)

src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
```diff
@@ -351,17 +352,50 @@ export const MapBlockInternal = ({
     );
   }, [dataSetCategoryConfigs]);
 
+  const shouldGroupLocationOptions = dataSetCategories.some(
+    element => element.filter.level === 'localAuthority',
```
ntsim, line 356: What about if we have LAD's? They're essentially the same, and we already do this grouping in table tool

ntsim (changes requested):

src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
```diff
@@ -351,17 +352,50 @@ export const MapBlockInternal = ({
...
         value: dataSetCategory.filter.id,
       })),
       ['label'],
     );
-    locations.unshift({ label: 'None selected', value: '' });
-    return locations;
-  }, [dataSetCategories]);
+  }, [dataSetCategories, shouldGroupLocationOptions]);
+
+  // If has LAs, group them by region and group any others by level
```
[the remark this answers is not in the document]
amyb-hiveit: Not sure what you mean. This is a comment for the code when there are LAs.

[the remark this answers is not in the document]
ntsim: Oh sorry my bad, I actually meant to say 'If there are LAs'

ntsim:
Looks good, but looks like the build is failing @amyb-hiveit

ntsim (approved):

PR #3150, opened 2022-02-03
EES-2999 - added fix for provider and school level Location queries

duncan-at-hiveit:
This PR:
- fixes an existing bug in LocationPredicateBuilder (formerly ObservationPredicateBuilder) whereby School and Provider were not recognised as being able to provide Location Attributes and so did not add any clauses to the overall Location predicate.
- This caused the work in 2999 to fetch too many Locations when Provider or School codes were provided, thus returning table results that were not valid for the ranges of Providers and Schools chosen in queries.

I've also added in throws for any "misuse" of the LocationPredicateBuilder i.e. any scenarios that would not actually add any genuine clauses to the location predicate.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Services/ObservationService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Services/ObservationService.cs
@@ -51,7 +51,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Services
             {
                 locationIds = await _context
                     .Location
-                    .Where(LocationPredicateBuilder.Build(query.LocationIds, query.Locations))
+                    .Where(LocationPredicateBuilder.Build(query.LocationIds?.ToList(), query.Locations))
                     .Select(location => location.Id)
                     .ToListAsync(cancellationToken);
             }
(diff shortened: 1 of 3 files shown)


src/GovUk.Education.ExploreEducationStatistics.Data.Services/ObservationService.cs
```diff
@@ -51,7 +51,7 @@ public class ObservationService : AbstractRepository<Observation, long>, IObserv
             {
                 locationIds = await _context
                     .Location
-                    .Where(LocationPredicateBuilder.Build(query.LocationIds, query.Locations))
+                    .Where(LocationPredicateBuilder.Build(query.LocationIds?.ToList(), query.Locations))
```
ntsim, line 54: It seems like we're using `IEnumerable` in most of the `ObservationQueryContext` fields, but it kinda feels like we could replace most of these with `Set` or `List` types (or their interfaces) to avoid needing to fix potential multiple enumerations issues.

I'd probably lean toward `Set` to have built-in dupe removal, but I guess that depends on if order is important (there's always `SortedSet`).

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ObservationService.cs
duncan-at-hiveit: I think this is a great idea, and something I've been adding more of in 2999 in fact.  Had a discussion with Mark about it too and the reasoning behind it.  What a shame that IQueryable and IList both extend IEnumerable!

Def happy to extend this pattern beyond this piece of work :)

PR #3151, opened 2022-02-04
EES-2999 Improve get_data_block_responses script

markhiveit:
This PR makes a few minor improvements to the get_data_block_responses script.

- All console output of the script is now also saved in a "console_output" file.
- Removed "fails" directory - all data block files are saved in the same directory now.


duncan-at-hiveit (approved):
Looks great!

PR #3154, opened 2022-02-07
Ees 2990 allow change in app registrations for existing users

duncan-at-hiveit:
This PR:
* adds support for us to migrate from one App Registration to another on environments.
* adds Unit Testing around the login process.

The effects of moving App Registration (even when pointing against the same AD instance) means that the ProviderKey for a single user (unique identifier for a user provided form the login process) differ between the old and the new App Registration.

This means we must handle the ability for a single user (identified by email address) to have more than one set of ProviderKey information.  This supports not only having multiple App Registrations, but also multiple external OpenID Connect Identity Providers as well where we trust the Identity Provider and its ability to securely identify a user by their email address.  Thus we could support login from DfE's AD and also Google auth, for instance.

We can trust the details coming from the external Identity Provider because it is the one set up to communicate with in the appsettings.xml file.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Areas/Identity/Pages/Account/ExternalLoginTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Areas/Identity/Pages/Account/ExternalLoginTests.cs
@@ -9,7 +9,7 @@ using GovUk.Education.ExploreEducationStatistics.Admin.Areas.Identity.Data;
 using GovUk.Education.ExploreEducationStatistics.Admin.Areas.Identity.Data.Models;
 using GovUk.Education.ExploreEducationStatistics.Admin.Areas.Identity.Pages.Account;
 using GovUk.Education.ExploreEducationStatistics.Admin.Models;
-using GovUk.Education.ExploreEducationStatistics.Admin.Tests.Security.AuthorizationHandlers.Utils;
+using GovUk.Education.ExploreEducationStatistics.Admin.Tests.Security.Utils;
 using GovUk.Education.ExploreEducationStatistics.Content.Model;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
 using Microsoft.AspNetCore.Identity;
@@ -148,7 +148,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Areas.Identity.
         [Fact]
         public async Task Login_EmailClaimNotAvailable()
         {
-            var claimsPrincipal = AuthorizationHandlersTestUtil.CreateClaimsPrincipal(
+            var claimsPrincipal = ClaimsPrincipalUtils.CreateClaimsPrincipal(
                 Guid.NewGuid(),
                 new Claim(ClaimTypes.GivenName, "FirstName"),
                 new Claim(ClaimTypes.Surname, "LastName"));
@@ -532,7 +532,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Areas.Identity.
 
         private static ClaimsPrincipal CreateClaimsPrincipal(string emailAddress, string firstName, string lastName)
         {
-            return AuthorizationHandlersTestUtil.CreateClaimsPrincipal(
+            return ClaimsPrincipalUtils.CreateClaimsPrincipal(
                 Guid.NewGuid(),
                 new Claim(ClaimTypes.Email, emailAddress),
                 new Claim(ClaimTypes.GivenName, firstName),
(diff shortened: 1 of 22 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Areas/Identity/Pages/Account/ExternalLoginTests.cs
```diff
@@ -0,0 +1,576 @@
+#nullable enable
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using Castle.Core.Internal;
```
benoutram, line 7: This import looks a bit suspect. I think it's coming from using `IsNullOrEmpty` rather than our own.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Areas/Identity/Pages/Account/ExternalLoginTests.cs
```diff
@@ -0,0 +1,576 @@
...
+                
+            AssertRedirectedToLoginPageUponFailure(result, loginService);
+        }
+
+        [Fact]
+        public async Task Login_InvitedUser_Success()
+        {
+            var contextId = Guid.NewGuid().ToString();
+
+            var email = "inviteduser@example.com";
```
benoutram, line 219: Could use `const string` rather than `var` here and in a few other places throughout these tests to indicate these variables don't get reassigned.

src/GovUk.Education.ExploreEducationStatistics.Admin/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
```diff
@@ -61,196 +68,429 @@ public IActionResult OnPost(string provider, string returnUrl = null)
             return new ChallengeResult(provider, properties);
         }
 
-        public async Task<IActionResult> OnGetCallbackAsync(string returnUrl = null, string remoteError = null)
+        public async Task<IActionResult> OnGetCallbackAsync(string providedReturnUrl = null, string remoteError = null)
```
benoutram, line 71: Found this a bit difficult to follow with all of the branching `return` statements, so not sure if that can be avoided, but nice work with the comments! 🙂 .

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Areas/Identity/Pages/Account/ExternalLoginTests.cs
```diff
@@ -0,0 +1,576 @@
...
+using Xunit;
+using static GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.DbUtils;
+using static GovUk.Education.ExploreEducationStatistics.Common.Services.CollectionUtils;
+using static GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils.MockUtils;
+using static Moq.MockBehavior;
+using SignInResult = Microsoft.AspNetCore.Identity.SignInResult;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Areas.Identity.Pages.Account
+{
+    public class ExternalLoginTests
```
benoutram, line 29: Great to get all these tests written! 🙂

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Areas/Identity/Pages/Account/ExternalLoginTests.cs
duncan-at-hiveit: Good spot!  Changed to use ours

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Areas/Identity/Pages/Account/ExternalLoginTests.cs
duncan-at-hiveit: cool - done thanks!

src/GovUk.Education.ExploreEducationStatistics.Admin/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
duncan-at-hiveit: Yeah, I did try to make it as painless as possible, but there's too many routes that we can take I think and each one requiring subtly different handling.  I figured that with the test coverage, we could hopefully avoid accidentally shooting ourselves in the foot when we make changes here.

Additionally I've raised https://dfedigital.atlassian.net/browse/EES-3165 in the backlog to imrove the coverage around the "invited users" code branches, as it's not as extensive as the "new user" code branches.

benoutram (approved):

PR #3163, opened 2022-02-09
EES-2556 reorder datasets

amyb-hiveit:
Version 2 of reordering datasets, see origin PR https://github.com/dfe-analytical-services/explore-education-statistics/pull/3124 and PR to undo it https://github.com/dfe-analytical-services/explore-education-statistics/pull/3158/files for context.

This adds reordering to chart data sets to allow users to control the ordering in the chart.

The original solution would potentially have caused existing charts with groups in the major axis to change. To avoid this I've added an order property to data sets (don't worry, @markhiveit told me what to do). If the new order property is present then reordering the data sets also reorders the dataSetCategories on the basis of their order in the data sets. If no order is present then the previous ordering is used.

For testing I suggest setting up some charts on dev before this is merged so they can be checked to confirm they haven't changed.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartDataSet.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartDataSet.cs
@@ -11,6 +11,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Model.Chart
         public List<Guid> Filters = new List<Guid>();
         public ChartDataSetLocation Location;
         public string TimePeriod;
+        public int? Order;
 
         // TODO EES-1649 Migrate data set configs to legend item configs
         public ChartDataSetConfiguration Config;
--- a/src/explore-education-statistics-common/src/modules/charts/util/__tests__/createDataSetCategories.test.ts
+++ b/src/explore-education-statistics-common/src/modules/charts/util/__tests__/createDataSetCategories.test.ts
@@ -1340,4 +1340,339 @@ describe('createDataSetCategories', () => {
     expect(dataSetCategories[0].filter.label).toBe('State-funded primary');
     expect(dataSetCategories[1].filter.label).toBe('State-funded secondary');
   });
+
+  describe('ordering data sets with `order` property', () => {
+    const testOrderedAxisConfiguration: AxisConfiguration = {
+      type: 'major',
+      groupBy: 'timePeriod',
+      sortAsc: true,
+      dataSets: [
+        {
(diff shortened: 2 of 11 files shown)


src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartDataSet.cs
```diff
@@ -11,6 +11,7 @@ public class ChartDataSet
         public List<Guid> Filters = new List<Guid>();
         public ChartDataSetLocation Location;
         public string TimePeriod;
+        public int Order;
```
ntsim, line 14: As not all charts will have this currently, we should probably set the type to `int?`, otherwise it can end up being defaulted to 0 (that's the default for `int` in C#).

Have we checked that the `order` is coming back as null/undefined from the API for charts configured with the older data sets?

src/explore-education-statistics-common/src/modules/charts/util/__tests__/createDataSetCategories.test.ts
```diff
@@ -77,8 +78,18 @@ describe('createDataSetCategories', () => {
...
+          label: '2014/15',
+          order: 0,
+          year: 2014,
+        }),
+        new TimePeriodFilter({
+          code: 'AY',
+          label: '2015/16',
+          order: 1,
+          year: 2015,
+        }),
```
ntsim, line 92: This doesn't look right. We're working with a `TableDataResponse`, which should be the pure API response and shouldn't have any class versions of the filters as it won't have been passed into `mapFullTable` yet.

`TableDataReponse` also uses `TimePeriodOption`, which shouldn't have an  `order` field as the order is inferred from the ordering of the `timePeriodRange` array itself.

src/explore-education-statistics-common/src/modules/charts/util/createDataSetCategories.ts
```diff
@@ -283,6 +303,10 @@ export default function createDataSetCategories(
   results: TableDataResult[],
   meta: FullTableMeta,
 ): DataSetCategory[] {
+  // Check if the datasets have ordering applied, old charts won't so we can use this to preserve their ordering.
+  const hasDataSetOrder = axisConfiguration.dataSets.some(
+    dataSet => !!dataSet.order,
+  );
```
ntsim, line 309: No need to do this here, we can do it inside of `getSortedDataSetCategoriesRange`

ntsim (changes requested):

ntsim (approved):

PR #3170, opened 2022-02-14
EES-3170 Stop log spam when running Azure Functions locally

markhiveit:
This PR prevents log spam we were getting when running Azure Functions locally via azure-functions-core-tools. The log spam looked like this:
```
[2022-02-14T16:37:47.611Z] Request [b12a9236-e309-4c20-bb7c-54f06c9c824c] PUT http://data-storage:10000/devstoreaccount1/azure-webjobs-hosts/locks/baptistin-479599971/host?comp=lease
[2022-02-14T16:37:47.613Z] x-ms-lease-action:release
[2022-02-14T16:37:47.615Z] x-ms-lease-id:00000000-0000-0000-0000-0000a7df7835
[2022-02-14T16:37:47.617Z] x-ms-version:2020-08-04
[2022-02-14T16:37:47.618Z] Accept:application/xml
[2022-02-14T16:37:47.620Z] x-ms-client-request-id:b12a9236-e309-4c20-bb7c-54f06c9c824c
[2022-02-14T16:37:47.622Z] x-ms-return-client-request-id:true
[2022-02-14T16:37:47.624Z] User-Agent:azsdk-net-Storage.Blobs/12.9.0,(.NET 6.0.0-rc.2.21480.5; Microsoft Windows 10.0.19042)
[2022-02-14T16:37:47.626Z] x-ms-date:Mon, 14 Feb 2022 16:37:47 GMT
[2022-02-14T16:37:47.628Z] Authorization:REDACTED
[2022-02-14T16:37:47.630Z] client assembly: Azure.Storage.Blobs
```

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Processor/host.json
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Processor/host.json
@@ -3,8 +3,10 @@
   "functionTimeout": "04:00:00",
   "logging": {
     "logLevel": {
-      "default": "Information",
-      "Microsoft": "Warning"
+      "default": "Warning",
+      "Microsoft": "Warning",
+      "Function": "Information",
+      "GovUk.Education.ExploreEducationStatistics": "Information"
     }
   },
   "extensions": {
--- a/src/GovUk.Education.ExploreEducationStatistics.Notifier/host.json
+++ b/src/GovUk.Education.ExploreEducationStatistics.Notifier/host.json
@@ -2,7 +2,10 @@
     "version": "2.0",
     "logging": {
         "logLevel": {
-            "default": "Information"
+            "default": "Warning",
+            "Microsoft": "Warning",
+            "Function": "Information",
+            "GovUk.Education.ExploreEducationStatistics": "Information"
         }
     }
 }
--- a/src/GovUk.Education.ExploreEducationStatistics.Publisher/host.json
+++ b/src/GovUk.Education.ExploreEducationStatistics.Publisher/host.json
@@ -2,8 +2,10 @@
   "version": "2.0",
   "logging": {
     "logLevel": {
-      "default": "Information",
-      "Microsoft": "Warning"
+      "default": "Warning",
+      "Microsoft": "Warning",
+      "Function": "Information",
+      "GovUk.Education.ExploreEducationStatistics": "Information"
     }
   },
   "extensions": {


src/GovUk.Education.ExploreEducationStatistics.Data.Processor/local.settings.json
```diff
@@ -4,7 +4,10 @@
     "AzureWebJobsStorage": "UseDevelopmentStorage=true;DevelopmentStorageProxyUri=http://data-storage",
     "CoreStorage": "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://data-storage:10000/devstoreaccount1;QueueEndpoint=http://data-storage:10001/devstoreaccount1;TableEndpoint=http://data-storage:10002/devstoreaccount1",
     "FUNCTIONS_WORKER_RUNTIME": "dotnet",
-    "RowsPerBatch": 3000
+    "RowsPerBatch": 3000,
+    "AzureFunctionsJobHost__logging__logLevel__Default": "None",
+    "AzureFunctionsJobHost__logging__logLevel__Function": "Information",
+    "AzureFunctionsJobHost__logging__logLevel__Microsoft": "Warning"
```
benoutram, line 10: I assume you have to use this double underscore notation as it's the only way to override the default log configuration from `host.json` in the `local.settings.json`?

This PR says it's to stop the logging locally but isn't it also a problem in other environments where we won't be filtering these messages going into App Insights?

If there's a reason we want this for the local environment fair enough but if not then the equivalent of what we already have in `host.json` i.e.

```
"logging": {
  "logLevel": {
    "default": "Information",
    "Microsoft": "Warning"
  }
}
```

would just become

```
"logging": {
  "logLevel": {
    "default": "None",
    "Function": "Information",
    "Microsoft": "Warning"
  }
}
```

rather than these changes.

src/GovUk.Education.ExploreEducationStatistics.Publisher/local.settings.json
```diff
@@ -9,7 +9,10 @@
...
     "PublisherStorage": "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://data-storage:10000/devstoreaccount1;QueueEndpoint=http://data-storage:10001/devstoreaccount1;TableEndpoint=http://data-storage:10002/devstoreaccount1;",
-    "PublicStorage": "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://data-storage:10000/devstoreaccount1;TableEndpoint=http://data-storage:10002/devstoreaccount1;"
+    "PublicStorage": "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://data-storage:10000/devstoreaccount1;TableEndpoint=http://data-storage:10002/devstoreaccount1;",
+    "AzureFunctionsJobHost__logging__logLevel__Default": "None",
```
benoutram, line 13: I'm not sure that using `"Default: "None"` here and in the two other function projects is a good idea.
Have you tested this? I think if you do you will find you have supressed all of our own service code logging which will log under the `GovUk.Education.ExploreEducationStatistics` namespace and now be filtered out.

In addition to this you are supressing warnings and errors from any other dependencies that fall outside the `Microsoft` namespace. Compare this with the default logging configuration that you are overriding where we currently have `"Default": "Error"`. This means we wouldn't spot them when developing locally.

I would do the detective work to find the source of the log 'spam' that's causing the issue first and then you can target it's namespace to filter it out, if it's all being generated from the same source.

If that's too much then something like this might be better

```
"logging": {
  "logLevel": {
    "default": "Error",
    "Function": "Information",
    "GovUk.Education.ExploreEducationStatistics": "Information",
    "Microsoft": "Warning"
  }
```

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/local.settings.json
markhiveit: Talked to Ben about this - switching the `default` from `None` to `Error`. The log spam still isn't on.

src/GovUk.Education.ExploreEducationStatistics.Publisher/local.settings.json
markhiveit: Talked to Ben about this - switching the default from None to Error. The log spam still isn't on.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/local.settings.json
markhiveit: I checked on AppInsights on Dev, and the logs that were happening locally are also happening there. So I've moved the settings to the host.json so AppInsights also isn't spammed.

benoutram (approved):
This change looks good now. I suppose there's no need to have a separate `"Microsoft": "Warning"` given that the default level is also Warning, but no problem with separating them either.

PR #3180, opened 2022-02-18
EES-3167 Migrate Location codes to id's in Data Block queries, tables and charts (Location Id's change 3/4)

benoutram:
This PR migrates ONS location codes in data block queries, tables, and charts to location id's to help with the aim of EES-2955 to switch over to using location id's throughout the service with the exception of historical Permalinks.

Data block queries, tables and charts are migrated aside to three temporary columns `DataBlock_ChartsMigrated`, `DataBlock_QueryMigrated` and `DataBlock_TablesMigrated` allowing us to test the migration as it progresses through each environment before we commit to switching over to use them in https://github.com/dfe-analytical-services/explore-education-statistics/pull/3179 (EES-2955).

An additional flag stored in `DataBlock_LocationsMigrated` allows us to track the state of the migration. This defaults to `false` when a new data block is created and is also reset when a data block is updated so we can identify it as needing migrating again.

This migration is rerunnable so it can be used at any time to 'catch-up' with new and updated data blocks that need migrating.

We can also reset the migration by clearing the flag on all data blocks if any problems with the migration code are found so that we can try a complete re-run after deploying a fix.

By the point EES-2955 is deployed all data blocks must be migrated since it will switch to using the new values and assume they contain location id's. New and updated data blocks will be saved using location id's from that point in time onwards.

After EES-2955 has been deployed and tested in all environments  and we have confidence that the change has been successful the intention is to drop the original columns and rename these back to the original names `DataBlock_Charts`, `DataBlock_Query` and `DataBlock_Tables`.

Although migrations are traditionally run from the Admin app service, it isn't suitable to do this migration wholly inside a request/response cycle in an app service where a long running request could be terminated or a database connection could be reset.

The Publisher project has been temporarily enlisted to host a function that will dequeue messages for each data block, migrating them one at a time.

### Migration

Run the secured Admin endpoint `PATCH /api/releases/migrate-all-data-blocks` to generate migration queue messages for every data block that needs migrating.


src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/ContentDbContextModelSnapshot.cs
```diff
@@ -15,9 +17,10 @@ protected override void BuildModel(ModelBuilder modelBuilder)
         {
 #pragma warning disable 612, 618
             modelBuilder
-                .HasAnnotation("ProductVersion", "3.1.18")
-                .HasAnnotation("Relational:MaxIdentifierLength", 128)
-                .HasAnnotation("SqlServer:ValueGenerationStrategy", SqlServerValueGenerationStrategy.IdentityColumn);
+                .HasAnnotation("ProductVersion", "6.0.2")
+                .HasAnnotation("Relational:MaxIdentifierLength", 128);
+
+            SqlServerModelBuilderExtensions.UseIdentityColumns(modelBuilder, 1L, 1);
```
duncan-at-hiveit, line 23: Was this as a result of warnings or something?

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/ContentDbContextModelSnapshot.cs
```diff
@@ -960,8 +981,9 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.HasBaseType("GovUk.Education.ExploreEducationStatistics.Content.Model.ContentBlock");
 
                     b.Property<string>("Body")
-                        .HasColumnName("Body")
-                        .HasColumnType("nvarchar(max)");
+                        .ValueGeneratedOnUpdateSometimes()
```
duncan-at-hiveit, line 984: Whoa!

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/ContentDbContextModelSnapshot.cs
```diff
@@ -1019,6 +1050,12 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                         .WithOne()
                         .HasForeignKey("GovUk.Education.ExploreEducationStatistics.Content.Model.DataImport", "ZipFileId")
                         .OnDelete(DeleteBehavior.Restrict);
+
+                    b.Navigation("File");
```
duncan-at-hiveit, line 1054: I'm guessing as with other misc updates in this file, this is as a result of trying to create our first Migration since the .NET / EF upgrade?

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Model/Data/Query/ObservationQueryContextTests.cs
```diff
@@ -0,0 +1,116 @@
...
+                TimePeriod = new TimePeriodQuery
+                {
+                    StartCode = CalendarYear,
+                    StartYear = 2022,
+                    EndCode = CalendarYear,
+                    EndYear = 2022
+                }
+            };
+
+            var clone = original.Clone();
```
duncan-at-hiveit, line 59: Was assuming in this test that the first things we'd assert is that all of the fields in the clone were the same as the original, prior to the updates that we make to the clone.

I.e. pretty much a copy of all the assertions made on line 88 onwards, but applied to the clone directly after it's been created.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Model/Data/TableBuilderConfigurationTests.cs
```diff
@@ -0,0 +1,55 @@
...
+                    {
+                        new()
+                        {
+                            new TableHeader("original", TableHeaderType.TimePeriod)
+                        }
+                    }
+                }
+            };
+
+            var clone = original.Clone();
```
duncan-at-hiveit, line 42: Same as per comment on ObservationQueryContextTests.cs.

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Data/TableBuilderConfiguration.cs
```diff
@@ -1,7 +1,16 @@
...
     public class TableBuilderConfiguration
     {
-        public TableHeaders TableHeaders { get; set; }
+        public TableHeaders TableHeaders { get; set; } = new();
+
+        public TableBuilderConfiguration Clone()
+        {
+            return new TableBuilderConfiguration
+            {
+                TableHeaders = TableHeaders.Clone()
```
duncan-at-hiveit, line 12: Can this not use MemberwiseClone as well?

duncan-at-hiveit: Oh I see it does eventually.

duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/DataBlockMigrationController.cs
```diff
@@ -0,0 +1,35 @@
...
+using GovUk.Education.ExploreEducationStatistics.Admin.Models;
+using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Common.Model;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api.Bau
+{
+    /**
```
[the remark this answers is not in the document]
benoutram: Don't see why not, but depending on what the migration is doing you might just want to run it all from the Admin, rather than delegating to the Publisher project.

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/ContentDbContextModelSnapshot.cs
```diff
@@ -6,6 +6,8 @@
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
 
+#nullable disable
```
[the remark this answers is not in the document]
benoutram: Yes it was 👍

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/ContentDbContextModelSnapshot.cs
benoutram: It was added after updating the EF CLI tools from 3.1 to 6.0.2

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/ContentDbContextModelSnapshot.cs
benoutram: Since updating my own CLI tools I think. A new migration after the update but with the old version of the EF tools CLI wouldn't have these changes I believe.

You can update with `dotnet tool update --global dotnet-ef`

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Data/TableBuilderConfiguration.cs
benoutram: Ah, yep it doesn't use it here so we can call `Clone` on the `TableHeaders`.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Model/Data/Query/ObservationQueryContextTests.cs
benoutram: Good point, I totally missed that we might not be cloning the original correctly. I've made an update in https://github.com/dfe-analytical-services/explore-education-statistics/pull/3189 to test this using `AssertDeepEquals` before any updates are made to the clone.

PR #3185, opened 2022-02-24
EES-3086 EES-3060 move map indicator tile

amyb-hiveit:
The ticket was to move the indicator tiles to above maps so they are more noticeable.  However, doing this resulted in the map being pushed down the page, potentially off screen, when there were several tiles.

Chatted to Cam about it, he realised that all datasets were shown in the indicator tiles, not just the one that's selected. The same was true of the tooltip when you hover over the map.

So this PR:
- only shows the selected data set as an indicator tile and in the tooltip
- moves the indicator tile to under the map legend to make it more noticeable
- adds aria-live="polite" so that screen readers are informed when it changes
- removes aria-live="assertive" from the map legend to prevent useless information being shouted at screen reader users (https://dfedigital.atlassian.net/browse/EES-3060)

## Screenshots

map2

### DIFF
--- a/src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
+++ b/src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
@@ -417,6 +417,14 @@ export const MapBlockInternal = ({
   const selectedDataSetConfiguration =
     dataSetCategoryConfigs[selectedDataSetKey];
 
+  const selectedDataSet =
+    selectedFeature?.properties?.dataSets[selectedDataSetKey];
+
+  const {
+    config: selectedDataSetConfig,
+    dataSet: expandedSelectedDataSet,
+  } = dataSetCategoryConfigs[selectedDataSetKey];
+
   // initialise
   useEffect(() => {
     import('@common/modules/charts/files/ukGeoJson.json').then(imported => {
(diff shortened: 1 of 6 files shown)


src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
```diff
@@ -417,6 +417,14 @@ export const MapBlockInternal = ({
   const selectedDataSetConfiguration =
     dataSetCategoryConfigs[selectedDataSetKey];
 
+  const selectedDataSet =
+    selectedFeature?.properties?.dataSets[selectedDataSetKey];
+
+  const {
+    config: selectedDataSetConfig,
+    dataSet: expandedSelectedDataSet,
+  } = dataSetCategoryConfigs[selectedDataSetKey];
```
ntsim, line 426: We already assign `selectedDataSetConfiguration` from `dataSetCategoryConfigs[selectedDataSetKey]`. This would  be easily confused with the new destructured `selectedDataSetConfig` variable.

I'd probably lean towards getting rid of this destructuring and just using `selectedDataSetConfiguration` (maybe we want to rename it to `selectedDataSetConfig` for brevity).

src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
```diff
@@ -545,21 +553,13 @@ export const MapBlockInternal = ({
...
-                `<li>` +
-                `${dataSetConfig.config.label}: ${formatPretty(
-                  dataSet.value,
-                  dataSetConfig.dataSet.indicator.unit,
-                  dataSetConfig.dataSet.indicator.decimalPlaces,
-                )}` +
-                `</li>`
-              );
-            },
+          const dataSetConfig = dataSetCategoryConfigs[selectedDataSetKey];
```
ntsim, line 556: Not sure we need this? Couldn't we use `selectedDataSetConfiguration` / `selectedDataSetConfig` instead?

tests/robot-tests/tests/libs/charts.robot
```diff
@@ -141,11 +141,26 @@ user checks chart tooltip item contains
     user waits until element is visible    ${element}
     user waits until element contains    ${element}    ${text}
 
+user checks map tooltip label contains
+    [Arguments]    ${locator}    ${text}
+    user waits until parent contains element    ${locator}    css:[data-testid="chartTooltip-label"]
+    ${element}=    get child element    ${locator}    css:[data-testid="chartTooltip-label"]
```
ntsim, line 147: In cases where we use testids, we should be able to use our custom selector for it i.e: `testid:chartTooltip-label`

We might want to do this for the other testid selectors further down as well.

ntsim (changes requested):

src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
amyb-hiveit: yep, good spot. sorted.

src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
amyb-hiveit: yep, sorted

tests/robot-tests/tests/libs/charts.robot
amyb-hiveit: Done

ntsim (approved):

ntsim (approved):

PR #2809, opened 2021-08-24
Ees 2587 search with duplicate location codes

duncan-at-hiveit:
This PR:
- fixes Location search and various other React issues, by removing Locations with duplicated codes under a single geographic level from the set of Locations for a Subject.  It does this by combining those "duplicate" Locations into a single Location record for the front-end - in the example given in screenshots below we have combined 2 Providers with the same code together into a single record - "ANGLIA RUSKIN UNIVERSITY" and "ANGLIA RUSKIN UNIVERSITY HIGHER EDUCATION CORPORATION" become a single combined "ANGLIA RUSKIN UNIVERSITY / ANGLIA RUSKIN UNIVERSITY HIGHER EDUCATION CORPORATION" Location choice
- merges data from "duplicate" Locations together into single sets of table rows / cells, using a summation of each of the Locations' values for each table cell (each combination of Time Period, Filters and Measures)

## Provider search working
image

## Duplicate Locations combined in Data Block summary
image

## Table data combined for duplicate Locations, summing up numeric values where available or non-numerics when no numeric data exists for a table cell
image

## Chart working with combined duplicate Location
image
 

## Full successful Robot test run
Although this image shows 1 test failure, this was actually due to a duplicate test case name in a suite, which I have amended in this PR, within `prerelease_visibility.robot`.  This was causing subsequent successful runs to not be able to mark this as successful.

image

PR #3196, opened 2022-03-09
EES-2955 EES-3179 EES-3180 Switch to using Location Id's instead of codes (Location Id's change 4/4)

benoutram:
### ⚠️ Depends on Data block migration

This PR depends on https://github.com/dfe-analytical-services/explore-education-statistics/pull/3180 (EES-3167) which introduced code to migrate existing data blocks to use location id's.

It assumes that by the time this is deployed that no data blocks are waiting to be migrated and that they all have query, table and chart configuration that use location Id's in their temporary columns `DataBlock_QueryMigrated`, `DataBlock_TableMigrated` and `DataBlock_ChartMigrated`.

This should be tested immediately prior to the deploy. The count should be zero:
```
SELECT count(Id) from ContentBlock WHERE Type = 'DataBlock'
AND
(
  DataBlock_LocationsMigrated IS NULL
  OR
  DataBlock_LocationsMigrated = 0
)
```

If any data blocks are waiting to be migrated, re-run the migration to 'catch-up', following the instructions of https://github.com/dfe-analytical-services/explore-education-statistics/pull/3180.

EES-3212 has been created to eventually rename these columns back. They can replace the original `Query`, `Table` and `Chart` columns once we are satisfied with testing in all environments. At the same time we can also drop the migration code.

### ⚠️ Depends on clearing Public and Private cache

Warning: This will require the data blocks, fast tracks and subject meta JSON cache files to be cleared from the private and public caches in each environment.
* (Private)
```
DELETE /api/bau/private-cache
```
* (Public)

```
DELETE /api/bau/public-cache/releases
{
  "paths": [
    "data-blocks",
    "fast-track-results",
    "subject-meta",
  ]
}
```

###  What is this PR?

This PR builds upon the work of https://github.com/dfe-analytical-services/explore-education-statistics/pull/3073 , https://github.com/dfe-analytical-services/explore-education-statistics/pull/3074 and https://github.com/dfe-analytical-services/explore-education-statistics/pull/3180 allowing us to switch to working with Id's for locations instead of the ONS location codes.

Up until now location filter options have been a combination of geographic level and ONS location code of the corresponding location attribute within the location for that level.

The location field in table queries (including  the queries used in data blocks, fast tracks and permalinks) has been a dictionary of code arrays keyed by geographic level.

Subject meta responses used in table builder have included only the ONS codes and labels for locations to build the location filter options with.

Similarly table result subject meta data (returned for table builder results, data block tables and charts embedded in content, fast tracks and permalinks) has only included the codes.

Table result rows included the entire location view model containing code/label pairs for every attribute of the Observations location. **This created a lot of redundancy in responses since all the labels were already returned in the result subject meta data, and none of the codes except the one for the primary attribute corresponding with the geographic level had any relevance.**

**The way we rely on codes became problematic as soon as we allowed School and Provider data to be imported in https://github.com/dfe-analytical-services/explore-education-statistics/pull/2667 (and later on Institutions and Planning Areas) since these levels are allowed to have multiple attributes that reuse the same code.**

For example, this is a possible CSV upload where provider Rydell Primary has two sites that share the same code:

time_period | time_identifier | geographic_level | country_code | country_name | planning_area_code | planning_area_name | school_capacity
-- | -- | -- | -- | -- | -- | -- | --
202223 | Academic year | Planning Area | E92000001 | England | 9010001 | Rydell Primary (Site A) | 20
202223 | Academic year | Planning Area | E92000001 | England | 9010001 | Rydell Primary (Site B) | 7

Querying by code meant that it was impossible to restrict results to one of these locations without including the other. When displaying results it was also not possible to lookup a distinct location by code in the meta data in order to know the label.

This was temporarily resolved by the work of EES-2587 in https://github.com/dfe-analytical-services/explore-education-statistics/pull/2809. The FE code was changed to merge location attributes sharing the same code in meta data into one option and table results were merged by summing their values together.

In combination with https://github.com/dfe-analytical-services/explore-education-statistics/pull/3073 and https://github.com/dfe-analytical-services/explore-education-statistics/pull/3074 we can now query Observations by location Id alone without needing to match results by geographic level.

This PR switches to using location id's:
  * In table builder queries and in data blocks a new `locationIds` field replaces `locations`.
  * In subject meta data and table result subject meta data leaf nodes have a location id in addition to their code and label.
  * In table result rows Observations have a `locationId` which replaces `location`.

It removes the FE deduplication of Locations in Subject meta data, table result subject meta data, and table results.

Unfortunately because the FE is responsible for generating CSV downloads of built tables, the table result subject metadata still needs location codes which are populated into the CSV.

Due to our limited ability to migrate historical Permalinks, we also have to retain backwards compatibility with them by expecting their table results to contain a `location` rather than a `locationId`, and so must expect to match a location by the code relevant to its geographic level in the meta data which will not contain id's. This means that we can still expect historical Permalinks to have locations with duplicate codes and must deduplicate them.

### Also fixes
  * EES-3180 - Permalink tables for data with duplicate Location codes don't match original tables
    * We were not applying the FE deduplicating method to Permalinks, so in the one place where we need to retain this code we actually start applying it!
  * EES-3179 Clicking 'Explore data' on a table with duplicate locations takes you to different table with missing stats
    * This was caused because we also didn't apply the FE duduplicating method to Fast Tracks, but since we now no longer need to do so this should now be fixed.

### Other changes
  * Rework the data replacement service to expect that locations can be replaced in data blocks.
    * Previously this worked only by location codes, so analysts were free to correct labels and replace data which would have no affect on the location codes in data blocks and a replacement plan was valid if the replacement subject contained the same code. If this were to happen now and we didn't allow location id's to alter, any change to the location attributes would likely  invalidate the replacement plan. If the same location no longer exists for the subject we now allow targetting a different location on the same subject as long as it has the same code for its primary attribute.
  * Drop the query from Permalink view model so it's not returned in Permalink responses.
  * No longer need to specify the geographic level in a hierarchy for the level that the hierarchy is defined for.
    * For example,  the config of a hierarchy for Region-LA at the LA level becomes `"Locations": { "Hierarchies": { "LocalAuthority": [ "Region" ] } }`

### UI Test Result

UI Tests have been run with a new copy of the developer test database `ees-mssql-data-41.zip` which already has the migration of EES-3167 applied.

The private and public caches were also cleared prior to the run.

image


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Statistics/TableBuilderControllerTests.cs
```diff
@@ -44,15 +45,7 @@ public class TableBuilderControllerTests : BlobCacheServiceTestFixture
...
-                    LocationId = Guid.NewGuid(),
-                    Location = new LocationViewModel
-                    {
-                        Country = new CodeNameViewModel
-                        {
-                            Code = "code",
-                            Name = "name"
-                        }
-                    },
+                    LocationId = new Guid(),
```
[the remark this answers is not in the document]
benoutram: Good spot, no it doesn't `new Guid()` creates a blank Guid so we want `Guid.NewGuid()`. I've corrected this now and in a few other places in files that I'm touching. Thanks.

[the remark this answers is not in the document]
benoutram: Good spot, no it doesn't `new Guid()` creates a blank Guid so we want `Guid.NewGuid()`. I've corrected this now and in a few other places in files that I'm touching. Thanks.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/LocalAuthority.cs
```diff
@@ -1,9 +1,8 @@
 #nullable enable
-using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 
 namespace GovUk.Education.ExploreEducationStatistics.Data.Model
 {
-    public class LocalAuthority : ObservationalUnit, ILocationAttribute
+    public class LocalAuthority : LocationAttribute, ILocationAttribute
```
duncan-at-hiveit, line 5: Don't think we need the ILocationAttribute if this is already extending from LocationAtribute?

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/SubjectMetaServiceTests.cs
```diff
@@ -1159,10 +1160,7 @@ public async Task GetSubjectMetaForQuery_FilterItems()
             var query = new ObservationQueryContext
             {
                 SubjectId = subject.Id,
-                Locations = new LocationQuery
-                {
-                    Country = ListOf(_england.Code)!
-                },
+                LocationIds = ListOf(Guid.NewGuid()),
```
duncan-at-hiveit, line 1163: Was the england code not important here then?

src/explore-education-statistics-common/src/modules/table-tool/components/LocationFiltersForm.tsx
```diff
@@ -57,44 +57,45 @@ const LocationFiltersForm = ({
...
+      ),
+    [options],
   );
 
-  const allOptions = useMemo(() => Object.values(keyedOptions), [keyedOptions]);
-
-  const hasSingleOption = allOptions.length === 1;
+  const hasSingleOption =
+    levelKeys.length === 1 &&
+    Object.keys(keyedOptions[levelKeys[0]]).length === 1;
```
duncan-at-hiveit, line 82: Can't get my brain around what we're doing here :joy:  I'm guessing we're checking we've got one key and within that key only one value??

src/GovUk.Education.ExploreEducationStatistics.Data.Model/LocalAuthority.cs
benoutram: `ILocationAttribute` has two default interface methods `GetCacheKey` and `GetCodeOrFallback` which we want to override here in `LocalAuthority` to change their default behaviour.

I can't see anything about this in the docs for default interface methods but unless we repeat the interface declaration here, i.e. `LocalAuthority : LocationAttribute, ILocationAttribute` as opposed to `LocalAuthority : LocationAttribute` the methods are not treated as overrides.

If we remove it then unit tests that we've written around covering the old code for Local Authority data will fail

E.g.
`ImporterLocationServiceTests.FindOrCreate_UniqueByLocalAuthorityOldCode`
`ResultSubjectMetaServiceTests.GetSubjectMeta_LocationsForSpecialCases`
`SubjectMetaServiceTests.GetSubjectMeta_LocationsForSpecialCases`

Would be interested to know if you can spot something wrong with the way those methods are overriden?

Thanks.

src/explore-education-statistics-common/src/modules/table-tool/components/LocationFiltersForm.tsx
benoutram: Yes that's correct, just added additional comment around that.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/LocalAuthority.cs
duncan-at-hiveit: Will check later - lets get this onto Dev!

duncan-at-hiveit (approved):

PR #3203, opened 2022-03-11
EES-2459 publication.json cache file now generated by BlobCache rather than when an associated release is published

markhiveit:
Before this PR, the Azure storage cache file `publication.json` for each publication would be generated at publication time. This PR is to change this so it is produced whenever required, if no existing cache file exists, using the `BlobCache` method annotation. If the `publication.json` cache files doesn't already exist, it is now created when `Content.Services/PublicationService#Get` is called.

This change enables us to update the LegacyReleases on release pages immediately after they've been updated in Admin, so a change has been made to make this happen - basically deleting the publication.json cache file whenever LegacyReleases are changed.

The caching of releases is still done at publishing time, as this can take more time and could lead to a bad user experience, as the first time a release page is visited a user would have to wait for the cache file to be generated. As it is currently generated at publication time, this isn't currently an issue.


src/GovUk.Education.ExploreEducationStatistics.Content.Services/PublicationService.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        return
+            // Release itself must be live
+            release.Live
+            // It must also be the latest version unless the later version is a draft not included for publishing
+            && !release.Publication.Releases.Any(r =>
+                r.Live
+                && r.PreviousVersionId == release.Id
+                && r.Id != release.Id);
+    }
```
benoutram, line 108: I think you should try and reuse some of the logic in `Publication.cs` rather than duplicate this again here. It looks like adding a variation of `ListActiveReleases` like `GetPublishedReleases` which in turn uses `IsLatestPublishedVersionOfRelease` instead of  `IsLatestVersionOfRelease` would work.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Interfaces/IReleaseService.cs
```diff
@@ -3,16 +3,21 @@
...
     {
-        Task<Either<ActionResult, ReleaseViewModel>> Get(string publicationPath, string releasePath);
+        Task<Either<ActionResult, ReleaseViewModel>> Get(string publicationSlug, string? releaseSlug = null);
 
-        Task<Either<ActionResult, ReleaseSummaryViewModel>> GetSummary(string publicationPath, string releasePath);
+        Task<Either<ActionResult, ReleaseSummaryViewModel>> GetSummary(string publicationSlug, string? releaseSlug = null);
 
         Task<Either<ActionResult, List<ReleaseSummaryViewModel>>> List(string publicationSlug);
+
+        Task<Either<ActionResult, CachedReleaseViewModel?>> FetchCachedRelease(
```
benoutram, line 19: It would be more consistent to use `GetCachedRelease` here rather than introducing 'Fetch'.

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Controllers/PublicationController.cs
```diff
@@ -1,29 +1,34 @@
...
+        public PublicationController(IPublicationService publicationService)
         {
-            _fileStorageService = fileStorageService;
+            _publicationService = publicationService;
         }
 
         [HttpGet("publications/{slug}/title")]
         public async Task<ActionResult<PublicationTitleViewModel>> GetPublicationTitle(string slug)
         {
-            return await _fileStorageService
```
benoutram, line 25: Not sure if it's related to these changes or not but `PublicContentPublicationTitlePath` in `FileStoragePathUtils` has been left unused.

src/GovUk.Education.ExploreEducationStatistics.Common/Services/FileStoragePathUtils.cs
```diff
@@ -67,6 +67,11 @@ public static string PublicContentPublicationPath(string slug, string prefix = n
             return $"{PublicContentPublicationParentPath(slug, prefix)}/publication.json";
         }
 
+        public static string PublicContentPublicationTitlePath(string slug, string prefix = null)
+        {
+            return $"{PublicContentPublicationParentPath(slug, prefix)}/publication-title.json";
+        }
+
```
benoutram, line 74: Looks like this isn't required

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ReleaseService.cs
```diff
@@ -24,37 +25,83 @@ public class ReleaseService : IReleaseService
...
                 {
-                    var (viewModel, methodologies) = viewModelAndMethodologies;
-                    viewModel.Publication.Methodologies = methodologies;
-                    return viewModel;
+                    var publicationTask = _publicationService.Get(publicationSlug);
+                    var releaseTask = FetchCachedRelease(publicationSlug, releaseSlug);
+
+                    await Task.WhenAll(publicationTask, releaseTask);
+
+                    if (releaseTask.Result.IsRight
```
benoutram, line 61: I think we could improve on this by changing back to `OnSuccessCombineWith`. Elsewhere I've said that I think we could be getting the Publication by Id rather than slug so that kind of invalidates what I'm about to say as we'd need to look up the Publication by slug still somewhere, but this method could have started with `_publicationService.Get` and then `.OnSuccessCombineWith(model => _methodologyService.GetSummariesByPublication(model.Publication.Id))` rather than start with `persistenceHelper.CheckEntityExists<Publication>(query => query.Where(p => p.Slug == publicationSlug))`.

That's just duplicating what `_publicationService.Get` is already doing.

benoutram (changes requested):
Have not completed full review as I've not reviewed any tests yet. Will wait for @markhiveit to reply to comments so far.

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests.csproj
```diff
@@ -22,6 +22,7 @@
     <ItemGroup>
         <ProjectReference Include="..\GovUk.Education.ExploreEducationStatistics.Content.Api\GovUk.Education.ExploreEducationStatistics.Content.Api.csproj" />
         <ProjectReference Include="..\GovUk.Education.ExploreEducationStatistics.Common.Tests\GovUk.Education.ExploreEducationStatistics.Common.Tests.csproj" />
+        <ProjectReference Include="..\GovUk.Education.ExploreEducationStatistics.Content.Model.Tests\GovUk.Education.ExploreEducationStatistics.Content.Model.Tests.csproj" />
```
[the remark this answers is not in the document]
markhiveit: I thought it was being added for MockUtils or something like that, but removing the reference doesn't seem to make anything fail - so I'll remove it.

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Startup.cs
```diff
@@ -121,6 +121,7 @@ public void ConfigureServices(IServiceCollection services)
             services.AddTransient<IFilterRepository, FilterRepository>();
             services.AddTransient<IIndicatorRepository, IndicatorRepository>();
             services.AddTransient<IDataGuidanceService, DataGuidanceService>();
+            services.AddTransient<Content.Services.Interfaces.IPublicationService, Content.Services.PublicationService>();
```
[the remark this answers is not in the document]
markhiveit: I saw that but decided to include to it to be explicit - I'll remove when it isn't necessary.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -185,6 +191,10 @@ public class PublicationService : IPublicationService
                     if (publication.Live)
                     {
                         await _publishingService.PublicationChanged(publication.Id);
+
+                        await _publicBlobCacheService.DeleteItem(new PublicationCacheKey(publication.Slug));
```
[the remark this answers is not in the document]
markhiveit: `PublishPublicationFunction` called `contentService.UpdatePublication` - which set a new published date on the publication and deleted all the public taxonomy blobs. We're now doing those things in `PublicationService#UpdatePublication` instead, which means I've been able to remove `PublishPublicationFunction` entirely and `contentService.UpdatePublication`, which was only being used by that function.

This also entirely removes the `publish-publication` queue, so we could also delete that on environments after this change is deployed.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/PublicationService.cs
```diff
@@ -16,40 +13,17 @@ namespace GovUk.Education.ExploreEducationStatistics.Publisher.Services
...
-        {
-            var publication = await _contentDbContext.Publications
-                .Include(p => p.Contact)
-                .Include(p => p.LegacyReleases)
-                .Include(p => p.Topic)
-                .ThenInclude(topic => topic.Theme)
-                .SingleOrDefaultAsync(p => p.Id == id);
-
-            var publicationViewModel = _mapper.Map<CachedPublicationViewModel>(publication);
-            var latestRelease = await _releaseService.GetLatestRelease(publication.Id, includedReleaseIds);
```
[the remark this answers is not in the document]
markhiveit: Talked to Ben about this. Rider sold him up the river, and told him `ReleaseService#GetLatestRelease` wasn't being used when it was.

src/GovUk.Education.ExploreEducationStatistics.Common/Cache/PublicationCacheKey.cs
```diff
@@ -0,0 +1,15 @@
...
+using GovUk.Education.ExploreEducationStatistics.Common.Cache.Interfaces;
+using GovUk.Education.ExploreEducationStatistics.Common.Services;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Cache
+{
+    public record PublicationCacheKey(string PublicationSlug) : IBlobCacheKey
+    {
+        private string PublicationSlug { get; } = PublicationSlug;
+
+        public string Key => FileStoragePathUtils.PublicContentPublicationPath(PublicationSlug);
```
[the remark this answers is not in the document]
markhiveit: I've made a separate ticket (EES-3262) to address this. This ticket would change the way we store all publication-specific cache files from the current format of `public-cache/publication/{publicationSlug}` to `public-cache/publications/{publicationId}`

src/GovUk.Education.ExploreEducationStatistics.Content.Services/PublicationService.cs
```diff
@@ -0,0 +1,109 @@
...
+                    .OnSuccess(latestRelease =>
+                    {
+                        publicationViewModel.LatestReleaseId = latestRelease.Id;
+                        publicationViewModel.Releases = ListReleases(publication.Id);
+                        return publicationViewModel;
+                    });
+            });
+    }
+
+    private async Task<Either<ActionResult,Release>> GetLatestRelease(Guid publicationId)
```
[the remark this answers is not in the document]
markhiveit: Reformatted to add a space between `ActionResult, Release`

src/GovUk.Education.ExploreEducationStatistics.Content.Services/PublicationService.cs
```diff
@@ -0,0 +1,109 @@
...
+            .OrderBy(release => release.Year)
+            .ThenBy(release => release.TimePeriodCoverage)
+            .LastOrDefault();
+
+        if (release == null)
+        {
+            return new NotFoundResult();
+        }
+
+        return release;
```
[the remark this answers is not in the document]
markhiveit: Fixed

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/ContentService.cs
```diff
@@ -134,7 +133,9 @@ public async Task UpdateContent(PublishContext context, params Guid[] releaseIds
 
             foreach (var publication in publications)
             {
-                await CachePublication(publication.Id, context, releaseIds);
+                await _blobCacheService.DeleteItem(new PublicationCacheKey(publication.Slug));
```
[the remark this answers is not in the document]
markhiveit: I've made the change so now we don't delete the cache file here, but instead delete it in the `PublishStagedReleaseContent` method as you suggested.

We also delete the cache file alongside the deletion of other cache files in PublishReleaseContentFunction#PublishReleaseContent`. I think this is the correct place as in addition to the bug in the above comment, I think it would also be possible for someone to generate a new publication cache file before `_releaseService.SetPublishedDates` which I believe would lead to an issue.

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Controllers/PublicationController.cs
markhiveit: It is. I originally intended to create a new cache file in additiona to `publication.json`: `publication-title.json`. But in the end I went back on that decision and the publication title is fetched from `publication.json`.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/PublicationService.cs
```diff
@@ -0,0 +1,70 @@
...
+                .Include(p => p.Contact)
+                .Include(p => p.LegacyReleases)
+                .Include(p => p.Topic)
+                .ThenInclude(topic => topic.Theme)
+                .Where(p => p.Slug == publicationSlug))
+            .OnSuccess(publication =>
+            {
+                var publicationViewModel = _mapper.Map<PublicationViewModel>(publication);
+                // NOTE: BlobCache won't cache result if return a Either.IsLeft - if there is no latestRelease
+                return GetLatestRelease(publication)
```
benoutram, line 46: `var publicationViewModel = _mapper.Map<PublicationViewModel>(publication);` could move inside the `OnSuccess` below as it's not relevant if there's no latest Release.

Alternatively `OnSuccess(publication` above could change to `OnSuccessCombineWith` and then get the latest release and then you would reduce the level of nesting here down to one block where you have the publication and latest release, and then build the view model.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
```diff
@@ -1318,20 +1410,21 @@ await using (var context = InMemoryApplicationDbContext(contextId))
...
                 context,
                 AdminMapper(),
                 new PersistenceHelper<ContentDbContext>(context),
                 userService ?? AlwaysTrueUserService().Object, 
-                publicationRepository ?? new Mock<IPublicationRepository>().Object,
-                publishingService ?? new Mock<IPublishingService>().Object, 
-                methodologyVersionRepository ?? new Mock<IMethodologyVersionRepository>().Object);
+                publicationRepository ?? Mock.Of<IPublicationRepository>(),
+                methodologyVersionRepository ?? Mock.Of<IMethodologyVersionRepository>(),
+                publicBlobCacheService ?? Mock.Of<IBlobCacheService>());
```
benoutram, line 1427: I think you should use Strict behaviour on all the mocks here like we do elsewhere.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
```diff
@@ -646,8 +655,12 @@ await using (var context = InMemoryApplicationDbContext(contextId))
                         TopicId = topic.Id
                     }
                 );
+
+                // NOTE: Not called because publication.Live is false
+                publicBlobCacheService.Verify(mock => mock.DeleteItem(It.IsAny<PublicationCacheKey>()),
```
benoutram, line 660: If you use strict behaviour on all the mocks you can get away without this here and in other places throughout the tests. If something unexpected is called it will fail the tests which is why we use the strict behaviour (wish this was the default behaviour!).

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
```diff
@@ -740,7 +758,18 @@ await using (var context = InMemoryApplicationDbContext(contextId))
...
+                publicBlobCacheService.Verify(mock => mock.DeleteItem(new PublicationTreeCacheKey(null)),
+                    Times.Once);
+                publicBlobCacheService.Verify(mock => mock.DeleteItem(new PublicationTreeCacheKey(PublicationTreeFilter.AnyData)),
+                    Times.Once);
+                publicBlobCacheService.Verify(mock => mock.DeleteItem(new PublicationTreeCacheKey(PublicationTreeFilter.LatestData)),
+                    Times.Once);
+                publicBlobCacheService.Verify(mock => mock.DeleteItem(It.IsAny<PublicationCacheKey>()),
+                    Times.Once);
+
+                VerifyAllMocks(methodologyVersionRepository, publicBlobCacheService);
```
benoutram, line 772: Unless your concerned about this happening more than once then I would change to using strict behaviour, and make these verifications the setups instead.

Anything not setup but invoked will fail due to the strict behaviour.
Anything setup and not invoked will fail the `VerifyAllMocks.

I think this comment applies elsewhere to to your test changes in this PR where I see you are using default behaviour.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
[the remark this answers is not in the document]
benoutram: Great, that is what I was expecting! 🙂

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -184,13 +189,26 @@ public class PublicationService : IPublicationService
 
                     if (publication.Live)
                     {
-                        await _publishingService.PublicationChanged(publication.Id);
+                        publication.Published = DateTime.UtcNow;
+                        await _context.SaveChangesAsync();
+
+                        await DeleteCachedTaxonomyBlobs();
+                        await _publicBlobCacheService.DeleteItem(new PublicationCacheKey(publication.Slug));
+                        // @MarkFix EES-3149 Need to handle superseded publication.json cache files here too?
```
benoutram, line 197: If you use `// TODO` rather than your own annotation it will benefit us all as we can see all TODO items in the Rider TODO tool window and we also get a prompt if we are altering code with a TODO still remaining.

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/PublicationControllerTests.cs
```diff
@@ -17,54 +19,50 @@ public async Task GetPublicationTitle()
...
 
+            Assert.NotNull(publicationTitleViewModel);
             Assert.IsType<PublicationTitleViewModel>(publicationTitleViewModel);
-
-            Assert.Equal(publicationId, publicationTitleViewModel.Id);
+            Assert.Equal(publicationId, publicationTitleViewModel!.Id);
             Assert.Equal("Test title", publicationTitleViewModel.Title);
 
-            MockUtils.VerifyAllMocks(fileStorageService);
+            MockUtils.VerifyAllMocks();
```
benoutram, line 40: Should be `publicationService` here?

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/ContentService.cs
```diff
@@ -144,18 +142,6 @@ public async Task UpdateContent(PublishContext context, params Guid[] releaseIds
             }
         }
 
-        public async Task UpdatePublication(PublishContext context, Guid publicationId)
-        {
-            var publication = await _publicationService.Get(publicationId);
-
-            await CachePublication(publication.Id, context);
-            await _publicationService.SetPublishedDate(publication.Id, context.Published);
```
benoutram, line 152: Pretty sure this time, this is one that can be removed! 😆

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/DataGuidanceServiceTests.cs
```diff
@@ -119,101 +119,100 @@ public async Task Get()
...
+            IDataGuidanceSubjectService? dataGuidanceSubjectService = null,
+            Interfaces.IPublicationService? publicationService = null,
+            Interfaces.IReleaseService? releaseService = null)
         {
             return new DataGuidanceService(
-                fileStorageService ?? Mock.Of<IFileStorageService>(),
-                dataGuidanceSubjectService ?? Mock.Of<IDataGuidanceSubjectService>()
+                dataGuidanceSubjectService ?? Mock.Of<IDataGuidanceSubjectService>(),
+                publicationService ?? Mock.Of<Interfaces.IPublicationService>(),
+                releaseService ?? Mock.Of<Interfaces.IReleaseService>()
```
benoutram, line 215: I think you could try using strict behaviour on these mocks.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Services/DataGuidanceService.cs
```diff
@@ -1,36 +1,39 @@
...
+            _publicationService = publicationService;
+            _releaseService = releaseService;
         }
 
-        public async Task<Either<ActionResult, DataGuidanceViewModel>> Get(string publicationPath, string releasePath)
+        public async Task<Either<ActionResult, DataGuidanceViewModel>> Get(string publicationSlug, string? releaseSlug = null)
         {
-            var publicationTask = _fileStorageService.GetDeserialized<CachedPublicationViewModel>(publicationPath);
-            var releaseTask = _fileStorageService.GetDeserialized<CachedReleaseViewModel>(releasePath);
+            var publicationTask = _publicationService.Get(publicationSlug);
```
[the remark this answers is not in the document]
markhiveit: I've changed this in all places that a `Task.WhenAll` is used. There is one instance where I believe we could still use `Task.WhenAll` (`ReleaseService#GetSummary`), but I figure the future risk isn't worth the performance benefit, which I suspect is negligible. So to prevent a future bug I've awaited `_publicationService.Get` and `GetCachedRelease` separately there too.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/PublicationService.cs
```diff
@@ -0,0 +1,70 @@
...
+                        publicationViewModel.LatestReleaseId = latestRelease.Id;
+                        publicationViewModel.Releases = ListPublishedReleases(publication);
+                        return publicationViewModel;
+                    });
+            });
+    }
+
+    private static Either<ActionResult, Release> GetLatestRelease(Publication publication)
+    {
+        // @MarkFix EES-3149 exclude superseded releases here
```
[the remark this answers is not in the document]
markhiveit: Fair point - changed the wording.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/PublicationService.cs
markhiveit: Made the change. Had to create a new overload of `OnSuccessCombineWith` that doesn't return a Task to get this to work.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
markhiveit: I've switched them all to Strict, but I am actually inclined to think using MockBehavior.Default in these instances actually made it easier for someone to read what is going on. I've left a comment in the tests about why no Setup for publicBlobCacheService is necessary.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ReleaseService.cs
```diff
@@ -24,37 +26,76 @@ public class ReleaseService : IReleaseService
...
+                        var result = new Either<ActionResult, ReleaseViewModel>(new ReleaseViewModel(
+                            _mapper.Map<CachedReleaseViewModel>(cachedRelease.Right),
+                            _mapper.Map<PublicationViewModel>(publication)
+                        ));
+
+                        result.Right.Publication.Methodologies = methodologies;
+                        return result;
+                    }
+
+                    return new NotFoundResult();
```
benoutram, line 71: I'm not sure that publication can be null here? 
I'm not sure if you consider this any better/simpler but you might be interested to see it can be written combining another `OnSuccessCombineWith` and without  needing to wrap the `ReleaseViewModel` in an `Either`, and without the references to `.Right` as follows:

```
            return await _publicationService.Get(publicationSlug)
                .OnSuccessCombineWith(publication => _methodologyService.GetSummariesByPublication(publication.Id))
                .OnSuccessCombineWith(_ => GetCachedRelease(publicationSlug, releaseSlug))
                .OnSuccess<
                    ActionResult,
                    Tuple<PublicationViewModel, List<MethodologyVersionSummaryViewModel>, CachedReleaseViewModel?>,
                    ReleaseViewModel>(tuple =>
                {
                    var (publication, methodologies, cachedRelease) = tuple;
                    if (cachedRelease != null)
                    {
                        return new ReleaseViewModel(
                            _mapper.Map<CachedReleaseViewModel>(cachedRelease),
                            _mapper.Map<PublicationViewModel>(publication))
                        {
                            Publication =
                            {
                                Methodologies = methodologies
                            }
                        };
                    }

                    return new NotFoundResult();
                });
```

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ReleaseService.cs
```diff
@@ -24,37 +26,76 @@ public class ReleaseService : IReleaseService
...
+                    .Where(p => p.Slug == publicationSlug))
+                .OnSuccess(async _ =>
+                {
+                    var publication = await _publicationService.Get(publicationSlug);
+                    var release = await GetCachedRelease(publicationSlug, releaseSlug);
+
+                    if (release.IsRight
+                        && release.Right is not null
+                        && publication.IsRight
+                        && publication.Right is not null)
```
benoutram, line 89: Isn't `_publicationService.Get` going to mean the `.CheckEntityExists` is a bit redundant here? Same comment as before that you might want to consider using `OnSuccessCombineWith`

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Publication.cs
```diff
@@ -36,6 +36,10 @@ public class Publication
 
         public DateTime? Updated { get; set; }
 
+        public Guid? SupersededById { get; set; }
+
+        public Publication? SupersededBy { get; set; }
+
```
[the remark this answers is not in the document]
markhiveit: Good spot. Not sure how I've managed that...

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Controllers/PreReleaseAccessListController.cs
```diff
@@ -32,23 +34,21 @@ public async Task<ActionResult<PreReleaseAccessListViewModel>> GetLatest(string
...
-
-            await Task.WhenAll(publicationTask, releaseTask);
+            var publication = await _publicationService.Get(publicationSlug);
+            var release = await _releaseService.GetCachedRelease(publicationSlug, releaseSlug);
 
-            if (releaseTask.Result.IsRight && publicationTask.Result.IsRight)
+            if (release.IsRight && publication.IsRight)
             {
-                return new PreReleaseAccessListViewModel(releaseTask.Result.Right, publicationTask.Result.Right);
+                return new PreReleaseAccessListViewModel(release.Right!, publication.Right);
```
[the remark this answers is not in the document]
markhiveit: I've changed this in PreReleaseAccessListController and ReleaseService and DataGuidanceService, as all three contained a similar pattern.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ReleaseService.cs
markhiveit: Good point. Removed `CheckEntityExists`

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ReleaseService.cs
markhiveit: I changed it slightly differently, but still follows the broad strokes of what you suggest.

PR #3220, opened 2022-03-22
    EES-3268 - add health checks to admin, content & data APIs

luke-h1:
This PR adds health checks to the admin, data & content API. This has been added in order to provide an endpoint that Azure can use during deployments (in an attempt to fix EES-949)

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
@@ -103,6 +103,9 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin
         // This method gets called by the runtime. Use this method to add services to the container.
         public void ConfigureServices(IServiceCollection services)
         {
+
+            services.AddHealthChecks();
+            
             services.AddApplicationInsightsTelemetry()
                 .AddApplicationInsightsTelemetryProcessor<SensitiveDataTelemetryProcessor>();
 
@@ -467,7 +470,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
         public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
         {
-            // Enable caching and register any caching services.
+        // Enable caching and register any caching services.
             CacheAspect.Enabled = true;
             BlobCacheAttribute.AddService("default", app.ApplicationServices.GetService<IBlobCacheService>());
 
@@ -533,6 +536,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin
             app.UseAuthentication();
             app.UseIdentityServer();
             app.UseAuthorization();
+            app.UseRouting();
+            app.UseHealthChecks("/api/health");
 
             // deny access to all Identity routes other than /Identity/Account/Login and
             // /Identity/Account/ExternalLogin
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/web.config
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/web.config
@@ -15,10 +15,9 @@
         <requestLimits maxAllowedContentLength="104857600" />
       </requestFiltering>
     </security>
-   <applicationInitialization
-      doAppInitAfterRestart="true"
-    >
-      <add initializationPage="/"/> 
+
+    <applicationInitialization doAppInitAfterRestart="true">
+      <add initializationPage="/api/health" />
     </applicationInitialization>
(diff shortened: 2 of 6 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
```diff
@@ -462,11 +465,13 @@ public void ConfigureServices(IServiceCollection services)
                     }
                 });
             });
-        }
+       }
 
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
         public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
         {
+
```
duncan-at-hiveit, line 473: Just check over the formatting in this file?

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Startup.cs
```diff
@@ -199,6 +201,8 @@ public void ConfigureServices(IServiceCollection services)
                 .AllowAnyHeader());
 
             app.UseMvc();
+            app.UseHealthChecks("/api/health");
+
```
duncan-at-hiveit, line 205: Empty line here

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Startup.cs
```diff
@@ -236,6 +238,8 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
                 .AllowAnyHeader());
 
             app.UseMvc();
+            app.UseHealthChecks("/api/health");
+
```
duncan-at-hiveit, line 242: Empty line here

src/GovUk.Education.ExploreEducationStatistics.Admin/web.config
```diff
@@ -15,10 +15,12 @@
...
       </requestFiltering>
     </security>
-   <applicationInitialization
+
+    <applicationInitialization
       doAppInitAfterRestart="true"
     >
-      <add initializationPage="/"/> 
+      <add initializationPage="/api/health" 
+      />
```
duncan-at-hiveit, line 23: Can remove the newline here

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
duncan-at-hiveit: got some comments that are out of alignment and some empty lines

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
```diff
@@ -103,6 +103,9 @@ public Startup(IConfiguration configuration, IHostEnvironment hostEnvironment)
         // This method gets called by the runtime. Use this method to add services to the container.
         public void ConfigureServices(IServiceCollection services)
         {
+
+            services.AddHealthChecks();
```
duncan-at-hiveit, line 107: This is cool!  So do we need to implement something ourselves at /api/health or is there a pre-built one that gives back basic information on this route?  I guess at the very least it's going to error until the API is up and running.

luke-h1: Nope, by default, I believe it returns a 200 OK if it's deemed to be in the `healthy state` or a `503 service unavailable` if it's deemed to be in an `unhealthy` state. This is quite a basic implementation but you could take it further by adding more granular health checks for individual services i.e. 

```
using Microsoft.Extensions.Diagnostics.HealthChecks;
using System.Data.SqlClient;
 
namespace HealthCheck.Demo
{
    public static class DbHealthCheckProvider
    {
        public static HealthCheckResult Check(string connectionString)
        {
            try
            {
                using var connection = new SqlConnection(connectionString);
                connection.Open();
                return HealthCheckResult.Healthy();
            }
            catch
            {
                return HealthCheckResult.Unhealthy();
            }
        }
    }
}
```

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
luke-h1: Ta, I think some formatting went awry when I used Rider

benoutram: I've just checked this and I can still see problems with the formatting in the diff shown in Github. When I reformat with Rider there are differences.

Maybe try reformatting with Rider again?

benoutram (changes requested):
Check formatting of Admin `Statup.cs` again.

benoutram (approved):

PR #3225, opened 2022-03-24
EES-3149 Add ability to archive publication

markhiveit:
This work adds the ability to archive a publication by setting a superseding publication. The superseding publication is set in a new column on the `Publications` table, `SupersededById`. A publication becomes archived once the superseding publication contains a live release.

Once a publication is archived, it will no longer appear on the Find statistics page or the Table tool page. It will, however, still appear on the Data guidance page, and the release pages themselves will still be accessible, but will on those pages, any data / releases will no longer be marked as the "Latest data".

A superseding publication can be set on the Manage publication page.

After deploying this change, we will need to clean up existing tree json cache blobs.

Edit: I've pushed the UI tests from EES-3272 on to this PR for convenience


src/explore-education-statistics-frontend/src/modules/find-statistics/__tests__/PublicationReleasePage.test.tsx
```diff
@@ -17,6 +17,24 @@ describe('PublicationReleasePage', () => {
     ).toBeDefined();
   });
 
+  test('renders latest data tag', () => {
+    render(<PublicationReleasePage release={testRelease} />);
+
+    expect(screen.queryByText('This is the latest data')).toBeInTheDocument();
+  });
+
+  test('does not render latest data tag publication is superseded', () => {
```
ntsim, line 26: 'does not render latest data tag **when** publication is superseded'

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationRepository.cs
```diff
@@ -149,7 +149,24 @@ public async Task<List<Release>> ListActiveReleases(Guid publicationId)
...
+        }
+
+        private bool IsSuperseded(Publication publication)
+        {
+            return publication.SupersededById != null
+                   // To be superseded, superseding publication must have Live release
+                   && _context.Releases
+                       .Where(r => r.PublicationId == publication.SupersededById)
+                       .ToList()
+                       .Any(r => r.Live);
```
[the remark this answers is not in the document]
markhiveit: I think I originally did do this and it didn't like the r.Live as part of the queryable, but I'll double check.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
```diff
@@ -1121,6 +1210,59 @@ await using (var context = InMemoryApplicationDbContext(contextId))
             }
         }
 
+        [Fact]
+        public async Task UpdatePublication_NoPermissionToChangeSupersededBy()
```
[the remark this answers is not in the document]
markhiveit: I've moved all permission tests to the `PermissionTests` file. I've switched the existing permission tests to use the same type of BuildService method that we tend to use elsewhere in the service, removing the `var mock = Mocks();` style.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Functions/PublishReleaseContentFunction.cs
```diff
@@ -83,17 +81,24 @@ public class PublishReleaseContentFunction
...
+
+                // Invalidate publication cache for superseded publications, as potentially affected. If newly
+                // published release is first Live release for the publication, the superseding is now enforced
+                var supersededPublications = await _contentDbContext.Publications
+                    .Where(p => p.SupersededById == release.Publication.Id)
+                    .ToListAsync();
+                foreach (var p in supersededPublications)
+                {
+                    await _blobCacheService.DeleteItem(new PublicationCacheKey(p.Slug));
+                }
```
benoutram, line 101: You could use `ForEachAwaitAsync` like

```
                await _contentDbContext.Publications
                    .Where(p => p.SupersededById == release.Publication.Id)
                    .ToAsyncEnumerable()
                    .ForEachAwaitAsync(publication =>
                        _blobCacheService.DeleteItem(new PublicationCacheKey(publication.Slug)));
```

You could also refactor the blob service to be able to delete multiple cache keys at once.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/PublishingService.cs
```diff
@@ -58,7 +63,19 @@ public async Task PublishStagedReleaseContent(Guid releaseId, string publication
             await _releaseService.SetPublishedDates(releaseId, DateTime.UtcNow);
 
             await _publicBlobCacheService.DeleteItem(new PublicationCacheKey(publicationSlug));
-            // TODO: @MarkFix EES-3149 Delete superseded publication's cache here too?
+
+            // Invalidate publication cache for superseded publications, as potentially affected. If newly
+            // published release is first Live release for the publication, the superseding is now enforced
+            var supersededPublications = await _contentDbContext.Publications
```
benoutram, line 69: Same comment as elsewhere about using the `ForEachAwaitAysnc` but just wondering if any additional testing has gone into this because in all of our automated testing and most of our manual testing we publish immediately, but this change here affects publishing with scheduled releases where the content is moved from the staging directory.

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -0,0 +1,180 @@
...
+Check public superseding-publication release page displays correctly
+    user clicks element    testid:View stats link for ${PUB_NAME_SUPERSEDE}
+
+    user waits until h1 is visible    ${PUB_NAME_SUPERSEDE}    %{WAIT_MEDIUM}
+    user waits until page contains    This is the latest data
+
+Check public archive-publication release page displays correctly
+    go to    ${PUB_ARCHIVE_URL}
+    user waits until h1 is visible    ${PUB_NAME_ARCHIVE}    %{WAIT_MEDIUM}
+    user checks page does not contain    This is the latest data
```
benoutram, line 137: Not related to this test but on Permalink pages we show a warning a soon as the Permalink is *not* the latest data, i.e. not from the latest published release.

See the code in `Data.Api.Services.PermalinkService.BuildViewModel` which sets that flag.

If a Permalink table is for the latest published release of that publication, but that publication is now superseded, should we now show the same warning?

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
```diff
@@ -144,6 +156,80 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
...
+                Assert.True(publicationViewModel.IsSuperseded);
+
+                Assert.Single(publicationViewModel.Releases);
+                Assert.Equal(publication.Releases[0].Id, publicationViewModel.LatestReleaseId);
+
+                Assert.Equal(publication.Releases[0].Id, publicationViewModel.Releases[0].Id);
+                Assert.Equal(publication.Releases[0].Slug, publicationViewModel.Releases[0].Slug);
+                Assert.Equal(publication.Releases[0].Title, publicationViewModel.Releases[0].Title);
+
+                Assert.Equal(publication.Topic.Theme.Title, publicationViewModel.Topic.Theme.Title);
```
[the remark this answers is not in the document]
markhiveit: The test data in this case is required, as a publication is only superseded if the superseding publication has a live release. I've trimmed the assertions.

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -0,0 +1,293 @@
...
+Resource            ../../libs/public-common.robot
+
+Suite Setup         user signs in as bau1
+Suite Teardown      user closes the browser
+Test Setup          fail test fast if required
+
+Force Tags          Admin    Local    Dev    AltersData
+
+*** Variables ***
+${PUB_NAME_ARCHIVE}=            UI tests - archived publication %{RUN_IDENTIFIER}
```
luke-h1, line 13: Would rename this & any other instance of `PUB` to either `publication` or `public` as it could lead to confusion about whether this means 'public' or 'publication'

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -0,0 +1,293 @@
...
+Navigate to archive-publication release
+    user navigates to editable release summary from admin dashboard    ${PUB_NAME_ARCHIVE}
+    ...    ${RELEASE_NAME_ARCHIVE} (not Live)
+
+Import archive-publication subject to release
+    user uploads subject    ${SUBJECT_NAME_ARCHIVE}    upload-file-test.csv    upload-file-test.meta.csv
+
+Add data guidance to archive-publication subject
+    user clicks link    Data guidance
+    user waits until h2 is visible    Public data guidance
```
luke-h1, line 35: Would add `%{WAIT_SMALL}` to prevent issues in CI

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -0,0 +1,293 @@
...
+    ${PUB_ID_SUPERSEDE}=    user creates test publication via api    ${PUB_NAME_SUPERSEDE}
+    user create test release via api    ${PUB_ID_SUPERSEDE}    FY    2000
+
+Set archive-publication to be superseded by superseding-publication
+    user navigates to admin dashboard
+    user opens accordion section    ${PUB_NAME_ARCHIVE}
+    ${ARCHIVE_ACCORDION}=    user gets accordion section content element    ${PUB_NAME_ARCHIVE}
+    user clicks link    Manage publication    ${ARCHIVE_ACCORDION}
+
+    user waits until h1 is visible    Manage publication
```
luke-h1, line 59: Would add %{WAIT_SMALL} to prevent issues in CI

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -0,0 +1,293 @@
...
+Navigate to superseding-publication release on Admin site
+    user navigates to editable release summary from admin dashboard    ${PUB_NAME_SUPERSEDE}
+    ...    ${RELEASE_NAME_SUPERSEDE} (not Live)
+
+Import superseding-publication subject to release
+    user uploads subject    ${SUBJECT_NAME_SUPERSEDE}    upload-file-test.csv    upload-file-test.meta.csv
+
+Add data guidance to superseding-publication subject
+    user clicks link    Data guidance
+    user waits until h2 is visible    Public data guidance
```
luke-h1, line 132: Would add %{WAIT_SMALL} to prevent issues in CI

luke-h1 (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationRepository.cs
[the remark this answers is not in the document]
markhiveit: Yeah doesn't like it:
```
fail: Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware[1]
      An unhandled exception has occurred while executing the request.
      System.InvalidOperationException: The LINQ expression 'DbSet<Release>()
          .Where(r => !(r.SoftDeleted))
          .Any(r => (Guid?)r.PublicationId == __publication_SupersededById_0 && r.Live)' could not be translated. Additional information: Translation of member 'Live' on entity type 'Release' failed. This commonly occurs when the specified member is unmapped. Either rewrite the query in a form that can be translated, or switch to client evaluation explicitly by inserting a call to 'AsEnumerable', 'AsAsyncEnumerable', 'ToList', or 'ToListAsync'. See https://go.microsoft.com/fwlink/?linkid=2101038 for more information.
```

src/GovUk.Education.ExploreEducationStatistics.Content.Services/PublicationService.cs
```diff
@@ -66,4 +69,14 @@ private List<ReleaseTitleViewModel> ListPublishedReleases(Publication publicatio
...
     }
+
+    private bool IsSuperseded(Publication publication)
+    {
+        return publication.SupersededById != null
+               // To be superseded, superseding publication must have Live release
+               && _contentDbContext.Releases
+                   .Where(r => r.PublicationId == publication.SupersededById)
+                   .ToList()
+                   .Any(r => r.Live);
```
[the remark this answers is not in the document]
markhiveit: Talked to Nick about this on Slack. I've temporarily removed the `NotMapped` annotation from `Live` in `Release.cs`, and that allows me to now use the line of code above. I created a migration after removing the annotation, and it didn't create a column, so it seems like this should be fine.

src/explore-education-statistics-admin/src/pages/publication/components/__tests__/PublicationForm.test.tsx
```diff
@@ -393,4 +400,246 @@ describe('PublicationForm', () => {
...
+      title: 'Publication 3',
+      releases: [testRelease2, testRelease3],
+    };
+
+    const testPublication4: MyPublication = {
+      ...testPublication1,
+      id: 'publication-id-4',
+      title: 'Publication 4',
+      releases: [testRelease4],
+    };
```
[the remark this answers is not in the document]
markhiveit: I've removed some of the test data.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -82,6 +82,22 @@ public class PublicationService : IPublicationService
                     }));
         }
 
+        public async Task<Either<ActionResult, List<TitleAndIdViewModel>>> ListPublications()
+        {
+            return await _userService
+                .CheckCanManageAllUsers()
```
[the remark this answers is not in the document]
markhiveit: It was my intention to leave this endpoint as BAU only. The only reason I've moved it is because it's now used for both user management and on the Manage publication page - but in both cases its only used by BAU users.

src/GovUk.Education.ExploreEducationStatistics.Admin/Mappings/MyPublicationPermissionsResolver.cs
```diff
@@ -32,6 +32,10 @@ public MyPublicationPermissionsResolver(IUserService userService)
                     .CheckCanUpdatePublicationTitle()
                     .Result
                     .IsRight,
+                CanUpdatePublicationSupersededBy = _userService
```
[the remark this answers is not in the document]
markhiveit: I double checked with Cam and only BAU users should have the ability to set a superseding publication.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/PublicationService.cs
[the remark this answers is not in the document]
markhiveit: Discussed this with Nick again. He says `Think it'd be good if we could try something else. Can you make a PublicationExtensions and make the IsLive method an extension method of Publication? Then try it again`

src/explore-education-statistics-admin/src/pages/publication/components/PublicationForm.tsx
```diff
@@ -152,6 +175,32 @@ const PublicationForm = ({
               />
             </FormFieldset>
 
+            {publications && showSupersededBy && (
+              <FormFieldset
+                id="supersede"
+                legend="Archive this publication"
+                legendSize="m"
+                hint="Select the publication that will be superseding this publication."
```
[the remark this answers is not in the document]
markhiveit: Talked to Nick and Cam about this. Updated the hint text and moved it to the `FormFieldSelect`.

src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/GovUk.Education.ExploreEducationStatistics.Publisher.Tests.csproj
```diff
@@ -20,6 +20,7 @@
     </ItemGroup>
 
     <ItemGroup>
+        <ProjectReference Include="..\GovUk.Education.ExploreEducationStatistics.Admin.Tests\GovUk.Education.ExploreEducationStatistics.Admin.Tests.csproj" />
```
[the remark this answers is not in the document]
markhiveit: This was to access `DbUtils` in `PublishingServiceTests.cs` for the test `PublishStagedReleaseContent`.

`PublishingService#PublishStagedReleaseContent` now needs access to the Content DB to determine what releases are superseded so it can delete the relevant `PublicationCacheKey`s.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServicePermissionTests.cs
```diff
@@ -145,6 +145,21 @@ public void GetMyPublicationsAndReleasesByTopic_CanViewRelatedReleases()
             publicationRepository.VerifyNoOtherCalls();
         }
 
+        [Fact]
+        public async Task ListPublications()
+        {
+            var mocks = Mocks();
+
+            await PermissionTestUtils.PolicyCheckBuilder<SecurityPolicies>()
+                .ExpectCheckToFail(SecurityPolicies.CanManageUsersOnSystem)
```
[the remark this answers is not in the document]
markhiveit: As above, double checked with Cam and this should be BAU only.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Functions/PublishReleaseContentFunction.cs
markhiveit: I've just used `ForEachAwaitAsync` rather than refactored the blob service. I reached for `foreach` here because of that previous issue with the `ForEachAwaitAsync` stuff and now I don't trust it. But if you're happy that's good here I'll go with that.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/PublishingService.cs
markhiveit: I did manually test this at one point locally - but probably made a lot of changes since then. I'll make sure to highlight to @luke-h1 that scheduled releases also need testing with the archiving stuff.

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
markhiveit: Discussed this with Ben on Slack. Made the change to ensure permalinks under superseded publications now show an out-of-date warning. Also updated unit tests and add UI tests to check permalinks work correctly around this change.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -82,6 +82,22 @@ public class PublicationService : IPublicationService
...
+                .OnSuccess(_ =>
+                {
+                    return _context.Publications
+                        .Select(p => new TitleAndIdViewModel
```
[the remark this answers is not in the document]
markhiveit: There is already a `PublicationSummaryViewModel`, so I've switched to that.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Cache/PublicationTreeCacheKey.cs
```diff
@@ -23,9 +23,11 @@ public static string GetKey(PublicationTreeFilter? filter = null)
             {
                 PublicationTreeFilter.AnyData => "publication-tree-any-data.json",
                 PublicationTreeFilter.LatestData => "publication-tree-latest-data.json",
+                PublicationTreeFilter.LatestDataNotSuperseded => "publication-tree-latest-data-not-superseded.json",
+                PublicationTreeFilter.NotSuperseded => "publication-tree-not-superseded.json",
```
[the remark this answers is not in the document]
markhiveit: Refactored this so that we now only use one cache file - `publication-tree.json`. The appropriate publication tree is generated by filtering the data in this cache file, based on the filter provided to `GetPublicationTree`

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
```diff
@@ -1441,55 +1358,649 @@ await using (var context = InMemoryContentDbContext(contextId))
...
+
+                var publications = result[0].Topics[0].Publications;
+
+                Assert.Single(publications);
+                Assert.Equal("Publication A", publications[0].Title);
+            }
+        }
+
+        [Fact]
+        public async Task GetPublicationTree_DataTables_DoesNotFilterPublicationWithDataOnLatestRelease()
```
ntsim, line 1808: Not sure this makes sense in the context of what the test actually asserts.  The latest release doesn't have any data on it, so it's actually filtered out.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
```diff
@@ -1441,55 +1358,649 @@ await using (var context = InMemoryContentDbContext(contextId))
...
+                Assert.Equal("Topic A", result[0].Topics[0].Title);
+
+                var publications = result[0].Topics[0].Publications;
+                Assert.Single(publications);
+                Assert.Equal("Publication A", publications[0].Title);
+            }
+        }
+
+        [Fact]
+        public async Task GetPublicationTree_DataCatalogue_SingleThemeTopic()
```
ntsim, line 1974: These test cases are a little inconsistent as we don't seem to have the `SingleThemeTopic` on some of the other filter types. In the majority of test cases, it seems like we assert to a single theme/topic so I'm not sure if this case adds an value as it's covered many times over.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
```diff
@@ -1544,94 +2039,122 @@ await using (var context = InMemoryContentDbContext(contextId))
...
+                Assert.Equal("publication-b", publications[1].Slug);
+                Assert.Equal("Publication B", publications[1].Title);
+                Assert.Equal(PublicationType.NationalAndOfficial, publications[1].Type);
+                Assert.Null(publications[1].LegacyPublicationUrl);
             }
         }
 
         [Fact]
-        public async Task GetPublicationTree_AnyData_DoesNotFilterPublicationWithNoDataOnLatestRelease()
+        public async Task GetPublicationTree_DataCatalogue_MultipleThemesTopics()
```
ntsim, line 2075: These test cases are a little inconsistent as we don't seem to have the `MultipleThemesTopics` on some of the other filter types. We might want to add this case for all filter types.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
```diff
@@ -2086,73 +2596,58 @@ await using (var context = InMemoryContentDbContext(contextId))
         }
 
         [Fact]
-        public async Task GetPublicationTree_LatestData_FiltersThemesWithNoVisiblePublications()
+        public async Task GetPublicationTree_DataCatalogue_DoesNotFilterPublicationWithDataOnLatestRelease()
```
ntsim, line 2599: This probably makes more sense as `DoesNotFilterPublicationWithoutDataOnLatestRelease`

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
```diff
@@ -2387,23 +2842,32 @@ await using (var context = InMemoryContentDbContext(contextId))
...
+                Assert.Equal("publication-b", publications[1].Slug);
+                Assert.Equal("Publication B", publications[1].Title);
+                Assert.Equal(PublicationType.NationalAndOfficial, publications[1].Type);
+                Assert.Null(publications[1].LegacyPublicationUrl);
             }
         }
 
         [Fact]
-        public async Task GetPublicationTree_LatestData_DoesNotFilterPublicationWithDataOnLatestRelease()
+        public async Task GetPublicationTree_DataCatalogue_FiltersPublicationWithNoDataOnLatestAmendedRelease()
```
ntsim, line 2870: Seems like this is test case would assert the same thing as another test case above: `GetPublicationTree_DataCatalogue_FiltersPublicationWithNoDataOnAmendedLatestRelease`

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests.csproj
```diff
@@ -19,6 +19,7 @@
     </ItemGroup>
 
     <ItemGroup>
+      <ProjectReference Include="..\GovUk.Education.ExploreEducationStatistics.Admin.Tests\GovUk.Education.ExploreEducationStatistics.Admin.Tests.csproj" />
```
ntsim, line 22: Why are we including this reference? If we're trying to access some util, it should probably be moved to another test project like `Common.Tests` or `Data.Model.Tests`

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
```diff
@@ -97,14 +109,38 @@ private static JsonSerializerSettings BuildJsonSerializerSettings()
...
+                .Select(r => r.Publication)
+                .Single();
+
+            return publication.SupersededById != null
+                   && _contentDbContext.Releases
+                       .Include(p => p.Publication)
+                       .Where(r => r.PublicationId == publication.SupersededById)
+                       .ToList()
+                       .Any(r => r.IsLatestPublishedVersionOfRelease());
+        }
```
ntsim, line 143: Seems like we're doing a lot of the same queries that are being performed in `_subjectRepository.IsSubjectForLatestPublishedRelease`. This is pretty inefficient as we're doing a lot of unnecessary database calls.

Is it possible to combine these two? Or at the very least, consolidate this logic into `_subjectRepository` with a new method?

src/explore-education-statistics-admin/src/pages/publication/components/__tests__/PublicationForm.test.tsx
```diff
@@ -393,4 +400,207 @@ describe('PublicationForm', () => {
...
+            contactName: 'John Smith',
+          }}
+          showTitleInput
+          showSupersededBy={false}
+          onSubmit={noop}
+        />,
+      );
+
+      await waitFor(() => {
+        expect(screen.getByLabelText('Publication title'));
```
ntsim, line 516: Mentioned this with @amyb-hiveit, but we should be consistent and add the `toBeInTheDocument` in Jest tests

src/explore-education-statistics-admin/src/pages/publication/components/__tests__/PublicationForm.test.tsx
```diff
@@ -393,4 +400,207 @@ describe('PublicationForm', () => {
...
+            contactTelNo: '0123456789',
+            contactName: 'John Smith',
+          }}
+          showSupersededBy
+          onSubmit={noop}
+        />,
+      );
+
+      await waitFor(() => {
+        expect(screen.getByLabelText('Superseding publication'));
```
ntsim, line 589: Would add `toBeInTheDocument`

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
```diff
@@ -2471,98 +2936,106 @@ await using (var context = InMemoryContentDbContext(contentDbContextId))
...
+                var publications = result[0].Topics[0].Publications;
+                Assert.Equal(2, publications.Count);
+
+                Assert.Equal("Publication A", publications[0].Title);
+                Assert.False(publications[0].IsSuperseded);
+
+                Assert.Equal("Publication B", publications[1].Title);
+                Assert.True(publications[1].IsSuperseded);
             }
         }
```
ntsim, line 3041: Where are all the `FastTrack` tests? :wink: 

Not going to ask you to do these, but you should probably be able to appreciate why having filters for specific screens might not be the best idea.

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
markhiveit: I'm not sure I have an issue with this - the FastTrack tests are covered by the DataCatalogue tests - if they differ in the future, unit tests for FastTrack can be written at that point. I don't see much of a downside.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
markhiveit: The current code does make a DB call to get the `ReleaseSubject` twice
, so it's possible I could make that call once and then pass the `ReleaseSubject` to the relevant methods - but the downside to making one less call is the code will be harder to understand.




I did consider moving `PermalinkService#IsSuperseded` to `SubjectRepository` - but that would have required adding `ContentDbContext` to `SubjectRepository` and so decided against it

, but maybe that is fine?


On an aside, I'm not actually a fan of `SubjectRepository#IsSubjectForLatestPublishedRelease`, as it only checks the Stats DB, and that means that if a new release is published without any subject, then no new rows get created in the Stats DB, and the method actually says the subject is for the latest published release when it isn't

. I think we've accepted this so far because it is expected that all releases will have subjects attached - but it seems plausible that in the future that may not be the case - I'd prefer to cover that eventuality now.

I'm inclined to just leave this. This code is going to get changed anyway with the permalink error stuff (EES-3143). But if you still want me to , I'm happy to pull the `ReleaseSubject` DB call out so it only gets made once and/or move the `IsSuperseded` method to `SubjectRepository`.

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests.csproj
markhiveit: This reference is for `DbUtils`. 

If I move it to `Common.Tests`, I'll have to add things like `Content.Model` to `Common.Tests`, and that doesn't seem right.

I think the real home for `DbUtils` is `Data.Model.Tests`, but I'll have to add a lot of references, like adding `IdentityServer4` to `Data.Model.Tests`, which maybe means maybe that should be moved to?

I don't see a big downside to just adding the Admin.Tests reference here, as it is just for unit tests, and it's probably going to take a while to move DbUtils to Data.Model.Tests as it is used a lot. So I'm inclined to leave this now and create a ticket for it. But happy to go along with what you'd prefer.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
```diff
@@ -97,14 +109,38 @@ private static JsonSerializerSettings BuildJsonSerializerSettings()
...
+                .Where(r => r.Id == releaseId)
+                .Select(r => r.Publication)
+                .Single();
+
+            return publication.SupersededById != null
+                   && _contentDbContext.Releases
+                       .Include(p => p.Publication)
+                       .Where(r => r.PublicationId == publication.SupersededById)
+                       .ToList()
+                       .Any(r => r.IsLatestPublishedVersionOfRelease());
```
benoutram, line 142: Has the publication got _all_ of it's Releases included to be able to do this check? I suspect it might only have one of it's releases loaded.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
markhiveit: Ben suggested how I should go about this on Slack so I've done that - probably just over complicating it in my mind.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
markhiveit: They appear to be doing the same thing so I've removed one

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
markhiveit: I've gone with `DoesFilterPublicationWithoutDataOnLatestRelease` - I imagine this is what you intended.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
markhiveit: There is one for `FindStatistics`, `DataTables`, and `DataCatalogue`, which covers everything - I've renamed the `FindStatistics` one from `GetPublicationTree_MultipleThemesTopics` to `GetPublicationTree_FindStatistics_MultipleThemesTopics` so it's clearer

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Controllers/ThemeController.cs
```diff
@@ -28,9 +29,8 @@ public class ThemeController : ControllerBase
         }
 
         [HttpGet("themes")]
-        [BlobCache(typeof(PublicationTreeCacheKey))]
-        public async Task<IList<ThemeTree<PublicationTreeNode>>> GetThemes(
-            [FromQuery(Name = "publicationFilter")] PublicationTreeFilter? filter = null)
+        public async Task<IList<ThemeTree<PublicationTreeNode>>> GetPublicationTree(
+            [Required][FromQuery(Name = "publicationFilter")] PublicationTreeFilter filter)
```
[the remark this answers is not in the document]
markhiveit: So I tried this with an without `[Required]` and it made no difference to the behaviour.

If you hit the endpoint with `GET {content_api_url}/themes?publicationFilter=blah` or just `GET {content_api_url}/themes`, it'll default to `FindStatistics` because it's first in the `PublicationTreeFilter` enum. This wasn't the behaviour I was expecting.

`GET {content_api_url}/themes?publicationFilter=DataTables` behaves as expected.

I'm inclined to leave this - I don't think defaulting to `FindStatistics` is a problem. I've removed the `[Required]` since it doesn't appear to do anything. But happy to do something else if others prefer.

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -0,0 +1,346 @@
...
+    user clicks radio    ${PUBLICATION_NAME_SUPERSEDE}
+    user clicks button    Next step
+    user waits until page contains    Choose a release
+    user waits until page contains    ${RELEASE_NAME_SUPERSEDE}
+
+    user checks page contains radio    ${RELEASE_NAME_SUPERSEDE}
+    user checks page contains    This is the latest data
+
+Check archive-publication permalink has out-of-date warning
+    go to    ${PERMALINK_URL}
```
luke-h1, line 250: I believe you'll need to use the `user goes to public frontend` KW in order to avoid issues with how basic auth works on the public FE

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -0,0 +1,346 @@
...
+    user clicks radio    ${PUBLICATION_NAME_SUPERSEDE}
+    user clicks button    Next step
+    user waits until page contains    Choose a release
+    user waits until page contains    ${RELEASE_NAME_SUPERSEDE}
+
+    user checks page contains radio    ${RELEASE_NAME_SUPERSEDE}
+    user checks page contains    This is the latest data
+
+Check archive-publication permalink no longer has out-of-date warning after archive-publication has been unarchived
+    go to    ${PERMALINK_URL}
```
luke-h1, line 341: I believe you'll need to use the user goes to public frontend KW in order to avoid issues with how basic auth works on the public FE

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -0,0 +1,346 @@
...
+${RELEASE_NAME_ARCHIVE}=            Financial Year 3000-01
+${SUBJECT_NAME_ARCHIVE}=            Subject for archived publication
+
+${PUBLICATION_NAME_SUPERSEDE}=      UI tests - superseding publication %{RUN_IDENTIFIER}
+${RELEASE_NAME_SUPERSEDE}=          Financial Year 2000-01
+${SUBJECT_NAME_SUPERSEDE}=          Subject for superseding publication
+
+*** Test Cases ***
+Create new publication to be archived and release via API
+    ${PUB_ID_ARCHIVE}=    user creates test publication via api    ${PUBLICATION_NAME_ARCHIVE}
```
luke-h1, line 23: I would change this to `PUBLICATION_ID_ARCHIVE` to make it more obvious that this means publication and not public

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -0,0 +1,346 @@
...
+    user waits until modal is visible    Confirm publication changes
+    user clicks button    Confirm
+    user waits until modal is not visible    Confirm publication changes
+
+Validate archive warning is on Admin dashboard for archive-publication release
+    ${accordion}=    user opens publication on the admin dashboard    ${PUBLICATION_NAME_ARCHIVE}
+    user checks element should contain    ${accordion}
+    ...    This publication will be archived when its superseding publication has a live release published.
+
+Check public site, that archive-publication still appears correctly
```
luke-h1, line 74: Would word this a little better. Something such as Validate that archive-publication appears correctly on public frontend`

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -0,0 +1,346 @@
...
+    user opens details dropdown    %{TEST_TOPIC_NAME}
+    user waits until details dropdown contains publication    %{TEST_TOPIC_NAME}    ${PUBLICATION_NAME_ARCHIVE}
+    user checks publication bullet contains link    ${PUBLICATION_NAME_ARCHIVE}    View statistics and data
+
+    user clicks element    testid:View stats link for ${PUBLICATION_NAME_ARCHIVE}
+
+    user waits until h1 is visible    ${PUBLICATION_NAME_ARCHIVE}    %{WAIT_MEDIUM}
+    user waits until page contains    This is the latest data
+    ${PUB_ARCHIVE_URL}=    user gets url
+    set global variable    ${PUB_ARCHIVE_URL}
```
luke-h1, line 90: What's the difference between `set suite variable` and `set global variable`?

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -0,0 +1,346 @@
...
+
+Check Data tables and Data catalogue pages, that archive-publication still appears correctly
```
luke-h1, line 92: Would word this differently. Something such as `Check that archive-publication appears correctly on data-tables & data-catalogue pages`

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -0,0 +1,346 @@
...
+    user opens details dropdown    %{TEST_TOPIC_NAME}
+
+    user checks page does not contain    ${PUBLICATION_NAME_SUPERSEDE}
+
+    user clicks radio    ${PUBLICATION_NAME_ARCHIVE}
+    user clicks element    id:publicationForm-submit
+    user waits until table tool wizard step is available    2    Choose a subject
+    user checks previous table tool step contains    1    Publication    ${PUBLICATION_NAME_ARCHIVE}
+
+    user checks page contains    ${SUBJECT_NAME_ARCHIVE}
```
luke-h1, line 105: Can we be more specific with this selector?

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -0,0 +1,346 @@
...
+
+    user waits until results table appears    %{WAIT_SMALL}
+    user waits until page contains button    Generate shareable link
+
+    user clicks button    Generate shareable link
+    user waits until page contains testid    permalink-generated-url    %{WAIT_MEDIUM}
+    user checks generated permalink is valid
+
+    user clicks link    View share link
+    user waits until h1 is visible
```
luke-h1, line 134: Would add a wait here to prevent issues in CI `%{WAIT_SMALL}`

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -0,0 +1,346 @@
...
+    ...    '${SUBJECT_NAME_ARCHIVE}' from '${PUBLICATION_NAME_ARCHIVE}'
+
+    ${PERMALINK_URL}=    user gets url
+    set global variable    ${PERMALINK_URL}
+
+    user checks page does not contain    WARNING
+
+Check Data catalogue page, that archive-publication still appears correctly
```
luke-h1, line 142: would reword this, similar to my other comments

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -0,0 +1,346 @@
...
+    ...    ${SUBJECT_NAME_SUPERSEDE} data guidance content
+
+    user clicks button    Save guidance
+    user waits until page contains button    Edit guidance
+
+Go to "Sign off" page and approve superseding-publication release
+    user clicks link    Sign off
+    user approves original release for immediate publication
+
+Check public Find stats page, that archive-publication is now archived and superseding-publication appears
```
luke-h1, line 182: would reword this, similar to my other comments

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -0,0 +1,346 @@
...
+
+    user waits until h1 is visible    ${PUBLICATION_NAME_SUPERSEDE}    %{WAIT_MEDIUM}
+    user waits until page contains    This is the latest data
+
+Check public archive-publication release page displays correctly
+    go to    ${PUB_ARCHIVE_URL}
+    user waits until h1 is visible    ${PUBLICATION_NAME_ARCHIVE}    %{WAIT_MEDIUM}
+    user checks page does not contain    This is the latest data
+
+Check public data tables page is correct
```
luke-h1, line 206: Would reword this to be more descriptive

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -0,0 +1,346 @@
...
+    user checks page does not contain    ${PUBLICATION_NAME_ARCHIVE}
+
+    user clicks radio    ${PUBLICATION_NAME_SUPERSEDE}
+    user clicks element    id:publicationForm-submit
+    user waits until table tool wizard step is available    2    Choose a subject
+    user checks previous table tool step contains    1    Publication    ${PUBLICATION_NAME_SUPERSEDE}
+
+    user checks page contains    ${SUBJECT_NAME_SUPERSEDE}
+
+Check data catalogue page is correct
```
luke-h1, line 221: Would reword this to be more descriptive

luke-h1 (changes requested):

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
markhiveit: `set global variable` sets it for all tests and test suites. I'll switch to using `set suite variable`.

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
markhiveit: There is no way that this is going to be detected on a different part of the page, so I don't see any issue with checking the whole page here.

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Controllers/ThemeController.cs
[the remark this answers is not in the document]
markhiveit: After discussing on Slack with Ben, I've made `filter` nullable and return a `BadRequestResult` if it is null.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -205,8 +239,25 @@ private async Task DeleteCachedTaxonomyBlobs()
...
+        {
+            // NOTE: When a publication's SupersededById column is updated, we only need to invalidate the cache
+            // if the updated publication is now superseded OR if the updated publication is now no longer superseded.
+            // If the updated publication has a new superseding publication or a different superseding publication,
+            // any old or new superseding publication will be unaffected, so there is no need to invalidate their caches
+            var supersededPublications = await _context.Publications
+                .Where(p =>
+                    p.SupersededById == publication.Id
+                    || (originalSupersededById == null
+                        && p.Id == originalSupersededById))
```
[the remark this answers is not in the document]
markhiveit: Not sure how I messed up the logic of this. It seems everything after the `||` is unnecessary - so only deletes cache blobs that it doesn't need to.

I'll write / update tests.

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests.csproj
ntsim: Let's leave it for now as it sounds like it'd be a lot of work. You're right that just having to a reference to `Admin.Tests` isn't really causing any harm

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
ntsim: Ideally, we want any tests to be value for money. With the current enum approach, we're having to do a lot of duplication of tests, but is necessary as each enum can represent drastically different scenarios.

It's entirely possible that the conditions for the `FastTracks` enum don't align with the `DataCatalogue` enum in the future and there are no tests to actually catch any problems that might arise from that.

Of course, it's not great to have to do a bunch of duplication to test a different code path, but that's the architecture we've gone with currently. Unfortunately, the flip-side is that leaving the tests to be written by someone in the future isn't very responsible :wink:

We can leave this for now as my money is on us re-writing a lot of this in the near future anyway.

ntsim (approved):

ntsim:
@markhiveit  the build is currently failing

benoutram (approved):

PR #3229, opened 2022-03-28
EES-3258 fix table layout when missing region

amyb-hiveit:
Fixes table layout when LADs don't have regions, which causes rows to be mis-aligned.

The fix:
- adds an empty header in the case of a location filter with an empty string for group (when a group isn't expected group is undefined)
- when the table is rendered converts this header into an empty <td> as empty <th> cause accessibility problems
- adds some styling to make it look ok.

For testing
- the original files are attached to the ticket
- please test with reordering the rows and columns, including moving the locations into columns
- please check there are no problems with other location groupings / combinations

### DIFF
--- a/src/explore-education-statistics-common/src/modules/table-tool/components/MultiHeaderTable.tsx
+++ b/src/explore-education-statistics-common/src/modules/table-tool/components/MultiHeaderTable.tsx
@@ -161,9 +161,9 @@ const MultiHeaderTable = forwardRef<HTMLTableElement, MultiHeaderTableProps>(
       >
         <thead className={styles.tableHead}>
           {expandedColumnHeaders.map((columns, rowIndex) => {
+            const headingRowKey = `row-${rowIndex}`;
             return (
-              // eslint-disable-next-line react/no-array-index-key
-              <tr key={rowIndex}>
+              <tr key={headingRowKey}>
                 {rowIndex === 0 && (
                   <td
                     colSpan={rowHeaderColumnLength}
@@ -174,6 +174,17 @@ const MultiHeaderTable = forwardRef<HTMLTableElement, MultiHeaderTableProps>(
 
                 {columns.map((column, columnIndex) => {
                   const key = `${column.text}_${columnIndex}`;
+                  // Add an empty td instead of a th for empty group headers
+                  if (column.id === '') {
+                    return (
+                      <td
+                        key={key}
+                        colSpan={column.span}
+                        rowSpan={column.crossSpan}
+                        className={styles.emptyColumnHeaderCell}
+                      />
+                    );
+                  }
 
                   return (
                     <th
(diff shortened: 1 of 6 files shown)


src/explore-education-statistics-common/src/modules/table-tool/components/TimePeriodDataTable.tsx
```diff
@@ -125,9 +125,14 @@ function optimizeFilters(filters: Filter[], headerConfig: Filter[][]) {
           header => header.group !== firstSubGroup,
         );
 
+        // LADs can be added without regions, which results as an empty string for `group` and can cause table layout problems.
+        // So include it as an empty group and change to a td later so don't have empty header cells
+        const isMissingLocationGroup =
+          filter instanceof LocationFilter && filter.group === '';
```
benoutram, line 131: I think you could reword the comment about the LAD level and it's region attribute here as the problem isn't specific to LAD's and regions, it's more general than that.

The location hierarchy functionality is generic and can be applied for any level of data. The attribute which is grouped is always the principal attribute for that level of data, so for LA level data the LA attribute gets grouped. The attributes which it's grouped by can be any other attribute.

Our current configuration:

* LA level data - LA attribute is grouped by Region.
* LAD level data - LAD attribute is grouped by Region.

Tomorrow we could introduce a new attribute like "Provider Group" and configure Providers to be grouped by Provider Group.

It should also be the case that we can can configure multiple levels of hierarchies, e.g. we could have grouped LAD by Country and Region or Provider by Country as well as Provider Group. The BE supports this although I'm not sure how well tested that's been on the FE. 

The problem we are solving here is that there's nothing requiring analysts to import data that includes all of the attributes of the hierarchy. The screener might warn against that but it doesn't mean the data won't exist and historically data may exist without all the attributes too.

So with our current LA config, this example data imported would be grouped by an empty Region.

```
geographic_level    country_code    country_name    region_code    region_name    la_code    la_name
====================================================================================================
Local authority     E92000001       England                                       E06000015  Derby
Local authority     E92000001       England                                       E06000015  Derby
Local authority     E92000001       England                                       E06000018  Nottingham
```

I'm not sure how likely it is but there could be a scenario where *some* of the rows have location attributes not present in the hierarchy, but not all. E.g. Derby appears with and without a Region.

```
geographic_level    country_code    country_name    region_code    region_name    la_code    la_name
====================================================================================================
Local authority     E92000001       England         E12000004      East Midlands  E06000015  Derby
Local authority     E92000001       England                                       E06000015  Derby
Local authority     E92000001       England         E12000004      East Midlands  E06000018  Nottingham
```

src/explore-education-statistics-common/src/modules/table-tool/components/MultiHeaderTable.tsx
```diff
@@ -197,40 +208,63 @@ const MultiHeaderTable = forwardRef<HTMLTableElement, MultiHeaderTableProps>(
...
+                  const key = `header-${headerIndex}`;
+
+                  // Add an empty td instead of a th for empty group headers
+                  if (header.id === '') {
+                    return (
+                      <td
+                        key={key}
+                        className={classNames(
+                          styles.emptyRowHeaderCell,
+                          // 'fixed',
```
benoutram, line 225: Should this comment be cleaned up here?

amyb-hiveit: yep, sorted now thanks.

benoutram (approved):

PR #3247, opened 2022-04-05
EES-3252 / EES-2459 / EES-3310 LegacyReleaseService and PublicationService was deleting blobs from wron…

markhiveit:
LegacyReleaseService and PublicationService were deleting blobs from wrong blob storage container.

In `Startup.cs`, both services weren't explicitly wired up to use the public cache, and so were incorrectly trying to delete `PublicationCacheKey` from the private cache.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
@@ -291,7 +291,19 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin
             );
             services.AddTransient<IThemeService, ThemeService>();
             services.AddTransient<ITopicService, TopicService>();
-            services.AddTransient<IPublicationService, PublicationService>();
+            services.AddTransient<IPublicationService, PublicationService>(provider =>
+                new PublicationService(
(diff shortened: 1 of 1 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
```diff
@@ -298,7 +294,17 @@ public void ConfigureServices(IServiceCollection services)
             services.AddTransient<IPublicationService, PublicationService>();
             services.AddTransient<IPublicationRepository, PublicationRepository>();
             services.AddTransient<IMetaService, MetaService>();
-            services.AddTransient<ILegacyReleaseService, LegacyReleaseService>();
+            services.AddTransient<ILegacyReleaseService, LegacyReleaseService>(provider =>
```
benoutram, line 297: Same problem exists with `_publicBlobCacheService` that was recently added to `PublicationService`.

Also because of this, while we can easily clear the cached tree json blobs, I don't think we have a way of clearing all publication json blobs that may be now out of date without using the publisher functions method of republishing all content.

benoutram (changes requested):

benoutram (approved):

PR #3257, opened 2022-04-11
EES-3318 - update help & support page

luke-h1:
This PR: 
- updates the help & support page to mention 'Data catalogue' instead of `download latest data` due to that page no longer existing. Previously this went unnoticed due to the redirect we have in place for this old slug.


cjrace:
Change to lower case 'c' for catalogue, as that's what we've used elsewhere. Good to merge once that's updated though

cjrace (changes requested):
Change to lower case 'c' for catalogue, as that's what we've used elsewhere. Good to merge once that's updated though

cjrace (approved):

PR #3259, opened 2022-04-12
Ees 3142 retain existing latest map boundary level when latest chosen

duncan-at-hiveit:
EDIT: added do not merge to this so we can set up some test data prior to Boundary Level becoming mandatory.

### Overview
The focus of this PR is to address Maps that do not have specific Boundary Level GeoJSON chosen as part of their configuration.

The absence of a particular choice was supposed to indicate that the content creator was choosing the latest Boundary Level choice at the time that was available at the point that the content was created.

The upshot of this was that no Boundary Level Id (of the latest at the time) was stored in the database.  When an even newer Boundary Level was introduced for the matching Geographic Level of the Map, this broke a lot of Maps as the newer version of the GeoJSON was incompatible (missing LAs in the newer version of the GeoJSON, for example due to updated codes).

An additional problem was that a Data Block that a Map is created from can encompass several Geographic Levels but the intention is for a Map to be tied to a specific single Geographic Level.  Maps that were created from Data Blocks like this and without a specific Boundary Level set were ambiguous in terms of what Geographic Level they were supposed to be representing.  This led to some odd behaviour.

### Work included in this PR

The main portion of the fix here is to run a migration to set a specific Boundary Level for every existing Map per environment.  This is manually triggered.  We identify the appropriate Boundary Level to set for each map based on the following rules:
- If the map contains data sets for only a single Geographic Level, then choose the "latest at map creation time" Boundary Level for that Geographic Level
- If the map contains data sets for National Geographic Level and a single other Geographic Level, then choose the "latest at map creation time" Boundary Level for the non-National one
- If the map contains data sets for multiple non-National Geographic Levels, then we need to contact the teams to determine the correct boundary level to set

The secondary change in this PR is to make the selection of Boundary Level mandatory for Maps.  This not only sets a specific Boundary Level for its GeoJSON with no ambiguity at all, but also locks the Map to a particular GeographicLevel, that of the Boundary Level.  It would be nice at this point to limit the ability to add data sets only of the matching Geographic Level to the chart, but this has not been done as a part of this work.  It's therefore possible to choose a Boundary Level for Regions and then subsequently add data sets for LAs or National, and the upshot will just be that that data will not be visible in the map itself.  Given that it seems that 99% of the Analysts are specifically choosing data sets of a single Geo Level manually right now anyway, this isn't a biggie - just not 100% intuitive.  I've raised https://dfedigital.atlassian.net/browse/EES-3329 for looking into this.

### Additional work in this PR and future work.

I have included `BoundaryLevel` within the MapChart configuration now, as this is what is used within the ChartRenderer components.  This means we should be able to extract it from the ObservationQueryContext in the future.  I have raised https://dfedigital.atlassian.net/browse/EES-3328 to look at this in the future.  

I have removed the `IncludeGeoJson` flag.  This was a flag to indicate to the back end that the front end needs GeoJSON to be returned within the table builder "results" array items.  Instead, we now return GeoJSON when a Boundary Level Id is sent instead, given that this is now a mandatory field for maps.  I've raised https://dfedigital.atlassian.net/browse/EES-3332 as we're asking for GeoJson where we don't need to, for example when viewing a plain table.  This is existing behaviour as seen on the dev branch currently.

### Rolling out the work
- Deploy the code.
- Back up the ContentBlock table.
- Perform a dry run of the migration using `PATCH /releases/migrate-all-maps` to see how the various Maps on the environment will be migrated.  If any report that manual intervention needs to take place to determine the correct Boundary Level to set on a Map, get the teams to carry out the appropriate changes.
- Perform a full run of the migration to migrate all the Maps using `PATCH /releases/migrate-all-maps?dryRun=false`.
- Run a full content refresh (i.e. the destructive one run by the Publisher that clears all Release json and then recreates it).

### Follow-on work
https://dfedigital.atlassian.net/browse/EES-3319

image


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockMigrationService.cs
```diff
@@ -1,40 +1,456 @@
...
+                    return migrationResults;
                 });
         }
+
+        private async Task SaveChanges(DataBlock dataBlock, bool dryRun)
+        {
+            if (!dryRun)
+            {
+                _contentDbContext.DataBlocks.Update(dataBlock);
+                await _contentDbContext.SaveChangesAsync();
```
markhiveit, line 196: Would it be possible to call `SaveChangesAsync` once at the end to prevent the migration from partially running?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockMigrationService.cs
```diff
@@ -1,40 +1,456 @@
...
+                                        plan.Strategy,
+                                        null,
+                                        null);
+                                case FixMapConfigInvalidDataSetGeographicLevels:
+                                    return new MapMigrationResult(
+                                        plan.DataBlockId,
+                                        plan.Type,
+                                        plan.Strategy,
+                                        null,
+                                        null);
```
markhiveit, line 179: I think you could stack these three cases and just have one return for all three? Realise this is getting a bit pedantic for migration code :)

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockMigrationService.cs
```diff
@@ -1,40 +1,456 @@
...
+            var boundaryLevelsWithDateRanges = GetBoundaryLevelsWithDateRanges(boundaryLevels);
+            return boundaryLevelsWithDateRanges
+                .Single(boundary =>
+                {
+                    var (level, start, end, _) = boundary;
+
+                    return level == geographicLevel &&
+                           dataBlockCreated.CompareTo(start) >= 0 &&
+                           dataBlockCreated.CompareTo(end) <= 0;
+                })
```
markhiveit, line 272: As I understand it, this works because `boundaryLevelsWithDateRanges` is ordered? So it selects the earliest created boundary level that the data block's creation date is within its start/end?

markhiveit (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Data/Query/ObservationQueryContext.cs
```diff
@@ -10,27 +9,13 @@ public class ObservationQueryContext
         public Guid SubjectId { get; set; }
         public TimePeriodQuery TimePeriod { get; set; }
         public IEnumerable<Guid> Filters { get; set; }
+        // TODO EES-3328 - remove BoundaryLevel from ObservationQueryContext as we now store it on
```
luke-h1, line 12: Just wondering if it's better to use the `obsolete` attribute in order to make TODOs more obvious in the logs i.e. 
```
[obsolete("TODO EES-3319 - remove backwards-compatibility for Map Configuration without its
                    // own Boundary Level selection")]
```

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/DataBlockMigrationController.cs
```diff
@@ -23,12 +27,14 @@ public DataBlockMigrationController(IDataBlockMigrationService dataBlockMigratio
         {
             _dataBlockMigrationService = dataBlockMigrationService;
         }
-
-        [HttpPatch("releases/migrate-data-blocks")]
-        public async Task<ActionResult<Unit>> MigrateDataBlocks()
+        
+        [AllowAnonymous]
```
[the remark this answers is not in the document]
duncan-at-hiveit: Nyah nice one, this was accidentally left in from testing.  There's still a permission check in the service that would stop anyone from executing the migration but yeah, I'll dispose of this now, thanks!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockMigrationService.cs
```diff
@@ -1,40 +1,456 @@
...
 
-        public async Task<Either<ActionResult, Unit>> Migrate()
+        public async Task<Either<ActionResult, List<DataBlockMapMigrationPlan>>> GetMigrateMapPlans()
+        {
+            var locations = GetAllLocations();
+            var boundaryLevels = await GetAllBoundaryLevels();
+            var mapsNeedingMigration = (await GetAllDataBlocksWithMaps())
+                .Where(dataBlock => 
+                    (dataBlock.Charts[0] as MapChart)?.BoundaryLevel == null 
+                    || dataBlock.Query.BoundaryLevel == null);
```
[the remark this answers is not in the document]
duncan-at-hiveit: So this will be covered by the work in https://dfedigital.atlassian.net/browse/EES-3328

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockMigrationService.cs
```diff
@@ -1,40 +1,456 @@
...
+                FixMapConfigInvalidDataSetGeographicLevels,
+            }
+            
+            public Guid DataBlockId { get; }
+
+            public MigrationType Type { get; } = Undefined;
+
+            public MigrationStrategy Strategy { get; } = FixMapConfigNoDataSetGeographicLevels;
+
+            public long? SpecificBoundaryLevelId { get; }
```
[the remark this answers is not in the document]
duncan-at-hiveit: Good spot - removed!

Originally I had 2 endpoints - one for generating plans like this and another for running the plans.  When I combined the 2 together with the use of the dryRun flag, this became redundant.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockMigrationService.cs
```diff
@@ -1,40 +1,456 @@
...
+            [JsonConverter(typeof(StringEnumConverter))]
+            public GeographicLevel? SpecificBoundaryLevelGeographicLevel { get; }
```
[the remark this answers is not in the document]
duncan-at-hiveit: Removed ta

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockMigrationService.cs
```diff
@@ -1,40 +1,456 @@
...
+                                .DataBlocks
+                                .SingleAsync(db => db.Id == plan.DataBlockId);
+
+                            var map = (MapChart)dataBlock.Charts[0];
+
+                            switch (plan.Strategy)
+                            {
+                                case RetainChosenBoundaryLevel:
+                                {
+                                    map.BoundaryLevel = dataBlock.Query.BoundaryLevel ?? 0;
```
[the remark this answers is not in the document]
duncan-at-hiveit: This field will be removed as part of https://dfedigital.atlassian.net/browse/EES-3329

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockMigrationService.cs
```diff
@@ -1,40 +1,456 @@
...
+using GovUk.Education.ExploreEducationStatistics.Data.Model;
+using GovUk.Education.ExploreEducationStatistics.Data.Model.Database;
 using Microsoft.AspNetCore.Mvc;
-using Microsoft.Extensions.Logging;
+using Microsoft.EntityFrameworkCore;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+using static GovUk.Education.ExploreEducationStatistics.Admin.Services.DataBlockMigrationService.DataBlockMapMigrationPlan.MigrationStrategy;
+using static GovUk.Education.ExploreEducationStatistics.Admin.Services.DataBlockMigrationService.DataBlockMapMigrationPlan.MigrationType;
+using static GovUk.Education.ExploreEducationStatistics.Common.Services.CollectionUtils;
```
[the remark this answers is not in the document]
duncan-at-hiveit: Cool removed thanks

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockMigrationService.cs
duncan-at-hiveit: I'm happy to allow a partial migration as it's rerunnable.  It'll only attempt to migrate Maps that aren't yet migrated, so happy if a partial failure doesn't roll back all the successful operations to date.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockMigrationService.cs
duncan-at-hiveit: Hey no, it's totally valid and a 1 second change - done!

For throwaway migration code, if it was something more involved then I might push back, but this is all good!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockMigrationService.cs
duncan-at-hiveit: Yup that's correct.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/IDataBlockMigrationService.cs
```diff
@@ -1,15 +1,17 @@
...
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces
 {
     /**
-     * Interface used to migrate data blocks. Intended to be adapted as needed when migrations are required.  
+     * Temporary interface used to migrate data blocks for EES-3167 and EES-3142.
      */
     public interface IDataBlockMigrationService
     {
-        Task<Either<ActionResult, Unit>> Migrate();
+        Task<Either<ActionResult, List<DataBlockMigrationService.DataBlockMapMigrationPlan>>> GetMigrateMapPlans();
```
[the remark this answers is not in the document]
duncan-at-hiveit: Sure can!  Removed - as before, when I added the dryRun flag, then this became redundant

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Data/Query/ObservationQueryContext.cs
duncan-at-hiveit: I could do, but it's not always in places where I need to add attributes that I need to comment up.  In the ticket description though for EES-3319 I have said to look for this specific TODO text

markhiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Data/Query/ObservationQueryContext.cs
luke-h1: Cool, sounds good

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ResultSubjectMetaService.cs
```diff
@@ -281,58 +281,39 @@ private static string OrderLocationAttributes(LocationAttributeNode node)
...
+                        var locationAttributeCodes = locationAttributes
+                            .SelectMany(locationAttribute => locationAttribute.GetLeafAttributes())
+                            .Select(locationAttribute => locationAttribute.GetCodeOrFallback())
+                            .WhereNotNull();
+
+                        return _geoJsonRepository.FindByBoundaryLevelAndCodes(boundaryLevelId, locationAttributeCodes);
+                    });
             }
-
-            private bool GeoJsonRequested => _query.IncludeGeoJson != null && _query.IncludeGeoJson.Value;
```
benoutram, line 335: Previously `IncludeGeoJson` on the query was set to true in `GetReleaseDataBlockResults` if the data block contained a map, and here we see `GeoJsonRequested ` was true when `IncludeGeoJson`, and that in turn controlled whether GeoJson was returned or not.

It now seems like we are returning GeoJson all of the time meaning that we could be returning it for table builder results where the geojson is not required  because no chart is rendered in the table builder.

EES-3332 relates to this and suggests this isn't new behaviour but I think it's worth checking again as I think an adverse change could be being made here. We wouldn't want the geojson being returned in table builder responses.

benoutram (changes requested):

benoutram:
I'm not sure that I understand ticket EES-3330 that's been raised and is mentioned in the description of this PR.

I thought the GeoJson is returned within the locations in the `ResultSubjectMetaViewModel` which is going to be included only once in each response. It has a map of the locations keyed by geographic level. There should be one location for each of the locations in the results.

If a location appears in multiple results it should only appear once in the subject meta.

The ticket seems to suggest that GeoJson can be repeated more than once in table results in the 'results' array but those results contain nothing more than a level and location id, or in the case of historical permalinks a `LocationViewModel` containing location codes.

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/Charts.cs
```diff
@@ -59,6 +59,9 @@ public class VerticalBarChart : Chart
     public class MapChart : Chart
     {
         public override ChartType Type => Map;
+        
+        // TODO EES-3319 - make mandatory when all Map Charts are migrated to have a Boundary Level set
+        public long? BoundaryLevel { get; set; }
```
benoutram, line 64: I think it's worth building something into the test plan for this so that we don't forget, that a full content refresh is required after running the migration (i.e. the destructive one run by the Publisher that clears all Release json and then recreates it).

The cache clearing mechanism that we recently introduced for clearing cached data blocks from the public blob cache actually clears cached data block (table) results.

The data block config here that is being altered by the migration won't change because it's cached within Release blobs and they don't get cleared.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockMigrationService.cs
```diff
@@ -1,40 +1,438 @@
...
+                        {
+                            var level = dataSet.Location?.Level;
+
+                            // If no level is specified on this Data Set, it is intended for all
+                            // Geographic Levels, so return all Geo Levels for this Data Block in
+                            // this case.
+                            if (level == null)
+                            {
+                                return DataBlockGeographicLevels;
+                            }
```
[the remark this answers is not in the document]
duncan-at-hiveit: It's the Location that can be null, yeah.  If someone selects the "All locations" option.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockMigrationService.cs
```diff
@@ -1,40 +1,438 @@
...
+                .Single(boundary =>
+                {
+                    var (level, start, end, _) = boundary;
+
+                    return level == geographicLevel &&
+                           dataBlockCreated.CompareTo(start) >= 0 &&
+                           dataBlockCreated.CompareTo(end) <= 0;
+                })
+                .boundaryLevel;
+        }
```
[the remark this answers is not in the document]
duncan-at-hiveit: Yeah true, would've been simpler in hindsight!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/IDataBlockMigrationService.cs
```diff
@@ -1,15 +1,16 @@
...
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using GovUk.Education.ExploreEducationStatistics.Common.Model;
 using Microsoft.AspNetCore.Mvc;
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces
 {
     /**
-     * Interface used to migrate data blocks. Intended to be adapted as needed when migrations are required.  
+     * Temporary interface used to migrate data blocks for EES-3167 and EES-3142.
```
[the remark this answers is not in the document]
duncan-at-hiveit: I reckon so yeah, unless you've a strong urge to keep it?

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ResultSubjectMetaService.cs
duncan-at-hiveit: Yup, checked it out on the dev branch and we are indeed currently returning GeoJSON when viewing a data block table, if that data block has a Map configured.

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/Charts.cs
duncan-at-hiveit: Cool cheers, I have included that in the PR notes above.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/IDataBlockMigrationService.cs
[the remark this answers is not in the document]
benoutram: No, getting rid of these after this migration is fine by me. I was wondering if we might try and keep some kind of framework for migrating data blocks but as we've seen each migration is so different that a lot has to change anyway.

benoutram (approved):

PR #3261, opened 2022-04-19
EES-3256 - update legacy release UI tests

luke-h1:
This PR: 
- updates the legacy release UI test suite in order to assert that changes made to legacy releases are instantly shown on the public frontend (if the publication already has live releases)

image

### DIFF
--- a/tests/robot-tests/tests/admin_and_public/bau/legacy_releases.robot
+++ b/tests/robot-tests/tests/admin_and_public/bau/legacy_releases.robot
@@ -9,12 +9,17 @@ Suite Teardown      user closes the browser
 Test Setup          fail test fast if required
 
 *** Variables ***
-${PUBLICATION_NAME}     UI tests - legacy releases %{RUN_IDENTIFIER}
+${PUBLICATION_NAME}         UI tests - legacy releases %{RUN_IDENTIFIER}
+${DESCRIPTION}              legacy release description
+${UPDATED_DESCRIPTION}      updated legacy release description
 
 *** Test Cases ***
 Create new publication for topic
-    environment variable should be set    RUN_IDENTIFIER
-    user creates test publication via api    ${PUBLICATION_NAME}
+    ${PUBLICATION_ID}=    user creates test publication via api    ${PUBLICATION_NAME}
+    Set Suite Variable    ${PUBLICATION_ID}
+
+Create new release
+    user create test release via api    ${PUBLICATION_ID}    AY    2020
 
 Verify new publication
     user selects theme and topic from admin dashboard    %{TEST_THEME_NAME}    %{TEST_TOPIC_NAME}
@@ -33,7 +38,7 @@ Create legacy release
     user clicks button    OK
     user waits until modal is not visible    Create legacy release
     user waits until h1 is visible    Create legacy release
-    user enters text into element    id:legacyReleaseForm-description    Test collection
+    user enters text into element    id:legacyReleaseForm-description    ${DESCRIPTION}
     user enters text into element    id:legacyReleaseForm-url    http://test.com
     user clicks button    Save legacy release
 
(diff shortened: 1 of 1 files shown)


tests/robot-tests/tests/admin/bau/legacy_releases.robot
```diff
@@ -33,36 +33,87 @@ Create legacy release
...
     user checks element count is x    css:tbody tr    1
-    # Changing order to 2 should not actually change it to 2
-    # as there isn't another release to swap the order with.
     user checks results table cell contains    1    1    1
-    user checks results table cell contains    1    2    Test collection 2
-    user checks results table cell contains    1    3    http://test-2.com
+    user checks results table cell contains    1    2    updated description
+    user checks results table cell contains    1    3    http://test2.com
+
+Validate public frontend shows changes made to legacy release after saving publication
```
ntsim, line 97: Should this suite be in `admin_and_public` now that it's testing the public frontend as well?

markhiveit (changes requested):

tests/robot-tests/tests/admin/bau/legacy_releases.robot
```diff
@@ -33,36 +33,87 @@ Create legacy release
...
+    user navigates to public frontend    ${PUBLIC_RELEASE_LINK}
+
+    user clicks element    testid:Expand Details Section See other releases (1)
+
+    user checks page contains element    //*[@data-testid="other-release-item"]//a[text()="updated description"]
+
+    user checks element attribute value should be
+    ...    //*[@data-testid="other-release-item"]//a[text()="updated description"]    href    http://test2.com/
+
+Navigate to admin dashboard to update legacy release
```
markhiveit, line 111: This test case name appears twice - I'd just add ` again` on the end of this to avoid confusion when investigating failures

markhiveit (changes requested):

tests/robot-tests/tests/admin_and_public/bau/legacy_releases.robot
```diff
@@ -33,36 +38,86 @@ Create legacy release
...
+    user navigates to public frontend    ${PUBLIC_RELEASE_LINK}
+
+    user clicks element    testid:Expand Details Section See other releases (1)
+
+    user checks page contains element    //*[@data-testid="other-release-item"]//a[text()="${DESCRIPTION}"]
+
+    user checks element attribute value should be
+    ...    //*[@data-testid="other-release-item"]//a[text()="${DESCRIPTION}"]    href    http://test.com/
+
+Navigate to admin dashboard to update legacy release
```
ntsim, line 77: Seems like `Navigate to publication to update legacy releases` would be more appropriate

tests/robot-tests/tests/admin_and_public/bau/legacy_releases.robot
```diff
@@ -33,36 +38,86 @@ Create legacy release
...
     user checks element count is x    css:tbody tr    1
     user checks results table cell contains    1    1    1
-    user checks results table cell contains    1    2    Test collection
+    user checks results table cell contains    1    2    ${DESCRIPTION}
     user checks results table cell contains    1    3    http://test.com
 
+Navigate to admin dashboard to create new release
+    user navigates to admin dashboard    Bau1
+
+Navigate to release
```
ntsim, line 55: Could just combine these two into `Navigate to release`. Also, the first `Navigate to admin dashboard to create new release` doesn't make sense anyway as we've already created the release.

tests/robot-tests/tests/admin_and_public/bau/legacy_releases.robot
```diff
@@ -33,36 +38,86 @@ Create legacy release
...
+
+Check legacy release appears on public frontend
+    user navigates to public frontend    ${PUBLIC_RELEASE_LINK}
+
+    user clicks element    testid:Expand Details Section See other releases (1)
+
+    user checks page contains element    //*[@data-testid="other-release-item"]//a[text()="${DESCRIPTION}"]
+
+    user checks element attribute value should be
+    ...    //*[@data-testid="other-release-item"]//a[text()="${DESCRIPTION}"]    href    http://test.com/
```
ntsim, line 75: This could do with a bit of cleanup. We probably don't need these ugly xpaths, maybe we could do something like:

```robot
user opens details dropdown  See other releases (1)

${other_releases}=  user gets details content element  See other releases (1)

user checks list has x items  css:ul  1  ${other_releases}

${other_release_1}=  user gets list item element  css:ul  1  ${other_releases}
${other_release_1_link}=  get child element  ${other_release_1}  link:${DESCRIPTION}
user checks element attribute value should be  link:${other_release_link}  href  http://test.com/
```

Note that I've asserted that the list of other releases is the right size.

We should also do this for the other test case further down.

ntsim (changes requested):

tests/robot-tests/tests/admin_and_public/bau/legacy_releases.robot
markhiveit: I don't consider this a huge deal, but I would keep a reference to the admin site here - `Navigate to release` itself is ambiguous as it isn't clear if it's the release on admin or public.

My preference would for test case names to always indicate when the tests are switching from admin to public or vice versa.

ntsim: Could just call it `Navigate to release in admin`?

markhiveit (approved):

PR #3265, opened 2022-04-21
EES-3142 - added additional way to look up Created date for Data Bloc…

duncan-at-hiveit:
This PR is raised in reaction to hitting some data in the Dev environment whereby there's no route from a ContentBlock back to its owning Release via the ReleaseContentBlock join table.

Now we have an alternative route via ReleaseContentSection, and failing that, we fall back to using UtcNow.

We also have a try-catch surrounding the migration process to allow us to continue if any exceptions occur.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockMigrationService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockMigrationService.cs
@@ -185,26 +185,39 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
             }
         }
 
-        private async Task<DateTime> GetDataBlockCreatedDate(DataBlock dataBlock)
+        private async Task<DateTime?> GetDataBlockCreatedDate(DataBlock dataBlock)
         {
-            DateTime createdDate;
-
             if (dataBlock.Created != null)
             {
-                createdDate = dataBlock.Created.Value;
+                return dataBlock.Created.Value;
             }
-            else
+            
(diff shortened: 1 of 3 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockMigrationService.cs
```diff
@@ -185,26 +204,47 @@ private async Task SaveChanges(DataBlock dataBlock, bool dryRun)
...
+                        .Include(rcs => rcs.ContentSection)
+                        .ThenInclude(cs => cs.Content)
+                        .Where(rcs => rcs.ContentSection.Content.Contains(dataBlock))
+                        .FirstOrDefaultAsync())
+                    ?.Release;
+
+                if (owningReleaseByContentSection != null)
+                {
+                    return owningReleaseByContentSection.Created;
+                }
```
benoutram, line 239: I think the name `ReleaseContentBlock` has been a bit misleading as we don't link any other types of `ContentBlock` through it apart from data blocks. So to find the Release for a datablock _if_ it's been linked (plenty of data blocks aren't - many are created just to provide table highlights or remain unused), we'd lookup the Release through `ReleaseContentBlock`. Is it necessary to look for them in `ReleaseContentSection` as a fallback? I guess no harm in doing so but maybe we should log this to find out why a data block is in a ContentSection without a link in ReleaseContentBlock. If it ever happened nobody would be able to load the table fast track link.

benoutram (approved):

PR #3266, opened 2022-04-22
EES-3143 Update permalink error messages

markhiveit:
This PR adds a Status field to permalinks, so the frontend can display custom warnings if a permalink is invalid.

We add `PermalinkStatus`es for the following:
- If the permalink is current - i.e. doesn't fall under any of the below statuses.
- If the subject has been removed -- I don't believe this is possible through the service itself. This error would only display if we manually removed a subject from the database.
- If a new release has been published for the permalink's publication - i.e. the permalink's subject was for a 2000 release and there is now a 2001 release.
- If the permalink's subject has been replaced in a release amendment. NOTE: This message also displays if the permalink's subject is removed in an amendment.


src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
```diff
@@ -110,44 +106,62 @@ private async Task<PermalinkViewModel> BuildViewModel(Permalink permalink)
...
 
-        private async Task<bool> IsValid(Guid subjectId)
+        private async Task<PermalinkStatus> GetPermalinkStatus(Guid subjectId)
         {
             var subject = await _subjectRepository.Get(subjectId);
             if (subject == null)
             {
-                return false;
+                return PermalinkStatus.SubjectRemoved;
             }
```
benoutram, line 120: This is using `_subjectRepository` which is accessing the public stats db.

The description of this PR suggests that this status is for the following case:

> If the subject has been removed -- I don't believe this is possible through the service itself. This error would only display if we manually removed a subject from the database.

I think there's four possibilities of how a subject will end up being removed:

* Analysts delete a Subject from the first version of a Release that's not published yet. Removing data file soft deletes subject from private stats db, hard deleted overnight. Not published = no permalinks = we don't care.
* Analysts delete a Subject while amending a Release.  Subject remains in private stats db, linked to previous version(s) of Release through ReleaseSubject. In the public stats db the previous Release and all  its ReleaseSubjects are removed and the Subject is soft deleted, hard deleted overnight.
* Analysts can delete a Subject by amending the Release, replacing the data, publishing. Replacement is a new Subject. Same as previous step - Replaced subject remains in private stats db, linked to previous version(s) of Release through ReleaseSubject but in the public stats db all  the ReleaseSubjects for the replaced Subject are soft deleted and hard deleted overnight.
* Subjects have historically been deleted, possibly before we had amendments. Review of Permalinks for SOW8 found 61 Permalinks using 6 distinct Subjects that aren't in either Public or Private Stats db's.

A review of the Permalinks for SOW8 found at the time that 13% of the Permalinks don't have a Subject in the Public stats db. The majority will be due to points 2 and 3 which will grow over time, with a small number being due to point 4.

So with this first step checking the public stats db for the presence of the subject, you aren't going to be able to differentiate between subjects removed and subjects replaced, as you will never reach the block for determining replacements.

I think you might have to rethink this before we review further, and I think that the comment we put on this ticket about needing a generic approach for deleted vs. replaced might still be all you can do, but please try and prove this as I might be wrong! 🙂 

Here's my thoughts:

* If you can identify the release from the subject (through subject id on the content db's ReleaseFiles?) and the publication is superseded or has a newer release in the series then it's not for the latest data.

* If you can identify the subject is not linked to the latest version of the release (also through the subject id on the content db's ReleaseFiles then it's was either replaced or removed.

* Subject not on any content db ReleaseFiles? Might be able to say it's been removed in this case but would probably go with the generic message.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
```diff
@@ -110,44 +106,62 @@ private async Task<PermalinkViewModel> BuildViewModel(Permalink permalink)
...
+                .Where(rf =>
+                    rf.File.SubjectId == subjectId
+                    && rf.File.Type == FileType.Data
+                    && rf.Release.Published.HasValue && DateTime.UtcNow >= rf.Release.Published.Value)
+                .Select(rf => rf.Release)
+                .ToList();
+            if (releasesWithSubject.Count == 0)
             {
-                return false;
+                throw new Exception("The subject should be attached to at least one live release");
```
benoutram, line 135: This might change depending on whether you agree with my other comment and have to rework this or not, but I don't think we should be throwing any exceptions here. We have a Permalink and regardless of the state of any of the data in our database we should find a way of showing that table.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
markhiveit: So I've changed `PermalinkService#GetPermalinkStatus` so now it only relies on the content DB. This means that we avoid a lot of the complexity around the differences between what is on the public statistics DB vs. the private statistics DB and when the database consolidation work happens, this code won't have to be updated.

I investigated a little to try and find a way we could distinguish between a removed subject and a replaced subject using just the content DB - it may be possible by comparing all release files to see if two match in such a way that would indicate a replacement, but even if that were possible (and I'm doubtful) the extra complexity in code and performance cost probably doesn't warrant it.

But I will emphasize that I only investigated this a little. So, as I'm now off on holiday, someone else might find a superior way.

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Services/PermalinkServiceTests.cs
```diff
@@ -317,71 +289,47 @@ public async Task Get()
...
-                MockUtils.VerifyAllMocks(
-                    blobStorageService,
-                    subjectRepository,
-                    releaseRepository);
+                MockUtils.VerifyAllMocks(blobStorageService);
 
                 Assert.Equal(permalink.Id, result.Id);
                 Assert.InRange(DateTime.UtcNow.Subtract(result.Created).Milliseconds, 0, 1500);
-                Assert.False(result.Invalidated);
+                Assert.Equal(PermalinkStatus.Current, result.Status);
```
benoutram, line 332: I think it would be worth either amending this test or adding another variation which checks the status is Current when the subject is linked through ReleaseFiles to multiple versions of the release.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
```diff
@@ -110,44 +106,56 @@ private async Task<PermalinkViewModel> BuildViewModel(Permalink permalink)
...
+                .Include(rf => rf.File)
+                .Include(rf => rf.Release)
+                .ThenInclude(r => r.Publication)
+                .ThenInclude(p => p.Releases)
+                .Where(rf =>
+                    rf.File.SubjectId == subjectId
+                    && rf.File.Type == FileType.Data
+                    && rf.Release.Published.HasValue && DateTime.UtcNow >= rf.Release.Published.Value)
+                .Select(rf => rf.Release)
+                .ToList();
```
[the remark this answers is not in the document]
benoutram: One other point, might be worth checking if there are duplicate releases in this list and doing a distinct before it's returned.

benoutram (approved):

PR #3268, opened 2022-04-26
EES-3149 Find Stats page should show publications with live releases …

markhiveit:
The Find Statistics page should display publications that have live releases without any data.

Also added more unit tests to ensure something like this doesn't get missed in the future.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
@@ -451,7 +451,6 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Services.Tests
                     Type = ReleaseType.OfficialStatistics,
                     Published = new DateTime(2020, 1, 1),
                 },
-                File = new File { Type = FileType.Data },
             };
 
             var releaseD = new Release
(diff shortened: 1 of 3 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
```diff
@@ -1985,6 +2152,149 @@ await using (var context = InMemoryContentDbContext(contextId))
...
+                Assert.Equal(publicationA.Id, publication.Id);
+                Assert.False(publication.IsSuperseded);
+                Assert.True(publication.HasLiveRelease);
+                Assert.True(publication.LatestReleaseHasData);
+                Assert.True(publication.AnyLiveReleaseHasData);
+            }
+        }
+
+        [Fact]
+        public async Task GetPublicationTree_DataTables_PreviousReleaseHasData()
```
ntsim, line 2229: Is this already covered by `GetPublicationTree_DataTables_DoesFilterPublicationWithoutDataOnLatestRelease`? We might want to adjust the name to make it more readable/obvious.

These new assertions should probably be in there instead.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
```diff
@@ -2879,6 +3189,158 @@ await using (var context = InMemoryContentDbContext(contextId))
             }
         }
 
+        [Fact]
+        public async Task GetPublicationTree_DataCatalogue_ReleaseHasData()
```
ntsim, line 3193: Seems like this is already covered by: `GetPublicationTree_DataCatalogue_LatestData`. We might want to adjust the name to make it more readable/obvious.

These new assertions should probably be in there instead.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
```diff
@@ -1985,6 +2152,149 @@ await using (var context = InMemoryContentDbContext(contextId))
             }
         }
 
+        [Fact]
+        public async Task GetPublicationTree_DataTables_ReleaseHasData()
```
ntsim, line 2156: Is this already covered by `GetPublicationTree_DataTables_LatestDataNotSuperseded`? We might want to adjust the name to make it more readable/obvious.

These new assertions should probably be in there instead.

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
markhiveit: I think this is already covered - I've removed this test.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ThemeService.cs
```diff
@@ -164,8 +165,8 @@ private async Task<PublicationTreeNode> BuildPublicationNode(Publication publica
             {
                 case PublicationTreeFilter.FindStatistics:
                     return !publicationTreeNode.IsSuperseded
-                           && (publicationTreeNode.LatestReleaseHasData ||
-                               !string.IsNullOrEmpty(publicationTreeNode.LegacyPublicationUrl));
+                           && (publicationTreeNode.HasLiveRelease
```
benoutram, line 168: Could this be simplified to `&& (publicationTreeNode.HasLiveRelease || publicationTreeNode.Type == Legacy)`?

markhiveit: I've changed it. I've also changed a similar bit of code in `ThemeService#BuildTopicTree`.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
```diff
@@ -1818,7 +1984,7 @@ await using (var context = InMemoryContentDbContext(contextId))
         }
 
         [Fact]
-        public async Task GetPublicationTree_DataTables_DoesFilterPublicationWithoutDataOnLatestRelease()
+        public async Task GetPublicationTree_DataTables_DoesNotFilterPublicationWithDataOnPreviousRelease()
```
ntsim, line 1987: Just to be consistent with some of these other test cases (and for brevity), maybe we could call this `GetPublicationTree_DataTables_PreviousReleaseHasData`

ntsim (approved):

PR #3269, opened 2022-04-26
EES-3019 Remove full table title button from table caption

ntsim:
This PR removes the 'View/hide full table title' button that was previously visible beneath table captions that were using auto-generated titles.

We've removed this button as the full table title isn't particularly useful to users, especially when they have selected many options. For screen readers this is even worse, as the generated title can take a very long time to read out in its entirety. We now just always use the shortened title which takes the following form when locations/filters exceed 5 selections:

> '{subject}' for {filters} and x other filters in {locations} and x other locations between {start} and {end}

As part of this:

- We have reduced the shortening/truncation limit from 10 locations to 5 locations
- Added the truncation logic to filters as well. We would previously show _all_ filters, which could lead to excessively long titles regardless of the number of locations.

## Relevant changes

- Moved `generateTableTitle` function into its own file so that other components don't have to import the `DataTableCaption` component to access that function.


src/explore-education-statistics-common/src/modules/table-tool/utils/__tests__/generateTableTitle.test.ts
```diff
@@ -81,20 +81,30 @@ describe('generateTableTitle', () => {
...
 
-    expect(result).toBe(
+    expect(title).toBe(
+      "Authorised absence rate for 'Absence by characteristic' in England for 2015/16",
+    );
+  });
+
+  test('with no filters', () => {
+    const title = generateTableTitle({
+      ...testFullTableMeta,
```
duncan-at-hiveit, line 91: Could rename this to `testMeta` as you did in another test file in this PR, but I'm not overly fussed :)

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/DataTableCaption.test.tsx
```diff
@@ -67,163 +66,22 @@ describe('DataTableCaption', () => {
...
+    render(<DataTableCaption meta={testMeta} />);
 
-    expect(screen.getByTestId('dataTableCaption')).toMatchInlineSnapshot(`
-      <strong
-        data-testid="dataTableCaption"
-      >
-        Authorised absence rate for 'Absence by characteristic' for Female and Male in Barnet and Barnsley for 2015/16
-      </strong>
-    `);
+    expect(screen.getByTestId('dataTableCaption')).toHaveTextContent(
```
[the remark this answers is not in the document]
ntsim: Nope, these were using `toMatchInlineSnapshot` which doesn't generate snapshot files - it literally dumps the snapshot into the test case itself.

src/explore-education-statistics-common/src/modules/table-tool/utils/__tests__/generateTableTitle.test.ts
ntsim: Done!

PR #3270, opened 2022-04-26
EES-1238 Allow ordering filters and indicators

benoutram:
This PR adds the back-end support to allow ordering filters, filter groups, filter items, indicator groups and indicators with the intention that the front end will then order them in respect to their order in API responses.

It allows analysts to change the ordering between versions of a Release for the same subjects by storing the sequences on `ReleaseSubject` as two new configuration attrbitures `FilterSequence` and `IndicatorSequence` rather than on new `Order` attributes added directly to the `Filter`, `FilterGroup` `FilterItem`, `IndicatorGroup` and `Indicator` types.

The latter approach would not be versionable. New `ReleaseSubjects` are created with each new version of a release which is the reason why this solution was chosen. It also doesn't need to be concerned with ensuring the sequence of any `Order` attributes remains consistent which could have been challenging when it came to data replacements.

Note: Future work in EES-2793 will cover the updates required when subjects are replaced.

To keep things simple there's no migration of existing subject data, and no configuration is created when importing new subjects.

If the order is yet to be defined by analysts then there's a fallback which matches existing ordering behaviour, i.e. alphabetical by label. For filter groups and filter items this fallback behaviour retains a special case where "Total" is ordered as the first entry if it exists.

Now that analysts have the ability to change the order they are free to choose where they position the "Total" items in future.

To update the order of filters including filter groups and their filter items:
`PATCH /api/data/release/{releaseId}/meta/subject/{subjectId}/filters`

To update the order of indicator groups and their indicators:
`PATCH /api/data/release/{releaseId}/meta/subject/{subjectId}/indicators`

The desired sequence is inferred from the json array order and nested orders in the request.

This PR makes the front-end changes to integrate the new "Reorder filters and indicators" tab with these endpoints.

Responsibility for building filter and indicator view models for subject and table result subject meta data with is shifted away from `SubjectMetaService`, `ResultSubjectMetaService`, and the abstract class `AbstractSubjectMetaService`, to a dedicated `FilterAndIndicatorViewModelBuilders` class. This allows the services and their reliance on the abstract class to be simplified and the logic for building the view models as well as identifying the "Total"'s and performing the ordering to be unit tested independently of them.

### Other changes
* Corrects nullability of the `Name` and `Label` database columns for the Filter, FilterGroup, FilterItem, IndicatorGroup and Indicator tables.


src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/Meta/FilterMetaViewModels.cs
```diff
@@ -0,0 +1,89 @@
...
+            Value = filterItem.Id;
+        }
+
+        public FilterItemMetaViewModel(string label, Guid value)
+        {
+            Label = label;
+            Value = value;
+        }
+
+        public FilterItemMetaViewModel()
```
duncan-at-hiveit, line 71: Are these default constructors needed for some purpose?  JSON-related maybe?

src/explore-education-statistics-admin/src/pages/release/datablocks/__data__/tableToolServiceData.ts
```diff
@@ -6,25 +6,30 @@ import {
...
         Gender: {
+          id: 'gender',
           label: 'Gender',
           options: [
             {
               label: 'Gender female',
               value: 'gender-female',
             },
           ],
+          order: 0,
```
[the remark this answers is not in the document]
benoutram: In the previous front end work to add the reordering tab to the Admin we added an order field to the types that are objects in the API responses for subject meta data and table result subject meta data which is filters, filter groups and indicator groups. I think these corresponded with the 'legend options' view model types on the back end previously. The options (which will be the filter items or indicators) are arrays so we can infer the order from the array position. For the others I did this because I thought with them being key/values of an object that their position shouldn't be relied on for an order. A lot of the changes on this PR will be because I now made the order field mandatory whereas it was optional before.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/Meta/FilterMetaViewModels.cs
benoutram: This particular one is required for deserialization of JSON data that's set up in a test of `TableBuilderControllerTests`.

The other constructors in `FilterMetaViewModel` and `FilterGroupMetaViewModel` are needed for setting up test data in `SubjectMetaServiceTests` in the way that we are initialising them using the object initialisation constructor.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/ReleaseSubject.cs
```diff
@@ -23,5 +32,31 @@ public ReleaseSubject CopyForRelease(Release release)
...
+        public DateTime? Updated { get; set; }
+    }
+
+    public interface ISequenceEntry<TChildSequenceEntry>
+    {
+        Guid Id { get; init; }
+        List<TChildSequenceEntry> ChildSequence { get; init; }
+    }
+
+    public class FilterSequenceEntry : ISequenceEntry<FilterGroupSequenceEntry>
```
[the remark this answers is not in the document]
benoutram: Yes, that seems to be fine. Altered on a new commit.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
```diff
@@ -32,91 +36,168 @@ private enum SubjectMetaQueryStep
...
             _observationService = observationService;
-            _persistenceHelper = persistenceHelper;
+            _releaseSubjectRepository = releaseSubjectRepository;
             _timePeriodService = timePeriodService;
-            _userService = userService;
             _locationOptions = locationOptions.Value;
         }
 
-        public Task<Either<ActionResult, SubjectMetaViewModel>> GetSubjectMeta(Guid subjectId)
+        public async Task<Either<ActionResult, SubjectMetaViewModel>> GetSubjectMeta(Guid subjectId)
```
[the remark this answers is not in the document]
benoutram: Thanks, gone with `GetCachedSubjectMeta`.

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/SubjectMetaServiceTests.cs
```diff
@@ -42,33 +48,42 @@ public class SubjectMetaServiceTests
         private readonly LocalAuthority _sunderland = new("E08000024", "", "Sunderland");
 
         [Fact]
-        public async Task GetSubjectMeta_SubjectNotFound()
+        public async Task GetCachedSubjectMeta_SubjectNotFound()
```
ntsim, line 51: Are we going with the convention of prefixing cached methods with 'Cached'? This conversation has also cropped up in Duncan's PR: https://github.com/dfe-analytical-services/explore-education-statistics/pull/3277#discussion_r861799083

Personally, I'm not a big fan for the aforementioned reasons in that discussion, but perhaps we need to discuss this a little further.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/AbstractSubjectMetaService.cs
```diff
@@ -124,9 +79,25 @@ protected static List<IndicatorMetaViewModel> BuildIndicatorViewModels(IEnumerab
...
+                    .Include(rs => rs.Subject)
+                    .Where(rs => rs.ReleaseId == releaseId
+                                 && rs.SubjectId == subjectId)
+            );
+        }
+
+        protected async Task<Either<ActionResult, ReleaseSubject>> CheckCanViewSubjectData(
+            ReleaseSubject releaseSubject)
+        {
+            if (await _userService.MatchesPolicy(releaseSubject.Subject, DataSecurityPolicies.CanViewSubjectData))
```
ntsim, line 95: Would it be a good idea to put this on a `UserServiceExtensions` method like we do for the content and admin security policies?

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
```diff
@@ -32,91 +36,168 @@ private enum SubjectMetaQueryStep
...
+        {
+            return await CheckReleaseSubjectExists(releaseId, subjectId)
+                .OnSuccessVoid(async rs =>
+                {
+                    // TODO EES-3345 Validate all entries are present and only entries that relate to the subject
+
+                    // Set the sequence based on the order of filters, filter groups and indicators observed
+                    // in the request
+                    rs.FilterSequence = request.Select(filter =>
+                            new FilterSequenceEntry
```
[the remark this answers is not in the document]
benoutram: Yes, since I'm only storing the Id's I was trying to be a bit more explicit about the types, to still make it clearer that Filter has child FilterGroup etc.

I could have done something like this

`SequenceEntry<Guid, SequenceEntry<Guid, Guid>>` for a filter id with child filter group id's that in turn have child filter item id's, and

`SequenceEntry<Guid, Guid>` for an indicator group id that has child indicator id's.

What I've ended up going with after changing to record types is:

```
    public abstract record SequenceEntry<TEntry, TChild>(TEntry Id, List<TChild> ChildSequence);

    public record FilterSequenceEntry(Guid Id, List<FilterGroupSequenceEntry> ChildSequence) :
        SequenceEntry<Guid, FilterGroupSequenceEntry>(Id, ChildSequence);

    public record FilterGroupSequenceEntry(Guid Id, List<Guid> ChildSequence) :
        SequenceEntry<Guid, Guid>(Id, ChildSequence);

    public record IndicatorGroupSequenceEntry(Guid Id, List<Guid> ChildSequence) :
        SequenceEntry<Guid, Guid>(Id, ChildSequence);
```

src/GovUk.Education.ExploreEducationStatistics.Data.Services/FilterAndIndicatorViewModelBuilders.cs
```diff
@@ -0,0 +1,296 @@
...
+            IEnumerable<FilterSequenceEntry>? sequence = null)
+        {
+            var filters = GroupFilterItemsByFilter(values);
+            return BuildFilters(filters, sequence);
+        }
+
+        public static Dictionary<string, FilterMetaViewModel> BuildFilters(IEnumerable<Filter> values,
+            IEnumerable<FilterSequenceEntry>? sequence = null)
+        {
+            FilterMetaViewModel BuildViewModel(Ordered<Filter, FilterSequenceEntry> input, int index)
```
ntsim, line 26: Not a biggy, but how come we haven't inlined this into a lambda on `resultSelector`? It's just a little unusual to see a full function declaration like this?

Seems like we've also done this in the other methods as well, so same question applies for those too

src/explore-education-statistics-admin/src/pages/release/data/components/ReleaseDataReorderSection.tsx
```diff
@@ -33,13 +34,27 @@ const ReleaseDataReorderSection = ({ releaseId, canUpdateRelease }: Props) => {
   const [reorderingFilters, setReorderingFilters] = useState<Subject>();
   const [reorderingIndicators, setReorderingIndicators] = useState<Subject>();
 
-  const handleSaveFilters = (updatedFilters: UpdateFiltersRequest) => {
+  const handleSaveFilters = (
```
ntsim, line 37: For async/await stuff, we should write our handlers accordingly e.g.
```ts
const handleSaveFilters = async (...) => {
  await adminTableBuilderService.updateFilters(...);
}
```

Similar applies for `handleSaveIndicators` below as well

src/explore-education-statistics-admin/src/pages/release/data/components/ReleaseDataReorderSection.tsx
```diff
@@ -33,13 +34,27 @@ const ReleaseDataReorderSection = ({ releaseId, canUpdateRelease }: Props) => {
...
     setReorderingFilters(undefined);
-    // TO DO: send the update request.
+    adminTableBuilderService.updateFilters(
+      releaseId,
+      subjectId,
+      updatedFilters,
+    );
   };
-  const handleSaveIndicators = (updatedIndicators: UpdateIndicatorsRequest) => {
+  const handleSaveIndicators = (
```
ntsim, line 48: Would add a new line between event handler declarations

src/explore-education-statistics-admin/src/pages/release/data/components/ReleaseDataReorderSection.tsx
```diff
@@ -33,13 +34,27 @@ const ReleaseDataReorderSection = ({ releaseId, canUpdateRelease }: Props) => {
...
   const [reorderingIndicators, setReorderingIndicators] = useState<Subject>();
 
-  const handleSaveFilters = (updatedFilters: UpdateFiltersRequest) => {
+  const handleSaveFilters = (
+    subjectId: string,
+    updatedFilters: UpdateFiltersRequest,
+  ) => {
     setReorderingFilters(undefined);
-    // TO DO: send the update request.
+    adminTableBuilderService.updateFilters(
```
ntsim, line 42: Wouldn't it make more sense to toggle `setReorderingFilters` _after_ the service request has completed?

```ts
await adminTableBuilderService.updateFilters(...);
setReorderingFilters(undefined);
```

Similar applies for `handleSaveIndicators` below as well

src/GovUk.Education.ExploreEducationStatistics.Data.Services/FilterAndIndicatorViewModelBuilders.cs
```diff
@@ -0,0 +1,296 @@
...
+            ? OrderByLabel<TValue, TSequence>(values, labelSelector)
+            : OrderBySequence(values, idSelector, sequenceIdSelector, sequence)).ToList();
+
+        var result = new Dictionary<string, TResult>(ordered.Count);
+        ordered.ForEach((ordered1, index) =>
+        {
+            var key = labelSelector(ordered1).PascalCase();
+            var value = resultSelector(ordered1, index);
+            result.Add(key, value);
+        });
```
ntsim, line 224: Perhaps it might be convenient to create a `GroupByIndexed` enumerable extension method instead of manually creating this? Same applies for the grouping in `OrderAsDictionaryWithTotalFirst` below.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/AbstractSubjectMetaService.cs
benoutram: Ah yeah, good idea, this makes things one step closer to getting rid of this abstract class. Thanks.

src/explore-education-statistics-admin/src/pages/release/data/components/ReleaseDataReorderSection.tsx
benoutram: Thanks have made this change.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/FilterAndIndicatorViewModelBuilders.cs
benoutram: Thanks that does simplify things a bit. I've made this change now.

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/SubjectMetaServiceTests.cs
benoutram: Thanks, I'm not sure. I made the change to prefix the name with 'Cached' after Duncan chatted with me about it and left a comment (see above).

I think we do need to discuss this further as I see this cropping up on 3 PR's now! 🙂

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/ReleaseSubjectRepository.cs
```diff
@@ -48,6 +48,26 @@ public async Task DeleteAllReleaseSubjects(Guid releaseId, bool softDeleteOrphan
...
+                .Where(releaseSubject => releaseSubject.SubjectId == subjectId)
+                .ToListAsync();
+
+            var publishedReleaseSubjects = releaseSubjects
+                .Where(releaseSubject => releaseSubject.Release.Live)
+                .ToList();
+
+            var previousReleaseVersionIds = publishedReleaseSubjects
+                .Select(releaseSubject => releaseSubject.Release.PreviousVersionId)
+                .Where(previousVersionId => previousVersionId.HasValue);
```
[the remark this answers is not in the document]
benoutram: They could be merged like you say but I want to use `publishedReleaseSubjects ` in a third statement below and I'm not sure I can achieve merging all 3 while returning a `ReleaseSubject`, or at least not in a very readable way.

This is the code including the third statement:

```
            var publishedReleaseSubjects = releaseSubjects
                .Where(releaseSubject => releaseSubject.Release.Live)
                .ToList();

            var previousReleaseVersionIds = publishedReleaseSubjects
                .Select(releaseSubject => releaseSubject.Release.PreviousVersionId)
                .Where(previousVersionId => previousVersionId.HasValue);

            return publishedReleaseSubjects.SingleOrDefault(
                releaseSubject => !previousReleaseVersionIds.Contains(releaseSubject.ReleaseId));
```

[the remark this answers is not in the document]
ntsim: Ah sorry, I see! Probably should have read the code a little closer :man_facepalming:

src/GovUk.Education.ExploreEducationStatistics.Data.Services/FilterAndIndicatorViewModelBuilders.cs
benoutram: Update: just renamed this to `ToDictionaryIndexed` as it doesn't actually do a GroupBy. It works similar to the built in `ToDictionary` in the sense that if multiple values evaluate to the same key it will throw an error, rather than grouping the values into a list.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
```diff
@@ -32,91 +34,128 @@ private enum SubjectMetaQueryStep
...
             _logger = logger;
             _observationService = observationService;
-            _persistenceHelper = persistenceHelper;
             _timePeriodService = timePeriodService;
             _userService = userService;
             _locationOptions = locationOptions.Value;
         }
 
-        public Task<Either<ActionResult, SubjectMetaViewModel>> GetSubjectMeta(Guid subjectId)
+        public async Task<Either<ActionResult, SubjectMetaViewModel>> GetSubjectMeta(ReleaseSubject releaseSubject)
```
ntsim, line 74: Not sure if I'm missing something here but how come we've gone with passing in `ReleaseSubject` as a parameter? Kinda would have expected still expected a `subjectId` and optional `releaseId` so that we could fetch the `ReleaseSubject` within the service.

Ideally it'd be good to avoid passing around entities as much as possible. There's obviously caveats around whether relationships have been loaded beforehand or not. It'd be nice to keep as much of the release subject fetching in this service too.

If this mostly due to the cache key stuff, perhaps it'd be good to have overloads for `ReleaseSubject` and `releaseId` / `subjectId`.

Same applies for `FilterSubjectMeta`.

benoutram: Yes, some requests don't have a `releaseId` so we need to lookup the `ReleaseSubject` for the latest version of the release which could all be done in the service if it wasn't for caching.

Prior to your comment I made the `GetSubjectMeta` and `FilterSubjectMeta` methods in the service take `ReleaseSubject` for consistency and updated everything across both the Admin and Data.API, thinking this would be more readable and have less variations of tests plus permissions tests.

I've applied your suggestion now and altered the Admin controller to always use `releaseId` and `subjectId` for both `GetSubjectMeta` and `FilterSubjectMeta`. In the Admin `releaseId` is always known. This tidies things up nicely for that controller.

In the Data.API it now always calls the service with `releaseId?` and `subjectId` for `FilterSubjectMeta` but still uses a `releaseSubject` overload for `GetFilterSubjectMeta` due to the caching.

For the caching, in order to know the release slug we can use `ReleaseSubject.Release.Slug` but because we also need to know the publication slug we lookup the Content release including publication and then we have both. Either way we need to have identified the release from `subjectId`.

We could potentially do away with wrapping the `ReleaseSubject` in the cacheable entity as long as it still held the required parameters for caching, i.e. both slugs and the subject id. However as the ReleaseSubject has already been fetched I've decided to continue using the overload.

Just as an aside, for the case where we don't have the `releaseId` I've been contemplating changing it so that rather than looking up the latest ReleaseSubject using `subjectId` with `_releaseSubjectRepository.GetReleaseSubjectForLatestPublishedVersion`, I could lookup the publication from any ReleaseSubject using the `subjectId` along with looking up the latest release in the series for that publication. Then call the service with `subjectId` and the latest release in the series' `releaseId`.

Maybe this is the more 'correct' thing to do as it would stop someone fetching the metadata without a release id for a subject of a release that's no longer the latest release in the series.

At the moment you would be able to query without release id for a subject  of a release that's no longer the latest release in the series. 

It must always be the latest version of that release however, and that's enforced with a permission check on the service, so I've not made this change for now.

benoutram:
image

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
ntsim: Fair enough, sounds reasonable.

As to the final point about how we fetching the latest published release, I'm not sure I understand the following:

> Maybe this is the more 'correct' thing to do as it would stop someone fetching the metadata without a release id for a subject of a release that's no longer the latest release in the series.

How would someone be able to get the meta for the non-latest release without a `releaseId`? I would have assumed `_releaseSubjectRepository.GetReleaseSubjectForLatestPublishedVersion` should already prevent that?

ntsim (approved):

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
benoutram: Short answer: It's allowing them to get it for the latest version of all published releases, which I think is ok? 🙂 

Long answer:
`_releaseSubjectRepository.GetReleaseSubjectForLatestPublishedVersion ` works by assuming a subject can be linked to more than one release through `ReleaseSubject` but working out which is the most recent release version by dismissing any older versions that are linked through `previousVersionId` and dismissing any where the release is not `live`. This ends up with either one result or no result.

It's not aware of content releases and that there may be a newer published release _in the series_ for a later time period.  I put 'version' in the name of the method to try and make it a bit clearer what it's returning.

Getting the meta data either with a specific `releaseId` from a previous _version_ that's been replaced, or by `subjectId` alone where the subject has been removed from the latest _version_ will both be forbidden by the permissions check on `DataSecurityPolicies.CanViewSubjectData` which checks the content release is the latest published _version_. I think this is the important bit, i.e. that we don't allow returning meta data where the specified release id, or the releasesubject that's been found aren't on a published release.

Getting the meta data either with a specific `releaseId` from the latest version or by `subjectId` alone where the subject is on the latest version, but in both cases where that version is not the latest release in the series by time period will both be allowed because a ReleaseSubject will be found and the `DataSecurityPolicies.CanViewSubjectData` is happy because that release is published, just not the latest year/time period.

PR #3273, opened 2022-04-27
EES-3136 Invalidate Admin data blocks cache files after footnotes get updated

markhiveit:
This PR fixes a bug where Admin data block cache files wouldn't get invalidated after updating footnotes. This meant that if a data block has been cached and then someone updated footnotes associated with that data block, in Admin Content the data block would display the wrong footnotes.

We now delete all data block cache files associated with a release whenever a footnote is added, updated, or removed. This was for speed and simplicity of implementing it. Also, ascertaining what specific data blocks are affected by a footnote change will probably be an expensive operation, so it isn't clear that doing that is actually preferable.

This PR also adds unit tests for `FootnoteService#DeleteFootnote` and `FootnoteService#UpdateFootnote`, and enables a `Failing` UI test that checks that this PR actually fixes the bug.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/FootnoteServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/FootnoteServiceTests.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services;
+using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Services;
 using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces;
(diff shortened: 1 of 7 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
```diff
@@ -329,6 +331,15 @@ private List<DataBlock> GetDataBlocks(Guid releaseId, Guid? subjectId = null)
                 .OnSuccessVoid();
         }
 
+        public async Task InvalidateCachedDataBlocks(Guid releaseId)
+        {
+            var dataBlocks = GetDataBlocks(releaseId);
+            foreach (var dataBlock in dataBlocks)
+            {
+                 await InvalidateDataBlockCache(releaseId, dataBlock.Id);
```
ntsim, line 339: Would be tempted to rename this to `InvalidateCachedDataBlock` for consistency with our new method

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/FootnoteServiceTests.cs
```diff
@@ -538,11 +544,490 @@ void AssertFootnoteDetailsCopiedCorrectly(Footnote originalFootnote, Footnote ne
...
         }
 
+        [Fact]
+        public async Task DeleteFootnote()
+        {
+            var release = new Release();
+
+            var subject = new Subject
+            {
+                Id = Guid.NewGuid(),
```
ntsim, line 554: Do we need to generate ids for the Subject/Filter/Indicator models?

Think we could remove all of these ids in favour of linking the models directly (would save a bunch of lines). Seems like we have done this for `indicatorGroup`, but not the others?

Would do this for the other test cases further down as well.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/FootnoteServiceTests.cs
```diff
@@ -538,11 +544,490 @@ void AssertFootnoteDetailsCopiedCorrectly(Footnote originalFootnote, Footnote ne
...
+                    new List<Guid>{ subject.Id });
+
+                result.AssertRight();
+            }
+
+            MockUtils.VerifyAllMocks(dataBlockService);
+
+            await using (var statisticsDbContext = InMemoryStatisticsDbContext(contextId))
+            {
+                var dbFootnote = Assert.Single(statisticsDbContext.Footnote.ToList());
```
ntsim, line 827: Not a big deal, but 'saved' could be a better prefix than 'db' e.g. `savedFootnote`. Could do this for the others as well.

ntsim (changes requested):

PR #3277, opened 2022-04-28
Ees 3156 use cached methodologies for get latest release

duncan-at-hiveit:
This PR:
- reuses cached Methodologies from the "All Methodologies" page (the "All Methodologies Theme Tree") to display Methodology Summaries on public Release pages.
- This is to address some heavy database calls that were identified as having been mainly responsible for site performance issues when large traffic volumes were viewing Release pages concurrently.
- 
image


src/GovUk.Education.ExploreEducationStatistics.Content.Api/Controllers/ReleaseController.cs
```diff
@@ -30,7 +30,7 @@ public async Task<ActionResult<List<ReleaseSummaryViewModel>>> ListReleases(stri
         [HttpGet("publications/{publicationSlug}/releases/latest")]
         public async Task<ActionResult<ReleaseViewModel>> GetLatestRelease(string publicationSlug)
         {
-            return await _releaseService.Get(publicationSlug)
+            return await _releaseService.GetCachedViewModel(publicationSlug)
```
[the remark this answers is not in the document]
ntsim: Yeh I'd personally say we shouldn't be prefixing stuff with 'cached'. Caching is an implementation detail and naming methods like this would be quite limiting if we had different implementations of the underlying interfaces.

That said, we're not exactly running an OOP purity class here, so if people are finding it problematic to know whether a method is cached or not, we can go with it.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
```diff
@@ -3,21 +3,47 @@
...
+    [Collection(BlobCacheServiceTests)]
+    public class PublicationServiceTests : BlobCacheServiceTestFixture
     {
+        // TODO - this constructor sets up default expectations for the caching service that are called by each test
+        // method within this test class.  In future it would be good to do targeted testing around caching within
+        // each method under test.  The reason this test class needs to be identified as part of the
+        // "BlobCacheServiceTests" collection is that otherwise these test methods will fail with unmatched CacheService
+        // setups if run in parallel with another test class that is of the "BlobCacheServiceTests" collection.
+        // This would happen because the other test class enables caching and sets a static CacheService mock at the
+        // same time as this class is running.
```
ntsim, line 33: Seems like we're excessively explaining how the `BlobCacheServiceTests` collection stuff works. We've already got this all over the place, so it's a bit weird to put this here. It also seems like a bit of a tangent to the earlier parts of the comment (which is a TODO).

It'd probably be a better idea to put any explanations  next to the `BlobCacheServiceTests` constant in `BlobCacheServiceTestFixture`

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
```diff
@@ -4,18 +4,42 @@
...
+    [Collection(BlobCacheServiceTests)]
+    public class ThemeServiceTests : BlobCacheServiceTestFixture
     {
+        // TODO - this constructor sets up default expectations for the caching service that are called by each test
+        // method within this test class.  In future it would be good to do targeted testing around caching within
+        // each method under test.  The reason this test class needs to be identified as part of the
+        // "BlobCacheServiceTests" collection is that otherwise these test methods will fail with unmatched CacheService
+        // setups if run in parallel with another test class that is of the "BlobCacheServiceTests" collection.
+        // This would happen because the other test class enables caching and sets a static CacheService mock at the
+        // same time as this class is running.
```
ntsim, line 29: Seems like we're excessively explaining how the `BlobCacheServiceTests` collection stuff works. We've already got this all over the place, so it's a bit weird to put this here. It also seems like a bit of a tangent to the earlier parts of the comment (which is a TODO).

It'd probably be a better idea to put any explanations  next to the `BlobCacheServiceTests` constant in `BlobCacheServiceTestFixture`

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/MethodologyServiceCachingTests.cs
```diff
@@ -0,0 +1,646 @@
...
+        
+        var methodologyVersionRepository = new Mock<IMethodologyVersionRepository>(MockBehavior.Strict);
+
+        methodologyVersionRepository
+            .Setup(mock => mock.GetLatestPublishedVersionByPublication(publication.Id))
+            .ReturnsAsync(publicationLatestVersions);
+        
+        methodologyVersionRepository
+            .Setup(mock => mock.GetLatestPublishedVersionByPublication(publication.Id))
+            .ReturnsAsync(publicationLatestVersions);
```
ntsim, line 84: Is this actually being called twice? If so, it probably shouldn't be?

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/MethodologyServiceCachingTests.cs
```diff
@@ -0,0 +1,646 @@
...
+                                    new MethodologyVersionSummaryViewModel
+                                    {
+                                        Id = Guid.NewGuid(),
+                                        Title = "Theme 2 Topic 1 Publication 1 Methodology 1",
+                                        Slug = "Theme 2 Topic 1 Publication 1 Methodology 1 Slug"
+                                    })
+                            })
+                    })
+            }
+        );
```
ntsim, line 195: Do we need all of this test data? Feels like we could simplify this down and save ourselves a bunch of lines.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/MethodologyServiceCachingTests.cs
```diff
@@ -0,0 +1,646 @@
...
+    public async Task GetCachedSummariesByPublication_PublicationNotFound()
+    {
+        await using var contentDbContext = InMemoryContentDbContext();
+        var service = SetupMethodologyService(contentDbContext);
+        var result = await service.GetCachedSummariesByPublication(Guid.NewGuid());
+        result.AssertNotFound();
+    }
+
+    [Fact]
+    public async Task GetCachedSummariesTree_NoCachedMethodologyThemeTreeExists()
```
ntsim, line 302: The test case names for `GetCachedSummariesTree_` are a little on the verbose side, we could shorten these to variants of `GetCachedSummariesTree_NoCachedTreeExists` / `GetCachedSummariesTree_CacheTreeExists`

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
duncan-at-hiveit: Actually I've just removed it!

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/MethodologyServiceCachingTests.cs
duncan-at-hiveit: I kind've wanted to have at least another theme and topic, another publication in the same theme and topic for this test case, just to be thorough and prove that we're just matching the correct data in this method.

Note that in other tests within this same test class, I'm not going overboard with them - but for this "happy path" one I just wanted to prove that only a very particular set of results is coming back.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/MethodologyServiceCachingTests.cs
duncan-at-hiveit: Good shout, done

ntsim (approved):

PR #3282, opened 2022-05-04
EES-3345 Validate all filters and indicators for subject are present on update

benoutram:
This PR builds upon https://github.com/dfe-analytical-services/explore-education-statistics/pull/3270 adding validation to ensure that when filters, filter groups, filter items, indicator groups and indicators are updated that the request contains all elements for the subject and that there are no unexpected elements.

When the update request is processed the sequence of filters and indicators groups is set based on the order in the request along with the sequences of their respective filter groups, filter items, and indicators. It's important to make sure that no elements are missing and that no elements unknown to the subject are included.

### Other changes

* Add async support to `FilterRepository` and `IndicatorGroupRepository`.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FilterRepository.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FilterRepository.cs
@@ -2,24 +2,29 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 using GovUk.Education.ExploreEducationStatistics.Data.Model.Database;
 using GovUk.Education.ExploreEducationStatistics.Data.Model.Repository.Interfaces;
 using Microsoft.EntityFrameworkCore;
 
 namespace GovUk.Education.ExploreEducationStatistics.Data.Model.Repository
 {
-    public class FilterRepository : AbstractRepository<Filter, Guid>, IFilterRepository
+    public class FilterRepository : IFilterRepository
     {
-        public FilterRepository(StatisticsDbContext context) : base(context)
+        private readonly StatisticsDbContext _context;
+
+        public FilterRepository(StatisticsDbContext context)
         {
+            _context = context;
         }
 
-        public List<Filter> GetFiltersIncludingItems(Guid subjectId)
+        public Task<List<Filter>> GetFiltersIncludingItems(Guid subjectId)
         {
-            return FindMany(filter => filter.SubjectId == subjectId)
+            return _context.Filter
                 .Include(filter => filter.FilterGroups)
                 .ThenInclude(group => group.FilterItems)
-                .ToList();
+                .Where(filter => filter.SubjectId == subjectId)
+                .ToListAsync();
         }
     }
 }
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
@@ -347,12 +347,10 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Services
                 "Requested filters do not match subject filters").OnSuccess(_ =>
             {
                 var requestMap = requestFilters.ToDictionary(filter => filter.Id);
-                var results = filters.Select(filter =>
+                return filters.Select(filter =>
                         ValidateFilterGroupsForSubject(filter, requestMap[filter.Id].FilterGroups))
-                    .ToList();
-
-                var failure = results.FirstOrDefault(result => result.IsLeft)?.Left;
-                return failure ?? new Either<ActionResult, Unit>(Unit.Instance);
+                    .OnSuccessAll()
+                    .OnSuccessVoid();
             });
         }
 
@@ -368,15 +366,13 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Services
                 .OnSuccess(_ =>
                 {
                     var requestMap = requestFilterGroups.ToDictionary(filterGroup => filterGroup.Id);
-                    var results = filter.FilterGroups.Select(filterGroup =>
+                    return filter.FilterGroups.Select(filterGroup =>
                             AssertCollectionsAreSameIgnoringOrder(
                                 filterGroup.FilterItems.Select(filterItem => filterItem.Id),
                                 requestMap[filterGroup.Id].FilterItems,
                                 $"Requested filter items do not match subject for filter group: {filterGroup.Id}"))
-                        .ToList();
-
-                    var failure = results.FirstOrDefault(result => result.IsLeft)?.Left;
-                    return failure ?? new Either<ActionResult, Unit>(Unit.Instance);
+                        .OnSuccessAll()
+                        .OnSuccessVoid();
                 });
         }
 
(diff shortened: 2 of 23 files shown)


src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FilterRepository.cs
```diff
@@ -2,24 +2,29 @@
...
+
+        public FilterRepository(StatisticsDbContext context)
         {
+            _context = context;
         }
 
-        public List<Filter> GetFiltersIncludingItems(Guid subjectId)
+        public Task<List<Filter>> GetFiltersIncludingItems(Guid subjectId)
         {
-            return FindMany(filter => filter.SubjectId == subjectId)
```
duncan-at-hiveit, line 19: Are there more usages of FindMany() or can it go now?

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
```diff
@@ -335,5 +333,114 @@ private static string OrderLocationAttributes(LocationAttributeNode node)
...
+        private async Task<Either<ActionResult, Unit>> ValidateFiltersForSubject(
+            Guid subjectId,
+            List<FilterUpdateViewModel> requestFilters)
+        {
+            var filters = await _filterRepository.GetFiltersIncludingItems(subjectId);
+            return AssertCollectionsAreSameIgnoringOrder(filters,
+                requestFilters,
+                filter => filter.Id,
+                requestFilter => requestFilter.Id,
+                "Requested filters do not match subject filters").OnSuccess(_ =>
```
duncan-at-hiveit, line 346: Would you not want to add to the ValidationMessages Enum here to produce and error code rather than needing an plain English string here?  More machine readable for API clients to use?

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
```diff
@@ -335,5 +333,114 @@ private static string OrderLocationAttributes(LocationAttributeNode node)
...
+            {
+                var requestMap = requestFilters.ToDictionary(filter => filter.Id);
+                var results = filters.Select(filter =>
+                        ValidateFilterGroupsForSubject(filter, requestMap[filter.Id].FilterGroups))
+                    .ToList();
+
+                var failure = results.FirstOrDefault(result => result.IsLeft)?.Left;
```
duncan-at-hiveit, line 353: I think we have a helper method for this somewhere... will pop it on here if I find it

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
```diff
@@ -335,5 +333,114 @@ private static string OrderLocationAttributes(LocationAttributeNode node)
...
+        }
+
+        private static Either<ActionResult, Unit> AssertCollectionsAreSameIgnoringOrder<TLeft, TRight, TId>(
+            IEnumerable<TLeft> left,
+            IEnumerable<TRight> right,
+            Func<TLeft, TId> leftIdSelector,
+            Func<TRight, TId> rightIdSelector,
+            string error)
+        {
+            var rightIdList = right.Select(rightIdSelector).ToList();
```
duncan-at-hiveit, line 414: I tend to equate "left" and "right" with error handling.  Mayeb change the terminology slightly?  First, Second?  TFirst, TSecond?

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
```diff
@@ -335,5 +333,114 @@ private static string OrderLocationAttributes(LocationAttributeNode node)
...
+            return AssertCollectionsAreSameIgnoringOrder(left, rightIdList, leftIdSelector, error);
+        }
+
+        private static Either<ActionResult, Unit> AssertCollectionsAreSameIgnoringOrder<TLeft, TId>(
+            IEnumerable<TLeft> left,
+            IEnumerable<TId> right,
```
duncan-at-hiveit, line 420: Why do we have one of these lists as an Id list and the other as entities with an Id selector?

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FilterRepository.cs
benoutram: With a bit of refactoring of `BoundaryLevelRepository` and things that use it, I've been able to get rid of it now. Thanks.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
benoutram: Renamed to first/second/TFirst/TSecond etc, thanks.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
benoutram: It was for convenience because in a couple of cases we want to compare a list of either filter items or indicators, with a list of Id's that are supplied in the request.

In the other cases we want to compare two lists of entities where the lists are not the same type and supply Id selectors for both.

I've simplified things a bit now so we don't have this overload anymore. Now we either make the comparison with two Id lists or two entity lists with Id selectors.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
benoutram: I've added an `OnSuccessAll` and a `OnSuccessVoid` to `EitherExtensions` so that I can get all the comparisons and then return `OnSuccessAll().OnSuccessVoid()`. These are non-async extension methods as opposed to our other task extensions.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
duncan-at-hiveit: Oh cool - but I'm sorry, I thought that our ValidationResults could take arguments as well as the error code, whereas I see that I don't think we have that right now.  Happy if you wanted to back this out :grimacing: or equally happy for you to keep it as codes.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
duncan-at-hiveit: :+1:

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
duncan-at-hiveit: Cool that works better for me personally, ta!

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
duncan-at-hiveit: Nice one, I see.  Cool yeah I think a bit simpler is probably better in the long run.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
benoutram: I think it's cleaner with the codes rather than the strings so will keep it like it is for now. Being able to add arguments will be a good improvement for the future though!

duncan-at-hiveit (approved):

benoutram:
image

Latest UI test report run with temporary fix `user sets focus to element  xpath://button[text()="Resolve"]  ${block}` in `release_content_authoring.robot` to prevent failure in test `Switch back to analyst1 to resolve second text block` when clicking the Resolve button.

PR #3303, opened 2022-05-20
EES-3395 Fix bug with location deduplication and tidy-up building location view models

benoutram:
There was a bug with the back-end location deduplication code as it stood, because for attributes where label AND value are the same, it tried to deduplicated these by adding geographic level to the label.

_I don’t think_  it’s correct anymore to try and add geographic level to the label because the view models for each set of attributes are built and deduplicated per geographic level, so there can't be any duplicates in the list which then vary by level. Level is also not a field that’s populated on leaf `LocationAttributeNode` so when it’s appended it appeared as an empty set of curly braces.

image

We also take the opportunity in this PR to tidy up the way we build the location view models.

The deduplication code was in `AbstractSubjectMetaService`, the base class of `SubjectMetaService` and `ResultSubjectMetaService`. This meant it was only being indirectly unit-tested and led to the question of which service should be responsible for testing it? We ended up with an attempt to test it in both, but this probably wasn't exhaustive.

There was also a lot of duplication in the way that we build the view models between `SubjectMetaService`  and `ResultSubjectMetaService`. These services were already complicated and now benefit from the common code being abstracted out into a new helper class. 

# Other changes
 - Use `record` type rather than `class` for all the location attribute types which make up a `Location`, e.g. `Country`, `Region` etc. Each type has a primary constructor which gives it public init-only properties for `Code` and `Name`. `RscRegion` gets only `Code`  and `LocalAuthority` has an `OldCode`. This simplifies the declaration of all these types which are moved to `LocationAttribute.cs`, replacing the previous class declarations which were made one file per class.


src/GovUk.Education.ExploreEducationStatistics.Data.Model/LocationAttribute.cs
```diff
@@ -1,35 +1,77 @@
...
+
+    public override string GetCodeOrFallback()
+    {
+        return string.IsNullOrEmpty(Code) ? OldCode ?? string.Empty : Code;
+    }
+}
+
+public record LocalAuthorityDistrict(string? Code, string? Name) : LocationAttribute(Code, Name);
+
+public record LocalEnterprisePartnership(string? Code, string? Name) : LocationAttribute(Code, Name);
```
duncan-at-hiveit, line 55: Very nice simplification!

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/LocationViewModelBuilder.cs
```diff
@@ -0,0 +1,123 @@
...
+public class LocationViewModelBuilder
+{
+    public static Dictionary<GeographicLevel, List<LocationAttributeViewModel>> BuildLocationAttributeViewModels(
+        IList<Location> locations,
+        Dictionary<GeographicLevel, List<string>>? hierarchies,
+        Dictionary<GeographicLevel, Dictionary<string, GeoJson>>? geoJson = null)
+    {
+        var locationAttributes = locations.GetLocationAttributesHierarchical(hierarchies);
+        return locationAttributes.ToDictionary(
+            pair => pair.Key,
```
duncan-at-hiveit, line 20: Might be tempted to call "pair" something that hints at what is contained in the pair, like "codeAndLocationAttributes", but not massively fussed.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/LocationViewModelBuilder.cs
```diff
@@ -0,0 +1,123 @@
...
+        // TODO EES-2954 enhance the deduplication to append other attributes of the location
+        var notDistinctByLabelAndValue = locations
+            .GroupBy(model => (model.Value, model.Label))
+            .Where(grouping => grouping.Count() > 1)
+            .SelectMany(grouping => grouping)
+            .ToList();
+
+        // Excluding those, find duplicates due to having the same label
+        var notDistinctByLabel = locations
+            .Except(notDistinctByLabelAndValue)
```
duncan-at-hiveit, line 90: Cool, didn't know you could do "Except"

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/LocationViewModelBuilder.cs
```diff
@@ -0,0 +1,123 @@
...
+        };
+    }
+
+    private static List<LocationAttributeViewModel> DeduplicateLocationViewModels(
+        List<LocationAttributeViewModel> locations)
+    {
+        // Find duplicates by label which also have the same value.
+        // They are unique by some other attribute of location and appending the value of this attribute won't deduplicate them.
+        // TODO EES-2954 enhance the deduplication to append other attributes of the location
+        var notDistinctByLabelAndValue = locations
```
duncan-at-hiveit, line 82: So at the moment after this work goes in, we'll still have the duplicate Locations in the front end that are appearing because of some other hidden attribute, but we'll not see the empty braces where empty value was being appended anymore?  And 2954 will address the hidden attribute surfacing?

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/LocationViewModelBuilder.cs
```diff
@@ -0,0 +1,123 @@
...
+            {
+                // Append the value to make them differentiable by label
+                location.Label += $" ({location.Value})";
+            }
+
+            return location;
+        }).ToList();
+    }
+
+    private static string OrderLocationAttributes(LocationAttributeNode node)
```
duncan-at-hiveit, line 113: I'd expect this method to do ordering, so I feel like the name might not be quite right or I'm not understanding it?

duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/LocationViewModelBuilder.cs
benoutram: Yep, that's the plan if we decide to do it. There's some considerations added to the comments of EES-2954 and a plan for implementing it.

With this refactoring if should be easier to do now as we have the 'full' locations closer to where we do the deduplication.

We could also move the point where the deduplication happens to the area where the narrowing from Location to LocationAttribute occurs when building the location hierarchy rather than it operating on view models as it does at present.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/LocationViewModelBuilder.cs
benoutram: Thanks have renamed it `levelAndLocationAttributes` although I think it would be really nice if the language allowed something like

```
.ToDictionary(
    (level, _) => level,
    (level, locationAttributes) => DoSomething(level, locationAttributes)
)

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/LocationViewModelBuilder.cs
benoutram: Thanks, it's picking the field that is used for the ordering, so I've renamed it `OrderingKey` rather `OrderLocationAttributes`. Let me know if you think of a better name.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/LocationAttribute.cs
benoutram: Thanks! 🙂

PR #3307, opened 2022-05-25
EES-3015 EES-3026 improve table header reordering accessibility

amyb-hiveit:
Fixes two accessibility problems with reordering table headers:

- EES-3015: Nested active focusable elements in table tool reordering
- EES-3026: Instructions provided to screen reader users do not adequately indicate how to access content

Changes:
- only reordering groups or items can be active at once, when groups or items are inactive they don't have the role of button so there aren't nested active elements any more.
- buttons have been added to row & column groups to switch to/from reordering groups, by default reorder items is selected
- added move to rows / columns buttons to groups to make it clearer that you can do this and because moving them to the other axis with the keyboard doesn't work consistently. 
- swapped columns and rows so column groups is at the top
- used names of indicators etc for groups instead of 'row group x'
- changed the drag icon and added it to groups 
- drag and drop was sometimes failing to load correctly on SSR pages (fast tracks) so have put in a mounted check
- updated the help text and added some screen reader announcements.
- removing the nested active elements means that the drag n drop screen reader announcements work.
- I've fixed the width of the groups and made it scroll horizontally if they don't fit, previously they got squashed and could become unusable if there were several groups on a small screen. Note that there is a warning about nested scroll containers not being supported by react-beautiful-dnd but I've not found it to cause any problems.

## Screenshots

reorderingpr

### DIFF
--- a/src/explore-education-statistics-common/src/modules/table-tool/components/FormFieldSortableListGroup.module.scss
+++ b/src/explore-education-statistics-common/src/modules/table-tool/components/FormFieldSortableListGroup.module.scss
@@ -60,7 +60,9 @@
   }
 
   legend {
-    padding-right: govuk-spacing(4);
+    display: flex;
+    justify-content: space-between;
+    width: 100%;
   }
 }
 
@@ -90,11 +92,10 @@
   .dragIcon {
     display: block;
     fill: currentColor;
-    height: 18px;
-    position: absolute;
-    right: 16px;
-    top: 12px;
-    width: 18px;
+    flex-shrink: 0;
+    height: 1rem;
+    margin-top: 0.2rem;
+    width: 1rem;
   }
 }
 
--- a/src/explore-education-statistics-common/src/modules/table-tool/components/FormFieldSortableListGroup.module.scss
+++ b/src/explore-education-statistics-common/src/modules/table-tool/components/FormFieldSortableListGroup.module.scss
@@ -1,59 +1,118 @@
 @import '~govuk-frontend/govuk/base';
 
-.groupsFieldset {
-  border: 4px solid govuk-colour('light-grey');
-  padding: govuk-spacing(2);
-  width: 100%;
-}
-
-.isDraggingOver {
+.container {
   background-color: govuk-colour('mid-grey');
-  border-color: $govuk-focus-colour;
+  border: 4px solid govuk-colour('light-grey');
+  overflow-x: auto;
+  padding: govuk-spacing(2);
+  position: relative;
+  width: 100%;
+
+  &.isDragDisabled {
+    background: none;
+  }
+
+  &.isDraggingOver {
+    border-color: $govuk-focus-colour;
+  }
 }
 
-.listsContainer {
+.toggleReorderGroups {
+  @include govuk-media-query($from: desktop) {
+    position: absolute;
+    right: govuk-spacing(2);
+    top: govuk-spacing(2);
+
+    &:active {
+      top: govuk-spacing(2);
+    }
+  }
+}
+
+.groupsContainer {
   display: flex;
   flex-direction: column;
-  padding: govuk-spacing(2);
   width: 100%;
 
   @include govuk-media-query($from: desktop) {
     flex-direction: row;
   }
+
+  &.isActive {
+    position: relative;
+  }
 }
 
-.list {
+.groupContainer {
+  padding: 0 govuk-spacing(1);
+}
+
+.group {
   background: govuk-colour('light-grey');
+  box-sizing: border-box;
   display: block;
-  flex-basis: 100%;
   margin-bottom: govuk-spacing(2);
-  margin-right: govuk-spacing(2);
   padding: govuk-spacing(2);
+  position: relative;
 
   @include govuk-media-query($from: desktop) {
-    flex-basis: 33%;
+    width: 320px;
   }
 
-  &:focus {
-    border: $govuk-focus-width solid #000;
+  legend {
+    padding-right: govuk-spacing(4);
+  }
+}
+
+.dragIcon {
+  display: none;
+}
+
+.groupIsActive {
+  background: govuk-colour('white');
+
+  &:focus-visible {
     outline: $govuk-focus-colour solid $govuk-focus-width;
   }
 
   &:hover {
-    background: govuk-colour('mid-grey');
-
+    background: govuk-colour('light-grey');
     // Fallback if grab is unsupported
     cursor: move;
     cursor: grab;
   }
+
+  &.isDragging {
+    background: govuk-colour('blue');
+    outline: $govuk-focus-width solid $govuk-focus-colour;
+  }
+
+  .dragIcon {
+    display: block;
+    fill: currentColor;
+    height: 18px;
+    position: absolute;
+    right: 16px;
+    top: 12px;
+    width: 18px;
+  }
 }
 
 .isDragging {
-  outline: $govuk-focus-width solid $govuk-focus-colour;
-  outline-offset: 0;
+  legend {
+    color: govuk-colour('white');
+  }
 }
 
-.optionDraggedOutside {
+.isDraggedOutside {
   opacity: 0.4;
   outline: none;
+}
+
+.moveButton {
+  max-width: 100%;
+
+  @include govuk-media-query($from: desktop) {
+    width: 320px;
+  }
 }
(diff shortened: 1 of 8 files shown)


src/explore-education-statistics-common/src/modules/table-tool/components/FormSortableList.module.scss
```diff
@@ -26,47 +35,60 @@
...
   border: 0;
   opacity: 0.2;
   outline: none;
 }
 
+.dragIcon {
+  display: block;
+  fill: currentColor;
+  height: 18px;
+  position: absolute;
```
ntsim, line 73: Do we need to absolute positioning? Feels like flex should work fine here?

src/explore-education-statistics-common/src/modules/table-tool/components/FormSortableList.module.scss
```diff
@@ -26,47 +35,60 @@
...
-  flex-direction: column;
+  &:hover {
+    background: darken(govuk-colour('blue'), 5);
+  }
 }
 
-.optionDragging {
+.isDragging {
   background: govuk-colour('blue');
   border: $govuk-focus-width solid govuk-colour('black');
```
ntsim, line 54: Just noticed that when dragging, the border causes displacement of the child elements inside of the box. Think we can fix this with:

```scss
box-shadow: inset 0 0 0 $govuk-focus-width govuk-colour('black');
```

src/explore-education-statistics-common/src/modules/table-tool/components/FormFieldSortableListGroup.module.scss
```diff
@@ -1,59 +1,118 @@
...
+  &.isDragging {
+    background: govuk-colour('blue');
+    outline: $govuk-focus-width solid $govuk-focus-colour;
+  }
+
+  .dragIcon {
+    display: block;
+    fill: currentColor;
+    height: 18px;
+    position: absolute;
```
ntsim, line 94: Do we need to use absolute positioning here? Feels like flex should work here?

src/explore-education-statistics-common/src/modules/table-tool/components/FormFieldSortableListGroup.module.scss
```diff
@@ -1,59 +1,118 @@
...
+
+  &.isDragging {
+    background: govuk-colour('blue');
+    outline: $govuk-focus-width solid $govuk-focus-colour;
+  }
+
+  .dragIcon {
+    display: block;
+    fill: currentColor;
+    height: 18px;
```
ntsim, line 93: Not a big deal, but we typically declare sizes with rems rather than px. Could also do this for the drag icon in `FormSortableList`

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersForm.tsx
```diff
@@ -50,113 +56,202 @@ const TableHeadersForm = ({
     [onSubmit],
   );
 
+  const handleMoveGroupToOtherAxis = useMemo(
```
ntsim, line 59: Shouldn't this be `useCallback`?

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersForm.tsx
```diff
@@ -50,113 +56,202 @@ const TableHeadersForm = ({
...
+      const nextSourceValue = [...form.values[sourceId]];
+      const sourceItem = nextSourceValue.splice(groupIndex, 1);
+      const nextDestinationValue = [...form.values[destinationId]].concat(
+        sourceItem,
+      );
+
+      form.setFieldValue(sourceId, nextSourceValue);
+      form.setFieldValue(destinationId, nextDestinationValue);
+      form.setFieldTouched(sourceId);
+      form.setFieldTouched(destinationId);
```
ntsim, line 77: We're doing a bunch of the same stuff in the DragDropContext's `onDragEnd`. Perhaps we could refactor this out into a function and re-use it in both places.

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersForm.tsx
```diff
@@ -50,113 +56,202 @@ const TableHeadersForm = ({
...
-      </Formik>
-    </Details>
+                        0,
+                        sourceItem,
+                      );
+
+                      form.setFieldValue(sourceId, nextSourceValue);
+                      form.setFieldValue(destinationId, nextDestinationValue);
+                      form.setFieldTouched(sourceId);
+                      form.setFieldTouched(destinationId);
```
ntsim, line 196: We might want to just extract this out into a `handleDragEnd` callback as it's a bit unwieldy currently.

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersForm.tsx
```diff
@@ -50,113 +56,202 @@ const TableHeadersForm = ({
...
+                    <FormGroup>
+                      <div className="govuk-!-margin-bottom-2">
+                        <FormFieldSortableListGroup<
+                          PickByType<TableHeadersConfig, Filter[][]>
+                        >
+                          name="columnGroups"
+                          legend="Column groups"
+                          isDraggingGroup={isDraggingGroup}
+                          onChangeReorderingType={handleSwitchReorderingType}
+                          onMoveGroupToOtherAxis={(groupIndex: number) =>
```
ntsim, line 209: Don't need an explicit `number` type here

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersForm.tsx
```diff
@@ -50,113 +56,202 @@ const TableHeadersForm = ({
...
+
+                      <div className="govuk-!-margin-bottom-2">
+                        <FormFieldSortableListGroup<
+                          PickByType<TableHeadersConfig, Filter[][]>
+                        >
+                          name="rowGroups"
+                          legend="Row groups"
+                          isDraggingGroup={isDraggingGroup}
+                          onChangeReorderingType={handleSwitchReorderingType}
+                          onMoveGroupToOtherAxis={(groupIndex: number) =>
```
ntsim, line 227: Don't need an explicit `number` type here

tests/robot-tests/tests/general_public/table_tool_absence_by_characteristic.robot
```diff
@@ -129,22 +128,29 @@ Reorder Gender to be column group
...
-    user presses keys    ARROW_DOWN
     user presses keys    ARROW_LEFT
     user presses keys    ${SPACE}
 
 Reorder Gender male to be second
+    ${columnGroups}=    user gets testid element    columnGroups
+    user clicks button    Re-order items    columnGroups
     # The /../.. to get to a focusable element
-    user sets focus to element    xpath://strong[text()="Gender male"]/../..
+    user sets focus to element    xpath://span[text()="Gender male"]/../..
```
ntsim, line 145: Would consider using `*` instead of `span`. If there are potentially other candidate elements, we should constrain the selector using the `parent` argument:

```robot
user sets focus to element   xpath://*[text()="..."]/../..   testid:sortableListGroup-columns
```
I've used a testid selector here (may need to add one to the parent container), but it could be whatever selector.

Same applies for the other keywords below

ntsim (changes requested):

src/explore-education-statistics-common/src/modules/table-tool/components/FormFieldSortableListGroup.tsx
```diff
@@ -34,59 +62,100 @@ function FormFieldSortableListGroup<FormValues>({
...
+            >
+              {isDragDisabled
+                ? `Re-order ${legend.toLowerCase()}`
+                : 'Re-order items'}
+            </Button>
+            <div
+              className={classNames(styles.groupsContainer, {
+                [styles.isActive]: !isDragDisabled,
+              })}
+              id="groupsContainer"
```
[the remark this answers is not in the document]
amyb-hiveit: Nope, I've removed it now.

src/explore-education-statistics-common/src/modules/table-tool/components/FormFieldSortableListGroup.tsx
```diff
@@ -34,59 +62,100 @@ function FormFieldSortableListGroup<FormValues>({
...
+              className={styles.toggleReorderGroups}
+              variant="secondary"
+              onClick={() => {
+                onChangeReorderingType(legend, isDragDisabled);
+                toggleDragDisabled();
+              }}
+            >
+              {isDragDisabled
+                ? `Re-order ${legend.toLowerCase()}`
+                : 'Re-order items'}
```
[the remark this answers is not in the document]
amyb-hiveit: I chatted to Marv about this earlier, he's had some feedback on it from the user testing and thinks it needs some more thought as well. I'm going to mark this PR as do not merge for now and book in a meeting for us, Marv and Cam to have a sketching session for it.

PR #3312, opened 2022-05-27
Ees 3321 publication approver role as mirroring release approver role

duncan-at-hiveit:
:warning: This PR will need to be accompanied by the `ees-mssql-data-45.zip` file in Google Drive.  It contains a new piece of bootstrapped data.

This PR
- adds in a new "ReleaseApprover" Publication Role that allows users with it to act as if they have the Release "Approver" role on all Releases within that Publication.  I called it ReleaseApprover rather than plain "Approver" because I felt that an "Approver" Publication Role would be confusing given that it does nothing to permissions around the Publications themselves, but rather only to Releases under them.
- effectively gives the user the equivalent of the Release Approver role on every Release within the Publication.

I have updated the Permissions Matrix to reflect this new role.

### UserReleaseRole, UserPublicationRole and the new base UserResourceRole
UserReleaseRole and UserPublicationRole both share a lot of very similar traits, so much so that I thought about trying to combine as many of their shared behaviours as possible and deduplicate various parts of the codebase where we effectively run the same codepaths for the 2 types of Roles but duplicated.

In the end, I stopped short of this but combined similar traits into a `UserResourceRole` base class, and an `AbstractUserResourceRoleRepository` from which the 2 concrete role repositories extend.

### Adding AuthorizationHandlerResourceRoleService, removing AuthorizationHandlerUtil
Whilst adding in the new permissions for the new Publication-level role, I found that we were repeating a lot of common patterns in our Auth Handlers, and relying on various helper methods within the static AuthorizationHandlerUtil to inspect the results of calls to role repositories.

Rather than adding more helper methods and lots more additional calls to the publication role repository to individual handlers, I decided it might be better to encapsulate those repeating patterns into a new service, and do away with AuthorizationHandlerUtil class altogether.

The methods available on AuthorizationHandlerResourceRoleService identify those repeating patterns.

Note that AuthorizationHandlerResourceRoleService currently has no interface - this is because currently we are not needing to mock it.  Instead, test classes now create a new AuthorizationHandlerResourceRoleService and pass in Mock role repositories into it.  However, there is a great opportunity to reduce the complexity of our Auth Handler tests by mocking AuthorizationHandlerResourceRoleService.  I chose not to do this ~~because I was too lazy~~ because this would have meant drastic changes to lots of Auth Handler tests, and would make it very hard to determine the actual changes occurring in this PR.  We should definitely do it in the future though, meaning that we don't have to burn hundreds of test lines setting up UserReleaseRoles and UserPublicationRoles in different combinations again and again.

## Dashboard Theme, Topic and Publication services

I have updated various calls around the Admin Dashboard for supplying a user's Themes, Topics, Publications and Releases based on their Publication and Release Roles to accommodate this new role.

## UI tests

Raised https://dfedigital.atlassian.net/browse/EES-3406 to create UI tests for this ticket.  Not doing immediately as it requires more seeded data and don't want to hold up the test user accounts work.

image

UPDATE:  added tests from 3406 to this PR:

image


src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/ViewSpecificPublicationAuthorizationHandlers.cs
```diff
@@ -62,12 +63,29 @@ public HasRoleOnAnyChildReleaseAuthorizationHandler(ContentDbContext context)
...
+            protected override async Task HandleRequirementAsync(
+                AuthorizationHandlerContext context, 
+                ViewSpecificPublicationRequirement requirement,
+                Publication publication)
+            {
+                if (await _authorizationHandlerResourceRoleService
+                        .HasRolesOnPublication(
+                            context.User.GetUserId(),
+                            publication.Id,
+                            Owner))
```
markhiveit, line 85: `ViewSpecificReleaseAuthorizationHandlers` allows publication owners and publication approvers - I think I'd expect this to do the same?

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/AssignPrereleaseContactsToSpecificReleaseAuthorizationHandlers.cs
```diff
@@ -37,19 +33,14 @@ public class AssignPrereleaseContactsToSpecificReleaseAuthorizationHandler
...
             
-            var releaseRoles = await _userReleaseRoleRepository.GetDistinctRolesByUserAndRelease(context.User.GetUserId(), release.Id);
-
-            if (ContainsEditorOrApproverRole(releaseRoles))
+            if (await _authorizationHandlerResourceRoleService
+                    .HasRolesOnPublicationOrRelease(
+                        context.User.GetUserId(),
+                        release.PublicationId,
+                        release.Id,
+                        ListOf(PublicationRole.Owner),
```
markhiveit, line 42: Should PublicationRole.ReleaseApprover be included here, since ReleaseRole.Approver is included?

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/PublishSpecificReleaseAuthorizationHandler.cs
```diff
@@ -37,10 +37,14 @@ public PublishSpecificReleaseAuthorizationHandler(IUserReleaseRoleRepository use
...
-            var releaseRoles = await _userReleaseRoleRepository.GetDistinctRolesByUserAndRelease(context.User.GetUserId(), release.Id);
-
-            if (ContainsApproverRole(releaseRoles))
+            
+            if (await _authorizationHandlerResourceRoleService
+                    .HasRolesOnPublicationOrRelease(
+                        context.User.GetUserId(),
+                        release.PublicationId,
+                        release.Id,
+                        ListOf(PublicationRole.Owner),
```
markhiveit, line 46: Should PublicationRole.ReleaseApprover be included here, since ReleaseRole.Approver is included?

markhiveit (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationRepository.cs
```diff
@@ -62,7 +66,7 @@ public async Task<List<MyPublicationViewModel>> GetAllPublicationsForTopic(Guid
             var publicationViewModels = new List<MyPublicationViewModel>();
 
             // Add publication view models for the Publications granted by the Publication Owner role
```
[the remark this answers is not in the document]
duncan-at-hiveit: Ooh good spot - updated.

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/ViewSpecificPublicationAuthorizationHandlers.cs
duncan-at-hiveit: Totally agree, and surprised this didn't cause issues when I was doing my testing!  Updated.

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/PublishSpecificReleaseAuthorizationHandler.cs
duncan-at-hiveit: yes it should!  Added, thanks!

markhiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Release.cs
```diff
@@ -225,6 +225,10 @@ public PartialDate NextReleaseDate
                 {
                     _nextReleaseDate = value;
                 }
+                else if (value.IsEmpty())
+                {
+                    _nextReleaseDate = null;
+                }
```
markhiveit, line 231: What's this change for?

tests/robot-tests/tests/admin/analyst/edit_and_approve_release_as_publication_release_approver.robot
```diff
@@ -0,0 +1,139 @@
+*** Settings ***
```
markhiveit, line 1: I think this file should live in `tests/admin_and_public` since it does visit the public frontend at the end.

Or maybe you could remove the bit that checks the public frontend - not sure how much value that provides?

markhiveit (changes requested):

tests/robot-tests/tests/admin/analyst/edit_and_approve_release_as_publication_release_approver.robot
duncan-at-hiveit: Yeah I agree, good shout.  I think I'll just remove it and just make sure we're checking that the Release status makes it to Published (which we already do)

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Release.cs
duncan-at-hiveit: Oh yeah, forgot about this.  A bug that Nusrath and I spotted when writing the UI tests.  You couldn't unset a Next Release Date during a Release status update if it had already been set to something once.

markhiveit (approved):

PR #3314, opened 2022-05-27
EES-3386 Add ability to assign user release roles when inviting a new user

markhiveit:
This work allows a BAU user to assign user release roles when inviting a new user. These user release roles are then created when the user logs in the for the first time.

This work also ensures that if a pending invite is cancelled, any associated user release role invites are also cancelled.

Additional work to change the invite email text to include the user release roles they've been invited to (EES-3403) and listing the user release roles of pending invites on the Pending user invites page (EES-3404) will be done in separate PRs.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserManagementServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserManagementServiceTests.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using GovUk.Education.ExploreEducationStatistics.Admin.Areas.Identity.Data;
+using GovUk.Education.ExploreEducationStatistics.Admin.Areas.Identity.Data.Models;
 using GovUk.Education.ExploreEducationStatistics.Admin.Models;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces;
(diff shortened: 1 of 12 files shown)


src/explore-education-statistics-admin/src/pages/users/UserInvitePage.tsx
```diff
@@ -91,67 +120,175 @@ const UserInvitePage = ({
...
+            onSubmit={handleSubmit}
+          >
+            {form => {
+              return (
+                <Form id={formId}>
+                  <FormFieldset
+                    id="email-fieldset"
+                    legend="Provide the email address for the user"
+                    legendSize="m"
+                    hint="The invited user must have a @education.gov.uk email address"
```
ntsim, line 155: This specifies that the email should be a DfE email, but is this actually true? Aren't MoJ users using the service?

If we are supposed to restrict this to just DfE users, then we should probably also add a custom Yup validation to enforce this.

src/explore-education-statistics-admin/src/pages/users/UserInvitePage.tsx
```diff
@@ -91,67 +120,175 @@ const UserInvitePage = ({
...
+                        <thead>
+                          <tr>
+                            <th scope="col">Release Title</th>
+                            <th scope="col">Release Role</th>
+                            <th scope="col">Options</th>
+                          </tr>
+                        </thead>
+                        <tbody>
+                          {form.values.selectedUserReleaseRoles.map(
+                            selectedUserReleaseRole => (
```
ntsim, line 247: Simplify this to `userReleaseRole` in-line with other suggestion to remove 'selected' prefixes.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserManagementServiceTests.cs
```diff
@@ -203,14 +206,305 @@ await using (var userAndRolesDbContext = InMemoryUserAndRolesDbContext(userAndRo
...
+                Assert.Equal("test@test.com", userReleaseInvite.Email);
+                Assert.Equal(release.Id, userReleaseInvite.ReleaseId);
+                Assert.Equal(Approver, userReleaseInvite.Role);
+                Assert.False(userReleaseInvite.Accepted);
+                Assert.True(userReleaseInvite.EmailSent);
+                Assert.InRange(DateTime.UtcNow.Subtract(userReleaseInvite.Created).Milliseconds, 0, 1500);
+                Assert.Equal(_createdById, userReleaseInvite.CreatedById);
+            }
+
+            VerifyAllMocks(emailTemplateService);
```
ntsim, line 293: Discussed this before with @duncan-at-hiveit, and think we agreed that mock verification should occur immediately after the method is called so that failure feedback occurs sooner. 

Same applies to any other changes, but also might want to update the rest of this file accordingly too.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserManagementServiceTests.cs
```diff
@@ -203,14 +206,305 @@ await using (var userAndRolesDbContext = InMemoryUserAndRolesDbContext(userAndRo
...
+                        },
+                    });
+
+                result.AssertRight();
+            }
+
+            await using (var usersAndRolesDbContext = InMemoryUserAndRolesDbContext(usersAndRolesDbContextId))
+            {
+                var userInvites = usersAndRolesDbContext.UserInvites
+                    .Where(invite => invite.Email == "test@test.com")
```
ntsim, line 266: If we have this `Where`, we don't really prove that only exactly one `UserInvite` was created. Perhaps there's a bug that creates an additional, invalid invite for example.

Would suggest removing the `Where` so that we can rest assured that only is created. We could do a similar thing across the rest of the test cases and checks against `UserReleaseInvites`.

src/explore-education-statistics-admin/src/services/userService.ts
```diff
@@ -43,6 +43,7 @@ export interface UserPublicationRoleSubmission {
 export interface UserInvite {
   email: string;
   roleId: string;
+  userReleaseRoles: { releaseId: string; releaseRole: string }[];
```
ntsim, line 46: Couldn't this just be `releaseRoles`?

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/users/UserInvitePage.tsx
```diff
@@ -91,67 +120,175 @@ const UserInvitePage = ({
...
+                <Form id={formId}>
+                  <FormFieldset
+                    id="email-fieldset"
+                    legend="Provide the email address for the user"
+                    legendSize="m"
+                    hint="The invited user must have a @education.gov.uk email address"
+                  >
+                    <FormFieldTextInput<FormValues>
+                      label="User email"
+                      name="userEmail"
```
ntsim, line 159: This field is way too wide for the content expected in it. Setting `width={20}` would be more appropriate.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/AddUserReleaseRoleViewModel.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.ViewModels
 {
-    public class AddReleaseRoleViewModel
+    public class AddUserReleaseRoleViewModel
```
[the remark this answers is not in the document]
markhiveit: I've renamed it to `UserReleaseRoleAddViewModel`.

I added the `User` to distinguish between `ReleaseRole`s like `Approver` and an object that contains a release ID and a release role. `UserReleaseRole` and `UserPublicationRole` are commonly used for this type of object elsewhere in the codebase. I think `User` does make sense because these objects are attached to specific users. It's possible there is a better way to name it - something like `ReleaseRolePermission`? - but with so much of the codebase already using `UserReleaseRole` already, my preference would be to stick to that.

src/explore-education-statistics-admin/src/pages/users/UserInvitePage.tsx
```diff
@@ -12,12 +16,23 @@ import { mapFieldErrors } from '@common/validation/serverValidations';
...
+import React from 'react';
 import { RouteComponentProps } from 'react-router';
+import useAsyncHandledRetry from '@common/hooks/useAsyncHandledRetry';
+import LoadingSpinner from '@common/components/LoadingSpinner';
+import { IdTitlePair } from '@admin/services/types/common';
 
 interface FormValues {
   userEmail: string;
   selectedRoleId: string;
+  selectedUserReleaseRoles: {
```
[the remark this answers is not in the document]
markhiveit: As above, I prefer `userReleaseRoles` to distinguish a releaseId+releaseRole combo from `releaseRoles` like Approver.

src/explore-education-statistics-admin/src/pages/users/UserInvitePage.tsx
```diff
@@ -61,25 +89,26 @@ const UserInvitePage = ({
...
+        <div className="govuk-grid-row">
+          <div className="govuk-grid-column-two-thirds">
+            <h1 className="govuk-heading-xl">
+              <span className="govuk-caption-xl">
+                Manage access to the service
+              </span>
+              Invite a new user
+            </h1>
+          </div>
+          {/* EES-2464
```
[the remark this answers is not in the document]
markhiveit: I've removed it and left a comment on the EES-2464 ticket.

src/explore-education-statistics-admin/src/pages/users/UserInvitePage.tsx
markhiveit: Talked to Cam and he suggested some alternative text which I've replaced this hint with.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
```diff
@@ -228,10 +234,22 @@ await _usersAndRolesDbContext.UserInvites
...
                         roleId: roleId,
                         createdById: _userService.GetUserId());
+
+                    foreach (var userReleaseRole in userReleaseRoles)
+                    {
+                        await _userReleaseInviteRepository.Create(
+                            releaseId: userReleaseRole.ReleaseId,
+                            email: email,
+                            releaseRole: userReleaseRole.ReleaseRole,
+                            emailSent: true, // EES-3403
```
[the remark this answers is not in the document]
markhiveit: I created EES-3403 after running the need to change the invite email text past Cam - I'll raise this before I do the work.

src/explore-education-statistics-admin/src/services/userService.ts
markhiveit: See previous comment: using `UserReleaseRole` to indicate it's a releaseId AND a releaseRole, not just a releaseRole.

ntsim (approved):

PR #3317, opened 2022-05-30
EES-3022 improve tabs accessibility

amyb-hiveit:
Accessibility improvements for tabs:
- when there's only one tab, for instance when datablocks with just a table are embedded in release content, the tab list is `aria-hidden` to prevent screen readers from seeing it.
- added screen reader only text to the tabs so they make sense when read out of context. This is via a new tabLabel field as the title is used in GA events so had to remain just a string
- changed it so the up and down keys move between tabs and prevented being able to loop over the tabs with the keyboard, this makes it more consistent with expected tabs behaviour and the GDS design system

## Related changes

- changed `user clicks link` to use xpath so that data block tab links can be clicked now that there is adjacent visually hidden text
- I noticed that the Headlines Key Stats block on releases has the title 'Headlines' in the admin and 'Summary' in the frontend, so have changed the admin one to be consistent with the frontend

### DIFF
--- a/src/explore-education-statistics-common/src/components/Tabs.tsx
+++ b/src/explore-education-statistics-common/src/components/Tabs.tsx
@@ -44,6 +44,7 @@ const Tabs = ({ children, id, modifyHash = true, testId, onToggle }: Props) => {
       section,
       {
         'aria-labelledby': `${sectionId}-tab`,
+        hasSiblings: filteredChildren.length > 1,
         hidden: selectedTabIndex !== index,
         id: sectionId,
         key: sectionId,
@@ -125,6 +126,8 @@ const Tabs = ({ children, id, modifyHash = true, testId, onToggle }: Props) => {
     };
   }, [sections, selectTab, selectedTabIndex]);
 
+  const hasSingleTab = sections.length === 1;
+
   return (
     <div
       className={classNames('govuk-tabs', styles.tabs)}
@@ -132,9 +135,16 @@ const Tabs = ({ children, id, modifyHash = true, testId, onToggle }: Props) => {
       data-testid={testId}
       ref={ref}
     >
-      <ul className="govuk-tabs__list" role="tablist">
+      <ul
+        aria-hidden={hasSingleTab}
+        className={classNames('govuk-tabs__list', {
+          [styles.mobileHidden]: hasSingleTab,
+        })}
+        role="tablist"
+      >
         {sections.map(({ props }, index) => {
           const sectionId = props.id || `${id}-${index + 1}`;
+
           return (
             <li
               className={classNames('govuk-tabs__list-item', {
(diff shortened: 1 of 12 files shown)


src/explore-education-statistics-common/src/components/TabsSection.tsx
```diff
@@ -56,14 +59,16 @@ const TabsSection = forwardRef<HTMLElement, TabsSectionProps>(
 
     return hasRendered ? (
       <section
-        aria-labelledby={onMedia(tabProps['aria-labelledby'])}
+        aria-labelledby={
+          !isSingleTab ? onMedia(tabProps['aria-labelledby']) : ''
```
ntsim, line 63: Would suggesting making the single tab branch use `undefined` to prevent the `aria-labelledby` attribute appearing in the markup as it would just be empty.

src/explore-education-statistics-common/src/components/Tabs.tsx
```diff
@@ -132,92 +135,82 @@ const Tabs = ({ children, id, modifyHash = true, testId, onToggle }: Props) => {
...
           const sectionId = props.id || `${id}-${index + 1}`;
           return (
             <li
               className={classNames('govuk-tabs__list-item', {
                 'govuk-tabs__list-item--selected': selectedTabIndex === index,
+                [styles.mobileHidden]: isSingleTab,
               })}
               key={sectionId}
-              role="presentation"
+              role={!isSingleTab ? 'presentation' : 'listitem'}
```
ntsim, line 148: What's the purpose of changing the role here? 

Also, given that this `li` element, it already has the implicit role of `listitem`. Should be able to set it to `undefined` instead

src/explore-education-statistics-common/src/components/Tabs.tsx
```diff
@@ -46,12 +46,15 @@ const Tabs = ({ children, id, modifyHash = true, testId, onToggle }: Props) => {
...
         hidden: selectedTabIndex !== index,
         id: sectionId,
+        isSingleTab: filteredChildren.length === 1,
         key: sectionId,
       },
       section.props.children,
     );
   });
 
+  const isSingleTab = sections.length === 1;
```
ntsim, line 56: In most usages of this, we've had to go with a negation of `!isSingleTab`. Perhaps it would be better to change this to a `isMultiTab` variable instead so that usages of it don't need the negation.

src/explore-education-statistics-common/src/components/Tabs.tsx
```diff
@@ -132,92 +135,82 @@ const Tabs = ({ children, id, modifyHash = true, testId, onToggle }: Props) => {
...
+                  tabIndex={selectedTabIndex !== index ? -1 : undefined}
+                  onClick={event => {
+                    event.preventDefault();
+                    selectTab(index);
+
+                    if (onToggle) {
+                      onToggle({
+                        title: props.title,
+                        id: sectionId,
+                      });
```
ntsim, line 169: Could change this to use optional chaining

src/explore-education-statistics-common/src/components/Tabs.tsx
```diff
@@ -132,92 +135,82 @@ const Tabs = ({ children, id, modifyHash = true, testId, onToggle }: Props) => {
...
+                      case 'ArrowUp': {
+                        event.preventDefault();
+                        const nextTabIndex = selectedTabIndex - 1;
+                        if (nextTabIndex >= 0) {
+                          selectTab(nextTabIndex);
+                          focusTab(nextTabIndex);
+
+                          onToggle?.({
+                            title: `${props.title}`,
+                            id: `${sectionId}`,
```
ntsim, line 184: These don't need to be template strings

src/explore-education-statistics-common/src/components/Tabs.tsx
```diff
@@ -132,92 +135,82 @@ const Tabs = ({ children, id, modifyHash = true, testId, onToggle }: Props) => {
...
+                        event.preventDefault();
+                        const nextTabIndex = selectedTabIndex + 1;
+
+                        if (nextTabIndex < sections.length) {
+                          selectTab(nextTabIndex);
+                          focusTab(nextTabIndex);
+
+                          onToggle?.({
+                            title: `${props.title}`,
+                            id: `${sectionId}`,
```
ntsim, line 201: These don't need to be template strings

src/explore-education-statistics-common/src/components/Tabs.tsx
```diff
@@ -132,92 +135,82 @@ const Tabs = ({ children, id, modifyHash = true, testId, onToggle }: Props) => {
...
-                          id: `${sectionId}`,
-                        });
+                  }}
+                  onKeyDown={event => {
+                    switch (event.key) {
+                      case 'ArrowLeft':
+                      case 'ArrowUp': {
+                        event.preventDefault();
+                        const nextTabIndex = selectedTabIndex - 1;
+                        if (nextTabIndex >= 0) {
```
ntsim, line 178: Nitpick, but could add a line break here to format this better

src/explore-education-statistics-common/src/components/Tabs.tsx
```diff
@@ -46,12 +46,15 @@ const Tabs = ({ children, id, modifyHash = true, testId, onToggle }: Props) => {
         'aria-labelledby': `${sectionId}-tab`,
         hidden: selectedTabIndex !== index,
         id: sectionId,
+        isSingleTab: filteredChildren.length === 1,
```
ntsim, line 49: In most usages of this, we've had to go with a negation of `!isSingleTab`. Perhaps it would be better to change this to a `isMultiTab` variable instead so that usages of it don't need the negation.

src/explore-education-statistics-common/src/components/TabsSection.tsx
```diff
@@ -56,14 +59,16 @@ const TabsSection = forwardRef<HTMLElement, TabsSectionProps>(
...
+        }
         hidden={hidden}
         className={classNames(classes.panel, styles.panel, {
           [styles.panelHidden]: hidden,
         })}
         id={id}
         ref={ref}
-        role={onMedia('tabpanel')}
+        role={!isSingleTab ? onMedia('tabpanel') : ''}
```
ntsim, line 71: would suggest making the single tab branch use `undefined` to prevent the `role` attribute appearing in the markup as it would just be empty.

src/explore-education-statistics-common/src/modules/find-statistics/components/DataBlockTabs.tsx
```diff
@@ -55,14 +55,24 @@ const DataBlockTabs = ({
     typeof additionalTabContent === 'function'
       ? additionalTabContent({ dataBlock })
       : additionalTabContent;
-
   return (
     <LoadingSpinner loading={isLoading}>
       <Tabs id={id} testId={testId(dataBlock)} onToggle={onToggle}>
         {firstTabs}
 
         {dataBlock.charts?.length && (
```
ntsim, line 63: Not a biggy, but probably worth just adding a `!!` to make sure this coerces correctly to a boolean. In other cases, we've made the mistake of accidentally showing 0s where this check is being performed.

tests/robot-tests/tests/libs/common.robot
```diff
@@ -477,6 +475,10 @@ user clicks link
     [Arguments]    ${text}    ${parent}=css:body
     user clicks element    link:${text}    ${parent}
 
+user clicks link by visible text
+    [Arguments]    ${text}    ${parent}=css:body
+    user clicks element    xpath:.//a[text()="${text}"]    ${parent}
```
ntsim, line 480: I'm surprised that the `xpath` selector is working compared to the `link` selector used by `user clicks link`. Would have thought these are basically identical in what they're doing.

If there is actually a difference here, would suggest scrapping this new keyword and change `user clicks link` to use the xpath selector instead. It's simpler and if it also covers cases that `link` can't, then it'd be more beneficial to be there instead.

ntsim (changes requested):

ntsim (approved):

PR #3324, opened 2022-06-08
EES-3387 Add ability to assign user publication roles when inviting a new user

markhiveit:
This work allows a BAU user to assign user publication roles when inviting a new user. These user publication roles are then created when the user logs in the for the first time.

image

This work also ensures that if a pending invite is cancelled, any associated user publication role invites are also cancelled. An additional bug was fixed that meant that an accepted invite could be cancelled / removed from the DB.

Additional work to change the invite email text to include the user release roles and user publication roles they've been invited to (EES-3403) and listing the user release roles and user publication roles of pending invites on the Pending user invites page (EES-3404) will be done in separate PRs.

In changes in response to PR comments commit, this PR now has removed the `Accepted` column from `UserReleaseInvites` - this means that now when any release or publication invite is processed, it is just deleted. To ensure the migration to remove the column doesn't cause issues, we delete all accepted `UserReleaseInvites` before removing the column.

This PR also now adds an `Updated` column to `UserReleaseInvites` for better auditability.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserManagementServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserManagementServiceTests.cs
@@ -223,8 +223,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                 await usersAndRolesDbContext.SaveChangesAsync();
             }
 
-            var release = new Release { Id = Guid.NewGuid() };
-            var publication = new Publication { Id = Guid.NewGuid() };
+            var releaseId = Guid.NewGuid();
+            var publicationId = Guid.NewGuid();
 
             var emailTemplateService = new Mock<IEmailTemplateService>(Strict);
 
@@ -247,7 +247,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                     {
                         new ()
                         {
-                            ReleaseId = release.Id,
+                            ReleaseId = releaseId,
                             ReleaseRole = Approver,
                         },
                     },
@@ -255,7 +255,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                     {
                         new ()
                         {
-                          PublicationId = publication.Id,
+                          PublicationId = publicationId,
                           PublicationRole = Owner,
                         },
                     });
@@ -284,9 +284,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                     .ToList();
                 var userReleaseInvite = Assert.Single(userReleaseInvites);
                 Assert.Equal("test@test.com", userReleaseInvite.Email);
-                Assert.Equal(release.Id, userReleaseInvite.ReleaseId);
+                Assert.Equal(releaseId, userReleaseInvite.ReleaseId);
                 Assert.Equal(Approver, userReleaseInvite.Role);
-                Assert.False(userReleaseInvite.Accepted);
                 Assert.True(userReleaseInvite.EmailSent);
                 Assert.InRange(DateTime.UtcNow.Subtract(userReleaseInvite.Created).Milliseconds, 0, 1500);
                 Assert.Equal(_createdById, userReleaseInvite.CreatedById);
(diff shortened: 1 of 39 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Model/UserPublicationInvite.cs
```diff
@@ -0,0 +1,35 @@
+#nullable enable
+using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Model
+{
+    public class UserPublicationInvite
```
ntsim, line 7: We've done work to manage these in the BAU pages, but it doesn't seem like there's anything to handle the conversion of them to user publication roles when the user is created?

Are we planning on doing this in a separate ticket?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
```diff
@@ -249,6 +253,16 @@ await _usersAndRolesDbContext.UserInvites
...
 
+                    foreach (var userPublicationRole in userPublicationRoles)
+                    {
+                        await _userPublicationInviteRepository.Create(
+                            publicationId: userPublicationRole.PublicationId,
+                            email: email,
+                            publicationRole: userPublicationRole.PublicationRole,
+                            emailSent: true, // EES-3403
+                            createdById: _userService.GetUserId());
+                    }
```
ntsim, line 264: This is a bit inefficient, shouldn't we do something like `UserReleaseInviteRepository.CreateManyIfNotExists`?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserManagementServiceTests.cs
```diff
@@ -288,11 +290,24 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+                var userPublicationInvite = Assert.Single(userPublicationInvites);
+                Assert.Equal("test@test.com", userPublicationInvite.Email);
+                Assert.Equal(publication.Id, userPublicationInvite.PublicationId);
+                Assert.Equal(Owner, userPublicationInvite.Role);
+                Assert.False(userPublicationInvite.Accepted);
+                Assert.True(userPublicationInvite.EmailSent);
+                Assert.InRange(DateTime.UtcNow.Subtract(userPublicationInvite.Created).Milliseconds, 0, 1500);
+                Assert.Equal(_createdById, userPublicationInvite.CreatedById);
+
+
```
ntsim, line 305: Can remove these empty line breaks

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserManagementServiceTests.cs
```diff
@@ -288,11 +290,24 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
                 Assert.True(userReleaseInvite.EmailSent);
                 Assert.InRange(DateTime.UtcNow.Subtract(userReleaseInvite.Created).Milliseconds, 0, 1500);
                 Assert.Equal(_createdById, userReleaseInvite.CreatedById);
+
+                var userPublicationInvites = contentDbContext.UserPublicationInvites
+                    .ToList();
+                var userPublicationInvite = Assert.Single(userPublicationInvites);
```
ntsim, line 296: Seems like we've done a bunch of tests around a single release and publication role being added, but we should probably have an additional test case for having multiple of each.

src/explore-education-statistics-admin/src/pages/users/UserInvitePage.tsx
```diff
@@ -160,6 +183,31 @@ const UserInvitePage = ({
                   }}
                 />
 
+                <InviteUserPublicationRoleForm
+                  publications={model?.publications}
+                  publicationRoles={model?.resourceRoles.Publication}
+                  userPublicationRoles={form.values.userPublicationRoles}
+                  onAddUserPublicationRole={(
+                    newUserPublicationRole: InviteUserPublicationRole,
```
ntsim, line 191: Don't need an explicit type (same could be done for the `InviteUserReleaseRoleForm` above as well.

src/explore-education-statistics-admin/src/pages/users/UserInvitePage.tsx
```diff
@@ -160,6 +183,31 @@ const UserInvitePage = ({
...
+                  ) => {
+                    form.setFieldValue('userPublicationRoles', [
+                      ...form.values.userPublicationRoles,
+                      newUserPublicationRole,
+                    ]);
+                  }}
+                  onRemoveUserPublicationRole={(
+                    userPublicationRoleToRemove: InviteUserPublicationRole,
```
ntsim, line 199: Don't need an explicit type (same could be done for the `InviteUserReleaseRoleForm` above as well.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/UserPublicationInvite.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        [Required] public Publication Publication { get; set; } = null!;
+
+        [Required]
+        public Guid PublicationId { get; set; }
+        
+        [Required]
+        public PublicationRole Role { get; set; }
+        
+        public bool Accepted { get; set; }
```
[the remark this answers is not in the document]
markhiveit: > Do we need this? I know we set it for UserReleaseInvites but UserInvite has its own Accepted field which should be all we need. Tbh, even that field is a little redundant because you could just check if a user exists on the system or not to determine if the invite was accepted.

Investigating a little, it seems like the original intention of `Accepted` on `UserReleaseInvites` was that an existing user could be invited to have a new release role - an email is also sent when this happens. Not sure if `UserReleaseInvites` are still used in this case - if the user already exists on the system, then a new `UserReleaseRole` can be created immediately - I believe that's what happens with the `ContributorInvite` stuff. Looking at the code, I don't see where `UserReleaseInvites` get processed if the user already exists. And `UserRoleService#AddReleaseRole` instantly creates the role - it doesn't create a UserReleaseInvite.

So I'm going to go ahead and delete `Accepted` from `UserReleaseInvites` and `UserPublicationInvites`, and anywhere `Accepted` was set, I'll instead remove the release/publication invite entirely.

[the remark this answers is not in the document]
markhiveit: `ReleaseInviteService#CreateExistingUserContributorInvite` was creating accepted `UserReleaseInvites` after creating the `UserReleaseRoles` themselves - so it looks like it was being used for auditing. Anyway, I'm going to continue and remove the `Accepted` columns.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/UserPublicationInvite.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public bool EmailSent { get; set; }
+
+        public DateTime Created { get; set; }
```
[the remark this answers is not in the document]
markhiveit: Also added this to `UserReleaseInvite`.

I think it makes sense to keep UserReleaseInvite and UserPublicationInvite aligned in case we create an abstract class `UserResourceInvite` in the future, much like Dunc's `ResourceRole`.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/UserPublicationInvite.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public User CreatedBy { get; set; } = null!;
+
+        public Guid CreatedById { get; set; }
+        
+        public bool SoftDeleted { get; set; }
```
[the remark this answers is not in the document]
markhiveit: I've removed it. A little torn about whether to keep it because it seems worthwhile to keep `UserReleaseInvite` and `UserPublicationInvite` aligned, but this difference doesn't seem like a big deal.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/UserPublicationInvite.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        [Required]
+        public Guid PublicationId { get; set; }
+        
+        [Required]
+        public PublicationRole Role { get; set; }
+        
+        public bool Accepted { get; set; }
+
+        public bool EmailSent { get; set; }
```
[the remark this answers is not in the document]
markhiveit: I've removed it.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/UserPublicationInvite.cs
markhiveit: You mean the creation of the `UserPublicationRole`s from the `UserPublicationInvites` when a newly invited user first logs in? That code is in `ExternalLogin.cshtml.cs` in method `HandleNewInvitedUser`.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/UserPublicationInvite.cs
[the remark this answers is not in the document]
markhiveit: I've removed `Updated` from `UserPublicationInvite`, because it wasn't being used.

ntsim (approved):
Looks good :+1:

PR #3330, opened 2022-06-14
EES-3017 table tool publication step changes

amyb-hiveit:
The DAC report raised accessibility issues with the current 'Choose a publication' step of the table tool. To fix these we've changed the component to not use the Details component.

Instead:
- shows a list of themes as radio buttons
- on selecting a theme the list of publications for the theme is shown as radio buttons
- topics aren't shown
- searching populates the list of publications with those that match the search (a basic text match as currently)

This change also applies to the Data Catalogue.

step1

### DIFF
--- a/src/explore-education-statistics-common/src/components/form/FormTextSearchInput.tsx
+++ b/src/explore-education-statistics-common/src/components/form/FormTextSearchInput.tsx
@@ -1,9 +1,9 @@
 import { FormTextInputProps } from '@common/components/form/FormTextInput';
 import { FormTextInput } from '@common/components/form/index';
+import styles from '@common/components/form/FormTextSearchInput.module.scss';
 import classNames from 'classnames';
 import debounce from 'lodash/debounce';
-import React, { ChangeEvent } from 'react';
-import styles from './FormTextSearchInput.module.scss';
+import React, { ChangeEvent, useEffect, useState } from 'react';
 
 interface Props extends FormTextInputProps {
   debounce?: number;
@@ -15,6 +15,12 @@ const FormTextSearchInput = ({
   onChange,
   ...props
 }: Props) => {
+  const [value, setValue] = useState<string>(props.value ?? '');
+
+  useEffect(() => {
+    setValue(props.value ?? '');
+  }, [props.value]);
+
   const handleChange = debounce((event: ChangeEvent<HTMLInputElement>) => {
     if (onChange) {
       onChange(event);
@@ -25,8 +31,10 @@ const FormTextSearchInput = ({
     <FormTextInput
       {...props}
       className={classNames(className, styles.searchInput)}
+      value={value}
       onChange={event => {
         event.persist();
+        setValue(event.target.value);
         handleChange(event);
       }}
     />
--- a/src/explore-education-statistics-common/src/modules/table-tool/components/PublicationForm.tsx
+++ b/src/explore-education-statistics-common/src/modules/table-tool/components/PublicationForm.tsx
@@ -1,22 +1,23 @@
-import DetailsMenu from '@common/components/DetailsMenu';
 import {
   Form,
   FormFieldRadioGroup,
   FormFieldset,
   FormGroup,
+  FormRadioGroup,
   FormTextSearchInput,
 } from '@common/components/form';
 import SummaryList from '@common/components/SummaryList';
 import SummaryListItem from '@common/components/SummaryListItem';
 import WizardStepSummary from '@common/modules/table-tool/components/WizardStepSummary';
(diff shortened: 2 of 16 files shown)


src/explore-education-statistics-common/src/modules/table-tool/components/PublicationForm.tsx
```diff
@@ -78,142 +115,131 @@ const PublicationForm = ({
...
+                      id={`${formId}-publications`}
+                      legend={
+                        <>
+                          Select a publication
+                          <span
+                            className="govuk-visually-hidden"
+                            aria-live="polite"
+                            aria-atomic="true"
+                          >
+                            {' '}
```
ntsim, line 189: This space can just be added to the front of the adjacent template string to avoid splitting the node up:

```ts
` ${publications.length} ...`
```

src/explore-education-statistics-common/src/modules/table-tool/components/PublicationForm.tsx
```diff
@@ -30,29 +31,68 @@ const formId = 'publicationForm';
...
 }: Props) => {
   const { isActive, goToNextStep } = stepProps;
 
   const [searchTerm, setSearchTerm] = useState('');
-  const lowercaseSearchTerm = searchTerm.toLowerCase();
+  const [selectedThemeId, setSelectedThemeId] = useState<string>('');
+
+  const publications = useMemo(() => {
+    if (selectedThemeId) {
+      setSearchTerm('');
```
ntsim, line 53: Do we need to clear this here? State changes like this would be better within the event handler where the change happens and we already are doing this.

src/explore-education-statistics-common/src/modules/table-tool/components/PublicationForm.tsx
```diff
@@ -30,29 +31,68 @@ const formId = 'publicationForm';
...
+      return (
+        themes
+          .find(theme => theme.id === selectedThemeId)
+          ?.topics.flatMap(topic => topic.publications) ?? []
+      );
+    }
+    if (searchTerm) {
+      setSelectedThemeId('');
```
ntsim, line 61: Do we need to clear this here? State changes like this would be better within the event handler where the change happens and we already are doing this.

src/explore-education-statistics-common/src/modules/table-tool/components/PublicationForm.tsx
```diff
@@ -78,142 +115,131 @@ const PublicationForm = ({
...
+                    })}
+                    value={selectedThemeId}
+                    onChange={e => {
+                      setSelectedThemeId(e.target.value);
+                      form.setFieldValue('publicationId', '');
+                      setSearchTerm('');
+                    }}
+                  />
+
+                  <div className={styles.publicationsList} id="publications">
```
ntsim, line 178: This doesn't seem like a very safe id. Couldn't the skip link use the `#${formId}-publications` to jump to the radio group?

src/explore-education-statistics-common/src/modules/table-tool/components/PublicationForm.tsx
```diff
@@ -78,142 +115,131 @@ const PublicationForm = ({
...
+                      )}
+                    />
+
+                    {!publications.length && (
+                      <>
+                        <p>Search or select a theme to view publications</p>
+                        {(searchTerm || selectedThemeId) && (
+                          <p>
+                            <em>No publications found</em>
+                          </p>
```
ntsim, line 221: Italics are a little horrible because they aren't really supported by the font. Could use `InsetText` instead:

image

ntsim (changes requested):

src/explore-education-statistics-common/src/components/form/FormTextSearchInput.tsx
```diff
@@ -15,6 +15,12 @@ const FormTextSearchInput = ({
   onChange,
   ...props
 }: Props) => {
+  const [value, setValue] = useState<string>(props.value ?? '');
+
+  useEffect(() => {
+    setValue(props.value ?? '');
+  }, [props.value]);
```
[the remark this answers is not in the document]
amyb-hiveit: It's to be able to clear the field when a theme is selected, I couldn't get it to work otherwise but maybe I missed something?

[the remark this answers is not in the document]
ntsim: Cool fair enough! Think if we remove this, the input also seems to stop working!

ntsim (approved):

PR #3334, opened 2022-06-17
EES-3459 - bringing test users in Keycloak up to date, adding additio…

duncan-at-hiveit:
This PR:
- brings our Keycloak users up to date with our new test users
- adds convenience methods and more README instructions

On this branch you should be able to run:

```
cd useful-scripts
./run.js idp adminKeycloak
```

The log into Admin with a standard seed user email address and the password `password`.

The flow should work as shown here: https://www.awesomescreenshot.com/video/9490155?key=4140f4b7bf0dd91ebaeda18943051f78


src/keycloak/keycloak-ees-realm.json
```diff
@@ -492,7 +492,7 @@
     "emailVerified" : true,
     "firstName" : "Prerelease1",
     "lastName" : "User1",
-    "email" : "ees-prerelease1@education.gov.uk",
+    "email" : "ees-test.prerelease1@education.gov.uk",
```
luke-h1, line 495: At the moment we haven't migrated to DFE-based test pre-release users yet. I believe any pre-release users can stay the same (in order to avoid failures in the UI tests)

luke-h1 (changes requested):

src/keycloak/keycloak-ees-realm.json
duncan-at-hiveit: Done, ta!

luke-h1:
Looks good. I think we also need to rename analyst users to:
```
ees-test.analyst1@education.gov.uk
ees-test.analyst2@ezducation.gov.uk
ees-test.analyst3@ezducation.gov.uk
```

luke-h1 (approved):

src/keycloak/keycloak-ees-realm.json
```diff
@@ -448,7 +448,7 @@
     "emailVerified" : true,
     "firstName" : "Bau1",
     "lastName" : "User1",
-    "email" : "ees-bau1@education.gov.uk",
+    "email" : "ees-test.bau1@education.gov.uk",
```
markhiveit, line 451: Just out of interest, I think this config here is setting up these users? If so, I assume that `secretData` is basically where the password is set?

I ask because I can see a potential pain point of trying to get passwords aligned between AAD and KeyCloak - maybe if someone does change the UI tests so they work with KeyCloak, might be worth to create separate environment variables to store the passwords used by KeyCloak, so people don#'t need to change their `.env` files if they're switching between AAD and KeyCloak. Maybe another solution is they could also specify a separate `.env` file for KeyCloak

luke-h1: Not too sure on this but I'm planning to load `.env.keycloak`  in the UI tests if a user wants to use keycloak as an identity provider

PR #3348, opened 2022-06-24
EES-3478 - remove login.robot test

luke-h1:


benoutram (approved):
Removed since we won't be needing to test that these users can log in on Prod any longer

PR #3351, opened 2022-06-26
Ees 3372 add import tests to k6 suite

duncan-at-hiveit:
:warning: This PR requires an Infrastructure deploy to get the Open ID Connect refresh token config out to environments :warning: 

## Overview
This PR:
* adds new "data file import" and "table builder query" performance tests to the performance test suite, focusing on the Admin app
* adds a new Dashboard to Grafana tailored around custom metrics captured in these new tests
* adds the support of refresh tokens to Local, Dev and Test environments allowing extended load testing without the need to continuously log in.  This also enables the silent login flow for the Admin app, so it might be worth rolling out to Pre Prod and Prod at some point after we're happy that it's not causing any unforeseen issues.
* updates Keycloak's settings to allow redirects back to both "localhost" and "ees.local"

## New tests

There are 2 new performance tests, `import.test.ts` and `tableBuilderQuery.test.ts` that both target Admin.  They both look for data that they need in order to run.  If it exists already it uses it.  If it doesn't, it'll firstly set it up before the Virtual Users (VUs) begin running.

These tests both capture custom metrics that are reported in Grafana under an EES-specific Dashboard:

image

Examples are the time it takes to reach each data file import stage, and the time it takes to complete a table query.

### Puppeteer and K6 tests

In order to support the authentication needed for these Admin tests, I have adopted and amended Luke's Puppeteer script to work with both Azure AD and Keycloak.  This is run prior to any tests, and the output captured in an environment variable.  Tests are then run that use the auth tokens in that environment variable to get started.  They then use refresh tokens to obtain fresh access tokens so that they don't expire.

This can be improved.  Currently due to time restrictions, each VU starts its job by grabbing a fresh access token, ensuring that they're always fresh.  This would better be moved into `httpClient.ts`, which'll grab fresh tokens as and when they are due to expire / have expired.

Also, it might be nice to store the initial access / refresh tokens from the login script in a file, like `.access-tokens.local` rather than an environment variable, with an expiry so that they can be used across multiple terminal sessions - at the moment each Bash session for instance would need to sign in to set its own environment variables.

### Docker access to localhost from K6 / Puppeteer

In order for K6 and Puppeteer to access the host machine from within a Docker container, we add `ees.local` as a host entry in the containers' hosts files.  In order to allow access from non-Localhost network interfaces, we need to start up Admin, Data and Content APIs on `0.0.0.0`, not `127.0.0.1`.
  
### Updates to Admin API's Open ID Connect setup

In order to set up support for refresh tokens, the SPA's Open ID Connect client needs to be granted the "offline_access" scope.  We enable this with the `AllowOfflineAccess` flag, and various other setup parameters.  In order to allow several VUs to spawn off their own access tokens from a single initial access token and refresh token, we allow the `ReUse` TokenUsage config setting on Dev and Test only for the time being, these being our candidate environments for load testing.  We use the `Sliding` window TokenExpiration setting to give us extra assurance that we can use our refresh tokens during a long test run, where perhaps certain individual VU jobs might take a very long time to run.

The ARM templates have been updated to support this new config.  Note that refresh tokens have been disabled completely for Prod right now, but there's no reason not to enable in the future, just without the ReUse and Sliding options!

Again, this can all be improved with more time :)

### Possible security fix in Startup.cs

Previously, we were disabling Token Issuer validation in non-local environments.  I think that this was a mistake, and so have flipped it.  Now we disable Token Issuer validation *only* in local environments.  We need this to allow us to sign in successfully from `localhost` and from `ees.local`.  Note that I also tried to achieve the same thing by adding both entries to "ValidIssuers" as well, but it didn't work.  Might be something we can improve on though in the future.

### Keycloak config updates

I have added both `ees.local` and `localhost` as valid callees of Keycloak login, allowing us to run Admin with either hostname. 

## Typescript support

Originally we decided against pursuing Typescript support for the K6 tests, as we had time restrictions initially and also the benefit was not obvious with our first round of tests.  This PR introduces Typescript support again though, partially due to the fact that the Puppeteer scripts were Typescript and it seemed good to keep it like tihs, and partially because there was now a good K6 / Typescript template out there courtesy of the K6 team.  With the more complex K6 tests now being added to the suite, it seems good to keep it as it definitely adds benefit in these cases.  

## Test results

Test results showed that the read replica appeared to be holding up well during file imports!  Details on EESthe -3372 ticket itself.

## Robot tests

image


tests/performance-tests/Dockerfile-node-chromium
```diff
@@ -0,0 +1,7 @@
+FROM node:18
```
luke-h1, line 1:
```suggestion
FROM node:16
```

tests/performance-tests/src/auth/getAuthDetails.ts
```diff
@@ -0,0 +1,166 @@
...
+  id_token: string;
+  access_token: string;
+  refresh_token: string;
+  expires_at: Date;
+}> => {
+  const browser = await puppeteer.launch({
+    headless: true,
+    ignoreHTTPSErrors: true,
+    product: 'chrome',
+    executablePath: '/usr/bin/google-chrome',
```
luke-h1, line 36: Would suggest creating a function to get the chrome executable path to cater for different environments i.e: 

```typescript
import os from 'os';


function getChromePath(): string {
  const platform = os.platform();

  switch(platform) {
    case 'win32':
      return 'C:\\Program Files (x86)\\Google\\Chrome\\Application\\chrome.exe';

    case 'darwin':
      return '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome';

    default:
      return '/usr/bin/google-chrome';
  }
};

```

tests/performance-tests/src/auth/getAuthDetails.ts
```diff
@@ -0,0 +1,166 @@
...
+      case 'keycloak':
+        await getAuthTokensKeycloak(page, email, password);
+        break;
+      default:
+        throw new Error(`No login strategy for IDP ${idp}`);
+    }
+
+    return await page.evaluate(() => {
+      /* eslint-disable-next-line no-plusplus */
+      for (let i = 0; i < localStorage.length; i++) {
```
luke-h1, line 64: This should get rid of the lint error: `/* eslint-disable-next-line no-plusplus */`

```suggestion
      for (let i = 0; i < localStorage.length; i+= 1) {
```

tests/performance-tests/src/utils/dataService.ts
```diff
@@ -0,0 +1,511 @@
...
+  locations: {
+    [geographicLevel: string]: {
+      options: {
+        id: string;
+      }[];
+    };
+  };
+}
+
+type Topic = {
```
luke-h1, line 44: Just wondering why you've used `type` here & not `interface`? Typically for objects, I think `interface` is the preferred convention and `type` for things such as unions or intersections

tests/performance-tests/src/utils/dataService.ts
```diff
@@ -0,0 +1,511 @@
...
+
+type PublicationAndReleases = {
+  id: string;
+  topicId: string;
+  title: string;
+  releases: Release[];
+};
+
+class DataService {
+  client;
```
luke-h1, line 69: Would suggest narrowing the type of client here like so (not sure if it should be public or private in this case): 
```suggestion
  private readonly client;
```

tests/performance-tests/src/utils/dataService.ts
```diff
@@ -0,0 +1,511 @@
...
+
+    return {
+      id: json.id,
+      response,
+    };
+  }
+
+  getOrCreateTheme(params: { title: string }) {
+    return {
+      id: this.getTheme(params)?.id ?? this.createTheme(params).id,
```
luke-h1, line 113: Nice 😎

tests/performance-tests/src/utils/httpClient.ts
```diff
@@ -0,0 +1,94 @@
+import http, { RefinedParams, RefinedResponse, RequestBody } from 'k6/http';
+
+type HttpHeaders = { [name: string]: string };
```
luke-h1, line 3: Would suggest using an interface here instead of `type`

tests/performance-tests/tsconfig.json
```diff
@@ -0,0 +1,29 @@
+{
+  "compilerOptions": {
+    "target": "es2017",
+    "moduleResolution": "node",
+    "module": "commonjs",
+    "noEmit": true,
+    "allowJs": true,
+    "removeComments": false,
+
```
luke-h1, line 9: Just wondering why there are spaces in this file?

luke-h1 (changes requested):

luke-h1:
Overall it looks really good. I would suggest getting rid of the old tests since they are out of date and Webpack won't move these old tests into the dist folder

tests/performance-tests/Dockerfile-node-chromium
duncan-at-hiveit: Sure thing!  I guess this aligns with other Node versions we use elsewhere?

tests/performance-tests/.babelrc
```diff
@@ -0,0 +1,22 @@
+{
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "targets": {
+          "node": "8"
```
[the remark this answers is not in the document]
duncan-at-hiveit: Hehe, probably a remnant of me desperately trying to get TS to work with the K6 tests after bringing in the Puppeteer scripts.  Have taken it out entirely and it seems to work fine, thanks!

tests/performance-tests/src/auth/getAuthDetails.ts
duncan-at-hiveit: This is actually the path that Chrome lives at inside the Docker container in which this script runs, but it's always possible that we wanna run this directly on our host machines, in which case the above would be useful, so I'm be happy to put it in!  Thanks for providing the snippet.

duncan-at-hiveit: Just switched "/usr/bin/chrome" for "/usr/bin/google-chrome" as that was the path in the container and on my local machine too.

tests/performance-tests/src/auth/getAuthDetails.ts
duncan-at-hiveit: Thanks, done

tests/performance-tests/src/auth/refreshAuthTokens.ts
```diff
@@ -0,0 +1,60 @@
...
+    return undefined;
+  }
+
+  /* eslint-disable camelcase */
+  const authTokens = (response.json() as unknown) as {
+    access_token: string;
+    refresh_token: string;
+    expires_in: Date;
+  };
+  /* eslint-enable camelcase */
```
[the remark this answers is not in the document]
duncan-at-hiveit: It's to re-enable the rule past this point in the code. So we disable it for a few lines, and then re-enable it.

tests/performance-tests/Dockerfile-node-chromium
luke-h1: Yeah, that's right. Also since Node 18 is a bit of a high version to be on at the moment (the majority of things are a bit early to be adopted).

tests/performance-tests/src/auth/getAuthDetails.ts
luke-h1: Ah right ok cool, keep forgetting this stuff runs purely through docker now

tests/performance-tests/src/utils/dataService.ts
duncan-at-hiveit: Yup good shout.  I've also extracted the contents of httpClient.ts into an HttpClient class, and assigned `client` here the appropriate type too.

tests/performance-tests/src/utils/dataService.ts
duncan-at-hiveit: Ta :smile:

tests/performance-tests/src/utils/httpClient.ts
duncan-at-hiveit: Done here, and also for all the `types` in dataService.ts too.  Ta.

tests/performance-tests/tsconfig.json
duncan-at-hiveit: I think that's just how it came from the K6 / TS template project - I've removed them now, thanks!

tests/performance-tests/src/tests/testData.ts
```diff
@@ -0,0 +1,10 @@
+const testData = {
+  themeName: 'UI test theme - Performance tests',
+  topicName: 'UI test topic - Performance tests',
+  maxImportWaitTimeMillis: 2 * 60 * 60 * 1000,
+};
+export default testData;
+  themeName: 'UI test theme - Performance tests',
```
luke-h1, line 7: Think the suggestion has gone a bit awry here

duncan-at-hiveit: Cheers, manually fixed!  Shame, it's a good idea for a mechanism - I may have used it incorrectly...

luke-h1 (approved):

infrastructure/templates/template.json
```diff
@@ -535,6 +535,29 @@
...
+    "openIdConnectSpaClientAllowOfflineAccess": {
+      "type": "bool",
+      "defaultValue": false,
+      "metadata": {
+        "description": "Whether or not this environment's Admin API supports refresh tokens"
+      }
+    },
+    "openIdConnectSpaClientRefreshTokenUsage": {
+      "type": "string",
+      "defaultValue": false,
```
benoutram, line 547: Should the default value here be `"OneTimeOnly"` rather than `false`?

```suggestion
      "defaultValue": "OneTimeOnly",
```

infrastructure/templates/template.json
```diff
@@ -535,6 +535,29 @@
...
+      "allowedValues": ["ReUse", "OneTimeOnly"],
+      "metadata": {
+        "description": "Whether or not this environment's Admin API supports reusable or one-time refresh tokens"
+      }
+    },
+    "openIdConnectSpaClientRefreshTokenExpiration": {
+      "type": "string",
+      "defaultValue": false,
```
benoutram, line 555: Should the default value here be `"Absolute"` rather than `false`?

```suggestion
      "defaultValue": "Absolute",
```

src/GovUk.Education.ExploreEducationStatistics.Admin/appsettings.json
```diff
@@ -28,6 +28,11 @@
     "TokenValidationParameters.NameClaimType": "name",
     "TokenValidationParameters.RoleClaimType": "role"
   },
+  "OpenIdConnectSpaClient": {
+    "AllowOfflineAccess": true,
```
benoutram, line 32: Should `AllowOfflineAccess` here be `false`?

```suggestion
    "AllowOfflineAccess": false,
```

I was going to say maybe don't even include this block here so there's no confusion over whether these settings might be taking effect over the settings defined in the App Service configuration in Azure, but I guess you've defined them here to provide a reference of all the possible settings?

infrastructure/templates/template.json
duncan-at-hiveit: Right you are!  Good spot, thanks!

infrastructure/templates/template.json
duncan-at-hiveit: Amended this one too

src/GovUk.Education.ExploreEducationStatistics.Admin/appsettings.json
duncan-at-hiveit: Not really, so happy to remove :)  

Blatted.  I'm always confused by this file.  Presumably Dev uses some of this if it doesn't have override values specified in the ARM template?  Always confused as to why we have the Hive AD settings in this file rather than just appsettings.Development.json but always assumed that this was just inherited by Dev and that was the onlyreason to have it in here?

benoutram: Yes that's correct, the settings in this `appsettings.json` file will take effect unless they are overridden by environment variables in the app service, (which are maintained by the infrastructure build through the ARM template).

In terms of precedence:

* command line options (highest)
* environment variables
* appsettings.{Environment}.json
* appsettings.json (lowest)

The three built-in environment names are `Development`, `Staging` and `Production`. All of our Azure environments run in `Production` and our local environment runs in `Development`.

So in our local environment, `appsettings.Development.json` has precedence over `appsettings.json`.

duncan-at-hiveit: Great info, thanks for the reply!

benoutram (approved):

PR #3357, opened 2022-06-28
EES-3483 Create UserReleaseInvite for existing users to ensure prerel…

markhiveit:
…ease email is sent on release approval

This fixes a bug introduced in EES-3387. We stopped creating `UserReleaseInvites` for existing prerelease users. This caused an issue when an existing user was invited to a release that wasn't `Approved`. Because they had no `UserReleaseInvite`, not prerelease invite email would be sent to them.

This fix creates a `UserReleaseInvite` for existing users when they're invited to prerelease.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PreReleaseUserService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PreReleaseUserService.cs
@@ -342,6 +342,13 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
                     }
                 }
 
+                await _userReleaseInviteRepository.Create(
+                    releaseId: release.Id,
+                    email: email,
+                    releaseRole: PrereleaseViewer,
+                    emailSent: sendEmail,
+                    createdById: _userService.GetUserId());
+
                 await _userReleaseRoleRepository.CreateIfNotExists(
                     userId: user.Id,
                     releaseId: release.Id,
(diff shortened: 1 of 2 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PreReleaseUserService.cs
```diff
@@ -342,6 +342,13 @@ await validEmails
                     }
                 }
 
+                await _userReleaseInviteRepository.Create(
+                    releaseId: release.Id,
+                    email: email,
+                    releaseRole: PrereleaseViewer,
+                    emailSent: sendEmail,
+                    createdById: _userService.GetUserId());
+
```
benoutram, line 351: No issue with this path for inviting existing users but just wondering if there's still potential for a bug though on the new user path:

1. Analyst invites a new user to the PreRelease of a Draft Release by e-mail address. E-mail not sent because the Release is still draft.
2.  UserReleaseInvite is created as per existing code since they aren't an existing user.
3. Same e-mail address gets invited to be a Contributor of Release or Publication.
4. New user logs in for the first time.
5. All UserReleaseInvites including the PreRelease invite get removed. See ExternalLogin.cshtml.cs:309-321 where the invites are removed where new user invites are removed.
6. Release with the PreRelease list is eventually approved and PreRelease. PreRelease e-mails are usually sent at this point.
7. Email to the user is not sent because UserReleaseInvite does not exist and invites are only sent on approval based on the list of UserReleaseInvites.

markhiveit: Discussed this with Ben on Slack. We thought that since this bug isn't that likely, and we're not sure how to approach this, we thought we would continue to deploy this hotfix without fixing the above issue, and instead decide how to deal with that in a meeting when all the devs are around. I'll create a ticket to address the above bug.

benoutram (approved):

PR #3359, opened 2022-06-29
EES-3453 Fix ReleaseDataAccordion styling issue

markhiveit:
This fixes a styling issue in EES-3453 where a HR dividing line wasn't being displayed between the "Explore data and files" and "Related dashboard(s)" accordion sections

image

### DIFF
--- a/src/explore-education-statistics-common/src/modules/release/components/ReleaseDataAccordion.module.scss
+++ b/src/explore-education-statistics-common/src/modules/release/components/ReleaseDataAccordion.module.scss
@@ -8,11 +8,13 @@
     margin-top: 0;
   }
 
-  @include govuk-media-query($from: tablet) {
+  /* stylelint-disable max-nesting-depth */
+  :global(.govuk-accordion__section:is(:first-child)) {
     :global(.govuk-accordion__section-header) {
       border-top: 0;
     }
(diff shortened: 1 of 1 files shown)


src/explore-education-statistics-common/src/modules/release/components/ReleaseDataAccordion.module.scss
```diff
@@ -3,16 +3,6 @@
...
-
-  @include govuk-media-query($until: tablet) {
-    margin-top: 0;
-  }
-
-  @include govuk-media-query($from: tablet) {
-    :global(.govuk-accordion__section-header) {
-      border-top: 0;
-    }
-  }
```
amyb-hiveit, line 15: These styles are needed for spacing on small screens and so it doesn't look like there's a double width bottom border on the summary box on large screens.

The margin one can safely be added back, I think, and you can change the border one so it only targets the first header with something convoluted like:
```
:global(.govuk-accordion__section:is(:first-child)) {
    :global(.govuk-accordion__section-header) {
      border-top: 0;
    }
  }
```

amyb-hiveit (changes requested):

amyb-hiveit (approved):

PR #3360, opened 2022-06-30
Updating UI snapshots

markhiveit:


markhiveit:
Closed this. This change occurred because the new release didn't have any subjects attached - the data was either linked to or provided in ancillary files. The fix will be that an amendment will be made to add a subject to the release - so no changes to the snapshot should be required.

PR #3365, opened 2022-07-01
EES-3489 EES-3491 Changes to invites

benoutram:
This PR makes a couple of minor changes to invites:

1. When inviting an existing user to Pre-release, only create a `UserReleaseInvite` if the release is not approved, i.e. when the e-mail is not being sent immediately.

 An invite has to exist in order for the e-mail to be sent at a later date which is why we create it, but it's not necessary to create one if the release is already approved and the e-mail is being sent at the same time.

2. When a new user is logging in for the first time, don't delete their `UserReleaseInvite`'s and `UserPublicationInvite`'s.

The reason for removing the deletion of `UserReleaseInvites`'s is that there can be invites where e-mails haven't been sent yet, i.e. in the case of Pre-release invites where e-mails are not sent until a release is approved.

If _all_ invites  are deleted when the user first logs in it will include deleting invites which aren't yet 'active' yet like the Pre-release invites where e-mails haven't been sent yet. This results in e-mails potentially not being sent.

Here's an example of how sending an e-mail could be missed at present:

* Invite a new user by e-mail address to the service. This could be as a contributor to a release or pre-release user on a release, or some other means but do it in a way that they are sent an e-mail.
* Invite the same new user to the Pre-release of a **draft** release, so they are not sent an e-mail just yet.
* User follows the link in the e-mail from the initial invite and logs in to the service for the first time.
* UserReleaseInvite for the Pre-release role on the draft release is deleted when the login is processed.
* Approve the draft release. Pre-release e-mails would normally be sent out at this point, where invites exist.
* No e-mail gets sent to the user because the invite has been deleted.

This scenario seems most likely to occur where a new user is invited to the Pre-release of multiple draft releases being prepared for publishing at a similar time.

There's another benefit to keeping the invites as well which is an audit trail of when the user was invited and who by, which is why I've also made the same change to stop deleting `UserPublicationInvite`'s.

### Other changes
- Rename UserInviteRepository `Create` methods to `CreateIfNotExists` to reflect behaviour of that method.
- Add tests to `ExternalLoginTests` to make sure that `UserPublicationRole`'s are being created for invites.

### UI test report

image

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Areas/Identity/Pages/Account/ExternalLoginTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Areas/Identity/Pages/Account/ExternalLoginTests.cs
@@ -51,21 +51,21 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Areas.Identity.
             Assert.Null(loginService.ErrorMessage);
             Assert.Empty(loginService.ModelState);
         }
-        
+
         [Fact]
         public async Task Login_ExistingUser_Failed()
         {
             var (result, loginService) = await DoExistingUserLogin(SignInResult.Failed);
             AssertRedirectedToLoginPageUponFailure(result, loginService);
         }
-        
+
         [Fact]
         public async Task Login_ExistingUser_NotAllowed()
         {
             var (result, loginService) = await DoExistingUserLogin(SignInResult.NotAllowed);
             AssertRedirectedToLoginPageUponFailure(result, loginService);
         }
-        
+
         [Fact]
         public async Task Login_ExistingUser_LockedOut()
         {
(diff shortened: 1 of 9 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Areas/Identity/Pages/Account/ExternalLoginTests.cs
```diff
@@ -329,21 +346,6 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contextId))
                     .AsQueryable()
                     .SingleAsync(invite => invite.Email == email);
                 Assert.True(newUsersInvite.Accepted);
-
```
duncan-at-hiveit, line 332: Why remove this check out of interest?  It's here to make sure that the processed invites are limited to those of the currently logging-in user.

duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Areas/Identity/Pages/Account/ExternalLoginTests.cs
benoutram: Good spot, I think I intended to delete a block below it and made a mistake. I've reinstated it now.

I've also added some new checks to make sure the release role invites and publication role invites are left alone.

PR #3368, opened 2022-07-06
EES-3512 Limit number of 'CreateIfNotExists' calls to Azure Storage

markhiveit:
This PR ensures that `CreateIfNotExists` calls to Azure Storage only happen once per container/queue/table each time a service is started. See EES-3512 for why I chose this method over creating all containers/queues/tables when services start.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Services/BlobStorageServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Services/BlobStorageServiceTests.cs
@@ -11,6 +11,7 @@ using Azure.Storage.Blobs.Models;
 using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Services;
 using GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils;
+using GovUk.Education.ExploreEducationStatistics.Common.Utils.Interfaces;
 using Microsoft.Extensions.Logging;
 using Moq;
 using Newtonsoft.Json;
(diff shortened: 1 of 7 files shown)


src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Services/BlobStorageServiceTests.cs
```diff
@@ -615,7 +610,8 @@ public async Task DeleteBlobs_NestedReleaseBlobs_ExcludeRegex()
             return new BlobStorageService(
                 connectionString: "",
                 blobServiceClient ?? new Mock<BlobServiceClient>().Object,
-                new Mock<ILogger<BlobStorageService>>().Object
+                new Mock<ILogger<BlobStorageService>>().Object,
```
duncan-at-hiveit, line 613: Can use Mock.Of<BlobStorageService>() instead, when not needing to use the actual mock from any setups.

src/GovUk.Education.ExploreEducationStatistics.Common/Services/BlobStorageService.cs
```diff
@@ -40,12 +42,22 @@ public class BlobStorageService : IBlobStorageService
...
+            string connectionString,
+            BlobServiceClient client,
+            ILogger<IBlobStorageService> logger,
+            IStorageInstanceCreationUtil? storageInstanceCreationUtil = null)
         {
             _connectionString = connectionString;
             _client = client;
             _logger = logger;
+
+            if (storageInstanceCreationUtil != null)
```
duncan-at-hiveit, line 57: You could use the same syntax that we use in tests?

```
_storageInstanceCreationUtil = storageInstanceCreationUtil ?? new StorageInstanceCreationUtil()
```

Alternatively, you could set up DI to give you an instance of storageInstanceCreationUtil as per any other autowired dependency - it doesn't store any state (other than the static, so it's not unique to any instance of StorageInstanceCreationUtil, so it wouldn't matter if it's a singleton, a transient, or a new one for every wring event.

PR #3387, opened 2022-07-18
EES-3529 Remove usages of CheckBlobExists and GetBlob

benoutram:
⚠️ Depends on EES-3548 to secure methodology images.
⚠️ Depends on EES-3547/EES-3549 to migrate blob ContentLength and ContentType properties to database. See https://github.com/dfe-analytical-services/explore-education-statistics/pull/3389
⚠️ Has post-deploy cleanup task EES-3552.

This PR eliminates most usages of `BlobStorageService#CheckBlobExists` and `BlobStorageService#GetBlob`. Both require HTTP requests to the storage account. `GetBlob` doesn't return blob content but retrieves blob properties including any custom metadata.

Both are left in the codebase as they are still used for streaming a zip of files, and for caching the 'All files' zip file.

Blob properties which we can now store in the database:

- Content length
- Content type

Custom blob metadata which was previously used but can be fetched from elsewhere:

- Number of rows, only applicable to release data files.
- Release published datetime, only applicable to published files.

`ContentLength` property is used to create a formatted display label of the file size.
`ContentType` property is used to stream files in both admin and public release ancillary, chart, data and image files, and methodology images.
Number of rows is being used to display the rows in the Data and files tab of the Admin.
Release published datetime metadata is being used to secure the access to public methodology images.

By storing the file content type in the database when the file is uploaded then there's no need to get the blob properties before streaming it using a `FileStreamResult`.

Similarly by storing the content length in the database then there's no need to get the blob properties for a file or a list of files to create the file info response.

By retrieving the 'Rows' property from the `DataImport` related to a data file instead means there's no need to store and retrieve the number of rows from custom metadata on the blob.

Additionally we can make the following optimisations:

Uploading a blob was always followed by `GetBlob` in order to build a file info response for the newly uploaded file. Without Size or Rows this additional call is no longer required.

Before streaming a blob we were checking if it exists, following in the method call to stream a blob another call to check if it exists, throwing an exception if it didn't.
This meant two HTTP requests to check if the blob exists before opening the stream. By changing `BlobStorageService#DownloadToStream` to return a `NotFound` result if the blob doesn't exist, this is now reduced to one.

Securing methodology images can be achieved without using the release published datatime metadata and this will be done by EES-3548 in https://github.com/dfe-analytical-services/explore-education-statistics/pull/3402.

### Other changes:

- Remove lots of code which actively expected blobs corresponding with database file entries not to exist.
- Remove the 'SetAttributes' callback when copying files between private and public storage during publishing which set - metadata attributes now that it's no longer required.
- Remove unused userName parameter from ReleaseDataFileService Upload and UploadAsZip methods.
- Remove unused releaseId parameter from DataArchiveValidationService#ValidateDataArchiveFile
-Remove the metadata parameters from BlobStorageService UploadFile and UploadStream so that it's no longer possible to set blob metadata.
- Rename DataImportViewModel to DataImportStatusViewModel.
- Remove Size from BlobInfo in favour of ContentLength which is retained.
- Remove the eager loading of the Errors collection in DataImportRepository#GetByFileId and change it to be explicitly loaded in the one case where required in DataImportService#GetImportStatus.
- Add missing unit tests DataImportService#Import
- Change File to use ICreatedTimestamp interface for setting the created date.

### UI Test report

image


src/GovUk.Education.ExploreEducationStatistics.Data.Processor.Tests/Services/DataImportServiceTests.cs
```diff
@@ -148,7 +149,8 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
                 var updated = await contentDbContext.DataImports.FindAsync(import.Id);
                 Assert.NotNull(updated);
                 Assert.Equal(1000, updated.RowsPerBatch);
-                Assert.Equal(10000, updated.TotalRows);
+                Assert.Equal(10000, updated.Rows);
+                Assert.Equal(5000, updated.TotalRows);
```
ntsim, line 153: This seems a little confusing. These values don't seem to correspond well with the parameters `totalRowCount` and `importableRowCount` passed to `service.Update`.

Would expect that:
- `totalRowCount` => `TotalRows`
- `importableRowCount` => `Rows`

However, it seems that the reverse is true. Perhaps the service method needs to be updated to have more appropriate parameters.

It also seems strange that the total rows would be less than rows, shouldn't total rows always be larger? That said, I'm just generally confused what the difference between rows, total rows even is. The terminology is a little confusing and could use some clarification (perhaps in a docblock or comment).

src/explore-education-statistics-admin/src/pages/release/data/components/DataFileDetailsTable.tsx
```diff
@@ -113,33 +113,27 @@ const DataFileDetailsTable = ({
...
           <th scope="row">Data file size</th>
           <td data-testid="Data file size">
-            {!dataFile.isQueuedZipUpload && (
-              <>
-                {dataFile.fileSize.size.toLocaleString()}{' '}
-                {dataFile.fileSize.unit}
-              </>
-            )}
+            <>
+              {dataFile.fileSize.size.toLocaleString()} {dataFile.fileSize.unit}
```
ntsim, line 117: Would suggest making this a single template literal string:
```tsx
{`${dataFile.fileSize.size.toLocaleString()} ${dataFile.fileSize.unit}`}
```

It's a not a big difference, but if we don't do this, it can make it harder for tests to find specific pieces of text (particularly the Robot tests).

src/explore-education-statistics-admin/src/pages/release/data/components/DataFileDetailsTable.tsx
```diff
@@ -113,33 +113,27 @@ const DataFileDetailsTable = ({
...
             <td data-testid="Replacement Data file size">
-              {!replacementDataFile.isQueuedZipUpload && (
-                <>
-                  {replacementDataFile.fileSize.size.toLocaleString()}{' '}
-                  {replacementDataFile.fileSize.unit}
-                </>
-              )}
+              <>
+                {replacementDataFile.fileSize.size.toLocaleString()}{' '}
+                {replacementDataFile.fileSize.unit}
```
ntsim, line 124: Same as above - could use a template literal string here

src/explore-education-statistics-admin/src/pages/release/data/components/DataFileDetailsTable.tsx
```diff
@@ -113,33 +113,27 @@ const DataFileDetailsTable = ({
...
+              </>
             </td>
           )}
         </tr>
         <tr>
           <th scope="row">Number of rows</th>
           <td data-testid="Number of rows">
-            {!dataFile.isQueuedZipUpload && dataFile.rows.toLocaleString()}
+            {dataFile.rows?.toLocaleString()}
```
ntsim, line 132: Not a biggy, but if there are no detected rows, we could show something like 'Unknown'?

src/explore-education-statistics-admin/src/pages/release/data/components/DataFileDetailsTable.tsx
```diff
@@ -113,33 +113,27 @@ const DataFileDetailsTable = ({
...
           </td>
           {replacementDataFile && (
             <td data-testid="Replacement Number of rows">
-              {!replacementDataFile.isQueuedZipUpload &&
-                replacementDataFile.rows.toLocaleString()}
+              {replacementDataFile.rows?.toLocaleString()}
```
ntsim, line 136: Same as above - could show 'Unknown' if this is null

tests/robot-tests/tests/admin/bau/upload_files.robot
```diff
@@ -48,7 +48,7 @@ Upload a ZIP file subject
     user checks headed table body row contains    Data file    absence_in_prus.csv    ${section}
     user checks headed table body row contains    Metadata file    absence_in_prus.meta.csv    ${section}
     user checks headed table body row contains    Data file size    141 Kb    ${section}
-    user checks headed table body row contains    Number of rows    613    ${section}
+    user checks headed table body row contains    Number of rows    612    ${section}
```
ntsim, line 51: How come all of these row counts have shrunk by 1?

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Data.Processor.Tests/Services/DataImportServiceTests.cs
benoutram: That's correct, the reverse was true, and total rows was always less than rows which was the real total.

It really needed the underlying columns of `DataImport` renaming to correct this which I didn't want to do but I'm glad now you've raised it as it's justified the renaming a bit more!

https://github.com/dfe-analytical-services/explore-education-statistics/pull/3405 has renamed `TotalRows` -> `ImportedRows` and `Rows` -> `TotalRows`.

This means the service call now looks something like:

```
service.Update(import.Id,
    importedRows: 5000,
    totalRows: 10000,
    numBatches: 10,
    ...
);
```
and the assertions now correspond:

```
    Assert.Equal(5000, updated.ImportedRows);
    Assert.Equal(10000, updated.TotalRows);
````

File: src/GovUk.Education.ExploreEducationStatistics.Data.Services/Interfaces/IReleaseService.cs [hunk 7]
```diff
@@ -14,11 +13,5 @@ public interface IReleaseService
         Task<Either<ActionResult, List<SubjectViewModel>>> ListSubjects(Guid releaseId);
 
         Task<Either<ActionResult, List<FeaturedTableViewModel>>> ListFeaturedTables(Guid releaseId);
-
-        // TODO: EES-2343 Remove when file sizes are stored in database
```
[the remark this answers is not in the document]
benoutram: Looks like [https://dfedigital.atlassian.net/browse/EES-1810](https://dfedigital.atlassian.net/browse/EES-1810) can go as well!

I will close these when this work goes in, thanks.

I agree about the original ticket EES-2343 which looks like it was deleted.

src/explore-education-statistics-admin/src/pages/release/data/components/DataFileDetailsTable.tsx
benoutram: Thanks, made this change

src/explore-education-statistics-admin/src/pages/release/data/components/DataFileDetailsTable.tsx
benoutram: Thanks, made this change

tests/robot-tests/tests/admin/bau/upload_files.robot
benoutram: If you uploaded a non-zip data file in the Admin prior to this change, or uploaded a data zip and the importer extracted a data file from a zip archive, this dictionary was created and set on the blob metadata of the data file blob:

```
var dataInfo = GetDataFileMetaValues(
    metaFileName: metaFile.Filename,
    numberOfRows: CalculateNumberOfRows(dataFormFile.OpenReadStream())
);
```

`CalculateNumberOfRows` read the number of _lines_ in the file.

The same number was written to the `DataImport.Rows` field in the db which has now been renamed `TotalRows`.

Calculating the number of rows is now always deferred to 'Stage 1' of the importer so the Admin app is not potentially iterating over a huge number of rows just to count them.

DataImport.TotalRows is now set after validating the data file  in 'Stage 1' and it's the number of rows in the `DataRowCollection` object from the CSV, which doesn't include the header row.

Now 'TotalRows' and 'ImportedRows' should match each unless any rows are excluded from the import.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -311,13 +292,17 @@ await _contentDbContext.Entry(dataFile)
...
+                                                        contentType: zipFormFile.ContentType,
                                                         releaseId: releaseId,
                                                         createdById: _userService.GetUserId());
 
                                                     var dataFile = await _releaseDataFileRepository.Create(
                                                         releaseId: releaseId,
                                                         subjectId: subjectId,
                                                         filename: archiveFile.DataFileName,
+                                                        contentLength: archiveFile.DataFileSize,
+                                                        contentType: "text/csv",
```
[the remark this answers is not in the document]
benoutram: During the files migration except for a handful of cases all the files got a `ContentType` that's not `application/octet-stream` so I would say this was never an issue in Prod or it got fixed a long time ago. I'll recommend that EES-1460 is retested or just closed off once this work is complete.

Setting the content type for non-zip data files using `dataFormFile.ContentType`, `metaFormFile.ContentType`, `upload.File.ContentType` (ancillary files), and `formFile.ContentType` (chart files)  seems to have been working.

I opened [https://dfedigital.atlassian.net/browse/EES-3572](https://dfedigital.atlassian.net/browse/EES-3572)  and did a bit of investigation into the ones with `application/octet-stream`, and fixed those up in Prod. They look like they are large zip file uploads where we probably uploaded them during support requests manually using storage explorer and never set the content type property.

Interestingly that work spotted one chart not displaying correctly in the Admin. I've corrected the ContentType on the File in the db now so hopefully it will be fixed when this PR gets deployed.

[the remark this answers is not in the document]
benoutram: That sounds good as I can't think of a reason why the data file or meta file shouldn't be `test/csv`. I've made this change now, thanks.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -168,37 +168,32 @@ await ids.Select(id => _releaseFileRepository.CheckFileExists(releaseId, id, Fil
...
                 });
         }
 
         public async Task<Either<ActionResult, DataFileInfo>> Upload(Guid releaseId,
             IFormFile dataFormFile,
             IFormFile metaFormFile,
-            string userName,
             Guid? replacingFileId = null,
-            string subjectName = null)
+            string? subjectName = null)
```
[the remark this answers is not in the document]
benoutram: It's because when replacing a file the subject name is not provided and the subject name from the original is retrieved and used. @markhiveit has spotted the same issue in https://github.com/dfe-analytical-services/explore-education-statistics/pull/3408 and taken it one step further already, validating the subject name, so if's ok with you we could leave this here and review the proposed fix on that PR.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataImportService.cs
```diff
@@ -74,38 +76,42 @@ public async Task<bool> HasIncompleteImports(Guid releaseId)
...
+                .Collection(i => i.Errors)
+                .LoadAsync();
+
+            return new DataImportStatusViewModel
             {
                 Errors = import.Errors.Select(error => error.Message).ToList(),
                 PercentageComplete = import.PercentageComplete(),
                 StagePercentageComplete = import.StagePercentageComplete,
-                NumberOfRows = import.Rows,
+                NumberOfRows = import.Rows > 0 ? import.Rows : null,
```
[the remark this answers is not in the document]
benoutram: No never negative but I wanted to avoid returning a value of 0 as that's not strictly true, the importer just hasn't processed the data or zip file yet so we don't know the number of rows.

The value isn't set when the `DataImport` is created by the admin but the field isn't nullable so it's initialised as 0.

To display the number of rows as 'Unknown` rather than 0 and avoid the expression you've highlighted here I've made `TotalRows` nullable along with a few other changes to `DataImport`. I didn't change everything as I wanted to keep Importer changes to a minimum.

Changes are on [this commit](https://github.com/dfe-analytical-services/explore-education-statistics/pull/3387/commits/80c29ce1c22cc19f607af57ff647d76d9bd7213c).

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -168,37 +168,32 @@ await ids.Select(id => _releaseFileRepository.CheckFileExists(releaseId, id, Fil
...
-                    await filesExcludingReplacements
+                    return await filesExcludingReplacements
                         .ToAsyncEnumerable()
-                        .ForEachAwaitAsync(async file =>
-                            fileList.Add(await GetDataFileInfo(releaseId, file.File)));
-
-                    return fileList
-                        .OrderBy(file => file.Name)
-                        .AsEnumerable();
+                        .SelectAwait(async file => await GetDataFileInfo(releaseId, file.File))
```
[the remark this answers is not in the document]
benoutram: Cool, good suggestion. I've made this change in [this commit](https://github.com/dfe-analytical-services/explore-education-statistics/pull/3387/commits/30fe5e924a895555b5327975b542794e6aa86054).

I get the `DataImport` which we were already getting, and then use it to get the data file and the meta file, avoiding the old way of looking up the meta file which associated the two files together by subject id.

If it's a list of files then this happens in one query, although I used `AsSplitQuery` to avoid repetition in the result set.

It works out quite nicely because this change also removes a lookup on the subject to get the subject name which cuts out a bunch of setup in the unit tests.

However it highlights an inefficiency that still exists getting the permissions for each file using a auth handler.

We know the data import status so if we could work out if the user has permission to cancel any import (which Bau users can), then we wouldn't need to make repeated calls to an auth handler for each file.

That's probably not too difficult however I left this as a TODO for another day 🙂.

benoutram:
Updated UI test report after changes from PR feedback

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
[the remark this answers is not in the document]
ntsim: Nice one!

ntsim (approved):
Looks good!

PR #3389, opened 2022-07-18
EES-3547 Copy ContentLength and ContentType from blob to database to support EES-3529

benoutram:
⚠️ Has deploy task EES-3549.
⚠️ Has post-deploy cleanup task EES-3552.

This PR copies blob content type and length properties to the database to support https://github.com/dfe-analytical-services/explore-education-statistics/pull/3387.

It contains a database migration to add `ContentLength` and `ContentType` to `File`.

It contains a new secured Admin endpoint which updates all File entries found in the database, retrieves the corresponding blob from the correct storage container and sets the new `ContentLength` and `ContentType` properties based on `ContentLength` and `ContentType` blob properties.

Although migrations are traditionally run from the Admin app service, it might not be suitable to do this migration wholly inside a request/response cycle in an app service where a long running request could be terminated or a database connection could be reset.

The Publisher project has been used to host a temporary function that will dequeue messages for each file, retrieving the blob properties one file at a time.

This migration is rerunnable so it can be used at any time to 'catch-up' with new files that don't yet have `ContentLength` and `ContentType` values.

### Other changes

- Alters Content database File table column Filename to be not-null.

### Running the migration
Run the secured Admin endpoint `PATCH https://<admin-url>/api/bau/migrate-files` once to generate queue messages for all file that need migrating.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/FileMigrationServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/FileMigrationServiceTests.cs
@@ -54,21 +54,15 @@ public class FileMigrationServiceTests
     }
 
     [Fact]
-    public async Task MigrateAll_IgnoresFilesWithContentTypeOrSize()
+    public async Task MigrateAll_IgnoresFilesWithContentTypeAndSize()
     {
-        // Set up files that have content type, size, or both content type and size
+        // Set up files that already have values for ContentLength and ContentType so don't need migrating
         var files = new List<File>
         {
             new()
             {
                 Id = Guid.NewGuid(),
                 ContentType = "application/pdf",
-                ContentLength = null
-            },
-            new()
-            {
-                Id = Guid.NewGuid(),
-                ContentType = null,
                 ContentLength = 1024
             },
             new()
@@ -97,7 +91,67 @@ public class FileMigrationServiceTests
 
             var result = await service.MigrateAll();
 
-            // There should be no other interactions with the storage queue service
+            // There should be no further interactions with the storage queue service
+
+            VerifyAllMocks(storageQueueService);
+
+            result.AssertRight();
+        }
+    }
+
+    [Fact]
+    public async Task MigrateAll_MigratesPartiallyMigratedFilesWithContentTypeOrSize()
+    {
+        var files = new List<File>
+        {
+            // Partially migrated, missing ContentType
+            new()
+            {
+                Id = Guid.NewGuid(),
+                ContentLength = 1024,
+                ContentType = null
+            },
+            // Ignored as already migrated
+            new()
+            {
+                Id = Guid.NewGuid(),
+                ContentType = "application/pdf",
+                ContentLength = 1024
+            },
+            // Partially migrated, missing ContentLength
+            new()
+            {
+                Id = Guid.NewGuid(),
+                ContentLength = null,
+                ContentType = "application/pdf"
+            }
+        };
+
+        var contentDbContextId = Guid.NewGuid().ToString();
+
+        await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId))
+        {
+            await contentDbContext.Files.AddRangeAsync(files);
+            await contentDbContext.SaveChangesAsync();
+        }
+
+        var storageQueueService = MockStorageQueueService(messageCount: 0);
+
+        storageQueueService.Setup(mock =>
+                mock.AddMessages(MigrateFilesQueue, new List<MigrateFileMessage>
+                {
+                    new(files[0].Id),
+                    new(files[2].Id)
+                }))
+            .Returns(Task.CompletedTask);
+
+        await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId))
+        {
+            var service = SetupService(
+                contentDbContext: contentDbContext,
+                storageQueueService: storageQueueService.Object);
+
+            var result = await service.MigrateAll();
 
             VerifyAllMocks(storageQueueService);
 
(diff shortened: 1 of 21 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/FileMigrationServiceTests.cs
```diff
@@ -0,0 +1,188 @@
...
+
+        var service = SetupService(
+            contentDbContext: InMemoryContentDbContext(),
+            storageQueueService: storageQueueService.Object);
+
+        var result = await service.MigrateAll();
+
+        VerifyAllMocks(storageQueueService);
+
+        var left = result.AssertLeft();
```
duncan-at-hiveit, line 37: You can use:

```
result.AssertBadRequest();
```

instead

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/FileMigrationServiceTests.cs
```diff
@@ -0,0 +1,188 @@
...
+
+    [Fact]
+    public async Task MigrateAll_IgnoresFilesWithContentTypeOrSize()
+    {
+        // Set up files that have content type, size, or both content type and size
+        var files = new List<File>
+        {
+            new()
+            {
+                Id = Guid.NewGuid(),
```
duncan-at-hiveit, line 66: But do we not want to be able to fix up Files that have been partially migrated upon a rerun?  E.g. I'd assume that a File with only ContentType but no Size isn't valid, so wouldn't we want to revisit those on a subsequent run?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/FileMigrationServiceTests.cs
```diff
@@ -0,0 +1,188 @@
...
+        await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId))
+        {
+            await contentDbContext.Files.AddRangeAsync(files);
+            await contentDbContext.SaveChangesAsync();
+        }
+
+        var storageQueueService = MockStorageQueueService(messageCount: 0);
+
+        storageQueueService.Setup(mock =>
+                mock.GetApproximateMessageCount(MigrateFilesQueue))
```
duncan-at-hiveit, line 95: I think this Setup is already done within `MockStorageQueueService()`?

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20220713144101_EES3529_AddFileContentTypeAndSize.cs
```diff
@@ -0,0 +1,53 @@
...
+                nullable: false,
+                defaultValue: "",
+                oldClrType: typeof(string),
+                oldType: "nvarchar(max)",
+                oldNullable: true);
+
+            migrationBuilder.AddColumn<string>(
+                name: "ContentType",
+                table: "Files",
+                type: "nvarchar(max)",
```
duncan-at-hiveit, line 24: Not a biggie. but could *probably* limit this to nvarchar(255).  Happy to leave it though.

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20220713144101_EES3529_AddFileContentTypeAndSize.cs
```diff
@@ -0,0 +1,53 @@
...
+                nullable: true);
+
+            migrationBuilder.AddColumn<long>(
+                name: "Size",
```
duncan-at-hiveit, line 28: Any worth in being explicit and calling it SizeBytes or ContentLength?  I'd assume tbh that we'd all know it's bytes...

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/FileMigrationService.cs
```diff
@@ -0,0 +1,80 @@
...
+                            $"Unexpected null message count for queue {MigrateFilesQueue}");
+                    case > 0:
+                        return new BadRequestObjectResult(
+                            $"Found non-empty queue {MigrateFilesQueue}. Message count: {messageCount}");
+                }
+
+                // Queue one message per file for all files that don't have the new content type and size values
+                var files = await _contentDbContext.Files
+                    .AsNoTracking()
+                    .Where(file => file.ContentType == null && file.Size == null)
```
duncan-at-hiveit, line 60: I suspect this should be an "or", so that we can revisit Files that are only partially migrated.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/FileMigrationService.cs
```diff
@@ -0,0 +1,103 @@
...
+
+                return new Either<ActionResult, Unit>(Unit.Instance);
+            });
+    }
+
+    private async Task<IBlobContainer> GetPrivateContainer(File file)
+    {
+        return file.Type switch
+        {
+            FileType.Ancillary => BlobContainers.PrivateReleaseFiles,
```
duncan-at-hiveit, line 69: Not one for this PR, but I think if you have multiple switch cases that do the same thing, you can basically list the cases and then in the final one specify the behaviour.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/FileMigrationService.cs
benoutram: Thanks as mentioned in the other comment, this has been changed now.

PR #3400, opened 2022-07-22
EES-3567 Updating create_emulator_release_files.py after EES-3529

benoutram:
This PR makes an update to the `create_emulator_release_files.py` python script which we use to create blobs in the storage emulator for the local environment.

In future after EES-3529 the blobs only need to exist so that they can be zipped, or a successful download can be initiated for them.

All other properties about file uploads such as content type (used to open a stream to download them), content length, and  number of rows (applicable to data files) are now stored in either the `File` or `DataImport` in the database.

This PR:
* Removes setting metadata on blobs, meaning name, published date and number of rows is no longer relevant in the code too.
* Sets a content type on blobs.
* Stops uploading the 'All files' zip which is now generated on demand.

### DIFF
--- a/tests/robot-tests/tests/libs/create_emulator_release_files.py
+++ b/tests/robot-tests/tests/libs/create_emulator_release_files.py
@@ -1,53 +1,30 @@
 from abc import ABC, abstractmethod
 from azure.core.exceptions import ResourceExistsError
 from azure.storage.blob import BlobServiceClient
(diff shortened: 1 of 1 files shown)


markhiveit (approved):

benoutram:
Adding Do Not Merge until https://github.com/dfe-analytical-services/explore-education-statistics/pull/3387 is merged in

PR #3401, opened 2022-07-25
EES-3552 Cleanup to remove EES-3547 migration and make fields non nullable

benoutram:
⚠️ Depends on https://github.com/dfe-analytical-services/explore-education-statistics/pull/3387 which begins setting `ContentLength` and `ContentType` on all new files, and the migration https://github.com/dfe-analytical-services/explore-education-statistics/pull/3389 which needs to be run in all environments.

==================================================

This PR is a cleanup task to remove the migration introduced by EES-3547 in https://github.com/dfe-analytical-services/explore-education-statistics/pull/3389 that was run in the deploy task EES-3549.

Now that all Files have values for `ContentLength` and `ContentType` it also makes those fields non nullable and migrates them to be not-null in the database.


ntsim (approved):
My kinda PR :smile:

PR #3404, opened 2022-07-25
EES-3223 publication contact details

amyb-hiveit:
Adds the Contact page to the new Manage Publication section.

This work will be hidden to users until it's all ready, so for this you can only access the new page directly, at /publication//contact, there are no links to it yet.

This PR and https://github.com/dfe-analytical-services/explore-education-statistics/pull/3391 both make changes to PublicationPageContainer and publicationRoutes, when one is merged I'll rebase the other so all changes are kept.

contact1
contact2

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/publication/__tests__/PublicationContactPage.test.tsx
+++ b/src/explore-education-statistics-admin/src/pages/publication/__tests__/PublicationContactPage.test.tsx
@@ -51,32 +51,14 @@ describe('PublicationContactPage', () => {
       screen.getByText('Contact for this publication'),
     ).toBeInTheDocument();
 
-    expect(screen.getByTestId('Team name-key')).toHaveTextContent('Team name');
-    expect(screen.getByTestId('Team name-value')).toHaveTextContent(
-      'Team Smith',
-    );
-
-    expect(screen.getByTestId('Team email-key')).toHaveTextContent(
-      'Team email',
-    );
-    expect(screen.getByTestId('Team email-value')).toHaveTextContent(
+    expect(screen.getByTestId('Team name')).toHaveTextContent('Team Smith');
+    expect(screen.getByTestId('Team email')).toHaveTextContent(
       'john.smith@test.com',
     );
-
-    expect(screen.getByTestId('Contact name-key')).toHaveTextContent(
-      'Contact name',
-    );
-    expect(screen.getByTestId('Contact name-value')).toHaveTextContent(
-      'John Smith',
-    );
-
-    expect(screen.getByTestId('Contact telephone-key')).toHaveTextContent(
-      'Contact telephone',
-    );
-    expect(screen.getByTestId('Contact telephone-value')).toHaveTextContent(
+    expect(screen.getByTestId('Contact name')).toHaveTextContent('John Smith');
+    expect(screen.getByTestId('Contact telephone')).toHaveTextContent(
       '0777777777',
     );
-
     expect(
       screen.getByRole('button', { name: 'Edit contact details' }),
     ).toBeInTheDocument();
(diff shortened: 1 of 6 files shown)


src/explore-education-statistics-admin/src/pages/publication/__tests__/PublicationContactPage.test.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+    userEvent.click(
+      screen.getByRole('button', { name: 'Edit contact details' }),
+    );
+
+    expect(screen.getByLabelText('Team name')).toHaveValue('Team Smith');
+
+    userEvent.click(screen.getByRole('button', { name: 'Cancel' }));
+
+    expect(screen.getByTestId('Team name-key')).toHaveTextContent('Team name');
```
ntsim, line 117: This doesn't seem right as it asserts that the key is itself. This would be better:

```ts
expect(screen.getByTestId('Team name')).toHaveTextContent('Team Smith');
```

We should also to assert other things to ensure that the form is no longer there e.g. the submit button, the fields, etc.

```ts
expect(screen.queryByLabelText('Team name')).not.toBeInTheDocument();
expect(screen.queryByRole('button', { name: 'Update contact details')).not.toBeInTheDocument();
```

src/explore-education-statistics-admin/src/pages/publication/components/PublicationContactForm.tsx
```diff
@@ -0,0 +1,133 @@
...
+  onCancel: () => void;
+  onSubmit: (values: FormValues) => void;
+}
+
+const PublicationContactForm = ({
+  initialValues,
+  onCancel,
+  onSubmit,
+}: Props) => {
+  const [showConfirmSubmitModal, toggleShowConfirmSubmitModal] = useToggle(
```
ntsim, line 32: Could simplify this to `show/toggleConfirmModal`

src/explore-education-statistics-admin/src/pages/publication/components/PublicationContactForm.tsx
```diff
@@ -0,0 +1,133 @@
...
+          teamName: '',
+          teamEmail: '',
+          contactName: '',
+          contactTelNo: '',
+        }),
+      }}
+      validationSchema={validationSchema}
+      onSubmit={useFormSubmit((values: FormValues) => {
+        onSubmit(values);
+      })}
```
ntsim, line 59: Could simplify this to a one-liner:
```tsx
onSubmit={useFormSubmit(onSubmit)}
```

src/explore-education-statistics-admin/src/pages/publication/components/PublicationContactForm.tsx
```diff
@@ -0,0 +1,133 @@
...
+    >
+      {form => (
+        <>
+          <Form id="publicationContactForm">
+            <FormFieldset
+              id="contact"
+              legend="Contact for this publication"
+              legendSize="m"
```
ntsim, line 67: The size of this is different to the read-only mode. Would at least make them the same size to be consistent.

We might want to consider just removing this fieldset and just use the heading and help text in both read-only and editable modes.

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/publication/components/PublicationContactForm.tsx
amyb-hiveit: I've gone for removing the fieldset as it's not really adding anything.

src/explore-education-statistics-admin/src/pages/publication/components/PublicationContactForm.tsx
```diff
@@ -0,0 +1,122 @@
...
+  return (
+    <Formik<FormValues>
+      enableReinitialize
+      initialValues={{
+        ...(initialValues ?? {
+          teamName: '',
+          teamEmail: '',
+          contactName: '',
+          contactTelNo: '',
+        }),
```
ntsim, line 52: `initialValues` should be set so this shouldn't need a default value

ntsim (changes requested):

ntsim (approved):

PR #3407, opened 2022-07-27
EES-3224 publication details page

amyb-hiveit:
Adds the Details page to the new Manage Publication section.

This work will be hidden to users until it's all ready, so for this you can only access the new page directly, at /publication/details, there are no links to it yet.

This PR, https://github.com/dfe-analytical-services/explore-education-statistics/pull/3391 and https://github.com/dfe-analytical-services/explore-education-statistics/pull/3404 all make changes to PublicationPageContainer and publicationRoutes, when one is merged I'll rebase the others so all changes are kept.

Other things to note:
- archive publication is included in this section (it's not in the designs but I checked with Marv that it should go here)
- added the message when a publication is archived to the header of the Manage Publication section so it'll appear on all tabs
- added an `inline` option to `FormThemeTopicSelect` to control the style differences when it's used on this new page and the existing create/edit publication page

details1
details2

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/publication/PublicationDetailsPage.tsx
+++ b/src/explore-education-statistics-admin/src/pages/publication/PublicationDetailsPage.tsx
@@ -1,0 +1,108 @@
+import PublicationDetailsForm, {
+  FormValues,
+} from '@admin/pages/publication/components/PublicationDetailsForm';
+import usePublicationContext from '@admin/pages/publication/contexts/PublicationContext';
+import publicationService from '@admin/services/publicationService';
+import themeService from '@admin/services/themeService';
+import Button from '@common/components/Button';
+import LoadingSpinner from '@common/components/LoadingSpinner';
+import SummaryList from '@common/components/SummaryList';
+import SummaryListItem from '@common/components/SummaryListItem';
+import useAsyncHandledRetry from '@common/hooks/useAsyncHandledRetry';
+import useToggle from '@common/hooks/useToggle';
+import React from 'react';
+
+const PublicationDetailsPage = () => {
+  const { publication, onReload } = usePublicationContext();
+  const {
+    contact,
+    id,
+    permissions,
+    supersededById,
+    themeId,
+    title,
+    topicId,
+  } = publication;
+  const [readOnly, toggleReadOnly] = useToggle(true);
+
+  const { value, isLoading } = useAsyncHandledRetry(async () => {
+    const themes = await themeService.getThemes();
+    if (!supersededById && !permissions.canUpdatePublicationSupersededBy) {
+      return { themes };
+    }
+
+    const allPublications = await publicationService.getPublicationSummaries();
+    const publications = allPublications.filter(pub => pub.id !== id);
+
+    return {
+      themes,
+      publications,
+    };
+  });
+
+  const { themes, publications } = value ?? {};
+
+  const currentTheme = themes?.find(theme => theme.id === themeId);
+  const currentTopic = currentTheme?.topics.find(topic => topic.id === topicId);
+  const currentSupersededByPublication = publications?.find(
+    pub => pub.id === supersededById,
+  );
+
+  const handleSubmit = async (values: FormValues) => {
+    await publicationService.updatePublication(publication.id, {
+      ...values,
+      contact: {
+        teamName: contact?.teamName ?? '',
+        teamEmail: contact?.teamEmail ?? '',
+        contactName: contact?.contactName ?? '',
+        contactTelNo: contact?.contactTelNo ?? '',
+      },
+    });
+    onReload();
+  };
+
+  return (
+    <LoadingSpinner loading={isLoading}>
+      {readOnly ? (
+        <>
+          <h2>Publication details</h2>
+          <SummaryList>
+            <SummaryListItem term="Publication title">{title}</SummaryListItem>
+            <SummaryListItem term="Theme">
+              {currentTheme?.title}
+            </SummaryListItem>
+            <SummaryListItem term="Topic">
+              {currentTopic?.title}
+            </SummaryListItem>
+            <SummaryListItem term="Superseding publication">
+              {supersededById
+                ? currentSupersededByPublication?.title
+                : 'This publication is not archived'}
+            </SummaryListItem>
+          </SummaryList>
+          <Button variant="secondary" onClick={toggleReadOnly.off}>
+            Edit publication details
+          </Button>
+        </>
+      ) : (
+        <PublicationDetailsForm
+          canUpdatePublicationSupersededBy={
+            permissions.canUpdatePublicationSupersededBy
+          }
+          canUpdatePublicationTitle={permissions.canUpdatePublicationTitle}
+          initialValues={{
+            supersededById,
+            title,
+            topicId,
+          }}
+          publications={publications}
+          themes={themes}
+          onCancel={toggleReadOnly.on}
+          onSubmit={handleSubmit}
+        />
+      )}
+    </LoadingSpinner>
+  );
+};
+
+export default PublicationDetailsPage;
--- a/src/explore-education-statistics-admin/src/pages/publication/PublicationDetailsPage.tsx
+++ b/src/explore-education-statistics-admin/src/pages/publication/PublicationDetailsPage.tsx
@@ -1,5 +1,5 @@
 import PublicationDetailsForm, {
-  FormValues,
+  PublicationDetailsFormValues,
 } from '@admin/pages/publication/components/PublicationDetailsForm';
 import usePublicationContext from '@admin/pages/publication/contexts/PublicationContext';
 import publicationService from '@admin/services/publicationService';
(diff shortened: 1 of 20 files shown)


src/explore-education-statistics-admin/src/pages/publication/PublicationDetailsPage.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  const handleSubmit = async (values: FormValues) => {
+    await publicationService.updatePublication(publication.id, {
+      ...values,
+      contact: {
+        teamName: contact?.teamName ?? '',
+        teamEmail: contact?.teamEmail ?? '',
+        contactName: contact?.contactName ?? '',
+        contactTelNo: contact?.contactTelNo ?? '',
+      },
```
ntsim, line 59: Checked with @benoutram and contacts are no longer nullable locally or in production. Contacts are always created when creating new publications as well so we should change the `MyPublication.contact` type to be non-optional so we don't need to do this kind of thing anymore.

The following would suffice:

```ts
await publicationService.updatePublication(publication.id, {
  ...values,
  contact
});
```

src/explore-education-statistics-admin/src/pages/publication/PublicationPageContainer.tsx
```diff
@@ -41,6 +43,17 @@ const PublicationPageContainer = ({
...
+                <WarningMessage className="govuk-!-margin-bottom-0">
+                  This publication is archived.
+                </WarningMessage>
+              )}
+              {publication.supersededById && !publication.isSuperseded && (
+                <WarningMessage className="govuk-!-margin-bottom-0">
+                  This publication will be archived when its superseding
+                  publication has a live release published.
+                </WarningMessage>
+              )}
```
ntsim, line 56: Might want to have some tests for this

src/explore-education-statistics-admin/src/pages/publication/PublicationDetailsPage.tsx
```diff
@@ -0,0 +1,108 @@
...
+            </SummaryListItem>
+            <SummaryListItem term="Topic">
+              {currentTopic?.title}
+            </SummaryListItem>
+            <SummaryListItem term="Superseding publication">
+              {supersededById
+                ? currentSupersededByPublication?.title
+                : 'This publication is not archived'}
+            </SummaryListItem>
+          </SummaryList>
```
ntsim, line 82: Would suggest not showing this unless the publication's superseded as it's not really important otherwise.

src/explore-education-statistics-admin/src/pages/publication/components/PublicationDetailsForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const [showConfirmModal, toggleConfirmModal] = useToggle(false);
+
+  const errorMappings = [
+    mapFieldErrors<FormValues>({
+      target: 'title',
+      messages: {
+        SlugNotUnique: 'Choose a unique title',
+      },
+    }),
+  ];
```
ntsim, line 52: Typically error mappings are placed the module level. Other than preventing them from being re-constructed when the component re-renders, they also take up a lot of space and don't need to be within the component itself.

src/explore-education-statistics-admin/src/pages/publication/components/PublicationDetailsForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  const validationSchema = Yup.object<FormValues>({
+    title: Yup.string().required('Enter a title'),
+    topicId: Yup.string().required('Choose a topic'),
+  });
```
ntsim, line 57: Not a biggy, but would keep the validation schema closer to where it's being used i.e. either in-line with the prop or after callback declarations

src/explore-education-statistics-admin/src/pages/publication/components/PublicationDetailsForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  const handleSubmit = useFormSubmit((values: FormValues) => {
+    onSubmit(values);
+  }, errorMappings);
```
ntsim, line 61: Can be simplified to a one-liner:

```ts
const handleSubmit = useFormSubmit(onSubmit, errorMappings);
```

src/explore-education-statistics-admin/src/pages/publication/components/PublicationDetailsForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+        }),
+      }}
+      validationSchema={validationSchema}
+      onSubmit={handleSubmit}
+    >
+      {form => (
+        <>
+          <Form id={id}>
+            <FormFieldset id="details" legend="Publication details">
+              {canUpdatePublicationTitle && (
```
ntsim, line 82: It doesn't actually make a lot of sense why we'd be able to update the publication title but still be able to change the theme and topic as well.

Might want to check with Cam on this, but would suggest that if you can't change the title, you also shouldn't be able to change the theme or topic. 

I'd probably want to go further and change the backend permission naming to something more appropriate like `canUpdatePublicationDetails`. Having permissions for individual fields like the title is totally overkill. We don't have to do this as part of this work, but would suggest we ticket it up as an improvement / tech debt.

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/publication/PublicationDetailsPage.tsx
amyb-hiveit: Great, that's always been annoying!

src/explore-education-statistics-admin/src/pages/publication/PublicationDetailsPage.tsx
amyb-hiveit: I disagree as it makes the archive setting harder to find - you have to know it's hidden in the edit details form.

src/explore-education-statistics-admin/src/pages/publication/components/PublicationDetailsForm.tsx
amyb-hiveit: As discussed in slack, have wrapped `canUpdatePublicationTitle` around the theme and topic, as well as title, and will raise a ticket to tidy the permissions up.

src/explore-education-statistics-admin/src/pages/publication/PublicationDetailsPage.tsx
```diff
@@ -0,0 +1,95 @@
...
+      theme,
+      supersedingPublication,
+    };
+  });
+
+  const { theme, supersedingPublication } = value ?? {};
+
+  const topic = theme?.topics.find(themeTopic => themeTopic.id === topicId);
+
+  const handleSubmit = async (values: FormValues) => {
```
ntsim, line 46: Just a minor thing, but would export this as `PublicationDetailsFormValues` as other form components might also export their form values. It'd be good to be able to differentiate between exported form values.

src/explore-education-statistics-admin/src/pages/publication/PublicationDetailsPage.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    return {
+      theme,
+      supersedingPublication,
+    };
+  });
+
+  const { theme, supersedingPublication } = value ?? {};
+
+  const topic = theme?.topics.find(themeTopic => themeTopic.id === topicId);
```
ntsim, line 44: Could just do this in the `useAsyncHandledRetry`

src/explore-education-statistics-admin/src/pages/publication/__tests__/PublicationDetailsPage.test.tsx
```diff
@@ -0,0 +1,408 @@
...
+    expect(topics[0]).toHaveValue('theme-1-topic-1');
+    expect(topics[1]).toHaveTextContent('Theme 1 Topic 2');
+    expect(topics[1]).toHaveValue('theme-1-topic-2');
+
+    userEvent.selectOptions(screen.getByLabelText('Select theme'), ['theme-2']);
+
+    const updatedTopics = within(
+      screen.getByLabelText('Select topic'),
+    ).getAllByRole('option');
+
```
ntsim, line 246: We should count the length to make sure we have the right amount of topics

src/explore-education-statistics-admin/src/pages/publication/__tests__/PublicationDetailsPage.test.tsx
```diff
@@ -0,0 +1,408 @@
...
+    ]);
+
+    userEvent.click(
+      screen.getByRole('button', { name: 'Update publication details' }),
+    );
+
+    userEvent.click(screen.getByRole('button', { name: 'Confirm' }));
+
+    await waitFor(() => {
+      expect(publicationService.updatePublication).toHaveBeenCalledWith(
```
ntsim, line 383: Not a biggy, but to help with type-safety, you can declare these like:
```ts
toHaveBeenCalledWith<Parameters<typeof publicationService.updatePublication>>(...)
```

src/explore-education-statistics-admin/src/pages/publication/__tests__/PublicationDetailsPage.test.tsx
```diff
@@ -0,0 +1,408 @@
...
+    await waitFor(() => {
+      expect(screen.getByText('Publication details')).toBeInTheDocument();
+    });
+
+    expect(screen.getByTestId('Superseding publication')).toHaveTextContent(
+      'Publication 2',
+    );
+  });
+
+  test('shows the form when the edit button is clicked', async () => {
```
ntsim, line 157: Just for better organisation, might be nice if we grouped all the form related tests into their own `describe` block

src/explore-education-statistics-admin/src/pages/publication/components/PublicationDetailsForm.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+    return {
+      themes,
+      publications,
+    };
+  });
+
+  const { themes, publications } = value ?? {};
+
+  const id = 'publicationDetailsForm';
```
ntsim, line 73: This can also be defined at module scope

ntsim (approved):
Noice! :+1:

ntsim (approved):

PR #3408, opened 2022-07-27
EES-2425 Allow ordering of subjects

markhiveit:
This PR allows users to reorder subjects in Admin. The subjects are displayed in that order on Admin on the Data and files page and Data guidance page, and on Public on the Data tables page, Data catalogue page, and release Data guidance pages.

We decided to put the `Order` column on the `ReleaseFile` table. This is because in places where we display subjects, both Admin and Public, we currently get the subjects from the Content DB. While the column's more logical home is in `ReleaseSubject`, this would require us to make extra queries to the Statistics DB. So to prevent these extra queries, we are putting the column in `ReleaseFile`, and accepting its redundancy for non-data files.

Newly imported subject are always put last in the list by taking the current max order value and adding one - done in `ReleaseDataFileService#GetReleaseFileOrder`. This prevents a need to update order values when subjects are removed because we can allow gaps in the order numbers, like `0, 1, 3, 4`.

Extra code was required to ensure that the current order is preserved during a data replacement. This basically involves copying the old subjects order value to the new replacement subject.

There were a few additional code changes that were required after adding `#nullable enable`.


src/explore-education-statistics-admin/src/pages/release/data/components/ReleaseDataUploadsSection.tsx
```diff
@@ -245,84 +246,115 @@ const ReleaseDataUploadsSection = ({
...
-                        <>
-                          {dataFile.status === 'COMPLETE' && (
-                            <>
-                              <Link
-                                className="govuk-!-margin-right-4"
-                                to={generatePath<ReleaseDataFileRouteParams>(
-                                  releaseDataFileRoute.path,
-                                  {
+          <EditingContextProvider editingMode="edit">
+            <EditableAccordion
```
amyb-hiveit, line 250: I'm not sure that reusing EditableAccordion is the best option here. Naming-wise it's confusing as it's not editable in this case and it ends up adding complexity to the EditableAccordion components to hide most of the functionality if it's not editable.

I suggest that we instead have ReorderableAccordion and ReorderableAccordionSection components that have the same reordering stuff as the Editable ones but with all the editing functionality taken out. This:
- would make the components simpler
- would be easier to reuse if we need more reorderable accordions in future, and
- wouldn't need to be inside the EditingContextProvider 

It'll be duplicating some code, which I'm guessing is what you were trying to avoid, but I think it's worth it here.

Happy to chat about this and help.

amyb-hiveit (changes requested):

src/explore-education-statistics-admin/src/components/editable/EditableAccordionSection.tsx
```diff
@@ -58,7 +58,7 @@ const EditableAccordionSection = (props: EditableAccordionSectionProps) => {
   const [newHeading, setNewHeading] = useState(heading);
 
   const saveHeading = useCallback(async () => {
-    if (isEditingHeading && onHeadingChange && newHeading !== heading) {
+    if (onHeadingChange && isEditingHeading && newHeading !== heading) {
```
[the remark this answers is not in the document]
markhiveit: Seems like if it's not included it complains that the onHeadingChnage call below might be null.

src/explore-education-statistics-admin/src/pages/release/data/components/ReleaseDataUploadsSection.tsx
```diff
@@ -245,84 +246,115 @@ const ReleaseDataUploadsSection = ({
...
+                    return acc;
+                  },
+                  {},
+                );
+                await releaseDataFileService.updateDataFilesOrder(
+                  releaseId,
+                  order,
+                );
+
+                // NOTE: Doesn't like it if we pass the result of the PUT to setDataFiles so we do this
```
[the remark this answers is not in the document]
markhiveit: In `releaseDataFileService.ts`, the new `updateDataFilesOrder` endpoint call needed to do the same transformation of `DataFileInfo[] -> DataFile[]` that `getDataFiles` does. Now it successfully can `setDataFiles` on the result of `updateDataFilesOrder`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -184,26 +185,69 @@ await ids.Select(id => _releaseFileRepository.CheckFileExists(releaseId, id, Fil
...
+                .OnSuccess(release => _userService.CheckCanUpdateRelease(release))
+                .OnSuccess(async _ =>
+                {
+                    var releaseFiles = await _contentDbContext.ReleaseFiles
+                        .Include(releaseFile => releaseFile.File)
+                        .Where(releaseFile =>
+                            releaseFile.File.Type == FileType.Data
+                            && releaseFile.ReleaseId == releaseId
+                            && !releaseFile.File.ReplacingId.HasValue)
+                        .ToListAsync();
```
benoutram, line 199: You could use `ToDictionary` here to provide a quicker lookup of the files by id because the way `releaseFiles.Find` is used below is iterating over all the files to find the correct one by id, and that's happening for every file within the loop.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -403,10 +461,10 @@ private async Task<string> GetSubjectName(Guid releaseId, File file)
         }
 
         private async Task<Either<ActionResult, string>> ValidateSubjectName(Guid releaseId,
-            string subjectName,
+            string? subjectName,
             File? replacingFile)
         {
-            if (replacingFile == null)
+            if (replacingFile == null && subjectName != null)
```
benoutram, line 467: Just wondering if we should enforce here that the name is present. Seems like we allow subjectName to be null when it's not a replacement and that will throw an exception rather than a return a validation result. If that's dealt with as early as possible then I think we avoid the optional subjectName in fileUploadsValidatorService.ValidateSubjectName.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseDataFileServiceTest.cs
```diff
@@ -995,6 +996,211 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+            };
+            var releaseMetaFile5 = new ReleaseFile
+            {
+                Release = releaseDataFile5.Release,
+                File = new File
+                {
+                    Type = FileType.Metadata,
+                    SubjectId = releaseDataFile5.File.SubjectId
+                }
+            };
```
benoutram, line 1098: I don't think you need to add SubjectId's here anymore as to build the view models, the meta file is now found via the `DataImport`

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseDataFileServiceTest.cs
```diff
@@ -1684,15 +1892,183 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+                        metaFormFile,
+                        null))
+                    .ReturnsAsync(Unit.Instance);
+
+                dataImportService
+                    .Setup(s => s.Import(
+                        It.IsAny<Guid>(),
+                        It.Is<File>(file => file.Type == FileType.Data && file.Filename == dataFileName),
+                        It.Is<File>(file => file.Type == Metadata && file.Filename == metaFileName)))
+                    .ReturnsAsync(new DataImport { Status = QUEUED }); // @MarkFix
```
benoutram, line 2005: Is ther esomething that you still need to fix up here?

markhiveit: That was me during the nightmare merge conflict - and then I forgot about it. I'll sort it.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -508,5 +563,35 @@ private async Task DeleteBatchFiles(File dataFile)
...
+                        && rf.File.Id == replacingFile.Id);
+                return replacedReleaseDataFile.Order;
+            }
+
+            var existingReleaseFilesOrders = _contentDbContext.ReleaseFiles
+                .Include(releaseFile => releaseFile.File)
+                .Where(releaseFile => releaseFile.ReleaseId == releaseId
+                                      && releaseFile.File.Type == FileType.Data
+                                      && releaseFile.File.ReplacingId == null)
+                .Select(releaseFile => releaseFile.Order)
```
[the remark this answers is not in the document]
markhiveit: I originally did this. I changed it to how it currently is as this query might find no release files, and then the `Max` call failed. So I did this instead and then added the `IsNullOrEmpty` check below.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -508,5 +563,35 @@ private async Task DeleteBatchFiles(File dataFile)
...
+                .ToList();
```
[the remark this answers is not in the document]
markhiveit: Changed to async

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/ReleaseDataFileRepository.cs
```diff
@@ -137,6 +140,8 @@ private IQueryable<File> ListDataFilesQuery(Guid releaseId)
                           && rf.File.ReplacingId == null
                           && rf.File.SubjectId.HasValue
                 )
+                .OrderBy(rf => rf.Order)
+                .ThenBy(rf => rf.Name)
```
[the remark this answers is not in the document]
markhiveit: Discussed this with Ben - I think the time to create a migration to give correct order numbers to existing subjects and then make sure it works / test it isn't worth the benefit, so I'm going to leave this. I will leave a comment in the code to discuss why the ordering by Name happens.

markhiveit:
After a discussion with Ben on Slack, decided to not allow gaps to happen in the order - so added some code to `ReleaseFileRepository#Delete` to ensure order numbers are always sequential. I've also updated a unit test to ensure this is working correctly.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -392,8 +445,13 @@ private async Task<List<DataFileInfo>> BuildDataFileViewModels(List<ReleaseFile>
...
         }
 
-        private async Task<string> GetSubjectName(Guid releaseId, File file)
+        private async Task<string> GetSubjectName(Guid releaseId, File? file)
         {
+            if (file == null)
+            {
+                throw new ArgumentException("file should not be null");
+            }
+
```
ntsim, line 454: Similar to earlier - this seems like a big code smell as we accept an optional parameter that immediately throws if it's not present.

We should revert this change and handle any null `file` parameter in the consuming method i.e. `ValidateSubjectName`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -203,6 +250,8 @@ await ids.Select(id => _releaseFileRepository.CheckFileExists(releaseId, id, Fil
                                 await _releaseRepository
                                     .CreateStatisticsDbReleaseAndSubjectHierarchy(releaseId);
 
+                            int releaseDataFileOrder = GetReleaseDataFileOrder(replacingFile, releaseId);
```
ntsim, line 253: This can be `var`

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ReleaseService.cs
```diff
@@ -83,31 +83,20 @@ private async Task<List<SubjectViewModel>> GetSubjects(Guid releaseId, List<Guid
...
                             );
                         }
                     ))
-                .OrderBy(svm => svm.Name)
+                .OrderBy(svm => svm.Order)
+                .ThenBy(svm => svm.Name)
                 .ToList();
         }
 
-        private async Task<string> GetSubjectName(Guid releaseId, Guid subjectId)
```
ntsim, line 98: Nice bit of tidy!

src/explore-education-statistics-admin/src/components/editable/ReorderableAccordion.tsx
```diff
@@ -0,0 +1,128 @@
...
+              Reorder<span className="govuk-visually-hidden"> data files</span>
+            </Button>
+          ) : (
+            <Button
+              className="govuk-!-font-size-16 govuk-!-margin-bottom-0"
+              onClick={saveOrder}
+            >
+              Save order
+            </Button>
+          ))}
```
ntsim, line 100: Buttons like this should not be in the heading as it's not accessible. This should be placed below.

src/explore-education-statistics-admin/src/components/editable/ReorderableAccordionSection.module.scss
```diff
@@ -0,0 +1,50 @@
...
+
+  label {
+    display: none;
+  }
+
+  input {
+    position: relative;
+    z-index: 1000;
+  }
+}
```
ntsim, line 14: We don't need this.

src/explore-education-statistics-admin/src/components/editable/ReorderableAccordionSection.tsx
```diff
@@ -0,0 +1,88 @@
...
+          // eslint-disable-next-line react/jsx-props-no-spreading
+          {...draggableProvided.draggableProps}
+          // eslint-disable-next-line react/jsx-props-no-spreading
+          {...draggableProvided.dragHandleProps}
+          ref={draggableProvided.innerRef}
+          className={classNames({
+            [styles.dragContainer]: isReordering,
+            [styles.isDragging]: snapshot.isDragging,
+          })}
+          data-testid="editableAccordionSection"
```
ntsim, line 66: `reorderableAccordionSection` is probably a better test id

tests/robot-tests/tests/files/four.csv
```diff
@@ -0,0 +1,17 @@
+time_period,time_identifier,geographic_level,country_code,country_name,old_la_code,new_la_code,la_name,filter_one,filter_two,filter_group_one,filter_three,filter_four,filter_group_four,ind_one
```
ntsim, line 1: Are there some more appropriate names for these new test files? Names like one, two, three and four are super vague and don't really tell us what they're intended for.

tests/robot-tests/tests/admin_and_public/bau/subject_reordering.robot
```diff
@@ -0,0 +1,246 @@
...
+
+    user clicks radio    %{TEST_THEME_NAME}
+    user clicks radio    ${PUBLICATION_NAME}
+    user clicks element    id:publicationForm-submit
+
+    user waits until table tool wizard step is available    2    Choose a subject
+    user checks previous table tool step contains    1    Publication    ${PUBLICATION_NAME}
+
+    user checks page contains element
+    ...    xpath://*[contains(@class, "govuk-radios__item")][1]//label[contains(text(), "One")]
```
ntsim, line 196: Would extract a keyword like `user checks radio exists`. 

Would also use a better selector more inline with other radio keywords `xpath://label[text()="${label}"]/../input[@type="radio"]` where we don't rely on a class selector like `govuk-radios__item`

tests/robot-tests/tests/admin_and_public/bau/subject_reordering.robot
```diff
@@ -0,0 +1,246 @@
...
+    user waits until page contains    Choose a release
+    user waits until page contains    ${RELEASE_NAME}
+
+    user clicks radio    ${RELEASE_NAME}
+    user clicks button    Next step
+
+    user waits until page contains    Choose files to download
+
+    user checks page contains element
+    ...    xpath://*[contains(@class, "govuk-checkboxes__item")][1]//label[contains(text(), "One")]
```
ntsim, line 220: Similar to above - would extract a keyword like `user checks checkbox exists` and use a better selector that doesn't rely on a class i.e `xpath://label[text()="${label}"]/../input[@type="radio"]`

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -508,5 +563,35 @@ private async Task DeleteBatchFiles(File dataFile)
         {
             await _blobStorageService.DeleteBlobs(PrivateReleaseFiles, dataFile.BatchesPath());
         }
+
+        private int GetReleaseDataFileOrder(File? replacingFile, Guid releaseId)
```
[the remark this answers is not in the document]
ntsim: Has this change been done?

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleasesController.cs
```diff
@@ -96,6 +96,17 @@ public async Task<ActionResult<IEnumerable<DataFileInfo>>> GetDataFileInfo(Guid
                 .HandleFailuresOrOk();
         }
 
+        [HttpPut("release/{releaseId}/data/order")]
+        [Produces("application/json")]
+        [ProducesResponseType(200)]
+        [ProducesResponseType(404)]
+        public async Task<ActionResult<IEnumerable<DataFileInfo>>> ReorderDataFiles(Guid releaseId,
+            Dictionary<Guid,int> newOrder)
```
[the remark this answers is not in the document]
markhiveit: I did it this way because I was copying how it was done in `ContentService#ReorderContentSections`.

The change was easy to make so I've switched to using a list of fileIds now instead of a dictionary.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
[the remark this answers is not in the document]
markhiveit: I've just not pushed it yet. The same will be true for other comments too.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -184,26 +185,69 @@ await ids.Select(id => _releaseFileRepository.CheckFileExists(releaseId, id, Fil
...
-                        .OrderBy(file => file.Name)
                         .AsEnumerable();
                 });
         }
 
+        public async Task<Either<ActionResult, IEnumerable<DataFileInfo>>> ReorderDataFiles(
+            Guid releaseId,
+            Dictionary<Guid, int> newOrder)
+        {
+            return await _persistenceHelper
```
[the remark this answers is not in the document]
markhiveit: Added validate that the provided `List<Guid>` contains unique values, the same count as the number in the database, and that all fileIds match one in the database.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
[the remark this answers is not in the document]
markhiveit: Managed to get `Max` working by using `DefaultIfEmpty`

src/explore-education-statistics-admin/src/components/editable/ReorderableAccordion.tsx
markhiveit: Discussed this with Nick on Slack. The buttons have been moved out of the `h2` and some styling added to ensure things still look the right. Since the original code was copied from `EditableAccordion`, I've also made the same change there.

tests/robot-tests/tests/admin_and_public/bau/subject_reordering.robot
markhiveit: I've extracted this to the new keyword `user checks x position radio has label`.

In order to avoid the `govuk-radios__item`, i've had to resort to `*[contains(@data-testid, "Radio item for ")]`. This is necessary as I need to select the Xth radio first, then check the label. I tried starting the selector an `input`, but `//input[@type="radio"][2]` didn't work.

tests/robot-tests/tests/admin_and_public/bau/subject_reordering.robot
markhiveit: Did practically the same as with the radio case above, except there was no `data-testid` on check box items, so I added one.

markhiveit:
image

tests/robot-tests/tests/files/four.csv
markhiveit: Discussed this with Nick and have renamed the files

ntsim (changes requested):

ntsim (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/ReleaseDataFileRepository.cs
```diff
@@ -15,7 +15,7 @@ public class ReleaseDataFileRepository : IReleaseDataFileRepository
     {
         private readonly ContentDbContext _contentDbContext;
 
-        private static readonly List<FileType> SupportedFileTypes = new()
+        private static readonly List<FileType> SupportedFileTypes = new ()
```
luke-h1, line 18: accidental whitespace here

src/explore-education-statistics-admin/src/components/editable/EditableAccordion.tsx
```diff
@@ -95,7 +97,7 @@ const EditableAccordion = (props: EditableAccordionProps) => {
               className="govuk-!-font-size-16 govuk-!-margin-bottom-0"
               onClick={toggleReordering.on}
             >
-              Reorder <span className="govuk-visually-hidden">sections</span>
+              Reorder<span className="govuk-visually-hidden"> sections</span>
```
luke-h1, line 100: accidental whitespace here?

tests/robot-tests/tests/admin_and_public/bau/subject_reordering.robot
```diff
@@ -0,0 +1,238 @@
...
+    user chooses file    id:dataFileUploadForm-metadataFile
+    ...    ${FILES_DIR}ordering-test-3-replacement.meta.csv
+    user clicks button    Upload data files
+
+    user waits until page contains element    testid:Replacement Subject title
+    user checks headed table body row cell contains    Subject title    1    Three
+    user checks headed table body row cell contains    Data file    1    ordering-test-3.csv
+    user checks headed table body row cell contains    Subject title    2    Three
+    user checks headed table body row cell contains    Data file    2    ordering-test-3-replacement.csv
+    user checks headed table body row cell contains    Status    2    Complete    wait=%{WAIT_LONG}
```
luke-h1, line 116: Don't think you need `wait=` here

src/explore-education-statistics-admin/src/components/editable/EditableAccordion.tsx
ntsim: Nope, this is correct :wink:

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/ReleaseDataFileRepository.cs
markhiveit: I believe this is the way Rider normally formats this - rather than having no space. In the codebase, we sometimes have a space and sometimes don't, so I'm going to leave this.

PR #3413, opened 2022-07-28
EES-3585 - fix manage_users_page.robot test

luke-h1:


luke-h1:
> I'm guessing this should get fixed by EES-3578? Probably just need some kind of reminder to put these back in once the bug is fixed

No this is failing because we don't have pre-release user 3 on dev at the moment (last I checked. Currently can't get ssms on the new dfe laptop). I think this was just missed during the phase of having 0 test users. This will likely be fixed by https://dfedigital.atlassian.net/browse/EES-3546

ntsim (approved):
Approved, but lets remember to get the resolving issue sorted out ASAP so that we're not without tests for too long

luke-h1:
Added a dot not merge label due to investigating other problems with pre-prelease users on dev that could be affecting this test

luke-h1:
Think the failures were due to the pre-release users being used manually (outside of UI tests). This test clears down any pre-release user to have  'no role' so that they are included in the list of users on the `manage user`. In addition to this, it doesn't look like there are any conflicts/tests editing pre-release user roles (when running the test in parallel against dev) which could be causing pre-release users on dev to have a role (and thus being filtered out of the list of users on the `manage-user` page. I've made amends to the test to use pre-release user 2 and changed a few things that were causing flakiness when ran against dev

tests/robot-tests/tests/admin/bau/manage_users_page.robot
```diff
@@ -1,6 +1,7 @@
 *** Settings ***
 Resource            ../../libs/admin-common.robot
 Resource            ../../libs/common.robot
+Resource            ../../libs/common.robot
```
duncan-at-hiveit, line 4: Accidental double import here

tests/robot-tests/tests/admin/bau/manage_users_page.robot
```diff
@@ -69,16 +62,18 @@ Select a user to manage
...
+    user waits until h1 is visible    Prerelease2 User2    10
 
 Check the initial manage user page
-    user checks select contains x options    name:selectedRoleId    4
-    user checks select contains option    name:selectedRoleId    Choose role
-    user checks select contains option    name:selectedRoleId    Analyst
-    user checks select contains option    name:selectedRoleId    BAU User
-    user checks select contains option    name:selectedRoleId    Prerelease User
-    user checks selected option label    name:selectedRoleId    Choose role
+    user checks select contains x options    //*[@name="selectedRoleId"]    4
```
duncan-at-hiveit, line 71: What's the difference in behaviour here between the old and the new?

markhiveit (approved):

PR #3415, opened 2022-07-28
EES-3550, EES-3589 - review UI tests for TODOs, failing tags & other tags

luke-h1:
This PR: 
- fixes a few TODOs and removes either unnecessary or failing tags from the tests. 

Relevant changes:
* Starts to scaffold out possibilities of how to validate data-guidance directory structure/files 
* Adds assertion to check new release status history entries get created on amendments
* Introduces a data-guidance related TODO in `release_page_absence.robot`
* Increases the timeout for `user clears element text` & `user presses keys` due to periodic failures I noticed when running the tests locally
* Adds `desired_capabilities=DesiredCapabilities.CHROME` parameter when creating a new instance of chrome driver. This seems to silence the various logging some people get on other consoles 
* Adds assertions to check for regressions of [EES-3535](https://dfedigital.atlassian.net/browse/EES-3535)
* Adds `%{WAIT_SMALL}` in various tests that check table rows due to changes made to how table sizes/rows are fetched which seems to have introduced flakiness in these areas (fixes https://dfedigital.atlassian.net/browse/EES-3589)
Local run: 
image

### DIFF
--- a/tests/robot-tests/tests/admin_and_public/bau/data_reordering.robot
+++ b/tests/robot-tests/tests/admin_and_public/bau/data_reordering.robot
@@ -212,7 +212,7 @@ Replace subject data
     user checks headed table body row cell contains    Subject title    1    ${SUBJECT_NAME}
     user checks headed table body row cell contains    Data file    1    grouped-filters-and-indicators.csv
     user checks headed table body row cell contains    Metadata file    1    grouped-filters-and-indicators.meta.csv
-    user checks headed table body row cell contains    Number of rows    1    100
+    user checks headed table body row cell contains    Number of rows    1    100    wait=%{WAIT_SMALL}
     user checks headed table body row cell contains    Data file size    1    13 Kb
     user checks headed table body row cell contains    Status    1    Data replacement in progress    wait=%{WAIT_LONG}
 
@@ -220,8 +220,8 @@ Replace subject data
     user checks headed table body row cell contains    Data file    2    grouped-filters-and-indicators-replacement.csv
     user checks headed table body row cell contains    Metadata file    2
     ...    grouped-filters-and-indicators-replacement.meta.csv
-    user checks headed table body row cell contains    Number of rows    2    140
-    user checks headed table body row cell contains    Data file size    2    19 Kb
+    user checks headed table body row cell contains    Number of rows    2    140    wait=%{WAIT_SMALL}
+    user checks headed table body row cell contains    Data file size    2    19 Kb    wait=%{WAIT_SMALL}
     user checks headed table body row cell contains    Status    2    Complete    wait=%{WAIT_LONG}
 
 Confirm data replacement
@@ -530,8 +530,6 @@ Validate table cells
     user checks results table cell contains    10    2    39,761,905
 
 Go back to locations step
-    # TODO not sure why button with test Id isn't found? Using label instead
-    # user clicks button    testid:wizardStep-3-goToButton
     user clicks button    Edit locations
     user waits until page contains element    xpath://h2[text()="Go back to previous step"]
     user clicks button    Confirm
--- a/tests/robot-tests/tests/admin_and_public/bau/publish_data.robot
+++ b/tests/robot-tests/tests/admin_and_public/bau/publish_data.robot
@@ -455,10 +455,6 @@ Select subject "${SUBJECT_2_NAME}" in table tool
     user checks previous table tool step contains    2    Subject    ${SUBJECT_2_NAME}
 
 Select locations in table tool
-    # LH: Not using the 'user opens details dropdown' keyword here as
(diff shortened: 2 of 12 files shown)


tests/robot-tests/tests/admin_and_public/bau/publish_data.robot
```diff
@@ -455,10 +455,9 @@ Select subject "${SUBJECT_2_NAME}" in table tool
     user checks previous table tool step contains    2    Subject    ${SUBJECT_2_NAME}
 
 Select locations in table tool
-    # LH: Not using the 'user opens details dropdown' keyword here as
+    # don't use the 'user opens details dropdown' keyword here as
```
duncan-at-hiveit, line 458: I think you can probably just remove this whole coment now in my opinion.  I don't think people would question it if they saw something getting selected with its testid, so I don't think we need to maintain a explanation :)

tests/robot-tests/tests/admin_and_public/bau/data_reordering.robot
```diff
@@ -530,9 +530,7 @@ Validate table cells
     user checks results table cell contains    10    2    39,761,905
 
 Go back to locations step
-    # TODO not sure why button with test Id isn't found? Using label instead
-    # user clicks button    testid:wizardStep-3-goToButton
-    user clicks button    Edit locations
+    user clicks element    testid:wizardStep-3-goToButton
```
duncan-at-hiveit, line 533: Just a personal preference, but I actually prefer the label version of this, because it's more obvious to a reader of the test as to what "wizardStep-3-goToButton" is actually doing.

tests/robot-tests/tests/admin_and_public/bau/publish_release_and_amend.robot
```diff
@@ -315,8 +315,8 @@ Navigate to data replacement page
     user checks headed table body row contains    Subject title    Dates test subject
     user checks headed table body row contains    Data file    dates.csv
     user checks headed table body row contains    Metadata file    dates.meta.csv
-    user checks headed table body row contains    Number of rows    118
-    user checks headed table body row contains    Data file size    17 Kb
+    user checks headed table body row contains    Number of rows    118    wait=%{WAIT_SMALL}
```
duncan-at-hiveit, line 318: Is there something about these cells specifically that require the WAITs?  Are they lazy?

tests/robot-tests/tests/admin_and_public/bau/publish_release_and_amend.robot
```diff
@@ -372,7 +372,20 @@ Update existing data guidance for amendment
...
+Add a Footnote
+    user waits until page contains link    Create footnote
+    user clicks link    Create footnote
+    user waits until h2 is visible    Create footnote
+    user clicks footnote subject radio    Dates test subject    Applies to all data
+    user clicks element    id:footnoteForm-content
+    user enters text into element    id:footnoteForm-content
+    ...    A footnote
+    user clicks button    Save footnote
+    user waits until h2 is visible    Footnotes
```
duncan-at-hiveit, line 388: Should we here do a cursory check that the footnote has actually been added at this point?  Could be super simple - just checking that the page contains the new footnote title in a read-only form.

tests/robot-tests/tests/general_public/release_page_absence.robot
```diff
@@ -82,6 +82,7 @@ Validate subject files file type and file unit style
 Validate absence_in_prus.csv file can be downloaded
     [Documentation]    DFE-958    DFE-562
     [Tags]    NotAgainstLocal    Failing
+    # TODO: This needs data-guidance utils to be created to support the new way of downloading files
```
duncan-at-hiveit, line 85: I think a ticket ref would be good here, to tie together the commented-out code in get_auth_tokens.py with this comment, and to let another person who doesn't know the details understand what is needed from the data-guidance utils to support to new way of downloading, and what the new way of downloading is.

luke-h1: Think this was a mistake. I was thinking of data guidance when I mentioned a 'new way of downloading files' (due to how the zip file gets dynamically generated). I'll update the comment to reflect what needs to be added.

tests/robot-tests/tests/admin_and_public/bau/publish_release_and_amend.robot
luke-h1: No, they aren't lazy loaded. The way the backend returns this information has changed which I think has introduced a bit of flakiness in this area (see https://dfedigital.atlassian.net/browse/EES-3571, https://github.com/dfe-analytical-services/explore-education-statistics/pull/3387) I noticed that tests that check file sizes and rows became flaky. They now transition from 'Unknown' > '<file size>' which may be the cause.

PR #3417, opened 2022-07-29
EES-3591 Publication owner and release approver roles couldn't access Prerelease content

benoutram:
There appears to be a mistake in `ViewSpecificPreReleaseSummaryAuthorizationHandler` which uses `CompoundAuthorizationHandler`  where one of its three sub-handlers is `ViewSpecificReleaseAuthorizationHandler.HasUnrestrictedViewerRoleOnReleaseAuthorizationHandler`.

Note how this handler is from a different class ViewSpecific**Release**AuthorizationHandler and configured for the requirement `ViewReleaseRequirement` not `ViewSpecificPreReleaseSummaryRequirement`.

image

This looks like it isn't intentional and should be `ViewSpecificPreReleaseSummaryAuthorizationHandler.HasUnrestrictedViewerRoleOnReleaseAuthorizationHandler` instead. If this handler was in use it would succeed for the Publication role `ReleaseApprover`.

The handler actually in use can only handle `ViewReleaseRequirement` so it isn't being called to evaluate the release and is effectively being ignored.

`ViewSpecificPreReleaseSummaryAuthorizationHandler.HasUnrestrictedViewerRoleOnReleaseAuthorizationHandler` isn't being used at all except by unit tests.

It seems this highlights a disadvantage of using the `CompoundAuthorizationHandler` and individually unit testing each of the sub-handlers, whereby nothing tests the overall behaviour of `ViewSpecificPreReleaseSummaryAuthorizationHandler`.

This PR rewrites `ViewSpecificPreReleaseSummaryAuthorizationHandler` to not use `CompoundAuthorizationHandler` but instead simplify it to make a claims check followed by a combined publication and release roles check.

It also takes the opportunity to add in Publication role `Owner`.

### UI test report

image


duncan-at-hiveit (approved):

duncan-at-hiveit:
Great stuff @benoutram!  Hopefully we can do away with the Compound handlers altogether soon.  Seemed like a good idea at the time :)

PR #3418, opened 2022-08-01
EES-3225 publication team access page

amyb-hiveit:
Adds the Team Access section to Manage Publication, including the Invite new user and Manage release contributors pages. 

This work will be hidden to users until it's all ready, so for this you can only access the new page directly, at /publication/<id>/team, there are no links to it yet.

I've made some changes to existing components that will be visible in the current version but I figured these were small enough changed that it wasn't worth having them under a flag:
- added a table header and changed the remove button text in ReleaseContributorTable
- reduced the size of the checkboxes on the invite user and manage contributors forms

Other notes:
- I've renamed some files that are used here and in the current version so the naming makes more sense for both
- I added `live` to ReleaseSummary as was missing from the type
- I've updated the existing UI tests, but not added new ones (those will be done when this work is ready to go live)

team


src/explore-education-statistics-admin/src/pages/publication/components/__tests__/PublicationManageTeamAccess.test.tsx
```diff
@@ -0,0 +1,195 @@
...
+        release={testRelease}
+      />,
+    );
+
+    await waitFor(() => {
+      expect(
+        screen.getByRole('heading', {
+          name: 'Release 1 (Not live) Draft',
+        }),
+      ).toBeInTheDocument();
```
ntsim, line 74: Don't use `getByRole` selectors (and variants) in `waitFor`. Use faster selectors like `getByText` within the `waitFor` and then perform `getByRole` after it has completed.

src/explore-education-statistics-admin/src/services/releaseService.ts
```diff
@@ -51,6 +51,7 @@ export interface ReleaseSummary {
   publishScheduled: string;
   nextReleaseDate?: PartialDate;
   latestInternalReleaseNote: string;
+  live?: boolean;
```
ntsim, line 54: This is optional? Would have expected it to non-optional. Sounds like a potential bug in the backend. Probably needs a ticket to sort out if so.

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/publication/PublicationTeamAccessPage.tsx
```diff
@@ -0,0 +1,131 @@
...
+  return (
+    <>
+      <div className="govuk-grid-row govuk-!-margin-bottom-8">
+        <div className="govuk-grid-column-two-thirds">
+          <h2>Update release access</h2>
+
+          <FormSelect
+            id="currentRelease"
+            name="release"
+            label="Select release"
```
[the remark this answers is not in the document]
amyb-hiveit: It's managed at both levels:
- 'Invite new users' is at the publication level. You invite a user to have access to all or some of the releases in the publication.
- 'Add or remove users' is on the release level. You're adding/removing users (who already have access to some releases in the publication) to the selected release.

I've had a chat with Marv about this. It was user tested when it was done originally (not that long ago) so he's going to look through the notes from that and decide how to make it clearer, especially the 'Invite new users' button. I'll raise a new ticket under the main dashboard story for this so it'll get done before the work is released.

ntsim (approved):

PR #3419, opened 2022-08-02
EES-3226 manage publication legacy releases page

amyb-hiveit:
Adds the Legacy releases section to Manage Publication, including the create and edit legacy release pages.

This work will be hidden to users until it's all ready, so for this you can only access the new page directly, at /publication/id/legacy, there are no links to it yet.

I've made some changes to the existing LegacyReleasesTable component that will be visible in the current version but I figured these were small enough changes that it wasn't worth having them under a flag:
- change edit and delete button styles and text

legacy


src/explore-education-statistics-admin/src/pages/publication/components/LegacyReleasesTable.tsx
```diff
@@ -41,9 +44,15 @@ const getModalTitle = (confirmAction?: ConfirmAction): string => {
 };
 
 interface Props {
+  createRoute?: string; // EES-3217 remove when live
+  editRoute?: (id: string) => string; // EES-3217 remove when live
```
ntsim, line 48: Not a biggy, but would get into the convention of prefix things like this `TODO`. IDEs and VS Code extensions can detect these in the codebase more easily

src/explore-education-statistics-admin/src/pages/publication/components/LegacyReleasesTable.tsx
```diff
@@ -41,9 +44,15 @@ const getModalTitle = (confirmAction?: ConfirmAction): string => {
...
   publication: MyPublication;
 }
-const LegacyReleasesTable = ({ publication }: Props) => {
+const LegacyReleasesTable = ({
+  createRoute,
+  editRoute,
+  publication,
```
ntsim, line 54: Would be good to decouple the dependency on `publication` as we end up having to pass in a full `MyPublication` when we only need `legacyReleases` and a `publicationId`.

The upcoming backend work in [EES-3575](https://dfedigital.atlassian.net/browse/EES-3575) will split the list of legacy releases out into a separate endpoint so we should save ourselves the hassle of separating this out later.

This might also simplify some of the tests as we won't need to setup some of the repetitive test data.

src/explore-education-statistics-admin/src/routes/publicationRoutes.ts
```diff
@@ -50,3 +58,21 @@ export const publicationContactRoute: PublicationRouteProps = {
...
+
+export const publicationLegacyReleasesRoute: PublicationRouteProps = {
+  path: '/publication/:publicationId/legacy',
+  title: 'Legacy releases',
+  component: PublicationLegacyReleasesPage,
+};
+
+export const publicationCreateLegacyReleaseRoute: PublicationRouteProps = {
+  path: '/publication/:publicationId/legacy/create',
+  title: 'Create Legacy Release',
```
ntsim, line 70: A little weird to capitalise every word. We typically don't do this in other places (e.g. `releaseRoutes`).

Would change this to 'Create legacy release'.

src/explore-education-statistics-admin/src/routes/publicationRoutes.ts
```diff
@@ -50,3 +58,21 @@ export const publicationContactRoute: PublicationRouteProps = {
...
+  component: PublicationCreateLegacyReleasePage,
+};
+
+export const publicationEditLegacyReleaseRoute: PublicationRouteProps = {
+  path: '/publication/:publicationId/legacy/:legacyReleaseId/edit',
+  title: 'Edit Legacy Release',
```
ntsim, line 76: As above - no need to capitalise every word.

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/publication/__tests__/PublicationCreateLegacyReleasePage.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+    expect(screen.getByText('Create legacy release')).toBeInTheDocument();
+    expect(screen.getByLabelText('Description')).toBeInTheDocument();
+    expect(screen.getByLabelText('URL')).toBeInTheDocument();
+    expect(
+      screen.getByRole('button', { name: 'Save legacy release' }),
+    ).toBeInTheDocument();
+    expect(screen.getByRole('link', { name: 'Cancel' })).toBeInTheDocument();
+  });
+});
+
```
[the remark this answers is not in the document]
amyb-hiveit: I've added a test case for submitting. Cancel is just a link so don't think we need to test clicking a link works, instead I've added a check that it has the correct href.

src/explore-education-statistics-admin/src/pages/publication/__tests__/PublicationEditLegacyReleasePage.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+    );
+    expect(screen.getByLabelText('URL')).toHaveValue('http://gov.uk/3');
+    expect(screen.getByLabelText('Order')).toHaveValue(3);
+    expect(
+      screen.getByRole('button', { name: 'Save legacy release' }),
+    ).toBeInTheDocument();
+    expect(screen.getByRole('link', { name: 'Cancel' })).toBeInTheDocument();
+  });
+});
+
```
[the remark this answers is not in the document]
amyb-hiveit: As above for the create tests.

src/explore-education-statistics-admin/src/pages/publication/__tests__/PublicationCreateLegacyReleasePage.test.tsx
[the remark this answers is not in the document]
ntsim: Cool, fair enough!

ntsim (approved):

PR #3421, opened 2022-08-04
EES-3075 - add 'zip' to download all data button

luke-h1:
This PR: 

 * adds `(zip)` to the end of the download all data button.

Other notes:
* Not sure what `downloads.title` is but there seems to be a reference to it in one of the GA logEvents. I've removed the stray curly bracket but potentially it can be removed all together

```typescript 
           onClick={() => {
                logEvent({
                  category: 'Downloads',
                  action: `Release page all files downloads.title, Release: ${release.title}, File: All files`,
                });
              }}
```

### DIFF
--- a/src/explore-education-statistics-frontend/src/modules/find-statistics/PublicationReleasePage.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/find-statistics/PublicationReleasePage.tsx
@@ -247,7 +247,7 @@ const PublicationReleasePage: NextPage<Props> = ({ release }) => {
                         });
                       }}
                     >
-                      Download all data
+                      Download all data (zip)
                     </ButtonLink>
                   </li>
                 )}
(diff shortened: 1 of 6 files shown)


src/explore-education-statistics-frontend/src/modules/find-statistics/PublicationReleasePage.tsx
```diff
@@ -404,11 +404,11 @@ const PublicationReleasePage: NextPage<Props> = ({ release }) => {
               onClick={() => {
                 logEvent({
                   category: 'Downloads',
-                  action: `Release page all files downloads.title}, Release: ${release.title}, File: All files`,
+                  action: `Release page all files downloads.title, Release: ${release.title}, File: All files`,
```
ntsim, line 407: Probably would suggest changing this to 'Release page all files download. ....'

Would suggest checking with Laura / Cam.

ntsim (approved):

src/explore-education-statistics-frontend/src/modules/find-statistics/PublicationReleasePage.tsx
luke-h1: Checked &  changing this might be more effort than it's worth since they'd have to review if that introduces any breaking changes.

PR #3424, opened 2022-08-05
EES-3514 - create python CLI to run tests in CI

luke-h1:


tests/robot-tests/scripts/run_tests_pipeline.py
```diff
@@ -0,0 +1,56 @@
...
+
+    print('Installing dependencies')
+    subprocess.check_call(['google-chrome-stable', '--version'])
+    subprocess.check_call(["python -m", "pip install --upgrade pip"])
+    subprocess.check_call(["pip install", "pipenv"])
+    subprocess.check_call(["pipenv", "install"])
+
+    subprocess.run(
+        f'pipenv run python run_tests.py --admin-pass {args.admin_password} --analyst-pass {args.analyst_password} --slack-webhook-url {args.slack_webhook_url} --env {args.env} --file {args.file} --ci --processes 4 --enable-slack',
+        shell=True)
```
markhiveit, line 29: At first I wondered whether we could do this in python without resorting to subprocess calls, but having thought about it I think this is fine. I think if we're pragmatic, trying to do this without subprocess calls is more work and/or requires adding several tasks to DevOps. This way is quicker and simpler.

You could possibly argue that we should have just updated the bash script, but I personally think python is less bug prone than bash, so I think this is ok.

That said, others might disagree, so I'll hold back from approving until they've given an opinion.

tests/robot-tests/scripts/pipeline-run-rf-tests-snapshots.sh
```diff
@@ -1,53 +0,0 @@
...
-[[ "$analyst_pass" == "" ]] && { echo "Provide an analyst password with an '--analyst-pass PASS' argument"; exit 1; }
-[[ "$env" == "" ]] && { echo "Provide an environment with an '--env ENV' argument"; exit 1; }
-[[ "$file" == "" ]] && { echo "Provide a file/dir to run with an '--file FILE/DIR' argument"; exit 1; }
-
-google-chrome-stable --version
-
-python -m pip install --upgrade pip
-pip install pipenv
-pipenv install
-pipenv run python run_tests.py --admin-pass "$admin_pass" --analyst-pass "$analyst_pass" --slack-webhook-url "$slack_webhook_url" -e $env --ci --file $file --processes 3
```
benoutram, line 53: The difference between the two deleted bash script files seems to be that `--enable-slack` was missing here for the `pipeline-run-rf-tests-snapshots.sh` version. Is that important, since the `run_tests_pipeline.sh` includes it?

tests/robot-tests/scripts/run_tests_pipeline.py
```diff
@@ -0,0 +1,56 @@
...
+        help='slack webhook URL for sending test reports'
+    )
+
+    parser.add_argument(
+        '--env',
+        dest='env',
+        help=f'environment to run tests against (dev, test, pre-prod, prod & ci)'
+    )
+
+    parser.add_argument('--file', dest='file', help='directory to test')
```
benoutram, line 53: Would 'file or directory` be more helpful for the help text as I think it can be a file as well?

tests/robot-tests/scripts/pipeline-run-rf-tests-snapshots.sh
luke-h1: Good spot. So the snapshot tests use `SLACK_BOT_TOKEN` to send a message if the snapshots do not match. We'll also get alerted if the CI task for the snapshot tests fails. I'll add something to not pass `--enable-slack` if the snapshot tests are being ran

tests/robot-tests/scripts/run_tests_pipeline.py
```diff
@@ -0,0 +1,63 @@
...
+    subprocess.check_call(["python -m", "pip install --upgrade pip"])
+    subprocess.check_call(["pip install", "pipenv"])
+    subprocess.check_call(["pipenv", "install"])
+
+    # utility to not send slack test report if snapshot tests are being run
+    def should_send_test_reports() -> bool:
+        if args.file == 'tests/general_public/check_snapshots.robot':
+            return False
+        else:
+            return True
```
benoutram, line 32: I think with the comment it's still pretty clear what this does if you shorten it down to just the return statement like so

```
def should_send_test_reports() -> bool:
    return args.file != 'tests/general_public/check_snapshots.robot'
```

luke-h1: Nice, yeah that looks a lot cleaner

benoutram (approved):

PR #3428, opened 2022-08-09
EES-3581 Reduce number of footnotes in collapsible list before collapsing from 5 to 2

benoutram:
This PR reduces the number of footnotes shown in the collapsible list below data blocks before collapsing them from 5 to 2.

### DIFF
--- a/tests/robot-tests/tests/libs/common.robot
+++ b/tests/robot-tests/tests/libs/common.robot
@@ -423,6 +423,19 @@ user checks element is not visible
     [Arguments]    ${element}    ${wait}=${timeout}
     element should not be visible    ${element}    ${wait}
 
+user checks element is visually hidden
+    [Arguments]    ${selector}    ${parent}=css:body
+    user checks element has class
+    ...    ${selector}
+    ...    govuk-visually-hidden
+    ...    ${parent}
+
+user checks element has class
+    [Arguments]    ${selector}    ${class}    ${parent}=css:body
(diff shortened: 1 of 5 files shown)


tests/robot-tests/tests/libs/common.robot
```diff
@@ -423,6 +423,19 @@ user checks element is not visible
     [Arguments]    ${element}    ${wait}=${timeout}
     element should not be visible    ${element}    ${wait}
 
+user checks element is visually hidden
+    [Arguments]    ${selector}    ${parent}=css:body
+    user checks element has class
+    ...    ${selector}
+    ...    govuk-visually-hidden
+    ...    ${parent}
```
markhiveit, line 431: Selenium Library has keywords [Element Should Be Visible](https://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#Element%20Should%20Be%20Visible) and [Wait Until Element Is Visible](https://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#Wait%20Until%20Element%20Is%20Visible)

Although I don't think they worked for me well in the past, so not sure it's worth exploring, but I thought I'd mention it.

markhiveit (approved):

tests/robot-tests/tests/libs/common.robot
benoutram: Chatted to Mark to say the footnotes are still considered visible with these keywords due to the way the `gov-uk-visually-hidden` style is hiding them.

PR #3429, opened 2022-08-09
EES-3582 Remove Pre-release email validation requiring education.gov.uk email addresses

benoutram:
This PR removes the email validation on the Admin Pre-release user access page emails field which restricted email addresses to the education.gov.uk domain. Addresses no longer have any domain restriction.

It also changes the validation so that the error identifies the email address which is invalid. This is useful if a long list of addresses is entered containing an invalid value somewhere within it.

Update (suggested by @ntsim): A single error is returned only for the first invalid email to avoid creating a long error message in the case of multiple invalid email addresses. This might have also been quite annoying for screen reader users.

image

# UI test report

image

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
@@ -61,7 +61,7 @@ const PreReleaseUserAccessForm = ({
   const [invitePlan, setInvitePlan] = useState<PreReleaseInvitePlan>();
 
   const isValidEmail = (input: string) => {
-    return /^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i.test(input);
+    return /^[A-Z0-9.'_%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i.test(input);
   };
 
   const splitAndTrimLines = (input: string) =>
--- a/src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
@@ -26,7 +26,7 @@ const errorMappings = [
   mapFieldErrors<FormValues>({
     target: 'emails',
     messages: {
-      InvalidEmailAddress: 'Enter only @education.gov.uk email addresses',
+      InvalidEmailAddress: 'Enter valid email addresses',
       NoInvitableEmails:
         'All of the email addresses have already been invited or accepted',
     },
@@ -59,6 +59,10 @@ const PreReleaseUserAccessForm = ({
   ]);
 
   const [invitePlan, setInvitePlan] = useState<PreReleaseInvitePlan>();
+
+  const isValidEmail = (input: string) => {
+    return /^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i.test(input);
+  };
 
   const splitAndTrimLines = (input: string) =>
     input
@@ -119,7 +123,7 @@ const PreReleaseUserAccessForm = ({
               .trim()
               .required('Enter 1 or more email addresses')
               .test({
-                name: 'number of lines',
+                name: 'number-of-lines',
                 message: `Enter between 1 and ${inviteLimit} lines of email addresses`,
                 test: (value: string) => {
                   if (value) {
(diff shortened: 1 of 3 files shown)


src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
```diff
@@ -130,27 +134,27 @@ const PreReleaseUserAccessForm = ({
...
+                      path: 'emails',
+                      params: {
+                        value: emails[indexOfFirstInvalid],
+                      },
                     });
                   }
-                  return false;
+                  return true;
                 },
               }),
```
benoutram, line 159: It seems like we should be able to use `context: Yup.TestContext` here as a parameter and use arrow syntax rather than `this` but I struggled to get this to work. I think it's fixed in a more recent version of `@types/yup` but upgrading that led to other type errors elsewhere.

Example:

```
                test: (context: TestContext, value?: string) => {
                  if (value) {
                    const emails = splitAndTrimLines(value);
                    const indexOfFirstInvalid = emails.findIndex(
                      email => !isValidEmail(email),
                    );
                    if (indexOfFirstInvalid < 0) {
                      return true;
                    }
                    return context.createError({
                      path: 'emails',
                      params: {
                        value: emails[indexOfFirstInvalid],
                      },
                    });
                  }
                  return true;
                },
```

src/explore-education-statistics-admin/src/pages/release/pre-release/components/__tests__/PreReleaseUserAccessForm.test.tsx
```diff
@@ -36,11 +36,11 @@ describe('PreReleaseUserAccessForm', () => {
 
       const row1Cells = within(rows[1]).getAllByRole('cell');
 
-      expect(row1Cells[0]).toHaveTextContent('test1@education.gov.uk');
+      expect(row1Cells[0]).toHaveTextContent('test1@test.com');
```
amyb-hiveit, line 39: This was already here, but it'd be good to improve the performance of these tests by moving `getAllByRole` and `getAllByRole` to after awaits. In this case you could check for the text that appears above the table in the await then the rest after it, eg:
```
await waitFor(() => {
expect(screen.getByText('These people will have access to a preview of the release 24 hours before the scheduled publish date.').toBeInTheDocument();
})
 const rows = screen.getAllByRole('row');
 expect(rows).toHaveLength(3);
  etc...
```

There are a few other tests in this file that could have a similar improvement.

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
```diff
@@ -60,6 +60,10 @@ const PreReleaseUserAccessForm = ({
 
   const [invitePlan, setInvitePlan] = useState<PreReleaseInvitePlan>();
 
+  const isValidEmail = (input: string) => {
+    return /^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i.test(input);
+  };
+
```
amyb-hiveit, line 66: This validation should allow apostrophes, eg. amy.o'benson@hiveit.co.uk (I have friends with the surname O'Neil who are always complaining about this!)

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
amyb-hiveit: Shame about this, but I think you've done the right thing here to get it working. We should keep an eye out for when Yup v1 is ready and upgrade then.

amyb-hiveit (changes requested):
This works well, just a couple of small improvements that could be made.

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
benoutram: Thanks, good point. Made a change to accept this and added the unit test `accepts a range of valid values without causing a validation error` to check the special chars and some other valid styles of address.

src/explore-education-statistics-admin/src/pages/release/pre-release/components/__tests__/PreReleaseUserAccessForm.test.tsx
benoutram: Thanks, I've made some improvements in https://github.com/dfe-analytical-services/explore-education-statistics/pull/3429/commits/1c17ad19878460fdf4f69d1cb0cc29fa6669d3ba. Let me know if I've missed anything 🙂 .

amyb-hiveit (approved):

PR #3442, opened 2022-08-12
EES-3638 - fix general_public dev UI test failures

luke-h1:


benoutram (approved):
If this isn't running in Prod it shouldn't be too difficult to fix up the data in environments to be consistent, if you want to give that a try first?

luke-h1:
> If this isn't running in Prod it shouldn't be too difficult to fix up the data in environments to be consistent, if you want to give that a try first?

Yeah, that's a good shout. I suppose we could just go into each environment's content DB & manually edit these publish dates

PR #3443, opened 2022-08-12
EES-3600 Fix methodology content and annexe content section reordering

benoutram:
This PR fixes a bug in the Admin where content sections and annexe content sections could not be reordered on the methodology content page.

This was due to the `MethodologyVersionContent` related to the `MethodologyVersion` not being loaded when processing the request to reorder the sections.

It might be preferable that in future we change the `MethodologyContent` property to not be initialised by default.

Currently in `MethodologyVersion.cs`:

`public MethodologyVersionContent MethodologyContent { get; set; } = new();`

This could change to using the null-forgiving operator:

`public MethodologyVersionContent MethodologyContent { get; set; } = null!;`

This way we should observe a null `MethodologyVersionContent` if it hasn't been set due a bug where loading the related entity was missed, rather than being  initialised as an empty `MethodologyVersionContent` which was the case here.

Combined with changes like throwing an exception where the related entity is null would make it easier to spot a bug.

```
            if (methodologyVersion.MethodologyContent == null)
            {
                throw new ArgumentException("MethodologyContent must be hydrated");
            }
```

## Other changes

* Add missing unit tests for `MethodologyContentService.ReorderContentSections`.
* Fix a bug where the ordering request could be missing section one or more section id's but still be a valid.
* Enhance the UI tests to add multiple content sections and annexes when creating a methodology. This checks the public order of the sections once the methodology is published.

The UI test reorders them on an amendment and again checks the order once the amendment is published.

# UI test report

image

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyContentService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyContentService.cs
@@ -68,7 +68,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services.Methodologie
         public Task<Either<ActionResult, List<T>>> GetContentBlocks<T>(Guid methodologyVersionId) where T : ContentBlock
         {
             return _persistenceHelper
-                .CheckEntityExists<MethodologyVersion>(methodologyVersionId, q => 
+                .CheckEntityExists<MethodologyVersion>(methodologyVersionId, q =>
                     q.Include(version => version.MethodologyContent))
                 .OnSuccess(CheckCanViewMethodology)
                 .OnSuccess(methodology =>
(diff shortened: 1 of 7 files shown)


luke-h1 (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyContentService.cs
```diff
@@ -458,42 +464,44 @@ private List<ContentSectionViewModel> OrderedContentSections(List<ContentSection
...
         }
 
-        private Either<ActionResult, Tuple<MethodologyVersionContent, List<ContentSection>>> FindContentList(
-            MethodologyVersionContent content,
-            params Guid[] contentSectionIds)
+        private static List<ContentSection> GetContentListWithSection(MethodologyVersionContent content,
+            ContentSection contentSection)
         {
-            return FindContentList(content, contentSectionIds.ToList());
+            return FindContentListWithSectionId(content, contentSection.Id).Right.Item2;
```
markhiveit, line 470: I think you can do something like this and then make this line more readable than having `Item2`:
```
        private static Either<ActionResult, (MethodologyVersionContent Content, List<ContentSection> ContentSectionList)>
            FindContentListWithSectionId(MethodologyVersionContent content,
                Guid contentSectionId)
        {
            if (content.Content.Any(section => section.Id == contentSectionId))
            {
                return (Content: content, ContentSectionList: content.Content);
            }

            if (content.Annexes.Any(section => section.Id == contentSectionId))
            {
                return (Content: content, ContentSectionList: content.Annexes);
            }

            return new NotFoundResult();
        }
```

markhiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyContentService.cs
benoutram: I ended up rewriting `RemoveContentSection` to find the section to remove from the list again.

`var sectionToRemove = contentSections.Single(section => section.Id == contentSectionId);`

This isn't ideal because it's already in one of the other tuples, but I was then able to remove this extra method, and `FindContentListWithSectionId` remains consistent in its use of tuples with the other methods.

I think the use of tuples as method parameters and return types in this class is a bit crazy and it could be a lot simpler if we didn't use them. We'd probably just need to make more use of `OnSuccessCombineWith` to clean a lot of them up.

markhiveit (approved):

PR #3444, opened 2022-08-15
EES-3621 new table headers reordering

amyb-hiveit:
New UI for reordering table headers in the table tool. This affects:
- frontend table tool
- admin data blocks table tool
- admin release preview table tool
- admin prerelease table tool

This also fixes two accessibility problems with reordering table headers:
- EES-3015: Nested active focusable elements in table tool reordering
- EES-3026: Instructions provided to screen reader users do not adequately indicate how to access content

Summary of the changes:
- changed the toggle for seeing the form to be a button to make it more prominent
- only reordering groups or items in a single group can be active at once, when groups or items are inactive they don't have the role of button so there aren't nested active elements any more.
- by default reordering groups is enabled
- added move to rows / columns buttons to groups to make it clearer that you can do this and because moving them to the other axis with the keyboard doesn't work consistently.
- swapped columns and rows so column groups is at the top
- used names of indicators etc for groups instead of 'row group x'
- changed the drag icon and added it to groups (when they're draggable)
- updated the help text and added some screen reader announcements.
removing the nested active elements means that the drag n drop screen reader announcements work.
- I've fixed the width of the groups and made it scroll horizontally if they don't fit, previously they got squashed and could become unusable if there were several groups on a small screen. Note that there is a warning about nested scroll containers not being supported by react-beautiful-dnd but I've not found it to cause any problems.

Other notes:
- I've renamed and refactored a lot of files here as I found the previous naming confusing. I've tried to make the names and content of the files more specific.
- I've added a TableHeadersFormContext to avoid having to pass too many params down from the main component.
- I've updated ButtonText to have the same accessibility features as Button

ttreordering1

ttreordering2


src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersDraggableGroup.tsx
```diff
@@ -0,0 +1,69 @@
...
+              id={groupId}
+              legend={legend}
+              name={name}
+            />
+          </div>
+        )}
+      </Draggable>
+      {!groupDraggingActive && (
+        <TableHeadersGroupControls
+          axisName={name as string}
```
luke-h1, line 58: Don't think we need to cast `name` here

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersForm.tsx
```diff
@@ -1,41 +1,43 @@
 import Button from '@common/components/Button';
-import Details from '@common/components/Details';
 import { FormGroup } from '@common/components/form';
+import useMounted from '@common/hooks/useMounted';
+import TableHeadersAxis from '@common/modules/table-tool/components/TableHeadersAxis';
+import useTableHeadersContext, {
```
luke-h1, line 5: unused import here (`useTableHeadersContext`)

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersForm.tsx
```diff
@@ -50,114 +52,229 @@ const TableHeadersForm = ({
...
-    </Details>
+                      Update and view reordered table
+                    </Button>
+                  </Form>
+                );
+              }}
+            </Formik>
+          </div>
+
+          <div
```
luke-h1, line 260: Just wondering if we want to not render a div if there is no `screenReaderMessage` set?

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersForm.tsx
```diff
@@ -50,114 +52,229 @@ const TableHeadersForm = ({
...
+          >
+            {screenReaderMessage}
+          </div>
+        </>
+      )}
+    </TableHeadersContextProvider>
   );
 };
 
+// const TableHeadersForm = (props: TableHeadersFormProps) => (
```
luke-h1, line 274: Looks like there's a bit of unused stuff in this file

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersNotDraggableGroup.tsx
```diff
@@ -0,0 +1,49 @@
...
+            id={groupId}
+            isDraggable={false}
+            legend={legend}
+            name={name}
+          />
+        )}
+      </div>
+
+      <TableHeadersGroupControls
+        axisName={name as string}
```
luke-h1, line 40: Don't think we need to cast `name` here

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersReorderableList.tsx
```diff
@@ -0,0 +1,299 @@
+import DragIcon from '@common/components/DragIcon';
```
luke-h1, line 1: Looks like there is a couple of unused imports in this file (`useTableHeadersContext` & `useRef`)

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersReorderableList.tsx
```diff
@@ -0,0 +1,299 @@
...
+              <DragDropContext
+                onDragStart={start => {
+                  if (!selectedIndices.includes(start.source.index)) {
+                    setSelectedIndices([]);
+                  }
+
+                  setDraggingIndex(start.source.index);
+                }}
+                onDragEnd={result => {
+                  if (result.destination?.index == null) {
```
luke-h1, line 167: We should use triple equals here for strict equality

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersReorderableList.tsx
```diff
@@ -0,0 +1,299 @@
...
+                    nextValue.reduce<number[]>((acc, option, index) => {
+                      if (oldOptions.includes(option)) {
+                        acc.push(index);
+                      }
+
+                      return acc;
+                    }, []),
+                  );
+
+                  form.setFieldValue(name as string, nextValue);
```
luke-h1, line 197: Don't think we need to cast `name` here (but could be wrong). Not getting any Typescript errors about it in my editor

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/TableHeadersGroupControls.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+    });
+  });
+});
+
+function render({
+  activeList = undefined,
+  expandedLists = [],
+  totalItems = 2,
+}: {
+  activeList?: string | undefined;
```
luke-h1, line 103:
```suggestion
  activeList?: string;
```

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/TableHeadersNotDraggableGroup.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+  testCategoryFilters,
+  testIndicators,
+  testLocationFilters,
+  testTimePeriodFilters,
+} from '@common/modules/table-tool/components/__tests__/__data__/TableHeadersConfig.data';
+import { Formik } from 'formik';
+import { render as baseRender, screen, within } from '@testing-library/react';
+import noop from 'lodash/noop';
+import React from 'react';
+import _ from 'lodash';
```
luke-h1, line 14: Don't think we need this lodash import

src/explore-education-statistics-common/src/modules/table-tool/contexts/TableHeadersContext.tsx
```diff
@@ -0,0 +1,109 @@
+import useToggle from '@common/hooks/useToggle';
+import { noop } from 'lodash';
```
luke-h1, line 2: Not too certain but should we import noop like `import noop from 'lodash/noop';` to avoid importing the entire library here?

src/explore-education-statistics-common/src/modules/table-tool/contexts/TableHeadersContext.tsx
```diff
@@ -0,0 +1,109 @@
...
+import React, {
+  createContext,
+  ReactNode,
+  useContext,
+  useMemo,
+  useState,
+} from 'react';
+
+export interface TableHeadersContextState {
+  activeList: string | undefined;
```
luke-h1, line 12:
```suggestion
  activeList?: string;
```

luke-h1 (changes requested):

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersForm.tsx
amyb-hiveit: We do, with `aria-live` screen readers looks for changes in an existing div (or whatever element the `aria-live` is on) to know whether to announce the changes.

luke-h1: Ah right cool, that's good to know. Thanks

luke-h1:
Just had a look locally & it seems that there are a couple of final issues:

* It looks like when you move everything to rows (leaving no columns) and move an item from rows to columns, the 'Must have at least one column group' warning persists:

Screenshot 2022-08-16 at 12 52 26

* There's a beautiful-dnd warning when re-ordering Draggable elements (not sure whether this matters or not): 

```typescript 
react_devtools_backend.js:4026 react-beautiful-dndDroppable: unsupported nested scroll container detected.A Droppable can only have one scroll parent (which can be itself)Nested scroll containers are currently not supported.We hope to support nested scroll containers soon: https://github.com/atlassian/react-beautiful-dnd/issues/131👷‍ This is a development only message. It will be removed in production builds.
```

amyb-hiveit:
> Just had a look locally & it seems that there are a couple of final issues:
> 
> * It looks like when you move everything to rows (leaving no columns) and move an item from rows to columns, the 'Must have at least one column group' warning persists:
> 
> Screenshot 2022-08-16 at 12 52 26
> 
> * There's a beautiful-dnd warning when re-ordering Draggable elements (not sure whether this matters or not):
> 
> ```ts
> react_devtools_backend.js:4026 react-beautiful-dndDroppable: unsupported nested scroll container detected.A Droppable can only have one scroll parent (which can be itself)Nested scroll containers are currently not supported.We hope to support nested scroll containers soon: https://github.com/atlassian/react-beautiful-dnd/issues/131👷‍ This is a development only message. It will be removed in production builds.
> ```

Thanks @luke-h1

1. that's the form validation, it does it on the current version as well, but I'll have a look to see if it's fixable.
2. I think that warning is ok. I've made the axis groups horizontally scrollable so if you have lots of groups, or a small screen, you can scroll to access them all, this causes the warning but doesn't cause any problems as far as I can tell.

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/__data__/TableHeadersConfig.data.ts
```diff
@@ -0,0 +1,95 @@
+import {
+  CategoryFilter,
+  Indicator,
+  LocationFilter,
+  TimePeriodFilter,
+} from '@common/modules/table-tool/types/filters';
+import { TableHeadersConfig } from '@common/modules/table-tool/types/tableHeaders';
+
+export const testTimePeriodFilters = [
```
ntsim, line 9: Would add explicit types to all of the exported variables in this file

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/__data__/TableHeadersConfig.data.ts
```diff
@@ -0,0 +1,95 @@
+import {
```
ntsim, line 1: Test data file names are typically camelCased as they don't refer to a single thing e.g. `tableHeadersConfig.data.ts`

src/explore-education-statistics-common/src/modules/table-tool/contexts/TableHeadersContext.tsx
```diff
@@ -0,0 +1,109 @@
...
+  createContext,
+  ReactNode,
+  useContext,
+  useMemo,
+  useState,
+} from 'react';
+
+export interface TableHeadersContextState {
+  activeList?: string;
+  defaultNumberOfItems: number;
```
ntsim, line 13: Don't think we need this a context value. It creates an unnecessary dependency on the context and it'd be simpler for this to be controlled the components themselves via props.

src/explore-education-statistics-common/src/modules/table-tool/contexts/TableHeadersContext.tsx
```diff
@@ -0,0 +1,109 @@
...
+  expandedLists: string[];
```
ntsim, line 14: Not keen on tracking expanded list states in context for the following reasons:

- Toggling any group will cause all groups to re-render - this is a little wasteful
- It creates unnecessary coupling to the context for determining if a group can be expanded/collapsed.
- This is more complicated than just passing state via props.

Couldn't we track this state for each group individually without using context?

src/explore-education-statistics-common/src/modules/table-tool/contexts/TableHeadersContext.tsx
```diff
@@ -0,0 +1,109 @@
...
+  groupDraggingActive: boolean;
+  groupDraggingEnabled: boolean;
+  setActiveList: (listName?: string) => void;
+  toggleExpandedList: (listName: string) => void;
+  toggleGroupDraggingActive: () => void;
```
ntsim, line 19: Would suggest that this requires an explicit boolean parameter like `active`.

The main problem with just calling the toggle function without one is that if the user does things too quickly, the toggle state can end up inverted and the UI would end up looking messed up.

src/explore-education-statistics-common/src/modules/table-tool/contexts/TableHeadersContext.tsx
```diff
@@ -0,0 +1,109 @@
...
+  toggleGroupDraggingEnabled: () => void;
```
ntsim, line 20: Same as above - would suggest requiring a boolean parameter like `enabled` to determine the next state.

src/explore-education-statistics-common/src/modules/table-tool/contexts/TableHeadersContext.tsx
```diff
@@ -0,0 +1,109 @@
...
+import React, {
+  createContext,
+  ReactNode,
+  useContext,
+  useMemo,
+  useState,
+} from 'react';
+
+export interface TableHeadersContextState {
+  activeList?: string;
```
ntsim, line 12: Shouldn't this be `activeGroup` / `setActiveGroup`?

ntsim (changes requested):

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersReorderableList.module.scss
```diff
@@ -26,45 +30,46 @@
...
+  justify-content: space-between;
+  word-break: break-word;
 }
 
-.optionDraggedOutside {
-  border: 0;
-  opacity: 0.2;
-  outline: none;
+.dragIcon {
+  flex-shrink: 0;
```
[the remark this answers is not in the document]
amyb-hiveit: To prevent the icon shrinking when there's a lot of text in the list item.

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersReorderableList.module.scss
```diff
@@ -26,45 +30,46 @@
...
+  margin: 0.2rem 0 0 0.2rem;
```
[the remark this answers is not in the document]
amyb-hiveit: I've used the margin instead to keep the icon at a consistent place at the top of the item as it looks a better that way when the text wraps onto multiple lines.

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersDraggableGroup.tsx
```diff
@@ -0,0 +1,69 @@
...
+          >
+            <TableHeadersReadOnlyList
+              id={groupId}
+              legend={legend}
+              name={name}
+            />
+          </div>
+        )}
+      </Draggable>
+      {!groupDraggingActive && (
```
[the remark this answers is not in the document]
amyb-hiveit: The controls are removed because they're not in the draggable (as they would be nested buttons which causes accessibility problems) so get left behind when the groups move if they're not hidden. Agree it is a bit janky but I couldn't think of a better solution, any ideas?

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersGroupControls.module.scss
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  button {
+    flex-grow: 1;
+    margin-bottom: 0;
+  }
+
+  &.noShowMoreButton {
+    padding-top: govuk-spacing(9);
+  }
```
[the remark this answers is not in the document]
amyb-hiveit: It's so the groups without the 'show more' button are the same height as the ones with it as I though that looked better. (Though as I type that have realised that if there's only one option then it's shorter).

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersDraggableGroup.tsx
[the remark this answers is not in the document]
amyb-hiveit: No, the buttons would need to be in the Draggable to be dragged with the group, which we can't have because nested buttons aren't allowed. Disabling them on drag wouldn't make any difference as they need to be active when nothing is being dragged, so causing the nested buttons problem.

src/explore-education-statistics-common/src/modules/table-tool/contexts/TableHeadersContext.tsx
amyb-hiveit: This was because when I just had it on each group then when they were reordered the state wasn't maintained correctly. For example, if I expanded Group 1, then moved it to after Group 2, Group 2 would then be expanded and Group 1 wouldn't be.
Can you suggest a better way to avoid this?

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersDraggableGroup.tsx
```diff
@@ -0,0 +1,69 @@
...
+            className={classNames(styles.group, {
+              [styles.isDragging]: draggableSnapshot.isDragging,
+              [styles.isDraggingActive]:
+                !draggableSnapshot.isDragging && groupDraggingActive,
+              [styles.isDraggedOutside]:
+                draggableSnapshot.isDragging && !draggableSnapshot.draggingOver,
+            })}
+            ref={draggableProvided.innerRef}
+            role="button"
+            tabIndex={0}
```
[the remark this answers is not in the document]
amyb-hiveit: The list of items are there to provide context for what's in the group so it's clearer what you're moving, so I think having it read for screen reader users is ok, especially as by default it's only two items.
I don't think adding aria-label to the Draggable would help as it is makes it so you can't get to the list easily (or possibly at all, I'm not sure if it's just my poor NVDA skills, but I can't get it to read them out).
I'll be interested to see what DAC says about this though.

src/explore-education-statistics-common/src/components/ButtonText.module.scss
```diff
@@ -30,3 +31,7 @@
 .warning {
   color: govuk-colour('red');
 }
+
+.disabled:hover {
+  color: govuk-colour('blue');
+}
```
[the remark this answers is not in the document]
amyb-hiveit: Yep, I've made it light-blue when disabled, including on hover.

src/explore-education-statistics-admin/src/pages/release/content/components/ReleasePreviewTableToolFinalStep.tsx
```diff
@@ -44,19 +49,36 @@ const ReleasePreviewTableToolFinalStep = ({
...
+      {!showTableHeadersForm ? (
+        <div className="govuk-!-margin-bottom-3 dfe-flex dfe-justify-content--flex-end ">
+          <Button
+            className="govuk-!-margin-bottom-0"
+            ariaControls={tableHeadersFormId}
+            ariaExpanded={showTableHeadersForm}
+            onClick={toggleShowTableHeadersForm}
+          >
+            Move and reorder table headers
+          </Button>
```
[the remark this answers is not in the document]
amyb-hiveit: Good point, sorted now.

src/explore-education-statistics-common/src/components/ButtonText.tsx
```diff
@@ -14,28 +26,74 @@ interface Props {
...
+      ariaDisabled,
+      disabled,
+      disableDoubleClick,
+      isMountedRef,
+      onClick,
+      toggleClicking,
+    ],
+  );
+
+  const isDisabled = ariaDisabled || disabled || isClicking;
```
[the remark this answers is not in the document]
amyb-hiveit: Moved this into a hook as suggested.

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersAxis.tsx
```diff
@@ -0,0 +1,129 @@
...
+                {field.value.map((group: Filter[], index: number) => {
+                  const key = `group-${index}`;
+
+                  return (
+                    <div
+                      className={styles.groupContainer}
+                      data-testid={`${name}-${index}`}
+                      key={key}
+                    >
+                      <TableHeadersDraggableGroup
```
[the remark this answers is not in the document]
amyb-hiveit: Not sure I can do anything about this, I've tried a few things but end up making it misaligned in other ways.

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/TableHeadersAxis.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+    expect(
+      screen.getByRole('group', { name: 'Test legend' }),
+    ).toBeInTheDocument();
+
+    const buttons1 = within(screen.getByTestId('rowGroups-0')).getAllByRole(
+      'button',
+    );
+    // Draggable groups have the role button and are the first button in the container.
+    const draggableGroup1 = within(buttons1[0]);
```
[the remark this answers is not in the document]
amyb-hiveit: I've added testids to the groups and updated the tests to use them.

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersAxis.tsx
```diff
@@ -0,0 +1,129 @@
...
+function TableHeadersAxis({ id, legend, name, onMoveGroupToOtherAxis }: Props) {
+  const {
+    groupDraggingActive,
+    groupDraggingEnabled,
+  } = useTableHeadersContext();
+  const [field, meta] = useField(name);
+
+  if (groupDraggingEnabled) {
+    return (
+      <Droppable droppableId={name} direction="horizontal">
```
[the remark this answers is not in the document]
amyb-hiveit: It had all got in a bit of a tangle, but I went a bit overboard untangling it :) It's simplified now and just has one `Table HeadersGroup`.

src/explore-education-statistics-common/src/modules/table-tool/contexts/TableHeadersContext.tsx
ntsim: This will be due to re-rendering groups in completely different parts of the DOM i.e. the components are being unmounted/re-mounted with completely fresh state.

You're right that there there probably isn't a way to maintain state locally! Happy to go with the current solution :+1:

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersDraggableGroup.tsx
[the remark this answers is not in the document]
ntsim: Yep, thought that'd be the case. It's a little bit of a tricky one tbh. I can imagine that if the user expands the list, every time they refocus the group, it could be annoying having them all read out though.

Guess we can just press on for now and just see what happens from the re-test.

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersDraggableGroup.tsx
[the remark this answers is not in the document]
ntsim: Yep, gave it a little play and I think we can do something like this :smile: 
```tsx
<Draggable>
  {(draggableProvided, draggableSnapshot) => (
    <div
      {...draggableProvided.draggableProps}
      ref={draggableProvided.innerRef}
    >
      <div 
        {...draggableProvided.dragHandleProps}
      >
      </div>
      <TableHeadersGroupControls ... />
    </div>
  )}
</Draggable>
```

Should give us this:

image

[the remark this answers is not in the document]
amyb-hiveit: That totally works, I'm kicking myself for not thinking of it! I've moved the focus to be around the whole group as well.

amyb-hiveit:
> > * It looks like when you move everything to rows (leaving no columns) and move an item from rows to columns, the 'Must have at least one column group' warning persists:

@luke-h1 this validation should only show when you try to submit the form now.

ntsim (approved):

PR #3445, opened 2022-08-16
EES-3623 - remove userPublicationRoles, userReleaseRoles endpoints

luke-h1:


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserRoleService.cs
```diff
@@ -428,6 +429,57 @@ await _usersAndRolesPersistenceHelper
...
+            return await _userService.CheckCanManageAllUsers()
+                .OnSuccess(async _ =>
+                {
+                    return await _contentPersistenceHelper
+                        .CheckEntityExists<User>(userId)
+                        .OnSuccess(async _ =>
+                        {
+
+                            var userReleaseRoles =
+                                await _contentDbContext.UserReleaseRoles.Where(urr => urr.UserId == userId)
```
duncan-at-hiveit, line 443: I'd say that the method name might be misleading here, or we're missing some checks in our code.

If the intention is to clear down a user's Prerelease Roles across the board: 

You'll need to add a check for `urr.Role == PreRelease` as well, as otherwise this'll clear down ALL of the user's ReleaseRoles for both prerelease and non-prerelease role types.

It'd be good to have a unit test that covers this scenario - i.e. set up a user with various ReleaseRoles, some of which are prerelease and some which are not, and then check that the non-prerelease ones have been correctly left alone.

duncan-at-hiveit: Or if the intention is to remove all Roles from a User, who at the current time is just a prerelease user:

I'd change the method name to `RemoveAllRolesFromPreReleaseUser`.

I'd then do a check to make sure that the user is legitimately only a Prerelease User, by ensuring that they have the Prerelease User global role.  If not, then this method should throw a wobbly.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
```diff
@@ -438,6 +438,9 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             modelBuilder.Entity<UserPublicationRole>()
                 .Property(r => r.Role)
                 .HasConversion(new EnumToStringConverter<PublicationRole>());
+            
+            modelBuilder.Entity<UserPublicationRole>()
```
duncan-at-hiveit, line 442: Good catch!

tests/robot-tests/tests/libs/admin_api.py
```diff
@@ -161,6 +162,40 @@ def user_adds_publication_role_to_user_via_api(
         f'Adding publication role to user API request failed with {response.status_code} and {response.text}'
 
 
+def user_removes_roles_from_prerelease_user_via_api(user_id: str) -> None:
+    response = admin_client.delete(
+        f'/api/user-management/user/{user_id}/release-role/pre-release'
```
duncan-at-hiveit, line 167: Given that this endpoint specifically mentions `release-role`, I'd not expect it to delete any UserPublicationRoles from the User - going back to my other comments on the `RemoveUserPreReleaseRoles` method though, I think we'd be removing that behaviour anyway.

tests/robot-tests/tests/libs/admin_api.py
```diff
@@ -161,6 +162,40 @@ def user_adds_publication_role_to_user_via_api(
...
+
+
+def user_removes_prerelease_user_roles_if_needed_via_api(user_emails: list) -> None:
+    allowed_users = [
+        'ees-prerelease1@education.gov.uk',
+        'ees-prerelease2@education.gov.uk',
+        'ees-prerelease3@education.gov.uk']
+
+    if len(set(user_emails).intersection(allowed_users)) < 2:
+        raise AssertionError('`User emails` must contain at least two allowed prerelease users')
```
duncan-at-hiveit, line 180: Why at least 2?  Could we not call this with just a singleton list?

Shows you've been working around code too long when you call a list with one item in it "singleton" :)

tests/robot-tests/tests/libs/admin_api.py
```diff
@@ -161,6 +162,40 @@ def user_adds_publication_role_to_user_via_api(
...
+            BuiltIn().log('User with email {} does not exist'.format(user_email))
+
+        try:
+            result = [user_removes_roles_from_prerelease_user_via_api(user_id) for user_id in user_ids]
+            assert(all(result))
+
+        except AssertionError:
+            BuiltIn().log(f'Failed to remove roles from user {user_email}')
+
+    BuiltIn().log(f'All userReleaseRoles & userPublicationRoles removed from prerelease user: {user_email}')
```
duncan-at-hiveit, line 196: Again, I wouldn't expect a prerelease user to have any publication-level roles, or else they shouldn't be a prerelease user :)

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserRoleService.cs
duncan-at-hiveit: Throw a wobbly === `return ValidationActionResult(NameOfYourErrorMessageHere);`.

Have a search for "return ValidationActionResult" in the IDE and you'll see plenty of examples :+1:

luke-h1: Cool, yeah I get why the method name is misleading. The pre-release users that the UI tests could theoretically 
become an analyst, bau, etc if someone ends up changing their role.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserRoleService.cs
```diff
@@ -428,6 +429,57 @@ await _usersAndRolesPersistenceHelper
...
+                                );
+                                    
+                            }
+
+                            var userPublicationRoles =
+                                await _contentDbContext.UserPublicationRoles
+                                    .Where(upr => upr.UserId == userId)
+                                    .ToListAsync();
+                            
+                            if (userPublicationRoles.Any())
```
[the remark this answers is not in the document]
luke-h1: Think I may have confused the intention behind this endpoint a bit with the naming. When we hit this endpoint, we're assuming that the designated pre-release users have been used manually. There's not a problem with this but to stop certain tests from failing we want to check if they have release or publication roles. If they do, clear them down and downgrade them to 'no role'

tests/robot-tests/tests/libs/admin_api.py
luke-h1: I've renamed this to `user_removes_all_release_and_publication_roles_from_user` now

tests/robot-tests/tests/libs/admin_api.py
luke-h1: No reason really, I've removed this restriction now

tests/robot-tests/tests/libs/admin_api.py
luke-h1: I've renamed this to `BuiltIn().log(f'All userReleaseRoles & userPublicationRoles reset for user: {user_email}')` now

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/UserManagement/UserManagementController.cs
```diff
@@ -87,7 +87,21 @@ public async Task<ActionResult<Unit>> DeleteUserReleaseRole(Guid id)
...
                 .HandleFailuresOrOk();
         }
-
+        
+        /// <summary>
+        ///   Removes all userReleaseRoles and userPublicationRoles for the given user.
+        /// </summary>
+        /// <param name="userId"></param>
+        /// <returns></returns>
+        [HttpDelete("user/{userId:guid}/release-publication-roles/all")]
```
duncan-at-hiveit, line 96: I would update this name to `[HttpDelete("user/{userId:guid}/resource-roles/all")]`, as ResourceRole is the common base class of both ReleaseRole and PublicationRole.  This way, this name is a bit more RESTful.

Similarly, `DeleteAllUserReleaseAndUserPublicationRoles` can become `DeleteAllUserResourceRoles`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/IUserRoleService.cs
```diff
@@ -34,5 +34,8 @@ public interface IUserRoleService
         Task<Either<ActionResult, Unit>> RemoveUserPublicationRole(Guid userPublicationRoleId);
 
         Task<Either<ActionResult, Unit>> RemoveUserReleaseRole(Guid userReleaseRoleId);
+        
+        Task<Either<ActionResult, Unit>> removeAllUserReleaseAndPublicationRoles(Guid userId);
```
duncan-at-hiveit, line 38: Don't forget to uppercase the first letter :)

duncan-at-hiveit: And again, suggest this becomes `RemoveAllUserResourceRoles`.

tests/robot-tests/tests/libs/admin_api.py
```diff
@@ -161,6 +162,39 @@ def user_adds_publication_role_to_user_via_api(
         f'Adding publication role to user API request failed with {response.status_code} and {response.text}'
 
 
+def user_removes_all_release_and_publication_roles_from_user(user_id: str) -> None:
+    response = admin_client.delete(
+        f'/api/user-management/user/{user_id}/release-publication-roles/all'
```
duncan-at-hiveit, line 167: Again in line with my other comments, I'd update this URL to be `resource-roles` rather than `release-publication-roles`.

PR #3450, opened 2022-08-16
EES-3554 Update Admin API for fetching publication releases

markhiveit:
This PR makes several changes:
- Remove MyReleaseViewModel - This involved moving `PermissionsSet` from MyReleaseViewModel to ReleaseViewModel and making it nullable. Because it's now optional whether PerrmissionsSet is returned, it is no longer set by AutoMapper/MappingProfile.cs, which always set these values, and is now done by methods in a new `PermissionsUtils` class. 

- Another consequence of removing MyReleaseViewModel was that release and publication repositories that were returning view models now need to return `Release`s or `Publication`s. This is because work AutoMapper was previously doing inside repositories would now require adding `UserService` as a dependency to those repositories, if they were going to continue to return view models. So instead the repositories don't return view models and the relevant service creates/hydrates the view models.

- Added ability to filter releases based whether they have been published - `GetActiveReleases` can now return all live releases only, or all nonlive releases, or all releases regardless of live status. Whether live or nonlive releases are returned is determined by the new `live` boolean query parameter. All releases regardless of published status are returned if no `live` query parameter is provided.

- Added pagination when fetching publication releases - There is a detailed comment on EES-3554's Jira ticket detailing why we still get all the results from the database rather than only fetching the results for the page required. The endpoint returns all results on a single page if no `page` or `pageSize` query parameters are provided.

Other notes:
- I've also fixed the bug EES-3536 on this PR - ReleaseApprovers weren't being shown all releases they had permission to see.
- I've also fixed a redundant `onRemoveSection` from `ReorderableAccordionSection`. Credit goes to Luke for spotting this.
- I've not added more of the data that is needed by the frontend to `ReleaseListItem` yet. As this can be easily be done in its own PR, I figure there is no need to further complicate this PR.


src/explore-education-statistics-admin/src/services/publicationService.ts
```diff
@@ -135,9 +131,20 @@ const publicationService = {
...
+    live: boolean | undefined = undefined,
+  ): Promise<PaginatedList<ReleaseListItem>> {
+    let params = {};
+    if (page !== undefined) params = { ...params, page };
+    if (pageSize !== undefined) params = { ...params, pageSize };
+    if (live !== undefined) params = { ...params, live };
+
+    return client.get<PaginatedList<ReleaseListItem>>(
       `/publication/${publicationId}/releases`,
+      { params },
```
amyb-hiveit, line 147: This could be simplified by making the params an optional object that gets passed in, something like:
```
export interface getReleasesParams {
  live?: boolean;
  page?: number;
  pageSize?: number;
}

 getReleases(
    publicationId: string,
    params?: getReleasesParams,
  ): Promise<PaginatedList<ReleaseListItem>> {
    return client.get<PaginatedList<ReleaseListItem>>(
      `/publication/${publicationId}/releases`,
      { params },
    );
  },
  ```
  
  which could then be called with or without the params:
  ```
  await publicationService.getReleases(
      publicationId,
      { page: 2 },
    );
   
    await publicationService.getReleases(publicationId);

File: src/explore-education-statistics-admin/src/pages/publication/PublicationManageTeamAccessPage.tsx [hunk 2]
```diff
@@ -35,10 +35,11 @@ const PublicationManageTeamAccessPage = ({
   const [currentReleaseId, setCurrentReleaseId] = useState(releaseId ?? '');
 
   const { value: model, isLoading } = useAsyncHandledRetry<Model>(async () => {
-    const [releases, publication] = await Promise.all([
+    const [paginatedReleases, publication] = await Promise.all([
```
amyb-hiveit, line 38: I'd maybe not call these paginatedReleases when we're getting all of them just because it's a bit confusing, I was reading this and thinking that it'd just get a page of them (or possibly I've not had enough coffee yet!).

markhiveit: Discussed this with Amy on Slack. Could go for `allReleases` or `allReleasesSinglePage`. Not sure if others have an opinion?

src/GovUk.Education.ExploreEducationStatistics.Common/ViewModels/PaginatedListViewModel.cs
```diff
@@ -0,0 +1,70 @@
...
+                var pageResults = allResults
+                    .Skip((page.Value - 1) * pageSize.Value)
+                    .Take(pageSize.Value)
+                    .ToList();
+
+                paginatedViewModel = new PaginatedListViewModel<T>(
+                    pageResults, allResults.Count, page.Value, pageSize.Value);
+            }
+
+            if (page <= 0 || paginatedViewModel.Paging.Page > paginatedViewModel.Paging.TotalPages)
```
luke-h1, line 50: Just wondering if we want to return some default results here if consumers enter dodgy paging sizes instead of 404ing?

luke-h1:
Looks like making release permissions potentially undefined has introduced a lot of Eslint non-null assertion warnings. Would suggest adding: 

```typescript 
// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
or using optional chaining where appropriate in order to reduce the noise Eslint produces

src/GovUk.Education.ExploreEducationStatistics.Common/ViewModels/PaginatedListViewModel.cs
markhiveit: My thought is that it's better to return no results or respond with an error in these situations. This protects us in the case where the frontend has a bug - we want the backend to reveal that bug rather than potentially hide it.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
```diff
@@ -32,6 +32,948 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
...
     {
+        [Fact]
+        public async Task GetMyPublicationsAndReleasesByTopic_CanViewAllReleases_ReleaseOrder()
+        {
+            var topic = new Topic
+            {
+                Id = Guid.NewGuid(),
+            };
+
+            // The order they should appear in result
```
duncan-at-hiveit, line 43: Might be worth saying in this comment what drives this order, as for me as a casual reader of this test, I don't know why it should be this order :)

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
```diff
@@ -32,6 +32,948 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
...
+                    publication2Release2,
+                    publication2Release1,
+                },
+            };
+
+
+            var publicationRepository = new Mock<IPublicationRepository>(Strict);
+
+            publicationRepository.Setup(s => s.GetAllPublicationsForTopic(topic.Id))
+                .ReturnsAsync(
```
duncan-at-hiveit, line 106: Can use `ReturnsAsync(ListOf(publication1, publication2))` as a space-saver.

duncan-at-hiveit: (where `ListOf` exists in `CollectionUtils` - not to be confused with `CollectionsUtil` :joy: )

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
```diff
@@ -32,6 +32,948 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
...
+                {
+                    release,
+                },
+                Methodologies = AsList(
+                    new PublicationMethodology
+                    {
+                        Methodology = new Methodology
+                        {
+                            Slug = "methodology-2-slug",
+                            Versions = AsList(methodology2Version2, methodology2Version1)
```
duncan-at-hiveit, line 204: Wherever we see "AsList", try to replace with "ListOf" from within the same util class!  We're trying to phase out the former in favour of the latter.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
```diff
@@ -62,11 +62,14 @@ public async Task<ActionResult<List<PublicationSummaryViewModel>>> ListPublicati
         }
 
         [HttpGet("api/publication/{publicationId}/releases")]
-        public async Task<ActionResult<List<ReleaseViewModel>>> ListActiveReleases(
-            [Required] Guid publicationId)
+        public async Task<ActionResult<PaginatedListViewModel<ReleaseListItemViewModel>>> ListActiveReleases(
+            [Required] Guid publicationId,
+            [FromQuery] int? page = null,
```
duncan-at-hiveit, line 67: Might be that as we add more pagination to the codebase that we have a `record` called something like `PaginationParams` which holds `page` and `pageSize` within it.

Then it also makes it easier to pass these multiple parameters around through from Controller methods to Service methods to Repository methods.

Also then makes it easier to add new pagination parameters in the future in one place if necessary.

Doesn't have to be done as part of this PR (or at all in fact), but might be worth considering.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/IPublicationService.cs
```diff
@@ -28,8 +28,16 @@ public interface IPublicationService
 
         Task<Either<ActionResult, PublicationViewModel>> GetPublication(Guid publicationId);
 
-        Task<Either<ActionResult, List<ReleaseViewModel>>> ListActiveReleases(Guid publicationId);
+        Task<Either<ActionResult, PaginatedListViewModel<ReleaseListItemViewModel>>> ListActiveReleasesPaginated(
```
duncan-at-hiveit, line 31: I wonder if we bother keeping both `ListActiveReleasesPaginated` and `ListActiveReleases`, or instead just have one method with the pagination support and just let code that needs all of them call it with "unlimited" page size?

duncan-at-hiveit: So in the case of my other comment about bundling `page` and `pageSize` into a `PaginationParams` record, we could have a factory method on `PaginationParams` called something like `NoPagination`.  Thus a caller would say:

`var allReleases = publicationService.ListActiveReleases(publicationId, PaginationParams.NoPagination());`

or:

`var paginatedReleases = publicationService.ListActiveReleases(publicationId, paginationParams);`

depending on the use case?

duncan-at-hiveit: Or better, a singleton called `NoPagination` within `PaginationParams` that represents no pagination. Thus callers say:

`var allReleases = publicationService.ListActiveReleases(publicationId, PaginationParams.NoPagination);`

duncan-at-hiveit: OOOOR!

Given that all pagination parameters could now be in a single nullable parameter:

```
Task<Either<ActionResult, PaginatedListViewModel<ReleaseListItemViewModel>>> ListActiveReleases(            
            Guid publicationId,
            PaginationParams pagination,
            bool? live = null);
```

a caller would say:

`var allReleases = publicationService.ListActiveReleases(publicationId);`

or:

`var paginatedReleases = publicationService.ListActiveReleases(publicationId, paginationParams);`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -357,5 +379,28 @@ public static IQueryable<Publication> HydratePublicationForPublicationViewModel(
                 .ThenInclude(p => p.Methodology)
                 .ThenInclude(p => p.Versions);
         }
+
+        private List<MyPublicationViewModel> HydrateMyPublicationsViewModel(List<Publication> publications)
```
duncan-at-hiveit, line 383: Given it's returning a list, I'd call this `HydrateMyPublicationsViewModels`

duncan-at-hiveit: And given that it;s also doing ordering behaviour, I'd probably either mention that in the method title or have whatever needs the ordering to do the ordering itself after calling this.

So maybe either rename to `HydrateAndOrderMyPublicationsViewModels`, or remove the ordering and have the caller do it.

Not a biggie tbh, but showing your intent / side effects with the method names is a good habit to get into :)

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ReleaseViewModels.cs
```diff
@@ -51,18 +52,31 @@ public class ReleaseViewModel
...
-        public string LatestInternalReleaseNote { get; set; }
+        public string? LatestInternalReleaseNote { get; set; }
 
         public bool Amendment { get; set; }
 
         public Guid? PreviousVersionId { get; set; }
+
+        public PermissionsSet? Permissions { get; set; }
+
+        public class PermissionsSet
```
duncan-at-hiveit, line 70: Things like this can probably be made a `record` rather than a class, but appreciate this was probably just moved so it's existing code.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/ViewModels/PaginatedListViewModelTests.cs
```diff
@@ -0,0 +1,161 @@
...
+        Assert.Equal(page, paginatedViewModel.Paging.Page);
+        Assert.Equal(pageSize, paginatedViewModel.Paging.PageSize);
+        Assert.Equal(allResults.Count, paginatedViewModel.Paging.TotalResults);
+        Assert.Equal(1, paginatedViewModel.Paging.TotalPages);
+    }
+
+    [Fact]
+    public void PaginatedListViewModel_NoResults()
+    {
+        var allResults = new List<int>();
```
duncan-at-hiveit, line 131: I'd call this `noResults` to be explicit about what it represents in the test personally.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/ViewModels/PaginatedListViewModelTests.cs
```diff
@@ -0,0 +1,161 @@
...
+    {
+        var allResults = new List<int>{ 1, 2, 3 };
+
+        var page = 2;
+        var pageSize = 10;
+
+        var result = PaginatedListViewModel<int>.Create(
+            allResults, page, pageSize);
+
+        result.AssertNotFound();
```
duncan-at-hiveit, line 159: Noice

src/GovUk.Education.ExploreEducationStatistics.Common/ViewModels/PaginatedListViewModel.cs
```diff
@@ -0,0 +1,70 @@
+#nullable enable
+using System.Collections.Generic;
+using System.Linq;
+using GovUk.Education.ExploreEducationStatistics.Common.Model;
+using Microsoft.AspNetCore.Mvc;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.ViewModels
+{
```
duncan-at-hiveit, line 8: For new classes, we're adopting the non-indented version of the namespace syntax.  So the majority of the class doesn't need to be indented.

src/GovUk.Education.ExploreEducationStatistics.Common/ViewModels/PaginatedListViewModel.cs
```diff
@@ -0,0 +1,70 @@
...
+            if (page <= 0 || paginatedViewModel.Paging.Page > paginatedViewModel.Paging.TotalPages)
+            {
+                return new NotFoundResult();
+            }
+
+            return paginatedViewModel;
+        }
+    }
+
+    public class PagingViewModel
```
duncan-at-hiveit, line 59: This can be a record

src/GovUk.Education.ExploreEducationStatistics.Common/ViewModels/PaginatedListViewModel.cs
```diff
@@ -0,0 +1,70 @@
+#nullable enable
+using System.Collections.Generic;
+using System.Linq;
+using GovUk.Education.ExploreEducationStatistics.Common.Model;
+using Microsoft.AspNetCore.Mvc;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.ViewModels
+{
+    public class PaginatedListViewModel<T>
```
duncan-at-hiveit, line 9: This is a perfect candidate for a record!

src/explore-education-statistics-admin/src/pages/publication/PublicationManageTeamAccessPage.tsx
duncan-at-hiveit: I'd probably vote for `allReleases` in this instance

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
duncan-at-hiveit: Or `PaginationRequest` rather than `PaginationParams` - it's a bit more explicit that it's something to do with a request rather than a response.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ReleaseViewModels.cs
```diff
@@ -18,21 +19,21 @@ public class ReleaseViewModel
...
+        public string? PublicationTitle { get; set; }
 
-        public string PublicationSlug { get; set; }
+        public string? PublicationSlug { get; set; }
 
-        public string ReleaseName { get; set; }
+        public string? ReleaseName { get; set; }
 
-        public string YearTitle { get; set; }
+        public string? YearTitle { get; set; }
```
ntsim, line 34: Why are all of these optional? These should be guaranteed to have values, so this is a strange change.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ReleaseViewModels.cs
```diff
@@ -51,18 +52,31 @@ public class ReleaseViewModel
         [JsonConverter(typeof(StringEnumConverter))]
         public ReleaseType Type { get; set; }
 
-        public Contact Contact { get; set; }
+        public Contact? Contact { get; set; }
```
ntsim, line 55: Same as above - why has this been made optional?

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ReleaseViewModels.cs
```diff
@@ -98,52 +112,52 @@ public class ReleaseCreateViewModel
...
-        [RegularExpression(@"^([0-9]{4})?$")] public string ReleaseName { get; set; }
+        [RegularExpression(@"^([0-9]{4})?$")] public string? ReleaseName { get; init; }
 
         public string Slug => SlugFromTitle(Title);
 
-        private string Title => Format(Year, TimePeriodCoverage);
+        private string? Title => Year.HasValue ? Format(Year.Value, TimePeriodCoverage) : null;
 
-        private int Year => int.Parse(ReleaseName);
+        private int? Year => ReleaseName != null ? int.Parse(ReleaseName) : null;
```
ntsim, line 123: Why are these optional? 

I'm not going to ask this again because we seem to be making the same kind of things optional everywhere. To me, this seems incorrect as these things are not optional and should have been set in the frontend.

Same questions for `ReleaseUpdateViewModel` as well.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ReleaseViewModels.cs
```diff
@@ -98,52 +112,52 @@ public class ReleaseCreateViewModel
             }
         }
 
-        [PartialDateValidator] public PartialDate NextReleaseDate { get; set; }
+        [PartialDateValidator] public PartialDate? NextReleaseDate { get; set; }
 
-        [RegularExpression(@"^([0-9]{4})?$")] public string ReleaseName { get; set; }
+        [RegularExpression(@"^([0-9]{4})?$")] public string? ReleaseName { get; init; }
```
ntsim, line 117: Why is this optional?

src/explore-education-statistics-admin/src/services/releaseService.ts
```diff
@@ -56,6 +53,25 @@ export interface ReleaseSummary {
...
+  results: T[];
+  paging: Pagination;
+}
+
+export interface Pagination {
+  page: number;
+  pageSize: number;
+  totalResults: number;
+  totalPages: number;
+}
```
ntsim, line 73: These should be in a common types file such as `services/types/pagination.ts`

src/explore-education-statistics-admin/src/services/publicationService.ts
```diff
@@ -94,6 +94,12 @@ export interface SavePublicationRequest {
   topicId: string;
 }
 
+export interface getReleasesParams {
```
ntsim, line 97: Types should be PascalCased i.e. `GetReleaseParams`. Also would consider naming this `ListReleasesParams`

src/explore-education-statistics-admin/src/services/publicationService.ts
```diff
@@ -135,9 +141,13 @@ const publicationService = {
     return client.get<MyPublication>(`/me/publication/${publicationId}`);
   },
 
-  getReleases(publicationId: string): Promise<ReleaseSummary[]> {
-    return client.get<ReleaseSummary[]>(
+  getReleases(
```
ntsim, line 144: Would consider renaming this `listReleases`

src/explore-education-statistics-admin/src/pages/publication/components/__tests__/PublicationScheduledReleases.test.tsx
```diff
@@ -114,7 +114,8 @@ describe('PublicationScheduledReleases', () => {
           {
             ...testRelease1,
             permissions: {
-              ...testRelease1.permissions,
+              // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+              ...testRelease1.permissions!,
```
ntsim, line 118: If we separate the permissions out into a separate variable with an explicit type, we don't need this non-null assertion

src/explore-education-statistics-admin/src/pages/publication/components/__tests__/PublicationDraftReleases.test.tsx
```diff
@@ -210,7 +210,8 @@ describe('PublicationDraftReleases', () => {
           {
             ...testRelease1,
             permissions: {
-              ...testRelease1.permissions,
+              // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+              ...testRelease1.permissions!,
```
ntsim, line 214: If we separate the permissions out into a separate variable with an explicit type, we don't need this non-null assertion. Same for below as well.

src/explore-education-statistics-admin/src/pages/publication/components/__tests__/PublicationPublishedReleases.test.tsx
```diff
@@ -318,7 +318,8 @@ describe('PublicationPublishedReleases', () => {
           {
             ...testRelease1,
             permissions: {
-              ...testRelease1.permissions,
+              // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+              ...testRelease1.permissions!,
```
ntsim, line 322: If we separate the permissions out into a separate variable with an explicit type, we don't need this non-null assertion

src/explore-education-statistics-admin/src/pages/admin-dashboard/components/__tests__/ScheduledReleasesTable.test.tsx
```diff
@@ -145,7 +145,8 @@ describe('ScheduledReleasesTable', () => {
             {
               ...testReleases[0],
               permissions: {
-                ...testReleases[0].permissions,
+                // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                ...testReleases[0].permissions!,
```
ntsim, line 149: If we separate the permissions out into a separate variable with an explicit type, we don't need this non-null assertion

src/explore-education-statistics-admin/src/pages/admin-dashboard/components/__tests__/DraftReleasesTable.test.tsx
```diff
@@ -370,7 +379,8 @@ describe('DraftReleasesTable', () => {
             {
               ...testReleases[0],
               permissions: {
-                ...testReleases[0].permissions,
+                // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                ...testReleases[0].permissions!,
```
ntsim, line 383: If we separate the permissions out into a separate variable with an explicit type, we don't need this non-null assertion. Same for below.

src/explore-education-statistics-admin/src/pages/admin-dashboard/components/__tests__/NonScheduledReleaseSummary.test.tsx
```diff
@@ -108,8 +109,10 @@ describe('NonScheduledReleaseSummary', () => {
         <NonScheduledReleaseSummary
           includeCreateAmendmentControls
           release={produce(testRelease, draft => {
-            draft.permissions.canUpdateRelease = true;
-            draft.permissions.canMakeAmendmentOfRelease = true;
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            draft.permissions!.canUpdateRelease = true;
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            draft.permissions!.canMakeAmendmentOfRelease = true;
```
ntsim, line 115: All of these non-null assertions are a little nasty. Would consider creating a new extended `ReleaseWithPermissions` type where `permissions` are non-null and use it for the `testRelease`

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/publication/PublicationManageTeamAccessPage.tsx
markhiveit: Done

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationRepositoryTests.cs
```diff
@@ -354,22 +354,18 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contextId))
                 Assert.NotNull(publication);
                 Assert.Equal(3, publication.Methodologies.Count);
 
-                // Check that the latest versions of the methodologies are returned in title order 
-
-                var link1 = publication.Methodologies[0];
-                Assert.True(link1.Owner);
-                Assert.Equal(methodology1Id, link1.Methodology.Id);
-                Assert.Equal("Methodology 1", link1.Methodology.Title);
+                // Methodologies aren't ordered
```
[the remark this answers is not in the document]
markhiveit: They are returned in the order they were added to the in-memory db. They were previously ordered by Automapper in the repository, but now `map` is called in the service.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationRepositoryTests.cs
```diff
@@ -1014,28 +951,22 @@ await using (var context = InMemoryApplicationDbContext(contextId))
...
+                Assert.Equal(publication.TopicId, resultPublication.TopicId);
+                Assert.Equal(publication.Contact.ContactName, resultPublication.Contact.ContactName);
+
+                Assert.Equal(2, resultPublication.Releases.Count);
+
+                Assert.Single(resultPublication.Methodologies);
+
+                Assert.Equal(3, resultPublication.LegacyReleases.Count);
+
+                Assert.Equal(publication.LegacyReleases[0].Id, resultPublication.LegacyReleases[0].Id);
```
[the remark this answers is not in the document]
markhiveit: After switching I got this failure:
```
Begin Differences (1 differences):
Types [Guid,Guid], Item Expected[0].Publication.Methodologies[0].Methodology.Versions[0].MethodologyContent.MethodologyVersionId != Actual[0].Publication.Methodologies[0].Methodology.Versions[0].MethodologyContent.MethodologyVersionId, Values (339b8c89-9b59-45b1-9a3a-6dc8e0a613a7,00000000-0000-0000-0000-000000000000)
End Differences (Maximum of 1 differences shown).
Expected: True
Actual:   False
```

I think this is due to the test data not having an Id set, rather than an actual issue. So I'll not switch to using `AssertDeppEqualTo`.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
markhiveit: Changed - and made other similar changes across `PublicationServiceTests` where appropriate.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
markhiveit: Ok - I will leave it as it is for now.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
markhiveit: Renamed - Did consider moving the ordering out of the method for the reasons you mentioned.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ReleaseViewModels.cs
markhiveit: I've changed all `PermissionsSet` classes across the codebase to `record`s.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/IPublicationService.cs
markhiveit: So my thought behind keeping them separate is to try to stop pagination concerns infecting lots of the codebase if it becomes more widespread. Right now, since `ListActiveReleases` returns all results, it has been simple to separate out pagination into its own thing. I was thinking that if in the future we wanted to reduce the load on the DB and only return results from the DB for the requested page, then maybe `ListActiveReleases` could return a `Queryable` which could then be used by a generic pagination method that gets the correct page for any given queryable.

This has two advantages in my eyes - pagination and the stuff being paginated can be unit tested independently (we would test pagination once rather than repeatedly for each individual endpoint that has pagination, and you don't need to think about pagination when testing methods like `ListActiveReleases`) and someone writing a new feature gets the pagination at minimal cost (they just need to ensure their new method provides a `Queryable` that a generic pagination method could consume).

What got me thinking along these lines is that there are several things in the codebase that probably have infected the codebase and are gotchas if you forget to account for them when adding a feature - and I suspect it didn't have to be this way. 

All that said, I doubt pagination will hold such sway so maybe it's not such a big concern in this case, but on the other hand if you entangle stuff I think you often get unforeseen consequences.

So my preference would be to keep the two different methods, but happy to do whatever you think is best. Happy to add paginationParams - although not sure I'm going to get the time today!

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ReleaseViewModels.cs
markhiveit: I was getting warnings for these after adding `#nullable enable`. Adding stuff like `= string.Empty;` seemed unnecessary, but I didn't want to leave the warning, so I went for this.

Not sure what the best approach is in this case?

src/explore-education-statistics-admin/src/services/publicationService.ts
markhiveit: Done - also renamed `GetReleaseParams` to `ListReleaseParams`.

src/explore-education-statistics-admin/src/pages/admin-dashboard/components/__tests__/NonScheduledReleaseSummary.test.tsx
markhiveit: Also switched all the other test files to not use a variable and to use this new `ReleaseWithPermissions` type

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ReleaseViewModels.cs
markhiveit: Talked to Nick on Slack - I've switched to initialising them either as `string.Empty` or `null!` as appropriate.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ReleaseViewModels.cs
markhiveit: See previous comment.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ReleaseViewModels.cs
markhiveit: See previous comment.

markhiveit:
Talked to Nick on Slack. He doesn't want this PR merged until ReleaseListItem has everything the frontend needs, since that was what EES-3554 was originally created to address.

Unfortunately, I'm going on holiday for a week after today, but I'll do as much as I can today, and then either I can pick it up in a week's time or someone else will have to pick it up.

markhiveit:
I thought we needed release errors and warnings on this endpoint, but that turned out not to be the case. So I've added various things to `ReleaseListItem`, but I'm guessing at what is required, so probably will be changed again. Also added a unit test to check all `ReleaseListItem` elements are hydrated.

ntsim (approved):

PR #3453, opened 2022-08-17
EES-3655 - check 'WAIT_MEMORY_CACHE_EXPIRY' exists in UI tests

luke-h1:
This PR: 
- adds an assertion to `run_tests.py` to check that the required environment variable `WAIT_MEMORY_CACHE_EXPIRY` exists

Other changes:
* changes assertion error messages to be a bit more helpful if environment variables are missing
* Updates `.env.example`

### DIFF
--- a/tests/robot-tests/run_tests.py
+++ b/tests/robot-tests/run_tests.py
@@ -164,6 +164,8 @@ assert os.getenv('WAIT_LONG') is not None
 assert os.getenv('WAIT_SMALL') is not None
 assert os.getenv('FAIL_TEST_SUITES_FAST') is not None
 assert os.getenv('IDENTITY_PROVIDER') is not None
+assert os.getenv('WAIT_MEMORY_CACHE_EXPIRY') is not None
+
 
 if args.slack_webhook_url:
     os.environ['SLACK_WEBHOOK_URL'] = args.slack_webhook_url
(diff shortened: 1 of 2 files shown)


tests/robot-tests/run_tests.py
```diff
@@ -152,18 +152,32 @@
...
-assert os.getenv('PUBLIC_AUTH_USER') is not None
-assert os.getenv('PUBLIC_AUTH_PASSWORD') is not None
-assert os.getenv('RELEASE_COMPLETE_WAIT') is not None
-assert os.getenv('WAIT_MEDIUM') is not None
-assert os.getenv('WAIT_LONG') is not None
-assert os.getenv('WAIT_SMALL') is not None
-assert os.getenv('FAIL_TEST_SUITES_FAST') is not None
-assert os.getenv('IDENTITY_PROVIDER') is not None
+
+assert os.getenv('TIMEOUT') is not None, 'Missing environment variable: TIMEOUT. refer to `.env.example` for more information.'
```
ntsim, line 156: All the extra messages seem a bit overkill. If you feel like it's actually helpful, then would suggest that we refactor this into a function so that we're not copy and pasting the same message over and over again.

luke-h1: Yeah on second thoughts I think the out of the box AssertionError message is good enough i.e.:

```
  File "/Users/lukehive/srv/dev/explore-education-statistics/tests/robot-tests/run_tests.py", line 186, in <module>
    assert os.getenv('WAIT_MEMORY_CACHE_EXPIRY')
AssertionError
```

ntsim (approved):

PR #3456, opened 2022-08-17
EES-3050, EES-3051 - add more context to 'More details' + 'help' buttons for accessibility

luke-h1:
This PR:
- addresses an accessibility issue whereby the 'More details; and 'help' buttons don't give enough context about the 'More details' and 'help' buttons

### DIFF
--- a/src/explore-education-statistics-common/src/components/Details.tsx
+++ b/src/explore-education-statistics-common/src/components/Details.tsx
@@ -11,6 +11,7 @@ import React, {
   useState,
 } from 'react';
 import styles from './Details.module.scss';
+import VisuallyHidden from './VisuallyHidden';
 
 let hasNativeDetails: boolean;
 let idCounter = 0;
@@ -41,6 +42,7 @@ export interface DetailsProps {
   open?: boolean;
   summary: string;
   summaryAfter?: ReactNode;
+  visuallyHiddenText?: string;
 }
 
 const Details = ({
@@ -53,6 +55,7 @@ const Details = ({
   onToggle,
   summary,
   summaryAfter,
+  visuallyHiddenText,
 }: DetailsProps) => {
   const [id] = useState(propId);
   const ref = useRef<HTMLElement>(null);
@@ -138,8 +141,10 @@ const Details = ({
           data-testid={formatTestId(`Expand Details Section ${summary}`)}
         >
           {summary}
+          {visuallyHiddenText && (
+            <VisuallyHidden> {visuallyHiddenText}</VisuallyHidden>
+          )}
         </span>
-
         {summaryAfter}
       </summary>
       <div
--- a/src/explore-education-statistics-common/src/modules/find-statistics/components/KeyStat.tsx
+++ b/src/explore-education-statistics-common/src/modules/find-statistics/components/KeyStat.tsx
@@ -73,6 +73,11 @@ const KeyStat = ({
               <Details
                 summary={summary?.dataDefinitionTitle[0] || 'Help'}
                 className={styles.definition}
+                visuallyHiddenText={
+                  !summary?.dataDefinitionTitle[0]
+                    ? `for ${keyStat.title}`
+                    : undefined
+                }
               >
                 <div data-testid={`${testId}-definition`}>
                   {summary.dataDefinition.map(data => (
--- a/src/explore-education-statistics-common/src/modules/release/components/ReleaseDataAccordion.tsx
+++ b/src/explore-education-statistics-common/src/modules/release/components/ReleaseDataAccordion.tsx
@@ -173,6 +173,7 @@ const ReleaseDataAccordion = ({
                           <Details
                             summary="More details"
                             className="govuk-!-margin-top-2"
+                            visuallyHiddenText={` for file ${file.name}`}
                           >
                             <div className="dfe-white-space--pre-wrap">
                               {file.summary}
--- a/src/explore-education-statistics-common/src/components/__tests__/Details.test.tsx
+++ b/src/explore-education-statistics-common/src/components/__tests__/Details.test.tsx
@@ -200,4 +200,20 @@ describe('Details', () => {
       expect.objectContaining({ target: summary }),
     );
   });
+
+  test('passing a visuallyHiddenText prop displays visually hidden text', () => {
+    const { container } = render(
+      <Details
+        summary="Test summary"
+        id="test-details"
+        visuallyHiddenText="publication 1"
+      >
+        Key stats
+      </Details>,
+    );
+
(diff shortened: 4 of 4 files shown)


amyb-hiveit (changes requested):
I don't think aria-labelledby is the right solution here, it should refer to another element that labels the element it's applied to (MDN is a good source for the intricacies of aria https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) .

We'd be better off passing in some screenreader only text that is added after `summary` using the `VisuallyHidden` component.

src/explore-education-statistics-common/src/modules/release/components/ReleaseDataAccordion.tsx
```diff
@@ -173,6 +173,7 @@ const ReleaseDataAccordion = ({
                           <Details
                             summary="More details"
                             className="govuk-!-margin-top-2"
+                            visuallyHiddenText={`More details for file ${file.name}`}
```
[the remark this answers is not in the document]
luke-h1: Cool, didn't realize this. Will have to spend some time & see if there are any good accessibility courses to do.

src/explore-education-statistics-common/src/modules/find-statistics/components/KeyStat.tsx
```diff
@@ -73,6 +73,7 @@ const KeyStat = ({
               <Details
                 summary={summary?.dataDefinitionTitle[0] || 'Help'}
                 className={styles.definition}
+                visuallyHiddenText={`Help for definition: ${keyStat.title}`}
```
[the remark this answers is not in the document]
luke-h1: `dataDefinitionTitle` is an optional text Analysts can set on keystat blocks. Here are a couple of examples from prod. I've pushed a solution but I'm not too sure about the text content.


Screenshot 2022-08-21 at 18 55 03

Screenshot 2022-08-21 at 18 56 01

src/explore-education-statistics-common/src/modules/find-statistics/components/KeyStat.tsx
```diff
@@ -73,6 +73,11 @@ const KeyStat = ({
               <Details
                 summary={summary?.dataDefinitionTitle[0] || 'Help'}
                 className={styles.definition}
+                visuallyHiddenText={
+                  summary?.dataDefinitionTitle[0]
+                    ? ` help for ${keyStat.title}`
+                    : ''
+                }
```
amyb-hiveit, line 80: I think this is the wrong way round, this will only show the help text when the dataDefinitionTitle is set and just 'Help' otherwise.
Also it should just be `for ${keyStat.title}` as 'Help' is already in the summary and this will be read out with it ('Help help for whatevs').

You could also put the space in the Details component as it's likely to get missed if it has to be passed in with the text.

src/explore-education-statistics-common/src/components/Details.tsx
```diff
@@ -138,8 +141,10 @@ const Details = ({
           data-testid={formatTestId(`Expand Details Section ${summary}`)}
         >
           {summary}
+          {visuallyHiddenText && (
+            <VisuallyHidden>{visuallyHiddenText}</VisuallyHidden>
```
amyb-hiveit, line 145: Could add the space here to make sure it's always there.

src/explore-education-statistics-common/src/modules/find-statistics/components/__tests__/KeyStat.test.tsx
```diff
@@ -107,9 +107,9 @@ describe('KeyStat', () => {
       );
 
       expect(
-        screen.getByRole('button', {
-          name: 'What is the number of applications received?',
-        }),
+        screen.getByTestId(
+          'Expand Details Section What is the number of applications received?',
```
amyb-hiveit, line 111: I don't think this test should need to change as it's for when there is a dataDefinitionTitle.

amyb-hiveit (changes requested):

src/explore-education-statistics-common/src/modules/find-statistics/components/__tests__/KeyStat.test.tsx
luke-h1: Looks like this test fails on the following without this change: 

```typescript
  <summary
              aria-controls="details-content-1"
              aria-expanded="false"
              class="govuk-details__summary"
              role="button"
              tabindex="0"
            >
              <span
                class="summaryText govuk-details__summary-text"
                data-testid="Expand Details Section What is the number of applications received?"
                id="details-heading-details-content-1"
              >
                What is the number of applications received?
                <span
                  class="govuk-visually-hidden"
                >

                  for Number of applications received
                </span>
              </span>
            </summary>
```

amyb-hiveit: Sorry I wasn't very clear, I meant that it shouldn't need to change if we have it so the extra visually hidden text is only added when there's not a `dataDefinitionTitle`.  
As `dataDefinitionTitle` is controlled by the analysts, and looking at the example you screenshotted above generally makes enough sense by itself, we shouldn't append the extra text to it as it won't necessarily make sense, as in the test here which would read " What is the number of applications received? for Number of applications received".

src/explore-education-statistics-common/src/modules/find-statistics/components/KeyStat.tsx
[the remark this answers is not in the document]
luke-h1: I've added a ternary so that the visually hidden text: `Help for keystat` only shows up when `dataDefinition` is equal to 'Help'.

src/explore-education-statistics-common/src/modules/find-statistics/components/KeyStat.tsx
```diff
@@ -73,6 +73,11 @@ const KeyStat = ({
               <Details
                 summary={summary?.dataDefinitionTitle[0] || 'Help'}
                 className={styles.definition}
+                visuallyHiddenText={
```
amyb-hiveit, line 76: Don't think this is quite right as it will only render the hidden text if an analyst has set the `dataDefinitionTitle` to 'Help', which is unlikely.  
I think it should render when the `dataDefinitionTitle` isn't set as that's when the visible text defaults to 'Help':
```
visuallyHiddenText={
 !summary?.dataDefinitionTitle[0]'
 ? `for ${keyStat.title}`
 : undefined
  }
```

luke-h1: Ah yeah, you're right. Think I got this one mixed up

amyb-hiveit (approved):

PR #3459, opened 2022-08-18
EES-3604 new find stats page initial build

amyb-hiveit:
Initial build for the new Find Statistics page.

The existing page should not be affected, to see the new one use the `newDesign=true` query parameter, e.g. http://localhost:3000/find-statistics?newDesign=true

Includes: 
- publications list, using a single page of static publications data (I've made some assumptions about the shape of the data so this may need updating)
- pagination, uses the new [GDS pattern](https://design-system.service.gov.uk/components/pagination/). We're a few versions behind so I've copied over the required styles.

Does not include:
- fetching real publications
- search
- filters
- sorting

This is a bit messy as we have to maintain the current version and the new API isn't ready yet, so I've put in notes about where things would be done and what can be tidied up later. 

findstats


src/explore-education-statistics-common/src/components/Pagination.tsx
```diff
@@ -0,0 +1,122 @@
...
+    children,
+    className,
+    pageNumber,
+    rel,
+  }: {
+    children: ReactNode;
+    className: string;
+    pageNumber: number;
+    rel: 'next' | 'prev';
+  }) => ReactNode;
```
ntsim, line 19: Do we need to separate a separate prop for this? Would suggest that we consolidate both of these into a single prop:

```ts
interface LinkRenderProps {
  'aria-current'?: 'page' | undefined;
  'aria-label'?: string;
  children: ReactNode;
  className: string;
  pageNumber: number;
  rel?: 'next' | 'prev';
}

interface PaginationProps {
  renderLink: (props: LinkRenderProps) => ReactNode;
} 
```

Note that aria props are using snake-case. The intention of this would be so that we can just spread these props straight into whatever link component we're using:

```tsx
<Pagination
  renderLink={{ pageNumber, ...props }) => 
    <Link {...props} to={`?page=${pageNumber}`} />
  }
/>
```

Think we could also go a little further and return the link's `href` (or `to` to better suit our Link components) as part of the render props too. We'd want to set the base URL with something like:
```ts
interface PaginationProps {
  baseUrl?: string;
}
```

The `href` / `to` render prop would then return a URL that concatenates the `baseUrl` if it has been set, or returns the query string like `?page=${pageNumber}` if it hasn't. Might need the `qs` package to help with parsing query parameters, but probably possible to do in a simple manner without.

Note that the `baseUrl` could include its own query string, so we should bear that in mind when concatenating the query string to it.

We should then be able to simplify our render prop to something like:
```tsx
<Pagination
  renderLink={props => <Link {...props} />}
/>
```

src/explore-education-statistics-common/src/components/Pagination.tsx
```diff
@@ -0,0 +1,122 @@
...
+                <span className="govuk-pagination__link-title">Previous</span>
+              </>
+            ),
+          })}
+        </div>
+      )}
+
+      <ul className="govuk-pagination__list">
+        {pageNumbers.map((pageNumber, index) => {
+          const key = `paginationKey-${index}`;
```
ntsim, line 75: Would just use the `index` directly as the result is the same. If we're just trying to get over lint warnings, we can just use `index.toString()`

src/explore-education-statistics-common/src/components/Pagination.tsx
```diff
@@ -0,0 +1,122 @@
...
+  return (
+    <nav
+      className="govuk-pagination"
+      role="navigation"
+      aria-label="Pagination navigation"
+    >
+      {currentPage !== 1 && (
+        <div className="govuk-pagination__prev">
+          {nextPrevLinkRenderer({
+            className: 'govuk-link govuk-pagination__link',
```
ntsim, line 60: Could refactor all instances of this `className` prop to use a single module scoped variable so we don't have to change it in multiple places in the future.

src/explore-education-statistics-common/src/components/__tests__/Pagination.test.tsx
```diff
@@ -0,0 +1,270 @@
...
+    const links = screen.getAllByRole('link');
+    expect(links).toHaveLength(7);
+
+    const listItems = screen.getAllByRole('listitem');
+    expect(listItems).toHaveLength(5);
+
+    // Previous link
+    expect(links[0]).toHaveAttribute('rel', 'prev');
+    expect(links[0]).toHaveAttribute('href', `/test-url?page=${2}`);
+    expect(within(links[0]).getByText('Previous')).toBeInTheDocument();
```
ntsim, line 22: Don't have to, but rather than using nested selectors like this, could just do:

```ts
expect(links[0]).toHaveTextContent('Previous');
```

Similar thing would need to be done for all other assertions throughout this test suite.

src/explore-education-statistics-common/src/components/__tests__/Pagination.test.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+    const links = screen.getAllByRole('link');
+    expect(links).toHaveLength(7);
+
+    const listItems = screen.getAllByRole('listitem');
+    expect(listItems).toHaveLength(5);
+
+    // Previous link
+    expect(links[0]).toHaveAttribute('rel', 'prev');
+    expect(links[0]).toHaveAttribute('href', `/test-url?page=${2}`);
```
ntsim, line 21: Why interpolate actual numbers into this assertion? Should be able to just do:
```ts
expect(links[0]).toHaveAttribute('href', `/test-url?page=2`);
```

Same applies for all other assertions like this throughout the test suite.

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPageNew.tsx
```diff
@@ -0,0 +1,223 @@
...
+    PublicationSummaryWithRelease[]
+  >(publications);
+
+  useEffect(() => {
+    // TODO EES-3517 - fetch selected page of publications here,
+    // something like this:
+    // async function fetchPublications(params) {
+    //   const nextPublications = await publicationService.getPublications(params);
+    //   setCurrentPublications(nextPublications);
+    // }
```
ntsim, line 52: Would suggest that we do all data loading via `getServerSideProps` to keep it all in one place. This should make things less complicated as we only need to work with the props that are passed through from there. We shouldn't need to maintain current state for page/publications with this approach.

Importantly, we would still need to handle loading states when clicking through pages, so we'll need something like this:

```ts
const router = useRouter();
const [isLoading, toggleLoading] = useToggle(false);

useEffect(() => {
  router.events.on('routeChangeStart', toggleLoading.on);
  router.events.on('routeChangeComplete', toggleLoading.off);

  return () => {
    router.events.off('routeChangeStart', toggleLoading.on);
    router.events.off('routeChangeComplete', toggleLoading.off);
  };
}, [toggleLoading, router]);
```

Would also be tempted to wrap this up into a re-usable frontend hook called `useRouterLoading`!

src/explore-education-statistics-frontend/src/modules/find-statistics/__tests__/FindStatisticsPageNew.test.tsx
```diff
@@ -0,0 +1,99 @@
+import FindStatisticsPageNew from '@frontend/modules/find-statistics/FindStatisticsPageNew';
+import { testPublications } from '@frontend/modules/find-statistics/__tests__/__data__/testPublications';
+import { render, screen, within } from '@testing-library/react';
+import React from 'react';
+
+describe('FindStatisticsPageNew', () => {
+  const testPaging = {
```
ntsim, line 7: This should have an explicit type

src/explore-education-statistics-common/src/services/publicationService.ts
```diff
@@ -10,6 +10,23 @@ import { contentApi } from './api';
 
 export type ReleaseApprovalStatus = 'Draft' | 'HigherLevelReview' | 'Approved';
 
+// TODO EES-3517 a guess at the new type for the find stats page.
+export interface PublicationSummaryWithRelease {
+  id: string;
+  legacyPublicationUrl?: string;
+  release?: {
```
ntsim, line 17: We might want to call this `latestRelease` to make it less ambiguous

src/explore-education-statistics-common/src/services/publicationService.ts
```diff
@@ -10,6 +10,23 @@ import { contentApi } from './api';
 
 export type ReleaseApprovalStatus = 'Draft' | 'HigherLevelReview' | 'Approved';
 
+// TODO EES-3517 a guess at the new type for the find stats page.
+export interface PublicationSummaryWithRelease {
```
ntsim, line 14: Would move this next to `PublicationSummary` as they're similar in name/context.

src/explore-education-statistics-common/src/services/publicationService.ts
```diff
@@ -10,6 +10,23 @@ import { contentApi } from './api';
...
+  id: string;
+  legacyPublicationUrl?: string;
+  release?: {
+    id: string;
+    published: string;
+    theme: {
+      title: string;
+    };
```
ntsim, line 22: Themes aren't associated to releases. It goes Theme > Topic > Publication > Release.

Would put this on the publication itself:

```ts
export interface PublicationSummaryWithRelease {
  theme: {
    title: string;
  }
}
```

src/explore-education-statistics-frontend/src/modules/find-statistics/components/PublicationSummary.tsx
```diff
@@ -0,0 +1,55 @@
...
+          <li>
+            <strong>Published:</strong>{' '}
+            <FormattedDate format="d MMM yyyy">
+              {release.published}
+            </FormattedDate>
+          </li>
+          <li>
+            <strong>Theme:</strong> {release.theme.title}
+          </li>
+        </ul>
```
ntsim, line 50: A definition list (`dl`, `dt`, `dd`) seems more semantically correct for what we're doing here.

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPageNew.tsx
```diff
@@ -0,0 +1,223 @@
...
+              </div>
+            ) : (
+              <>
+                {currentPublications.map(publication => (
+                  <PublicationSummary
+                    key={publication.id}
+                    publication={publication}
+                  />
+                ))}
+              </>
```
ntsim, line 180: Given this is a list of results, should we be using a list and list item for each summary?

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPageNew.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+            <div
+              className="govuk-visually-hidden"
+              id="searchResults"
+              ref={resultsRef}
+              tabIndex={-1}
+            >
+              Search results
+            </div>
+
```
ntsim, line 166: We'll want to show some loading spinner whenever the page (and sort, filters, etc) are changed

ntsim (changes requested):

src/explore-education-statistics-common/src/components/util/generatePageNumbers.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export default function generatePageNumbers({
+  currentPage,
+  maxPageNumbers = 8,
+  offset = 2,
+  totalPages,
+}: Options): (number | null)[] {
+  // return all if less than the max
+  if (totalPages <= maxPageNumbers) {
+    return new Array(totalPages).fill(null).map((_, i) => i + 1);
```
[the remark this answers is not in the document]
amyb-hiveit: I've gone for Lodash's range instead here.

src/explore-education-statistics-common/src/components/util/generatePageNumbers.ts
```diff
@@ -0,0 +1,51 @@
+interface Options {
+  currentPage: number;
+  maxPageNumbers?: number;
+  offset?: number;
```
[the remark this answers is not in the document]
amyb-hiveit: I've changed a lot of this now and added a docblock and comments to try to explain it better. 

It no longer cares about window size, just lets you defined a maximum number of items to return so for example if you had the pagination on a full width page you could increase it. The responsiveness is handled by the GDS styles instead off trying to change what's returned based on screensize.

src/explore-education-statistics-common/src/components/Pagination.tsx
amyb-hiveit: yep, totally just trying to shut the linter up :) Have changed it now.

src/explore-education-statistics-common/src/components/__tests__/Pagination.test.tsx
amyb-hiveit: No idea why I did that!

src/explore-education-statistics-common/src/components/Pagination.tsx
```diff
@@ -0,0 +1,116 @@
...
+  'aria-current'?: 'page' | undefined;
+  'aria-label'?: string;
+  children: ReactNode;
+  className: string;
+  rel?: 'next' | 'prev';
+  to: string;
+}
+
+export interface PaginationProps {
+  baseUrl: string;
```
ntsim, line 18: Would suggest making this optional and default it to an empty string. In these cases, the `to` render props would just be set to `?page=1` etc.

src/explore-education-statistics-common/src/components/Pagination.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  if (!pageNumbers.length) {
+    return null;
+  }
+
+  const queryString = queryParams
+    ? `&${Object.keys(queryParams)
+        .map(key => `${key}=${queryParams[key]}`)
+        .join('&')}`
+    : '';
```
ntsim, line 47: This is too simplistic for creating a query string as it doesn't handle URL encoding. We should use the `qs` to handle this instead. In fact, there is already an `appendQuery` function that uses this in the admin project (it also handles URLs that may already contain a query string).

Would suggest moving this into common and using it here:
```ts
to: appendQuery(baseUrl, {
  ...queryParams,
  page: currentPage - 1
})
```

src/explore-education-statistics-frontend/src/components/Pagination.tsx
```diff
@@ -0,0 +1,11 @@
+import BasePagination, { PaginationProps } from '@common/components/Pagination';
+import Link from '@frontend/components/Link';
+import React from 'react';
+
+const Pagination = ({ ...props }: Omit<PaginationProps, 'renderLink'>) => (
+  <BasePagination
+    {...props}
+    renderLink={({ ...linkProps }) => <Link {...linkProps} />}
+  />
```
ntsim, line 9: Would suggest integrating Next's router into the `baseUrl` and `queryParams` props as defaults:

```tsx
const router = useRouter();

<BasePagination
  baseUrl={baseUrl ?? router.pathname}
  queryParams={queryParams ?? router.query}
  renderLink={({ ...linkProps }) => <Link {...linkProps} />}
/>

File: src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPageNew.tsx [hunk 19]
```diff
@@ -0,0 +1,155 @@
...
+                      key={publication.id}
+                      publication={publication}
+                    />
+                  ))}
+                </ul>
+              )}
+
+              <Pagination
+                baseUrl="/find-statistics"
+                queryParams={{ newDesign: true }} // TO DO EES-3517 make sure params for filters / search are included here
```
ntsim, line 143: We should be able to remove these props if we implement Next router-based defaults as suggested in an earlier comment.

src/explore-education-statistics-frontend/src/modules/find-statistics/components/PublicationSummary.tsx
```diff
@@ -0,0 +1,70 @@
...
+            </a>
+          </p>
+        </>
+      ) : (
+        <>
+          <h3 className="govuk-!-margin-bottom-2">
+            <Link to={`/find-statistics/${slug}`}>{title}</Link>
+          </h3>
+          <p>{summary}</p>
+        </>
```
ntsim, line 40: In the case of legacy publications, wouldn't it be better to keep the link within the heading and strip out the link from 'Statistics at DfE'?

It currently seems quite inconsistent: image

Think it'd be a little confusing to users as to why the link has moved out of the heading. Users also don't really care if a link remains in a site or they get bounced somewhere else.

Moving it back into the title would have the side benefit of not needing extra visually hidden text.

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPageNew.tsx
```diff
@@ -0,0 +1,155 @@
...
+  // const [totalResults, setTotalResults] = useState<number>(paging.totalResults);
+  const [currentPage, setCurrentPage] = useState<number>(paging.page);
+  const [currentPublications, setCurrentPublications] = useState<
+    PublicationSummaryWithRelease[]
+  >(publications);
+
+  useEffect(() => {
+    setCurrentPage(paging.page);
+    setCurrentPublications(publications);
+  }, [paging.page, publications]);
```
ntsim, line 29: We shouldn't need any of this if all paging is being handled by `getServerSideProps`. Prop changes will re-render this component so we're currently just copying props into state (which is redundant).

ntsim (changes requested):

src/explore-education-statistics-common/src/components/util/__tests__/generatePageNumbers.test.ts
```diff
@@ -0,0 +1,453 @@
...
+  });
+
+  test('uses the minimum return size if maxReturnSize below it', () => {
+    expect(
+      generatePageNumbers({
+        currentPage: 1,
+        maxReturnSize: 3,
+        totalPages: 6,
+      }),
+    ).toEqual([1, 2, 3, null, 6]);
```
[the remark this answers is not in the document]
amyb-hiveit: There's a minimum size of 5 (to allow for to allow [first, spacer, current, spacer, last]), these tests where maxReturnSize is 3 are checking that the min size gets used if a smaller maxReturnSize is passed in.

src/explore-education-statistics-common/src/components/util/__tests__/generatePageNumbers.test.ts
```diff
@@ -0,0 +1,453 @@
...
+
+    expect(
+      generatePageNumbers({
+        currentPage: 3,
+        maxReturnSize: 3,
+        totalPages: 6,
+      }),
+    ).toEqual([1, 2, 3, null, 6]);
```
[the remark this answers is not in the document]
amyb-hiveit: That would take the size above the minimum max size (5), currently it will only return the max size or lower. If the numbers on either side of the current one don't fit into that then one is left out.
Not sure if that's necessarily right, but I haven't found much of a consensus when looking at pagination patterns.

src/explore-education-statistics-common/src/components/util/__tests__/generatePageNumbers.test.ts
```diff
@@ -0,0 +1,453 @@
...
+
+    expect(
+      generatePageNumbers({
+        currentPage: 2,
+        maxReturnSize: 3,
+        totalPages: 6,
+      }),
+    ).toEqual([1, 2, 3, null, 6]);
```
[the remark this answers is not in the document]
amyb-hiveit: It does return 1, 2, 3, null, 6

src/explore-education-statistics-common/src/components/util/generatePageNumbers.ts
```diff
@@ -0,0 +1,107 @@
+import last from 'lodash/last';
+import range from 'lodash/range';
+
+interface Options {
+  currentPage: number;
+  maxReturnSize?: number;
```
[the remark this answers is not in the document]
amyb-hiveit: I went for maxItems (which must be about the 5th thing I've called it as nothing feels quite right!)

src/explore-education-statistics-common/src/components/util/__tests__/generatePageNumbers.test.ts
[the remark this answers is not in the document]
ntsim: Ah, I made a little mistake in my comment - 1, 2, 3, 4, null, 6 would be what I would expect.

I think my main issue here is that we wouldn't be respecting the GDS pattern's expectations.

> For smaller screens, show page numbers for:
> 
> - the current page
> - previous and next pages
> - first and last pages
>
> For larger screens, show page numbers for:
> 
> - the current page
> - at least one page immediately before and after the current page
> - first and last pages
> 
> Use ellipses (…) to replace any skipped pages.

The main point is that prev/next numbers have to be covered so it's it's possible to have up to 7 items as the minimum (first, spacer, prev, current, next, spacer, last).

I'm kinda thinking that specifying a max number of items is causing confusion here and perhaps specifying a target window size might work better? A window size of 1 would be the bare minimum as this accounts for prev/next. Window sizes of 2+ would result in stuff like 1, null, 3, 4, 5, 6, 7, null, 10, etc.

ntsim (changes requested):

amyb-hiveit:
@ntsim this is updated now.

I think the GDS guidance for small screens is a bit confusing, it says to show page numbers for the previous and next pages, but the CSS actually hides all page numbers apart from the current, first and last, e.g. [1, null, 3, null, 5] or [1, null, 6] if the first or last page is selected.

Based on this and your suggestions I've changed:
- got rid of maxItems and dropped the idea of being able to configure it to show more than 7 items as that's probably unnecessary.
- added windowSize instead which can be either small or medium and is set based on screen size in the Pagination component
- on small screens it renders the items to match what the styles will show, on desktop it's the same as the default before (at least one page immediately before and after the current page)

ntsim (approved):

PR #3460, opened 2022-08-18
EES-3603 Add publication summary

markhiveit:
This ticket adds publications summaries, which are stored in the ContentDB's `Publications` table under `Summary`. This ticket only adds the ability to give a publication a summary when creating and updating a publication. This summary isn't currently displayed on the public site. 

The intention is for all existing publications on production to give their publication a summary before we display these summaries on the public site. In the migration to add the Summary column, the summary defaults to the empty string. For this reason, if an admin user updates a publication's details, they are forced to give the publication a summary at that point.

Originally, it was intended to add a full-text search index on the new Summary DB column. This hasn't been done because the SQL docker container we use for testing on our own computers doesn't have full-text search enabled. As I'm on holiday next week and have other PR comments to address before then, a new ticket will be created to do this work.

On the frontend, we're currently switching the way publication details are updated - the old being `PublicationForm.tsx` and the new being `PublicationDetailsForm`. This PR adds the ability to change the summary on both. We now also displays the publication summary on Admin Release summary pages and the new, currently hidden, `PublicationDetailsPage`. The publication summary was intentionally not added to the Admin dashboard.


src/GovUk.Education.ExploreEducationStatistics.Content.Model/Publication.cs
```diff
@@ -14,6 +14,9 @@ public class Publication
 
         [Required(AllowEmptyStrings = false)] public string Title { get; set; } = string.Empty;
 
+        [MaxLength(160)]
```
benoutram, line 17: I think this is ok but just pointing out all the other places where we've configured a max length on a table field (not many!) have been configured using the fluent API in `ContentDbContext` or `StatisticsDbContext` rather than with data annotations.

src/explore-education-statistics-admin/src/pages/publication/components/__tests__/PublicationForm.test.tsx
```diff
@@ -194,6 +194,11 @@ describe('PublicationForm', () => {
       'Test title',
     );
 
+    await userEvent.type(
+      screen.getByLabelText('Publication summary'),
+      'Test summary',
+    );
+
```
benoutram, line 201: I think you could also add a new test `shows validation error when there is no summary` given there are variations like `shows validation error when there is no title` etc.

src/explore-education-statistics-admin/src/pages/publication/components/PublicationDetailsForm.tsx
```diff
@@ -80,11 +81,15 @@ const PublicationDetailsForm = ({
...
             title: '',
+            summary: '',
             topicId: '',
           }),
         }}
         validationSchema={Yup.object<PublicationDetailsFormValues>({
           title: Yup.string().required('Enter a title'),
+          summary: Yup.string()
+            .required('Enter a summary')
+            .max(160, 'Summary must be 160 characters or less'),
```
benoutram, line 92: Not sure if PublicationDetailsForm is missing tests or if it's tested through some other component?

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/PublicationViewModels.cs
```diff
@@ -13,6 +13,9 @@ public class PublicationViewModel
 
         public string Title { get; set; }
 
+        [MaxLength(160)]
```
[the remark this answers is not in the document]
markhiveit: Talked to Ben on Slack - I'll leave it on `Publication` and the publication save view model - the rest I'll remove.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/MyPublicationViewModel.cs
```diff
@@ -10,6 +11,8 @@ public class MyPublicationViewModel
 
         public string Title { get; set; }
 
+        [MaxLength(160)] public string Summary { get; set; }
```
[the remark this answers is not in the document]
markhiveit: See previous comment.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ReleaseViewModels.cs
```diff
@@ -26,6 +26,9 @@ public class ReleaseViewModel
 
         public string PublicationTitle { get; set; }
 
+        [MaxLength(160)]
```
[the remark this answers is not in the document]
markhiveit: See previous comment.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationRepositoryTests.cs
```diff
@@ -1046,6 +1048,8 @@ public async Task GetPublicationForUser_NoReleaseMethodologyLegacyRelease()
 
             var publication = new Publication
             {
+                Title = "Test title",
+                Summary = "Test summary",
```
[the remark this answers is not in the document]
markhiveit: I've renamed the test.

I've not thought through what will happen after the further refactoring. My guess is the test will stick around in some shape or form, but perhaps `GetPublicationForUser` will return a `Publication` rather than a `PublicationViewModel` like has happened with the release repository methods in this PR.

src/explore-education-statistics-admin/src/pages/publication/components/PublicationDetailsForm.tsx
markhiveit: I think the tests are in `PublicationDetailsPage.test.tsx`. I've added a test to check for a publication summary validation error, the same as in `PublicationForm`.

src/explore-education-statistics-admin/src/pages/publication/__tests__/PublicationDetailsPage.test.tsx
```diff
@@ -303,6 +303,33 @@ describe('PublicationDetailsPage', () => {
...
+      await waitFor(() => {
+        expect(screen.getByText('Publication details')).toBeInTheDocument();
+      });
+
+      userEvent.click(
+        screen.getByRole('button', { name: 'Edit publication details' }),
+      );
+
+      await waitFor(() => {
+        expect(screen.getByLabelText('Publication title')).toBeInTheDocument();
```
benoutram, line 318: Not sure if this is a mistake or intentional, but should it be 'Publication summary'?

markhiveit: I've changed it in on PublicationDetailsPage.test and PublicationForm.test.

PR #3463, opened 2022-08-24
Ees 3643 update taxonomy trees rather than delete

duncan-at-hiveit:
This PR:
- swaps out all instances of deleting the cached Methodology Tree and Publication Tree for immediate cache updates instead
- adds an `UpdateOnly` flag to CacheAttribute, to allow us to use methods to perform purely cache updates as opposed to cache retrievals / storing
- moves the caching methods that were touched as a part of this work into a centralised `ContentCacheService` which currently resides in `Content.Services` but in the future will hopefully move to `Content.Api` and be invoked remotely by a request from Publisher (as per the high-level plan in https://dfedigital.atlassian.net/browse/EES-3528).  It's hoped that this will provide a more centralised, easy to maintain and understand place where the effects of various caching activities can be easily understood.  

Note that we could've gone down the line of having a new caching attribute like `[BlobCacheInvalidate(typeof(AllMethodologyCacheKey))]` which would firstly remove a cache entry before then calling an existing method annotated with `[BlobCache(typeof(AllMethodologyCacheKey))]`, but this would leave a window of time where there was no value int he cache, which is what we want to avoid if possible!

All tests passing, other than already-failing Manage Users:

image


src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/BauCacheController.cs
```diff
@@ -61,9 +64,24 @@ public async Task<ActionResult> ClearPublicCacheTrees(ClearPublicCacheTreePathsV
                 await request.Paths
                     .ToAsyncEnumerable()
                     .ForEachAwaitAsync(
-                        path =>
-                            _publicBlobStorageService.DeleteBlob(BlobContainers.PublicContent, path)
-                    );
+                        
+                        async path =>
+                        {
+                            var allowedPath =
```
duncan-at-hiveit, line 70: I'll need to update some BAU doco somewhere to reference these new allowed values.  This was done as a way to break the dependency between AllMethodologiesCacheKey and PublicationTreeCacheKey with the Admin project, as a precursor to moving all public-facing caching code into the Content.Api and Data.Api projects themselves.

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
```diff
@@ -395,6 +399,9 @@ public void ConfigureServices(IServiceCollection services)
 
             services.AddAutoMapper(AppDomain.CurrentDomain.GetAssemblies());
             
+            services.AddTransient<IContentThemeService, ContentThemeService>();
```
duncan-at-hiveit, line 402: These `Content.Services` services need to be included here in Admin because they are in use by the ContentCacheService which is currently in direct use by Admin.

When we move towards the plan in https://dfedigital.atlassian.net/browse/EES-3528, we would instead be sending a request to the Content API to update its cached items instead of it being done from Admin directly, and so these DI dependencies can then be removed along with ContentCacheService.

I can always add a TODO to this affect, referencing an existing or a new caching ticket.

src/GovUk.Education.ExploreEducationStatistics.Common/Cache/CacheAttribute.cs
```diff
@@ -45,11 +45,19 @@ public abstract class CacheAttribute : Attribute
...
         /// </summary>
         public Type Key { get; }
+        
+        /// <summary>
+        /// Flag signifying that this caching attribute performs caching updates only.
+        /// When an attribute with this flag set is invoked, it will retrieve a fresh
+        /// value rather than a cached item and then set or update it in the cache,
+        /// therefore always caching and then supplying a freshly retrieved value.
+        /// </summary>
+        public bool UpdateOnly { get; }
```
benoutram, line 55: Could also consider calling this `forceUpdate` or `forceRefresh` which might require less explanation?

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ContentCacheService.cs
```diff
@@ -0,0 +1,142 @@
...
+    {
+        return _themeService.GenerateFullPublicationTree();
+    }
+    
+    [BlobCache(typeof(PublicationTreeCacheKey), updateOnly: true)]
+    public Task<IList<ThemeTree<PublicationTreeNode>>> UpdatePublicationTree()
+    {
+        _logger.LogInformation("Updating cached Publication Tree");
+        return _themeService.GenerateFullPublicationTree();
+    }
```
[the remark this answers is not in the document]
duncan-at-hiveit: Yeah good call - will go with `Get`.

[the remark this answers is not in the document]
duncan-at-hiveit: As discussed, split into 2 Services - MethodologyCacheService and ThemeCacheService, and in addition moved them within a "Cache" namespace too, so they're all together.

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
benoutram: I think adding a TODO to the code would be a good idea to retain this reason for anyone looking through these services in the `Startup.cs`.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ReleaseService.cs
```diff
@@ -24,44 +25,50 @@ public class ReleaseService : IReleaseService
...
             _fileStorageService = fileStorageService;
-            _methodologyService = methodologyService;
+            _methodologyCacheService = methodologyCacheService;
             _userService = userService;
             _publicationService = publicationService;
             _mapper = mapper;
         }
 
+        // TODO EES-3643 - move into a ReleaseCacheService?
         public async Task<Either<ActionResult, ReleaseViewModel>> GetCachedViewModel(string publicationSlug, string? releaseSlug = null)
```
benoutram, line 51: `MethodologyService.GetCachedSummariesByPublication` was moved (renamed as MethodologyCacheService.GetSummariesByPublication`), so on that basis this one should probably be moved too. However there would be a dependency in `ReleaseCacheService` on `MethodologyCacheService` then which is not so nice.

If this isn't moved, then should we revert the other one back? 😅

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/MethodologyServiceTests.cs
```diff
@@ -515,6 +516,253 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
...
+                Topics = ListOf(topic)
+            };
+
+            var latestVersions = ListOf(
+                new MethodologyVersion
+                {
+                    Id = Guid.NewGuid(),
+                    MethodologyContent = new MethodologyVersionContent {
+                        Annexes = new List<ContentSection>(),
+                        Content = new List<ContentSection>(),
```
benoutram, line 545: I think `new MethodologyVersionContent()` would be sufficient here, since the two properties will be initialised by default. Same comment applies to the other version.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/MethodologyServiceTests.cs
```diff
@@ -515,6 +516,253 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
...
+
+            await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId))
+            {
+                var service = SetupMethodologyService(contentDbContext,
+                    methodologyVersionRepository: methodologyVersionRepository.Object);
+
+                var result = await service.GetSummariesTree();
+                VerifyAllMocks(methodologyVersionRepository);
+
+                // Assert that the cached result is what was returned by the method.
```
benoutram, line 596: I think this comment has been copied from `MethodologyServiceCachingTests` but isn't applicable anymore.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/BauCacheController.cs
duncan-at-hiveit: Updated on the testing ticket

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ThemeServiceTests.cs
```diff
@@ -34,20 +32,18 @@ public ThemeServiceTests()
         }
 
         [Fact]
-        public async Task GetPublicationTree_Empty()
+        public async Task GetPublicationTree_NoThemes()
```
benoutram, line 35: I think the `CacheServiceTestFixture` and `BlobCacheService` setup at the top of this class can all be removed now

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/Cache/MethodologyCacheServiceTests.cs
```diff
@@ -0,0 +1,207 @@
...
+using static GovUk.Education.ExploreEducationStatistics.Common.Services.CollectionUtils;
+using static GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils.MockUtils;
+using static Moq.MockBehavior;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Services.Tests.Cache;
+ 
+[Collection(CacheServiceTests)]
+public class MethodologyCacheServiceTests : CacheServiceTestFixture
+{
+    private List<AllMethodologiesThemeViewModel> MethodologyTree = ListOf(
```
benoutram, line 25: Should be `_methodologyTree` and could also be `readonly`.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
```diff
@@ -1731,7 +1740,7 @@ await using (var context = InMemoryApplicationDbContext(contextId))
                     }
                 );
 
-                VerifyAllMocks(methodologyVersionRepository, publicBlobCacheService);
+                VerifyAllMocks(methodologyVersionRepository, publicBlobCacheService, themeCacheService);
```
benoutram, line 1743: Missing `methodologyCacheService`

src/GovUk.Education.ExploreEducationStatistics.Common/Cache/CacheAttribute.cs
benoutram: Have chatted with @ntsim about this and decided to go with `forceUpdate`.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Cache/MethodologyCacheService.cs
```diff
@@ -0,0 +1,56 @@
...
+
+    public MethodologyCacheService(
+        IMethodologyService methodologyService, 
+        IThemeService themeService, 
+        ILogger<MethodologyCacheService> logger)
+    {
+        _methodologyService = methodologyService;
+        _themeService = themeService;
+        _logger = logger;
+    }
```
benoutram, line 31: `ThemeService` shouldn't be needed here

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Cache/ThemeCacheService.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        return _themeService.GetPublicationTree();
+    }
+    
+    [BlobCache(typeof(PublicationTreeCacheKey), forceUpdate: true)]
+    public Task<IList<ThemeTree<PublicationTreeNode>>> UpdatePublicationTree()
+    {
+        _logger.LogInformation("Updating cached Publication Tree");
+        return _themeService.GetPublicationTree();
+    }
```
benoutram, line 41: I think these `BlobCache` attributes are missing a `ServiceName = "target"` attribute to signal that during the get or get and update that it needs to find the appropriate `BlobCacheService` by target service name that's wired up with the a `BlobStorageService` with the **public** storage account.

I think this is only working locally during testing because the storage emulator has a single account that we use for both the private and public account connection strings.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Cache/MethodologyCacheService.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+        return _methodologyService.GetSummariesTree();
+    }
+
+    [BlobCache(typeof(AllMethodologiesCacheKey), forceUpdate: true)]
+    public Task<Either<ActionResult, List<AllMethodologiesThemeViewModel>>> UpdateSummariesTree()
+    {
+        _logger.LogInformation("Updating cached Methodology Tree");
+        return _methodologyService.GetSummariesTree();
+    }
```
benoutram, line 40: Same comment about the `BlobCache` attributes missing a `ServiceName` that will target a specifc `BlobCacheService` that's using the **public** storage account connection.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Publisher/Startup.cs
```diff
@@ -52,7 +56,17 @@ public override void Configure(IFunctionsHostBuilder builder)
...
+                {
+                    var publicBlobCacheService = GetBlobCacheService(provider, "PublicStorage");
+                    CacheAspect.Enabled = true;
+                    BlobCacheAttribute.AddService("default", publicBlobCacheService);
+                    return publicBlobCacheService;
+                })
+                .AddScoped<IContentMethodologyService, ContentMethodologyService>()
+                .AddScoped<IMethodologyCacheService, MethodologyCacheService>()
+                .AddScoped<IThemeCacheService, ThemeCacheService>()
+                .AddScoped<IThemeService, ThemeService>()
```
benoutram, line 69: `ThemeService` hasn't been aliased `ContentThemeService` like it is for `ContentMethodologyService`, and the same for the same services in the Admin's `Startup.cs`. I see there's no `ThemeService` in Publisher for it to clash with, but I think it would be good name these consistently.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Startup.cs
```diff
@@ -52,7 +56,22 @@ public override void Configure(IFunctionsHostBuilder builder)
...
                 .AddSingleton<IFileStorageService, FileStorageService>(provider =>
                     new FileStorageService(GetConfigurationValue(provider, "PublisherStorage")))
-                .AddScoped(provider => GetBlobCacheService(provider, "PublicStorage"))
+                .AddScoped(provider =>
+                {
+                    var publicBlobCacheService = GetBlobCacheService(provider, "PublicStorage");
+                    CacheAspect.Enabled = true;
+                    BlobCacheAttribute.AddService("default", publicBlobCacheService);
+                    return publicBlobCacheService;
+                })
```
benoutram, line 65: This is registered but only run when the `PublishReleaseContentFunction` is called for the first time because it has an `IBlobCacheService` dependency.

It's not very clear that if no other service depended on `IBlobCacheService` (excluding services where it's configured directly inline such as the way we set up `PublishingService` and `ContentService`) that caching would not be enabled and no caching services would be registered.

This is the case in EES-3648 where I discovered it broken after removing the DI of `IBlobCacheService` from `PublishReleaseContentFunction`.

benoutram: I left some TODO's related to this on my EES-3648 PR in https://github.com/dfe-analytical-services/explore-education-statistics/pull/3479/commits/070456b7f9a3784f9c18568c745dbd98f79f2da3 but I didn't find a good way of configuring this outside of the setup of the services.

It seems like the setup of logging happens after our custom Startup configuration, so if we try and do something like this...

```
   var serviceProvider = builder.Services
  .AddScoped...
  ...
  .BuildServiceProvider();

  ConfigureCachingAttribute(serviceProvider);

  private static void ConfigureCachingAttribute(IServiceProvider provider)
  {
      CacheAspect.Enabled = true;

      var publicBlobCacheService = GetBlobCacheService(provider, "PublicStorage");
      BlobCacheAttribute.AddService("default", publicBlobCacheService);
  }
```

then if  will fail on `provider.GetRequiredService<ILogger<BlobCacheService>>())` within `GetBlobStorageService` and `GetBlobCacheService`

benoutram (approved):

PR #3466, opened 2022-08-25
EES-3669 - change publication summary to be textarea

luke-h1:
This PR: 
- changes publication summary field from an input to text area
- fixes `PublicationDetailsPage.tsx` table formatting if no publication summary is set

image

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/publication/PublicationDetailsPage.tsx
+++ b/src/explore-education-statistics-admin/src/pages/publication/PublicationDetailsPage.tsx
@@ -60,7 +60,7 @@ const PublicationDetailsPage = () => {
           <SummaryList>
             <SummaryListItem term="Publication title">{title}</SummaryListItem>
             <SummaryListItem term="Publication summary">
-              {summary}
+              {summary ?? 'Not set'}
             </SummaryListItem>
             <SummaryListItem term="Theme">{theme?.title}</SummaryListItem>
             <SummaryListItem term="Topic">{topic?.title}</SummaryListItem>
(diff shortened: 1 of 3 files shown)


src/explore-education-statistics-admin/src/pages/publication/components/PublicationForm.tsx
```diff
@@ -139,7 +143,7 @@ const PublicationForm = ({
               />
             )}
 
-            <FormFieldTextInput<FormValues>
+            <FormFieldTextArea<FormValues>
```
benoutram, line 146: I think you could use `maxLength={160}` here

src/explore-education-statistics-admin/src/pages/publication/PublicationDetailsPage.tsx
```diff
@@ -60,7 +60,7 @@ const PublicationDetailsPage = () => {
           <SummaryList>
             <SummaryListItem term="Publication title">{title}</SummaryListItem>
             <SummaryListItem term="Publication summary">
-              {summary}
+              {summary ?? 'No summary set'}
```
benoutram, line 63: Could use `Not set` as the default value here, given that it's already labelled as the summary field.

src/explore-education-statistics-admin/src/pages/publication/components/PublicationForm.tsx
luke-h1: Yep, good call

benoutram (approved):

PR #3467, opened 2022-08-25
EES-3574: Separate fetching publications and methodologies for new Admin dashboard

benoutram:
This PR adds a new endpoint `GET /api/publication/{publicationId}/methodologies` to get publication methodologies.

It introduces a new back-end view model for listing methodology versions for a publication - `MethodologyVersionViewModel`, collapsing the old `MyPublicationMethodologyVersionViewModel` and `MyMethodologyVersionViewModel` into one and also collapsing the permissions models of both into one.

The old `MyPublication` (soon to be removed by EES-3577) is updated to use the new `MethodologyVersionViewModel` and the  old view models `MyPublicationMethodologyVersionViewModel` and `MyMethodologyVersionViewModel` are removed.

The new methodologies API endpoint is integrated with the Publication methodologies tab in the new Admin dashboard, removing the need to fetch the publication again to get the methodologies.

For backwards compatibility, front-end changes are also made to the current Admin dashboard after the methodologies view model type in `MyPublication` was changed to remove the old 'My' view models.

The new endpoint is also integrated into the Prerelease methodologies page which was also depending on fetching the Publication and using methodologies from the `MyPublication`.

### Future work

* `publicationService.ts` has a `TODO` for EES-3666 to add an additional endpoint to allow fetching an external methodology without the publication.

* `InternalReleaseNote` is added to `MethodologyVersionViewModel` for backwards compatability with the old Admin dashboard and can be removed by EES-3576.

### Other changes

* Don't use Automapper to build the view model and permissions model in order to make (i) how the latest methodology version is retrieved, (ii) how the view models are ordered, and (iii) how the permissions are derived more transparent.
* Add missing mock verifications to `PublicationServiceTests`.
* Constrain id parameters to Guid type in `MethodologyController`.


src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/MethodologyVersionViewModel.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.ViewModels;
+
+public record MethodologyVersionViewModel
+{
+    public Guid Id { get; set; }
+
+    public bool Amendment { get; set; }
+
+    // TODO EES-3596 Remove InternalReleaseNote added to support the old Admin dashboard
```
ntsim, line 15: Would probably drop this comment as we should probably keep this in. As we don't display it in the new dashboard, we should at least show it somewhere. This would probably be best shown in a methodology's 'Sign off' section's summary list.

markhiveit (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Either.cs
```diff
@@ -96,6 +96,11 @@ public static class EitherExtensions
 
     public static class EitherTaskExtensions
     {
+        public static async Task<bool> IsRight<TFailure, TSuccess>(this Task<Either<TFailure, TSuccess>> task)
```
ntsim, line 99: Should we add the `IsLeft` variant of this as well?

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -409,19 +415,63 @@ private MyPublicationViewModel HydrateMyPublicationViewModel(Publication publica
                 releaseViewModel.Permissions = PermissionsUtils.GetReleasePermissions(_userService, release);
             });
 
+            publicationViewModel.Methodologies = await HydrateMethodologyVersionViewModels(publication);
```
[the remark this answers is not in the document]
benoutram: The `.Ignore` is actually due to not wanting to use AutoMapper to build the `MethodologyVersionViewModels` themselves, not specifically the permissions, although determining the permissions are part of the reason for not using it.

As well as the complexity of needing to use a permissions resolver class with AutoMapper to determine the permissions, using Automapper to map the `MethodologyVersion` itself needs to work out the latest version of the `PublicationMethodology` and it's clearer for this to happen in the service rather than be obscured in AutoMapper config.

I'm not sure it's any less efficient and when it comes to the cleanup we'll be removing the entire mapping from `Publication` to `MyPublicationViewModel` imminently anyway like you say.

I'm not sure how the `.Ignore` could be misunderstood but I've added a comment to explain why it's there now.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -68,12 +71,7 @@ public class PublicationService : IPublicationService
                     .OnSuccess(() => _publicationRepository.GetAllPublicationsForTopic(topicId))
                     .OrElse(() => _publicationRepository.GetPublicationsForTopicRelatedToUser(topicId, userId))
                 )
-                .OnSuccess(publicationViewModels =>
-                {
-                    return HydrateMyPublicationsViewModels(publicationViewModels)
-                        .OrderBy(publicationViewModel => publicationViewModel.Title)
-                        .ToList();
-                });
+                .OnSuccess(async publicationViewModels => await HydrateMyPublicationsViewModels(publicationViewModels));
```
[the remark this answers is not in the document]
benoutram: The reason for moving it inside was it's that outside it needs braces around the `await` and an extra call `ToList`, but I've moved it back outside now.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -129,14 +130,63 @@ public class MethodologyService : IMethodologyService
...
+                                    await _userService.CheckCanUpdateMethodologyVersion(latestVersion).IsRight(),
+                                CanApproveMethodologyVersion =
+                                    await _userService.CheckCanApproveMethodologyVersion(latestVersion).IsRight(),
+                                CanMarkMethodologyVersionAsDraft =
+                                    await _userService.CheckCanMarkMethodologyVersionAsDraft(latestVersion).IsRight(),
+                                CanMakeAmendmentOfMethodology =
+                                    await _userService.CheckCanMakeAmendmentOfMethodology(latestVersion).IsRight(),
+                                CanRemoveMethodologyLink =
+                                    await _userService.CheckCanDropMethodologyLink(publicationMethodology).IsRight()
+                            };
```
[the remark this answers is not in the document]
benoutram: There's a few reasons why I originally decided not to do this:

* The existing method within `PermissionUtils` which is `GetReleasePermissions` isn't working asynchronously whereas I wanted this to be. So I'd have either had to change that and in turn change things calling that to all work asynchronously, or leave the two methods inconsistent.
* `GetReleasePermissions` only needs the `Release` but a  `GetMethodologyVersionPermissions` will need more than just the `MethodologyVersion`, since it also needs the PublicationMethodology link.
This can be navigated through from the `MethodologyVersion` but we'd probably want to guarantee that it's set before relying on it, or have an extra parameter where the `PublicationMethodology` is passed explicitly. It just means the implementation is not quite as clean when compared to the example of `GetReleasePermissions`.
* Very soon the methodology permissions should only need to be worked out in one place rather than two, and having it inline when building the view model appeared to me to be clearer where they actually come from, rather than needing to click through.

Saying all this though, I've decided to move it here, but not updated `GetReleasePermissions` to work asynchronously just yet.

src/explore-education-statistics-admin/src/services/methodologyService.ts
```diff
@@ -33,14 +33,24 @@ export interface BasicMethodologyVersion {
   methodologyId: string;
 }
 
-export interface MyMethodologyVersion extends BasicMethodologyVersion {
+export interface MethodologyVersionListItem {
```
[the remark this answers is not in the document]
benoutram: Thanks, renamed `MethodologyVersion` now. Good point about a `MethodologyVersionSummary`

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/MethodologyVersionViewModel.cs
```diff
@@ -0,0 +1,42 @@
...
+    public Guid MethodologyId { get; set; }
+
+    public Guid? PreviousVersionId { get; set; }
+
+    public record MethodologyVersionPermissions
+    {
+        public bool CanDeleteMethodologyVersion { get; set; }
+        public bool CanUpdateMethodologyVersion { get; set; }
+        public bool CanApproveMethodologyVersion { get; set; }
+        public bool CanMarkMethodologyVersionAsDraft { get; set; }
```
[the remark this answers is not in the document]
benoutram: Thanks, removed 'version' from all the permissions now to make things less verbose.

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Either.cs
benoutram: Added `IsLeft` variant now, albeit not used currently.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/MethodologyVersionViewModel.cs
benoutram: The existing and new dashboard now both use the new `GET /api/publication/{publicationId}/methodologies` added by this PR which returns the new `MethodologyVersionViewModel` and maps to `MethodologyVersion` on the front end (renamed from `MethodologyVersionListItem` just now). This has this temporary `InternalReleaseNote` to keep the existing dashboard working, but not displayed on the new dashboard, hence the TODO to remove it in future.

The Methodology page / 'Sign off' tab are using `GET /api/methodology/{methodologyVersionId}/summary` which returns `MethodologyVersionSummaryViewModel` and maps to `BasicMethodologyVersion` . This hasn't changed but already has a `LatestInternalReleaseNote`. Since we weren't ready to display multiple approval histories yet for methodologies this should probably have been left as `InternalReleaseNote` for now rather than 'latest'`. 

I've just added 3ce3eb8e3da7c94290e05490b36e9e0f3ca320dd to display it on the 'Sign off' tab.

It does seem a bit strange that the variant returned for the content page is the 'Summary' view model.

benoutram: Added 850ffec809e9829f7ce36a0e0a6f17dfc85c708c to display it on the 'Sign off' tab.

The existing and new dashboard now both use the new `GET /api/publication/{publicationId}/methodologies` added by this PR which was returning a new `MethodologyVersionViewModel` which mapped to `MethodologyVersion` on the front end.

In 227af3491faa80974e172b20ff169334bd6e3bdc I've done a big switcharoo as discussed, renaming as follows:

`MethodologyVersionSummaryViewModel` -> `MethodologyVersionViewModel` (BE) with corresponding `BasicMethodologyVersion` -> `MethodologyVersion` (FE), extending from new `BaseMethodologyVersion`.

`MethodologyVersionViewModel` -> `MethodologyVersionSummaryViewModel` (BE) with corresponding `MethodologyVersion` -> `MethodologyVersionSummary` (FE), also extending from the new `BaseMethodologyVersion`.

`LatestInternalReleaseNote` was already on the view model used by the content page, which is now displayed on the 'Sign off' tab. Since we aren't ready to display multiple approval histories yet for methodologies I've renamed if from `LatestInternalReleaseNote` to `InternalReleaseNote`.

The TODO on `MethodologyVersionSummaryViewModel` has been left as we can remove it in future once the old dashboard is removed.




It does seem a bit strange that the variant returned for the content page is the 'Summary' view model.

markhiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -170,7 +209,7 @@ public class MethodologyService : IMethodologyService
                 .OnSuccess(_ => GetSummary(id));
         }
 
-        private async Task<MethodologyVersionSummaryViewModel> BuildMethodologySummaryViewModel(
+        private async Task<MethodologyVersionViewModel> BuildMethodologySummaryViewModel(
```
ntsim, line 212: We should rename this method to `BuildMethodologyVersionViewModel`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -121,22 +122,60 @@ public class MethodologyService : IMethodologyService
                 });
         }
 
-        public async Task<Either<ActionResult, MethodologyVersionSummaryViewModel>> GetSummary(Guid id)
+        public async Task<Either<ActionResult, MethodologyVersionViewModel>> GetSummary(Guid id)
```
ntsim, line 125: Would rename this method to `Get` or `GetMethodology`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Util/PermissionsUtils.cs
```diff
@@ -1,28 +1,52 @@
...
-       };
-   }
+    public static ReleasePermissions GetReleasePermissions(IUserService userService, Release release)
+    {
+        return new ReleasePermissions
+        {
+            CanAddPrereleaseUsers = CheckHasPermission(userService.CheckCanAssignPrereleaseContactsToRelease(release)),
+            CanUpdateRelease = CheckHasPermission(userService.CheckCanUpdateRelease(release)),
+            CanDeleteRelease = CheckHasPermission(userService.CheckCanDeleteRelease(release)),
+            CanMakeAmendmentOfRelease = CheckHasPermission(userService.CheckCanMakeAmendmentOfRelease(release))
```
ntsim, line 22: We could make all these async using the `IsRight` method? We could get rid of the `CheckHasPermission` method as well

src/explore-education-statistics-admin/src/pages/methodology/edit-methodology/status/components/MethodologyStatusForm.tsx
```diff
@@ -25,7 +25,7 @@ export interface FormValues {
 
 interface Props {
   isPublished?: string;
-  methodologySummary: BasicMethodologyVersion;
+  methodologySummary: MethodologyVersion;
```
ntsim, line 28: Would rename this prop to `methodology`

src/explore-education-statistics-admin/src/pages/methodology/edit-methodology/status/MethodologyStatusEditPage.tsx
```diff
@@ -16,7 +16,7 @@ interface FormValues {
 }
 
 interface Props {
-  methodologySummary: BasicMethodologyVersion;
+  methodologySummary: MethodologyVersion;
```
ntsim, line 19: This could be renamed to `methodology`

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Methodologies/MethodologyController.cs
```diff
@@ -28,47 +28,47 @@ public class MethodologyController : ControllerBase
...
         {
             return await _methodologyService
                 .GetAdoptableMethodologies(publicationId)
                 .HandleFailuresOrOk();
         }
 
-        [HttpGet("methodology/{methodologyVersionId}/summary")]
-        public async Task<ActionResult<MethodologyVersionSummaryViewModel>> GetMethodologySummary(Guid methodologyVersionId)
+        [HttpGet("methodology/{methodologyVersionId:guid}/summary")]
+        public async Task<ActionResult<MethodologyVersionViewModel>> GetMethodologySummary(Guid methodologyVersionId)
```
ntsim, line 64: We could:
- Remove the summary part of the URL i.e. `methodology/{methodologyVersionId:guid}` (requires associated frontend change to `methodologyService` too)
- Rename the method to `GetMethodology`

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Util/PermissionsUtils.cs
benoutram: The reason why I didn't do this before was because of down stream changes that it would require, which were nothing to do with my original work, but I've done this now in d03401b3b51428e652d67a14a9fb3c0c7560d2f7 🙂 .

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
benoutram: Thanks, done this renaming in 2027c7b9a00eb143fd485c50b0786cab239974a4 and the same applies to all the other comments left about naming.

ntsim (approved):

PR #3468, opened 2022-08-25
EES-3671 - fix user resets user roles keyword

luke-h1:
This PR fixes the 'user resets user roles' keyword

I tested this by: 
* changing prelease user2 to a user to pre-release viewer
* changing prerelease user1 to an analyst user 
* running `manage_users_page.robot` and checked these users were reset to 'no role' & the test passed

image

full run: 
image

### DIFF
--- a/tests/robot-tests/tests/libs/admin_api.py
+++ b/tests/robot-tests/tests/libs/admin_api.py
@@ -186,15 +186,15 @@ def user_resets_user_roles_via_api_if_required(user_emails: list) -> None:
             user_ids = [get_prerelease_user_details_via_api(user_email)['id']]
             _ = [user_removes_all_release_and_publication_roles_from_user(user_id) for user_id in user_ids]
             BuiltIn().log(f'All userReleaseRoles & userPublicationRoles reset for user: {user_email}')
-        except IndexError:
-            BuiltIn().log(f'User with email {user_email} does not exist in pre-release user list')
+        except TypeError or IndexError:
+            BuiltIn().log(f'User with email {user_email} does not exist in pre-release user list', 'WARN')
 
             try:
                 user_ids = [get_user_details_via_api(user_email)['id']]
                 _ = [user_removes_all_release_and_publication_roles_from_user(user_id) for user_id in user_ids]
                 BuiltIn().log(f'All userReleaseRoles & userPublicationRoles reset for user: {user_email}')
-            except IndexError:
-                BuiltIn().log(f'User with email {user_email} does not exist in user list')
+            except TypeError or IndexError:
+                BuiltIn().log(f'User with email {user_email} does not exist in user list', 'WARN')
 
 
 def user_create_test_release_via_api(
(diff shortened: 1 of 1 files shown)


tests/robot-tests/tests/libs/admin_api.py
```diff
@@ -183,17 +183,18 @@ def user_resets_user_roles_via_api_if_required(user_emails: list) -> None:
...
+            BuiltIn().log(f'User with email {user_email} does not exist in pre-release user list')
 
-        except AssertionError:
-            BuiltIn().log(f'Failed to remove roles from user {user_email}')
+            try:
+                user_ids = [get_user_details_via_api(user_email)['id']]
+                _ = [user_removes_all_release_and_publication_roles_from_user(user_id) for user_id in user_ids]
+                BuiltIn().log(f'All userReleaseRoles & userPublicationRoles reset for user: {user_email}')
+            except IndexError:
+                BuiltIn().log(f'User with email {user_email} does not exist in user list')
```
benoutram, line 197: Is there a concept of logging this at a warn or error level to make it more prominent in the console?

tests/robot-tests/tests/libs/admin_api.py
```diff
@@ -183,17 +183,18 @@ def user_resets_user_roles_via_api_if_required(user_emails: list) -> None:
...
-            BuiltIn().log(f'User with email {user_email} does not exist')
-
-        try:
-            result = [user_removes_all_release_and_publication_roles_from_user(user_id) for user_id in user_ids]
-            assert(all(result))
+            user_ids = [get_prerelease_user_details_via_api(user_email)['id']]
+            _ = [user_removes_all_release_and_publication_roles_from_user(user_id) for user_id in user_ids]
             BuiltIn().log(f'All userReleaseRoles & userPublicationRoles reset for user: {user_email}')
+        except IndexError:
+            BuiltIn().log(f'User with email {user_email} does not exist in pre-release user list')
```
benoutram, line 190: I suppose this log message could get annoying if the list was changed to allow a mix of pre-release and non pre-release users, and might be better if there was just one log message at the end "User with email blah does not exist in user list or pre-release user list", but probably not an issue right now.

benoutram (approved):

tests/robot-tests/tests/libs/admin_api.py
luke-h1: Yep looks like there is. I've changed this to a warning.

tests/robot-tests/tests/libs/admin_api.py
luke-h1: Cool, yep one to keep an eye out for if that list gets bigger or more users are going to be added to the `allowed_users` list

luke-h1:
In addition to the changes made here I've added an additonal assertion error to catch TypeErrors which could occur if pre-release users aren't found: 

`TypeError: string indices must be integers`

benoutram (approved):

PR #3470, opened 2022-08-26
EES-3062 - explore data button accesibility fixes

luke-h1:
This PR: 
- fixes an accessibility issue whereby the 'explore data' button is repeated without additional context. For now, I've added visually hidden text inside the `ExploreDataButton` component and used `block.name` to provide additional context. Not too sure about whether the block is the right thing to use here as it just contains: ` Generic data block - National`. This potentially needs revisiting to decide whether something such as the datablock name would be a more appropriate text

### DIFF
--- a/src/explore-education-statistics-frontend/src/modules/find-statistics/components/ExploreDataButton.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/find-statistics/components/ExploreDataButton.tsx
@@ -4,14 +4,15 @@ import ButtonLink from '@frontend/components/ButtonLink';
 import { DataBlock } from '@common/services/types/blocks';
 import { logEvent } from '@frontend/services/googleAnalyticsService';
 import React from 'react';
+import VisuallyHidden from '@common/components/VisuallyHidden';
 
 interface Props {
   block: DataBlock;
+  hiddenText?: string;
 }
 
-const ExploreDataButton = ({ block }: Props) => {
+const ExploreDataButton = ({ block, hiddenText }: Props) => {
   const [buttonClicked, toggleButtonClicked] = useToggle(false);
-
   return (
     <>
       <ButtonLink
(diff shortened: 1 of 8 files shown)


src/explore-education-statistics-frontend/src/modules/find-statistics/components/PublicationSectionBlocks.tsx
```diff
@@ -56,7 +56,10 @@ const PublicationSectionBlocks = ({
                     </h3>
 
                     <p>Use our table tool to explore this data.</p>
-                    <ExploreDataButton block={block} />
+                    <ExploreDataButton
+                      block={block}
+                      visuallyHiddenText={`for ${block.name}`}
```
amyb-hiveit, line 61: Instead of block.name could it use the title that's used for the table caption?

src/explore-education-statistics-frontend/src/modules/find-statistics/components/ExploreDataButton.tsx
```diff
@@ -4,14 +4,15 @@ import ButtonLink from '@frontend/components/ButtonLink';
 import { DataBlock } from '@common/services/types/blocks';
 import { logEvent } from '@frontend/services/googleAnalyticsService';
 import React from 'react';
+import VisuallyHidden from '@common/components/VisuallyHidden';
 
 interface Props {
   block: DataBlock;
+  visuallyHiddenText?: string;
```
ntsim, line 11: As mentioned in another PR - this could be shortened to `hiddenText` as the 'visually' part is a bit verbose

luke-h1: I've merged some PRs that include `visuallyHiddenText` as the prop name so I'll have a look around & change them to `hiddenText` as part of this PR

ntsim (approved):

PR #3471, opened 2022-08-26
EES-3064 - fix breadcrumb accessibility issues

luke-h1:
This PR: 
- fixes accessibility issues whereby breadcrumbs can be confused for normal links when using a screen reader


amyb-hiveit (approved):
Looks good 👍

PR #3472, opened 2022-08-26
EES-3648 Update cached publications rather than deleting them from the cache

benoutram:
This PR builds on https://github.com/dfe-analytical-services/explore-education-statistics/pull/3463, adding a new `Content.Services.PublicationCacheService` with methods to get and update a cached publication.

Prior to this change, when publishing a release immediately the Publisher function would delete the cached publication, including any superseded publications where the superseding is now potentially in effect.

Rather than deleting the cached publication, we now update it to ensure that there's never a window where a cached publication doesn't exist.

Additionally, all cases in the Admin where cached publications were deleted from blob storage (including superseded publications) are now swapped to update the cached publication instead.

### Other changes

* In order to avoid needing to configure the Admin app service and Publisher function projects with the AutoMapper configuration to map a publication to a public `Content.Services.ViewModels.PublicationViewModel`, and in turn map the entire object graph, AutoMapper is no longer used. View models constructors are used to make the code a little less verbose. 

* Registering a `BlobCacheService` with `BlobCacheAttribute` can't take place in the default `FunctionsStartup.Configure` method as part of its own factory initialisation because that won't happen until the service is required for dependency injection. A cached method could be invoked before it's been required by some other component in which case it won't be initialised and it won't be registered with `BlobCacheAttribute`. If no component requires an `IBlobCacheService` to be injected, it would never be initialised and registered.
It's also not possible in the `FunctionsStartup.Configure` method to create a `BlobCacheService` or eagerly retrieve one by getting it from the `IServiceProvider` because of its dependency on `ILogger`. You can't request an `ILogger` during service initialisation because it's not set up yet.

For these reasons, this PR creates and registers the public `BlobCacheService` using an extension. Its creation is now independent of whether its required by DI and done at a stage during startup where logging is already set up.

### Alternative approach without separate cache services

As part of this work I looked at how different this approach to caching could work if we don't create separate cache services, instead in-lining the cached methods alongside the `Get`  un-cached resource method in the service, adding in `GetCached{Resource}` and `UpdateCached{Resourced}`.

[V2 of this PR](https://github.com/dfe-analytical-services/explore-education-statistics/pull/3479) has no `PublicationCacheService` and so no additional test class for it either. However it adds an extra complication when it comes to setting up tests for `GetCachedPublication` and `UpdateCachedPublication` in `PublicationServiceTests`.

To avoid repeating all the set-up of data required to support their internal call to `Get` for the un-cached publication, it creates a stub of `Get` in a `PublicationServiceStub` and uses this as the class under test rather than `PublicationService`.

```
        private class PublicationServiceStub : PublicationService
        {
            public PublicationServiceStub(ContentDbContext contentDbContext,
                IPersistenceHelper<ContentDbContext> contentPersistenceHelper) : base(
                contentDbContext,
                contentPersistenceHelper)
            {
            }

            public override Task<Either<ActionResult, PublicationViewModel>> Get(string publicationSlug)
            {
                // Stub a response value to allow testing cached methods which use this
                return Task.FromResult(new Either<ActionResult, PublicationViewModel>(PublicationViewModel));
            }
        }
```

With separate cache services, the un-cached service call only requires standard setup to mock behaviour of that method. Also if there are any additional operations that happen post retrieval of the cached resource before returning the response, such is the case in `ThemeCacheService` where the cached result is filtered, the cache service provides a clearer separation in the tests between testing the cached and un-cached methods.

After discussing with the team, we felt the benefit of having less classes doesn't warrant going with the [V2 approach](https://github.com/dfe-analytical-services/explore-education-statistics/pull/3479).

### UI Test Report

image

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
@@ -8,9 +8,7 @@ using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.Security;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Util;
 using GovUk.Education.ExploreEducationStatistics.Admin.ViewModels;
-using GovUk.Education.ExploreEducationStatistics.Common.Cache;
 using GovUk.Education.ExploreEducationStatistics.Common.Model;
-using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces.Security;
 using GovUk.Education.ExploreEducationStatistics.Common.Utils;
 using GovUk.Education.ExploreEducationStatistics.Common.ViewModels;
@@ -38,7 +36,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
         private readonly IUserService _userService;
         private readonly IPublicationRepository _publicationRepository;
         private readonly IMethodologyVersionRepository _methodologyVersionRepository;
-        private readonly IBlobCacheService _publicBlobCacheService;
+        private readonly IPublicationCacheService _publicationCacheService;
         private readonly IMethodologyCacheService _methodologyCacheService;
         private readonly IThemeCacheService _themeCacheService;
 
@@ -49,7 +47,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
             IUserService userService,
             IPublicationRepository publicationRepository,
             IMethodologyVersionRepository methodologyVersionRepository,
-            IBlobCacheService publicBlobCacheService, 
+            IPublicationCacheService publicationCacheService, 
             IMethodologyCacheService methodologyCacheService, 
             IThemeCacheService themeCacheService)
         {
@@ -59,7 +57,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
             _userService = userService;
             _publicationRepository = publicationRepository;
             _methodologyVersionRepository = methodologyVersionRepository;
-            _publicBlobCacheService = publicBlobCacheService;
+            _publicationCacheService = publicationCacheService;
             _methodologyCacheService = methodologyCacheService;
             _themeCacheService = themeCacheService;
         }
@@ -244,34 +242,28 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
                         publication.Published = DateTime.UtcNow;
                         await _context.SaveChangesAsync();
 
-                        await UpdateCachedTaxonomyBlobs();
+                        await _methodologyCacheService.UpdateSummariesTree();
+                        await _themeCacheService.UpdatePublicationTree();
+                        await _publicationCacheService.UpdatePublication(publication.Slug);
 
-                        await _publicBlobCacheService.DeleteItem(new PublicationCacheKey(publication.Slug));
-
-                        await DeleteCachedSupersededPublicationBlobs(publication);
+                        await UpdateCachedSupersededPublications(publication);
                     }
 
                     return await GetPublication(publication.Id);
                 });
         }
 
-        private async Task UpdateCachedTaxonomyBlobs()
-        {
-            await _methodologyCacheService.UpdateSummariesTree();
-            await _themeCacheService.UpdatePublicationTree();
-        }
-
-        private async Task DeleteCachedSupersededPublicationBlobs(Publication publication)
+        private async Task UpdateCachedSupersededPublications(Publication publication)
         {
             // NOTE: When a publication is updated, any publication that is superseded by it can be affected, so
-            // invalidate the superseded publications' caches
+            // update any superseded publications that are cached
             var supersededPublications = await _context.Publications
                 .Where(p => p.SupersededById == publication.Id)
                 .ToListAsync();
-            foreach (var p in supersededPublications)
-            {
-                await _publicBlobCacheService.DeleteItem(new PublicationCacheKey(p.Slug));
-            }
+
+            await supersededPublications
+                .ToAsyncEnumerable()
+                .ForEachAwaitAsync(p => _publicationCacheService.UpdatePublication(p.Slug));
         }
 
         private async Task<Either<ActionResult, Unit>> ValidateSelectedTopic(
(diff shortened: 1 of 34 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -244,34 +241,25 @@ public class PublicationService : IPublicationService
...
 
-                        await UpdateCachedTaxonomyBlobs();
-
-                        await _publicBlobCacheService.DeleteItem(new PublicationCacheKey(publication.Slug));
+                        await _methodologyCacheService.UpdateSummariesTree();
+                        await _themeCacheService.UpdatePublicationTree();
+                        await _publicationCacheService.UpdatePublication(publication.Slug);
 
-                        await DeleteCachedSupersededPublicationBlobs(publication);
+                        await UpdateCachedSupersededPublications(publication);
```
duncan-at-hiveit, line 248: I was wondering if this code could eventually move into `_publicationCacheService.UpdatePublication()` itself.  I guess that kind of refactoring would happen at the point where we change the communication mechanism between Publisher and the API projects (if we do end up doing that).

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Mappings/MappingProfiles.cs
```diff
@@ -58,18 +53,6 @@ public MappingProfiles()
                 .ForMember(
                     dest => dest.Content,
                     m => m.MapFrom(r => r.GenericContent.OrderBy(s => s.Order)));
```
duncan-at-hiveit, line 55: Whoa, cleared the decks here!

src/GovUk.Education.ExploreEducationStatistics.Content.Services/PublicationService.cs
```diff
@@ -43,16 +37,11 @@ public class PublicationService : IPublicationService
...
-            .OnSuccess(tuple =>
+            .OnSuccess(async tuple =>
             {
                 var (publication, latestRelease) = tuple;
-                var publicationViewModel = _mapper.Map<PublicationViewModel>(publication);
-                publicationViewModel.LatestReleaseId = latestRelease.Id;
-                publicationViewModel.IsSuperseded = IsSuperseded(publication);
-                publicationViewModel.Releases = ListPublishedReleases(publication);
-                return publicationViewModel;
+                return await BuildPublicationViewModel(publication, latestRelease);
```
duncan-at-hiveit, line 44: Why did you decide to do away with AutoMapper here out of interest?  Not that I'm complaining :D

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ViewModels/ExternalMethodologyViewModel.cs
```diff
@@ -1,8 +1,17 @@
-﻿namespace GovUk.Education.ExploreEducationStatistics.Content.Services.ViewModels
+﻿#nullable enable
+
+using GovUk.Education.ExploreEducationStatistics.Content.Model;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Services.ViewModels;
+
+public record ExternalMethodologyViewModel(
```
duncan-at-hiveit, line 7: Nice to have more records coming into the codebase :+1:

src/GovUk.Education.ExploreEducationStatistics.Publisher/BlobCacheStartup.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace GovUk.Education.ExploreEducationStatistics.Publisher;
+
+/// <summary>
+/// <p>
+/// Registering a BlobCacheService with BlobCacheAttribute can't take place in the default FunctionsStartup
+/// Configure method as part of its own factory initialisation because that won't happen until the service is
+/// required for dependency injection. A cached method could be invoked before it's been required by some other
+/// component in which case it won't be initialised and it won't be registered with BlobCacheAttribute. If no
+/// component requires a IBlobCacheService to be injected, it would never be initialised and registered.
```
duncan-at-hiveit, line 17: Wow, good find!

src/GovUk.Education.ExploreEducationStatistics.Publisher/Config/BlobCacheConfigProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+using Microsoft.Azure.WebJobs.Host.Config;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
+namespace GovUk.Education.ExploreEducationStatistics.Publisher.Config;
+
+[Extension("BlobCache")]
+// ReSharper disable once ClassNeverInstantiated.Global
+public class BlobCacheConfigProvider : IExtensionConfigProvider
```
duncan-at-hiveit, line 18: Nicely done - how did you find out about this approach?  I wanted to do something like this when enabling AOP caching in the Publisher initially but couldn't find out how.

duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Services/PublicationService.cs
benoutram: I thought I was going to need to copy the AutoMapper configuration from `Content.Services.Mappings.MappingProfiles` for `PublicationViewModel` to Admin and Publisher, and thought that would get extremely confusing, especially in the case of Admin where it's got it's own Admin `PublicationViewModel` (in fact it's got an additional one already for the manage content page!) and mappings for those already.

I can't remember now if I saw an error about the missing mappings and then decided on this approach, or if I just imagined this based on a change to Publisher's MappingProfiles in https://github.com/dfe-analytical-services/explore-education-statistics/pull/3463.

It's possible that I didn't need to make this change if the `Content.Services.Mappings.MappingProfiles` are loaded by Admin and Publisher but thought moving away from AutoMapper also makes it clearer how the view modesl are built.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
benoutram: Ah good point, thanks. I could look to see if the work I'm doing for EES-3645 would be a good place to do this because I can see similar code in multiple places across Admin and Publisher.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Config/BlobCacheConfigProvider.cs
benoutram: Mainly stack overflow! 😆

I read a few Github issues related to people running into the problem trying to acquire the logging instance before setup completes, but no ideal solution or workaround has been proposed. There's just repeated guidance out there about not trying to acquire the logger during Startup currently. That doesn't really apply to us since we don't want to do any actually logging in this phase, but just use it for construction of the BlobCacheService and construction of the BlobStorageService which both depend on an ILogger.

PR #3474, opened 2022-08-30
EES-3554 Part 2 - Integrate publication releases endpoint with Manage publication > Releases page

ntsim:
This PR integrates the recently updated publication releases endpoint (`/api/{publicationId}/releases`) with the 'Manage publication > Releases' page.

We have had to reorganise some of the existing component structure to correctly integrate this endpoint i.e.
- `PublicationUnpublishedReleases` - handles draft and scheduled release tables
- `PublicationPublishedReleases` handles the live/published release table

Each of the above components makes its own request to fetch the relevant releases from the publication releases endpoint.

## Introducing React Query

As part of this, we have introduced [React Query](https://tanstack.com/query) as a dependency to manage async requests. This is essentially a more feature complete version of the `useAsync*` collection of hooks that we use throughout the service. It provides useful features such as client-side caching and automatic re-fetching for stale data.

The main driver for introducing React Query was to handle the new paginated publication releases, for which we are using `useInfiniteQuery`. The alternative would have been to hand-roll our own version of this and it did not seem worthwhile to re-engineer something that is already handled by a well maintained package.

## Implications for `useAsync*` hooks

A side-effect of introducing React Query is that we also will want to consider swapping `useQuery` (from React Query) in place of `useAsyncRetry` and `useAsyncHandledRetry`. We have already introduced `useQuery` for the `ReleaseUnpublishedReleases` component, where there is no pagination.

My suggestion would be that we start to adopt React Query and aim to slowly migrate existing usages of the `useAsync*` hooks to the React Query APIs. React Query is better maintained, more feature rich, has extensive documentation and means we don't need to reinvent the wheel.

## Testing with React Query

One notable caveat to using React Query is that we need to wrap our test rendering code with a `QueryClientProvider`. This provider needs to be configured in a fairly consistent and specific way, meaning that it is useful for us to use a new custom `render` function that pre-configures this in a wrapper.

This means, that in test code which involves rendering a React Query consuming component, we should do:
```tsx
import render from '@admin-test/render';

test('my test...', () => {
  render(<MyReactQueryComponent />);
});
```

Note that currently, React Query is only in admin, hence why the alias is `@admin-test`. In the future, we will add React Query to the frontend project as well, upon which it will be aliased like `@frontend-test` instead.

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/publication/PublicationReleasesPage.tsx
+++ b/src/explore-education-statistics-admin/src/pages/publication/PublicationReleasesPage.tsx
@@ -1,53 +1,17 @@
 import ButtonLink from '@admin/components/ButtonLink';
-import PublicationDraftReleases from '@admin/pages/publication/components/PublicationDraftReleases';
 import PublicationPublishedReleases from '@admin/pages/publication/components/PublicationPublishedReleases';
-import PublicationScheduledReleases from '@admin/pages/publication/components/PublicationScheduledReleases';
+import PublicationUnpublishedReleases from '@admin/pages/publication/components/PublicationUnpublishedReleases';
 import usePublicationContext from '@admin/pages/publication/contexts/PublicationContext';
 import { PublicationRouteParams } from '@admin/routes/publicationRoutes';
 import { releaseCreateRoute } from '@admin/routes/routes';
-import publicationService from '@admin/services/publicationService';
-import LoadingSpinner from '@common/components/LoadingSpinner';
-import WarningMessage from '@common/components/WarningMessage';
-import useAsyncHandledRetry from '@common/hooks/useAsyncHandledRetry';
-import React, { useMemo } from 'react';
+import noop from 'lodash/noop';
+import React, { useRef } from 'react';
 import { generatePath } from 'react-router';
 
 const PublicationReleasesPage = () => {
-  const { publicationId } = usePublicationContext();
+  const { publicationId, publication } = usePublicationContext();
 
-  // To ensure the releases are up to date when you switch to this tab
-  // we're re-fetching the publication here instead of using it from the context.
-  // This is not ideal and will be replaced by a call to just get the publication
-  // releases when an endpoint for this is ready - EES-3554 .
-  const {
-    value: publication,
-    isLoading: loadingPublication,
-    retry: reloadPublication,
-  } = useAsyncHandledRetry(() =>
-    publicationService.getMyPublication(publicationId),
-  );
-
-  const draftReleases = useMemo(() => {
-    return (
-      publication?.releases.filter(
-        release =>
-          release.approvalStatus === 'Draft' ||
-          release.approvalStatus === 'HigherLevelReview',
-      ) ?? []
-    );
-  }, [publication?.releases]);
-
-  const publishedReleases = useMemo(() => {
-    return publication?.releases.filter(release => release.live) ?? [];
-  }, [publication?.releases]);
-
-  const scheduledReleases = useMemo(() => {
-    return (
-      publication?.releases.filter(
-        release => !release.live && release.approvalStatus === 'Approved',
-      ) ?? []
-    );
-  }, [publication?.releases]);
+  const publishedReleasesRefetchRef = useRef<() => void>(noop);
 
   return (
     <>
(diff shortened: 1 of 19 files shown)


src/explore-education-statistics-admin/src/pages/publication/components/PublicationPublishedReleases.tsx
```diff
@@ -1,139 +1,169 @@
...
 
   return (
     <>
-      <table
-        className="dfe-hide-empty-cells"
-        data-testid="publication-published-releases"
+      <h3 aria-atomic aria-live="polite">
+        {`Published releases${
+          lastPage ? ` (${lastPage.paging.totalResults})` : ''
+        }`}
```
amyb-hiveit, line 114: This is meant to show the number of publications currently show and the total e.g. "Published releases (5 of 7)".

src/explore-education-statistics-admin/src/pages/publication/components/PublicationUnpublishedReleases.tsx
```diff
@@ -0,0 +1,98 @@
...
+  onAmendmentDelete,
+}: Props) {
+  const { data: releases, isLoading, isSuccess, refetch } = useQuery(
+    ['publicationUnpublishedReleases', publicationId],
+    () =>
+      publicationService.listReleases<ReleaseSummaryWithPermissions>(
+        publicationId,
+        {
+          live: false,
+          pageSize: 20,
```
amyb-hiveit, line 26: What happens if there are more than 20 unpublished releases? Seems like they wouldn't all be shown as there's no pagination for these releases.

src/explore-education-statistics-admin/src/pages/publication/PublicationReleasesPage.tsx
```diff
@@ -1,53 +1,17 @@
...
-  }, [publication?.releases]);
-
-  const scheduledReleases = useMemo(() => {
-    return (
-      publication?.releases.filter(
-        release => !release.live && release.approvalStatus === 'Approved',
-      ) ?? []
-    );
-  }, [publication?.releases]);
+  const publishedReleasesRefetchRef = useRef<() => void>(noop);
```
amyb-hiveit, line 14: I doesn't look like the releases aren't being refetched when you switch tabs any more.

src/explore-education-statistics-admin/src/pages/publication/components/PublicationDraftReleases.tsx
```diff
@@ -28,13 +34,16 @@ const PublicationDraftReleases = ({ releases, onChange }: Props) => {
   const [showDraftStatusGuidance, toggleDraftStatusGuidance] = useToggle(false);
   const [showIssuesGuidance, toggleIssuesGuidance] = useToggle(false);
 
+  if (releases.length === 0) {
+    return <InsetText>There are no draft releases.</InsetText>;
+  }
+
```
amyb-hiveit, line 40: The spacing is a bit wrong when there are no draft releases.
spacing

src/explore-education-statistics-admin/src/pages/publication/components/PublicationPublishedReleases.tsx
```diff
@@ -167,6 +197,96 @@ const PublicationPublishedReleases = ({ publicationId, releases }: Props) => {
       )}
     </>
   );
-};
+}
+
+interface PublishedReleasesTableProps {
```
amyb-hiveit, line 202: I'd split this out into it's own file as it's not a small component.

amyb-hiveit (changes requested):

src/explore-education-statistics-admin/src/pages/publication/components/PublicationUnpublishedReleases.tsx
ntsim: Indeed, but it's not likely for this to ever happen. Think in prod, the most draft/scheduled releases is like 4/5. Will just bump it up to something unreasonably high anyway, just in case.

src/explore-education-statistics-admin/src/pages/publication/components/PublicationPublishedReleases.tsx
ntsim: Done

src/explore-education-statistics-admin/src/pages/publication/PublicationReleasesPage.tsx
ntsim: Good point. It's because I'd set the React Query `staleTime` option to infinity unnecessarily. Just revert this back to using the defaults!

src/explore-education-statistics-admin/src/pages/publication/components/PublicationPublishedReleases.tsx
ntsim: I spoke to Marv about this last week, as it didn't really make a whole load of sense to me at the time as just displaying the total number of releases seemed sufficient. He wasn't too fussed either way so I just got rid of the lower bound.

Having implemented the pagination properly now, the lower bound kind of makes sense again. Have added it back in!

amyb-hiveit (approved):

PR #3476, opened 2022-09-01
EES-3679 fix showing all locations for non-school data

amyb-hiveit:
Fixes a silly mistake (by me 😃) in the table tool locations step that caused ungrouped locations with more than 500 options to be treated like school level data - only the search form with no options shown until you search.

School level data should still only show the search form with no options shown until you search.


src/explore-education-statistics-common/src/components/form/FormCheckboxSearchGroup.tsx
```diff
@@ -117,7 +117,7 @@ const FormCheckboxSearchGroup = ({
           value={value}
           name={name}
           options={filteredOptions}
-          searchOnly
+          searchOnly={searchOnly}
```
ntsim, line 120: Only just looked at these search changes and got a few issues with what we've done here.

`BaseFormCheckboxGroup` is only supposed to be a simple component that handles the rendering of checkboxes. Search is something that should be handled by`FormCheckboxSearchGroup` (it's in the name).

From what I can see, it seems strange that we opted to mix both of these concerns into `BaseFormCheckboxGroup` and would call this a smell.

Would appreciate some insight into the thought process here, or otherwise, I'd recommend that we pull all of this logic out of `BaseFormCheckboxGroup` and into `FormCheckboxSearchGroup` and refactor everything accordingly.

ntsim (changes requested):

ntsim (approved):
Nice, looking a lot better! :+1:

PR #3478, opened 2022-09-01
EES-3607 find stats sorting

amyb-hiveit:
Adds sorting to the new Find Statistics page. The existing Find Statistics page should not be affected, to see the new one use the newDesign=true query parameter, e.g. http://localhost:3000/find-statistics?newDesign=true

- doesn't actually do any sorting yet as using dummy data
- displays as a select dropdown on mobile and a radio group on desktop
- I've added the mobile filters button just to sort out the styling for that section, it doesn't do anything yet.
- includes a non-javascript only submit button to the controls so you can submit the form and sort the results when js is disabled. 
- pagination gets reset when results are sorted
- a Google Analytics event is fired when change the sorting
- I've added an inline option to FormSelect for use on the mobile version

sorting-mobile
sorting-desktop

### DIFF
--- a/src/explore-education-statistics-common/src/components/form/FormSelect.tsx
+++ b/src/explore-education-statistics-common/src/components/form/FormSelect.tsx
@@ -6,6 +6,7 @@ import React, {
   ChangeEventHandler,
   CSSProperties,
   FocusEventHandler,
+  ReactNode,
 } from 'react';
 import ErrorMessage from '../ErrorMessage';
 
@@ -23,6 +24,7 @@ export interface FormSelectProps extends FormLabelProps {
   disabled?: boolean;
   error?: string;
   id: string;
+  inline?: boolean;
   hint?: string;
   name: string;
   onBlur?: FocusEventHandler;
@@ -42,6 +44,7 @@ const FormSelect = ({
   disabled,
   error,
   id,
+  inline = false,
   hint,
   hideLabel,
   label,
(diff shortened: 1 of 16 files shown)


src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPageNew.tsx
```diff
@@ -1,29 +1,95 @@
...
+  >(publications);
+  const [currentSortBy, setCurrentSortBy] = useState<
+    PublicationSortOptions | undefined
+  >(sortBy ?? undefined);
+  const [isLoadingPublications, toggleIsLoadingPublications] = useToggle(false);
+
+  useEffect(() => {
+    setCurrentPage(page);
+    setCurrentPublications(publications);
+  }, [page, publications]);
```
ntsim, line 51: How come we're doing everything in state again?

Noticed your comment that there are 'usability problems as the page reloads while you're interacting with the sort controls' . What actually happens?

I stripped out all the state and changed the `router.push` back to be non-shallow and everything seems to work fine. Focus is not lost and I can cycle through the sort options easily enough. Is there a screen reader issue here?

src/explore-education-statistics-frontend/src/modules/find-statistics/components/SortControls.tsx
```diff
@@ -0,0 +1,66 @@
...
+  { label: 'Oldest', value: 'oldest' },
+  { label: 'A to Z', value: 'alphabetical' },
+];
+
+interface FormValues {
+  sortBy: PublicationSortOptions;
+}
+
+interface Props {
+  initialValues: FormValues;
```
ntsim, line 19: Don't need to do this as `initialValues` as this isn't a particularly complicated form (we're just concerned with sorting). Could just use a `sortBy` prop to simplify things.

src/explore-education-statistics-frontend/src/modules/find-statistics/components/__tests__/SortControls.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+    expect(sortOptions[0]).toEqual(sortGroup.getByLabelText('Newest'));
+    expect(sortOptions[0]).toBeChecked();
+    expect(sortOptions[1]).toEqual(sortGroup.getByLabelText('Oldest'));
+    expect(sortOptions[1]).not.toBeChecked();
+    expect(sortOptions[2]).toEqual(sortGroup.getByLabelText('A to Z'));
+    expect(sortOptions[2]).not.toBeChecked();
+
+    expect(screen.queryByLabelText('Sort results')).not.toBeInTheDocument();
+  });
+
```
ntsim, line 35: We should have a test to check that providing a different initial `sortBy` sets a different selected value (for desktop and mobile).

src/explore-education-statistics-common/src/components/form/FormSelect.tsx
```diff
@@ -57,69 +60,104 @@ const FormSelect = ({
 }: FormSelectProps) => {
   return (
     <>
-      <FormLabel id={id} label={label} hideLabel={hideLabel} />
+      <FormSelectWrapper inline={inline}>
```
ntsim, line 63: Don't need to split out a separate component for this. It's a bit of unnecessary indirection.

Would just implement this as a wrapping `div` with conditional `className` for simplicity.

ntsim (changes requested):

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPageNew.tsx
amyb-hiveit: I'm not sure what I was doing before to make it not work properly! I've taken it back out again and all works fine.

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPageNew.tsx
```diff
@@ -83,41 +149,53 @@ const FindStatisticsPageNew: NextPage<Props> = ({ paging, publications }) => {
...
+                  variant="secondary"
+                >
+                  Filter results
+                </Button>
+              )}
+              <SortControls
+                initialValues={{ sortBy: currentSortBy ?? 'newest' }}
+                onChange={sortPublications}
+              />
+            </div>
```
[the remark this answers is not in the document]
amyb-hiveit: Showed this to Marv and he agreed so have changed it as you suggest.

src/explore-education-statistics-common/src/components/form/FormSelect.tsx
amyb-hiveit: Unnecessary divs make me sad.

ntsim (approved):

PR #3480, opened 2022-09-02
EES-3681 - fix fast track UI test

luke-h1:
This PR:
fixes the fast track UI test. This was failing due to visually hidden text being inside the 'explore data' link which caused the keyword `user clicks link` to fail. We've changed this to `user clicks link by visible text` which will ignore any screen reader only text

### DIFF
--- a/tests/robot-tests/tests/general_public/fast_track.robot
+++ b/tests/robot-tests/tests/general_public/fast_track.robot
@@ -18,10 +18,10 @@ Click fast track link for 'Pupil absence rates' data block
     user opens accordion section    Pupil absence rates    id:content
     user scrolls to accordion section content    Pupil absence rates    id:content
     user waits until h3 is visible    Explore and edit this data online
-    user clicks link    Explore data
+    user clicks link by visible text    Explore data    testid:Data block - Generic data block - National
 
 Validate Publication selected step option
-    user waits until h1 is visible    Create your own tables
+    user waits until h1 is visible    Create your own tables    %{WAIT_SMALL}
     user waits until page contains element    css:table
     user checks previous table tool step contains    1    Publication    Pupil absence in schools in England
 


tests/robot-tests/tests/general_public/fast_track.robot
```diff
@@ -9,19 +9,11 @@ Test Setup          fail test fast if required
...
-    user navigates to public frontend    %{PUBLIC_URL}/find-statistics/pupil-absence-in-schools-in-england
-
-Click fast track link for 'Pupil absence rates' data block
-    user waits until h1 is visible    Pupil absence in schools in England
-    user opens accordion section    Pupil absence rates    id:content
-    user scrolls to accordion section content    Pupil absence rates    id:content
-    user waits until h3 is visible    Explore and edit this data online
-    user clicks link    Explore data
+Navigate to fast track link for 'Pupil absence rates' data block
+    user navigates to public frontend    %{PUBLIC_URL}/data-tables/fast-track/34beb07c-cac1-492b-8829-864abf8ac0e6
```
benoutram, line 13: Does this data block have the same id in all environments? It would be safer to click through to this from the button in the content rather than rely on the id.

tests/robot-tests/tests/general_public/fast_track.robot
```diff
@@ -9,19 +9,11 @@ Test Setup          fail test fast if required
...
-Navigate to publication release page
-    environment variable should be set    PUBLIC_URL
-    user navigates to public frontend    %{PUBLIC_URL}/find-statistics/pupil-absence-in-schools-in-england
-
-Click fast track link for 'Pupil absence rates' data block
-    user waits until h1 is visible    Pupil absence in schools in England
-    user opens accordion section    Pupil absence rates    id:content
-    user scrolls to accordion section content    Pupil absence rates    id:content
-    user waits until h3 is visible    Explore and edit this data online
-    user clicks link    Explore data
```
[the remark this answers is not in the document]
luke-h1: Looks like this ended up being because of screen reader only text being inside the 'explore data' lin

tests/robot-tests/tests/general_public/fast_track.robot
```diff
@@ -9,19 +9,15 @@ Test Setup          fail test fast if required
 
 
 *** Test Cases ***
-Navigate to publication release page
-    environment variable should be set    PUBLIC_URL
-    user navigates to public frontend    %{PUBLIC_URL}/find-statistics/pupil-absence-in-schools-in-england
-
```
ntsim, line 15: We've still deleted this? How would we actually navigate to the right place without this?

luke-h1: Ah yeah, good catch. I've added this back in

ntsim (approved):

PR #3481, opened 2022-09-02
EES-3677-2 disable reorder controls when release is published

luke-h1:
This PR:

disables re-order controls if a user can't update a release (such as when it's being published/approved). Previously, the re-order data-files still showed up even when a release was in the process of publishing or had the status of approved

### DIFF
--- a/src/explore-education-statistics-admin/src/components/editable/ReorderableAccordion.tsx
+++ b/src/explore-education-statistics-admin/src/components/editable/ReorderableAccordion.tsx
@@ -1,5 +1,6 @@
 import Accordion, { AccordionProps } from '@common/components/Accordion';
 import Button from '@common/components/Button';
+import VisuallyHidden from '@common/components/VisuallyHidden';
 import useToggle from '@common/hooks/useToggle';
 import { OmitStrict } from '@common/types';
 import reorder from '@common/utils/reorder';
@@ -24,10 +25,19 @@ export interface ReorderableAccordionProps
   extends OmitStrict<AccordionProps, 'openAll'> {
   heading?: string;
   onReorder: (sectionIds: string[]) => void;
+  canReorder: boolean;
+  reorderHiddenText?: string;
 }
 
 const ReorderableAccordion = (props: ReorderableAccordionProps) => {
-  const { children, id, heading, onReorder } = props;
+  const {
+    children,
+    id,
+    heading,
+    onReorder,
+    canReorder,
+    reorderHiddenText = 'sections',
+  } = props;
 
   const [isReordering, toggleReordering] = useToggle(false);
 
@@ -82,13 +92,16 @@ const ReorderableAccordion = (props: ReorderableAccordionProps) => {
         <h2 className="govuk-heading-l govuk-!-margin-bottom-0">{heading}</h2>
 
         {sections.length > 1 &&
+          canReorder &&
           (!isReordering ? (
             <Button
               variant="secondary"
               className="govuk-!-font-size-16 govuk-!-margin-bottom-0"
               onClick={toggleReordering.on}
+              data-testid="reorder-files"
             >
-              Reorder<span className="govuk-visually-hidden"> data files</span>
+              Reorder
+              <VisuallyHidden> {reorderHiddenText}</VisuallyHidden>
             </Button>
           ) : (
             <Button
--- a/src/explore-education-statistics-admin/src/components/editable/__tests__/ReorderableAccordion.test.tsx
+++ b/src/explore-education-statistics-admin/src/components/editable/__tests__/ReorderableAccordion.test.tsx
@@ -1,0 +1,73 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import noop from 'lodash/noop';
+import ReorderableAccordion from '../ReorderableAccordion';
+import ReorderableAccordionSection from '../ReorderableAccordionSection';
+
+describe('ReorderableAccordion', () => {
+  test("doesn't show controls if `canReorder` is false", () => {
+    render(
+      <ReorderableAccordion
+        canReorder={false}
+        heading="test accordion"
+        reorderHiddenText="files"
+        id="1"
+        onReorder={noop}
+      >
+        <ReorderableAccordionSection id="1" heading="test section one">
+          test accordion content section one
+        </ReorderableAccordionSection>
+
+        <ReorderableAccordionSection id="1" heading="test section two">
+          test accordion content section two
+        </ReorderableAccordionSection>
+      </ReorderableAccordion>,
+    );
+
+    expect(
+      screen.queryByRole('button', { name: 'Reorder files' }),
+    ).not.toBeInTheDocument();
+
+    expect(
+      screen.getByText('test accordion content section one'),
+    ).toBeInTheDocument();
+    expect(
+      screen.getByText('test accordion content section two'),
+    ).toBeInTheDocument();
+  });
+
+  test('shows controls if `canReorder` is true', () => {
+    const { container } = render(
+      <ReorderableAccordion
+        canReorder
+        heading="test accordion"
+        reorderHiddenText="files"
+        id="1"
+        onReorder={noop}
+      >
+        <ReorderableAccordionSection id="1" heading="test section one">
+          test accordion content section one
+        </ReorderableAccordionSection>
+
+        <ReorderableAccordionSection id="2" heading="test section two">
+          test accordion content section two
+        </ReorderableAccordionSection>
+      </ReorderableAccordion>,
+    );
+
+    expect(container.querySelector('.govuk-visually-hidden')).toHaveTextContent(
+      'files',
+    );
+
+    expect(
+      screen.getByRole('button', { name: 'Reorder files' }),
+    ).toBeInTheDocument();
+
+    expect(
+      screen.getByText('test accordion content section one'),
+    ).toBeInTheDocument();
+    expect(
+      screen.getByText('test accordion content section two'),
+    ).toBeInTheDocument();
+  });
+});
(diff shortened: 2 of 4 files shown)


src/explore-education-statistics-admin/src/components/editable/ReorderableAccordion.tsx
```diff
@@ -24,10 +25,11 @@ export interface ReorderableAccordionProps
   extends OmitStrict<AccordionProps, 'openAll'> {
   heading?: string;
   onReorder: (sectionIds: string[]) => void;
+  canUpdateRelease: boolean;
```
ntsim, line 28: Would name this more prop more generically i.e. `canReorder`. Where possible, we should think about creating components that can be used in broader contexts than just a single specific use-case :wink:

src/explore-education-statistics-admin/src/components/editable/ReorderableAccordion.tsx
```diff
@@ -82,21 +84,24 @@ const ReorderableAccordion = (props: ReorderableAccordionProps) => {
...
         {sections.length > 1 &&
+          canUpdateRelease &&
           (!isReordering ? (
             <Button
               variant="secondary"
               className="govuk-!-font-size-16 govuk-!-margin-bottom-0"
               onClick={toggleReordering.on}
             >
-              Reorder<span className="govuk-visually-hidden"> data files</span>
+              Reorder<VisuallyHidden> data files</VisuallyHidden>
```
ntsim, line 94: Would create a `reorderHiddenText` prop for this text. Would suggest defaulting it to 'sections'.

Think you've done something like this in a previous PR so it's basically more of the same!

src/explore-education-statistics-admin/src/components/editable/ReorderableAccordion.tsx
```diff
@@ -82,21 +84,24 @@ const ReorderableAccordion = (props: ReorderableAccordionProps) => {
...
             </Button>
           ) : (
-            <Button
-              className="govuk-!-font-size-16 govuk-!-margin-bottom-0"
-              onClick={saveOrder}
-            >
-              Save order
-            </Button>
+            <>
```
ntsim, line 97: Unnecessary wrapping fragment

src/explore-education-statistics-admin/src/components/editable/__tests__/ReorderableAccordion.test.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import noop from 'lodash/noop';
+import Button from '@common/components/Button';
+import ReorderableAccordion from '../ReorderableAccordion';
+
+describe('ReorderableAccordion', () => {
+  test("it doesn't show controls if `canUpdateRelease` is false", () => {
```
ntsim, line 8: Bit of nitpick, but don't need to write 'it' at the start of test cases as it's a bit of unnecessary filler.

Same with the other test case.

src/explore-education-statistics-admin/src/components/editable/__tests__/ReorderableAccordion.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+    const canUpdateRelease = false;
+
+    render(
+      <ReorderableAccordion
+        canUpdateRelease={canUpdateRelease}
+        id="test"
+        onReorder={() => noop}
+        heading="test"
+        onSectionOpen={() => noop}
+        onToggleAll={() => noop}
```
ntsim, line 18: This isn't the best way of using `noop` We typically do this:

```tsx
onReorder={noop}
onSectionOpen={noop}
onToggleAll={noop}
```

Same with the other test case

src/explore-education-statistics-admin/src/components/editable/__tests__/ReorderableAccordion.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+      >
+        {canUpdateRelease && <Button>Reorder</Button>}
```
ntsim, line 20: Why try to render this? Given that `canUpdateRelease` is false, this would never render, but doesn't seem like we need this for the test at all

src/explore-education-statistics-admin/src/components/editable/__tests__/ReorderableAccordion.test.tsx
```diff
@@ -0,0 +1,46 @@
...
```
ntsim, line 40: Same as earlier - I'm not sure why we need this?

The assertion should be able to look for a single 'Reorder sections' (as suggested) button.

ntsim (changes requested):

src/explore-education-statistics-admin/src/components/editable/__tests__/ReorderableAccordion.test.tsx
luke-h1: Cool good to know. I assumed that since you'd use `() => {}` to denote a function that returns void (if you're not using lodash) you'd do the same with lodash.

src/explore-education-statistics-admin/src/components/editable/__tests__/ReorderableAccordion.test.tsx
luke-h1: Yep, not sure why I put this in. Think this is just a hangover from trying to rush this fix.

luke-h1: I was getting a failure around no button being rendered since in the test it's rendering no children I assume

ntsim: If there are no children, we don't render the button so that should be another test case.

For the test cases on `canReorder`, it'd be better to try and render a more realistic test case e.g.

```tsx
<ReorderableAccordion canReorder>
  <ReorderableAccordionSection ...>
    Some test text
  </ReorderableAccordionSection>
</ReorderableAccordion>
```

Same applies for my other comment below.

src/explore-education-statistics-admin/src/components/editable/ReorderableAccordion.tsx
```diff
@@ -82,13 +92,18 @@ const ReorderableAccordion = (props: ReorderableAccordionProps) => {
...
             <Button
               variant="secondary"
               className="govuk-!-font-size-16 govuk-!-margin-bottom-0"
               onClick={toggleReordering.on}
             >
-              Reorder<span className="govuk-visually-hidden"> data files</span>
+              Reorder
+              <VisuallyHidden>
+                {' '}
+                {reorderHiddenText ?? 'sections'}
```
ntsim, line 105: Would do this as a single template string to avoid nodes splitting the text

src/explore-education-statistics-admin/src/components/editable/ReorderableAccordion.tsx
```diff
@@ -24,10 +25,19 @@ export interface ReorderableAccordionProps
...
 
 const ReorderableAccordion = (props: ReorderableAccordionProps) => {
-  const { children, id, heading, onReorder } = props;
+  const {
+    children,
+    id,
+    heading,
+    onReorder,
+    canReorder,
+    reorderHiddenText,
```
ntsim, line 39: Could set the default here instead of in the markup

src/explore-education-statistics-admin/src/components/editable/__tests__/ReorderableAccordion.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+        onSectionOpen={noop}
+        onToggleAll={noop}
+      >
+        {canReorder && <Button>Reorder</Button>}
+      </ReorderableAccordion>,
+    );
+
+    expect(
+      screen.queryByRole('button', { name: 'Reorder' }),
+    ).toBeInTheDocument();
```
ntsim, line 47: This doesn't actually test that the 'Reorder sections' buttons exists. All we're testing is that the button in the children renders.

ntsim (changes requested):

tests/robot-tests/tests/admin_and_public/bau/subject_reordering.robot
```diff
@@ -182,6 +182,17 @@ Add data guidance for all subjects
...
+Check subjects can no longer be re-ordered after release has been published
+    user clicks link    Data and files
+    user waits until page contains element    id:uploadedDataFiles
+
+    user checks accordion is in position    One    1    id:uploadedDataFiles
+    user checks accordion is in position    Two    2    id:uploadedDataFiles
+    user checks accordion is in position    Three    3    id:uploadedDataFiles
+    user checks accordion is in position    Four    4    id:uploadedDataFiles
+
+    user checks element is not visible    //*[button[text()="Reorder"]]    %{WAIT_SMALL}
```
ntsim, line 194: This selector would never fail as it's matching on exact text and the button has visually hidden text that is breaking up the element.

Would probably be best to use a test id instead

src/explore-education-statistics-admin/src/pages/release/data/components/ReleaseDataUploadsSection.tsx
```diff
@@ -231,6 +231,7 @@ const ReleaseDataUploadsSection = ({
       <LoadingSpinner loading={isLoading}>
         {dataFiles.length > 0 ? (
           <ReorderableAccordion
+            canReorder={canUpdateRelease}
```
ntsim, line 234: Just noticed this, but given that we're reordering files, shouldn't this have `reorderHiddenText="files"`?

ntsim (approved):

PR #3482, opened 2022-09-05
EES-3028, EES-3029, EES-3030, EES-3032 - address combobox + page search accessibility issues

luke-h1:
This PR addresses the following accessibility issues related to `PageSearchForm` & `FormComboBox` which are used on Find stats / Methodology pages.

* [EES-3028](https://dfedigital.atlassian.net/browse/EES-3028) - Ensure that all interactive components can be found both in and out of context for screen reader users
  * [EES-3029](https://dfedigital.atlassian.net/browse/EES-3029) - remove search button from `PageSearchForm` 
  * [EES-3030](https://dfedigital.atlassian.net/browse/EES-3030) - Fix the ability for screen reader users to find the search element 
  * [EES-3032](https://dfedigital.atlassian.net/browse/EES-3032) - Ensure that all interactive components can be found both in & out of context for screen reader users

Relevant changes:
* Added `aria-live="polite"` & `aria-atomic` to results label
* Adds two new aria-labels: `aria-posinset` (inform user where they are in the list) & `aria-setsize` (inform user how many items are in the list)
* Moves `role="combobox"` to `input` element

Example:
https://www.loom.com/share/66f7fb26d16c4b2f99ef7774724f025b (demonstrates using tabs + arrows etc.)

UI test run:
image

### DIFF
--- a/src/explore-education-statistics-common/src/components/form/FormComboBox.tsx
+++ b/src/explore-education-statistics-common/src/components/form/FormComboBox.tsx
@@ -7,9 +7,11 @@ import React, {
   InputHTMLAttributes,
   KeyboardEvent,
   ReactNode,
+  useEffect,
   useRef,
   useState,
 } from 'react';
+import inputStyles from '@common/components/form/FormTextSearchInput.module.scss';
 import styles from './FormComboBox.module.scss';
 
 interface ComboBoxRenderProps {
@@ -51,11 +53,11 @@ const FormComboBox = ({
   const inputRef = useRef<HTMLInputElement>(null);
   const optionRefs = useRef<Dictionary<HTMLLIElement>>({});
 
-  const [value, setValue] = useState('');
-  const [selectedOption, setSelectedOption] = useState(initialOption);
+  const [value, setValue] = useState<string>('');
+  const [selectedOption, setSelectedOption] = useState<number>(initialOption);
   const [showOptions, toggleShowOptions] = useToggle(false);
 
-  React.useEffect(() => {
+  useEffect(() => {
     setValue(inputValue);
   }, [inputValue]);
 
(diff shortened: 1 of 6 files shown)


src/explore-education-statistics-common/src/components/form/FormComboBox.tsx
```diff
@@ -156,158 +158,210 @@ const FormComboBox = ({
...
+                    aria-posinset={index + 1}
+                    className={
+                      selectedOption === index ? styles.selected : undefined
+                    }
+                    role="option"
+                    tabIndex={-1}
+                    ref={el => {
+                      if (el) {
+                        if (selectedOption === index) {
+                          el.focus();
```
ntsim, line 295: This is a smell - we shouldn't be focusing elements like this. Refs shouldn't be used for side effects.

Instead, we should do something like `PublicationPublishedReleasesTable` which refocuses whenever a `focusReleaseId` changes. In our case, it'd look something like:

```tsx
const optionRefs = useRef<Dictionary<HTMLLIElement>>({});

useEffect(() => {
  if (selectedOption) {
    optionRefs.current[selectedOption]?.focus();
  }
}, [selectedOption]);


// Revert the `ref` back to what it was previously

<li
  ref={el => {
    if (el) {
      optionRefs.current[key] = el;
    }
  }}
>
```

src/explore-education-statistics-common/src/components/form/FormComboBox.tsx
```diff
@@ -156,158 +158,210 @@ const FormComboBox = ({
...
+          >
+            {renderedOptions &&
+              renderedOptions.length > 0 &&
+              renderedOptions.map((item, index) => {
+                const key = index;
+
+                return (
+                  // eslint-disable-next-line jsx-a11y/click-events-have-key-events
+                  <li
+                    aria-selected={selectedOption === index}
```
ntsim, line 282: We could refactor all of the `selectedOption === index` expressions into a single `isSelected` variable and re-use this across the various props

src/explore-education-statistics-common/src/components/form/__tests__/FormComboBox.test.tsx
```diff
@@ -173,632 +179,552 @@ describe('FormComboBox', () => {
...
+
+        const input = getByLabelText('Choose option') as HTMLInputElement;
+
+        fireEvent.change(input, {
+          target: { value: 'Test' },
+        });
+
+        const comboBox = container.querySelector(
+          '[role="combobox"]',
+        ) as HTMLElement;
```
ntsim, line 202: Does `getByRole` work here? Same question for other places where we're doing this kind of selector as well.

src/explore-education-statistics-common/src/components/form/FormComboBox.tsx
```diff
@@ -156,158 +158,210 @@ const FormComboBox = ({
...
+          onInputChange(event);
+          toggleShowOptions(true);
+        }}
+        onKeyDown={event => {
+          if (event.key === 'ArrowUp' || event.key === 'ArrowDown') {
+            selectNextOption(event);
+
+            if (listBoxRef.current) {
+              listBoxRef.current.focus();
             }
```
ntsim, line 199: We don't need this anymore - it'll also interfere with the `useEffect` based way of focusing an option

src/explore-education-statistics-common/src/components/form/__tests__/FormComboBox.test.tsx
```diff
@@ -173,632 +179,552 @@ describe('FormComboBox', () => {
...
+            onSelect={noop}
+            options={['Option 1', 'Option 2', 'Option 3']}
+          />,
+        );
+
+        const input = getByLabelText('Choose option') as HTMLInputElement;
+
+        fireEvent.change(input, {
+          target: { value: 'Test' },
+        });
```
ntsim, line 198: We could probably change this to use `userEvent.type`?

Would refactor all the other test cases where we're changing this field if so.

src/explore-education-statistics-common/src/components/form/__tests__/FormComboBox.test.tsx
```diff
@@ -173,632 +179,552 @@ describe('FormComboBox', () => {
...
-      );
-
-      const input = getByLabelText('Choose option') as HTMLInputElement;
-
-      fireEvent.change(input, {
-        target: {
-          value: 'Test',
-        },
+    describe('list box', () => {
+      test('pressing ArrowDown will cycle the selected item through the entire list', () => {
```
ntsim, line 183: Shouldn't this test case be 'pressing ArrowUp will cycle the selected item through the entire list'?

We've already got another test case further down that is testing 'pressing ArrowDown will cycle the selected item through the entire list'

src/explore-education-statistics-common/src/components/form/__tests__/FormComboBox.test.tsx
```diff
@@ -173,632 +179,552 @@ describe('FormComboBox', () => {
...
+        expect(option2).toHaveAttribute('aria-selected', 'false');
+        expect(option3).toHaveAttribute('aria-selected', 'false');
+
+        fireEvent.keyDown(comboBox, { key: 'ArrowDown' });
+
+        expect(option1).toHaveAttribute('aria-selected', 'true');
+        expect(option2).toHaveAttribute('aria-selected', 'false');
+        expect(option3).toHaveAttribute('aria-selected', 'false');
+
+        fireEvent.keyDown(comboBox, { key: 'ArrowDown' });
```
ntsim, line 218: This seems to be testing pressing arrow down to move through the list, but the test case originally was testing that you could press ArrowUp and go to to the end of the list and back up it. Why have we changed this?

Again, we have another test case further down called 'pressing ArrowDown will cycle the selected item through the entire list' that already tests going down the list. We should be testing going up the list!

src/explore-education-statistics-common/src/components/form/__tests__/FormComboBox.test.tsx
```diff
@@ -173,632 +179,552 @@ describe('FormComboBox', () => {
...
+          <FormComboBox
+            id="test-combobox"
+            inputLabel="Choose option"
+            onInputChange={noop}
+            onSelect={noop}
+            options={['Option 1', 'Option 2', 'Option 3']}
+          />,
+        );
+
+        const input = getByLabelText('Choose option') as HTMLInputElement;
```
ntsim, line 194: If we refactor `fireEvent.change` to `userEvent.type` below, pretty sure we don't need a type assertion that this is a `HTMLInputElement` anymore. We could simplify this code to something like:

```ts
await userEvent.type(screen.getByLabelText('Choose option'), 'Test');
```

Same would apply to other test cases where we're doing this.

ntsim (changes requested):

src/explore-education-statistics-common/src/components/form/__tests__/FormComboBox.test.tsx
```diff
@@ -51,40 +50,46 @@ describe('FormComboBox', () => {
 
   describe('input field', () => {
     test('pressing ArrowDown focuses the list box', () => {
-      const { container, getByLabelText } = render(
+      const { getByLabelText, getByText } = render(
```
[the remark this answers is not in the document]
luke-h1: Sounds good to me

src/explore-education-statistics-common/src/components/form/__tests__/FormComboBox.test.tsx
luke-h1: Yep think you're right. Must have duplicated this one accidentaly

src/explore-education-statistics-common/src/components/form/FormComboBox.tsx
```diff
@@ -156,158 +158,210 @@ const FormComboBox = ({
...
+                          case 'Tab':
+                            setValue('');
+                            toggleShowOptions(false);
+                            break;
+
+                          default:
+                            break;
+                        }
+                      }
+                    }}
```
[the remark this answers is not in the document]
luke-h1: Not too sure why this was duplicated. I've removed this from the `li` & just placed it on the unordered list instead

src/explore-education-statistics-common/src/components/form/__tests__/FormComboBox.test.tsx
luke-h1: Yeah looks like it does, I've refactored this & other parts to use `screen.getByRole` instead

src/explore-education-statistics-common/src/components/form/FormComboBox.tsx
luke-h1: Had a look at this & it seems for some reason that the selectedOption gets set twice when doing this which results in two li elements sometimes having the `selected` class at the same time. In addition to that, the `useEffect` doesn't seem to be setting the focus on the list item instead & the focus stays on the input. For now, I've done the following which is a bit hacky but seems to work. In the future, it might be worth re-working this component to get around this problem.

```typescript 
 ref={el => {
                      if (el) {
                        optionRefs.current[index] = el;

                        if (isSelected) {
                          // this is a bit hacky but it seems to be the only way to focus
                          // without two list items being selected at the same time
                          // (and the focus being on the input instead of the list item)
                          el.focus();
                        }
                      }
                    }}

```

src/explore-education-statistics-common/src/components/form/__tests__/FormComboBox.test.tsx
```diff
@@ -1,804 +1,812 @@
...
+      .create(
+        <FormComboBox
+          id="test-combobox"
+          inputLabel="Choose option"
+          onInputChange={noop}
+          onSelect={noop}
+          options={['Option 1', 'Option 2', 'Option 3']}
+        />,
+      )
+      .toJSON();
```
ntsim, line 34: Why do this? It's okay to use `container` from the `render` above to test that the HTML matches the snapshot

src/explore-education-statistics-common/src/components/form/__tests__/FormComboBox.test.tsx
```diff
@@ -1,804 +1,812 @@
...
 
-      fireEvent.keyDown(getByLabelText('Choose option'), { key: 'ArrowUp' });
+      fireEvent.keyDown(screen.getByLabelText('Choose option'), {
+        key: 'ArrowUp',
+      });
 
-      expect(option1).toHaveAttribute('aria-selected', 'false');
-      expect(option2).toHaveAttribute('aria-selected', 'false');
-      expect(option3).toHaveAttribute('aria-selected', 'true');
+      await waitFor(() => {
```
ntsim, line 215: Do we need this `waitFor` anymore? As we discussed, it seems like the ref-based focus method seems to work better so don't think we need it?

Would also do it in any other place where we've done something similar to tidy up the test code a bit.

src/explore-education-statistics-common/src/components/form/__tests__/FormComboBox.test.tsx
```diff
@@ -1,804 +1,812 @@
...
     });
-  });
+    describe('list box', () => {
+      test('pressing ArrowUp will cycle the selected item through the entire list', async () => {
+        render(
+          <FormComboBox
+            id="test-combobox"
+            inputLabel="Choose option"
+            onInputChange={() => {}}
+            onSelect={() => {}}
```
ntsim, line 233: Could use `noop` here?

src/explore-education-statistics-common/src/components/form/__tests__/FormComboBox.test.tsx
```diff
@@ -1,804 +1,812 @@
...
+          id="test-combobox"
+          inputLabel="Choose option"
+          onInputChange={noop}
+          onSelect={noop}
+          options={['Option 1', 'Option 2', 'Option 3']}
+        />,
+      )
+      .toJSON();
+
+    expect(tree).toMatchSnapshot();
```
ntsim, line 71: Same question as earlier - not sure why we're doing this? We can just use `container` from the earlier `render` to check the snapshot HTML

src/explore-education-statistics-common/src/components/form/__tests__/FormComboBox.test.tsx
```diff
@@ -1,804 +1,812 @@
...
-          onInputChange={() => {}}
-          onSelect={() => {}}
-          options={['Option 1', 'Option 2', 'Option 3']}
-        />
-      </div>,
-    );
+        expect(onSelect).not.toHaveBeenCalled();
 
-    const input = getByLabelText('Choose option') as HTMLInputElement;
+        fireEvent.click(screen.getByText('Option 2'));
```
ntsim, line 740: We can change uses of `fireEvent.click` to `userEvent.click`

ntsim (changes requested):

src/explore-education-statistics-common/src/components/form/__tests__/FormComboBox.test.tsx
luke-h1: Cool, didn't realize that we could use both the destructuring way of doing things as well as using `screen` from `@testing-library`

ntsim (approved):

PR #3484, opened 2022-09-06
EES-3686, EES-3684 - fix table-tool + data-catalogue accessibility issues

luke-h1:
This PR:
* Adds visually hidden text to the 'More details' dropdown displayed on step 2 + step 3 of the table tool & data-catalogue pages respectively. 

This addresses an accessibility issue where 'More details' are repeated multiple times on a page which can confuse screen readers / not provide them with enough context

Relevant changes:
* Updates `Details` components to ensure any visually hidden text has the appropriate spacing in the markup.

### DIFF
--- a/src/explore-education-statistics-common/src/components/Details.tsx
+++ b/src/explore-education-statistics-common/src/components/Details.tsx
@@ -141,7 +141,7 @@ const Details = ({
           data-testid={formatTestId(`Expand Details Section ${summary}`)}
         >
           {summary}
-          {hiddenText && <VisuallyHidden> {hiddenText}</VisuallyHidden>}
+          {hiddenText && <VisuallyHidden>{` ${hiddenText}`}</VisuallyHidden>}
         </span>
         {summaryAfter}
       </summary>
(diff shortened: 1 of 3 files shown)


src/explore-education-statistics-common/src/components/Details.tsx
```diff
@@ -141,7 +141,9 @@ const Details = ({
           data-testid={formatTestId(`Expand Details Section ${summary}`)}
         >
           {summary}
-          {hiddenText && <VisuallyHidden> {hiddenText}</VisuallyHidden>}
+          {hiddenText && (
+            <VisuallyHidden>{`${' '} ${hiddenText}`}</VisuallyHidden>
```
ntsim, line 145: This is a template string so there's no need to interpolate a string into it.  Additionally, this is placing 2 spaces rather than one. The following would be sufficient:

```tsx
<VisuallyHidden>{` ${hiddenText}`}</VisuallyHidden>
```

ntsim (changes requested):

ntsim (approved):

PR #3488, opened 2022-09-08
EES-3645 Update cached publications at 9.30 rather than deleting them from the cache

benoutram:
This PR alters the publisher to update cached publications in the public blob storage cache with `PublicationCacheService` when the function runs at 9.30, rather than deleting them.

As part of this work an issue was spotted around where this happens.

The `PublishingService.PublishStagedReleaseContent` method was being invoked once _per release_ that's published, rather than once per function execution. It was performing 4 actions:

* Moving all cached releases in the staging directory of the public content cache to the root to make them live.
* Deleting the release's publication from the public cache.
* Working out any superseded publications and deleting those from the public cache.
* Setting the published date on the release.

Moving the staging directory only needs to happen once per function execution and isn't dependent on a single release. Luckily  this hasn't been causing an error on subsequent attempts to move it. The staging directory no longer exists after the first move, but the move (which is a copy followed by a delete) is lenient and we'd have seen this in the logs:

```
[2022-09-08T09:30:00.807Z] Copying directory from cache/staging to cache/
[2022-09-08T09:30:00.847Z] Deleting blobs from public-cache/staging/
```

Deleting cached publications and cached superseded  publications only needs to happen based on the set of publications affected by the releases being published. It's unlikely but if there were multiple releases being published at the same time for the same publication it only needs to happen once.

The only action that needs to happen per release is setting the published date. The `PublishStagedReleaseContent` function has been changed to do this per release being published.

`PublishingService.PublishStagedReleaseContent` has been limited to only moving the staged release content.  It's now only called once by the `PublishStagedReleaseContent` function.

The set of publications affected by the published releases is then worked out and the publications are updated in the public cache instead of being deleted.

### UI Test report

image

### Other changes

* Small tidy-up of Publisher, removing unused classes and methods.


duncan-at-hiveit (approved):

benoutram:
Re-ran UI tests after merging latest changes and did manual testing with scheduled releases and archived publications.

PR #3497, opened 2022-09-13
EES-3685 - remove methodology heading from methodology listings

luke-h1:
This PR removes methodology headings from the methoolody index page in order to improve accesibility. 

- [x]  Fix methodology UI test python utils (`get_methodology_` variations) - add testid?

image

### DIFF
--- a/src/explore-education-statistics-frontend/src/modules/methodologies/components/MethodologyList.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/methodologies/components/MethodologyList.tsx
@@ -16,26 +16,17 @@ function MethodologyList({ publications }: Props) {
   }
 
   return (
-    <ul>
+    <ul data-testid="methodology-list">
       {filteredPublications.map(publication => (
-        <li className="govuk-!-margin-bottom-2" key={publication.id}>
-          <h3
-            className="govuk-heading-s govuk-!-margin-bottom-0"
-            id={`methodology-heading-${publication.id}`}
-          >
-            {publication.title}
-          </h3>
-
-          <ul>
-            {publication.methodologies.map(methodology => (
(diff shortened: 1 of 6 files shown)


src/explore-education-statistics-frontend/src/modules/methodologies/components/MethodologyList.tsx
```diff
@@ -16,26 +16,17 @@ function MethodologyList({ publications }: Props) {
...
-        </li>
+        <>
+          {publication.methodologies.map(methodology => (
+            <li key={methodology.id} className="govuk-!-margin-bottom-2">
+              <Link to={`/methodology/${methodology.slug}`}>
+                {methodology.title}
+              </Link>
+            </li>
+          ))}
+        </>
```
amyb-hiveit, line 29: Code-wise this is fine and is what the ticket said to do, but I think it will cause some different issues.

The problem is that we're listing the methodologies by publication, and methodologies often have the same name as the publication, so it looks like the methodology name is being duplicated. However, methodology names can be different to the publication and they can be shared among multiple publications, so by removing the heading we're removing the context and in some cases will end up with a list of duplicate links. For example Pupil Absence has 3 publications that share 'Pupil absence statistics: methodology', so without the publication titles we'd just have 3 links to the same methodology. 

methodologylist

I'm not sure what the best solution for this is, probably a question for @mrmarvino

amyb-hiveit (changes requested):

mrmarvino:
Based on @amyb-hiveit  comment I think we should revert this change. We need to look at a design refresh to make the page less cluttered with bullets and headers, possibly a new solution may have a structure similar to the new find stats page

PR #3500, opened 2022-09-14
EES-3666 New Admin API endpoints for getting/updating/removing ExternalMethodology

markhiveit:
This PR adds new endpoints for use on Admin pages that get / update / remove an external methodology from a publication.

This code also changes the Admin frontend code to use these new endpoints when appropriate instead of using the `publicationService.createPublication` and `publicationService.updatePublication` endpoints. It is no longer possible to update a publication's external methodology with the `publicationService.updatePublication` endpoint.

I haven't added any UI tests. UI tests for new publication management pages will be done at the same time that they're unhidden, and it doesn't seem worth adding UI tests for the current ways of changing an external methodology as they will soon be removed.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
@@ -7,6 +7,7 @@ using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Admin.ViewModels;
 using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.ViewModels;
+using GovUk.Education.ExploreEducationStatistics.Content.Model;
 using GovUk.Education.ExploreEducationStatistics.Content.Services.ViewModels;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
@@ -70,6 +71,13 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api
         {
             return await _publicationService
                 .GetPublication(publicationId)
+                .HandleFailuresOrOk();
+        }
+
+        [HttpGet("api/publication/{publicationId:guid}/external-methodology")]
+        public async Task<ActionResult<ExternalMethodologyViewModel>> GetExternalMethodology(Guid publicationId)
+        {
+            return await _publicationService.GetExternalMethodology(publicationId)
                 .HandleFailuresOrOk();
         }
 
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
@@ -6,6 +6,7 @@ using GovUk.Education.ExploreEducationStatistics.Admin.Requests;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Admin.ViewModels;
 using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Common.Model;
 using GovUk.Education.ExploreEducationStatistics.Common.ViewModels;
 using GovUk.Education.ExploreEducationStatistics.Content.Model;
 using GovUk.Education.ExploreEducationStatistics.Content.Services.ViewModels;
@@ -88,6 +89,15 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api
             return await _publicationService.UpdateExternalMethodology(publicationId, updatedExternalMethodology)
                 .HandleFailuresOrOk();
         }
+
+        [HttpDelete("api/publication/{publicationId:guid}/external-methodology")]
+        public async Task<ActionResult<Unit>> RemoveExternalMethodology(
+            Guid publicationId)
+        {
+            return await _publicationService.RemoveExternalMethodology(publicationId)
+                .HandleFailuresOrOk();
+        }
+
 
         [HttpGet("api/publication/{publicationId}/releases")]
         public async Task<ActionResult<PaginatedListViewModel<ReleaseSummaryViewModel>>> ListActiveReleases(
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
@@ -9,12 +9,8 @@ using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Model;
 using GovUk.Education.ExploreEducationStatistics.Common.ViewModels;
 using GovUk.Education.ExploreEducationStatistics.Content.Model;
-using GovUk.Education.ExploreEducationStatistics.Content.Services.ViewModels;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
-using LegacyReleaseViewModel = GovUk.Education.ExploreEducationStatistics.Admin.ViewModels.LegacyReleaseViewModel;
-using PublicationViewModel = GovUk.Education.ExploreEducationStatistics.Admin.ViewModels.PublicationViewModel;
-using ReleaseSummaryViewModel = GovUk.Education.ExploreEducationStatistics.Admin.ViewModels.ReleaseSummaryViewModel;
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api
 {
@@ -84,7 +80,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api
 
         [HttpPut("api/publication/{publicationId:guid}/external-methodology")]
         public async Task<ActionResult<ExternalMethodologyViewModel>> UpdateExternalMethodology(
-            Guid publicationId, ExternalMethodology updatedExternalMethodology)
+            Guid publicationId, ExternalMethodologySaveRequest updatedExternalMethodology)
         {
             return await _publicationService.UpdateExternalMethodology(publicationId, updatedExternalMethodology)
                 .HandleFailuresOrOk();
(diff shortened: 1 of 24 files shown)


src/explore-education-statistics-admin/src/pages/methodology/external-methodology/ExternalMethodologyPage.tsx
```diff
@@ -26,7 +25,12 @@ const ExternalMethodologyPage = ({
   const { value: model, isLoading } = useAsyncHandledRetry<Model>(async () => {
     const [publication, externalMethodology] = await Promise.all([
       publicationService.getPublication(publicationId),
-      publicationService.getExternalMethodology(publicationId),
+      publicationService.getExternalMethodology(publicationId).catch(err => {
+        if (err.response.status !== 404) {
+          throw err;
+        }
+        return undefined;
+      }),
```
ntsim, line 33: Let's do this in the service method so we don't need to do this repeatedly all over the place. Would either return null (for a 404) or re-throw the error (as you have). 

We'll also need to change all the test mocks to use `mockResolvedValue` again instead, sorry!

```ts
publicationService.getExternalMethodology.mockResolvedValue(null);
```

ntsim (dismissed):

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
```diff
@@ -73,6 +75,30 @@ public async Task<ActionResult<List<PublicationSummaryViewModel>>> ListPublicati
...
+        {
+            return await _publicationService.UpdateExternalMethodology(publicationId, updatedExternalMethodology)
+                .HandleFailuresOrOk();
+        }
+
+        [HttpDelete("api/publication/{publicationId:guid}/external-methodology")]
+        public async Task<ActionResult<Unit>> RemoveExternalMethodology(
+            Guid publicationId)
+        {
+            return await _publicationService.RemoveExternalMethodology(publicationId)
```
duncan-at-hiveit, line 97: Would make this `HandleFailuresOrNoContent()` as it's a DELETE

src/GovUk.Education.ExploreEducationStatistics.Admin/Mappings/MappingProfiles.cs
```diff
@@ -6,12 +6,19 @@
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Methodologies;
 using GovUk.Education.ExploreEducationStatistics.Admin.ViewModels;
 using GovUk.Education.ExploreEducationStatistics.Admin.ViewModels.ManageContent;
-using GovUk.Education.ExploreEducationStatistics.Admin.ViewModels.Methodology;
 using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Content.Model;
+using GovUk.Education.ExploreEducationStatistics.Content.Services.ViewModels;
 using GovUk.Education.ExploreEducationStatistics.Publisher.Model;
+using ContactViewModel = GovUk.Education.ExploreEducationStatistics.Admin.ViewModels.ContactViewModel;
```
duncan-at-hiveit, line 13: Out of interest, where have all these `usings` popped up from?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -325,6 +324,48 @@ await supersededPublications
...
+        public async Task<Either<ActionResult, ExternalMethodologyViewModel>> UpdateExternalMethodology(
+            Guid publicationId, ExternalMethodology updatedExternalMethodology)
+        {
+            return await _persistenceHelper
+                .CheckEntityExists<Publication>(publicationId)
+                .OnSuccessDo(_userService.CheckCanUpdatePublication)
+                .OnSuccess(publication =>
+                {
+                    _context.Update(publication);
+                    publication.ExternalMethodology = updatedExternalMethodology;
```
duncan-at-hiveit, line 346: Note that given we've allowed a user of the API to potentially update ANY field of the ExternalMethodology here, that they could potentially do something nefarious here, like amend the JSON of the POST to alter fields of the ExternalMethodology that we wouldn't intend them to alter, and that the front end wouldn't let them alter. 

Safer (albeit with more boilerplate) would be to have the Controller endpont accept an `UpdateExternalMethodologyRequest` / `ExternalMethodologyUpdateRequest` that includes just the fields that we want a user to be able to update, then manually apply those field updates to either an existing or a new ExternalMethodology object on this Publication.

src/GovUk.Education.ExploreEducationStatistics.Admin/Mappings/MappingProfiles.cs
markhiveit: I think because there are duplicate view models in `Content.Services.ViewModels`, when I added `ExternalMethodologyViewModel` it having to explicitly declare all of these.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
```diff
@@ -73,6 +75,30 @@ public async Task<ActionResult<List<PublicationSummaryViewModel>>> ListPublicati
                 .HandleFailuresOrOk();
         }
 
+        [HttpGet("api/publication/{publicationId:guid}/external-methodology")]
+        public async Task<ActionResult<ExternalMethodologyViewModel>> GetExternalMethodology(Guid publicationId)
```
benoutram, line 79: This is a `Content.Services.ViewModels` view model, intended to be returned to the public. You've been able to import this after `Content.Services` was made a dependency of Admin for other fairly recent changes, but I think the intention still stands that Admin has its own view models.

By sharing view models with the public, it potentially exposes us to problems where a change to a common view model required for the Admin unknowingly exposes some private data to the public, or alters it in a way that breaks the public site.

It also opens us up to potential friction in the future when the view models do need to diverge.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Mappings/MappingProfiles.cs
markhiveit: This has highlighted an issue - we shouldn't be depending on Content.Services in Admin, so I've made a new `ExternalMethodologyViewModel` that lives in `Admin.ViewModels`.

markhiveit: I've also made a clone of `PublicationSummaryViewModel` into Admin so we can remove the dependency on `Content.Services.ViewModel` from `PublicationController` and `PublicationService` in Admin.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
markhiveit: Created `ExternalMethodologySaveRequest` record - basically used in the same vein as "PublicationSaveRequest`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -325,6 +319,50 @@ await supersededPublications
                 .OnSuccess(publication => _mapper.Map<PublicationViewModel>(publication));
         }
 
+        public async Task<Either<ActionResult, ExternalMethodologyViewModel>> GetExternalMethodology(Guid publicationId)
+        {
+            return await _persistenceHelper
+                .CheckEntityExists<Publication>(publicationId)
+                .OnSuccessDo(_userService.CheckCanViewPublication)
+                .OnSuccess(publication => publication.ExternalMethodology != null
+                    ? _mapper.Map<ExternalMethodologyViewModel>(publication.ExternalMethodology)
```
benoutram, line 328: You could not bother with AutoMapper here and use `new ExternalMethodologyViewModel(publication.ExternalMethodology)` since you've added  a copy constructor to `ExternalMethodologyViewModel` which copies the `ExternalMethodology`.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ExternalMethodologyViewModel.cs
```diff
@@ -0,0 +1,21 @@
...
+
+    public ExternalMethodologyViewModel()
+    {
+    }
+
+    public ExternalMethodologyViewModel(ExternalMethodology externalMethodology)
+    {
+        Title = externalMethodology.Title;
+        Url = externalMethodology.Url;
+    }
```
benoutram, line 20: This looks like it's been cloned from `Content.Services.ViewModels` but I don't think this copy constructor for `ExternalMethodology` is used anywhere by Admin?

See my other comment about how this could be used though 🙂 .

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
markhiveit: Also added a constructor for `ExternalMethodologyViewModel` that accepts a `ExternalMethodlogySaveRequest`, so I can remove that from the mapping config too.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ExternalMethodologyViewModel.cs
```diff
@@ -0,0 +1,28 @@
...
+
+    public ExternalMethodologyViewModel(ExternalMethodologySaveRequest externalMethodology)
+    {
+        Title = externalMethodology.Title;
+        Url = externalMethodology.Url;
+    }
```
benoutram, line 27: I don't think it really matters but `return new ExternalMethodologyViewModel(publication.ExternalMethodology)` after saving the updated publication would mean this isn't needed.

benoutram (approved):

markhiveit:
Pushed a fix for `UpdateExternalMethodology` and `RemoveExternalMethodology`. We weren't updating the public `publication.json` cache file after calling these methods. I'll leave a comment on a relevant ticket to write UI tests to catch this in the future

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -325,6 +319,56 @@ await supersededPublications
...
+                .OnSuccess(async publication =>
+                {
+                    _context.Update(publication);
+                    publication.ExternalMethodology ??= new ExternalMethodology();
+                    publication.ExternalMethodology.Title = updatedExternalMethodology.Title;
+                    publication.ExternalMethodology.Url = updatedExternalMethodology.Url;
+                    await _context.SaveChangesAsync();
+
+                    // Clear cache because ExternalMethodology is in Content.Services.ViewModels.PublicationViewModel
+                    await _publicationCacheService.UpdatePublication(publication.Slug);
```
benoutram, line 347: Comment could say it updates the cached publication because technically it doesn't clear/delete it anymore. Same on line 365.

benoutram (approved):

PR #3501, opened 2022-09-15
EES-3701 - fix skip to search results link

luke-h1:
Adds missing hash to link on step 1 of table tool page

### DIFF
--- a/src/explore-education-statistics-common/src/modules/table-tool/components/PublicationForm.tsx
+++ b/src/explore-education-statistics-common/src/modules/table-tool/components/PublicationForm.tsx
@@ -142,7 +142,7 @@ const PublicationForm = ({
                   {searchTerm && publications.length > 0 && (
                     <p>
                       <a
-                        href={`${formId}-publications`}
+                        href={`#${formId}-publications`}
                         className="govuk-!-margin-top-3 govuk-!-font-size-14"
                       >
                         Skip to search results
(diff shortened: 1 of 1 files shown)


src/explore-education-statistics-common/src/modules/table-tool/components/PublicationForm.tsx
```diff
@@ -142,7 +142,7 @@ const PublicationForm = ({
                   {searchTerm && publications.length > 0 && (
                     <p>
                       <a
-                        href={`${formId}-publications`}
+                        href={`#${formId}-publications`}
```
amyb-hiveit, line 145: Looks like the id it's linking to is wrong so it doesn't take you to the publications list. It's `publicationForm-publicationForm-publications`  - probably want to find what's adding the second `publicationForm` and remove it (checking that doesn't break anything else).

amyb-hiveit (changes requested):

src/explore-education-statistics-common/src/modules/table-tool/components/PublicationForm.tsx
luke-h1: hmm, interesting. afaik it looks like `FormFieldRadioGroup` was inheriting the ID from the `Form` component. I've removed the `${formId}` from the id and replaced it with `id="publications"` which seems to fix the duplication & link

amyb-hiveit (approved):

PR #3502, opened 2022-09-15
EES-3080 - hide destination/breadcrumbs from screen reader users in s…

luke-h1:
This PR: 
- adds `aria-hidden` to the breadcrumbs text on page search forms in order to address an accessibility issue. This improves the screen-reader experience due to the length of text that can be under a given theme

### DIFF
--- a/src/explore-education-statistics-common/src/components/PageSearchForm.tsx
+++ b/src/explore-education-statistics-common/src/components/PageSearchForm.tsx
@@ -212,7 +212,7 @@ const PageSearchForm = ({
                   <>
                     <div className={styles.resultHeader}>{result.text}</div>
                     {result.location && (
-                      <div className={styles.resultLocation}>
+                      <div className={styles.resultLocation} aria-hidden>
                         {result.location}
                       </div>
                     )}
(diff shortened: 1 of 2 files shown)


src/explore-education-statistics-common/src/components/PageSearchForm.tsx
```diff
@@ -212,7 +212,7 @@ const PageSearchForm = ({
                   <>
                     <div className={styles.resultHeader}>{result.text}</div>
                     {result.location && (
-                      <div className={styles.resultLocation}>
+                      <div className={styles.resultLocation} aria-hidden="true">
```
ntsim, line 215: Can simplify this to `aria-hidden` by itself

src/explore-education-statistics-common/src/components/form/FormComboBox.tsx
```diff
@@ -52,7 +52,6 @@ const FormComboBox = ({
   const listBoxRef = useRef<HTMLUListElement>(null);
   const inputRef = useRef<HTMLInputElement>(null);
   const optionRefs = useRef<Dictionary<HTMLLIElement>>({});
-
```
ntsim, line 55: Nitpick, but no need to remove the spacing here. The refs were logically spaced out from the state

ntsim (changes requested):

PR #3507, opened 2022-09-16
 upgrading python dependencies

N-moh:


Pipfile
```diff
@@ -9,17 +9,17 @@ docker = "==4.1.0"
 [packages]
 robotframework = "~=5.0.1"
 robotframework-seleniumlibrary = "~=6.0.0"
-requests = "~=2.27.1"
+requests = "==2.28.1"
 python-dotenv = "~=0.20.0"
-robotframework-pabot = "~=2.5.3"
+robotframework-pabot = "==2.6.0"
```
luke-h1, line 14: We will want to keep the `~=` syntax instead of `==`. This means that we follow semver so that pipenv locks the major version of the package.

luke-h1 (changes requested):

luke-h1:
Just one small comment & this should be good to go. In addition to this, I'd recommend renaming this branch to fit in with the naming convention we follow. This is to keep the history clean & also provide a ticket reference for changes that have gone in. The branch name would look something like this: `EES-3688`

PR #3509, opened 2022-09-20
EES-3719 Tidy up the public ReleaseController's use of cached resources

benoutram:
### Problem with cached releases

`ReleaseViewModel` was built from a `CachedReleaseViewModel` and `PublicationViewModel`, where `PublicationViewModel` is built with the list of methodology summaries.

Properties only in cached view model `CachedReleaseViewModel`:
```
string DataGuidance
string PreReleaseAccessList
```

and properties only in response view model `ReleaseViewModel`:
```
bool HasPreReleaseAccessList
bool HasDataGuidance
bool LatestRelease
PublicationViewModel Publication
```

After a cached release is retrieved, it's always transformed to something else, e.g. `ReleaseViewModel`.

### Problem with cached publications

`PublicationViewModel` was retrieved from blob cache but `Methodologies` was always be empty in the cache.
`Methodologies` is only populated later while constructing the `PublicationViewModel` from the cached version of it.

It's not obvious without looking that `Methodologies` will always be empty for a publication retrieved from the cache.

### Solution

Unless we want to return `DataGuidance` and `PreReleaseAccessList` with every request then it seems there's a legitimate need to continue to have two versions of the release view models.

To get rid of one of them we would also need to refactor `Publication` out of `ReleaseViewModel` by implementing a new endpoint in `PublicationController` to get the `PublicationViewModel` . However this would mean the FE making two separate requests for the release and publication when viewing a release content page and the overhead of an additional request could outweigh the benefits of the cleaner code.

To have the same separation for publications as we do for releases between what's cached and what is returned by the controller response, we introduce a new `PublicationCacheViewModel` _without_ methodologies.

For consistency with this new class we rename `CachedReleaseViewModel` -> `ReleaseCacheViewModel`.

To make `ReleaseCacheService` more reusable and not depend on other caching services, it's now only responsible for returning `ReleaseCacheViewModel`.

`ReleaseController` is made responsible for building view models from a mixture of cached resources using different cache services.

### Other changes

- Move filtering out the current release from the publication's list of releases from BE to FE. A consumer of the publication wouldn't expect any releases to be filtered. Prior to https://github.com/dfe-analytical-services/explore-education-statistics/pull/3504 that filtering happened when constructing `ReleaseViewModel` in the BE. It's now moved to be filtered by the FE at the closest point to where that filtering needs to happen in `PublicationReleasePage.tsx`.

### UI test report

image


src/GovUk.Education.ExploreEducationStatistics.Content.Services/ViewModels/PublicationCacheViewModel.cs
```diff
@@ -0,0 +1,29 @@
+#nullable enable
+using System;
+using System.Collections.Generic;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Services.ViewModels
+{
```
ntsim, line 6: Could change this to use file namespacing

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ViewModels/PublicationViewModel.cs
```diff
@@ -1,31 +1,50 @@
-#nullable enable
+﻿#nullable enable
 using System;
 using System.Collections.Generic;
 
-namespace GovUk.Education.ExploreEducationStatistics.Content.Services.ViewModels
+namespace GovUk.Education.ExploreEducationStatistics.Content.Services.ViewModels;
+
+public record PublicationViewModel
 {
```
ntsim, line 8: Could change this to use file namespacing

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Controllers/ReleaseController.cs
```diff
@@ -35,39 +43,71 @@ public async Task<ActionResult<List<ReleaseSummaryViewModel>>> ListReleases(stri
...
                 .HandleFailuresOrOk();
         }
+
+        private Task<Either<ActionResult, ReleaseViewModel>> GetReleaseViewModel(
+            string publicationSlug,
+            string? releaseSlug = null)
+        {
+            return _publicationCacheService.GetPublication(publicationSlug)
+                .OnSuccessCombineWith(publication => _methodologyCacheService.GetSummariesByPublication(publication.Id))
+                .OnSuccessCombineWith(_ => _releaseCacheService.GetRelease(publicationSlug, releaseSlug))
```
ntsim, line 86: Not saying that we need to do this, but did we consider optimising this to make parallel requests rather than performing these sequentially?

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Controllers/ReleaseController.cs
```diff
@@ -35,39 +43,71 @@ public async Task<ActionResult<List<ReleaseSummaryViewModel>>> ListReleases(stri
...
+                    );
+                });
+        }
+
+        private Task<Either<ActionResult, ReleaseSummaryViewModel>> GetReleaseSummaryViewModel(
+            string publicationSlug,
+            string? releaseSlug = null)
+        {
+            return _publicationCacheService.GetPublication(publicationSlug)
+                .OnSuccessCombineWith(_ => _releaseCacheService.GetRelease(publicationSlug, releaseSlug))
```
ntsim, line 102: Same as above - could perform these in parallel?

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ViewModels/PublicationCacheViewModel.cs
benoutram: Thanks, I've made this change.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ViewModels/PublicationViewModel.cs
benoutram: I think this one is already using a file scoped namespace.

ntsim: D'oh, of course! :smile:

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Controllers/ReleaseController.cs
benoutram: I don't think we can easily do this for two reasons.

1. [EES-3516](https://dfedigital.atlassian.net/browse/EES-3516), mentions that `CacheAspect` is forcing all `CacheAttribute` annotated methods to run synchronously.
2. EF Core doesn't support multiple parallel operations being run on the same context instance. This would be a problem if two or more methods got cache-misses and run the underlying methods which do database work. We need to await each one before starting the next.

I commented out the `MemoryCache` annotation on the method calling this one to make sure this block always runs, and then did some investigations.

I also ignores the third task getting methodologies for simplicity and just work with two tasks.

### Notes on (1):

A task getting a result from a cache is run  synchronously by
https://github.com/dfe-analytical-services/explore-education-statistics/blob/dev/src/GovUk.Education.ExploreEducationStatistics.Common/Cache/CacheAspect.cs#L81 :

`var triggerResult = cacheTrigger.Get(cacheKey, cachedReturnType).Result;`

If there's a cache-miss, the task getting the underlying result is run synchronously by https://github.com/dfe-analytical-services/explore-education-statistics/blob/dev/src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/ReflectionExtensions.cs#L120 :

`task.Wait()`

So if I start two tasks as follows without awaiting them, we will see their status is always already complete and we get the log message "Tasks already ran to completion":

```
var publicationTask = _publicationCacheService.GetPublication(publicationSlug);
var releaseTask = _releaseCacheService.GetRelease(publicationSlug, releaseSlug);

if (publicationTask.Status == TaskStatus.RanToCompletion
    && releaseTask.Status == TaskStatus.RanToCompletion)
{
    _logger.LogInformation("Tasks already ran to completion");
}
```

We see the same result whether there's a cache-miss or not. The cache-misses are reproduced by clearing the cache before running. 

### Notes on (2):

We can force parallel running by commenting out the `BlobCache` annotation on `GetPublication` and `GetRelease` in the cache services.

```
    //[BlobCache(typeof(PublicationCacheKey), ServiceName = "public")]
    public Task<Either<ActionResult, PublicationCacheViewModel>> GetPublication(string publicationSlug)

    //[BlobCache(typeof(ReleaseCacheKey), ServiceName = "public")]
    public Task<Either<ActionResult, ReleaseCacheViewModel>> GetRelease(string publicationSlug, string? releaseSlug = null)
```

We then run into the database context blocking issue.

Here's the code running them in parallel:

```
var publicationTask = _publicationCacheService.GetPublication(publicationSlug);
var releaseTask = _releaseCacheService.GetRelease(publicationSlug, releaseSlug);

if (publicationTask.Status == TaskStatus.RanToCompletion
    && releaseTask.Status == TaskStatus.RanToCompletion)
{
    _logger.LogInformation("Tasks already ran to completion");
    return new ReleaseViewModel(
        releaseTask.Result.Right,
        new PublicationViewModel(publicationTask.Result.Right,
            new List<MethodologyVersionSummaryViewModel>())
    );
}

// Tasks are running in parallel
_logger.LogInformation("Running async");

var publication = await publicationTask;
var release = await releaseTask;

return new ReleaseViewModel(
    release.Right,
    new PublicationViewModel(publication.Right, new List<MethodologyVersionSummaryViewModel>())
);
```

and the error

```
System.InvalidOperationException: A second operation was started on this context instance before a previous operation completed. This is usually caused by different threads concurrently using the same instance of DbContext. For more information on how to avoid threading issues with DbContext, see https://go.microsoft.com/fwlink/?linkid=2097913.
```

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Controllers/ReleaseController.cs
benoutram: See reply to above. 🙂

benoutram:
Running UI tests found an issue caused by removing filtering of the current release from `Release.Publication.Releases` of the `ReleaseViewModel`. This was fixed in an additional commit https://github.com/dfe-analytical-services/explore-education-statistics/pull/3509/commits/fb2209709375abf397f2330b0467f7b2ec2e1c90.

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Controllers/ReleaseController.cs
ntsim: Ah bad times. Surprising behaviour that the AspectInjector wrapped tasks seem to run to completion immediately!

Oh well, happy to leave things as-is!

ntsim (approved):

PR #3512, opened 2022-09-20
EES-3052 - add more context to release headers

luke-h1:
This PR: 
* fixes an accessibility issue whereby the 'last updated' & 'see other releases' headings weren't providing enough context to screen-readers

TODO: 
* update UI tests

### DIFF
--- a/src/explore-education-statistics-frontend/src/modules/find-statistics/PublicationReleasePage.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/find-statistics/PublicationReleasePage.tsx
@@ -125,6 +125,7 @@ const PublicationReleasePage: NextPage<Props> = ({ release }) => {
                 <Details
                   id="releaseLastUpdates"
                   summary={`See all updates (${updates.length})`}
+                  hiddenText={`for ${release.title}`}
                   onToggle={open => {
                     if (open) {
                       logEvent({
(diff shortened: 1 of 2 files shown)


src/explore-education-statistics-frontend/src/modules/find-statistics/PublicationReleasePage.tsx
```diff
@@ -125,6 +125,7 @@ const PublicationReleasePage: NextPage<Props> = ({ release }) => {
                 <Details
                   id="releaseLastUpdates"
                   summary={`See all updates (${updates.length})`}
+                  hiddenText={`for ${release.publication.title}`}
```
ntsim, line 128: Updates aren't for a publication, they're for a release. We should change this to `for ${release.title}`.

src/explore-education-statistics-frontend/src/modules/find-statistics/__tests__/PublicationReleasePage.test.tsx
```diff
@@ -243,7 +243,9 @@ describe('PublicationReleasePage', () => {
     );
 
     userEvent.click(
-      screen.getByRole('button', { name: 'See all updates (2)' }),
+      screen.getByRole('button', {
+        name: `See all updates (2) for ${testRelease.publication.title}`,
```
ntsim, line 247: Would hardcode the expected string rather than use a variable. If this variable isn't set correctly, then the test would still pass.

ntsim (changes requested):

ntsim (approved):
Thought these changes would break the Robot tests, but turns out we're using a more flexible selector for details text that doesn't break with the new hidden text :smile:

PR #3513, opened 2022-09-22
EES-3127 Remove backwards compatibility with cached releases that use `ReleaseTypeViewModel`.

benoutram:
⚠️ Depends on https://github.com/dfe-analytical-services/explore-education-statistics/pull/3509
⚠️ Requires a content cache refresh

This PR removes backwards compatibility with publicly cached releases that have a release `Type` of type `ReleaseTypeViewModel`.

`Release.Type` was altered to be the enum `ReleaseType` by https://github.com/dfe-analytical-services/explore-education-statistics/pull/3119 but the cached model continued to use `ReleaseTypeViewModel` even for new releases. Code was written to interchange between the two types when building `ReleaseViewModel` from the cached view model.

This was done to avoid requiring refreshing the publicly cached releases at the time.

EES-3650 has made refreshing publicly cached releases much easier so we can now introduce this change and remove the workaround.


markhiveit (approved):

markhiveit (approved):

benoutram:
Updated UI test report after `EnumToStringConverter` change

image

PR #3515, opened 2022-09-23
EES-3695 Add GetContact and UpdateContact endpoints

markhiveit:
This PR adds new endpoints to get and update contacts on a publication, and also does other various bits of cleanup in preparation for switching to the new Admin dashboard:

- Adds a GetContact endpoint. This endpoint has an optional query parameter so services can request permissions if required.
- Adds a UpdateContact endpoint.
- Make the Publications table's `ContactId` nonnullable. Since a publication should always have a contact, there isn't any reason for the column to be nullable. :rotating_light: :rotating_light: :rotating_light: Due to this change, we need to double check there aren't any instances of `ContactId` that are null on Prod before deploying :rotating_light: :rotating_light: :rotating_light: I already updated all entries with null on Dev, and Test, Preprod and Prod didn't have any instances of null.
- Create `PublicationCreateRequest` `PublicationCreateViewModel`. This is necessary as updating the contact is no longer possible via the update publication endpoint, but when creating a publication you must also create an associated contact, as every publication must have a contact, so it makes sense for a contact to be in the create publication response.
- Remove `Contact` from `PublicationViewModel`. This means that several frontend components need to make additional get and update calls to get / update contact information.
- The `GetPublication` endpoint now hydrates all parts of the `PublicationViewModel` that it returns, specifically `Permissions` and `IsSuperseded`.
- Add a query parameter for the `GetPublication` endpoint so services can choose whether they want `Permissions` returned.
- Change the frontend's `PublicationContext` to use `Publication` instead of `MyPublication`. This change has meant that some components have had to make an additional call to get other information, like publication's contact information.
- Several frontend interfaces have been renamed to match their associated backend view models.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
@@ -64,10 +64,10 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api
 
         [HttpGet("api/publications/{publicationId}")]
         public async Task<ActionResult<PublicationViewModel>> GetPublication(
-            [Required] Guid publicationId)
+            [Required] Guid publicationId, [FromQuery] bool permissions = false)
         {
             return await _publicationService
-                .GetPublication(publicationId)
+                .GetPublication(publicationId, permissions)
                 .HandleFailuresOrOk();
         }
 
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
@@ -125,7 +125,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api
         }
 
         [HttpPost("api/publications")]
-        public async Task<ActionResult<PublicationViewModel>> CreatePublication(
+        public async Task<ActionResult<PublicationCreateViewModel>> CreatePublication(
             PublicationCreateRequest publication)
         {
             return await _publicationService
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
@@ -124,7 +124,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api
 
         [HttpPost("api/publications")]
         public async Task<ActionResult<PublicationViewModel>> CreatePublication(
-            PublicationSaveRequest publication)
+            PublicationCreateRequest publication)
         {
             return await _publicationService
                 .CreatePublication(publication)
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
@@ -96,10 +96,9 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api
 
         [HttpGet("api/publication/{publicationId:guid}/contact")]
         public async Task<ActionResult<ContactViewModel>> GetContact(
-            Guid publicationId,
-            [FromQuery] bool includePermissions = false)
+            Guid publicationId)
         {
-            return await _publicationService.GetContact(publicationId, includePermissions)
+            return await _publicationService.GetContact(publicationId)
                 .HandleFailuresOrOk();
         }
 
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
@@ -94,6 +94,13 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api
                 .HandleFailuresOrNoContent();
         }
 
+        [HttpGet("api/publication/{publicationId:guid}/contact")]
+        public async Task<ActionResult<ContactViewModel>> GetContact(Guid publicationId)
+        {
+            return await _publicationService.GetContact(publicationId)
+                .HandleFailuresOrOk();
+        }
+
 
         [HttpGet("api/publication/{publicationId}/releases")]
         public async Task<ActionResult<PaginatedListViewModel<ReleaseSummaryViewModel>>> ListActiveReleases(
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
@@ -101,6 +101,13 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api
                 .HandleFailuresOrOk();
         }
 
+        [HttpPut("api/publication/{publicationId:guid}/contact")]
+        public async Task<ActionResult<ContactViewModel>> UpdateContact(
+            Guid publicationId, Contact updatedContact)
+        {
+            return await _publicationService.UpdateContact(publicationId, updatedContact)
+                .HandleFailuresOrOk();
+        }
 
         [HttpGet("api/publication/{publicationId}/releases")]
         public async Task<ActionResult<PaginatedListViewModel<ReleaseSummaryViewModel>>> ListActiveReleases(
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
@@ -28,11 +28,11 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api
 
         [HttpGet("api/publications")]
         public async Task<ActionResult<List<PublicationViewModel>>> ListPublications(
-            [FromQuery] bool permissions,
+            [FromQuery] bool includePermissions,
             [FromQuery] Guid? topicId)
         {
             return await _publicationService
-                .ListPublications(permissions, topicId)
+                .ListPublications(includePermissions, topicId)
                 .HandleFailuresOrOk();
         }
 
@@ -64,10 +64,10 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api
 
         [HttpGet("api/publications/{publicationId}")]
         public async Task<ActionResult<PublicationViewModel>> GetPublication(
-            [Required] Guid publicationId, [FromQuery] bool permissions = false)
+            [Required] Guid publicationId, [FromQuery] bool includePermissions = false)
         {
             return await _publicationService
-                .GetPublication(publicationId, permissions)
+                .GetPublication(publicationId, includePermissions)
                 .HandleFailuresOrOk();
         }
 
@@ -97,9 +97,9 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api
         [HttpGet("api/publication/{publicationId:guid}/contact")]
         public async Task<ActionResult<ContactViewModel>> GetContact(
             Guid publicationId,
-            [FromQuery] bool permissions = false)
+            [FromQuery] bool includePermissions = false)
         {
-            return await _publicationService.GetContact(publicationId, permissions)
+            return await _publicationService.GetContact(publicationId, includePermissions)
                 .HandleFailuresOrOk();
         }
 
@@ -116,11 +116,11 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api
             [Required] Guid publicationId,
             [FromQuery, Range(1, double.PositiveInfinity)] int page = 1,
             [FromQuery, Range(0, double.PositiveInfinity)] int pageSize = 5,
-            [FromQuery] bool permissions = false,
-            [FromQuery] bool? live = null)
+            [FromQuery] bool? live = null,
+            [FromQuery] bool includePermissions = false)
         {
             return await _publicationService
-                .ListActiveReleasesPaginated(publicationId, page, pageSize, live, includePermissions: permissions)
+                .ListActiveReleasesPaginated(publicationId, page, pageSize, live, includePermissions)
                 .HandleFailuresOrOk();
         }
 
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
@@ -95,9 +95,11 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api
         }
 
         [HttpGet("api/publication/{publicationId:guid}/contact")]
-        public async Task<ActionResult<ContactViewModel>> GetContact(Guid publicationId)
+        public async Task<ActionResult<ContactViewModel>> GetContact(
+            Guid publicationId,
+            [FromQuery] bool permissions = false)
         {
-            return await _publicationService.GetContact(publicationId)
+            return await _publicationService.GetContact(publicationId, permissions)
                 .HandleFailuresOrOk();
         }
 
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20220914155611_EES3695MakePublicationsContactIdNonnullable.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20220914155611_EES3695MakePublicationsContactIdNonnullable.cs
@@ -1,0 +1,57 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Migrations.ContentMigrations
+{
+    public partial class EES3695MakePublicationsContactIdNonnullable : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropForeignKey(
+                name: "FK_Publications_Contacts_ContactId",
+                table: "Publications");
+
+            migrationBuilder.AlterColumn<Guid>(
+                name: "ContactId",
+                table: "Publications",
+                type: "uniqueidentifier",
+                nullable: false,
+                defaultValue: new Guid("00000000-0000-0000-0000-000000000000"),
+                oldClrType: typeof(Guid),
+                oldType: "uniqueidentifier",
+                oldNullable: true);
+
+            migrationBuilder.AddForeignKey(
+                name: "FK_Publications_Contacts_ContactId",
+                table: "Publications",
+                column: "ContactId",
+                principalTable: "Contacts",
+                principalColumn: "Id",
+                onDelete: ReferentialAction.Cascade);
+        }
+
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropForeignKey(
+                name: "FK_Publications_Contacts_ContactId",
+                table: "Publications");
+
+            migrationBuilder.AlterColumn<Guid>(
+                name: "ContactId",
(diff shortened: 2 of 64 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
```diff
@@ -64,10 +64,10 @@ public async Task<ActionResult<List<PublicationSummaryViewModel>>> ListPublicati
 
         [HttpGet("api/publications/{publicationId}")]
         public async Task<ActionResult<PublicationViewModel>> GetPublication(
-            [Required] Guid publicationId)
+            [Required] Guid publicationId, [FromQuery] bool permissions = false)
```
duncan-at-hiveit, line 67: At this point in reading through the PR, it's unclear what the purpose of the "permissions" flag is here.

Presumably it'll decide whether or not to include a user's permissions with the returned object?  If so, I'd be tempted to rename this to `includePermissions` or `includeUserPermissions`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20220914155611_EES3695MakePublicationsContactIdNonnullable.cs
```diff
@@ -0,0 +1,57 @@
...
+                oldType: "uniqueidentifier",
+                oldNullable: true);
+
+            migrationBuilder.AddForeignKey(
+                name: "FK_Publications_Contacts_ContactId",
+                table: "Publications",
+                column: "ContactId",
+                principalTable: "Contacts",
+                principalColumn: "Id",
+                onDelete: ReferentialAction.Cascade);
```
duncan-at-hiveit, line 32: Not 100% sure on this, but I think this is telling EF that if a Contact record is deleted, it'd then try to cascade that and delete its owning Publication too.  Worth checking with other devs?

duncan-at-hiveit: Assuming that's so, would suggest changing it to ReferentialAction.Restrict.  This'll mean EF would not allow the deletion of a Contact that is in use by a Publication.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -369,6 +357,54 @@ await supersededPublications
...
+                    }
+
+                    publication.Contact.ContactName = updatedContact.ContactName;
+                    publication.Contact.ContactTelNo = updatedContact.ContactTelNo;
+                    publication.Contact.TeamName = updatedContact.TeamName;
+                    publication.Contact.TeamEmail = updatedContact.TeamEmail;
+                    await _context.SaveChangesAsync();
+
+                    // Clear cache because Contact is in Content.Services.ViewModels.PublicationViewModel
+                    await _publicationCacheService.UpdatePublication(publication.Slug);
```
duncan-at-hiveit, line 402: Good spot!

src/explore-education-statistics-admin/src/services/publicationService.ts
```diff
@@ -47,28 +51,36 @@ export interface MyPublication {
...
 export interface Publication {
   id: string;
   title: string;
   summary: string;
   slug: string;
-  contact: PublicationContactDetails;
   theme: IdTitlePair;
   topic: IdTitlePair;
+  supersededById?: string;
+  isSuperseded?: boolean;
```
duncan-at-hiveit, line 82: Could this just be inferred from the existence of a non-null `supersededById`?  Not a biggie, probably makes it more readable...

src/explore-education-statistics-admin/src/services/publicationService.ts
```diff
@@ -117,8 +138,13 @@ const publicationService = {
     return client.put(`/publications/${publicationId}`, publication);
   },
 
-  getPublication(publicationId: string): Promise<Publication> {
-    return client.get<Publication>(`/publications/${publicationId}`);
+  getPublication(
+    publicationId: string,
+    permissions = false,
```
duncan-at-hiveit, line 143: Again, I'd suggest making the name of this variable a bit more explanatory - `includePermissions` or summat.

src/explore-education-statistics-admin/src/pages/publication/PublicationDetailsPage.tsx
```diff
@@ -15,39 +16,29 @@ import React from 'react';
...
-      theme,
-      topic,
-      supersedingPublication,
-    };
-  });
-
-  const { theme, topic, supersedingPublication } = value ?? {};
+      return publicationService.getPublication(supersededById);
+    },
+  );
```
duncan-at-hiveit, line 37: Does `supersededById` need to be added to a dependency array here in `useAsyncHandledRetry`?  Not 100% sure if it needs it in the same way that useEffect() needs it, but would suggest just looking to see if there are any warnings about this in the console when on this page.

src/explore-education-statistics-admin/src/services/publicationService.ts
markhiveit: Unfortunately not. For `isSuperseded` to be true, the `supersededById` publication must also have a live release.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
ntsim: Disagree somewhat here. We're already using this flag for a bunch of other endpoints, so this would introduce inconsistency at this point. Not sure it's worth the effort to change this.

Personally, don't think it's particularly unclear what the purpose is and it's not super hard to follow it through to figure it out. This said, could be convinced that `includePermissions` is better, but `includeUserPermissions` seems like overkill.

markhiveit: Discussed this on Slack. I've switched to using `includePermission` and changed it for all endpoints.

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20220914155611_EES3695MakePublicationsContactIdNonnullable.cs
markhiveit: I haven't tested it, but I think you're right. I've changed it to `Restrict`.

markhiveit: Ok Ben said that this actually means if a Publication gets deleted, the associated Contact will be deleted. We think this is fine so I've changed this back to Cascade. I will test this to make sure it's right

src/explore-education-statistics-admin/src/pages/publication/PublicationDetailsPage.tsx
markhiveit: I've added it to the deps - other places we use `useAsyncHandledRetry` do include variables used in side in the deps.

src/explore-education-statistics-admin/src/pages/publication/PublicationContactPage.tsx
```diff
@@ -11,54 +10,29 @@ import SummaryListItem from '@common/components/SummaryListItem';
...
-      contact: {
-        teamName,
-        teamEmail,
-        contactName,
-        contactTelNo,
-      },
-    });
+    await publicationService.updateContact(publicationId, updatedContact);
 
     onReload();
```
ntsim, line 29: Don't need this anymore as contact isn't on publication. We actually need to update the `contact` and switch back to the read-only mode during the submit by doing something like this:

```ts
const { value: contact, setValue: setContact } = useAsyncHandledRetry(...);

const handleSubmit = async (updatedContact: PublicationContactFormValues) => {
  ...

  const nextContact = await publicationService.updateContact(publicationId, updatedContact);

  setContact(nextContact);
  toggleReadOnly.off();
};
```

It also seems like we're missing a test case to check that the UI switches back to the read-only state upon successful submit as well.

src/explore-education-statistics-admin/src/pages/publication/__tests__/PublicationContactPage.test.tsx
```diff
@@ -244,23 +260,15 @@ describe('PublicationContactPage', () => {
...
-        {
-          ...testPublication,
-          contact: {
-            contactName: 'John Smith',
-            contactTelNo: '0777777777',
-            teamEmail: 'john.smith@test.com',
-            teamName: 'Team Smith',
-          },
-        },
+        testContact,
```
ntsim, line 265: Can appreciate you didn't write this test originally, but it might be a good basic check to change one of the contact form's fields to check that the updated value gets passed through correctly.

src/explore-education-statistics-admin/src/services/publicationService.ts
```diff
@@ -117,8 +138,13 @@ const publicationService = {
     return client.put(`/publications/${publicationId}`, publication);
   },
 
-  getPublication(publicationId: string): Promise<Publication> {
-    return client.get<Publication>(`/publications/${publicationId}`);
+  getPublication(
+    publicationId: string,
+    includePermissions = false,
+  ): Promise<Publication> {
```
ntsim, line 144: To make this a little more flexible so that we don't need to handle optional permissions in the places where we call this, would suggest we add a generic type for this (similar to the `listReleases` method). It should look something like:
```ts
listReleases<TPublication extends Publication = Publication>(...): Promise<TPublication>
```

Then at call sites, we can do this:
```ts
getPublication<PublicationWithPermissions>(publicationId, true); 
```

Where the return type will now be a `PublicationWithPermissions` rather than a `Publication`

src/explore-education-statistics-admin/src/pages/publication/contexts/PublicationContext.tsx
```diff
@@ -1,10 +1,10 @@
-import { MyPublication } from '@admin/services/publicationService';
+import { Publication } from '@admin/services/publicationService';
 import React, { createContext, ReactNode, useContext, useMemo } from 'react';
 
 export interface PublicationContextState {
-  publication: MyPublication;
+  publication: Publication;
```
ntsim, line 5: All of the types in here should be a `PublicationWithPermissions`

src/explore-education-statistics-admin/src/pages/publication/PublicationPageContainer.tsx
```diff
@@ -55,7 +55,7 @@ const PublicationPageContainer = ({
     isLoading: loadingPublication,
     retry: reloadPublication,
   } = useAsyncHandledRetry(() =>
-    publicationService.getMyPublication(publicationId),
+    publicationService.getPublication(publicationId, true),
```
ntsim, line 58: As mentioned in `publicationService.ts`, we should add a generic type parameter so that we can call this like:
```ts
publicationService.getPublication<PublicationWithPermissions>(publicationId, true)
```

src/explore-education-statistics-admin/src/pages/publication/PublicationDetailsPage.tsx
```diff
@@ -15,39 +16,30 @@ import React from 'react';
...
-    themeId,
     title,
     summary,
-    topicId,
-  } = publication;
+    permissions,
+    supersededById,
+    theme,
+    topic,
+  } = publication as Publication;
```
ntsim, line 26: We shouldn't need this `as Publication`

src/explore-education-statistics-admin/src/pages/publication/PublicationDetailsPage.tsx
```diff
@@ -70,7 +62,7 @@ const PublicationDetailsPage = () => {
                 : 'This publication is not archived'}
             </SummaryListItem>
           </SummaryList>
-          {permissions.canUpdatePublication && (
+          {permissions?.canUpdatePublication && (
```
ntsim, line 65: As mentioned in `PublicationContact`, we should be working with a `PublicationWithPermissions` so none of the `permissions` should be optional. Same for below this as well.

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ContactViewModels.cs
```diff
@@ -1,34 +1,34 @@
...
 
-        public string ContactName { get; set; }
+        public string ContactTelNo { get; set; } = string.Empty;
 
-        public string ContactTelNo { get; set; }
+        public ContactPermissions? Permissions { get; set; }
+
+        public record ContactPermissions
+        {
+            public bool CanUpdatePublication { get; set; }
```
ntsim, line 20: Why do we have this on a contact? Shouldn't we have a permission on publication permissions called `CanUpdateContact`?

src/explore-education-statistics-admin/src/pages/publication/PublicationMethodologiesPage.tsx
```diff
@@ -99,7 +99,7 @@ const PublicationMethodologiesPage = () => {
             to an external file that contains methodology details.
           </p>
 
-          {permissions.canCreateMethodologies && (
+          {permissions?.canCreateMethodologies && (
```
ntsim, line 102: Again, these shouldn't optional - we should be working with `PublicationWithPermissions` from `PublicationContext`

src/explore-education-statistics-admin/src/pages/publication/PublicationReleasesPage.tsx
```diff
@@ -19,7 +19,7 @@ const PublicationReleasesPage = () => {
 
       <p>View, edit or amend releases contained within this publication.</p>
 
-      {publication?.permissions.canCreateReleases && (
+      {publication?.permissions?.canCreateReleases && (
```
ntsim, line 22: Again, these shouldn't optional - we should be working with `PublicationWithPermissions` from `PublicationContext`

src/explore-education-statistics-admin/src/pages/publication/PublicationContactPage.tsx
```diff
@@ -88,7 +62,7 @@ const PublicationContactPage = () => {
               {contact.contactTelNo}
             </SummaryListItem>
           </SummaryList>
-          {publication.permissions.canUpdatePublication && (
+          {contact.permissions?.canUpdatePublication && (
```
ntsim, line 65: As mentioned in another comment, the permission for updating a contact should be on the `publication` itself i.e. we should have something like `publication.permissions.canUpdateContact`.

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20220914155611_EES3695MakePublicationsContactIdNonnullable.cs
duncan-at-hiveit: We've had a discussion on Slack - @markhiveit to test this out to be sure of the correct way round.

src/explore-education-statistics-admin/src/services/publicationService.ts
duncan-at-hiveit: I might be tempted to rename `isSuperseded` to something more explicit then, like `isSupersededByLivePublication` or put a code comment next to `isSuperseded` to explain the situation - cos for me personally I'd be confused in a scenario whereby `isSuperseded` is false but `supersededById` is not null :)

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20220914155611_EES3695MakePublicationsContactIdNonnullable.cs
benoutram: Ignore what I said now, I was thinking this FK was being added to the contacts table, but it's a FK on the publications table.

So it does want to be `onDelete: ReferentialAction.Restrict` but we are left with a potential problem of leaving orphaned contacts if publications are deleted.

markhiveit: Tested this.

`DELETE  FROM Contacts WHERE Id = '35424c89-7ee1-4951-b1e8-f786c69e137c';` also deleted the publication
`DELETE FROM Publications WHERE ContactId = '35424c89-7ee1-4951-b1e8-f786c69e137c';` didn't result in the Contact being deleted.

I'll change the ReferentialAction to Restrict.

markhiveit: Switched to Restrict and reset the test data locally to reapply the migration. Tried to delete a contact and it wouldn't let me while the associated publication still existed.

src/explore-education-statistics-admin/src/services/publicationService.ts
markhiveit: I've added a comment in both the backend and frontend that should make this clearer. With hindsight, I think I would have named it `isSupersededByLivePublication`, as that clears up the ambiguity. But I'm inclined not to rename it now as that means invalidating all publications' publication.json. And since this work is likely to be deployed alongside the read replica work, we probably want as few things to worry about in that deploy as possible.

But still happy to rename it if you'd prefer - or I could create a ticket to rename it at a later date.

ntsim: Just to chime in here, but a more terse suggestion would be `isSupersededByLive`

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ContactViewModels.cs
markhiveit: I've just changed the name to `CanUpdateContact` for now - it still uses `userService.CheckCanUpdatePublication`. The same work will also need to be done with external methodology, legacy releases, etc., so I intend to sort out all the permission changes in my current branch. I've left a comment in `PermissionsUtils` to remind me to do this for contacts.

src/explore-education-statistics-admin/src/pages/publication/PublicationContactPage.tsx
markhiveit: I've also changed `getContact` and added a `ContactWithPermissions` interface so I can remove the `?` here.

markhiveit: Turns out I cannot do that because `UpdateContact` just returns a `Contact` and we pass the result of `UpdateContact` to `setContact` that was expecting a `ContactWithPermissions`. I've left the extra code around `getContact` and the new `ContactWithPermissions` interface.

ntsim: Would really just move the contact permission onto the publication as part of this PR to save yourself the trouble of having to do it later!

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ContactViewModels.cs
ntsim: Would suggest just doing it as part of this PR. Like I'm not sure what time we're saving here by not doing it here?

markhiveit: Discussed this with Nick on Slack - I'll make the changes on this PR.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Util/PermissionsUtils.cs
```diff
@@ -37,6 +38,7 @@ public static async Task<ReleasePermissions> GetReleasePermissions(IUserService
             CanCreateMethodologies = await userService.CheckCanCreateMethodologyForPublication(publication).IsRight(),
             CanManageExternalMethodology =
                 await userService.CheckCanManageExternalMethodologyForPublication(publication).IsRight(),
+            CanUpdateContact = await userService.CheckCanUpdatePublication(publication).IsRight(),
```
ntsim, line 41: We already check `userService.CheckCanUpdatePublication` above, so we should consolidate this into one call to avoid an additional database query.

ntsim (changes requested):

ntsim (approved):

PR #3516, opened 2022-09-23
EES-3735 - fix keyStat hiddenText always evaluating to undefined

luke-h1:
This PR: 
* fixes a bug whereby the `hiddenText` variable for keyStat tiles is always undefined due to `summary?.dataDefinitionTitle[0]` being set to `Help` by default. 

 
* Previously we were checking if `summary?.dataDefinitionTitle[0]` was nullable and then setting `hiddenText` if so. The guidance text for a keyStat tile is never nullable and by default is set to the string: `Help`. This caused the previous ternary op's condition to always evaluate to `undefined` & thus visually hidden text was never set on keyStat tiles with the default text of `Help`

### DIFF
--- a/src/explore-education-statistics-admin/src/components/editable/EditableKeyStat.tsx
+++ b/src/explore-education-statistics-admin/src/components/editable/EditableKeyStat.tsx
@@ -141,6 +141,8 @@ const EditableKeyStat = ({
       );
     }
 
+    const dataDefinitionTitle = summary?.dataDefinitionTitle[0] || 'Help';
+
     return (
       <>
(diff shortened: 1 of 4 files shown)


duncan-at-hiveit (changes requested):
If "The guidance text for a keyStat tile is never nullable", then we don't need to check it being nullable.  We can instead just check that it is "Help" or not.

Looking into it more, looks like it's not mandatory to have a non-empty value, so this logic should work for empty / undefined and the case where it's still "Help".

amyb-hiveit (approved):

PR #3518, opened 2022-09-26
EES-3054 & 3055 screen reader messages for shareable links

amyb-hiveit:
- Announces to screen reader users when a shareable link is generated (EES-3054) and when it is copied to the clipboard (EES-3055).
- When the link is generated I've included the instructions for it in the screen reader announcement so that it's clear what to do and aria-hidden the displayed version of the same message.
- I've added a reusable component for screen reader messages and updated table tool reordering to use this as well.

### DIFF
--- a/src/explore-education-statistics-common/src/components/ScreenReaderMessage.tsx
+++ b/src/explore-education-statistics-common/src/components/ScreenReaderMessage.tsx
@@ -1,0 +1,30 @@
+import React, { useEffect, useState } from 'react';
+
+/**
+ * Announces changes to screen readers.
+ * Use when just using aria-live on an existing element won't work
+ * because it announces changes multiple times.
+ */
+export default function ScreenReaderMessage({ message }: { message: string }) {
+  const [screenReaderMessage, setScreenReaderMessage] = useState('');
+
+  useEffect(() => {
+    // Clear the message then repopulate to ensure the new message is read,
+    // and only read once.
+    setScreenReaderMessage('');
+    setTimeout(() => {
+      setScreenReaderMessage(message);
+    }, 200);
+  }, [message]);
+
+  return (
+    <div
+      aria-live="assertive"
+      aria-atomic="true"
+      aria-relevant="additions"
+      className="govuk-visually-hidden"
+    >
+      {screenReaderMessage}
+    </div>
+  );
+}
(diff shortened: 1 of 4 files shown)


src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
```diff
@@ -57,54 +64,55 @@ const TableToolShare = ({
...
-          Copy link
-        </Button>
-
-        <ButtonLink to={`/data-tables/permalink/${permalinkId}`}>
-          View share link
-        </ButtonLink>
-      </ButtonGroup>
-    </div>
+    <>
+      {!permalinkId ? (
```
ntsim, line 72: Why use a ternary? It seemed cleaner previously with the early return, plus we've ended up making loads of changes to the markup as a result

src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
```diff
@@ -13,6 +14,9 @@ import {
 import ButtonLink from '@frontend/components/ButtonLink';
 import React, { useEffect, useState } from 'react';
 
+const linkInstructions =
+  ' Use the link below to see a version of this page that you can bookmark for future reference, or copy the link to send on to somebody else to view.';
```
ntsim, line 18: There's a erroneous space at the start of this sentence

src/explore-education-statistics-common/src/components/ScreenReaderMessage.tsx
```diff
@@ -0,0 +1,30 @@
...
+export default function ScreenReaderMessage({ message }: { message: string }) {
+  const [screenReaderMessage, setScreenReaderMessage] = useState('');
+
+  useEffect(() => {
+    // Clear the message then repopulate to ensure the new message is read,
+    // and only read once.
+    setScreenReaderMessage('');
+    setTimeout(() => {
+      setScreenReaderMessage(message);
+    }, 200);
```
ntsim, line 17: Might want to write a little test suite to check that this stuff happens?

ntsim (changes requested):

src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
amyb-hiveit: It's because the screen reader message div needs to always be present for the message to be announced as it's looking for changes in the aria-live area.

ntsim: Oh I see! I thought the screen reader message was in the lower branch of the ternary! My bad :sweat_smile:

src/explore-education-statistics-common/src/components/ScreenReaderMessage.tsx
amyb-hiveit: Good point, I've added tests now.

PR #3520, opened 2022-09-28
EES-3083 remove tabs when only one

amyb-hiveit:
Removes tabs when there is only one pane of content as this can be confusing to screen reader users. 

This affects:
- Key stats when there are no secondary stats
- Data blocks when there's just a table
- Pre-release access page on amendments
- Data blocks page in the admin when creating them or viewing a published release - I've added a 'Data source' heading to the content here as its not clear otherwise what you're doing.

Tabbed content with more than one tab should be unaffected.

### Screenshots

notabs2
notabs1


ntsim (approved):
Nice, it's good to finally get this sorted out! :+1:

PR #3528, opened 2022-10-03
EES-3055 change permalink copy method

amyb-hiveit:
The previous method of copying a permalink to the clipboard wasn't great for screen reader users as it read the permalink url as well as the 'Link copied to the clipboard' message (see https://github.com/dfe-analytical-services/explore-education-statistics/pull/3518).

This changes the copy method to use `navigator.clipboard`, instead of having to focus on the url input field, so that only the 'Link copied...' message is announced.

### DIFF
--- a/src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
@@ -16,7 +16,6 @@ import React, { useEffect, useState } from 'react';
 
 const linkInstructions =
   'Use the link below to see a version of this page that you can bookmark for future reference, or copy the link to send on to somebody else to view.';
-const permalinkBaseUrl = `${window.location.origin}/data-tables/permalink`;
 
 interface Props {
   tableHeaders?: TableHeadersConfig;
@@ -29,12 +28,12 @@ const TableToolShare = ({
   query,
   selectedPublication,
 }: Props) => {
-  const [permalinkId, setPermalinkId] = useState<string>('');
+  const [permalinkUrl, setPermalinkUrl] = useState('');
   const [permalinkLoading, setPermalinkLoading] = useState<boolean>(false);
   const [screenReaderMessage, setScreenReaderMessage] = useState('');
 
   useEffect(() => {
-    setPermalinkId('');
+    setPermalinkUrl('');
   }, [tableHeaders]);
 
   const handlePermalinkClick = async () => {
@@ -53,20 +52,20 @@ const TableToolShare = ({
       selectedPublication.selectedRelease.id,
     );
 
-    setPermalinkId(id);
+    setPermalinkUrl(`${window.location.origin}/data-tables/permalink/${id}`);
     setPermalinkLoading(false);
 
     setScreenReaderMessage(`Shareable link generated. ${linkInstructions}`);
   };
 
   const handleCopyClick = () => {
-    navigator.clipboard.writeText(`${permalinkBaseUrl}/${permalinkId}`);
+    navigator.clipboard.writeText(permalinkUrl);
     setScreenReaderMessage('Link copied to the clipboard.');
   };
 
   return (
     <>
-      {!permalinkId ? (
+      {!permalinkUrl ? (
         <>
           <h3 className="govuk-heading-s">Save table</h3>
           <LoadingSpinner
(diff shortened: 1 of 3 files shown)


src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
```diff
@@ -16,6 +16,7 @@ import React, { useEffect, useState } from 'react';
 
 const linkInstructions =
   'Use the link below to see a version of this page that you can bookmark for future reference, or copy the link to send on to somebody else to view.';
+const permalinkBaseUrl = `${window.location.origin}/data-tables/permalink`;
```
ntsim, line 19: This seems like it could cause SSR issues as `window` isn't necessarily available in module scope.

Would suggest moving this into the component with state. We could also clean this up a little whilst we're here:

```tsx
const TableToolShare = (...) => {
  const [permalinkUrl, setPermalinkUrl] = useState(''); // replaces permalinkId
  
  ...

  const handlePermalinkClick = async () => {
    ...
    
    setPermalinkUrl(`${window.location.origin}/data-tables/permalink/${id}`);
  };

  const handleCopyClick = () => {
    navigator.clipboard.writeText(permalinkUrl);
  };

  ...
}
```

Note that the link for viewing the permalink should also use the `permalinkUrl` variable

ntsim (changes requested):

src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
```diff
@@ -59,11 +60,7 @@ const TableToolShare = ({
   };
 
   const handleCopyClick = () => {
-    const el = document.querySelector(
-      "[data-testid='permalink-generated-url']",
-    ) as HTMLInputElement;
-    el?.select();
-    document.execCommand('copy');
+    navigator.clipboard.writeText(`${permalinkBaseUrl}/${permalinkId}`);
     setScreenReaderMessage('Link copied to the clipboard.');
```
ntsim, line 64: Think we could probably use a test for this

src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
amyb-hiveit: doh! yes, good catch. That's sorted and a test added now.

src/explore-education-statistics-frontend/src/modules/table-tool/components/__tests__/TableToolShare.test.tsx
```diff
@@ -52,31 +52,78 @@ describe('TableToolShare', () => {
...
+    permalinkService.createPermalink.mockResolvedValue({
+      id: 'permalink-id',
+    } as Permalink);
+
+    const originalClipboard = { ...window.navigator.clipboard };
+    Object.assign(window.navigator, {
+      clipboard: {
+        writeText: jest.fn(),
+      },
+    });
```
ntsim, line 93: Would suggest doing all this mocking setup in common's `setupGlobals.js`:

```js
beforeAll(() => {
  window.navigator.clipboard = {
    writeText: jest.fn(),
  };
});
```

Note that we don't shouldn't need to mess around with `Object.assign` here

src/explore-education-statistics-frontend/src/modules/table-tool/components/__tests__/TableToolShare.test.tsx
```diff
@@ -52,31 +52,78 @@ describe('TableToolShare', () => {
...
-
-      expect(
-        screen.getByRole('link', {
-          name: 'View share link',
-        }),
-      ).toBeInTheDocument();
+    Object.assign(window.navigator, {
+      clipboard: {
+        writeText: originalClipboard,
+      },
```
ntsim, line 126: If we set this up in `setupGlobals`, we don't need to reset this back to what it was before. We can just leave it mocked out as we've configured Jest mocks to automatically reset themselves after every test.

ntsim (changes requested):

src/explore-education-statistics-frontend/src/modules/table-tool/components/__tests__/TableToolShare.test.tsx
amyb-hiveit: Good call, thanks.

ntsim (approved):

PR #3529, opened 2022-10-04
EES-3751 Add high-level info about Permalinks to database

benoutram:
This PR introduces a new content database `Permalink` entity and adds a migration endpoint to add high-level info about Permalinks to the database. It's being done as part of the work to create permalink table snapshots, investigated by [EES-3665](https://dfedigital.atlassian.net/browse/EES-3665).

The content database is chosen over the statistics database to keep statistics purely related to data about statistics only.

Storing high level information about permalinks in the database gives us:

* An easier way to view the set of permalinks in existence and know its size, without browsing Azure storage.
* A way to query all Permalink id’s that we need to create snapshots for.

* A place to track the snapshot migration status as snapshots are created, until the migration is verified as complete when this temporary migration field can be removed.

* Allow for easy reporting of figures on Permalinks created by release / by data set / by time.

* Provide a place to store the `SubjectId` used for calculating the permalink ‘status' which won’t be part of the snapshot.

* Provide a place to store the Permalink title and table title which won't be part of the snapshot.
* Provide for a quicker way of retracting a Permalink in future if the need arises without needing to browse and alter Azure storage.

This PR:

* Creates the new Permalink entity with a database migration.
* Develops a re-runnable migration of high-level Permalink information from Azure storage to the database.
  * Executable by running the secured BAU endpoint `PATCH https://{adminUrl}/api/bau/migrate-permalinks`.
* Begins tracking the high-level information of new permalinks in the database for every new permalink generated from this point on.

We could add 'absolutely everything' we know about the permalink into the database here, but we'd end up with the problem we're trying to solve. We are now moving away from storing a strongly typed query, table configuration and results, to a snapshot of the table that won't cause a problem for future model changes by requiring migrations.

Referential integrity won't be possible for `ReleaseId` and `SubjectId`. Aside from being in a different database,
we know from the investigation of [EES-3068](https://dfedigital.atlassian.net/browse/EES-3068) that many Subjects no longer exist in public-statistics and it also found 61 Permalinks using 6 distinct Subjects that don't exist in the private statistics database either.

`SubjectName` and `PublicationName` of the permalink are also stored in the database. They are combined by the FE to make the permalink title and will be required in future when moving to table snapshots, where the title will not be part of the snapshot.

These could be queried but we know from previous investigations we have permalinks where release and/or subject no longer exist in any database, and the title should stay the same as it was originally when the permalink was created.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20221003123755_EES3751_AddPermalink.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20221003123755_EES3751_AddPermalink.cs
@@ -1,0 +1,45 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Migrations.ContentMigrations
+{
+    public partial class EES3751_AddPermalink : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.CreateTable(
+                name: "Permalinks",
+                columns: table => new
+                {
+                    Id = table.Column<Guid>(type: "uniqueidentifier", nullable: false),
+                    PublicationTitle = table.Column<string>(type: "nvarchar(max)", nullable: false),
+                    DataSetTitle = table.Column<string>(type: "nvarchar(max)", nullable: false),
+                    ReleaseId = table.Column<Guid>(type: "uniqueidentifier", nullable: true),
+                    SubjectId = table.Column<Guid>(type: "uniqueidentifier", nullable: false),
+                    Created = table.Column<DateTime>(type: "datetime2", nullable: false)
+                },
+                constraints: table =>
+                {
+                    table.PrimaryKey("PK_Permalinks", x => x.Id);
+                });
+
+            migrationBuilder.CreateIndex(
+                name: "IX_Permalinks_ReleaseId",
+                table: "Permalinks",
+                column: "ReleaseId");
+
+            migrationBuilder.CreateIndex(
+                name: "IX_Permalinks_SubjectId",
+                table: "Permalinks",
+                column: "SubjectId");
+        }
+
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropTable(
+                name: "Permalinks");
+        }
+    }
+}
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
@@ -63,6 +63,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Model.Database
         public DbSet<HtmlBlock> HtmlBlocks { get; set; }
         public DbSet<MarkDownBlock> MarkDownBlocks { get; set; }
         public DbSet<MethodologyNote> MethodologyNotes { get; set; }
+        public DbSet<Permalink> Permalinks { get; set; } = null!;
         public DbSet<Contact> Contacts { get; set; }
         public DbSet<ReleaseContentSection> ReleaseContentSections { get; set; }
         public DbSet<ReleaseContentBlock> ReleaseContentBlocks { get; set; }
@@ -415,6 +416,18 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Model.Database
             modelBuilder.Entity<MarkDownBlock>()
                 .Property(block => block.Body)
                 .HasColumnName("Body");
+            
+            modelBuilder.Entity<Permalink>()
+                .Property(permalink => permalink.Created)
+                .HasConversion(
+                    v => v,
+                    v => DateTime.SpecifyKind(v, DateTimeKind.Utc));
+
+            modelBuilder.Entity<Permalink>()
+                .HasIndex(data => data.ReleaseId);
+
+            modelBuilder.Entity<Permalink>()
+                .HasIndex(data => data.SubjectId);
 
             modelBuilder.Entity<ReleaseContentSection>()
                 .HasKey(item => new {item.ReleaseId, item.ContentSectionId});
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Services/PermalinkServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Services/PermalinkServiceTests.cs
@@ -116,7 +116,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Api.Tests.Services
             blobStorageService.Setup(s => s.UploadAsJson(
                     Permalinks,
                     Capture.With(blobPathCapture),
-                    It.Is<Permalink>(p =>
+                    It.Is<LegacyPermalink>(p =>
                         p.Configuration.Equals(request.Configuration) &&
                         p.FullTable.IsDeepEqualTo(new PermalinkTableBuilderResult(tableResult)) &&
                         p.Query.Equals(request.Query)),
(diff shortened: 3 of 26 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20221003123755_EES3751_AddPermalink.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            migrationBuilder.CreateTable(
+                name: "Permalinks",
+                columns: table => new
+                {
+                    Id = table.Column<Guid>(type: "uniqueidentifier", nullable: false),
+                    PublicationTitle = table.Column<string>(type: "nvarchar(max)", nullable: false),
+                    DataSetTitle = table.Column<string>(type: "nvarchar(max)", nullable: false),
+                    ReleaseId = table.Column<Guid>(type: "uniqueidentifier", nullable: true),
+                    SubjectId = table.Column<Guid>(type: "uniqueidentifier", nullable: false),
```
duncan-at-hiveit, line 20: When read replica goes out, we can revisit these, as some of them can be made non-nullable if we choose to.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
```diff
@@ -415,9 +416,21 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+
+            modelBuilder.Entity<Permalink>()
+                .HasIndex(data => data.ReleaseId);
+
+            modelBuilder.Entity<Permalink>()
+                .HasIndex(data => data.SubjectId);
 
             modelBuilder.Entity<ReleaseContentSection>()
-                .HasKey(item => new {item.ReleaseId, item.ContentSectionId});
+            .HasKey(item => new {item.ReleaseId, item.ContentSectionId});
```
duncan-at-hiveit, line 433: A bit o' formatting gone awry here

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Services/PermalinkServiceTests.cs
```diff
@@ -176,10 +181,20 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
                     subjectRepository,
                     tableBuilderService);
 
-                Assert.Equal(Guid.Parse(blobPath), result.Id);
+                Assert.Equal(expectedPermalinkId, result.Id);
                 Assert.InRange(DateTime.UtcNow.Subtract(result.Created).Milliseconds, 0, 1500);
                 Assert.Equal(PermalinkStatus.Current, result.Status);
             }
+
+            await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId))
```
duncan-at-hiveit, line 189: I guess this means that the `data` user on Azure environments needs write access to this table.  Worth checking this out ahead of time - it'll also feed back into my db migrations on https://github.com/dfe-analytical-services/explore-education-statistics/pull/3530.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Mappings/MappingProfiles.cs
```diff
@@ -17,12 +17,12 @@ public MappingProfiles()
...
 
-            ConfigureForPermalinks();
+            ConfigureForLegacyPermalinks();
         }
 
-        private void ConfigureForPermalinks()
+        private void ConfigureForLegacyPermalinks()
         {
-            CreateMap<Permalink, PermalinkViewModel>();
+            CreateMap<LegacyPermalink, PermalinkViewModel>();
```
duncan-at-hiveit, line 25: Worth going wild and renaming `PermalinkViewModel` to `LegacyPermalinkViewModel`?  Or will this be gone soon enough (or the name reverted) so it's not worth doing?

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/PermalinkMigrationService.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            throw new InvalidOperationException("Permalink found with no SubjectName");
+        }
+
+        return new Permalink
+        {
+            Id = Guid.Parse(id),
+            PublicationTitle = publicationName,
+            DataSetTitle = subjectName,
+            ReleaseId = null,
```
duncan-at-hiveit, line 122: So there's no way we can retrieve the releaseId?  Perhaps using the `PublicationTitle` and `SubjectId` somehow?  Probably we couldn't infer this for every Release that has amendments, but possibly we could for a majority?

src/GovUk.Education.ExploreEducationStatistics.Publisher/Startup.cs
```diff
@@ -126,6 +126,17 @@ public override void Configure(IFunctionsHostBuilder builder)
                         logger: provider.GetRequiredService<ILogger<QueueService>>()))
                 .AddScoped<IReleasePublishingStatusService, ReleasePublishingStatusService>()
                 .AddScoped<IValidationService, ValidationService>()
+                .AddScoped<IPermalinkMigrationService, PermalinkMigrationService>(provider =>
+                {
+                    var connectionString = GetConfigurationValue(provider, "PublicStorage");
```
duncan-at-hiveit, line 131: Pull out `GetConfigurationValue(provider, "PublisherStorage")` into a variable here too for consistency, or inline this?

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
benoutram: Thanks, I've edited the original commit to fix this.

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Services/PermalinkServiceTests.cs
benoutram: Good point, since it's inserting into the `Permalinks` table I think we will need a minimum of 
`GRANT INSERT ON [dbo].[Permalinks] TO publicapp;`.

Prior to the read-replica work we've run these grants manually.

When this is merged in, depending on where we are with #3530 I will either create a faffy deploy ticket to run this grant or add it as a migration ticket.

I've left a comment on #3530 as I don't understand why this would be the `data` user (as things currently stand). The Data API app service uses the `content` db user `publicapp`.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/PermalinkMigrationService.cs
benoutram: The `ReleaseId` isn't used for anything currently, but I thought it would be a good idea to begin recording it for new permalinks generated from this point onwards.

Happy to remove it altogether if you don't agree with storing it, since it's not used.

I think we possibly could infer the release id from the subject id for the majority of permalinks, but there's going to be edge cases where we can't find a subject id in either private or public stats databases, and like you say, where the release has been amended we won't know which release it was created from.

I'm not sure it's worth doing this at the moment. Since nothing is using `ReleaseId` yet, we can back-fill them at any point in time.

benoutram:
Added an extra commit since the last review to optimise re-running by not retrieving permalinks from storage if they already exist in the database.

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20221003123755_EES3751_AddPermalink.cs
duncan-at-hiveit: Yeah, leave as nullable for now then :+1:

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Services/PermalinkServiceTests.cs
duncan-at-hiveit: So maybe just get this PR in, and then I'll add this permission to the database migration for stage 1 replica.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Mappings/MappingProfiles.cs
duncan-at-hiveit: Cool ta

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/PermalinkMigrationService.cs
duncan-at-hiveit: Yup agreed it's nothing that's needed for now, or perhaps ever!

src/GovUk.Education.ExploreEducationStatistics.Publisher/Startup.cs
duncan-at-hiveit: Cool - I'll look out for that merge conflict :)

duncan-at-hiveit (approved):

PR #3534, opened 2022-10-04
EES-2843 option for line chart legends on lines

amyb-hiveit:
- Adds an option on line charts to show the legend as labels on the lines in the chart.
- You can position the label above or below the line. This option only shows when you select 'line' for the legend position.
- Slightly abuses the recharts label functionality, this normally shows a label on each data point with the point's value, instead we're only showing a label on the end data point and rendering the legend in it.
- It'll work best with legends that have been edited to be shorter, the default ones tend to be quite long so are more likely to overlap the lines.
- I had a go at updating the backend to add the new position option... seems to work :)  

legend

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartDataSet.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartDataSet.cs
@@ -34,8 +34,5 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Model.Chart
 
         [JsonConverter(typeof(StringEnumConverter))]
         public ChartLineStyle? LineStyle;
-
-        [JsonConverter(typeof(StringEnumConverter))]
-        public ChartInlinePosition? InlinePosition;
     }
 }
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartDataSet.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartDataSet.cs
@@ -36,6 +36,6 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Model.Chart
         public ChartLineStyle? LineStyle;
 
         [JsonConverter(typeof(StringEnumConverter))]
-        public ChartLineChartLegendPosition? LineChartLegendPosition;
+        public ChartInlinePosition? InlinePosition;
     }
 }
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartDataSet.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartDataSet.cs
@@ -34,5 +34,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Model.Chart
 
         [JsonConverter(typeof(StringEnumConverter))]
         public ChartLineStyle? LineStyle;
+
+        [JsonConverter(typeof(StringEnumConverter))]
+        public ChartLineChartLegendPosition? LineChartLegendPosition;
     }
 }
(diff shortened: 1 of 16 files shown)


src/explore-education-statistics-common/src/modules/charts/types/legend.ts
```diff
@@ -13,8 +13,11 @@ export interface LegendItemConfiguration {
...
   symbol?: ChartSymbol;
   lineStyle?: LineStyle;
+  lineChartLegendPosition?: LineChartLegendPosition;
 }
 
 export interface LegendItem extends LegendItemConfiguration {
   dataSet: DataSet;
 }
+
+export type LineChartLegendPosition = 'above' | 'below';
```
ntsim, line 23: Similar to above - would name this `LegendInlinePosition`

src/explore-education-statistics-common/src/modules/charts/util/chartUtils.ts
```diff
@@ -25,3 +33,8 @@ export const symbols: ChartSymbol[] = [
 ];
 
 export const lineStyles: LineStyle[] = ['dashed', 'dotted', 'solid'];
+
+export const lineChartLegendPositions: LineChartLegendPosition[] = [
```
ntsim, line 37: As per other suggestions, variable could be `legendInlinePositions`

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/__tests__/ChartLegendConfiguration.test.tsx
```diff
@@ -551,6 +551,66 @@ describe('ChartLegendConfiguration', () => {
     expect(legendItem1.queryByLabelText('Style')).not.toBeInTheDocument();
   });
 
+  test('renders the legend item position field if a line chart with legend position `chart`', () => {
```
ntsim, line 554: Not sure this test case name makes sense

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartLegend.cs
```diff
@@ -19,7 +19,8 @@ public enum ChartLegendPosition
     {
         none,
         bottom,
-        top
+        top,
+        line
```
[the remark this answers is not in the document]
amyb-hiveit: yep good point, I've updated it everywhere.

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartDataSet.cs
```diff
@@ -34,5 +34,8 @@ public class ChartDataSetConfiguration
 
         [JsonConverter(typeof(StringEnumConverter))]
         public ChartLineStyle? LineStyle;
+
+        [JsonConverter(typeof(StringEnumConverter))]
+        public ChartInlinePosition? InlinePosition;
```
ntsim, line 39: We don't need this here as `ChartDataSetConfiguration` has been deprecated for a while and isn't used anymore. We should get rid of it at some point but there's potentially a bunch of old charts from way back when that have this configuration (would need a migration to fix).

src/explore-education-statistics-common/src/modules/charts/types/dataSet.ts
```diff
@@ -50,6 +51,7 @@ export interface DataSetConfigurationOptions {
   colour: string;
   symbol?: ChartSymbol;
   lineStyle?: LineStyle;
+  inlinePosition?: LegendInlinePosition;
```
ntsim, line 54: As mentioned earlier - this config is deprecated and doesn't need to have this field

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartDataSet.cs
amyb-hiveit: Ok, sorted now.

PR #3537, opened 2022-10-06
EES-3049 fix methodology section content layout

amyb-hiveit:
Makes the content full width in methodology sections when there is no headings index.

### DIFF
--- a/src/explore-education-statistics-frontend/src/modules/methodologies/components/MethodologyContentSection.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/methodologies/components/MethodologyContentSection.tsx
@@ -3,7 +3,7 @@ import { useMobileMedia } from '@common/hooks/useMedia';
 import useMounted from '@common/hooks/useMounted';
 import MethodologySectionBlocks from '@frontend/modules/methodologies/components/MethodologySectionBlocks';
 import { ContentBlock } from '@common/services/types/blocks';
-import React, { useRef } from 'react';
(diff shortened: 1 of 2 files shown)


src/explore-education-statistics-frontend/src/modules/methodologies/components/MethodologyContentSection.tsx
```diff
@@ -35,16 +37,23 @@ const MethodologyContentSection = ({
...
 
-      <div className="govuk-grid-column-three-quarters" ref={contentRef}>
+      <div
+        className={
+          hasIndex
+            ? 'govuk-grid-column-three-quarters'
+            : 'govuk-grid-column-full'
+        }
+        ref={contentRef}
+      >
```
markhiveit, line 56: When reading this, at first I thought this code could potentially apply a `div.govuk-grid-column-one-quarter` and a `div.govuk-grid-column-full`. 

Maybe it can be made more apparent that `ContentSectionIndex` returns null if there are no headings, like with an additional `display` boolean prop or something. Additionally, the `const headingElements` calculation could be done inside `MethodologyContentSection` and then the application of the `govuk-grid-column-*` class names can depend on a single variable and not require `hasIndex` to get the info we need.

All that said, happy to go along with whatever you think is best - I haven't thought through the my ideas above and maybe they don't make sense or would lead to wide changes that cannot be justified. But I do feel like there is an opportunity to make this code simpler and more readable.

markhiveit (approved):

PR #3545, opened 2022-10-11
EES-965 change view original buttons

amyb-hiveit:
Change the 'View original' buttons to 'View existing version' for releases and methodologies on
- admin dashboard main tab (current version only, it's not on the new one)
- admin dashboard draft releases tab (current and new versions)
- new Publication - Releases page, https://localhost:5021/publication/id/releases 
- new Publication - Methodologies page, https://localhost:5021/publication/id/methodologies


ntsim (approved):
Approved, but I think we might have a bit of a blindspot in the UI tests as we don't:

- Check that existing versions for release amendments are working
- Check that we can revert back to an existing release version

Probably worth writing up a ticket to add those at some point

PR #3548, opened 2022-10-11
EES-3735 - trimming Data definition title of any leading or trailing …

duncan-at-hiveit:
This PR:
- fixes the scenario whereby an analyst adds trailing or leading whitespace to the Data Definition title of a Key Stat tile, thus making the test `dataDefinitionTitle === 'Help'` fail, and thus not adding in the visually hidden help text


N-moh (approved):
All ok

PR #3553, opened 2022-10-14
EES-3767 In full zip download, rename ancillary to supporting-files

markhiveit:
This PR changes the name of the directory ancillary files are stored in in generated zip files from `ancillary/` to `supporting-files/`. This change doesn't affect how ancillary files are stored in blob storage - they're still stored in an `ancillary` directory.

Zip files are cached on the public side, so zips cached before this change will still have an `ancillary` directory, not `supporting-files`, unless we remove them so they're regenerated.

UI tests to test this will be created as part of `EES-3421`.


src/GovUk.Education.ExploreEducationStatistics.Content.Model/Extensions/FileExtensions.cs
```diff
@@ -80,7 +80,13 @@ public static string PublicPath(this File file, Guid releaseId)
 
         public static string ZipFileEntryName(this File file)
         {
-            return file.Type.GetEnumLabel() + "/" + file.Filename;
+            return file.Type switch
+            {
+                Ancillary => $"supporting-files/{file.Filename}",
+                Data => $"data/{file.Filename}",
+                DataGuidance => $"data-guidance/{file.Filename}",
```
benoutram, line 87: I only put `DataGuidance` in this sample code because I saw it in the screenshot on the ticket showing the contents of an existing zip file.

On review, I don't think `DataGuidance` needs to be here. It's not a file in the db like the `Ancillary` and `Data` files are, and gets added to the zip by another means:

```
                var entry = archive
                    .CreateEntry(FileType.DataGuidance.GetEnumLabel() + "/data-guidance.txt")
                    .SetUnixPermissions("664");
```

benoutram (approved):
I don't think cached zip files are a problem because reading through the code, it looks like they expire after 1 hour.

PR #3562, opened 2022-10-20
EES-3813 - replace print() statements with logger

luke-h1:
This PR: 
* replaces print() statements with a logger 
* Creates new logger utility function which returns a memoized logger via the `lru_cache` strategy (last recently used). Typically the `lru_cache` decorator is good for when a function is complex or invoked frequently (can save computation time, memory, etc.).

### DIFF
--- a/tests/robot-tests/listeners/KeywordListener.py
+++ b/tests/robot-tests/listeners/KeywordListener.py
@@ -1,6 +1,5 @@
-from logging import warning
-
 from robot.running.context import EXECUTION_CONTEXTS
+from tests.libs.logger import get_logger
 
 
 class KeywordListener:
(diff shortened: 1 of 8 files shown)


luke-h1:
Added do not merge as waiting for numerous UI test changes made in #3549 to get merged into dev first

markhiveit:
Just wondering generally what benefit we gain from switching to using a logger?

luke-h1:
> Just wondering generally what benefit we gain from switching to using a logger?

So I'm not too precious about adding a logger but thought it'd be a nice addition. It adds timestamps to logs, makes stack traces a bit better to read & we can configure the logger to emit where it's being called (i.e. down to the function it was called from).

One good thing that may be useful is if we need to debug something that's happening in CI, we could send the logs to a file & output it as part of the test results artifact instead of having to comb through CI logs.

markhiveit (approved):

PR #3563, opened 2022-10-20
EES-3829, EES-3812, EES-3810 - migrate to flake8, black & isort code formatters

luke-h1:
This PR: 
- replaces autopep8 with flake8, black & isort. Note: This shouldn't be merged until #3549 is merged. I've purposely not run the formatters as part of this PR due to the number of conflicts it will create when #3549 is merged into dev
- Performs general cleanup of python test code
- moves visual testing dependencies to root Pipfile 

Addresses:
* [EES-3829](https://dfedigital.atlassian.net/browse/EES-3829)
* [EES-3812](https://dfedigital.atlassian.net/browse/EES-3812)
* [EES-3810](https://dfedigital.atlassian.net/browse/EES-3810)

More info:
* flake8 - https://flake8.pycqa.org/en/latest/
* black - https://pypi.org/project/black/
* isort - https://pypi.org/project/isort/

- [x] Run formatting once #3549 is merged
- [x] Investigate refurb (too opinionated)


markhiveit (changes requested):
I think we could have a small addition to the relevant README.md, giving an example of how these will be used in a typical workflow so we don't all have to dive into the documentation of these 4 new dependencies.

luke-h1:
> I think we could have a small addition to the relevant README.md, giving an example of how these will be used in a typical workflow so we don't all have to dive into the documentation of these 4 new dependencies.

Sounds good. I'll add a section to the testing README around code style

tests/robot-tests/tests/visual_testing/tables_and_charts.py
```diff
@@ -5,45 +5,49 @@
 
 
 class DataBlockType(Enum):
-    CONTENT_BLOCK = 1,
-    FAST_TRACK = 2,
-    SECONDARY_STATS = 3,
+    CONTENT_BLOCK = (1,)
+    FAST_TRACK = (2,)
+    SECONDARY_STATS = (3,)
```
markhiveit, line 10: Seems like in the original code the commas were unnecessary - in fact, it was causing them to be tuples rather than ints, so this is why the code has been changed to this. I think the brackets and the commas can be removed?

markhiveit (approved):

tests/robot-tests/tests/visual_testing/tables_and_charts.py
luke-h1: hmm odd, looks like black was responsible for this. I've disabled formatting for this specific enum for now. Also added a `format:python:check` command to output the diff to avoid issues like this in the future

PR #3564, opened 2022-10-21
EES-3501 - fix prerelease.robot test

luke-h1:
This PR: 
- fixes a small failure some people experienced when running the `prerelease.robot` locally in parallel


benoutram (approved):
UI tests are passing for me with this change and it means I don't need the workaround described on the EES-3501 issue.

PR #3567, opened 2022-10-21
EES-3825 Fix ContentDbContext config for Publications table

markhiveit:
This PR adds config to ContentDbContext to be aligned with the state of our DBs. I created a migration after this change and it was blank.

Ideally, the config wouldn't be `WithMany`, but `WithOne`, as we want Publications and Contacts to have a one-to-one relationship. But currently all environments have contacts that are shared across multiple publications. The time to fix this isn't worth the benefit, so for now I've left it `WithMany`.

Checked for contacts that appear on more than one publication with this query (courtesy of Ben):
```
SELECT ContactId, Count(ContactId) FROM Publications
GROUP BY ContactId
HAVING Count(ContactId) > 1
```

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
@@ -27,7 +27,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Model.Database
         public ContentDbContext(DbContextOptions<ContentDbContext> options) : this(options, true)
         {
         }
-        
+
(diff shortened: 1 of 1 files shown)


benoutram (approved):
I think it would be worth enforcing the correct relationship sooner rather than later as we waste quite a bit of time checking the status of shared contacts in different environments with this ambiguity over whether they can occur or not.

We've had the code in place for creating contacts with new publications for some time now and did a clean-up of Prod data at the time we made that change, so the problem should be limited to old test Publications.

We've cleaned up Prod and the local database image, but it looks like bad data still exists in Dev, Test and Pre-prod.

I've added a note to [EES-3824](https://dfedigital.atlassian.net/browse/EES-3824) to mention this.

PR #3568, opened 2022-10-24
Ees 3798 optimise validator data table usage

duncan-at-hiveit:
# PR Overview

This PR:
* addresses a bug whereby the importer was running out of memory during Mime type and Mime encoding detection
* addresses issues whereby large files were consuming GBs of memory due to our use of DataTables to read entire CSVs into memory

## Reducing memory footprint during Mime activities

By default, it seems that the RedMine Mime library falls back to inspecting the *entire contents* of a file if it can't infer the file type from a Byte Order Mark (BOM).  Therefore, although the usage of the library seems to infer that only the first 1,024 bytes of the file are needed, it in fact pulls the whole CSV in instead (because plain text files don't generally include a BOM).

We now change this to provide the library with a *sample* of a maximum of 10,000 lines from the CSV (if in fact it is a CSV!).  We use a new specific method in FilterTypeValidator for this purpose.  The Admin API also uses this new specific method to reduce its own memory footprint during the validation of a user's file choices.

## Reducing memory during CSV-reading activities in the import

When we use DataTables to read a CSV in, it pulls in the whole CSV results into memory in the form of the populated DataTable.  Thereafter, subsequent operations with the DataTable are very quick, but at the price of memory.  In the case of the large files under test, this amounted to 4 - 6 GBs of memory usage.

We now replace this with StreamReaders, reading the CSV stream line-by-line which is much more memory-efficient.

To help with the readability of this, we now include new `ForEachRow` and `Select` methods in `CsvUtils`, to allow us to act on each row at a time.  In addition, there are some more helper methods in CsvUtils too, for getting CSV headers, total rows etc.

## More efficient cutting up of large CSVs into batches

As above, we're now using StreamReaders instead of DataTables to do this.  To take it a step further, we are not bothering to read the CSV lines into a CsvReader only to then create a new mini-CSV from it, but rather just taking the raw lines from the original CSV and adding them to the new mini-CSV as raw strings, so without any kind of transformation. 

## Reducing one whole up-front pass of the Data CSV

The pre-existing code did a full sweep of the Data CSV to check it had the correct number of cells on every row before moving on to the next stage of validation, which then scans the entire CSV again.  I've moved this cell-checking to the subseequent sweep and removed the first sweep altogether, thus removing one full pass of the CSV.


markhiveit (dismissed):

src/GovUk.Education.ExploreEducationStatistics.Common/Services/FileTypeService.cs
```diff
@@ -65,13 +95,100 @@ public bool HasMatchingEncodingType(Stream stream, IEnumerable<string> encodingT
...
+                    }
+                    
+                    lines.Add(nextLine);
+                    linesRead++;
+                }
+                
+            }
+            catch (Exception e)
+            {
+                _logger.LogError("Unable to read sample lines from CSV - {ErrorMessage}", e.Message);
```
benoutram, line 171: I think `_logger.LogError(e, "Unable to read sample lines from CSV")` would be sufficient here, rather than two separate errors which will be recorded in App Insights as two errors.

The message and stacktrace should still be available in the log event that gets recorded.

src/GovUk.Education.ExploreEducationStatistics.Common/Services/FileTypeService.cs
```diff
@@ -65,13 +95,100 @@ public bool HasMatchingEncodingType(Stream stream, IEnumerable<string> encodingT
             return encodingTypes.Any(pattern => pattern.Equals(encodingType));
         }
 
+        public async Task<bool> IsValidCsvDataOrMetaFile(
```
benoutram, line 98: Could call this `IsValidCsvFile` ?

src/GovUk.Education.ExploreEducationStatistics.Common/Services/Interfaces/IFileTypeService.cs
```diff
@@ -1,3 +1,4 @@
+using System;
```
benoutram, line 1: There's some unused imports that could be cleaned up here.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Utils/CsvUtil.cs
```diff
@@ -11,6 +14,237 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Processor.Utils
 {
     public static class CsvUtil
     {
+        /// <summary>
```
benoutram, line 17: I think it would be worth adding `#nullable enable` to this class now to catch any problems with all this new code (of which I'm sure there aren't any! 😆 )

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/FileUploadsValidatorService.cs
```diff
@@ -60,7 +60,7 @@ public class FileUploadsValidatorService : IFileUploadsValidatorService
 
         public async Task<Either<ActionResult, Unit>> ValidateFileForUpload(IFormFile file, FileType type)
         {
-            if (type != Ancillary && type != Chart && type != Image)
+            if (type is FileType.Data or Metadata)
```
benoutram, line 63: Should this include all the `FileType`'s except for `Ancillary`, `Chart` and `Image`?

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Functions/Processor.cs
```diff
@@ -127,6 +127,8 @@ ILogger<Processor> logger
                 _logger.LogError(ex, $"{GetType().Name} function FAILED for : Import: " +
                                      $"{message.Id} : {ex.Message}");
 
+                _logger.LogError(ex.StackTrace);
```
benoutram, line 130: Stacktrace should be available with previous log message I think, so is this needed?

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
```diff
@@ -220,42 +224,38 @@ public int GetYear(IReadOnlyList<string> rowValues, List<string> colValues)
...
-                        ((batchNo - 1) * import.RowsPerBatch) + i++ + 2);
-                    observations.Add(o);
+                        (batchNo - 1) * import.RowsPerBatch + index + 2);
                 }
-            }
 
-            return observations;
+                return null;
+            }))
+                .Where(observation => observation != null);
```
benoutram, line 253: Minor but could use `.WhereNotNull()` here.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterMetaService.cs
```diff
@@ -58,33 +67,30 @@ public SubjectMeta Get(DataColumnCollection cols, DataRowCollection rows, Subjec
...
-                DecimalPlaces = values[7] == null ? (int?) null : int.Parse(values[7])
-            });
+            return CsvUtil.BuildType(
+                cells, 
+                cols, 
+                Enum.GetNames(typeof(MetaColumns)), 
+                values => new MetaRow
+                {
+                    ColumnName = values[0],
+                    ColumnType = (ColumnType) Enum.Parse(typeof(ColumnType), values[1]),
```
benoutram, line 86: I know this isn't your change but this could be `Enum.Parse<ColumnType>(values[1])` without the cast.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ValidatorService.cs
```diff
@@ -90,89 +92,72 @@ public class ValidatorService : IValidatorService
...
-                    colCount = dr.FieldCount;
-                }
-            }
-
-            if (errors.Count > 0)
-            {
-                return errors;
+                return ListOf(isMetaFile
+                    ? new DataImportError($"{MetaFileMustBeCsvFile.GetEnumLabel()}")
+                    : new DataImportError($"{DataFileMustBeCsvFile.GetEnumLabel()}"));
```
benoutram, line 138: I don't think the string interpolation is doing anything here?

```
                return ListOf(isMetaFile
                    ? new DataImportError(MetaFileMustBeCsvFile.GetEnumLabel())
                    : new DataImportError(DataFileMustBeCsvFile.GetEnumLabel()));
```

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Utils/CsvUtil.cs
duncan-at-hiveit: Already done on the follow-up PR thanks :D

src/GovUk.Education.ExploreEducationStatistics.Common/Services/Interfaces/IFileTypeService.cs
duncan-at-hiveit: Cleaned up thanks

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
```diff
@@ -220,42 +224,38 @@ public int GetYear(IReadOnlyList<string> rowValues, List<string> colValues)
...
         }
 
         private Observation ObservationFromCsv(
             StatisticsDbContext context,
-            string[] rowValues,
+            List<string> rowValues,
```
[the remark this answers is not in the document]
duncan-at-hiveit: Hey!  I tested and profiled this before-and-after and didn't notice any particular speed changes, but it was a good shout to take a look.

As a great aside, whilst doing the profiling I found that we were losing a lot of time to matching CSV labels to enum values during import.  Adding this as an additional optimisation to the 3799 PR adds another big boost to import speed, so this goodness came out of this PR comment, so thanks!!

PR #3570, opened 2022-10-24
EES-257 data labels on charts

amyb-hiveit:
Adds the ability to add data labels to charts.
- For bar charts the labels can be set to appear outside or inside the bars (note that we'll have to rely on the analysts here to chose a colour for the bars that makes the label readable if it's set to inside)
- For line charts the labels can be above or below the line. Showing data labels isn't compatible with showing the legend inline on the chart, so if they are both set a warning is shown and the data labels take precedence.
- I've again bodged in the backend... let me know if it needs changing.

chart2
chart3
chart1


src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/Charts.cs
```diff
@@ -7,75 +7,94 @@
...
+    public abstract ChartType Type { get; }
+
+    public Dictionary<string, ChartAxisConfiguration>? Axes { get; set; }
+    public ChartLegend? Legend { get; set; }
+  }
+
+  public class LineChart : Chart
+  {
+    public override ChartType Type => Line;
+    public bool showDataLabels { get; set; }
```
luke-h1, line 55: Same as above. Would capitalize / camel-case this property

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/Charts.cs
```diff
@@ -7,75 +7,94 @@
...
+    public LineChartDataLabelPosition? DataLabelPosition { get; set; }
+  }
+
+  public class HorizontalBarChart : Chart
+  {
+    public override ChartType Type => HorizontalBar;
+
+    public int? BarThickness { get; set; }
+    public bool Stacked;
+    public bool showDataLabels { get; set; }
```
luke-h1, line 65: Same as above. Think this should be pascal cased

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/Charts.cs
```diff
@@ -7,75 +7,94 @@
...
+    public BarChartDataLabelPosition? DataLabelPosition { get; set; }
+  }
+
+  public class VerticalBarChart : Chart
+  {
+    public override ChartType Type => VerticalBar;
+
+    public int? BarThickness { get; set; }
+    public bool Stacked;
+    public bool showDataLabels { get; set; }
```
luke-h1, line 75: Same as above. Think this should be pascal cased

luke-h1 (dismissed):

luke-h1:
Just wondering if the text on the bar chart will ever be affected by the color of the bar chart line?  Just thinking of scenarios where the text might be hard to read/have contrast issues due to the color of the bar chart line

amyb-hiveit:
> Just wondering if the text on the bar chart will ever be affected by the color of the bar chart line? Just thinking of the first image where the text might be hard to read/have contrast issues due to the color of the bar chart line

Yeah potentially, but I don't think we can do anything to automatically fix it as the width of the bar charts varies meaning the text might be wider than them so will have to work against the bar colour and the white background. We'll have to rely on the analysts to make sure the bar colour works with the text.

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/Charts.cs
```diff
@@ -7,75 +7,94 @@
...
-    public class InfographicChart : Chart
-    {
-        public override ChartType Type => Infographic;
-        public string FileId { get; set; }
-    }
+
+  [JsonConverter(typeof(StringEnumConverter))]
+  public enum BarChartDataLabelPosition
+  {
+    inside, outside
```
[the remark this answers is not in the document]
amyb-hiveit: Thanks Luke, I've fixed this and the other casing issues now.

ntsim:
Have we checked that changing the casing doesn't break existing charts?

amyb-hiveit:
> Have we checked that changing the casing doesn't break existing charts?
> […](#)

It shouldn't do as these are new properties and seems fine for the existing ones on my local.

src/explore-education-statistics-common/src/modules/charts/components/LineChartLabel.tsx
```diff
@@ -0,0 +1,79 @@
+import { LineChartDataLabelPosition } from '@common/modules/charts/types/chart';
+import { LegendInlinePosition } from '@common/modules/charts/types/legend';
+import formatPretty from '@common/utils/number/formatPretty';
+import { LabelProps } from 'recharts';
+import React from 'react';
+
+interface Props extends LabelProps {
```
ntsim, line 7: Wouldn't extend `LabelProps`. It contains a whole bunch of things that we don't pass onto the final rendered element so it'd be misleading to include these.

We should just define the props `x`, `y` and `value`

src/explore-education-statistics-common/src/modules/charts/components/LineChartLabel.tsx
```diff
@@ -0,0 +1,79 @@
...
+  colour: string;
+  dataLabelPosition?: LineChartDataLabelPosition;
+  decimalPlaces?: number;
+  index: number;
+  name: string;
+  legendLabelPosition: LegendInlinePosition;
+  showDataLabels?: boolean;
+  showLegendAsLabel?: boolean;
```
ntsim, line 15: Would suggest renaming these to `isDataLabel` and `isLegendLabel`.

On a side note, I wouldn't normally encourage having multiple boolean props to express a single outcome like this. The alternative would be to have a `type` prop instead (which would have a value of either `legend` or `data`), but this makes the calling function a little bit messy so I'll allow it this time :laughing:

src/explore-education-statistics-common/src/modules/charts/components/LineChartLabel.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { LabelProps } from 'recharts';
+import React from 'react';
+
+interface Props extends LabelProps {
+  colour: string;
+  dataLabelPosition?: LineChartDataLabelPosition;
+  decimalPlaces?: number;
+  index: number;
+  name: string;
+  legendLabelPosition: LegendInlinePosition;
```
ntsim, line 13: This should be optional. I'd also potentially go bit further and merge this with `dataLabelPosition` as a single `position` prop:

```ts
position?: LineChartDataLabelPosition | LegendInlinePosition;
```

Which could be then called like:

```tsx
position={showDataLabels ? dataLabelPosition : config.inlinePosition}
```

ntsim (approved):

PR #3571, opened 2022-10-24
EES-3652 fixes for decimals, negatives and default colour

amyb-hiveit:
Fixes the following issues found by @N-moh during testing:

- group values with decimals were being rounded when the form is saved
- the logic to prevent overlapping groups didn't work for negative values 
- areas that aren't covered by custom groups could be hard to distinguish from those that were as they defaulted to grey which can be indistinguishable from the low end of the legend colour scale, I've changed them to white.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartCustomDataGroup.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartCustomDataGroup.cs
@@ -2,7 +2,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Model.Chart;
 
 public class ChartCustomDataGroup
 {
-    public int Min;
-    public int Max; 
+  public decimal Min;
+  public decimal Max;
 }
 
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartCustomDataGroup.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartCustomDataGroup.cs
@@ -2,7 +2,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Model.Chart;
 
 public class ChartCustomDataGroup
 {
-  public decimal Min;
-  public decimal Max;
+    public decimal Min;
+    public decimal Max;
 }
 
--- a/src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartMapCustomGroupsConfiguration.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartMapCustomGroupsConfiguration.tsx
@@ -72,16 +72,25 @@ export default function ChartMapCustomGroupsConfiguration({
                 .test('noOverlap', 'Groups cannot overlap', function noOverlap(
                   value: number,
                 ) {
-                  return !(
-                    groups.length &&
-                    groups.some(group => {
-                      return (
-                        (value >= group.min && value <= group.max) ||
-                        // eslint-disable-next-line react/no-this-in-sfc
-                        (value <= group.max && group.min <= this.parent.max)
-                      );
-                    })
+                  // show an error when:
+                  // - the min is in an existing group, or
+                  // - this new group overlaps with another group and the max isn't in a group
+                  // (if the max is in a group and the min isn't then we only want to show the
+                  // error on the max field as it's the one that requires action)
+                  /* eslint-disable react/no-this-in-sfc */
+                  const minIsInAGroup = groups.some(
+                    group => value >= group.min && value <= group.max,
                   );
+                  const maxIsInAGroup = groups.some(
+                    group =>
+                      this.parent.max >= group.min &&
+                      this.parent.max <= group.max,
+                  );
+                  const overlaps = groups.some(
+                    group => value <= group.max && group.min <= this.parent.max,
+                  );
+                  return !minIsInAGroup && !(overlaps && !maxIsInAGroup);
+                  /* eslint-enable react/no-this-in-sfc */
                 }),
               max: Yup.number()
                 .required('Enter a maximum value')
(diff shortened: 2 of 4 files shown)


src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartCustomDataGroup.cs
```diff
@@ -2,7 +2,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Model.Chart;
 
 public class ChartCustomDataGroup
 {
-    public int Min;
-    public int Max; 
+  public decimal Min;
+  public decimal Max;
```
ntsim, line 6: The indentation is wrong - should be 4.

src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
```diff
@@ -555,6 +555,11 @@ function generateFeaturesAndDataGroups({
     unit,
   });
 
+  // Default to white for areas not covered by custom data sets
+  // to make it clearer which aren't covered by the groups.
+  const defaultColour =
+    classification === 'Custom' ? 'rgb(255, 255, 255)' : 'rgba(0,0,0,0)';
```
ntsim, line 561: Would use `rgba` for consistency i.e. `rgba(255, 255, 255, 1)`

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartMapCustomGroupsConfiguration.tsx
```diff
@@ -75,7 +75,11 @@ export default function ChartMapCustomGroupsConfiguration({
                   return !(
                     groups.length &&
                     groups.some(group => {
-                      return value >= group.min && value <= group.max;
+                      return (
+                        (value >= group.min && value <= group.max) ||
+                        // eslint-disable-next-line react/no-this-in-sfc
+                        (value <= group.max && group.min <= this.parent.max)
```
ntsim, line 81: This condition is a little gnarly to me. Is there a potential issue with this resolving to true if the max field hasn't been filled in yet?

Can we also add a comment explaining what's going on here?

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartMapCustomGroupsConfiguration.tsx
```diff
@@ -88,7 +92,11 @@ export default function ChartMapCustomGroupsConfiguration({
                   return !(
                     groups.length &&
                     groups.some(group => {
-                      return value >= group.min && value <= group.max;
+                      return (
+                        (value >= group.min && value <= group.max) ||
+                        // eslint-disable-next-line react/no-this-in-sfc
+                        (value <= group.min && group.max <= this.parent.min)
```
ntsim, line 98: Same as above

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartMapCustomGroupsConfiguration.tsx
```diff
@@ -72,25 +72,51 @@ export default function ChartMapCustomGroupsConfiguration({
...
+                  const maxIsInAGroup = groups.some(
+                    group =>
+                      this.parent.max >= group.min &&
+                      this.parent.max <= group.max,
+                  );
+                  const overlaps = groups.some(
+                    group => value <= group.max && group.min <= this.parent.max,
+                  );
+                  return !minIsInAGroup && !(overlaps && !maxIsInAGroup);
+                  /* eslint-enable react/no-this-in-sfc */
```
ntsim, line 93: Would rewrite this to use early returns so that we don't perform any more work than is necessary. The ternary is also a little hard to understand still because of all the negations.

One way of writing this might be something like:

```ts
// Some comment about checking the min
if (groups.some(group => value >= group.min && value <= group.max)) {
  return false;
}

// Some comment about overlapping another group
const overlaps = groups.some(
  group => value <= group.max && group.min <= this.parent.max,
);

if (overlaps) {
  // Some comment about why we're checking the max instead
  return !groups.some(
      group =>
        this.parent.max >= group.min &&
        this.parent.max <= group.max,
    );
}

return true;
```

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartMapCustomGroupsConfiguration.tsx
```diff
@@ -72,25 +72,51 @@ export default function ChartMapCustomGroupsConfiguration({
...
+                      this.parent.min <= group.max,
+                  );
+                  const maxIsInAGroup = groups.some(
+                    group => value >= group.min && value <= group.max,
+                  );
+                  const overlaps = groups.some(
+                    group => this.parent.min <= group.max && group.min <= value,
                   );
+                  return !maxIsInAGroup && !(overlaps && !minIsInAGroup);
+                  /* eslint-enable react/no-this-in-sfc */
```
ntsim, line 119: Same as above - would suggest rewriting this to use early returns

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/__tests__/ChartMapCustomGroupsConfiguration.test.tsx
```diff
@@ -221,25 +213,137 @@ describe('ChartMapCustomGroupsConfiguration', () => {
...
+    )[1];
+
+    await waitFor(() => {
+      expect(
+        within(minCell).queryByText('Groups cannot overlap'),
+      ).not.toBeInTheDocument();
+
+      expect(
+        within(maxCell).getByText('Groups cannot overlap'),
+      ).toBeInTheDocument();
```
ntsim, line 254: Would make this error more specific to the field e.g. 'Max cannot overlap another group'

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/__tests__/ChartMapCustomGroupsConfiguration.test.tsx
```diff
@@ -221,25 +213,137 @@ describe('ChartMapCustomGroupsConfiguration', () => {
...
+      'cell',
+    )[1];
+
     await waitFor(() => {
       expect(
         within(minCell).getByText('Groups cannot overlap'),
       ).toBeInTheDocument();
+      expect(
+        within(maxCell).getByText('Groups cannot overlap'),
+      ).toBeInTheDocument();
```
ntsim, line 285: Same as earlier - would make these errors more specific to the field e.g. 'Min cannot overlap another group' and 'Max cannot overlap another group'

Same applies for the other test cases

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartMapCustomGroupsConfiguration.tsx
amyb-hiveit: I've changed both to use early returns and have clearer messages now.

ntsim (approved):

PR #3572, opened 2022-10-25
Revert EES-3812 and EES-3813

markhiveit:


markhiveit:
Ben's revert got merged instead

PR #3575, opened 2022-10-26
EES-3801 fix admin style issues

amyb-hiveit:
Fixes some styling issues in the new admin Publication - Releases and Methodologies pages and Dashboard - Draft releases tab:
- fixes the layout of buttons on smaller screens 
- makes the button order and colour consistent 

Also fixed some other things that have been bugging me:
- changed `dfe-width-container--wide` to improve it's responsiveness (currently there is no side margin on screen sizes between 1020 and 1300px)
- added `flex-wrap: wrap` to topic publications to prevent the 'Create publication' button squashing the list on small screens


src/explore-education-statistics-common/src/components/form/FormFieldset.tsx
```diff
@@ -59,7 +59,7 @@ const FormFieldset = ({
             'govuk-fieldset__legend',
             `govuk-fieldset__legend--${legendSize}`,
             {
-              'govuk-visually-hidden': legendHidden,
+              'govuk-visually-hidden govuk-!-font-size-14': legendHidden,
```
ntsim, line 62: I'm a little confused at how this works. How is the visually hidden legend actually causing overflow issues? And why is the solution to make the font size smaller?

Seems to me like there's a more foundational issue here that requires a different styling fix than changing font size. Like what if the legend is just so long is still causes this overflow issue even with the smaller font?

ntsim (changes requested):

ntsim (approved):

PR #3576, opened 2022-10-26
EES-3189 fix footnotes width in admin

amyb-hiveit:
Fixes a bug where the footnotes for datablocks are only using 2/3’s of the available space in the admin (they're fine on the public site). 

This was due to a style added to restrict the width of footnotes under tables in the table tool / on permalinks that was also affecting all footnotes in the admin.

I've checked that the print style for permalinks is unaffected by this change.

### DIFF
--- a/src/explore-education-statistics-common/src/modules/table-tool/components/FixedMultiHeaderDataTable.tsx
+++ b/src/explore-education-statistics-common/src/modules/table-tool/components/FixedMultiHeaderDataTable.tsx
@@ -12,8 +12,8 @@ interface Props extends OmitStrict<MultiHeaderTableProps, 'ariaLabelledBy'> {
   caption: ReactNode;
   captionId: string;
   innerRef?: Ref<HTMLElement>;
-  isInDataBlock?: boolean;
   footnotes?: FullTableMeta['footnotes'];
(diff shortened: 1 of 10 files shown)


src/explore-education-statistics-common/src/modules/table-tool/components/FixedMultiHeaderDataTable.tsx
```diff
@@ -133,10 +133,12 @@ const FixedMultiHeaderDataTable = forwardRef<HTMLElement, Props>(
           />
         </div>
 
-        <FigureFootnotes
-          footnotes={footnotes}
-          headingHiddenText={footnotesHeadingHiddenText}
-        />
+        <div className="govuk-!-width-two-thirds">
```
ntsim, line 136: Wouldn't this mean that the footnotes will always be 2/3 width? Wouldn't this cause data block footnotes to be too narrow? To me, it seems like it'd more appropriate for these to be full width.

amyb-hiveit: Doh yes, I've changed it now.

ntsim (approved):

PR #3578, opened 2022-10-27
Ees 3799 stage 2 rerunnable and faster

duncan-at-hiveit:
# PR overview

This PR:
* addresses a bug whereby the restart of Importer Stage 2 was accidentally attempting to re-persist various pieces of Subject Meta when Stage 2 restarts
* speeds up Stage 2 by simply first collecting all unique FilterGroups, FilterItems and Locations from scanning the CSV, and then persisting at the end

:warning: Changes in this PR expect that Subjects that have only partially imported Stage 2 but have not yet *failed* will only be saving information in atomic transactions (one for Filters, Indicators and IndicatorGroups, and another for FilterGroups, FilterItems and Locations).  Therefore when this work goes out, we don't want to do it at a time when there are any ongoing imports occurring at Stage 2.  

## Additional changes

* Subject Meta (Filters, Indicators and IndicatorGroups) is now stored in a single atomic transaction.  Therefore we don't need to support the pre-existing code that had to keep checking whether or not a given Filter or Indicator already exists before deciding whether or not to persist.  This cleans things up a bit, and it's very obvious now and in one place where new Filters, Indicators and IndicatorGroups are persisted. 
* FilterGroups and FilterItems are now stored in a single atomic transaction too.  Therefore this simplifies the code in the same way as above.
* Locations are still checked to see if they already exist and saved if not, and added to the shared importer cache when persisted or looked up.  This is so that parallel imports don't attempt to insert the same Location as before.  The Location-reading code and the Location-saving code is split up however again as above to simplify.
* The Stage 3 and Stage 4 code no longer has to use the code that checks to see of Filters, FilterGroups, FilterItems, Indicators, IndicatorGroups or Locations exist yet - instead, we already know that they will by this stage.  Therefore they just use code that looks up entities that it *knows* will be there already, thereby simplifying the code again.
* The ImporterMemoryCache is no longer periodically cleared out at various stages starting.  Instead, it now has a size limit and sliding expiry windows on cached entities.  This means that the parallel running of various batch file imports can take better advantage of the shared ImporterMemoryCache and not keep clearing out cached results that each other would get good use out of.
* At the very end of Stage 2, after everything is created in the database, it is all stashed in ImporterMemoryCache, ready for Stage 3 to make use of.

## New tests

There is a new integration test, ProcessorServiceTests.  This test only really mocks out the BlobCache but the rest is for real, testing Stage 2.  Because the InMemory databases don't support transactions, we need to work around this a bit with a test-specific version of a helper component that provides no-op versions of transactions.  We could do with extending this test class with lots of examples of various different CSVs.

## Other abandoned changes

I tried to replace the use of the `InsertObservations` and `InsertObservationFilterItems` stored procedures with the use of Thinktecture's `BulkInsert` EF extensions.  Unfortunately this resulted in database lockups between the various parallel batch imports though, so I pulled it back out though.  It would've meant another area that we could've removed DataTable usage, but seems like this might be one for another time.

## Other possible things to do if only I had time

* Move `GetFilterItemCacheKey` and `GetFilterGroupCacheKey` out of `ImporterFilterService` and into ImporterMemoryCache.  Remove the reliance on needing the DcContext as part of generating these keys (I suspect they don't *actually* need it).
* Encapsulate the stashing and grabbing of Locations, FilterGroups and FilterItems into entity-specific methods on ImporterMemoryCache e.g. `SetLocation(location)`, `GetLocation(<Location attributes from CSV>)`, so that the Cache Key-generation methods don't need to be public anymore.
* Add integration tests for Stages 1, 3 and 4 of the import process.  These should be fairly easy to add on using what's already in place in `ProcessorServiceTests` for wiring up the service.

## Memory and speed improvements

* Work done here locally shows memory usage dropping down from 4 - 6 GBs for very large files to 150 - 200 MBs.
* Work done here locally shows imports of 19 minutes dropping to 14 minutes, thus shaving off about 25% of the import time.  Hopefully this'll show better improvements in Azure environments where memory is a bit more of a precious commodity.

## Test results

image

## Updated test results (2/11/2023)

After an addition round of updates after profiling (and optimising some enum label lookups), I recorded a new import run with 40% of the time shaven off (from 32.5 minutes down to 18 minutes).  Importing the `e-and-t-geography-detailed_6years_reordered.csv` CSV on my laptop:

```
Against the "dev" barch:
[processor] [2022-11-02T13:25:47.008Z] Validating: e-and-t-geography-detailed_6years_reordered.csv
[processor] [2022-11-02T13:31:58.785Z] Processor Function processing import message for e-and-t-geography-detailed_6years_reordered.csv at stage STAGE_2
[processor] [2022-11-02T13:39:20.013Z] Processor Function processing import message for e-and-t-geography-detailed_6years_reordered.csv at stage STAGE_3
[processor] [2022-11-02T13:43:51.083Z] Update: e-and-t-geography-detailed_6years_reordered.csv STAGE_3 (100%) -> STAGE_4 (0%)
[processor] [2022-11-02T13:58:09.663Z] Update: e-and-t-geography-detailed_6years_reordered.csv STAGE_4 (99%) -> COMPLETE (100%)

Stage 1: 6mins 10s 
Stage 2: 7min 20s
Stage 3: 4min 30s
Stage 4: 14min 20s
Total: 32.5min
```

```
On this branch:
[processor] [2022-11-02T14:05:21.731Z] Validating: e-and-t-geography-detailed_6years_reordered.csv
[processor] [2022-11-02T14:07:33.678Z] Processor Function processing import message for e-and-t-geography-detailed_6years_reordered.csv at stage STAGE_2
[processor] [2022-11-02T14:11:21.322Z] Processor Function processing import message for e-and-t-geography-detailed_6years_reordered.csv at stage STAGE_3
[processor] [2022-11-02T14:12:21.294Z] Update: e-and-t-geography-detailed_6years_reordered.csv STAGE_3 (99%) -> STAGE_3 (100%)
[processor] [2022-11-02T14:23:12.380Z] Update: e-and-t-geography-detailed_6years_reordered.csv STAGE_4 (99%) -> COMPLETE (100%)


Stage 1: 2mins 10s (-4min)
Stage 2: 3mins 50s (-3min 30s)
Stage 3: 1min (-3min 30s)
Stage 4: 10min 50s (-3min 30s)
Total: 18min (-18.5 mins)
```

## Updated UI test results (2/11/2023)

image


src/GovUk.Education.ExploreEducationStatistics.Data.Processor.Tests/Functions/ProcessorStage4ImportObservationTests.cs
```diff
@@ -206,12 +213,411 @@ await using (var contentDbContext = InMemoryContentDbContext(_contentDbContextId
...
+            .Setup(s => s.StreamBlob(PrivateReleaseFiles, import.File.Path(), null))
+            .ReturnsAsync(() => System.IO.File.OpenRead(dataFilePath));
+
+        blobStorageService
+            .Setup(s => s.StreamBlob(PrivateReleaseFiles, import.MetaFile.Path(), null))
+            .ReturnsAsync(() => System.IO.File.OpenRead(metaFilePath));
+        
+        blobStorageService
+            .Setup(s => s.DeleteBlob(PrivateReleaseFiles, importObservationsMessage.ObservationsFilePath))
+            .Returns(Task.CompletedTask);
```
benoutram, line 335: You could use `MockBlobStorageServiceExtensions` here and in the test below, and potentially add a new extension for `StreamBlob` using a `File` or a file path?

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ValidatorService.cs
```diff
@@ -320,9 +317,10 @@ public class ValidatorService : IValidatorService
             );
         }
 
-        private int GetRowsPerBatch(ExecutionContext executionContext)
+        private int GetRowsPerBatch()
         {
-            return _overrideRowsPerBatch ?? Convert.ToInt32(LoadAppSettings(executionContext).GetValue<string>("RowsPerBatch"));
+            return Int32.Parse(Environment.GetEnvironmentVariable("RowsPerBatch")
```
benoutram, line 322: I would expect this change has left the method `LoadAppSettings` unused now?

benoutram (approved):

PR #3579, opened 2022-10-27
EES-3826 Add IncludeInResponse field for subject meta request

markhiveit:
This PR adds an `IncludeInResponse` field when fetching subject meta data. This means the request can specify what information it wants in the response. Changes have also been made to the frontend to take advantage of this.

It was decided that for now the request cannot return multiple different types of meta data - this was done to keep the endpoint simpler for now, although it would be simple enough to make it return multiple types of meta data in the future if necessary.

I've not added any unit tests as they're time consuming to write and I may have to rework this.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Data/Query/ObservationQueryContext.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Data/Query/ObservationQueryContext.cs
@@ -1,3 +1,4 @@
+#nullable enable
 using System;
 using System.Collections.Generic;
 using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
@@ -7,15 +8,18 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Model.Data.Query
     public class ObservationQueryContext
     {
         public Guid SubjectId { get; set; }
-        public TimePeriodQuery TimePeriod { get; set; }
-        public IEnumerable<Guid> Filters { get; set; }
+        public TimePeriodQuery? TimePeriod { get; set; }
+
+        public IEnumerable<Guid>? Filters { get; set; }
         // TODO EES-3328 - remove BoundaryLevel from ObservationQueryContext as we now store it on
         // MapChart configuration instead
         public long? BoundaryLevel { get; set; }
-        public IEnumerable<Guid> Indicators { get; set; }
+        public IEnumerable<Guid>? Indicators { get; set; }
         public List<Guid> LocationIds { get; set; } = new();
         [Obsolete("Legacy Location field that exists in queries of historical Permalinks", false)]
-        public LocationQuery Locations { get; set; }
+        public LocationQuery? Locations { get; set; }
+
+        public IncludeInResponseQuery? IncludeInResponse { get; set; }
 
         public override string ToString()
         {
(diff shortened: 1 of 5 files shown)


src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
```diff
@@ -168,21 +162,110 @@ private async Task<SubjectMetaViewModel> GetSubjectMetaViewModel(ReleaseSubject
...
+                        return new SubjectMetaViewModel
+                        {
+                            Indicators = indicators,
+                        };
+                    }
+                };
+
+            }
+
+            // NOTE: If IncludeInResponse not set, revert to previous behaviour...
```
ntsim, line 253: Why are we maintaining the old behaviour? Subject meta filtering only happens in table tool and we control how this endpoint is used so there's no need to maintain this at all.

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Data/Query/ObservationQueryContext.cs
```diff
@@ -7,15 +8,18 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Model.Data.Query
...
         // MapChart configuration instead
         public long? BoundaryLevel { get; set; }
-        public IEnumerable<Guid> Indicators { get; set; }
+        public IEnumerable<Guid>? Indicators { get; set; }
         public List<Guid> LocationIds { get; set; } = new();
         [Obsolete("Legacy Location field that exists in queries of historical Permalinks", false)]
-        public LocationQuery Locations { get; set; }
+        public LocationQuery? Locations { get; set; }
+
+        public IncludeInResponseQuery? IncludeInResponse { get; set; }
```
ntsim, line 22: Think `ObservationQueryContext` is the wrong place to make these changes. In fact, `ObservationQueryContext` shouldn't be used for the filter subject meta endpoint at all really.

We should make a new request for filtering subject meta. This would only contain the things we're actually interested in e.g.

```cs
public record SubjectMetaFilterRequest
{
    public Guid SubjectId { get; set; }
    public TimePeriodQuery? TimePeriod { get; set; }
    public IEnumerable<Guid>? Filters { get; set; }
    public IEnumerable<Guid>? Indicators { get; set; }
    public List<Guid>? LocationIds { get; set; }
    public SubjectMetaFacetType IncludeFacetType { get; set; }
}

public enum SubjectMetaFacetType 
{
    Locations,
    TimePeriods,
    FilterItems,
    Indicators,
}
```

As mentioned in planning, I had issue with the `IncludeInResponse` stuff, so would suggest we change this to something simpler and more generic with the `IncludeFacetType` parameter.

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectMetaService.cs
markhiveit: This is still in as a consequence of not doing the frontend work to separate filter and indicators steps in the same PR. So the request made in the final table tool step still needs the old functionality. 

Once the frontend work is done to separate them, then this can be removed. I'll update the comment so this is clearer.

markhiveit:
I'm going to close this because it's unclear how long before this work gets prioritised. I've left a note about the branch this work is on in the Jira ticket.

PR #3591, opened 2022-11-01
EES-3421 test to check correct files in downloaded zip -data catalogue

N-moh:
EES-3421 Robot tests to check correct files in downloaded zip -data catalogue


tests/robot-tests/tests/libs/common.robot
```diff
@@ -93,6 +94,11 @@ user opens chrome headlessly
     Call Method    ${c_opts}    add_argument    ignore-certificate-errors
     Call Method    ${c_opts}    add_argument    log-level\=3
     Call Method    ${c_opts}    add_argument    disable-logging
+    Call Method    ${c_opts}    add_argument    download.default_directory=${DOWNLOADS_DIR}
+
+    ${prefs}=    Create Dictionary
```
luke-h1, line 99: Looks like I get the following error due to this when running this test against dev (using 3.10 & the latest chromedriver): 

``` 
Calling method 'add_argument' failed: TypeError: ArgOptions.add_argument() got an unexpected keyword argument 'download.default_directory'
```

tests/robot-tests/tests/libs/file_operations.py
```diff
@@ -1,3 +1,5 @@
+
+from genericpath import isfile
```
luke-h1, line 2: Looks like this import is unused

tests/robot-tests/tests/libs/file_operations.py
```diff
@@ -43,3 +45,12 @@ def zip_should_contain_file(zipfilename, filename):
...
         raise AssertionError(f'File "{filename}" not found in "{zipfilename}", which contains {str(files_in_zip)}')
+
+
+def zip_should_contain_directories_and_files(zipfilename: str, expected_files: list[str]):
+    zip = zipfile.ZipFile(f'test-results/downloads/{zipfilename}')
+    files = zip.namelist()
+    files = [file for file in files if not file.endswith(os.sep)]
+
+    if sorted(files) != sorted(expected_files):
+        raise AssertionError(f"expected_files didn't match files: files: '{files}', expected_files: '{expected_files}'")
```
luke-h1, line 56: think we can remove the duplicate `files:` in the assertion exception here

tests/robot-tests/tests/admin_and_public/bau/data_catalogue.robot
```diff
@@ -135,3 +143,8 @@ Check page displays 'This is not the latest data' tag
 Select new subject and download new subject file
     user clicks checkbox    ${SUBJECT_NAME_1} (csv, 456 Kb)
     user clicks button    Download selected files
+
+Validate new zip contains correct files
+    sleep    5    # wait for file to download
```
luke-h1, line 148: we potentially may want to pass a timeout to the `zip should contain directories and files` instead of setting a hard-coded sleep in order to avoid transient errors when this is running in a CI environment. We would then raise an assertion error inside `zip should contain directories and files` if the file isn't downloaded within the expected time

tests/robot-tests/tests/admin_and_public/bau/data_catalogue.robot
```diff
@@ -115,12 +115,20 @@ Navigate to Next step
...
 Check step 3 displays correctly
     user waits until page contains    Choose files to download    10
+
     user checks element contains    testid:choose-files    This is the latest data
 
 Check checkbox and download file
     user clicks checkbox    UI test subject 3 (csv, 17 Kb)
     user clicks button    Download selected files
 
+# Check zip file for 2021 here
```
luke-h1, line 125: Think this comment can be removed

luke-h1 (changes requested):

luke-h1:
Think you'll need to rebase from dev in order to resolve the above conflicts: 
``` 
git fetch && git rebase origin/dev
# resolve conflicts 
git rebase --continue 
git push -u origin EES-3421 --force 
``` 
After that's done we'll probably want to clean up the commits to keep the commit history clean: 
``` 
git log 
git reset --soft HEAD~{number of commits to go back to first commit}
git add -A
git commit -am "commit msg"
git push -u origin EES-3421 --force

```

tests/robot-tests/tests/admin_and_public/bau/data_catalogue.robot
luke-h1: On second thoughts, we probably want to keep the sleep here but 15 seconds is a more lenient time, especially when testing against slower environments.

PR #3592, opened 2022-11-01
🐛 EES-3412 Attempt to get Notifier to use sku config by changing apiVersion

benoutram:
After deploying https://github.com/dfe-analytical-services/explore-education-statistics/pull/3590 I noticed that the Notifier service plan has not changed in the Dev environment. The plan was expected to change from S1 to B1.

I notified that it has a different `apiVersion` to the other two function projects.

This PR changes the `apiVersion` of the function apps to all be `2022-03-01`.

### DIFF
--- a/infrastructure/templates/template.json
+++ b/infrastructure/templates/template.json
@@ -1421,7 +1421,7 @@
         "Name": "[first(split(parameters('skuData'), ' '))]"
       },
       "name": "[variables('dataPlanName')]",
-      "apiVersion": "2015-08-01",
+      "apiVersion": "2022-03-01",
       "location": "[resourceGroup().location]",
       "tags": {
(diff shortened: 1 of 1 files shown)


markhiveit (approved):

luke-h1 (commented):
Just wondering if we want to keep the `apiVersion` in sync with other app services & whatnot so that we have access to the latest features across different app services/keep everything on the same API version?

benoutram:
> Just wondering if we want to keep the `apiVersion` in sync with other app services & whatnot so that we have access to the latest features across different app services/keep everything on the same API version?

@luke-h1 Good point, I was trying not to affect too many services, but have now made this change so let's see how it goes. Thanks.

luke-h1 (approved):

PR #3593, opened 2022-11-01
EES-2867 handle details open state

amyb-hiveit:
This changes the Details component to work correctly as a React component. Due to a known issue in React this requires preventing the default browser behaviour  so the open state can be controlled. See https://github.com/facebook/react/issues/15486

This fixes a specific bug on the filters step of the table tool where you couldn't open a filter group had an error. 

Also updated the tests to check for the visibility of the content as well as the aria attributes as they were passing incorrectly before.

The expected behaviour when you submit the filters form without selecting filters is:

If the filter group has a total option:
- total is selected
- no error is shown

If the filter group doesn't have a total option:
- an error is shown
- the group is expanded (including if it was closed before the form was submitted)
- you cannot close the group 
- when you select an option in the group the error should disappear and you can close/open the group as normal.

### DIFF
--- a/src/explore-education-statistics-common/src/components/Details.tsx
+++ b/src/explore-education-statistics-common/src/components/Details.tsx
@@ -3,23 +3,14 @@ import useToggle from '@common/hooks/useToggle';
 import findAllParents from '@common/utils/dom/findAllParents';
 import { formatTestId } from '@common/utils/test-utils';
 import classNames from 'classnames';
-import React, {
-  MouseEvent,
-  ReactNode,
-  useEffect,
-  useRef,
-  useState,
-} from 'react';
+import React, { ReactNode, useEffect, useRef, useState } from 'react';
 import styles from './Details.module.scss';
 import VisuallyHidden from './VisuallyHidden';
 
 let hasNativeDetails: boolean;
 let idCounter = 0;
 
-export type DetailsToggleHandler = (
-  isOpen: boolean,
-  event: MouseEvent<HTMLElement>,
-) => void;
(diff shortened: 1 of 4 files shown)


src/explore-education-statistics-common/src/components/Details.tsx
```diff
@@ -98,25 +97,20 @@ const Details = ({
...
-
-            if (event.isDefaultPrevented()) {
-              return;
-            }
+          // preventDefault here to allow the open state to be controlled,
+          // see https://github.com/facebook/react/issues/15486
+          event.preventDefault();
+          if (!preventToggle) {
+            onToggle?.(!open);
+            toggleOpen(!open);
```
ntsim, line 112: Could just call `toggleOpen` without any args to achieve the same thing?

src/explore-education-statistics-common/src/components/Details.tsx
```diff
@@ -71,8 +64,14 @@ const Details = ({
...
-    setOpened(open);
-  }, [open, setOpened]);
+    toggleOpen(defaultOpen);
+  }, [defaultOpen, toggleOpen]);
+
+  useEffect(() => {
+    if (preventToggle) {
+      toggleOpen(true);
+    }
+  }, [preventToggle, toggleOpen]);
```
ntsim, line 74: Would merge these into a single effect. If the parent component changes the `preventToggle` and `open` props, these would probably interfere with one another.

ntsim (changes requested):

ntsim (approved):

PR #3600, opened 2022-11-02
EES-3854 - remove deprecated SetCompatibilityVersion noop from startu…

luke-h1:
This PR: 
* removes a deprecated no-op function from Admin, content & data startup classes. Since Dotnet 3, this function is a no-op & has later been deprecated in Dotnet 6

Other changes:
* Seem to have modified the pipfile. Looks like this may be due to some leftover work in https://dfedigital.atlassian.net/browse/EES-3851

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,11 +1,11 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "a591778cf707b35ec7cfff3e47e24c8f630c14398e8aa78fec4225dd21c8a2f4"
+            "sha256": "80676ff8d37ab0d57717a6ddcf9f5f601e2071b3eadc0a6a348110531c7de91e"
         },
         "pipfile-spec": 6,
         "requires": {
-            "python_version": "3.11.0"
+            "python_version": "3.10.0"
         },
         "sources": [
             {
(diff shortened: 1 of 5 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Api/Startup.cs
```diff
@@ -68,8 +67,7 @@ public void ConfigureServices(IServiceCollection services)
                     options.Filters.Add(new OperationCancelledExceptionFilter());
                     options.EnableEndpointRouting = false;
                 })
-                .SetCompatibilityVersion(CompatibilityVersion.Version_3_0)
-                .AddNewtonsoftJson(options => {
+               .AddNewtonsoftJson(options => {
```
benoutram, line 70: I think there's a slight formatting issue here

```
            services.AddMvc(options =>
                {
                    options.Filters.Add(new OperationCancelledExceptionFilter());
                    options.EnableEndpointRouting = false;
                })
                .AddNewtonsoftJson(options =>
                {
                    options.SerializerSettings.NullValueHandling = NullValueHandling.Ignore;
                });
```

Pipfile.lock
```diff
@@ -1,11 +1,11 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "a591778cf707b35ec7cfff3e47e24c8f630c14398e8aa78fec4225dd21c8a2f4"
+            "sha256": "80676ff8d37ab0d57717a6ddcf9f5f601e2071b3eadc0a6a348110531c7de91e"
```
benoutram, line 4: The changes in this file aren't related to this change in any way I don't think? 

I guess there's no harm leaving it in this PR this though.

luke-h1: Yep, they're not. I've just included it as part of this PR since anyone doing a `pipenv install` would include this file as part of their PR

benoutram (approved):

PR #3603, opened 2022-11-03
EES-2419 - add missing template ID parameters in template.json

luke-h1:
This PR:
* adds a missing param to the `template.json` needed for a new notify template ID


benoutram (approved):

benoutram (changes requested):
Should it be pascal case like the others?

benoutram (approved):

PR #3608, opened 2022-11-07
EES-3867 - increase prod stats DB size by 300GB

luke-h1:
This PR: 
* Increase the prod statistics DB size by 300GB. I've used the following sum to calculate this: 

```
1073741824 x 300
``` 

References: 
https://learn.microsoft.com/en-us/azure/azure-sql/database/single-database-scale?view=azuresql#vcore-based-purchasing-model

### DIFF
--- a/infrastructure/parameters/prod.parameters.json
+++ b/infrastructure/parameters/prod.parameters.json
@@ -84,7 +84,7 @@
       "value": 2147483648
     },
     "maxStatsDbSizeBytes": {
-      "value": 697932185600
+      "value": 1020054732800
     }
   }
 }


infrastructure/parameters/prod.parameters.json
```diff
@@ -84,7 +84,7 @@
       "value": 2147483648
     },
     "maxStatsDbSizeBytes": {
-      "value": 697932185600
+      "value": 912680550400
```
ntsim, line 87: Not sure if this actually equates to a 300GB increase?

The old value of 697932185600 (650GB) + 322122547200 (300 GB) = 1020054732800

luke-h1: Not too certain.  I've made this change though. To get to `912680550400 ` I used `1073741824 x 300` & then added the initial value of `650GB` in bytes to it.

ntsim (approved):

PR #3614, opened 2022-11-08
EES-2946 reference line label position

amyb-hiveit:
Adds an option to chart reference lines to be able to set the position of the label along the line.

- On reference lines on the major axis the position on the line is set as a value on the other axis (so if X is the major axis you set a value on the Y axis as the vertical position of the label).
- On reference lines on the minor axis the position on the line is set as a percentage of other axis length (so if Y is the minor axis you set a percentage along the x axis to position the label at, with 0% being the start of the axis and 100% the end).  I considered selecting a value on the major axis for lines on the minor axis but that doesn't work so well as you might not have many specific points to choose from on the other axis so wouldn't be able to make very fine-grained adjustments to the position.
- If no position is set the labels default to be in the middle of the line.

reflines3

### Major axis config
reflines1

### Minor axis config
reflines2

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartReferenceLinesConfiguration.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartReferenceLinesConfiguration.tsx
@@ -4,8 +4,7 @@ import FormFieldSelect from '@common/components/form/FormFieldSelect';
 import FormFieldTextInput from '@common/components/form/FormFieldTextInput';
 import Tooltip from '@common/components/Tooltip';
 import {
-  AxisType,
-  ChartDefinition,
+  ChartDefinitionAxis,
   ReferenceLine,
   ReferenceLineStyle,
 } from '@common/modules/charts/types/chart';
@@ -20,14 +19,14 @@ import React, { useMemo } from 'react';
 
 interface AddFormValues {
   label: string;
+  otherAxisPosition?: number;
   position: string | number;
   style: ReferenceLineStyle;
 }
 
 export interface ChartReferenceLinesConfigurationProps {
-  axisType: AxisType;
+  axisDefinition?: ChartDefinitionAxis;
   dataSetCategories: DataSetCategory[];
-  definition: ChartDefinition;
   id: string;
   lines: ReferenceLine[];
   onAddLine: (line: ReferenceLine) => void;
@@ -35,16 +34,17 @@ export interface ChartReferenceLinesConfigurationProps {
 }
 
 export default function ChartReferenceLinesConfiguration({
-  axisType,
+  axisDefinition,
   dataSetCategories,
-  definition,
   id,
   lines,
   onAddLine,
   onRemoveLine,
 }: ChartReferenceLinesConfigurationProps) {
+  const { axis, referenceLineDefaults, type } = axisDefinition || {};
+
   const options = useMemo<SelectOption[]>(() => {
-    if (axisType === 'minor') {
+    if (type === 'minor') {
       return [];
     }
 
@@ -52,7 +52,7 @@ export default function ChartReferenceLinesConfiguration({
       label: filter.label,
       value: filter.value,
     }));
-  }, [axisType, dataSetCategories]);
+  }, [type, dataSetCategories]);
 
   const filteredOptions = useMemo<SelectOption[]>(() => {
     return options.filter(option =>
@@ -61,7 +61,7 @@ export default function ChartReferenceLinesConfiguration({
   }, [lines, options]);
 
   const referenceLines = useMemo(() => {
-    if (axisType === 'major') {
+    if (type === 'major') {
       return (
         lines?.filter(line =>
           options.some(option => option.value === line.position),
@@ -70,17 +70,15 @@ export default function ChartReferenceLinesConfiguration({
     }
 
     return lines ?? [];
-  }, [axisType, lines, options]);
+  }, [type, lines, options]);
 
   const getPositionLabel = (position: string | number) => {
-    if (axisType === 'major') {
+    if (type === 'major') {
       return options.find(option => option.value === position)?.label;
     }
 
     return position;
   };
-
-  const axisDefinition = definition.axes[axisType];
 
   return (
     <table className="govuk-table">
@@ -88,7 +86,20 @@ export default function ChartReferenceLinesConfiguration({
       <thead>
         <tr>
           <th>Position</th>
-          <th>Label</th>
+          <th>
+            {axis === 'x' ? 'Y axis position' : 'X axis position'}
+            {type === 'minor' && (
+              <>
+                {' '}
+                (%)
+                <br />
+                <span className="govuk-!-font-weight-regular">
+                  0 = start of axis
+                </span>
+              </>
+            )}
+          </th>
+          <th className="govuk-!-width-one-third">Label</th>
           <th>Style</th>
           <th className="dfe-align--right">Actions</th>
         </tr>
@@ -97,7 +108,8 @@ export default function ChartReferenceLinesConfiguration({
         {referenceLines.map(line => (
           <tr key={`${line.label}_${line.position}`}>
             <td>{getPositionLabel(line.position)}</td>
-            <td>{line.label}</td>
+            <td>{line.otherAxisPosition}</td>
+            <td className="govuk-!-width-one-third">{line.label}</td>
             <td>{upperFirst(line.style)}</td>
             <td>
               <Button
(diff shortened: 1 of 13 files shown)


src/explore-education-statistics-common/src/modules/charts/components/utils/getReferenceLineLabelPosition.ts
```diff
@@ -0,0 +1,47 @@
...
+  if (otherAxisPosition === undefined || !otherAxisDomain) {
+    return {
+      xPosition: defaultXPosition,
+      yPosition: defaultYPosition,
+    };
+  }
+
+  // Amount to offset from the top on horizontal bar charts so the label is visible when positioned at 100%
+  const yOffset = 15;
+  const yPosition = y + height * ((maxValue - otherAxisPosition) / maxValue);
```
ntsim, line 33: Seems like we need some x offset as well when setting a reference line on the minor axis to 100%. It's possible to get the following:

image

src/explore-education-statistics-common/src/modules/charts/components/utils/getReferenceLineLabelPosition.ts
```diff
@@ -0,0 +1,47 @@
...
+  axisType: AxisType;
+  otherAxisDomain?: [AxisDomain, AxisDomain];
+  otherAxisPosition?: number;
+  viewBox?: ViewBox;
+}) {
+  const { height = 0, width = 0, x = 0, y = 0 } = viewBox ?? {};
+  const defaultXPosition = axis === 'x' ? x : width / 2 + x;
+  const defaultYPosition = axis === 'y' ? y : height / 2 + y;
+  // otherAxisPosition is set as a percentage on minor axis lines so max is 100.
+  const maxValue = axisType === 'minor' ? 100 : (otherAxisDomain[1] as number);
```
ntsim, line 21: Would do a check to ensure that `otherAxisDomain[1]` is a number before casting it. 

We'll need to handle it appropriately if it isn't. Not sure if this means we want to throw an error or change it in some way, will leave that for you to decide :wink:

src/explore-education-statistics-common/src/modules/charts/components/utils/__tests__/getReferenceLineLabelPosition.test.ts
```diff
@@ -0,0 +1,121 @@
...
+      },
+    });
+
+    expect(result).toEqual({
+      xPosition: 205,
+      yPosition: 135,
+    });
+  });
+
+  test('returns the default position for an Y axis line when otherAxisPosition is not set', () => {
```
ntsim, line 23: Typo - 'for a Y axis'

src/explore-education-statistics-common/src/modules/charts/components/utils/getReferenceLineLabelPosition.ts
```diff
@@ -0,0 +1,47 @@
...
+  otherAxisDomain = [0, 0],
+  otherAxisPosition,
+  viewBox,
+}: {
+  axis: Axis;
+  axisType: AxisType;
+  otherAxisDomain?: [AxisDomain, AxisDomain];
+  otherAxisPosition?: number;
+  viewBox?: ViewBox;
+}) {
```
ntsim, line 16: Would specify an explicit return type like:
```tsx
interface LabelPosition {
  x: number;
  y: number;
}
```

Note that I'm also suggesting we simplify the property names for brevity. No need to postfix these with 'position'.

src/explore-education-statistics-common/src/modules/charts/components/utils/__tests__/getReferenceLineLabelPosition.test.ts
```diff
@@ -0,0 +1,121 @@
...
+        height: 250,
+      },
+    });
+
+    expect(result).toEqual({
+      xPosition: 424,
+      yPosition: 200,
+    });
+  });
+
```
ntsim, line 81: Would suggest a few more tests for the minor x axis label:

- When position is greater than 100%
- When position is less than 0%

src/explore-education-statistics-common/src/modules/charts/components/utils/__tests__/getReferenceLineLabelPosition.test.ts
```diff
@@ -0,0 +1,121 @@
...
+        height: 0,
+      },
+    });
+
+    expect(result).toEqual({
+      xPosition: 256,
+      yPosition: 158,
+    });
+  });
+});
```
ntsim, line 121: Would suggest a few more tests for the minor y axis label:

- When position is greater than 100%
- When position is less than 0%

src/explore-education-statistics-common/src/modules/charts/components/utils/__tests__/getReferenceLineLabelPosition.test.ts
```diff
@@ -0,0 +1,121 @@
...
+        height: 230,
+      },
+    });
+
+    expect(result).toEqual({
+      xPosition: 205,
+      yPosition: 66,
+    });
+  });
+
```
ntsim, line 61: Would suggest a few more tests for the major x axis label:

- When position is greater than the max
- When position is less than than the min

src/explore-education-statistics-common/src/modules/charts/components/utils/__tests__/getReferenceLineLabelPosition.test.ts
```diff
@@ -0,0 +1,121 @@
...
+        height: 0,
+      },
+    });
+
+    expect(result).toEqual({
+      xPosition: 432,
+      yPosition: 218,
+    });
+  });
+
```
ntsim, line 101: Would suggest a few more tests for the major y axis label:

- When position is greater than the max
- When position is less than than the min

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartReferenceLinesConfiguration.tsx
```diff
@@ -156,7 +168,21 @@ export default function ChartReferenceLinesConfiguration({
                     />
                   )}
                 </td>
-                <td className="dfe-vertical-align--bottom">
+                <td>
+                  <FormFieldNumberInput
+                    name="otherAxisPosition"
+                    id={`${id}-referenceLines-otherAxisPosition`}
```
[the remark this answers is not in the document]
amyb-hiveit: Good spot! Sorted now.

src/explore-education-statistics-common/src/modules/charts/components/utils/getReferenceLineLabelPosition.ts
amyb-hiveit: This is now handled using parseNumber in CustomReferenceLineLabel, that'll set it to undefined if not a number, then in getReferenceLineLabelPosition it'll return the default position if either domain value is undefined.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartReferenceLinesConfiguration.tsx
```diff
@@ -134,7 +136,34 @@ export default function ChartReferenceLinesConfiguration({
             }}
             validationSchema={Yup.object<AddFormValues>({
               label: Yup.string().required('Enter label'),
-              position: Yup.string().required('Enter position'),
+              otherAxisPosition: Yup.number().test({
+                name: 'otherAxisPosition',
+                message: 'Enter a valid position',
+                test: (value: number) => {
+                  if (!value) {
```
ntsim, line 143: If the value is 0, this validation will pass, even if it's actually out of bounds on the other axis. Note that users can change the minor axis' min/max, so we need to cover the possibility of this not being 0.

I'm guessing this is supposed to cover the 'blank' case, so we should really do something like this instead:

```ts
test: (value?: number) {
  if (typeof value !== 'number) {
    return true;
  }

  // do validation...
}
```

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartReferenceLinesConfiguration.tsx
```diff
@@ -134,7 +136,34 @@ export default function ChartReferenceLinesConfiguration({
             }}
             validationSchema={Yup.object<AddFormValues>({
               label: Yup.string().required('Enter label'),
-              position: Yup.string().required('Enter position'),
+              otherAxisPosition: Yup.number().test({
+                name: 'otherAxisPosition',
+                message: 'Enter a valid position',
```
ntsim, line 141: Think we could be a little clearer than this as it's not going to be super obvious why this isn't valid. Perhaps something like 'Enter a position within the Y axis min/max range' would be better? If there's something more succinct, happy to use that.

With longer messages, we'll might want to style the 'Add line' button to avoid it squashing like this:

image

We might want to set a custom width for the 'Actions' column?

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartReferenceLinesConfiguration.tsx
```diff
@@ -134,7 +136,34 @@ export default function ChartReferenceLinesConfiguration({
...
+                    ? value >= minorAxisDomain?.min &&
+                        value <= minorAxisDomain.max
+                    : true;
+                },
+              }),
+              position: Yup.string()
+                .required('Enter position')
+                .test({
+                  name: 'axisPosition',
+                  message: 'Enter a valid position',
```
ntsim, line 159: Same as the comment on the `otherPosition` validation, we could use a clearer validation message here

ntsim (changes requested):

ntsim (approved):

PR #3616, opened 2022-11-08
EES-3661 Add database free-text search support for publications

benoutram:
This PR:

- Changes the SQL Server docker image we use in Docker Compose to a Docker file which is based on the same `2019-latest ` image but installs the SQL Server full-text search component.

- Adds a full-text catalog to the database with an index on the publications `Title` and `Summary` columns.

- Adds support to EF core for free-text searching publications by mapping a queryable function to a database table-valued function called `PublicationsFreeTextTable` which in turn uses SQL Server built-in function `FREETEXTTABLE`.

Unlike it's non-table equivalent `FREETEXT` which is mapped by `EF.Functions.FreeText`, there's no EF Core support for `FREETEXTTABLE`. I managed to map `FREETEXTTABLE` in EF Core using code like  `modelBuilder.HasDbFunction(FreeTextTable).IsBuiltIn(true);` but I couldn't find a way of setting the table and column parameters. The workaround has been to add the custom table-valued function `PublicationsFreeTextTable` which wraps `FREETEXTTABLE`. See [Mapping a  queryable function to a table-valued function](https://learn.microsoft.com/en-us/ef/core/querying/user-defined-function-mapping#mapping-a-queryable-function-to-a-table-valued-function) for further information.

These 3 changes will allow us to perform a free text search over the publications:

### Example usage returning Id, Rank

```
[HttpGet("publications")]
public async Task<ActionResult<List<FreeTextRank>>> PublicationsFreeText([FromQuery] string search)
{
    return await _contentDbContext.PublicationsFreeTextTable(search).ToListAsync();
}
```

which translates FROM LINQ to SQL as

```
SELECT p.Id, p.Rank FROM PublicationsFreeTextTable(@search) AS p
```

and produces

```
HTTP/1.1 200 OK
[
  {
    "id": "c8756008-ed50-4632-9b96-01b5ca002a43",
    "rank": 156
  },
  {
    "id": "bfdcaae1-ce6b-4f63-9b2b-0a1f3942887f",
    "rank": 73
  }
]
```

### Example usage returning Id, Title, ordered by Rank

```
[HttpGet("publications")]
public async Task<ActionResult<List<PublicationTitleViewModel>>> PublicationsFreeText([FromQuery] string search)
{
    return await _contentDbContext.Publications
        .Join(_contentDbContext.PublicationsFreeTextTable(search),
            publication => publication.Id,
            freeTextRank => freeTextRank.Id,
            (publication, freeTextRank) => new { Publication = publication, freeTextRank.Rank })
        .OrderByDescending(tuple => tuple.Rank)
        .Select(tuple => new PublicationTitleViewModel
        {
            Id = tuple.Publication.Id,
            Title = tuple.Publication.Title
        })
        .ToListAsync();
}
```

which translates from LINQ to SQL as

```
SELECT p.Id, p.Title, ftt.Rank FROM Publications AS p
INNER JOIN PublicationsFreeTextTable(@search) AS ftt ON p.Id = ftt.Id
ORDER BY ftt.Rank DESC
```

and produces

```
HTTP/1.1 200 OK
[
  {
    "id": "c8756008-ed50-4632-9b96-01b5ca002a43",
    "title": "Revised GCSE and equivalent results in England"
  },
  {
    "id": "bfdcaae1-ce6b-4f63-9b2b-0a1f3942887f",
    "title": "GCSE and equivalent results"
  }
]
```


src/GovUk.Education.ExploreEducationStatistics.Admin/GovUk.Education.ExploreEducationStatistics.Admin.csproj
```diff
@@ -201,7 +201,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Folder Include="Migrations\ContentMigrations" />
     <Folder Include="Migrations\UsersAndRolesMigrations" />
```
duncan-at-hiveit, line 204: Was this file change intentional?

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20221108151343_EES3661_AddPublicationsFreeTextTableFunction.cs
```diff
@@ -0,0 +1,28 @@
...
+public partial class EES3661_AddPublicationsFreeTextTableFunction : Migration
+{
+    private const string MigrationId = "20221108151343";
+
+    protected override void Up(MigrationBuilder migrationBuilder)
+    {
+        // Add a database table-valued function wrapping the SQL Server built-in function freetexttable.
+        // This can be mapped to a queryable function in EF Core
+        migrationBuilder.SqlFromFile(ContentMigrationsPath, $"{MigrationId}_Routine_PublicationsFreeTextTable.sql");
+        migrationBuilder.Sql("GRANT SELECT ON dbo.PublicationsFreeTextTable TO [content]");
```
duncan-at-hiveit, line 20: Nice!

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ModelBuilderExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+    private static readonly MethodInfo? PublicationsFreeTextTableMethodInfo = typeof(ContentDbContext)
+        .GetMethod(nameof(ContentDbContext.PublicationsFreeTextTable),
+            new[] { typeof(string) });
+
+    public static ModelBuilder AddFreeTextTableSupport(this ModelBuilder modelBuilder)
+    {
+        modelBuilder.Entity<FreeTextRank>().HasNoKey().ToTable((string?) null);
+
+        // Map the PublicationsFreeTextTable queryable function to database table-valued function
+        modelBuilder.HasDbFunction(PublicationsFreeTextTableMethodInfo!);
```
duncan-at-hiveit, line 18: Some NASA-grade programming tricks here?

src/mssql-server/Dockerfile-mssql-server
```diff
@@ -0,0 +1,17 @@
+﻿# Base OS layer: Latest Ubuntu LTS
```
duncan-at-hiveit, line 1: Can't think of a reason why this wouldn't work on Linux, but might be worth getting a guinea pig to trial it before merging

benoutram: Ah, yep, I hadn't thought about that, thanks. Would it be possible for you to give it a try?

If you switch to this branch and `docker compose up db`.

If you execute this on the database and then get a result of '1' it's enabled:

```
SELECT fulltextserviceproperty('IsFulltextInstalled')
```

If you run our Admin service which will run the migrations on this branch the free text catalog with column indexes will be added. You should then be able to run a query like

```
SELECT Summary, Title
FROM dbo.Publications
WHERE FREETEXT ((Summary, Title), 'GCSE OR Social OR Outcomes OR Graduate')
```

If you want to go even further and try it out from EF Core there's some quick and easy code examples in the description of this PR which can be pasted into any controller in the Content API.

Thanks! 🙂

src/GovUk.Education.ExploreEducationStatistics.Admin/GovUk.Education.ExploreEducationStatistics.Admin.csproj
benoutram: Rider seems to remove this as soon as I add a new file to the directory. It doesn't matter whether that's a c# `.cs` file or any other type of file.

If I add a new directory I get a new `ItemGroup/Folder` entry in the `.csproj` file up until the point I add files to the directory. Presumably this is to stop it from being lost without any files existing under version control in Git if the project was to be restored from a clean checkout.

Do you know if we need to keep it for any reason?

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ModelBuilderExtensions.cs
benoutram: `PublicationsFreeTextTable` is the signature of a user-defined CLR function. EF Core only cares about the method signature not the body, since it will only be evaluated on the database, not client side: `public IQueryable<FreeTextRank> PublicationsFreeTextTable(string search) =>FromExpression(() => PublicationsFreeTextTable(search));`

`PublicationsFreeTextTableMethodInfo` is the `MethodInfo` used to map that function to the database function I've created with the same name.

```
private static readonly MethodInfo? PublicationsFreeTextTableMethodInfo = typeof(ContentDbContext)
    .GetMethod(nameof(ContentDbContext.PublicationsFreeTextTable),
        new[] { typeof(string) });
```

It's registered with `modelBuilder.HasDbFunction(PublicationsFreeTextTableMethodInfo!);`.

EF core expects to find the database function with the same name. If the name is not the same, or it's in a different schema, or the function is built-in to the database, you can call the methods `.HasName('...')`, `.HasSchema('...')`, `.IsBuiltIn(true)`.

I could have in-lined the configuration without using `MethodInfo` in which case it would have looked like this:

```
modelBuilder.HasDbFunction(typeof(ContentDbContext).GetMethod(nameof(PublicationsFreeTextTable), new[] { typeof(string) })!);
```

There's more info about doing this here, see [User-defined function mapping](https://learn.microsoft.com/en-us/ef/core/querying/user-defined-function-mapping).

src/GovUk.Education.ExploreEducationStatistics.Admin/GovUk.Education.ExploreEducationStatistics.Admin.csproj
duncan-at-hiveit: I've no idea I'm afraid - if it's not causing any issues and the migrations are being applied, I guess potentially we don't need it - should we remove the line underneath too?

src/mssql-server/Dockerfile-mssql-server
duncan-at-hiveit: All looks great!

duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/GovUk.Education.ExploreEducationStatistics.Admin.csproj
benoutram: Yes I think we can, I've removed the entries which I think are unnecessary now.

PR #3618, opened 2022-11-09
EES-3871 Fix broken key vault references to publisher client id/secret

ntsim:
This is fixed by adding the Key Vault resource's API version to the `reference` function call. As noted in the docs, this is required if the resource is provisioned outside of the template. See: https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-resource#reference


luke-h1 (approved):

ntsim:
Do not merge until performance testing is done

PR #3629, opened 2022-11-10
EES-3861 - increased number of running observation rows to delete

duncan-at-hiveit:


benoutram (commented):
Don't you want to merge https://github.com/dfe-analytical-services/explore-education-statistics/pull/3630 first and then merge the master branch back into dev, rather than make this change?

PR #3630, opened 2022-11-10
EES-3861 - increased number of running observation rows to delete in …

duncan-at-hiveit:


benoutram (approved):
Worth noting that if this runs at 7pm on Saturday and runs up to the timeout of 12 hours, it would overlap with the reindexing which starts at 2am on Sunday.

On the dev branch there's already a change to the schedule for deleting subjects, changing it to only occur Monday to Friday. This is because the schedule for reindexing is also changing to start earlier at 7pm on Saturday rather than 2am on Sunday, giving that up to 12 hours to run and still finish by the early morning.

PR #3634, opened 2022-11-11
EES-2959 Remove release invites when removing user release invites vi…

markhiveit:
…a platform admin

This PR solves a bug where an existing user couldn't be invited to have pre-release access on a release if a previous `UserReleaseRole` had been removed but the associated `UserReleaseInvite` hadn't been deleted. This situation would happen when you removed a user release role via platform administration. This PR fixes this bug by ensuring that user release invites also get removed when removing user release roles via platform administration.


ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserRoleService.cs
```diff
@@ -406,21 +409,26 @@ await _usersAndRolesPersistenceHelper
...
                 {
                     return await _contentPersistenceHelper
-                        .CheckEntityExists<Release>(query => query
-                            .Include(r => r.Publication)
-                            .Where(r => r.Id == role.ReleaseId)
+                        .CheckEntityExists<UserReleaseRole>(query => query
+                            .Include(userReleaseRole => userReleaseRole.User)
+                            .Include(userReleaseRole => userReleaseRole.Release)
+                            .ThenInclude(release => release.Publication)
+                            .Where(urr => urr.ReleaseId == role.ReleaseId)
```
ntsim, line 416: This doesn't look right. This would probably get the first UserReleaseRole that matches the `releaseId` (which there could be many). Additionally, there's another `CheckEntityExists` earlier on that already makes the call for the correct UserReleaseRole, so we don't need to repeat this work.

We should move these `Includes` into the initial `CheckEntityExists` i.e.

```cs
.CheckEntityExists<UserReleaseRoles>(userReleaseRoleId, 
    query => query.Includes(...)
)
```

ntsim (changes requested):

ntsim (approved):

PR #3636, opened 2022-11-11
EES-3835 Fix login redirect

markhiveit:
This PR fixes a bug where after logging in, a user wouldn't get redirected to page they were previously trying to access.

This was fixed by ensuring the argument name for the return URL was correct on `OnGetCallbackAsync`. It must be `returnUrl`, otherwise IdentityServer4 will always return null for that argument.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
@@ -25,7 +25,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Areas.Identity.Pages.
     public class ExternalLoginModel : PageModel
     {
         private const string LoginErrorMessage = "Sorry, there was a problem logging you in.";
-        
+
         private readonly ISignInManagerDelegate _signInManager;
         private readonly IUserManagerDelegate _userManager;
         private readonly ContentDbContext _contentDbContext;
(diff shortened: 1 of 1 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
```diff
@@ -68,15 +68,15 @@ public IActionResult OnPost(string provider, string returnUrl = null)
             return new ChallengeResult(provider, properties);
         }
 
-        public async Task<IActionResult> OnGetCallbackAsync(string providedReturnUrl = null, string remoteError = null)
+        public async Task<IActionResult> OnGetCallbackAsync(string returnUrl = null, string remoteError = null)
         {
-            var returnUrl = providedReturnUrl ?? Url.Content("~/");
+            var returnUrlOrFallback = returnUrl ?? Url.Content("~/");
```
ntsim, line 73: Not sure how this fixes the problem. It looks like we've just renamed some variables?

markhiveit: It's the renaming of the argument name to `returnUrl` that fixed it. Duncan noticed that newer version of IdentityServer, the `OnGetCallbackAsync` method's argument name wasn't `providedReturnUrl`, so we speculatively changed it to `returnUrl` and then redirection worked. Dunc suggested IdentityServer uses reflection or something.

ntsim: Wow that's just mad. Can you leave a comment on the method saying that the parameter names are important?

markhiveit: Done. I've also updated the PR description.

ntsim (approved):

PR #3639, opened 2022-11-14
EES-3784 - create permalink checker script

luke-h1:
This PR:
* creates a python script to check whether a permalink can be rendered or not on production. This aids in migrating to [permalink snapshots](https://dfedigital.atlassian.net/browse/EES-3754) 

**Structure**
* Currently the script gets data from a `prod-permalinks.csv` which it assumes already exists and is filled with permalinks via running the query included in `check_permalinks.py` 

* The script then outputs new permalink IDs into `checked-permalinks.csv` to ensure work is not repeated and it only checks permalinks that haven't already been tested 

* Any invalid permalinks that couldn't load or have error text (i.e. `There was a problem rendering the table.`) present are placed in `invalid-permalinks.csv` which stores the following details about the permalink that couldn't render: 

```
* permalink ID 
* subject name
* publication name
* response status code (if it can be fetched) 
* has table (boolean) 
* created date 
* out of date (boolean) 
* error message (i.e. 'table cannot be rendered') 
* warning message (i.e. 'out of date, superseeded')
* response_time (if it can be fetched)
  ```

Todo:

- [x] Find actual error & warning text within the warning/error div rather than grabbing the whole text in that element
- [x] Update .gitignore to include full path to csv files that need to be excluded
- [x] Add response_time field to csv

### DIFF
--- a/tests/robot-tests/scripts/permalink_snapshots/check_permalinks.py
+++ b/tests/robot-tests/scripts/permalink_snapshots/check_permalinks.py
@@ -1,0 +1,275 @@
+import csv
+import os
+import time
+
+import requests
+from bs4 import BeautifulSoup
+from requests import Response
+
+"""
+Script for checking the health of permalinks & whether they can be rendered.
+Some permalinks on production don't load i.e. https://explore-education-statistics.service.gov.uk/data-tables/permalink/aed871a3-61e5-4f2a-8a2b-cceeca680980
+This will cause issues when migrating to permalink snapshots (see https://dfedigital.atlassian.net/browse/EES-3784)
+
+Instructions:
+
+To generate the csv file of permalinks run the following query against the content DB:
+
+```
+SELECT *
+from [dbo].[Permalinks]
+```
+
+place these permalinks in a csv file in the same directory as '~/permalink-snapshots'
+with the filename 'prod-permalinks.csv'
+
+create a csv file with the name 'invalid-permalinks.csv' in the same directory as '~/permalink-snapshots'
+with the following headers:
+
+```
+permalink_id,subject_name,publication_name,status_code,has_table,created_date,out_of_date,error_message,warning_message,response_time
+```
+
+create a csv file with the name 'checked-permalinks.csv' in the same directory as '~/permalink-snapshots'
+with the following headers:
+
+```
+permalink_id, response_time
+```
+
+"""
+
+
+class PermalinkType:
+    id = str
+    publicationTitle = str
+    dataSetTitle = str
+    releaseId = str
+    subjectId = str
+    created = str
+
+
+requests.sessions.HTTPAdapter(pool_connections=100, pool_maxsize=100, max_retries=3)
+
+
+class PermalinkChecker:
+    def __init__(self):
+        # change 'expected_row_length' to the number of rows in the 'prod-permalinks.csv' file
+        self.expected_row_length = 23019
+        self.backoff = 1.75
+        self.request_timeout = 15
+        self.session = requests.Session()
+        self.prod_permalinks_csv = os.path.join(os.getcwd() + "/scripts/permalink_snapshots", "prod-permalinks.csv")
+        self.invalid_permalinks_csv = os.path.join(
+            os.getcwd() + "/scripts/permalink_snapshots", "invalid-permalinks.csv"
+        )
+        self.checked_permalinks_csv = os.path.join(
+            os.getcwd() + "/scripts/permalink_snapshots", "checked-permalinks.csv"
+        )
+        self.public_url = "https://explore-education-statistics.service.gov.uk"
+        self.permalink_base_url = f"{self.public_url}/data-tables/permalink/"
+
+        assert os.path.basename(os.getcwd()) == "robot-tests", "Must run from the robot-tests directory"
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/robot-tests/scripts/permalinks_snapshots/check_permalinks.py [hunk 1]
```diff
@@ -0,0 +1,195 @@
...
+                # should never happen as all permalinks should have a page title
+                # could be due to a massive data table in the permalink so add it to the csv
+                # so we can check manually
+
+                print(f"Writing {permalink_id} permalink to csv as it has no page title")
+                writer.writerow(
+                    [
+                        permalink_id,
+                        subject_name if subject_name else "NO SUBJECT NAME FOUND",
+                        publication_name if publication_name else "NO PUBLICATION NAME FOUND",
```
ntsim, line 138: Would just leave these as empty strings if nothing was found.

tests/robot-tests/scripts/permalinks_snapshots/check_permalinks.py
```diff
@@ -0,0 +1,195 @@
...
+                        response.url,
+                        response.status_code,
+                        has_table,
+                        created_date,
+                        permalink_out_of_date_message if has_permalink_out_of_date else "NO OUT-OF-DATE ERROR",
+                        permalink_error_message if has_permalink_error else "NO ERROR MESSAGE FOUND",
```
ntsim, line 144: Would just leave these as empty strings if nothing was found.

tests/robot-tests/scripts/permalinks_snapshots/check_permalinks.py
```diff
@@ -0,0 +1,195 @@
...
```
ntsim, line 160: Same as above - would just leave these as empty strings if nothing was found.

tests/robot-tests/scripts/permalinks_snapshots/check_permalinks.py
```diff
@@ -0,0 +1,195 @@
...
+            # should never happen as all permalinks should have a page title
+            print("No page title found")
+
+        # fmt off
+        subject_name = (
+            page_title.text.split("from")[0].strip() if page_title else "No subject title found in page title"
+        )
+
+        publication_name = (
+            page_title.text.split("from")[1].strip() if page_title else "No publication title found in page title"
```
ntsim, line 106: Would just leave these as empty strings if nothing was found.

tests/robot-tests/scripts/permalinks_snapshots/check_permalinks.py
```diff
@@ -0,0 +1,195 @@
...
+        return permalinks
+
+    def _build_permalink_urls(self):
+        permalinks = self._get_permalinks_from_csv()
+
+        return [f"{self.permalink_base_url}{permalink['id']}" for permalink in permalinks]
+
+    def _write_status_to_csv(self, response: Response, permalink_id: str):
+        permalink_error_message = "There was a problem rendering the table."
+        permalink_out_of_date_message = "WARNING - The data used in this table may now be out-of-date as a new release has been published since its creation."
```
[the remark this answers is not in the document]
luke-h1: Cool, good to know. I've switched things to use Selenium now as BS4 didn't interact well with big permalink pages so this shouldn't be to hard

tests/robot-tests/scripts/permalink_snapshots/checked-permalinks.csv
```diff
@@ -0,0 +1 @@
+permalink_id, response_time
```
duncan-at-hiveit, line 1: Do we want to .gitignore this file and also the resulting csvs from prepping and running the script?  Just not wanting anyone to accidentally check in all of the Permalink IDs from Prod into the codebase!

If we do this, I guess the script will need adjusting slightly just to add manual instructions to create this CSV, in the same way there are manual instructions to create `invalid-permalinks.csv`. 

Either that or we just make sure no-one checks in any files, but my preference would be the .gitignore route.

tests/robot-tests/scripts/permalink_snapshots/check_permalinks.py
```diff
@@ -0,0 +1,268 @@
...
+    dataSetTitle = str
+    releaseId = str
+    subjectId = str
+    created = str
+
+
+requests.sessions.HTTPAdapter(pool_connections=100, pool_maxsize=100, max_retries=3)
+
+
+class PermalinkChecker:
```
duncan-at-hiveit, line 48: Nice work :smiley:

tests/robot-tests/scripts/permalink_snapshots/checked-permalinks.csv
luke-h1: Yep, that sounds like a good idea. Wasn't sure on whether to exclude it or not

tests/robot-tests/scripts/permalink_snapshots/check_permalinks.py
luke-h1: ta, took a few tries to find the balance between the speed of the script vs the impact that had on the public frontend. Hopefully should be useful in getting a lay of the land for bad permalinks out there.

duncan-at-hiveit (approved):

PR #3640, opened 2022-11-15
EES-3799 - speed up Locations import and fix import bug with case-sensitive Filter Group and Filter Item creation

duncan-at-hiveit:
# Overview

This PR:
* addresses a bug whereby Filter Items and Filter Groups with a mix of case in a CSV was causing the importer to fail import
* speeds up School data import speed massively, by caching ALL Locations in the Importer during startup in a new ImporterLocationCache (addressing EES-3460)
* renames and refactors the ImporterMemoryCache to be ImporterFilterCache, as only Filter Groups and Filter Items get cached there now

## Case-sensitive Filter Group / Filter Item labels bug

During the recent Importer work for EES-3798 and EES-3799, a bug was accidentally introduced into the Importer process whereby Filter Groups and Filter Items of mixed case in a CSV was causing lookups of owning Filter Groups to pick the incorrect Filter Group (as SQL Server string comparison is case-insensitive by default).

The place where this was going wrong was when gathering all of the Filter Groups and Filter Items prior to saving them to the database during Stage 2 of the import process (in ImporterService.ImportFiltersAndLocations).

Now, we make the code which gathers all "unique" Filter Groups and Filter Items case insensitive, thus ending up with a single Filter Group or Filter Item where previously there may have been 2 or more (e.g. "Not applicable" versus "Not Applicable" in `school_blocks_randomised.csv`).

## Speeding up School data imports

The problem we have been facing is that the query that we have to run against the database to determine whether or not a Location from the CSV exists already in the database is a complex one, and is made especially bad with school data, which can contain thousands or tens of thousands of rows (22,000+ in `school_blocks_randomised.csv`).  Add to this the fact that environments like Prod have close to 100,000 Locations to scan through.

The solution to this issue within this PR is to load all Locations into memory during Importer startup.  We then do existence checks in memory rather than in the database, cutting the Stage 2 time for import from 15 minutes down to 40 seconds.  There can be alternative approaches to improving the database lookup speed too but we have not yet prioritised scoping those out, and at the current time I think this should be sufficient.  We don't expect this list to grow massively beyond where it's at but should it ever do so, we can address this.

Locally, 22,500 Locations can be loaded into memory in about 1 second and consume about 30MB of additional memory for the Importer Function, meaning that it'd likely consume about 150MB of additional memory for Prod.  This gives the Importer plenty of memory headroom to play with, given that it was consuming GBs of memory before EES-3798 and EES-3799 were worked on.  This should be tested properly on Azure environments though to check the impact there.

## Refactoring ImporterMemoryCache to be ImporterFilterCache

Given that only Filter Items and Filter Groups were being put into ImporterMemoryCache now, it felt better to be more specific about its usage.  This meant that we could internalise the `GetFilterGroupCacheKey` and `GetFilterItemCacheKey` methods into the cache itself and additionally make specific getters and setters for Filter Groups and Filter Items, thus making the Importer code a bit clearer to follow.


src/GovUk.Education.ExploreEducationStatistics.Data.Processor.Tests/Services/ImporterLocationServiceTests.cs
```diff
@@ -72,18 +74,12 @@ await using (var statisticsDbContext = InMemoryStatisticsDbContext(statisticsDbC
                 await statisticsDbContext.SaveChangesAsync();
```
markhiveit, line 74: Do we need call LoadLocations here?

markhiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Data.Processor.Tests/Services/ImporterLocationServiceTests.cs
duncan-at-hiveit: Ooh well spotted!  Have added it in.

markhiveit (approved):

PR #3645, opened 2022-11-16
EES-3877 - fix bad methodology ids being generated in Accordion sections

luke-h1:
This PR:
- replaces any methodology/accordion section IDs that start with a number at the start with `section-` which causes this error. This resolves an error whereby methodology URLs with a hash (linking to accordion content) wouldn't open

```
DOMException: Failed to execute 'querySelector' on 'Element': '#6.test-title' is not a valid selector.
```

image

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/methodology/edit-methodology/content/components/MethodologyAccordionSection.tsx
+++ b/src/explore-education-statistics-admin/src/pages/methodology/edit-methodology/content/components/MethodologyAccordionSection.tsx
@@ -129,9 +129,8 @@ const MethodologyAccordionSection = ({
       {...props}
       anchorLinkUrl={
         editingMode === 'preview'
-          ? `${PublicAppUrl}/methodology/${
-              currentMethodology.slug
-            }#${heading.toLowerCase().split(' ').join('-')}`
+          ? id =>
+              `${PublicAppUrl}/methodology/${currentMethodology.slug}/#${id}`
           : undefined
       }
       heading={heading}
--- a/src/explore-education-statistics-common/src/components/Accordion.tsx
+++ b/src/explore-education-statistics-common/src/components/Accordion.tsx
@@ -64,7 +64,7 @@ const Accordion = ({
   );
 
   const { isMounted } = useMounted(() => {
-    const goToHash = async () => {
+    const goToAndOpenHash = async () => {
       setTimeout(() => {
         if (!ref.current || !window.location.hash) {
           return;
@@ -74,7 +74,9 @@ const Accordion = ({
 
         try {
           locationHashEl = ref.current.querySelector(window.location.hash);
-        } catch (_) {
+        } catch (e) {
+          // eslint-disable-next-line no-console
+          console.error(e);
           return;
         }
 
@@ -108,10 +110,10 @@ const Accordion = ({
       }, 200);
     };
 
-    goToHash();
-    window.addEventListener('hashchange', goToHash);
+    goToAndOpenHash();
+    window.addEventListener('hashchange', goToAndOpenHash);
 
-    return () => window.removeEventListener('hashchange', goToHash);
+    return () => window.removeEventListener('hashchange', goToAndOpenHash);
   });
 
   useEffect(() => {
--- a/src/explore-education-statistics-common/src/components/AccordionSection.tsx
+++ b/src/explore-education-statistics-common/src/components/AccordionSection.tsx
@@ -2,6 +2,7 @@ import CopyLinkButton from '@common/components/CopyLinkButton';
 import useMounted from '@common/hooks/useMounted';
 import findAllParents from '@common/utils/dom/findAllParents';
 import classNames from 'classnames';
+import kebabCase from 'lodash/kebabCase';
 import React, { createElement, memo, ReactNode } from 'react';
 import styles from './AccordionSection.module.scss';
 import GoToTopLink from './GoToTopLink';
@@ -18,7 +19,7 @@ export interface AccordionSectionProps {
         contentId: string;
       }) => ReactNode);
   className?: string;
-  anchorLinkUrl?: string;
+  anchorLinkUrl?: (id: string) => string;
   goToTop?: boolean;
   header?: ReactNode;
   heading: string;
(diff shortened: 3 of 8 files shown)


amyb-hiveit (dismissed):

src/explore-education-statistics-common/src/components/Accordion.tsx
```diff
@@ -108,10 +108,31 @@ const Accordion = ({
       }, 200);
     };
 
+    const openHash = () => {
```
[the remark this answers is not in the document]
luke-h1: Yep, the existing code works. After a while of discussing it with you & looking through `AccordionSection`, If figured out that replacing any heading ID that starts with a number and giving it a value such as `section-` fixed the issue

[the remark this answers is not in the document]
amyb-hiveit: I don't think this is quite right - testing on http://localhost:3000/methodology/permanent-and-fixed-period-exclusions-in-england adding the id eg '#section-.-national-statistics-badging' doesn't open the section. Maybe the '.' makes the id invalid?
Looking more closely at it I see that the buttons in the accordion h2s have an id on them based on the section number, so I wonder if this could be re-used instead? It would have to have something appended or removed from it so that it's not exactly the same as the button id.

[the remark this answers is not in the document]
luke-h1: Yeah, I've had a look at it does seem that the erroneous `.` makes the id invalid. I think that re-using the button-id with something such as a `-` to provide variation is the best option as you say. In the unlikely event that a methodology accordion has the same title, performing `.replace(/[0-9]/, 'section-'),` would end up results in non-unique IDs for different accordion sections

src/explore-education-statistics-common/src/components/AccordionSection.tsx
```diff
@@ -81,7 +81,17 @@ const AccordionSection = ({
...
+                .toLowerCase()
+                .split(' ')
+                .join('-')
+                /* 
+                workaround for bad IDs that start with a number from 
+                being generated which throws the following error 
+                DOMException: Failed to execute 'querySelector' on 'Element': '#6.test-title' is not a valid selector.
+              */
+                .replace(/[0-9]/, 'section-')
+                .replace(/\./g, '-'),
```
ntsim, line 94: Not a big fan of this. The above will give weird results if the heading contains numbers in non-starting positions e.g. 'Some title with 5 words' would give us `some-title-with-section-words`.

As mentioned in standup, we could just use Lodash's `kebabCase` to do this in a nice clean way:

```ts
id: `section-${kebabCase(heading)}`
```

Note that to avoid the id starting with a number, we just need to prefix the id with something generic like `section-`

src/explore-education-statistics-common/src/components/Accordion.tsx
```diff
@@ -64,7 +64,7 @@ const Accordion = ({
   );
 
   const { isMounted } = useMounted(() => {
-    const goToHash = async () => {
+    const goToAndOpenHash = async () => {
```
ntsim, line 67: Why rename this? This doesn't actually open anything so it's misleading to call it this.

ntsim (changes requested):

src/explore-education-statistics-common/src/components/AccordionSection.tsx
luke-h1: thanks, was quite bit distracted yesterday so forgot about the lodash solution. Looks like this is a cleaner fix & works much better than trying to write the perfect regex for this

src/explore-education-statistics-common/src/components/AccordionSection.tsx
```diff
@@ -81,7 +82,7 @@ const AccordionSection = ({
             headingTag,
             {
               className: classes.sectionHeading,
-              id: heading.toLowerCase().split(' ').join('-'),
+              id: `section-${kebabCase(heading)}`,
```
ntsim, line 85: As discussed in person, think there's a few things we need to sort out here:

- `anchorLinkUrl` should be changed to callback so that we can provide an id parameter to the consumer (so it doesn't have to figure out what the anchor ID should be) e.g.
  
  ```tsx
  <EditableAccordionSection 
    anchorLinkUrl={id => `${PublicAppUrl}/methodology/${currentMethodology.slug}#${id}`}
  >
  ```
- We should check with Cam if we human-readable IDs are a necessity. We might want to use guids instead as these are more stable (analysts can change section titles, thus breaking links) and they are unique (analysts may create sections with the same title and this would cause ambiguity).
- We should preferably use the `headingId` that is already set in `AccordionSection` instead of making a new id. This will be dependent on the previous point about human-readable ids.
- We'll need to update the public frontend accordingly (see `MethodologyPage`). Currently, it's using the section's `order` for the id of each, but we should really use the ID.

src/explore-education-statistics-common/src/components/AccordionSection.tsx
```diff
@@ -53,6 +55,7 @@ const AccordionSection = ({
   heading,
   headingTag = 'h2',
   id = 'accordionSection',
+  anchorLinkIdPrefix = 'section',
```
ntsim, line 58: Didn't say we should change the changes for `anchorLinkUrl`. Those should still be the same as they were previously i.e. a callback function

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/methodology/edit-methodology/content/components/MethodologyAccordionSection.tsx
```diff
@@ -129,9 +130,8 @@ const MethodologyAccordionSection = ({
       {...props}
       anchorLinkUrl={
         editingMode === 'preview'
-          ? `${PublicAppUrl}/methodology/${
-              currentMethodology.slug
-            }#${heading.toLowerCase().split(' ').join('-')}`
+          ? () => `${PublicAppUrl}/methodology/${currentMethodology.slug}
+          #content-section-${kebabCase(heading)}`
```
ntsim, line 134: This isn't correct, we want something like this:

```tsx
anchorLinkUrl={
  editingMode === 'preview' 
    ? (id) => `${PublicAppUrl}/methodology/${currentMethodology.slug}/#${id}`
    : undefined
}
```

ntsim (approved):

PR #3650, opened 2022-11-17
EES-3847 collapsible component accessibility fixes

amyb-hiveit:
Changes to the CollapsibleList component to fix an accessibility error that aria-hidden should not contain focusable elements.

Previously all the list items were rendered and a visually hidden class added to hide some of them off screen. Removing the aria-hidden on the button left it functionally useless for screen reader users (as the visually hidden items are read out), so I've changed the component to only render the visible items and them add in the rest when the button is clicked.

Also: 
- added aria-expanded and aria-controls attributes to (hopefully) improve the accessibility.
- changed the button text to provide more context

Collapsible list is used on:
- footnotes in data blocks
- table tool - inactive filters and location steps
- data file replacement plan in the admin

### Related change:
The Details component caused the same error when it contained links so I've removed the unnecessary aria-hidden here as well.


tests/robot-tests/tests/admin_and_public/bau/publish_data.robot
```diff
@@ -351,13 +351,13 @@ Create table
...
 Check created table has footnotes
-    user checks list has x items    testid:footnotes    3
+    user checks list has x items    testid:footnotes    2
     user checks list item contains    testid:footnotes    1    ${FOOTNOTE_ALL}
     user checks list item contains    testid:footnotes    2    ${FOOTNOTE_ALL_INDICATOR}
-    user checks list item contains    testid:footnotes    3    ${FOOTNOTE_ALL_FILTER}
 
-    user checks page contains button    Show 1 more item
-    user checks list item is visually hidden    testid:footnotes    3
+    user clicks button    Show 1 more footnote
```
duncan-at-hiveit, line 358: Aha so we have a test for this case, just in the UI tests :)

duncan-at-hiveit (commented):
Note that the front end build is failing on the pipeline for this branch at the mo too, if you wouldn't mind just seeing what could be happening there?

src/explore-education-statistics-common/src/components/__tests__/CollapsibleList.test.tsx
```diff
@@ -169,12 +226,37 @@ describe('CollapsibleList', () => {
       </CollapsibleList>,
     );
 
-    const button = container.querySelector('button');
+    userEvent.click(
+      screen.getByRole('button', { name: 'Show 2 more footnotes' }),
```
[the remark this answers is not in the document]
amyb-hiveit: It's covered by 'renders 4 of 5 items when collapseAfter is 4' but probably best to have an explicit test case for it so have added one.

src/explore-education-statistics-common/src/components/CollapsibleList.tsx
```diff
@@ -1,78 +1,72 @@
...
 }: Props) => {
   const [collapsed, toggleCollapsed] = useToggle(isCollapsed);
 
-  const listClasses = classNames('govuk-list', styles.listContainer, {
+  const itemNamePlural = initialItemNamePlural ?? `${itemName}s`;
+
+  const listItems = Children.toArray(children);
+
+  const renderedListItems = collapsed
+    ? listItems.splice(0, collapseAfter)
```
ntsim, line 32: Don't use `splice`, this is mutative and affects the original `listItems` variable. Use `slice` instead.

src/explore-education-statistics-common/src/components/FigureFootnotes.tsx
```diff
@@ -17,7 +17,12 @@ const FigureFootnotes = ({ footnotes, headingHiddenText }: Props) => {
           <VisuallyHidden>{` ${headingHiddenText}`}</VisuallyHidden>
         )}
       </h3>
-      <CollapsibleList listStyle="number" collapseAfter={2} testId="footnotes">
+      <CollapsibleList
+        listStyle="number"
+        collapseAfter={2}
+        id="footnotes"
```
ntsim, line 23: We can have many data blocks in release content, so this ID is almost certain to have collisions with other footnotes on the page.

ntsim (changes requested):

src/explore-education-statistics-common/src/components/CollapsibleList.tsx
```diff
@@ -1,78 +1,72 @@
...
+  itemName = 'item',
+  itemNamePlural: initialItemNamePlural,
   isCollapsed = true,
   listStyle = 'none',
-  testId,
 }: Props) => {
   const [collapsed, toggleCollapsed] = useToggle(isCollapsed);
 
-  const listClasses = classNames('govuk-list', styles.listContainer, {
+  const itemNamePlural = initialItemNamePlural ?? `${itemName}s`;
```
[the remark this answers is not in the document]
amyb-hiveit: I thought about making itemNamePlural required if you specify an item name but as in all existing cases where it's used adding 's' does work it seemed a bit unnecessary, so went with making itemNamePlural optional so you can override the default pluralisation if needed.

src/explore-education-statistics-common/src/components/CollapsibleList.tsx
amyb-hiveit: Good point, switched to slice now.

src/explore-education-statistics-common/src/components/FigureFootnotes.tsx
amyb-hiveit: Good spot! Sorted.

src/explore-education-statistics-common/src/components/CollapsibleList.tsx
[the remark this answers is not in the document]
ntsim: My only concern would be that some future person would potentially neglect to set the `itemNamePlural` based on the existing examples not setting it. Of course, testing should flag this up, but this could be easily missed if the collapse threshold isn't hit.

Personally, I think the minor inconvenience of having a type that enforces this is worth not having this come back as a bug in the future.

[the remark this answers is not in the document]
amyb-hiveit: Fair enough, I've changed it now.

ntsim (approved):

PR #3633, opened 2022-11-11
EES-3881 Optimise usages of Publication.LatestPublishedRelease() and remove it

benoutram:
⚠️ This depends on [EES-3882](https://dfedigital.atlassian.net/browse/EES-3882) added by https://github.com/dfe-analytical-services/explore-education-statistics/pull/3631 and also the migration of https://github.com/dfe-analytical-services/explore-education-statistics/pull/3656 being run which is described by [EES-3894](https://dfedigital.atlassian.net/browse/EES-3894).

This PR swaps all usages of `Publication.LatestPublishedRelease()` to use `Publication.LatestPublishedReleaseId` or `Publication.LatestPublishedRelease` and removes the computed method `Publication.LatestPublishedRelease()`.

This means that all former usages of it now don't depend on fetching back all releases (including all release versions) of a publication in order to determine its latest published release, the latest published release ID, or to check if a publication is published.

### UI Test report

image

PR #3656, opened 2022-11-21
EES-3882 migration to set Publication.LatestPublishedReleaseId and change Publisher to set it for all publications affected when publishing

benoutram:
⚠️ This has faffy deploy ticket [EES-3894](https://dfedigital.atlassian.net/browse/EES-3894) associated with running the migration.

This PR:

- Adds a migration endpoint that will set `Publication.LatestPublishedReleaseId` for all publications.
- Makes a change to Publisher to update `Publication.LatestPublishedReleaseId` for all publications affected when publishing one or more releases.

Once the migration is run and the new field is set, we can get rid of the computed method `Publication.LatestPublishedRelease()`. This change has already been prepared by https://github.com/dfe-analytical-services/explore-education-statistics/pull/3633.

LatestPublishedRelease() which looks like this, had two problems:

```
  public Release? LatestPublishedRelease()
  {
      return Releases
          .Where(IsLatestPublishedVersionOfRelease)
          .OrderBy(r => r.Year)
          .ThenBy(r => r.TimePeriodCoverage)
          .LastOrDefault();
  }
```

- It’s a computed method that can’t be executed on the database - It depends on another method which stops translation to SQL. Also an SQL query wouldn’t be able to correctly determine the latest published release of a publication.

The natural order of time periods (NVARCHAR field), would mean releases Week 1, Week2, Week3, Week 10, Week21 would be ordered Week 1, Week 10, Week 2, Week 21, Week 3.

- `IsLatestPublishedVersionOfRelease` requires the Publication to be loaded from the database with all of its Releases (all versions) to work out which is the latest version of each release.

### Benefits to adding LatestPublishedReleaseId:

- Prevent unnecessary loading of all Releases where we only want to know the latest published release or it's id.
- Add's the ability to chain `.Include(p => p.LatestPublishedRelease)` where we want to get the latest published release.
- Means that the presence of `LatestPublishedReleaseId` can be used to determine if the publication is published or not. This could be useful to simplify determining if a publication's superseding publication is published or not (although `p.SupersededBy.Published.HasValue` looks like it will be just as useful in simplifying it).

### UI Tests report

image

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Publisher/Functions/PublishReleaseContentFunction.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Publisher/Functions/PublishReleaseContentFunction.cs
@@ -22,6 +22,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Publisher.Functions
         private readonly IContentService _contentService;
         private readonly INotificationsService _notificationsService;
         private readonly IPublicationCacheService _publicationCacheService;
+        private readonly IPublicationService _publicationService;
         private readonly IReleaseService _releaseService;
         private readonly IReleasePublishingStatusService _releasePublishingStatusService;
 
@@ -30,6 +31,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Publisher.Functions
             IContentService contentService,
             INotificationsService notificationsService,
             IPublicationCacheService publicationCacheService,
+            IPublicationService publicationService,
             IReleaseService releaseService,
             IReleasePublishingStatusService releasePublishingStatusService)
         {
@@ -37,6 +39,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Publisher.Functions
             _contentService = contentService;
             _notificationsService = notificationsService;
             _publicationCacheService = publicationCacheService;
+            _publicationService = publicationService;
             _releaseService = releaseService;
             _releasePublishingStatusService = releasePublishingStatusService;
         }
@@ -79,6 +82,9 @@ namespace GovUk.Education.ExploreEducationStatistics.Publisher.Functions
                 }
 
                 var release = await _releaseService.Get(message.ReleaseId);
+                
+                // TODO EES-3369 (Read Replica-2) This call needs moving to the new PublishingCompletionService 
+                await _publicationService.UpdateLatestPublishedRelease(release.PublicationId);
 
                 // Update the cached publication and any cached superseded publications.
                 // If this is the first live release of the publication, the superseding is now enforced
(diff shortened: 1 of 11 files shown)


duncan-at-hiveit:
Hey @benoutram , great writeup on ticket https://dfedigital.atlassian.net/browse/EES-3894!

Also, did you know you can also write:

```
SELECT
    CONCAT('https://explore-education-statistics.service.gov.uk/find-statistics/',
    P.Slug,
    '/',
    R.Slug) AS PublicUrl
```

as

```
SELECT
    ('https://explore-education-statistics.service.gov.uk/find-statistics/' + P.Slug + '/' + R.Slug) AS PublicUrl
```

if you wanted to?

src/GovUk.Education.ExploreEducationStatistics.Publisher/Functions/PublishReleaseContentFunction.cs
```diff
@@ -79,6 +82,9 @@ public class PublishReleaseContentFunction
                 }
 
                 var release = await _releaseService.Get(message.ReleaseId);
+                
+                // TODO EES-3369 (Read Replica-2) This call needs moving to the new PublishingCompletionService
```
duncan-at-hiveit, line 86: Cool, thanks for that!

src/GovUk.Education.ExploreEducationStatistics.Publisher/Functions/PublishStagedReleaseContentFunction.cs
```diff
@@ -96,6 +99,13 @@ public class PublishStagedReleaseContentFunction
                         .Distinct()
                         .ToList();
 
+                    // TODO EES-3369 (Read Replica-2) This needs moving to the new PublishingCompletionService
+                    await _contentDbContext.Publications
```
duncan-at-hiveit, line 103: And again

duncan-at-hiveit (approved):

benoutram:
> Hey @benoutram , great writeup on ticket https://dfedigital.atlassian.net/browse/EES-3894!
> 
> Also, did you know you can also write:
> 
> ```
> SELECT
>     CONCAT('https://explore-education-statistics.service.gov.uk/find-statistics/',
>     P.Slug,
>     '/',
>     R.Slug) AS PublicUrl
> ```
> 
> as
> 
> ```
> SELECT
>     ('https://explore-education-statistics.service.gov.uk/find-statistics/' + P.Slug + '/' + R.Slug) AS PublicUrl
> ```
> 
> if you wanted to?

Thanks, I've updated the description now. 🙂

PR #3657, opened 2022-11-23
Ees 3779 add database resiliency options

duncan-at-hiveit:
# Overview

The purpose of this PR is to add additional resilience to our service by allowing the service to handle transient database connection errors more robustly.  Particularly in a cloud environment, the service should be resilient enough to recover from network errors and general availability errors where possible, retrying when it fails an operation due to communication errors.

## Work involved

* Enables retry-on-failure behaviour everywhere where we configure DbContexts
* Adds in a custom Retry-on-Failure Execution Strategy that adds to the OOTB EF strategy, that inspects the text of error messages when an error number itself is not identified as being handled
* Adds in the DbContext delegate pattern anywhere where we are controlling our own transaction boundaries, as mentioned in [this advice](https://learn.microsoft.com/en-us/ef/core/miscellaneous/connection-resiliency#execution-strategies-and-transactions)
* Changes the way that FilterGroups, FilterItems and Indicators are constructed so as to avoid unnecessary cascade insert issues when saving new data with a DbContext delegate

## The need for a custom ExecutionStrategy

When simply enabling the OOTB EF "RetryOnFailure()" when configuring our DbContexts, testing this with a variety of database issues showed that it wasn't handling simple scenarios that we expected it to handle.  This was in part because some errors that are explicitly listed in EF's `SqlServerTransientExceptionDetector` are all expecting particular ErrorNumbers to be part of the underlying SqlErrror that is reported, but they were coming back as "0" or "2" whereas they should be "10060", for example.  Other errors simply weren't listed in this class and again, didn't have safe-to-use ErrorNumbers to add as additional cases to handle.

To this end, we create a custom ExecutionStrategy that firstly defers to the OOTB strategy to test a retry case and, if it's reported as false, then inspects the SqlError's *message text* instead to decide if it's an error that can be handled.

Unfortunately it's nigh on impossible to unit test this with a real db connection.

## FilterGroup, FilterItem and Indicator changes

We now pass their owning entities' IDs only, rather than the entities themselves.  This allows us to create these new entities without the DbContext delegate (which knows nothing of the StateTracker from the DbContext which spawned it) from accidentally trying to cascade create the owning entity itself.  Note there were a few different ways to approach this, but just passing the IDs worked out fine and also allowed us to do a bunch of cleanup too.

image


src/GovUk.Education.ExploreEducationStatistics.Common/Database/SqlServerDbContextOptionsBuilderExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+public static class SqlServerDbContextOptionsBuilderExtensions
+{
+    public static SqlServerDbContextOptionsBuilder EnableCustomRetryOnFailure(
+        this SqlServerDbContextOptionsBuilder builder,
+        int? maxRetryCount = null,
+        TimeSpan? maxRetryDelay = null)
+    {
+        return builder.ExecutionStrategy(c =>
+            new MessageAwareSqlServerRetryingExecutionStrategy(
+                c,
```
benoutram, line 16: I think `dependencies` might be a better variable name here, or whatever this actually is. I'm not sure what `c` is in this instance just from reading it.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Startup.cs
```diff
@@ -87,18 +88,22 @@ public void ConfigureServices(IServiceCollection services)
...
                     .UseSqlServer(Configuration.GetConnectionString("StatisticsDb"),
-                        builder =>
-                        {
-                            builder.MigrationsAssembly("GovUk.Education.ExploreEducationStatistics.Data.Model");
-                            builder.AddBulkOperationSupport();
-                        })
+                        providerOptions =>
+                            providerOptions
+                                .MigrationsAssembly("GovUk.Education.ExploreEducationStatistics.Data.Model")
+                                .AddBulkOperationSupport()
```
benoutram, line 94: I think `.EnableCustomRetryOnFailure()` is missing here?

src/GovUk.Education.ExploreEducationStatistics.Common/Database/MessageAwareSqlServerRetryingExecutionStrategy.cs
```diff
@@ -0,0 +1,95 @@
...
+    private static readonly Regex[] TransientErrorMessages = {
+        // This is the error message associated with Error Number 10060, but is reported with Error Number 0, and so
+        // isn't handled correctly by SqlServerTransientExceptionDetector correctly.
+        new(@"^A network-related or instance-specific error occurred while establishing a connection to SQL Server\."),
+        
+        // An error that can occur when the connection pool is exhausted for the target database.
+        new(@"^A connection was successfully established with the server, but then an error occurred during the pre-login handshake\."),
+      
+        // An error that can occur when communication is interrupted during re-establishing communication with the database.
+        new(@"^Connection Timeout Expired\.  The timeout period elapsed while attempting to consume the pre-login handshake acknowledgement\."),
```
benoutram, line 48: Two spaces here but I did a quick search for this error and looks like this is correct! 😱 😄

src/GovUk.Education.ExploreEducationStatistics.Common/Database/MessageAwareSqlServerRetryingExecutionStrategy.cs
```diff
@@ -0,0 +1,95 @@
...
+
+    public MessageAwareSqlServerRetryingExecutionStrategy(
+        ExecutionStrategyDependencies dependencies, 
+        int maxRetryCount, 
+        TimeSpan maxRetryDelay) 
+        : base(
+            dependencies, 
+            maxRetryCount, 
+            maxRetryDelay,
+            null)
```
benoutram, line 61: You could add a prefix label to this `null` to make it clearer what this is

benoutram (commented):
Looks like a good technique for handling the missing error codes, and a nice bit of cleanup! 🙂

src/GovUk.Education.ExploreEducationStatistics.Common/Database/SqlServerDbContextOptionsBuilderExtensions.cs
duncan-at-hiveit: I couldn't agree more!  Updated.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Startup.cs
duncan-at-hiveit: Don't know how I missed this one, but great spot!  Updated.

src/GovUk.Education.ExploreEducationStatistics.Common/Database/MessageAwareSqlServerRetryingExecutionStrategy.cs
duncan-at-hiveit: Done.

benoutram (approved):

PR #3659, opened 2022-11-24
EES-3908 - notify approvers

luke-h1:
This PR:
* Notifies publication owners & publication release approvers when a release is moved into higher review. Previously, only release approvers were notified however it turned out that the intention behind this work was to notify both publication owners/releaseApprovers and release approvers

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServiceTests.cs
@@ -1779,14 +1779,6 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                 result.AssertRight();
 
             }
-
-            await using (var context = InMemoryApplicationDbContext(contextId))
-            {
-                var saved = await context.Releases.Include(r => r.ReleaseStatuses)
-                    .FirstAsync(r => r.Id == release.Id);
-
-                Assert.False(saved.ReleaseStatuses.First().NotifyReleaseApprovers);
-            }
         }
         
         [Fact]
@@ -1847,14 +1839,6 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                 result.AssertRight();
 
             }
-
-            await using (var context = InMemoryApplicationDbContext(contextId))
-            {
-                var saved = await context.Releases.Include(r => r.ReleaseStatuses)
-                    .FirstAsync(r => r.Id == release.Id);
-
-                Assert.False(saved.ReleaseStatuses.First().NotifyReleaseApprovers);
-            }
         }
 
         [Fact]
@@ -1904,10 +1888,10 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                         mock.ListUserReleaseRolesByPublication(ReleaseRole.Approver, release.Publication.Id))
                     .ReturnsAsync(ListOf( userReleaseRole1, userReleaseRole2 ));
 
-                emailTemplateService.Setup(mock => mock.SendReleaseApproverEmail(userReleaseRole1.User.Email, It.Is<Release>(r => r.Id == release.Id)))
+                emailTemplateService.Setup(mock => mock.SendHigherReviewEmail(userReleaseRole1.User.Email, It.Is<Release>(r => r.Id == release.Id)))
                     .Returns(Unit.Instance);
                 
-                emailTemplateService.Setup(mock => mock.SendReleaseApproverEmail(userReleaseRole2.User.Email, It.Is<Release>(r => r.Id == release.Id)))
+                emailTemplateService.Setup(mock => mock.SendHigherReviewEmail(userReleaseRole2.User.Email, It.Is<Release>(r => r.Id == release.Id)))
                     .Returns(Unit.Instance);
 
                 var releaseService = BuildService(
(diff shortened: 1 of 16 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServiceTests.cs
```diff
@@ -1937,7 +1937,137 @@ await using (var context = InMemoryApplicationDbContext(contextId))
...
+                var result = await releaseService
+                    .CreateReleaseStatus(
+                        release.Id, new ReleaseStatusCreateViewModel
+                        {
+                            PublishMethod = PublishMethod.Scheduled,
+                            PublishScheduled = "2051-06-30",
+                            ApprovalStatus = ReleaseApprovalStatus.HigherLevelReview,
+                            LatestInternalReleaseNote = "Test internal note",
+                            NextReleaseDate = new PartialDate { Month = "12", Year = "2077" },
+                        });
```
benoutram, line 2001: I suspect this method isn't working as expected, and that it doesn't actually send the email, or that something's wrong with this unit test.

You haven't set up an `EmailTemplateService` with the expected behaviour (i.e. the method call for the email to be sent).

Without setting it up, it's defaulting to a mock service that has strict behaviour. If the strict mock is called to send the email then it would fail.

The fact that it's not failing, indicates the email isn't getting sent.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServiceTests.cs
```diff
@@ -1937,7 +1937,137 @@ await using (var context = InMemoryApplicationDbContext(contextId))
...
```
benoutram, line 2066: Same problem as before, suspect either the email isn't being sent or there's a problem with the unit test.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServiceTests.cs
```diff
@@ -1937,7 +1937,137 @@ await using (var context = InMemoryApplicationDbContext(contextId))
...
+            await using (var context = InMemoryApplicationDbContext(contextId))
+            {
+                await context.AddRangeAsync(release, userPublicationReleaseApproverRole);
+                await context.SaveChangesAsync();
+            }
+
+            await using (var context = InMemoryApplicationDbContext(contextId))
+            {
+                var contentService = new Mock<IContentService>(MockBehavior.Strict);
+                var preReleaseUserService = new Mock<IPreReleaseUserService>(MockBehavior.Strict);
```
benoutram, line 2039: Same problem as before, `preReleaseUserService` could be omitted here.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServiceTests.cs
```diff
@@ -1937,7 +1937,137 @@ await using (var context = InMemoryApplicationDbContext(contextId))
...
+                var userReleaseRoleService = new Mock<IUserReleaseRoleService>(MockBehavior.Strict);
+                
+                contentService.Setup(mock => mock.GetContentBlocks<HtmlBlock>(release.Id))
+                    .ReturnsAsync(new List<HtmlBlock>());
+
```
benoutram, line 2044: Minor, extra line break here

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseApprovalService.cs
```diff
@@ -140,16 +141,22 @@ public class ReleaseApprovalService : IReleaseApprovalService
...
                             release.Publication.Id);
-
-                    var notifyReleaseApprovers = request.ApprovalStatus == ReleaseApprovalStatus.HigherLevelReview &&
-                                                 userReleaseRoles.Any();
                     
+                    
+                    var userPublicationRoles = await _context.UserPublicationRoles
+                        .Where(upr => upr.PublicationId == release.PublicationId && (upr.Role == PublicationRole.Owner || upr.Role == PublicationRole.ReleaseApprover))
+                        .ToListAsync();
+
```
benoutram, line 149: Minor, extra line break here

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseApprovalService.cs
```diff
@@ -140,16 +141,22 @@ public class ReleaseApprovalService : IReleaseApprovalService
...
+                    var notifyHigherReviewers = request.ApprovalStatus == ReleaseApprovalStatus.HigherLevelReview &&
+                        userReleaseRoles.Any() || userPublicationRoles.Any();
+
                     var releaseStatus = new ReleaseStatus
                     {
                         Release = release,
                         InternalReleaseNote = request.LatestInternalReleaseNote,
                         NotifySubscribers = notifySubscribers,
-                        NotifyReleaseApprovers = notifyReleaseApprovers,
+                        NotifyHigherReviewers = notifyHigherReviewers,
```
benoutram, line 159: To me, `NotifySubscribers` makes sense, i.e. whether the user decided to notify subscribers or not, regardless of whether there are any subscribers.

`NotifyHigherReviewers` is not based on the user deciding anything, but on whether any user has those roles.

These fields work differently, one audits the users decision so we can display it in the UI, one audits whether emails were sent.

I raised a discussion on this because either the naming could be better to make the reader  realise they mean different things, or we might not need `NotifyHigherReviewers` at all.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseApprovalService.cs
luke-h1: Yep good point, as discussed after standup I've gotten rid of this field altogether

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseApprovalService.cs
```diff
@@ -140,16 +141,22 @@ public class ReleaseApprovalService : IReleaseApprovalService
...
                     
+                    
+                    var userPublicationRoles = await _context.UserPublicationRoles
+                        .Include(upr => upr.User)
+                        .Where(upr => upr.PublicationId == release.PublicationId && (upr.Role == PublicationRole.Owner || upr.Role == PublicationRole.ReleaseApprover))
+                        .ToListAsync();
+                    
+
+                    var notifyHigherReviewers = request.ApprovalStatus == ReleaseApprovalStatus.HigherLevelReview &&
+                        userReleaseRoles.Any() || userPublicationRoles.Any();
```
benoutram, line 153: You could move this closer to where it's referenced now, i.e. after the line `case ReleaseApprovalStatus.HigherLevelReview:`

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServiceTests.cs
```diff
@@ -1785,7 +1785,6 @@ await using (var context = InMemoryApplicationDbContext(contextId))
                 var saved = await context.Releases.Include(r => r.ReleaseStatuses)
```
benoutram, line 1785: I think you could remove this whole block now as it's not really doing anything. There's other tests checking the release status has been saved correctly, so I think this can go.

```
await using (var context = InMemoryApplicationDbContext(contextId))
{
    var saved = await context.Releases.Include(r => r.ReleaseStatuses)
        .FirstAsync(r => r.Id == release.Id);
}
```

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServiceTests.cs
```diff
@@ -1853,7 +1852,6 @@ await using (var context = InMemoryApplicationDbContext(contextId))
...
                     .FirstAsync(r => r.Id == release.Id);
 
-                Assert.False(saved.ReleaseStatuses.First().NotifyReleaseApprovers);
```
benoutram, line 1856: Same as other comment, I think this can go now.

benoutram (approved):

PR #3661, opened 2022-11-24
EES-3921 Update PublicationController.ListPublications to return publication Slug in response

benoutram:
This PR updates the Content API  `PublicationController.ListPublications` to return publication `Slug` in the response.

It's required so that we can link to publications from results in the Find Statistics page.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Services/ViewModels/PublicationSearchResultViewModel.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Services/ViewModels/PublicationSearchResultViewModel.cs
@@ -9,6 +9,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Services.ViewModels
 public record PublicationSearchResultViewModel
 {
     public Guid Id { get; init; }
+    public string Slug { get; init; }
     public string Summary { get; init; }
     public string Title { get; init; }
     public string Theme { get; init; }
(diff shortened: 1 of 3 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Services/ViewModels/PublicationSearchResultViewModel.cs
```diff
@@ -9,6 +9,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Services.ViewModels
 public record PublicationSearchResultViewModel
```
luke-h1, line 9: Just wondering if we need to add the legacy publication url to this view model? Not sure if it's needed with the new design but thought it'd be worth mentioning

luke-h1 (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ViewModels/PublicationSearchResultViewModel.cs
benoutram: Thanks, legacy publications "Not yet on this service" aren't going to be displayed, or at least not in the current plan. 

@cjrace is planning to have them all removed from the service by the time the new Find Statistics page goes live.

I think there's currently 5, with 2 due to be removed imminently.

`ListPublications` has been designed to only return publications with published releases.

PR #3663, opened 2022-11-24
EES-3668 hook up new find stats paginated list and sorting

amyb-hiveit:
Hooks up the paginated list and sort controls in the new Find Statistics UI to the new endpoint. See https://github.com/dfe-analytical-services/explore-education-statistics/pull/3459 for the original UI work and https://github.com/dfe-analytical-services/explore-education-statistics/pull/3635 for details of the endpoint.

Legacy release will no longer be shown on this page (they should all have been removed by the time it goes live).

The existing page should not be affected, to see the new one use the newDesign=true query parameter, e.g. http://localhost:3000/find-statistics?newDesign=true

You'll need to have the the latest database (ees-mssql-data-53.zip) locally for the list to be populated.

### DIFF
--- a/src/explore-education-statistics-common/src/services/publicationService.ts
+++ b/src/explore-education-statistics-common/src/services/publicationService.ts
@@ -5,6 +5,7 @@ import {
   MethodologySummary,
   ExternalMethodology,
 } from '@common/services/types/methodology';
+import { Paging } from '@common/services/types/pagination';
 import { PartialDate } from '@common/utils/date/partialDate';
 import { contentApi } from './api';
 
@@ -42,21 +43,20 @@ export interface PublicationSummary {
   title: string;
 }
 
-// TODO EES-3517 a guess at the new type for the find stats page.
-export interface PublicationSummaryWithRelease {
+export interface PublicationListSummary {
   id: string;
-  legacyPublicationUrl?: string;
-  latestRelease?: {
-    id: string;
-    published: string;
-    type: ReleaseType;
-  };
+  published: Date;
+  rank: number;
   slug: string;
   summary?: string;
-  theme: {
-    title: string;
-  };
+  theme: string;
   title: string;
+  type: ReleaseType;
+}
+
+interface PublicationsResponse {
+  results: PublicationListSummary[];
+  paging: Paging;
 }
 
 export interface Contact {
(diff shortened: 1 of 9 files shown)


src/explore-education-statistics-common/src/services/publicationService.ts
```diff
@@ -206,4 +210,11 @@ export default {
       `/publications/${publicationSlug}/releases/${releaseSlug}/prerelease-access-list`,
     );
   },
+  getPublications(
+    params: GetPublicationsRequest,
+  ): Promise<PublicationsResponse> {
+    return contentApi.get(`/publications?`, {
```
ntsim, line 216: This shouldn't need a `?` at the end. Think Axios should automatically append the `params` into a query string without any further intervention.

src/explore-education-statistics-common/src/services/publicationService.ts
```diff
@@ -91,18 +91,22 @@ export interface ContentSection<BlockType> {
...
 
 export type PublicationSortOption = typeof publicationSortOptions[number];
 
-// TODO EES-3517 expand to include filters and search
-export interface ListPublicationsRequest {
+export type PublicationSortParam = 'published' | 'title' | 'relevance';
+
+export type PublicationOrderParam = 'asc' | 'desc';
+
+interface GetPublicationsRequest {
```
ntsim, line 102: Would suggest naming this `PublicationListRequest`.

Whilst it's not a big big deal, would say it's a good idea for type names to follow similar naming conventions as component file names.

This means nouns should go from less to more specific terminology i.e. `Publication` is the least specific thing, whilst the `ListRequest` is the most specific thing hence `PublicationListRequest` instead of `ListPublicationRequest`.

We're already doing this with `PublicationSortParam` and `PublicationOrderParam` so we should be consistent.

ntsim (changes requested):

src/explore-education-statistics-common/src/services/publicationService.ts
amyb-hiveit: yeah that's typo, it's gone now.

src/explore-education-statistics-common/src/services/publicationService.ts
```diff
@@ -206,4 +210,11 @@ export default {
       `/publications/${publicationSlug}/releases/${releaseSlug}/prerelease-access-list`,
     );
   },
+  getPublications(
```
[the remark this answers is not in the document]
amyb-hiveit: ok, changed this and for the request.

src/explore-education-statistics-common/src/services/publicationService.ts
```diff
@@ -91,18 +91,22 @@ export interface ContentSection<BlockType> {
...
+export type PublicationSortParam = 'published' | 'title' | 'relevance';
+
+export type PublicationOrderParam = 'asc' | 'desc';
+
+interface PublicationListRequest {
+  order?: PublicationOrderParam;
   page?: number;
-  sortBy?: PublicationSortOption;
+  pageSize?: number;
+  releaseType?: string;
```
luke-h1, line 106: Think we could narrow the type constraints of `releaseType` here to be of `ReleaseType`

amyb-hiveit: Good spot, sorted now.

ntsim (approved):

PR #3669, opened 2022-11-25
Ees 1823 add expiry to admin invites

duncan-at-hiveit:
This PR:
- adds logic to detect if a User Invite is expired and, if so, redirect the user to a new page explaining that the invite is expired
- makes the existing "create user invite" service slightly more flexible to allow already-expired invites to be created for the purposes of aiding the Robot test data creation
- update 30/11/22 - hides expired invites from invitation lists
- update 30/11/22 - allows a user with an already active invite to be re-invited (as in the case of expired invites, they are not visible to be cancelled and resent).  In these cases, their previous "main" invite with their global role will be updated, and in the case of Release Invites and Publication Invites attached to the main invite, they will replace any existing Release Invites and Publication Invites for the user
- update 30/11/22 - moved "Invite a new user" button to the top of the User Invites page

## New page

Upon signing in using an expired invite, the user will see the following page:

image

Any subsequent attempts to login will take the user back to the "Sign in" page, as per existing behaviour with users who can log into the IdPs but have no permission to use the service itself.

## Update 30/11/22 - Removing expired Invites from the Invite Users page

We apply a Global EF filter to remove expired UserInvites from being visible by default.  This means that expired invites will by default not appear on any pages. 

## Update 30/11/22 - Updating invites upon inviting a person with an existing invite

Previously if a person had a prior invite outstanding and another invite was sent to them, the original "main" invite (with the user's Global role selection) would be left alone and only any new Release Invites and Publication Invites would be added.  Now, the existing "main" invite will be updated (and thus will update its creation date to "unexpire" any expired ones), and all Release and Publication Invites will be updated, including removing any previously set ones.

:warning: A new cleaned-up ees-mssql-data-54.zip is available at https://drive.google.com/file/d/1vxLDK5Tz0gB2bomtQZGQOX2fQWu0gQr0/view?usp=share_link

:warning: New Keycloak users have been added, so IDP containers will need to be recreated, using `docker-compose up --build --force-recreate idp`

:warning: 2 new variables, EXPIRED_INVITE_USER_EMAIL and EXPIRED_INVITE_USER_PASSWORD will need to be added to peoples' various Robot `.env.*` files

:warning: these 2 variables need adding to the Pipeline - ticket https://dfedigital.atlassian.net/browse/EES-3925 is available to track this

## Update 30/11/22 - Button move

image

## UI test results

image

Note that the UI tests have been tested against both Keycloak and Azure.

## Update 30/11/22 - UI test results

image


tests/robot-tests/tests/libs/admin_api.py
```diff
@@ -199,15 +218,48 @@ def user_create_test_release_via_api(
     ), f"Creating release API request failed with {response.status_code} and {response.text}"
 
 
-def get_user_details_via_api(user_email: str):
+def __get_user_details_via_api(user_email: str):
```
luke-h1, line 221: Only need one underscore to denote this function as private

tests/robot-tests/tests/libs/admin_api.py
```diff
@@ -199,15 +218,48 @@ def user_create_test_release_via_api(
...
-    return list(filter(lambda user: user["email"] == user_email, users))[0]
+    matching_users = list(filter(lambda user: user["email"] == user_email, users))
+
+    assert matching_users, f"Could not find user with email {user_email}"
+
+    return matching_users[0]
 
 
-def get_prerelease_user_details_via_api(user_email: str):
+def __get_prerelease_user_details_via_api(user_email: str):
```
luke-h1, line 232: Similar to above comment

tests/robot-tests/tests/libs/admin_api.py
```diff
@@ -199,15 +218,48 @@ def user_create_test_release_via_api(
...
 
-    return list(filter(lambda user: user["email"] == user_email, users))[0]
+    matching_users = list(filter(lambda user: user["email"] == user_email, users))
+
+    assert matching_users, f"Could not find user with email {user_email}"
+
+    return matching_users[0]
+
+
+def __get_global_role_id(role_name: str):
```
luke-h1, line 243: Similar to above comment

tests/robot-tests/tests/libs/admin_api.py
```diff
@@ -199,15 +218,48 @@ def user_create_test_release_via_api(
...
+
+    roles = response.json()
+    matching_roles = list(filter(lambda role: role["name"] == role_name, roles))
+
+    assert matching_roles, f"Could not find global role matching name {role_name}"
+
+    return matching_roles[0]["id"]
+
+
+def __get_user_invite(user_email: str):
```
luke-h1, line 258: Similar to above comment

tests/robot-tests/tests/libs/admin_api.py
```diff
@@ -1,5 +1,6 @@
 import json
```
luke-h1, line 1: The formatting looks a bit off here. Python linters & formatters should be catching this via pre-commit hooks

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PreReleaseUserService.cs
```diff
@@ -369,7 +369,7 @@ await _context.Entry(release)
 
             var template = _configuration.GetValue<string>("NotifyPreReleaseTemplateId");
 
-            var scheme = _httpContextAccessor.HttpContext.Request.Scheme;
+            var scheme = _httpContextAccessor.HttpContext!.Request.Scheme;
```
luke-h1, line 372: Just wondering why we're having to use the null forgiving op here?

README.md
```diff
@@ -234,7 +234,20 @@ The [Keycloak Admin login](http://ees.local:5030/auth/admin/) is available with
 
 The environment variable `IdpProviderConfiguration` lets Admin know to use 
 [appsettings.Keycloak.json](src/GovUk.Education.ExploreEducationStatistics.Admin\appsettings.Keycloak.json) 
-for its Open ID Connect configuration. 
+for its Open ID Connect configuration.
+
+Additional seed data users can be added to Keycloak by manually adding new entries to the "users" array in
+[keycloak-ees-realm.json](src/keycloak/keycloak-ees-realm.json), ensuring to supply unique GUIDs to the `user` and
+`credentials` ids.  If copying and pasting from an existing user record in the array, the new user password will be
```
luke-h1, line 241:
```suggestion
`credentials` ids. If copying and pasting from an existing user record in the array, the new user password will be
```

README.md
```diff
@@ -234,7 +234,20 @@ The [Keycloak Admin login](http://ees.local:5030/auth/admin/) is available with
...
+"password" also.
+
+After this, existing Keycloak Docker containers will need to be rebuilt in order to pick up the new user list. To
+do this, run:
+
+  ```bash
```
luke-h1, line 247:
```suggestion
```bash
```

README.md
```diff
@@ -234,7 +234,20 @@ The [Keycloak Admin login](http://ees.local:5030/auth/admin/) is available with
...
+  cd src/
+  docker-compose up --build --force-recreate idp
```
luke-h1, line 249:
```suggestion
docker-compose up --build --force-recreate idp
```

luke-h1 (changes requested):

tests/robot-tests/tests/admin/analyst/expired_invitation.robot
```diff
@@ -0,0 +1,42 @@
...
+    user waits until page contains title    Invitation expired
+    user waits until page contains    Your invitation to the service has expired.
+    user waits until page contains    explore.statistics@education.gov.uk
+
+Check that the expired invite is now removed
+    user navigates to admin frontend
+    user waits until h1 is visible    Sign in    %{WAIT_MEDIUM}
+    user clicks element    id:signin-button
+
+    ## Currently, users attempting to login with no existing user record or user invite
```
luke-h1, line 37: Don't think double hashtags are needed here

luke-h1, line 37: Don't think double hashtags are needed here

tests/robot-tests/tests/libs/admin-common.robot
```diff
@@ -9,6 +9,25 @@ ${ANALYST1_BROWSER}=    analyst1
 
 
 *** Keywords ***
+user logs in via identity provider
```
luke-h1, line 12: I think `get_auth_tokens.py` shares a lot of logic with this keyword. Potentially it could be modified to be used  in both robot tests and generic python test code

tests/robot-tests/tests/admin/analyst/expired_invitation.robot
```diff
@@ -0,0 +1,42 @@
...
+Resource            ../../libs/admin-common.robot
+
+Suite Setup         user opens the browser    %{ADMIN_URL}
+Suite Teardown      user closes the browser
+Test Setup          fail test fast if required
+
+Force Tags          Admin    Local    Dev
+
+
+*** Test Cases ***
```
markhiveit, line 12: I think we also need a UI test that creates an expired invite and then checks it doesn't appear on the `Manage pre-release user access` - before the associated user has tried to accept the invite.

I'm not sure if there are other pages that list invites that also shouldn't include expired invites.

src/GovUk.Education.ExploreEducationStatistics.Admin/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
```diff
@@ -343,6 +352,39 @@ private IActionResult RedirectToLoginPageWithError(string returnUrl)
...
+            return await Task.FromResult(pageRedirect);
+        }
+
+        private async Task<IActionResult> HandleExpiredInvite(UserInvite inviteToSystem, string email)
+        {
+            var releaseInvites = GetUserReleaseInvites(email);
+            var publicationInvites = GetUserPublicationInvites(email);
+            _usersAndRolesDbContext.UserInvites.Remove(inviteToSystem);
+            _contentDbContext.UserReleaseInvites.RemoveRange(releaseInvites);
+            _contentDbContext.UserPublicationInvites.RemoveRange(publicationInvites);
```
markhiveit, line 368: I think we might want to check the created dates of all these invites too, rather than deleting them all?

markhiveit (changes requested):

tests/robot-tests/tests/libs/admin_api.py
duncan-at-hiveit: Nice spot - I got my conventions wrong!  Updated.

tests/robot-tests/tests/libs/admin_api.py
duncan-at-hiveit: Done

tests/robot-tests/tests/libs/admin_api.py
duncan-at-hiveit: Done

tests/robot-tests/tests/libs/admin_api.py
duncan-at-hiveit: This file did get pulled up for manual fixes, so whatever ended up in here went through the linters.  If there's anything in particular you're thinking of, I can sort it manually or look into which rules to enable.

README.md
duncan-at-hiveit: Done ta

README.md
duncan-at-hiveit: Done

README.md
duncan-at-hiveit: Done

tests/robot-tests/tests/admin/analyst/expired_invitation.robot
duncan-at-hiveit: Right you are - just left single ones, thanks

duncan-at-hiveit: Right you are - just left single ones, thanks

tests/robot-tests/tests/libs/admin-common.robot
duncan-at-hiveit: Yeah I was surprised initially not to find any actual "real" login code in Robot itself, but I knew there was a bunch in Python code and also in the Puppeteer code too.  Would be nice to consolidate the Python code in some way - as you say, perhaps as a part of another piece of work.

src/GovUk.Education.ExploreEducationStatistics.Admin/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
```diff
@@ -238,6 +242,13 @@ private IActionResult RedirectToLoginPageWithError(string returnUrl)
             ExternalLoginInfo info,
             string returnUrl)
         {
+            var inviteExpiryTime = inviteToSystem.Created + InviteExpirySpan;
+
+            if (DateTime.UtcNow >= inviteExpiryTime)
+            {
+                return await HandleExpiredInvite(inviteToSystem, email);
+            }
```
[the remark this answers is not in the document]
duncan-at-hiveit: Yeah I think we should bring this up with Cam tomorrow.  As far as I understood, this ticket was actually around expiring invites to the service itself rather than adding users to entities after that.

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/DbContextUtils.cs
```diff
@@ -19,19 +19,31 @@ public static void UpdateTimestamps(object? sender, EntityEntryEventArgs e)
                 switch (e.Entry.Entity)
                 {
                     case ICreatedTimestamp<DateTime> entityWithCreated:
-                        entityWithCreated.Created = now;
+                        if (entityWithCreated.Created == null)
+                        {
+                            entityWithCreated.Created = now;
+                        }
```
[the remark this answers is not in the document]
duncan-at-hiveit: Yeah - tbf, it was never a use case before, because in general there's no need to set a CreatedDate manually.  This is really just in place to help us with supporting test data setup!

tests/robot-tests/tests/admin/analyst/expired_invitation.robot
duncan-at-hiveit: As mentioned elsewhere, Cam was happy for expired invites to remain on lists until it was deleted by the user logging in.

tests/robot-tests/tests/libs/admin_api.py
luke-h1: There seem to be spaces in the first part of `user_adds_user_invite_via_api `, `user_adds_release_role_to_user_via_api ` & `user_adds_publication_role_to_user_via_api `. Aside from that, all looks good

luke-h1: Going to sort these formatting issues in a separate PR as it affects quite a few python files

markhiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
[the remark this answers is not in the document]
markhiveit: I think maybe my comment isn't correct anyway - I've got a feeling that if an existing user is invited to have a release/publication permission, then they are instantly given that permission, I think.

PR #3672, opened 2022-11-28
EES-3682 Add ability to embed dashboards in release content

markhiveit:
This PR gives analysts the ability to embed dashboards in release content:

image

To do this, we have added a single column to the `ContentBlock` table, `EmbedBlockId`, and a new table `EmbedBlocks`. We've separated the `EmbedBlock` data out of the `ContentBlock` table to firstly, reduce the number of extra columns that table needs, and secondly as in the long term we intend to essentially do the same for existing block types -- `HtmlBlock`, `DataBlock` etc. So the new content block type is called `EmbedBlockLink` which links to an `EmbedBlock`

Because of this, we've had to include the embed block as part of SQL queries we've generated like so:
```
.ThenInclude(contentBlock => (contentBlock as EmbedBlockLink)!.EmbedBlock)
```
This will include `.EmbedBlock` *if* the content block is of type `EmbedBlockLink`.

Alongside allowing the creation, updating, and deleting of `EmbedBlock`s, we've also needed to handle the copying of `EmbedBlock`s and `EmbedBlockLink`s on the creation of an amendment.

To ensure the dashboard iframe is sized correctly, we imported a new dependency on the frontend: `iframe-resizer-react`.

Other changes:
- `ReleaseService#HydrateReleaseForAmendment` was refactored as per EES-3164. Now that Entity Framework has been upgraded, we can use `AsSplitQuery`. This should produce the same SQL query as the previous `_context.Entry(blah)` code.
- To give credit where credit is due, the vast majority of the frontend work was done by Amy.


src/GovUk.Education.ExploreEducationStatistics.Content.Model/EmbedBlock.cs
```diff
@@ -0,0 +1,24 @@
+﻿#nullable enable
+using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Model;
+
+public class EmbedBlock
```
luke-h1, line 7: Could we add `created` + `updated` fields here for auditing/best practices?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/EmbedBlockService.cs
```diff
@@ -0,0 +1,153 @@
...
+
+                        var embedBlock = new EmbedBlock
+                        {
+                            Id = Guid.NewGuid(),
+                            Title = request.Title,
+                            Url = request.Url,
+                        };
+
+                        var order = contentSection.Content.Any()
+                            ? contentSection.Content.Max(contentBlock => contentBlock.Order) + 1
```
luke-h1, line 69: Just wondering if we need to account for negative values here? i.e. if a user adds an order such as `-10`

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/EmbedBlockLinkViewModel.cs
```diff
@@ -0,0 +1,27 @@
+﻿#nullable enable
+using System;
+using System.Collections.Generic;
+using GovUk.Education.ExploreEducationStatistics.Content.Model;
+using JsonKnownTypes;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.ViewModels;
+
+[JsonKnownThisType(nameof(EmbedBlockLink))]
+public class EmbedBlockLinkViewModel : IContentBlockViewModel
```
luke-h1, line 10: Might be worth exposing `created` / `updated` fields here if they end up getting added

src/explore-education-statistics-admin/src/components/editable/__tests__/EditableEmbedBlock.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+        order: 0,
+        title: 'Dashboard title-edited',
+        type: 'EmbedBlock',
+        url:
+          'https://department-for-education.shinyapps.io/test-dashboard-edited',
+      });
+    });
+  });
+
+  test('when click the remove block button it shows the modal and calls `onDelete` on confirm', async () => {
```
luke-h1, line 129:
```suggestion
  test('when clicking the remove block button it shows the modal and calls `onDelete` on confirm', async () => {
```

src/explore-education-statistics-admin/src/components/editable/__tests__/EditableEmbedForm.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+    );
+    userEvent.click(screen.getByRole('button', { name: 'Save' }));
+
+    await waitFor(() => {
+      expect(handleSubmit).toHaveBeenCalledWith(
+        {
+          title: 'Dashboard title',
+          url: 'https://department-for-education.shinyapps.io/test-dashboard',
+        },
+        expect.anything(),
```
luke-h1, line 117: Just wondering why we need this?

src/explore-education-statistics-admin/src/pages/release/content/contexts/useReleaseContentActions.ts
```diff
@@ -234,6 +239,86 @@ export default function useReleaseContentActions() {
...
+        releaseId,
+        request,
+      );
+
+      dispatch({
+        type: 'ADD_SECTION_BLOCK',
+        payload: { meta: { sectionId, sectionKey }, block: newBlock },
+      });
+
+      await updateAvailableDataBlocks({ releaseId }); // need this?
```
luke-h1, line 264: leftover comment?

tests/robot-tests/tests/admin_and_public_2/bau/publish_release_and_amend.robot
```diff
@@ -157,6 +159,17 @@ Add test text to second accordion section
     user adds content to autosaving accordion section text block    Test text    1    Some test text!
     ...    css:#releaseMainContent
 
+Add embedded dashboard to third accordion section
+    user adds embedded dashboard to editable accordion section
+    ...    Test embedded dashboard section
+    ...    Test embedded dashboard title
+    ...    https://department-for-education.shinyapps.io/dfe-shiny-template/
+    ...    css:#releaseMainContent
```
luke-h1, line 167: could potentially use `id:releaseMainContent` here

tests/robot-tests/tests/admin_and_public_2/bau/publish_release_and_amend.robot
```diff
@@ -559,6 +582,15 @@ Update second accordion section text for amendment
     user adds content to autosaving accordion section text block    Test text    1    Updated test text!
     ...    css:#releaseMainContent
 
+Update embedded dashboard title
+    user updates embedded dashboard in editable accordion section
+    ...    Test embedded dashboard section
+    ...    Test embedded dashboard title updated
+    ...    https://department-for-education.shinyapps.io/dfe-shiny-template/
+    ...    css:#releaseMainContent
```
luke-h1, line 590: could potentially use id:releaseMainContent here

tests/robot-tests/tests/admin_and_public_2/bau/publish_release_and_amend.robot
```diff
@@ -559,6 +582,15 @@ Update second accordion section text for amendment
...
+
+    user closes accordion section    Test embedded dashboard section    css:#releaseMainContent
```
luke-h1, line 592: could potentially use id:releaseMainContent here

src/GovUk.Education.ExploreEducationStatistics.Content.Model/EmbedBlock.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public Guid Id { get; set; }
+
+    [Required]
+    public string Title { get; set; }
+
+    [Required]
```
luke-h1, line 14: Wonder if we should add some regex here to match the DFE shiny apps domain just to be extra safe

luke-h1 (dismissed):

luke-h1:
The build will be failing due to `iframe-resizer-react` having no types. I don't there are any definitely typed packages out there that provide types for this package. Given that this package doesn't have provide types and it hasn't been updated in a while, I would suggest looking at some alternatives if it's not too much work or writing the types in a declaration file

src/GovUk.Education.ExploreEducationStatistics.Content.Model/EmbedBlock.cs
markhiveit: I've added regex checks to `EmbedBlockCreateRequest` and `EmbedBlockUpdateRequest`. Apparently, adding such a check to `EmbedBlock` itself wouldn't get checked.

tests/robot-tests/tests/admin_and_public_2/bau/publish_release_and_amend.robot
markhiveit: I've changed all instances across the UI tests.

src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseBlock.tsx
```diff
@@ -20,6 +21,14 @@ const ReleaseBlock = ({ block, releaseId, visible }: Props) => {
     releaseId,
   });
 
+  if (block.type === 'EmbedBlockLink') {
+    return (
+      <Gate condition={!!visible} key={block.id}>
+        <EmbedBlock key={block.id} url={block.url} title={block.title} />
```
ntsim, line 27: Don't need a `key` here

src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseContentAccordionSection.tsx
```diff
@@ -123,100 +144,127 @@ const ReleaseContentAccordionSection = ({
...
+        title="Embed dashboard link"
+        open={showEmbedDashboardForm}
+        onExit={toggleEmbedDashboardForm.off}
+      >
+        <EditableEmbedForm
+          onCancel={toggleEmbedDashboardForm.off}
+          onSubmit={embedBlock => {
+            addEmbedBlock(embedBlock);
+            toggleEmbedDashboardForm.off();
+          }}
```
ntsim, line 264: Similar to another comment, would suggest we async/await the `addEmbedBlock` call to avoid the toggle being flipped before the work has actually finished

src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseContentAccordionSection.tsx
```diff
@@ -123,100 +144,127 @@ const ReleaseContentAccordionSection = ({
...
-                allowImages
-                block={block}
-                sectionId={sectionId}
-                sectionKey="content"
-                editable={!isReordering}
-                publicationId={release.publication.id}
-                releaseId={release.id}
-                visible={open}
-              />
+    <>
```
ntsim, line 147: Rather than having the modal at the root of the component, would just put it inside of the `EditableAccordionSection` children. There shouldn't be a difference and we don't end up creating a bunch of formatting related changes in this diff.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/EmbedBlockLinkViewModel.cs
markhiveit: I discussed this with Ben and Nick. I'll add the `Created` and `Updated` fields to `EmbedBlock`.  There are already `Created` and `Updated` fields on `ContentBlock`. We discussed whether it was worth surfacing these into `IContentBlockViewModel`, but since the frontend doesn't use these values we decided we wouldn't. I think adding these columns is mostly to help us debugging and future proofing if they're needed in the future.

src/explore-education-statistics-admin/src/pages/release/content/contexts/useReleaseContentActions.ts
```diff
@@ -234,6 +239,86 @@ export default function useReleaseContentActions() {
...
+      blockId: string;
+      sectionKey: ContentSectionKeys;
+    }) => {
+      await releaseContentService.deleteEmbedSectionBlock(releaseId, blockId);
+      dispatch({
+        type: 'REMOVE_SECTION_BLOCK',
+        payload: { meta: { sectionId, blockId, sectionKey } },
+      });
+
+      await updateAvailableDataBlocks({ releaseId });
```
ntsim, line 287: This doesn't need to be called as embed blocks don't affect data blocks. We should remove it from the dependency array of this `useCallback` as well.

src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseContentAccordionSection.tsx
```diff
@@ -123,100 +144,127 @@ const ReleaseContentAccordionSection = ({
...
+                    >
+                      Add data block
+                    </Button>
+                  )}
+                  {!showEmbedDashboardForm && (
+                    <Button
+                      variant="secondary"
+                      onClick={toggleEmbedDashboardForm.on}
+                    >
+                      Embed dashboard
```
ntsim, line 244: Shouldn't this be 'Add embed block' or 'Add embed'?

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Model/EmbedBlock.cs
markhiveit: I've added `Created` and `Updated` fields to `EmbedBlock`, but I've not put them in models since they wouldn't currently be used. I think we mainly add these for debugging and in case we need them in the future.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/EmbedBlockService.cs
markhiveit: I think it's ok to leave as it is. I'm pretty sure other areas of the codebase have the same potential flaw. 

It will preserve the order even if there were negative numbers - but it seems unlikely that would happen. I don't think there is any security issue with this, so while it would be possible for someone to add content blocks with unexpected order numbers, it wouldn't cause any issues - at least I don't think.

src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseContentAccordionSection.tsx
markhiveit: I think I prefer "Embed dashboard" - I know we've used embed as a noun in the codebase, but it does make sense as a verb here.

src/explore-education-statistics-admin/src/components/editable/__tests__/EditableEmbedForm.test.tsx
amyb-hiveit: It's because we're using `useFormSubmit` which means as well as the things we expect `toHaveBeenCalledWith` picks up various methods related to forms (e.g. validateForm, resetForm), so the `expect.anything()` just means we don't have to worry about them.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/EmbedBlockLinkViewModel.cs
luke-h1: Cool, initially just thought of this for the debugging purposes

src/explore-education-statistics-frontend/server.js
```diff
@@ -79,20 +73,16 @@ async function startServer(port = process.env.PORT || 3000) {
...
+          fontSrc: ["'self'"],
           connectSrc:
             process.env.NODE_ENV !== 'production' ? ['*'] : cspConnectSrc,
-          frameSrc: ["'self'", 'https://vars.hotjar.com '],
+          frameSrc: [
+            "'self'",
+            'https://department-for-education.shinyapps.io/',
+          ],
           frameAncestors: ["'self'"],
-          childSrc: ["'self'", 'https://vars.hotjar.com'],
```
luke-h1, line 95: Not necessary for this PR but since we're getting rid of the hotjar CSP, maybe it'd be worth deleting the hotjar import from `_app.tsx` since it's unused?

tests/robot-tests/tests/admin_and_public_2/bau/publish_release_and_amend.robot
```diff
@@ -339,6 +353,15 @@ Verify Test text accordion section contains correct text
     user waits until parent contains element    ${section}    xpath:.//p[text()="Some test text!"]
     user closes accordion section    Test text    id:content
 
+Verify embedded dashboard accordion section contains dashboard
+    user opens accordion section    Test embedded dashboard section    id:content
+    ${section}=    user gets accordion section content element    Test embedded dashboard section    id:content
+    user waits until parent contains element    ${section}    xpath:.//iframe[@title="Test embedded dashboard title"]
+
+    select frame    xpath://iframe[@title="Test embedded dashboard title"]
+    user waits until h1 is visible    DfE Analytical Services R-Shiny data dashboard template (h1)    90
```
luke-h1, line 362: Just wondering why we've got such a long wait here?

markhiveit: I think I had an instance when it failed once - I believe because it was taking a long time to load. I'll change it back to wait small.

amyb-hiveit:
The whitespace issue on the iframe embeds and it's caused by a combination of things, one is the heightCalculation setting so I've changed that. The other is to do with the styling on the dashboard itself which causes the iframe to not resize down if the content becomes smaller.
It works ok when I embed https://department-for-education.shinyapps.io/education-to-employment-dashboard/ but not the template one https://department-for-education.shinyapps.io/dfe-shiny-template/. I think this is likely to do with the height that's applied to the body element on the template (a bit of a guess but it's the most obvious difference between the two). So we might need to experiment a bit once we've got an example dashboard.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/EmbedBlockService.cs
```diff
@@ -0,0 +1,157 @@
...
+                            && rcs.ContentSectionId == contentBlock.ContentSectionId))
+                        {
+                            return ValidationUtils.ValidationActionResult(
+                                ValidationErrorMessages.ContentBlockNotAttachedToRelease);
+                        }
+
+                        await _contentBlockService.DeleteContentBlockAndReorder(contentBlock.Id);
+                        await _contentDbContext.SaveChangesAsync();
+
+                        return Unit.Instance;
```
ntsim, line 153: Shouldn't need to return anything if we use `OnSuccessVoid`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/ContentService.cs
```diff
@@ -155,14 +158,14 @@ public class ContentService : IContentService
...
-                            .OfType<DataBlock>()
-                            .ToList();
-
-                        dataBlocks.ForEach(dataBlock =>
-                            RemoveContentBlockFromContentSection(sectionToRemove, dataBlock, false));
+                            .ToList()
+                            .ToAsyncEnumerable()
+                            .ForEachAwaitAsync(async contentBlock =>
+                            {
+                                await _contentBlockService.DeleteContentBlockAndReorder(contentBlock.Id, reorder: false);
```
ntsim, line 167: Would suggest we get rid of the `reorder` flag. It's a bit of code smell to have flags that modify behaviour, especially if they do the exact opposite of what the method says.

Another approach would be to define a new `DeleteSectionContentBlocks(Guid contentSectionId)` method in `ContentBlockService` which handles deleting all of a section's content blocks for us without trying to reorder anything afterwards.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/ContentBlockService.cs
```diff
@@ -0,0 +1,66 @@
...
+                    .SingleAsync(cs => cs.Id == originalContentSectionId);
+
+                var contentBlocksRequiringOrderChange = contentSection.Content
+                    .Where(cb => cb.Order > originalBlockOrder)
+                    .ToList();
+
+                _context.ContentBlocks.UpdateRange(contentBlocksRequiringOrderChange);
+                contentBlocksRequiringOrderChange
+                    .ForEach(contentBlock => contentBlock.Order--);
+            }
```
ntsim, line 63: As this is the end of a service method, shouldn't we save any changes?

Otherwise, this puts the responsibility back on the consumer, which could definitely lead to bugs if they don't know how this service is supposed to work.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/EmbedBlockViewModels.cs
```diff
@@ -0,0 +1,27 @@
+#nullable enable
+using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.ViewModels;
```
ntsim, line 5: This file should go in the `Admin.Requests` namespace under `EmbedBlockRequests` (not `EmbedBlockViewModels`)

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/EmbedBlockViewModels.cs
```diff
@@ -0,0 +1,27 @@
...
+    [Required] public Guid ContentSectionId { get; init; }
+}
+
+public record EmbedBlockUpdateRequest
+{
+    [Required] public string Title { get; init; } = string.Empty;
+
+    [Required]
+    [Url]
+    [RegularExpression(@"^https://department-for-education\.shinyapps\.io/.+$", ErrorMessage="URL not permitted")]
```
ntsim, line 25: To save repeating ourselves, we could refactor this so that the pattern is re-used from a constant string e.g.

```cs
public record EmbedBlockCreateRequest
{
    public const string UrlPattern = @"^https://department-for-education\.shinyapps\.io/.+$";
}

public record EmbedBlockUpdateRequest
{
    [RegularExpression(EmbedBlockCreateRequest.UrlPattern, ErrorMessage="URL not permitted")]
    public string Url { get; init; } = string.Empty;
}
```

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/EmbedBlockViewModels.cs
markhiveit: I had to wrap the const variable in a static class so it can be accessed by both records:
```
public static class UrlPattern
{
    public const string Pattern = @"^https://department-for-education\.shinyapps\.io/.+$";
}
```

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/EmbedBlockService.cs
markhiveit: If I turn it into an `OnSuccessVoid` it complains because the same `OnSuccess` also returns a `ValidationErrorMessage` - I could split these particular lines into their own `OnSuccessVoid`, but that would mean that the OnSuccess for L140-148 would need to return a `Unit.Instance` anyway.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/ContentService.cs
markhiveit: I've removed the reorder bool and created a `DeleteSectionContentBlocks` method.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/EmbedBlockViewModels.cs
markhiveit: My bad - missed the `EmbedBlockCreateRequest.UrlPattern`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/EmbedBlockService.cs
ntsim: Ah yeh, of course, d'oh!

ntsim:
@markhiveit looks like you need to push your changes :wink:

ntsim (approved):

PR #3678, opened 2022-11-29
EES-3937 fix reordering methodology sections

amyb-hiveit:
Changing the id to not be the section id (with the guid in) was breaking the reordering, so I've reverted that and added and anchorId instead.

### DIFF
--- a/src/explore-education-statistics-common/src/components/AccordionSection.tsx
+++ b/src/explore-education-statistics-common/src/components/AccordionSection.tsx
@@ -63,6 +63,7 @@ const AccordionSection = ({
 
   const headingId = `${id}-heading`;
   const contentId = `${id}-content`;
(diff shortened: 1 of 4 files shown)


src/explore-education-statistics-common/src/components/AccordionSection.tsx
```diff
@@ -55,14 +55,17 @@ const AccordionSection = ({
   heading,
   headingTag = 'h2',
   id = 'accordionSection',
-  anchorLinkIdPrefix = 'section',
+  anchorLinkIdPrefix,
```
ntsim, line 58: Can we not set a default for this? We need to be able to prevent invalid anchor ids from headings that start with numbers

ntsim (approved):

PR #3682, opened 2022-11-30
EES-3609 find stats search

amyb-hiveit:
Adds search to the new Find Statistics page:

- minimum 3 characters for the search term
- the URL query params are updated to include the search term
- the search can be removed by clicking the tag style button above the search results
- a clear all filters button has been added
- shows a message if there are no search results (I've copied this from the gov.uk search message)
- when searching an extra sort option 'Relevance' is added and selected

search

### DIFF
--- a/src/explore-education-statistics-common/src/components/form/FormLabel.tsx
+++ b/src/explore-education-statistics-common/src/components/form/FormLabel.tsx
@@ -6,12 +6,20 @@ export interface FormLabelProps {
   id: string;
   hideLabel?: boolean;
   label: string | ReactNode;
+  labelSize?: 'xl' | 'l' | 'm' | 's';
 }
 
-const FormLabel = ({ className, id, hideLabel, label }: FormLabelProps) => {
+const FormLabel = ({
+  className,
+  id,
+  hideLabel,
+  label,
+  labelSize,
+}: FormLabelProps) => {
   return (
     <label
       className={classNames('govuk-label', className, {
+        [`govuk-label--${labelSize}`]: labelSize,
         'govuk-visually-hidden': hideLabel,
       })}
       htmlFor={id}
--- a/src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
@@ -1,4 +1,5 @@
 import publicationService, {
+  PublicationListRequest,
   PublicationListSummary,
   PublicationOrderParam,
   PublicationSortOption,
@@ -18,6 +19,7 @@ interface Props {
   newDesign?: boolean; // TODO EES-3517 flag
   paging?: Paging | null; // TODO EES-3517 won't be optional or null
   publications?: PublicationListSummary[]; // TODO EES-3517 won't be optional
+  search?: string;
   sortBy?: PublicationSortOption;
   themes: Theme[];
 }
@@ -26,6 +28,7 @@ const FindStatisticsPage: NextPage<Props> = ({
   newDesign = false,
   paging,
   publications,
+  search,
   sortBy,
   themes = [],
 }) => {
@@ -35,6 +38,7 @@ const FindStatisticsPage: NextPage<Props> = ({
       <FindStatisticsPageNew
         paging={paging}
         publications={publications}
+        searchTerm={search}
         sortBy={sortBy}
       />
     );
@@ -50,11 +54,18 @@ export const getServerSideProps: GetServerSideProps<Props> = async ({
     ? query.sortBy
     : 'newest';
   const { order, sort } = getSortParams(sortBy);
+  const minSearchCharacters = 3;
+  const searchQuery = Array.isArray(query.search)
+    ? query.search[0]
+    : query.search;
+  const search =
+    searchQuery && searchQuery.length >= minSearchCharacters ? searchQuery : '';
 
-  const params = {
+  const params: PublicationListRequest = {
     order,
     page: parseNumber(query.page) ?? 1,
     sort,
+    ...(search && { search }),
   };
 
   // TODO EES-3517 - remove newDesign check
@@ -82,6 +93,7 @@ export const getServerSideProps: GetServerSideProps<Props> = async ({
       newDesign: !!newDesign,
       paging: newDesign ? publicationsResponse.paging : null,
       publications: newDesign ? publicationsResponse.results : [],
+      search,
       sortBy,
       themes,
     },
--- a/src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPageNew.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPageNew.tsx
@@ -1,5 +1,5 @@
 import Button from '@common/components/Button';
-import InsetText from '@common/components/InsetText';
+import ButtonText from '@common/components/ButtonText';
 import LoadingSpinner from '@common/components/LoadingSpinner';
 import RelatedInformation from '@common/components/RelatedInformation';
 import VisuallyHidden from '@common/components/VisuallyHidden';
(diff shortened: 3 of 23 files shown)


src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
```diff
@@ -18,6 +18,7 @@ interface Props {
   newDesign?: boolean; // TODO EES-3517 flag
   paging?: Paging | null; // TODO EES-3517 won't be optional or null
   publications?: PublicationListSummary[]; // TODO EES-3517 won't be optional
+  search?: string;
   sortBy?: PublicationSortOption;
```
benoutram, line 22: We've talked about adding a 'relevance' sort option which becomes the default when using a search term.

I don't think it makes sense to order by any of the other sort options (title or release published date) when using search, or at least you'd expect results to be sorted by relevance by default.

The back-end API has support for this and also 'relevance' is the default option used by the API if a search term is present but no sort is specified.

See the description of https://github.com/dfe-analytical-services/explore-education-statistics/pull/3635

amyb-hiveit: Completely forgot about that, sorry! I'll have a chat with Marv about it.

src/explore-education-statistics-common/src/components/form/FormLabel.tsx
```diff
@@ -6,12 +6,20 @@ export interface FormLabelProps {
...
+  labelSize?: 'xl' | 'l' | 'm' | 's';
 }
 
-const FormLabel = ({ className, id, hideLabel, label }: FormLabelProps) => {
+const FormLabel = ({
+  className,
+  id,
+  hideLabel,
+  label,
+  labelSize,
```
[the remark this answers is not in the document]
amyb-hiveit: I was going to add a default but then I noticed that all of the label size classes, `govuk-label--m` etc, add a font weight which isn't present on `govuk-label`, this means that defaulting to one of them would make all labels bold.

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
amyb-hiveit: Chatted to Marv - I'm going to change it to add the relevance option and select it by default when you have a search term.

src/explore-education-statistics-frontend/src/modules/find-statistics/components/SearchForm.module.scss
```diff
@@ -0,0 +1,45 @@
+@import '~govuk-frontend/govuk/base';
+
+.button {
+  background: $govuk-brand-colour;
+  border: 2px solid govuk-colour('black');
+  border-left: 0;
```
ntsim, line 6: This border is a little inconsistent with [GOV.UK](https://www.gov.uk/):

image

Would suggest we remove the border? The current implementation is also a little inconsistent as we don't typically have borders around buttons anyway.

src/explore-education-statistics-frontend/src/modules/find-statistics/components/__tests__/SearchForm.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+    render(<SearchForm onSubmit={noop} />);
+    expect(screen.getByLabelText('Search')).toBeInTheDocument();
+    expect(screen.getByRole('button', { name: 'Search' })).toBeInTheDocument();
+  });
+
+  test('renders correctly with an initial search term', () => {
+    render(<SearchForm searchTerm="find me" onSubmit={noop} />);
+    expect(screen.getByLabelText('Search')).toHaveValue('find me');
+  });
+
```
ntsim, line 18: Would suggest adding a test case where the validation error is removed once the input becomes valid. We should be a little more rigorous here as we're not backed up by Formik automatically handling stuff like this for us.

src/explore-education-statistics-frontend/src/modules/find-statistics/components/FilterTag.tsx
```diff
@@ -0,0 +1,18 @@
+import VisuallyHidden from '@common/components/VisuallyHidden';
+import styles from '@frontend/modules/find-statistics/components/FilterTag.module.scss';
+import React from 'react';
+
+interface Props {
+  name: string;
+  onClick: () => void;
+}
+
+const FilterTag = ({ name, onClick }: Props) => {
```
ntsim, line 10: Not sure this name is accurate, perhaps `FilterClearButton` would be better?

src/explore-education-statistics-frontend/src/modules/find-statistics/components/FilterTag.module.scss
```diff
@@ -0,0 +1,24 @@
...
+  background: govuk-colour('light-grey');
+  border: 1px solid govuk-colour('mid-grey');
+  border-radius: 6px;
+  box-shadow: none;
+  line-height: 1;
+  margin: 0 govuk-spacing(1) 0 0;
+  padding: 8px;
+
+  &::before {
+    content: '✕';
```
ntsim, line 13: This kind of content can still show up for a screen reader. Would suggest we put this directly into a `span` element that's aria-hidden.

```tsx
<span aria-hidden>✕</span>
```

src/explore-education-statistics-frontend/src/modules/find-statistics/components/FilterTag.module.scss
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  &:focus {
+    background: $govuk-focus-colour;
+  }
+
+  &:hover {
+    background: darken(govuk-colour('light-grey'), 4);
+    cursor: pointer;
```
ntsim, line 22: We could just put this directly on the `.tag` / `.button` class

src/explore-education-statistics-frontend/src/modules/find-statistics/components/FilterTag.module.scss
```diff
@@ -0,0 +1,24 @@
+@import '~govuk-frontend/govuk/base';
+
+.tag {
```
ntsim, line 3: Might want to name this `.button` if we rename the component (see other comment)

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPageNew.tsx
```diff
@@ -120,18 +185,56 @@ const FindStatisticsPageNew: NextPage<Props> = ({
...
-            <p className="govuk-!-margin-top-1">
-              {`Page ${page} of ${paging.totalPages}, showing all publications`}
-              <VisuallyHidden>{` sorted by ${sortBy}`}</VisuallyHidden>
-            </p>
+            <div className="dfe-flex dfe-justify-content--space-between dfe-align-items-start ">
+              <div>
+                <p className="govuk-!-margin-bottom-2">
+                  {`${
+                    totalResults
+                      ? `Page ${page} of ${paging.totalPages}`
```
ntsim, line 211: Could destructure `totalPages` out of `paging`

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPageNew.tsx
```diff
@@ -120,18 +185,56 @@ const FindStatisticsPageNew: NextPage<Props> = ({
...
+            >
               {`${totalResults} ${totalResults !== 1 ? 'results' : 'result'}`}
             </h2>
 
-            <p className="govuk-!-margin-top-1">
-              {`Page ${page} of ${paging.totalPages}, showing all publications`}
-              <VisuallyHidden>{` sorted by ${sortBy}`}</VisuallyHidden>
-            </p>
+            <div className="dfe-flex dfe-justify-content--space-between dfe-align-items-start ">
+              <div>
```
ntsim, line 207: Don't think we need this?

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPageNew.tsx
```diff
@@ -140,14 +243,32 @@ const FindStatisticsPageNew: NextPage<Props> = ({
 
           {isMobileMedia && <Button>Filter results</Button>}
 
-          <SortControls sortBy={sortBy} onChange={handleSortPublications} />
+          <SortControls
+            hasSearch={!!searchTerm}
+            sortBy={sortBy}
+            onChange={handleSortPublications}
+          />
```
ntsim, line 250: If there aren't any results, maybe we could just hide these controls?

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPageNew.tsx
```diff
@@ -59,6 +65,65 @@ const FindStatisticsPageNew: NextPage<Props> = ({
     });
   };
 
+  const handleSearchPublications = async (nextSearchTerm: string) => {
+    await router.push(
```
ntsim, line 69: Would be a good idea to extract out a function so that we don't repeat ourselves and can ensure we're always correctly calling `router.push`. Something like this might work:

```ts
const updateQueryParams = (query: FindStatisticsPageQuery) => {
    await router.push(
      {
        pathname: '/find-statistics',
        query: query as ParsedUrlQuery,
      },
      undefined,
      {
        scroll: false,
      },
    );
}

// Then we would use it like:

const handleSearchPublication = async (nextSearchTerm: string) => {
  await updateQueryParams({...});
}
```

Note that I've suggested we define a new type `FindStatisticsPageQuery`. This would contain all the query parameters that can be used on this page and we would add this as a type parameter `getServerSideProps` as well:

```ts
export interface FindStatisticsPageQuery {
  ...
}

export const getServerSideProps: GetServerSideProps<Props, FindStatisticsPageQuery> = ({
  query
}) => {
  ...
}
```

ntsim (changes requested):

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPageNew.tsx
```diff
@@ -59,6 +79,40 @@ const FindStatisticsPageNew: NextPage<Props> = ({
...
+  const handleSearchPublications = async (nextSearchTerm: string) => {
+    await updateQueryParams({
+      ...omit(router.query, 'page'),
+      search: nextSearchTerm,
+      sortBy: 'relevance',
+    });
+
+    logEvent({
+      category: 'Find statistics and data',
+      action: 'Publications searched',
```
luke-h1, line 91: Just wondering if it might be useful to include what publication search term the user has specified?

amyb-hiveit: It sends the search term as the event label.

luke-h1: At right my bad, didn't notice at first glance

PR #3683, opened 2022-11-30
EES-3941 - fix chart tests

luke-h1:
This PR fixes chart visual tests (namely around the chart tests not being able to find a selector that's been removed. 

Other changes
* fixes import + path issues 
* fixes styling issues
* Updates visual testing README
* Updates linting rules

### DIFF
--- a/tests/robot-tests/scripts/visual-testing/generate_tables_and_charts_test_cases.py
+++ b/tests/robot-tests/scripts/visual-testing/generate_tables_and_charts_test_cases.py
@@ -3,57 +3,58 @@ import csv
 import os
 import shutil
 
-from tests.libs.logger import get_logger
-
-logger = get_logger(__name__)
-
-ap = argparse.ArgumentParser()
-ap.add_argument(
-    "-f", "--file", required=True, help="the datablocks csv to use to generate the robot test script from the template"
-)
-ap.add_argument("-t", "--target", required=True, help="the target filename of the generated robot test script")
-args = vars(ap.parse_args())
-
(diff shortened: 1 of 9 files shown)


tests/robot-tests/scripts/visual-testing/generate_tables_and_charts_test_cases.py
```diff
@@ -3,57 +3,62 @@
...
+            generated_robot_file.write(generated_test_script)
 
-    logger.warn(generated_test_script)
+        shutil.copy(datablocks_csv_filepath, f"{robot_tests_folder}/{target_robot_test_csv_filename}")
 
-    with open(f"{robot_tests_folder}/{target_robot_test_filename}", "w", encoding="utf-8-sig") as generated_robot_file:
-        generated_robot_file.write(generated_test_script)
 
-    shutil.copy(datablocks_csv_filepath, f"{robot_tests_folder}/{target_robot_test_csv_filename}")
+main()
```
markhiveit, line 64: How come wrap everything in a main function here?

If this file is only meant to be called directly, then you could put this `main()` call in an `if __name__ == "__main__":` to protect the code from accidentally getting run by a different file that imports this file. If that's not necessary, I'd remove the `main` function.

All that said, I don't think it's a big deal either way, since I highly doubt we're going to import these files into other files and accidentally end up running this code.

luke-h1: No reason as such, I'll wrap the function in `if "__name__ == "__main__"`

PR #3688, opened 2022-12-05
EES-3610 find stats filtering

amyb-hiveit:
Adds filtering to the new Find Statistics page:
- filter by theme
- filter by release type
- modal style filters on mobile (I've added a `fullscreen` option to the modal component for this)
- selected filters are shown above the publications list and can be removed individually or all at once.

**Desktop:**
findstats-fitlers-desktop

**Mobile:**
findstats-filters-mobile


src/explore-education-statistics-frontend/src/modules/find-statistics/components/FilterModals.tsx
```diff
@@ -0,0 +1,100 @@
...
+    open={open}
+    title="Release types guidance"
+    onExit={onClose}
+  >
+    <p>
+      This is a description list of the different types of publication we
+      currently publish.
+    </p>
+    {/* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */}
+    <div className={styles.content} tabIndex={0}>
```
ntsim, line 27: Does this need to be tabbable?

Is it related to the overflow styling? Why not allow the modal fill the entire screen and just tweak it so that the modal can be scrolled?

src/explore-education-statistics-frontend/src/modules/find-statistics/components/FilterModals.tsx
```diff
@@ -0,0 +1,100 @@
...
+}) => (
+  <Modal
+    className="govuk-!-width-one-half"
+    open={open}
+    title="Themes guidance"
+    onExit={onClose}
+  >
+    <p>This is a description list of our different publication themes.</p>
+    {/* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */}
+    <div className={styles.content} tabIndex={0}>
```
ntsim, line 87: Same as above - does this need to be tabbable?

src/explore-education-statistics-frontend/src/modules/find-statistics/components/ReleaseTypeFilters.tsx
```diff
@@ -0,0 +1,59 @@
...
+  return (
+    <form id="releaseTypeFilters">
+      <FormRadioGroup
+        hint={
+          <ButtonText onClick={toggleModal.on}>
+            What are release types?
+          </ButtonText>
+        }
+        id="releaseType"
+        legend="Filter by release type"
```
ntsim, line 37: The UI is a little busy with headings. Would suggest we use `legendHidden` to hide this. We would get this:

image

src/explore-education-statistics-frontend/src/modules/find-statistics/components/ReleaseTypeFilters.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const ReleaseTypeFilters = ({
+  releaseType: initialReleaseType = 'all',
+  onChange,
+}: Props) => {
+  const [showModal, toggleModal] = useToggle(false);
+  const releaseTypeOptions = [
+    { label: 'Show all', value: 'all' },
+    ...Object.keys(releaseTypes).map(type => ({
+      label: releaseTypes[type as keyof typeof releaseTypes],
```
ntsim, line 23: This should be `releaseTypes[type as ReleaseType]`

src/explore-education-statistics-frontend/src/modules/find-statistics/components/ReleaseTypeFilters.tsx
```diff
@@ -0,0 +1,59 @@
...
+        legend="Filter by release type"
+        name="releaseType"
+        options={releaseTypeOptions}
+        order={[]}
+        small
+        value={initialReleaseType}
+        onChange={e => {
+          onChange({
+            filterType: 'releaseType',
+            nextValue: e.target.value as ReleaseType,
```
ntsim, line 46: Not valid to cast this to `ReleaseType` as it can be 'all'. Think we could just remove the cast entirely as `nextValue` should just be a string

src/explore-education-statistics-frontend/src/modules/find-statistics/components/MobileFilters.tsx
```diff
@@ -0,0 +1,51 @@
...
+  onClose,
+}: Props) => {
+  const totalResultsString = `${totalResults} ${
+    totalResults !== 1 ? 'results' : 'result'
+  }`;
+
+  return (
+    <Modal fullScreen hideTitle title="Filter publications" onExit={onClose}>
+      {/* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */}
+      <div className={styles.container} tabIndex={0}>
```
ntsim, line 36: Does this need to be tabbable? It's not great in terms of the keyboard experience as the entire modal shows a focus on rendering. Upon tabbing again, we get a really faint second outline:

image

I actually didn't notice this outline initially so it was like the Tab press had been swallowed by some invisible element.

Would suggest we remove this so there's only a single focus when the modal renders. Otherwise, it's a little annoying to have to Tab twice to reach the actual content.

If this is related to the scroll container, would suggest that we also make the modal itself scrollable when it's in full screen mode.

src/explore-education-statistics-frontend/src/modules/find-statistics/components/__tests__/ThemeFilters.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+      />,
+    );
+
+    expect(screen.getByLabelText('Theme 3')).toBeChecked();
+  });
+
+  test('calls onChange when a theme is selected', () => {
+    const handleChange = jest.fn();
+    render(<ThemeFilters themes={testThemesList} onChange={handleChange} />);
+
```
ntsim, line 50: Not a biggy, but might want to test that `handleChange` hasn't been called yet to confirm it's the click that causes the handler to be called

ntsim (changes requested):

src/explore-education-statistics-frontend/src/modules/find-statistics/components/FilterModals.tsx
amyb-hiveit: as above

src/explore-education-statistics-frontend/src/modules/find-statistics/components/MobileFilters.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <div className={styles.header}>
```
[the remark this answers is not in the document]
amyb-hiveit: Yeah, I've added a border to the header so it looks like part of focus outline.

src/explore-education-statistics-frontend/src/modules/find-statistics/components/MobileFilters.tsx
amyb-hiveit: yeah you're right, I've taken that our and made the fullscreen modal scrollable.

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPageNew.tsx
```diff
@@ -226,7 +292,16 @@ const FindStatisticsPageNew: NextPage<Props> = ({
             Skip to search results
```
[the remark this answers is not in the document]
amyb-hiveit: yeah good point, it's not really needed so I've removed it now.

amyb-hiveit:
@ntsim I've made changes base on your feedback and also changed the filtering, search and pagination to be done client side to prevent unnecessary calls to the theme service as discussed. Filtering will still work server side as well.

I tried to test that changing the filters etc updates the publications but couldn't figure out how to get the query to update in the test. Any ideas on the best way to approach this?

ntsim (changes requested):

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPageNew.tsx
```diff
@@ -2,71 +2,108 @@ import Button from '@common/components/Button';
...
+      setPublications(updatedPublications.results);
+      setPaging(updatedPublications.paging);
+    }
+    if (
+      Object.keys(router.query).length &&
+      !isEqual(router.query, currentQuery)
+    ) {
+      setQuery(router.query);
+      fetchPublications(createPublicationListRequest(router.query));
+    }
```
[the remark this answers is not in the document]
amyb-hiveit: @ntsim I did think about using react-query originally and thought it might be overkill here, but have given it a go and it works well. 
I ran out of time today so haven't got very far with the tests, I'll pick that back up on Tuesday. If you can review the changes for react-query in the meantime that'd be good, I've kept it in a separate commit for now so it's clear what's changed.

src/explore-education-statistics-frontend/src/modules/find-statistics/__tests__/FindStatisticsPageNew.test.tsx
```diff
@@ -180,270 +200,282 @@ describe('FindStatisticsPageNew', () => {
...
+  //     publicationsList.getByRole('heading', { name: 'Publication 3' }),
+  //   ).toBeInTheDocument();
+
+  //   expect(
+  //     screen.queryByText('There are no matching results.'),
+  //   ).not.toBeInTheDocument();
+  // });
 });
+
+// Can't set using the common render to work, so doing it here instead.
```
ntsim, line 473: We'll want to sort this out so that we can use the common `render` function as it's pre-configured for use in tests.

If it's issues around using the same instance of `react-query`, due to how the project's currently setup, the `react-query` used as a dependency in common is not going to be the exact same one used in frontend.

To fix this, we'll need to add a `moduleNameMapper` to the Jest config in the frontend's `package.json`.

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
```diff
@@ -54,35 +37,34 @@ export const getServerSideProps: GetServerSideProps<Props> = async ({
...
+  const queryClient = new QueryClient();
+
+  // TODO EES-3517 - remove newDesign check
+  if (newDesign) {
+    await Promise.all([
+      queryClient.prefetchQuery(['listPublications', query], () =>
+        publicationService.listPublications(
+          createPublicationListRequest(query),
+        ),
+      ),
```
ntsim, line 56: Once we start using React Query more, managing query and their keys can be a bit problematic as it can lead to incorrect fetching from the cache if there are naming collisions.

We also want to coordinate keys/queries so that:

- we can re-use them easily in multiple places (e.g. in the server and client)
- they can be easily refactored in the future

There's a few ways to approach this, but the recommendation from the docs is to use the query key factory pattern. 

At it's crux, it consists of something like this:

```ts
// queries/publications.ts
import { useQuery, UseQueryOptions } from '@tanstack/react-query';

const publicationQueries = {
  list(query: ParsedUrlQuery): UseQueryOptions {
    return {
      queryKey: ['listPublications', query],
      queryFn: () =>
        publicationService.listPublications(
          createPublicationListRequest(query),
        ),
    };
  },
} as const;

export default publicationQueries;
```

Then we would use it, like so:

```ts
// getServerSideProps
queryClient.prefetchQuery(publicationQueries.list(query)),

// client-side
useQuery(publicationQueries.list(router.query));
useQuery({
  ...publicationQueries.list(router.query),
  keepPreviousData: true,
  staleTime: 60000,
});
```

This pattern is very similar to how we define services, but we're just defining these factory functions that generate query options instead.

Alternatively, if we really want to standardise the crap out of this, we could use the following package: https://github.com/lukemorales/query-key-factory

There's some benefit to using this package, but I'm not totally convinced we need it right away. We're probably not doing enough with React Query to warrant the additional dependency just yet. However, if you like what you see (would advise you to read the docs for some inspiration), happy to bring it in as well.

src/explore-education-statistics-frontend/src/modules/find-statistics/__tests__/FindStatisticsPageNew.test.tsx
amyb-hiveit: Sorted now, thanks for the tip about adding a moduleNameMapper for react-query.

The tests all pass but I'm getting a warning on them, `It looks like you're using the wrong act() around your test interactions.`, related to next/link which I can't figure out how to get rid of. Have you come across this before?

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
amyb-hiveit: Cool, I've done this now. Didn't use query-key-factory, but definitely something to consider later on if we use react-query in more places.

src/explore-education-statistics-frontend/src/modules/find-statistics/__tests__/FindStatisticsPageNew.test.tsx
```diff
@@ -446,4 +453,402 @@ describe('FindStatisticsPageNew', () => {
...
+    });
+
+    expect(sortOptions[0]).toEqual(sortGroup.getByLabelText('Newest'));
+    expect(sortOptions[0]).not.toBeChecked();
+    expect(sortOptions[1]).toEqual(sortGroup.getByLabelText('Oldest'));
+    expect(sortOptions[1]).not.toBeChecked();
+    expect(sortOptions[2]).toEqual(sortGroup.getByLabelText('A to Z'));
+    expect(sortOptions[2]).toBeChecked();
+  });
+
```
ntsim, line 789: Would be good to get some tests in here for at least the following (there might be some others in this general area):

- sorting with search (i.e. relevance)
- removing relevance sorting when search is cleared

ntsim (approved):

PR #3705, opened 2022-12-08
EES-3951 Footnotes tidy up

benoutram:
This is a back-end Footnotes tidy up to clean up a few things prior to beginning work on [EES-167](https://dfedigital.atlassian.net/browse/EES-167) which will add the ability to order footnotes. (Update the ordering changes in https://github.com/dfe-analytical-services/explore-education-statistics/pull/3713 have been approved and are now merged into this branch).

- `FootnotesRepository.GetFilteredFootnotes` is changed from using a stored procedure to be written entirely in EF Core. The reason for changing this is that as new columns get added to the footnote table they immediately break the stored procedure call in `GetFilteredFootnotes` which expects the result set to contain all columns of Footnote. Therefore adding `Order` would require changing it slightly anyway  and I noticed that it seems a bit more complicated than I think it needs to be, and just confusing with all of the table aliases. I have drafted an alternative version of the stored procedure but rather than go with that I thought we may as well switch the query to be written in EF Core which can then be unit tested more easily.
- `IFootnoteService` and `IFootnoteRepository` methods changed to return `List<Footnote>` rather than `IEnumerable` to allow preserving order when footnotes are ordered in future.
- `IFootnoteService` and `IFootnoteRepository` methods changed to be async to avoid mixing async and non async methods.
- `Footnote.Content` database column made not nullable.
- `FootnoteRepository.DeleteFootnote` - Simplified by changing foreign keys from FilterFootnote, FilterGroupFootnote, FilterItemFootnote, IndicatorFootnote, and SubjectFootnote tables to use cascade delete. Removed method `DeleteEntities`.
- `FootnoteServicePermissionTests` - Updated to use `PolicyCheckBuilder` rather than obsolete `AssertSecurityPoliciesChecked` method. Extended coverage by adding missing permissions tests.
- `FootnoteViewModel` changed from class to record and now simplfied by using a primary constructor.
- `FootnoteController.GetFootnotesMeta` - Changed to no longer depend on `EnumerableExtensions.SelectAsync`.
- `FootnoteService.CopyFootnotes` - Changed to no longer depend on `EnumerableExtensions.ForEachAsync`.
- `FootnoteService.CreateFootnotes` - Moved most of the body of this method to `FootnoteRepository` so that it can be reused by `CopyFootnotes` without doing a release permission check copying each individual footnote. Also simplified the way we build the links in new footnote when it's created.
- `FootnoteCreateViewModel` and `FootnoteUpdateViewModel` - Replaced with `FootnoteCreateRequest` and `FootnoteUpdateRequest` which have sets of id's rather than lists to prevent duplicates, after PR feedback.

### UI test report 

image


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/FootnoteService.cs
```diff
@@ -60,83 +56,70 @@ public class FootnoteService : IFootnoteService
...
-                    return footnote;
-                })
+                .OnSuccess(() => _footnoteRepository.CreateFootnote(
+                    releaseId,
+                    content,
+                    filterIds,
+                    filterGroupIds,
+                    filterItemIds,
+                    indicatorIds,
+                    subjectIds))
```
ntsim, line 66: Would suggest using named parameters here given that we have so many lists of ids. We wouldn't want to accidentally assign the wrong one if the parameter order changes.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/FootnoteService.cs
```diff
@@ -60,83 +56,70 @@ public class FootnoteService : IFootnoteService
...
+                            var subjects = footnote.Subjects
+                                .Select(subjectFootnote => subjectFootnote.SubjectId).ToList();
+
+                            return await _footnoteRepository.CreateFootnote(destinationReleaseId,
+                                footnote.Content,
+                                filters,
+                                filterGroups,
+                                filterItems,
+                                indicators,
+                                subjects);
```
ntsim, line 100: Same as above - would suggest using named parameters to avoid mixing up any of the id lists by accident.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/FootnoteService.cs
```diff
@@ -60,83 +56,70 @@ public class FootnoteService : IFootnoteService
...
+                            var filters = footnote.Filters
+                                .Select(filterFootnote => filterFootnote.FilterId).ToList();
+                            var filterGroups = footnote.FilterGroups
+                                .Select(filterGroupFootnote => filterGroupFootnote.FilterGroupId).ToList();
+                            var filterItems = footnote.FilterItems
+                                .Select(filterItemFootnote => filterItemFootnote.FilterItemId).ToList();
+                            var indicators = footnote.Indicators
+                                .Select(indicatorFootnote => indicatorFootnote.IndicatorId).ToList();
+                            var subjects = footnote.Subjects
+                                .Select(subjectFootnote => subjectFootnote.SubjectId).ToList();
```
ntsim, line 92: Not a biggy, but could suffix these with `Ids` e.g. `filterIds` to name them more appropriately.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
```diff
@@ -19,40 +18,134 @@ public FootnoteRepository(StatisticsDbContext context)
...
+                    FilterItemId = filterItemId
+                }).ToList(),
+                Indicators = indicatorIds.Select(indicatorId => new IndicatorFootnote
+                {
+                    IndicatorId = indicatorId
+                }).ToList(),
+                Subjects = subjectIds.Select(subjectId => new SubjectFootnote
+                {
+                    SubjectId = subjectId
+                }).ToList()
```
ntsim, line 58: Seems like it's possible for the code to attempt to insert multiple links with the same ids i.e. dupes.

I'm guessing this would get picked up by the DB when it tries to create the links, but I'd be tempted to ensure that we don't get into this situation by filtering out any dupe ids.

We could either do this by calling `Distinct().Select(...)` on each link, or change the method parameters to use `IReadOnlySet<Guid>` (the consuming code would have to create the sets beforehand).

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
```diff
@@ -19,40 +18,134 @@ public FootnoteRepository(StatisticsDbContext context)
...
+                                       filterItems.Contains(filterItemFootnote.FilterItemId)))
+                .Where(footnote => !footnote.FilterGroups.Any() ||
+                                   footnote.FilterGroups.Any(filterGroupFootnote =>
+                                       filterGroups.Contains(filterGroupFootnote.FilterGroupId)))
+                .Where(footnote => !footnote.Filters.Any() ||
+                                   footnote.Filters.Any(filterFootnote =>
+                                       filters.Contains(filterFootnote.FilterId)))
+                .Where(footnote => !footnote.Indicators.Any() ||
+                                   footnote.Indicators.Any(indicatorFootnote => 
+                                       indicators.Contains(indicatorFootnote.IndicatorId)))
```
ntsim, line 123: It's possible that either `filterItems` or `indicators` is empty, meaning it's possible to construct a `Where` against an empty list. 

I can't remember if EF protects against this. If it doesn't, we end up SQL like `WHERE IN ()` which always throws an SQL exception. 

Would suggest better protecting against this if EF doesn't automatically handle potentially empty lists.

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/FootnoteService.cs
benoutram: Have made this change now as part of the new commit altering from lists to sets, and done it in quite a few other places, thanks.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/FootnoteService.cs
benoutram: Made this change now as part of the new commit altering from lists to sets.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
benoutram: Decided to go with `IReadOnlySet<Guid>` all the way up to the request types `FootnoteCreateRequest`  and `FootnoteUpdateRequest`. Thanks.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
```diff
@@ -19,40 +18,134 @@ public FootnoteRepository(StatisticsDbContext context)
...
+                .ToListAsync();
+
+            var filters = await filterItemsQueryable
+                .Select(filterItem => filterItem.FilterGroup.FilterId)
+                .Distinct()
+                .ToListAsync();
+
+            var indicators = await _context.Indicator
+                .Where(indicator => indicatorIds.Contains(indicator.Id)
+                                    && indicator.IndicatorGroup.SubjectId == subjectId)
```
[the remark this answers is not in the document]
benoutram: Ah yeah, good spot! I'm surprised we've never seen this but I suppose it's because the table tool always requires at least one indicator to be selected. I've made this change in `ResultSubjectMetaService` on a new commit, thanks.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
```diff
@@ -19,40 +18,134 @@ public FootnoteRepository(StatisticsDbContext context)
...
+                .Select(indicator => indicator.Id)
+                .ToListAsync();
```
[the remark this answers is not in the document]
benoutram: I don't think we can return early because if both collections are empty we still want to return footnotes that are applied directly to the subject (i.e. those that have a `SubjectFootnote` matching the `subjectId` parameter).

I will add some unit tests to check it works in this case though, thanks.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20221208130742_EES-167_Footnotes_TidyUp.cs
```diff
@@ -0,0 +1,181 @@
...
+
+        migrationBuilder.DropForeignKey(
+            name: "FK_IndicatorFootnote_Footnote_FootnoteId",
+            table: "IndicatorFootnote");
+
+        migrationBuilder.DropForeignKey(
+            name: "FK_SubjectFootnote_Footnote_FootnoteId",
+            table: "SubjectFootnote");
+
+        // Recreate the same foreign keys but this time with a cascade delete action
```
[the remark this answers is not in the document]
benoutram: I'm thinking the same, just an oversight and code we haven't reviewed recently. I'm planning to do some additional manual testing on this though before it get's merged! Thanks.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
```diff
@@ -19,40 +18,134 @@ public FootnoteRepository(StatisticsDbContext context)
...
+
+            var filterGroups = await filterItemsQueryable
+                .Select(filterItem => filterItem.FilterGroupId)
+                .Distinct()
+                .ToListAsync();
+
+            var filters = await filterItemsQueryable
+                .Select(filterItem => filterItem.FilterGroup.FilterId)
+                .Distinct()
+                .ToListAsync();
```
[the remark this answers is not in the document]
benoutram: I think you are right, it can be done quite nicely all in one, so no need for the 3 separate queries.

I've changed this on the original commit.

I plan to unit test it as part of addressing other comments on this method in this PR and will review the translated SQL query, thanks.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
benoutram: I'm going to unit test and manually test this method to try and trip it up, especially after other changes have been made following comments on this PR. Will report back here when unit tests are sorted, thanks.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
benoutram: By the way, I've added tests now for creating footnotes, not in `FootnoteRepositoryTest` but testing it indirectly through `FootnoteServiceTests` in the Footnotes Ordering PR - https://github.com/dfe-analytical-services/explore-education-statistics/pull/3713/files#diff-1959d3889e4233e87d3fbd3aec8c7e32b54b035b3d563467c05c29adcf040283

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
benoutram: Added lots of unit tests which did show up a problem here, not with empty collections and a `WHERE IN ()` but with all the logic between the different `.Where` clauses.

I've simplified the logic now which has made the unit tests happy!

We still end up with lots of conditions like this

```
                footnote.FilterItems.Any(filterItemFootnote =>
                    filterItemIdsSanitised.Contains(filterItemFootnote.FilterItemId)) ||
```

We could prefix these with a check to make sure the lists aren't empty, e.g.

```
                filterItemIdsSanitised.Any() && footnote.FilterItems.Any(filterItemFootnote =>
                    filterItemIdsSanitised.Contains(filterItemFootnote.FilterItemId)) ||
```

but that doesn't seem to be necessary as EF is automatically protecting against this.

benoutram:
Updated UI test report

image

src/GovUk.Education.ExploreEducationStatistics.Data.Model.Tests/Repository/FootnoteRepositoryTests.cs
```diff
@@ -1,16 +1,1222 @@
...
+                // Footnote applies to a requested indicator 
+                Assert.Equal(releaseFootnotes[4].FootnoteId, results[4].Id);
+                Assert.Equal("Footnote 5", results[4].Content);
+
+                // Footnote applies to a various criteria related to the requested filter items and indicators
+                Assert.Equal(releaseFootnotes[5].FootnoteId, results[5].Id);
+                Assert.Equal("Footnote 6", results[5].Content);
+            }
+        }
+
```
ntsim, line 303: Might be a good idea to add other test cases for:

- returning footnotes that match only one filter item / filter group / filter (from a selection)
- returning footnotes that match only one indicator (from a selection)

In the above test, we're providing all the criteria, and returning all the footnotes. This means we can't be absolutely certain that the method isn't just returning _all_ the subject's footnotes by default.

Some of the other test cases semi-cover this, but it'd be good to have dedicated test cases for this it to be more airtight.

src/GovUk.Education.ExploreEducationStatistics.Data.Model.Tests/Repository/FootnoteRepositoryTests.cs
```diff
@@ -6,12 +6,2253 @@
...
+                Assert.Equal(releaseFootnotes[8].FootnoteId, results[5].Id);
+                Assert.Equal("Applies to filter 1 group 1 and indicator 1", results[5].Content);
+
+                Assert.Equal(releaseFootnotes[10].FootnoteId, results[6].Id);
+                Assert.Equal("Applies to filter item 1 and indicator 1", results[6].Content);
+            }
+        }
+
+        [Fact]
+        public async Task GetFilteredFootnotes_FilterByFilters()
```
duncan-at-hiveit, line 669: Should this be called "GetFilteredFootnotes_FilterByFilterItems"?  Only because from the test setup it seems that we're passing `filterItemIds` as the main filtering parameter.

src/GovUk.Education.ExploreEducationStatistics.Data.Model.Tests/Repository/FootnoteRepositoryTests.cs
```diff
@@ -6,12 +6,2253 @@
...
+                var results = await repository.GetFilteredFootnotes(
+                    releaseId: release.Id,
+                    subjectId: subject1.Id,
+                    filterItemIds: ListOf<Guid>(),
+                    indicatorIds: ListOf(indicatorGroup1.Indicators[0].Id,
+                        indicatorGroup1.Indicators[1].Id)
+                );
+
+                // Check that only footnotes applying to indicator 1 are returned.
+                // No footnotes apply to indicator 2 but including its id in the parameter shouldn't alter the result.
```
duncan-at-hiveit, line 1237: Good idea to throw this scenario into the mix!

src/GovUk.Education.ExploreEducationStatistics.Data.Model.Tests/Repository/FootnoteRepositoryTests.cs
```diff
@@ -6,12 +6,2253 @@
...
+                var repository = BuildFootnoteRepository(context);
+
+                //  Get the footnotes applying to subject 1 or any of its filter items or indicators
+                var results = await repository.GetFilteredFootnotes(
+                    releaseId: release.Id,
+                    subjectId: subject1.Id,
+                    filterItemIds: ListOf(subject1Filter1.FilterGroups[0].FilterItems[0].Id),
+                    indicatorIds: ListOf(subject1IndicatorGroup1.Indicators[0].Id));
+
+                // Check that only footnotes applying to subject 1 or any of its filter items or indicators are returned
```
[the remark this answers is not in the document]
duncan-at-hiveit: I guess here we're getting back ALL Footnotes that were originally put in, so it's hard to tell what the actual effects of the input parameters had.  But I think that the coverage of the other test cases cover our backs.  Just not 100% sure if you need to retain this particular test case.

src/GovUk.Education.ExploreEducationStatistics.Data.Model.Tests/Repository/FootnoteRepositoryTests.cs
```diff
@@ -6,12 +6,2253 @@
...
+                            }
+                        }
+                    }
+                }
+            };
+
+            // Create another release which has no footnotes
+            var release2 = new Release();
+
+            var releaseSubjects = new List<ReleaseSubject>
```
duncan-at-hiveit, line 1930: Is it worth adding a note here to say that we hope to be removing this ReleaseSubject link table soon, and so going forwards from that point this test case can be removed?

Just thinking that if we don't get round to that any time soon, and then when we do we run up against this explicit test case, we might second guess whether or not we should be removing the link table!

Worth mentioning at least that we *shouldn't* be seeing this kind of data under normal circumstances?

src/GovUk.Education.ExploreEducationStatistics.Data.Model.Tests/Repository/FootnoteRepositoryTests.cs
```diff
@@ -6,12 +6,2253 @@
...
+                // The filter item and indicator id's related to subject 1 should have been ignored
+                // Footnotes applying to subject 1 should be excluded
+                // The footnotes applying to specific data of subject 2 should also be excluded
+                Assert.Single(results);
+
+                Assert.Equal(releaseFootnotes[5].FootnoteId, results[0].Id);
+                Assert.Equal("Applies to subject 2", results[0].Content);
+            }
+        }
+
```
duncan-at-hiveit, line 2255: Great set of new tests!

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
```diff
@@ -78,25 +76,100 @@ public FootnoteRepository(StatisticsDbContext context)
...
         {
-            var releaseIdParam = new SqlParameter("releaseId", releaseId);
-            var subjectIdParam = new SqlParameter("subjectId", subjectId);
-            var filterItemListParam = CreateIdListType("filterItemList", filterItemIds);
-            var indicatorListParam = CreateIdListType("indicatorList", indicatorIds);
+            // Query the id's of the filter groups and filters corresponding with the filter item id's
+            // Sanitise all id's to make sure they belong to the subject
+            var filterIdsResult = await _context.FilterItem
+                .Where(filterItem => filterItemIds.Contains(filterItem.Id)
+                                     && filterItem.FilterGroup.Filter.SubjectId == subjectId)
```
duncan-at-hiveit, line 83: I thought you'd need to do a bunch of "Include" and "ThenInclude" to get this lookup to work: `filterItem.FilterGroup.Filter.SubjectId`?

But I'm also guessing that no way would the unit tests be working if you didn't have this right :)

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
```diff
@@ -78,25 +76,100 @@ public FootnoteRepository(StatisticsDbContext context)
...
+            }
+
+            // A footnote which applies directly to any subjects must include this subject id
+            footnotesQueryable = footnotesQueryable.Where(footnote =>
+                !footnote.Subjects.Any() || footnote.Subjects.Any(subjectFootnote =>
+                    subjectFootnote.SubjectId == subjectId));
+
+            // A footnote which applies to any indicators of this subject must include one of the indicator id's
+            footnotesQueryable = footnotesQueryable.Where(footnote =>
+                footnote.Indicators.All(indicatorFootnote =>
```
duncan-at-hiveit, line 145: Can't quite get my head round this left-hand part of this "or" clause - we're saying to include footnotes where all of the indicators belong to a different subject?

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
```diff
@@ -78,25 +76,100 @@ public FootnoteRepository(StatisticsDbContext context)
...
+                    indicatorFootnote.Indicator.IndicatorGroup.SubjectId != subjectId) ||
+                footnote.Indicators.Any(indicatorFootnote =>
+                    indicatorIdsSanitised.Contains(indicatorFootnote.IndicatorId)));
+
+            // A footnote which applies to any filter items of this subject must include one of the the filter item id's 
+            footnotesQueryable = footnotesQueryable.Where(footnote =>
+                footnote.FilterItems.All(filterItemFootnote =>
```
duncan-at-hiveit, line 152: Same as above with indicators

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Statistics/FootnoteControllerTests.cs
```diff
@@ -171,49 +169,46 @@ public FootnoteControllerTests()
...
-                Subjects = new List<Guid>()
-            });
-
-            Assert.IsType<FootnoteViewModel>(result.Value);
+            result.AssertOkResult();
         }
 
         [Fact]
-        public async Task Delete_DeleteFootnote_Returns_Ok()
+        public async Task UpdateFootnote()
```
duncan-at-hiveit, line 197: Is there a reason we've lost our DeleteFootnote() tests?  Not that they were testing much particularly :)

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
benoutram: I don't think you need to because my understanding is that includes alter what is selected, and that you would only need them if your referencing attributes of whatever your including after you've executed the query in memory.

EF core seems capable of translating what's required into SQL without them, and the only thing we return after the result is executed are the columns of Footnotes, so we don't need to include them.

Like you say, the unit tests should also cover us if they were needed and I've missed something.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ResultSubjectMetaService.cs
```diff
@@ -101,14 +101,18 @@ public class ResultSubjectMetaService : IResultSubjectMetaService
...
-
                     var indicatorViewModels = GetIndicatorViewModels(query, releaseSubject);
                     _logger.LogTrace("Got Indicators in {Time} ms", stopwatch.Elapsed.TotalMilliseconds);
                     stopwatch.Restart();
 
+                    var footnoteViewModels = await GetFootnoteViewModels(
+                        releaseId: releaseId,
+                        subjectId: query.SubjectId,
+                        filterItemIds: filterItems.Select(fi => fi.Id).ToList(),
+                        indicatorIds: indicatorViewModels.Select(i => i.Value).ToList());
```
duncan-at-hiveit, line 112: Cool - good to have those extra params handy.  How come we've not had to pass indicator ids before?

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
```diff
@@ -78,25 +76,100 @@ public FootnoteRepository(StatisticsDbContext context)
...
+            // If the list of filter and indicator id's are empty,
+            // return early with only footnotes that apply directly to this subject
+            if (!filterIdsResult.Any() && !indicatorIdsSanitised.Any())
+            {
+                return await footnotesQueryable.Where(footnote =>
+                        footnote.Subjects.Any(subjectFootnote => subjectFootnote.SubjectId == subjectId)
+
+                        // We don't allow a footnote to apply to directly to a subject and to specific subject data
+                        // of the same subject at the same time, so these additional conditions aren't strictly necessary
+                        && footnote.FilterItems.All(itemFootnote =>
```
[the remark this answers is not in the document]
benoutram: Yes it was mainly about bad data if there are any that are linked to a subject and also some specific data from the subject too. I added an example into the unit tests representing some bad data with a big comment around it too.

I'm happy to get rid of this though

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Statistics/FootnoteControllerTests.cs
benoutram: I was thinking about dropping the whole of the controller tests because generally the controller tests we've written in the past don't test very much.

In the case of this controller though, it has a complicated methods which build view models. So I've kept all test methods that end up returning view models which pass through code like `BuildFootnoteViewModel`, but removed this test which wasn't really testing anything.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ResultSubjectMetaService.cs
benoutram: It's because we previously passed `query` and then got the id's from `queryContext.Indicators`, but see the comment on the PR from @ntsim on this one starting "There's potentially a bit of a bug around the indicatorIds that get passed into this method"... for an explanation on the bug this had.

Basically the table tool forces a selection of indicators currently, but if it didn't and the list is empty the table result contains all indicators, as does the meta data.

With an empty `queryContext.Indicators` we would be getting footnotes that only apply to no indicators and miss all footnotes that apply to specific indicators if there are any.

I think we should be able to see this bug in prod currently if we query it via the API, although I've not double checked.

src/GovUk.Education.ExploreEducationStatistics.Data.Model.Tests/Repository/FootnoteRepositoryTests.cs
benoutram: Yes good spot, I think you are right. I've renamed it, thanks! 🙂

src/GovUk.Education.ExploreEducationStatistics.Data.Model.Tests/Repository/FootnoteRepositoryTests.cs
benoutram: It's the `ReleaseFootnote` table we are hoping to get rid of I think? See [EES-2979](https://dfedigital.atlassian.net/browse/EES-2979).

I'm not sure that we should get rid of this test even after we get rid of that link.

I've added this explanation into the test:

```
This test covers a case where a subject is shared by multiple releases. It makes sure that even though
a different release has footnotes for the subject, there's no way of retrieving the footnotes that
belong to another release by specifying the id, filter item id's and indicator id's of the subject.
```

src/GovUk.Education.ExploreEducationStatistics.Data.Model.Tests/Repository/FootnoteRepositoryTests.cs
benoutram: I've added these tests now, thanks!

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
benoutram: Left-hand part:
`footnote.Indicators.All(indicatorFootnote =>indicatorFootnote.Indicator.IndicatorGroup.SubjectId != subjectId)`

A footnote can apply to several subjects.

If all of the footnote indicator links are for other subjects it shouldn't exclude this footnote.
The footnote might apply to this subject in some other way (i.e. apply directly to the subject or apply to filters/filter groups/filters items of this subject)

Then if that's false, i.e. one of the links is for this subject we need to check it's contained in the list of indicator id's (the right-hand part).

I suppose there's a possibility that the footnote doesn't apply to this subject at all in which case the following could happen:

- Matches the left hand part here.
- Matches all the left hand parts below in this code for filters/filter groups/filters items because all of those links are for other subjects too.

and then it would still be included. I'm sure there's a test case for this but I'll double check this isn't a bug!

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
benoutram: See comment above for indicators.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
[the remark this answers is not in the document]
benoutram: Update: we can return early by creating a branch in the code similar to how you say, checking `if (!filterIdsResult.Any() && !indicatorIdsSanitised.Any())`

It returns footnotes that applicable directly to a subject.

```
            // If the list of filter and indicator id's are empty,
            // return early with only footnotes that apply directly to this subject
            if (!filterIdsResult.Any() && !indicatorIdsSanitised.Any())
            {
                return await footnotesQueryable.Where(footnote =>
                        footnote.Subjects.Any(subjectFootnote => subjectFootnote.SubjectId == subjectId)

                        // We don't allow a footnote to apply to directly to a subject and to specific subject data
                        // of the same subject at the same time, so these additional conditions aren't strictly necessary
                        && footnote.FilterItems.All(itemFootnote =>
                            itemFootnote.FilterItem.FilterGroup.Filter.SubjectId != subjectId)
                        && footnote.FilterGroups.All(filterGroupFootnote =>
                            filterGroupFootnote.FilterGroup.Filter.SubjectId != subjectId)
                        && footnote.Filters.All(filterFootnote =>
                            filterFootnote.Filter.SubjectId != subjectId)
                        && footnote.Indicators.All(indicatorFootnote =>
                            indicatorFootnote.Indicator.IndicatorGroup.SubjectId != subjectId))
                    .OrderBy(footnote => footnote.Order)
                    .ThenBy(footnote => footnote.Id)
                    .ToListAsync();
            }
```

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
benoutram: Update: there was a bug here, so good job you questioned this!

I didn't have any examples of footnotes which were only applicable exclusively to other subjects than the one being queried. It turns out that as soon as I added them they were being returned even though they were not relevant.

I had to largely rewrite the `GetFilteredFootnotes` method again to resolve this. It's very gnarly but it's got a lot more comments now.

I've added an entirely new unit test `GetFilteredFootnotes_FilterBySubject` which has two subjects and sets up a bunch of footnotes that apply to one subject or the other, but not both. It queries for `subject1` and checks that `subject2`'s footnotes are excluded. Prior to this change, all of `subject2`'s footnotes would have been returned due to passing all the clauses but never matching on anything.

Changing `GetFilteredFootnotes` then showed up some mistakes I'd got in `GetFilteredFootnotes_FilterByFilterItems` not expecting some footnotes which we should be expecting, so I've corrected that too. 🙂

src/GovUk.Education.ExploreEducationStatistics.Data.Model.Tests/Repository/FootnoteRepositoryTests.cs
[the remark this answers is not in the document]
duncan-at-hiveit: Ahh ok cool - thanks, that makes it clearer to me!

src/GovUk.Education.ExploreEducationStatistics.Data.Model.Tests/Repository/FootnoteRepositoryTests.cs
duncan-at-hiveit: Yup thanks - that was just me getting confused with the different join tables - ta.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
duncan-at-hiveit: Yup, happy that ti would spit out plenty of errors if it wasn't happy by this stage!

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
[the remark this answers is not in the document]
duncan-at-hiveit: Happy to leave it for now - perhaps it'd be an idea to raise a ticket to address this in the future?

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
duncan-at-hiveit: Ahh brill, happy conclusion then :D  Nicely done.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Statistics/FootnoteControllerTests.cs
duncan-at-hiveit: Yeah fair enough!

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ResultSubjectMetaService.cs
duncan-at-hiveit: Cool!

benoutram:
Updated UI test result

image

duncan-at-hiveit (approved):

PR #3714, opened 2022-12-14
EES-3900 fix map bug where areas not getting correct colour

amyb-hiveit:
Fixes a bug with maps where some areas weren't been given a colour. This due to the raw values being compared to the rounded values in the data groups, so not matching up correctly.

**Before:**
map-before

**After:**
map-after

### DIFF
--- a/src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
+++ b/src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
@@ -26,6 +26,7 @@ import { Dictionary } from '@common/types';
 import generateHslColour from '@common/utils/colour/generateHslColour';
 import locationLevelsMap from '@common/utils/locationLevelsMap';
 import formatPretty from '@common/utils/number/formatPretty';
+import countDecimalPlaces from '@common/utils/number/countDecimalPlaces';
 import classNames from 'classnames';
 import { Feature, FeatureCollection, Geometry } from 'geojson';
 import { Layer, Path, PathOptions, Polyline } from 'leaflet';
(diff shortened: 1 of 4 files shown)


luke-h1 (approved):

src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
```diff
@@ -567,9 +568,17 @@ function generateFeaturesAndDataGroups({
...
+            const dp = Math.max(
+              countDecimalPlaces(dataClass.minRaw) ?? 0,
+              countDecimalPlaces(dataClass.maxRaw) ?? 0,
+            );
+            const roundedValue = Number(value.toFixed(dp));
+            return (
+              dataClass.minRaw <= roundedValue &&
+              roundedValue <= dataClass.maxRaw
+            );
+          });
```
ntsim, line 581: Not too keen on this solution as it's not where the problem is. It's also little wasteful to re-do work figuring out decimal places again after we've already done all these calculations immediately beforehand (that could be re-used).

Seems like the issue is within `generateLegendDataGroups`. We could interpret the problem as the first/last groups having incorrect `minRaw` and `maxRaw` values set.

We could fix this by handling the first/last groups specifically e.g.

```ts
const minRaw = index === 0 ? min : roundNearest(groupMin, valueIncrement);
const maxRaw = index === groups - 1 
  ? max 
  : clamp(
      roundNearest(
        subtract(groupMin + groupSize, maxOffset),
        valueIncrement,
      ),
      roundNearest(max, valueIncrement),
    );
```

The above is for the `EqualIntervals` classification, so something similar would need to be done for the others. The main benefits of this are:

- it keeps the code for generating features untouched
- the min/max raw values are actually 'correct' (depending on your definition)

Another solution would be to just pass assign the decimal places used in each group i.e.

```ts
return {
  colour: colourScale(i + groupProportion),
  decimalPlaces,
  min: formatPretty(minRaw, unit, decimalPlaces),
  max: formatPretty(maxRaw, unit, decimalPlaces),
  minRaw,
  maxRaw,
};

// Could then be used in MapBlockInternal when generating features:

const matchingDataGroup = dataGroups.find(dataClass => {
  const roundedValue = Number(value.toFixed(dataClass.decimalPlaces));

  return (
    dataClass.minRaw <= roundedValue &&
    roundedValue <= dataClass.maxRaw
  );
});
```

Again, we'd need to do something similar for the other classification algorithms. It's probably the simpler solution, but our min/max raw values for the first/last group could be considered 'wrong' still.

I'm not quite sure which solution I prefer as there's pros/cons to both, although I might slightly lean towards just adding the `decimalPlaces` to each group. It's possibly a little more complicated for the custom grouping algorithm though, so you're probably in a better position to figure that one out :wink:

ntsim (changes requested):

ntsim (approved):

PR #3729, opened 2022-12-21
EES-3966 Fix deleting footnotes when removing a subject belonging to previous release versions

benoutram:
This PR fixes a bug where removing a data file fails to remove links between footnotes of the release and the subject or its filter items/filter groups/filters.

It manifests itself as a publishing failure in the data factory stage of publishing.
See [EES-3966](https://dfedigital.atlassian.net/browse/EES-3966) for the exact steps to reproduce that which involves replacing a file before deleting it.

Even in the local environment or an environment without data factory, the footnote links are still present after deleting a subject, and it doesn't require a data file replacement to observe that at the database level.

This PR also attempts to tidy up the `FootnoteRepository.DeleteFootnote` method which previously had a bit of a nasty signature and in certain cases did not delete the footnote.

```
        private async Task DeleteFootnote(Guid releaseId,
            Footnote footnote,
            bool canRemoveReleaseFootnote,
            bool canRemoveFootnote)
```

### Other changes

- Rename method `DeleteReleaseFootnoteLinkAsync` to `DeleteReleaseFootnoteLink`
- Rename method `DeleteAllFootnotesBySubject` to `DeleteFootnotesBySubject`
- Add lots of TODO's related to [EES-2979](https://dfedigital.atlassian.net/browse/EES-2979) which will remove `ReleaseFootnote` and make this code a lot cleaner.

UI test report

image

\* with local change to `tests/admin_and_public_2/bau/publish_release_and_amend.robot` to include recent fix not on master branch.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/FootnoteService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/FootnoteService.cs
@@ -130,7 +130,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
                 .OnSuccess(_ => _statisticsPersistenceHelper.CheckEntityExists<Footnote>(footnoteId)
                 .OnSuccessVoid(async footnote =>
                 {
-                    await _footnoteRepository.DeleteFootnote(releaseId, footnote.Id);
+                    await _footnoteRepository.DeleteFootnote(releaseId: releaseId,
+                        footnoteId: footnote.Id);
 
                     await _dataBlockService.InvalidateCachedDataBlocks(releaseId);
                 }));
@@ -174,7 +175,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
                     // TODO EES-2979 Remove this delete link and create call once all footnotes only belong to one release
                     // If this amendment of the footnote affects other release then break the link with the old
                     // and create a new one
-                    await _footnoteRepository.DeleteReleaseFootnoteLinkAsync(releaseId, footnote.Id);
+                    await _footnoteRepository.DeleteReleaseFootnoteLink(releaseId, footnote.Id);
 
                     return await _footnoteRepository.CreateFootnote(
                         releaseId,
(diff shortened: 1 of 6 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/FootnoteService.cs
```diff
@@ -130,7 +130,24 @@ public class FootnoteService : IFootnoteService
...
+                        // TODO EES-2979 Remove this once a footnote can only belong to one release
+                        // It's only be possible to reach this code if editing a very old draft release amendment,
+                        // created circa September-2020 when we began copying footnotes when creating new versions.
+                        await _footnoteRepository.DeleteReleaseFootnoteLink(releaseId: releaseId,
+                            footnoteId: footnoteId);
+
+                        // TODO Because the footnote isn't deleted this could be leaving footnote links to subjects or
+                        // filters/filter groups/filter items of subjects that only belong to this release.
+                        // It seems unlikely we will need to fix this, but prioritise EES-2979 instead. 
+                    }
```
ntsim, line 150: Couldn't all this logic be handled internally within `FootnoteRepository.DeleteFootnote` (similar to `DeleteFootnotesBySubject`)?

If we did that, we could just keep all the logic in there and just call `DeleteFootnote`.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/FootnoteRepository.cs
```diff
@@ -198,68 +198,106 @@ public FootnoteRepository(StatisticsDbContext context)
...
+                        // Therefore we can't rely on the cascade delete of all the footnote links related to the subject
+                        // which would happen when it's deleted. Explicitly delete the links instead
 
-            await _context.SaveChangesAsync();
-        }
+                        var filterItemFootnoteLinks = await _context.FilterItemFootnote
+                            .Where(filterItemFootnote =>
+                                filterItemFootnote.FootnoteId == footnote.Id &&
+                                filterItemFootnote.FilterItem.FilterGroup.Filter.SubjectId == subjectId)
+                            .ToListAsync();
```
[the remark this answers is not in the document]
benoutram: Done this, thanks.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/FootnoteService.cs
benoutram: Thanks, I actually moved this kind of logic out of `FootnoteRepository.DeleteFootnote` to make sure the only thing that was doing was actually deleting a foonote, but I agree it's a bit cleaner this way so I've moved it back.

It will be a lot cleaner all over the place once we've done EES-2979.

ntsim (approved):

PR #3734, opened 2022-12-22
EES-3970 - fix prod UI tests

luke-h1:
This PR:
* Removes all production find stats page assertions in `statistics_page.robot` due to the statistics page being covered by snapshot tests
* Adds a nicer exception when slack environment variables aren't set (causing local snapshot tests to fail with an exception that's quite hard to decipher)   
* Updates methodology snapshots


tests/robot-tests/tests/libs/snapshots.py
```diff
@@ -20,27 +20,25 @@
 
 slack_service = SlackService()
 
+public_url = os.getenv("PUBLIC_URL")
```
markhiveit, line 23: If you pull this out, you can remove all instances of `os.getenv('PUBLIC_URL')` from this file.

markhiveit (approved):

PR #3735, opened 2022-12-22
EES-3971 - amended logic that determines latest ReleaseSubject for a SubjectId

duncan-at-hiveit:
# Overview

This PR:
- fixes a bug whereby we had null "Published" dates in the statistics table that were being relied on during the working out of the latest ReleaseSubject for a given SubjectId - this led to 500s when attempting to query a Subject from a row of Release versions that had the above-mentioned gaps in its "Published" date history.  This is fixed by the next bulletpoint!
- swapping over to using the Content DB version of Release to determine the latest version of a series of Release amendments that are linked to a given SubjectId.  This is simpler than using the Statistics DB's Release table because we can use the "Version" column to determine the latest. 
- removes the use of "Published" date and "PreviousVersionId" from the statistics db's version of the Release table, instead using the existing content db's Release table with the equivalent information

## Work done

I have moved the `GetReleaseSubjectForLatestPublishedVersion()` method from `ReleaseSubjectRepository` into `SubjectMetaService` in order to access the ContentDbContext, and then rewritten it to use the Version column on the content Release to work out the latest version rather than the more convoluted original way of looking for live Releases that weren't referenced by lists of PreviousVersionIds. 

The reason that we had gaps in the "Published" dates on the statistics db's Release table is that historically we didn't used to capture the Published date in the private stats db at all, only the Content DB and the public-stats db.  When Read Replica stage 1 went out, we copied all Published dates from public-stats into private stats - but obviously this didn't account for them all, because the public stats db only had the Published date for whichever Release version was currently live at that time, not for any previously amended Release versions (they are deleted from public-stats as a newer amendment is published).

As part of the Read Replica cleanup ticket (EES-3763), we can drop the unused columns from the Statistics DB's Release table.

## Additional work done

I've updated the `get_data_block_responses.py` slightly to produce more meaningful console output, and stash the request URLs along with the request bodies.

image


src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/PolicyCheckBuilder.cs
```diff
@@ -89,5 +89,11 @@ public async Task AssertSuccess<T>(Func<Mock<IUserService>, Task<Either<ActionRe
             Assert.NotNull(result);
             Assert.True(result.IsRight);
         }
+
+        public async Task AssertSuccess<T>(Func<Mock<IUserService>, Task<T>> action)
```
luke-h1, line 93: just wondering if this is going to conflict with the existing `AssertSuccess` task on L83?

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/SubjectMetaServicePermissionTests.cs
```diff
@@ -103,48 +111,89 @@ await PolicyCheckBuilder<DataSecurityPolicies>()
...
 
                     return await service.FilterSubjectMeta(null, query, cancellationToken);
                 }
             );
     }
 
+    [Fact]
+    public async Task GetReleaseSubjectForLatestPublishedVersion()
+    {
+        // Note that there are on explicit security policy checks enforced for this method.
```
luke-h1, line 143: Small grammar issue here I think?

markhiveit (changes requested):
Does a migration need to be created to remove the columns from the Statistics DB Releases table?

Edit: Discussed on Slack - the migration will be done in another ticket.

markhiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/PolicyCheckBuilder.cs
duncan-at-hiveit: It doesn't - if you take a look via Rider, the original method is still used everywhere except for the single usage of the new method that was introduced in this PR.

It's how we have it set up with `OnSuccess` method usages too - the C# compiler is very good at finding the best candidate method for a generic call.

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/SubjectMetaServicePermissionTests.cs
duncan-at-hiveit: Typo - should've read "no", not "on" :)  Cheers!

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/PolicyCheckBuilder.cs
luke-h1: hmm interesting

luke-h1 (approved):

PR #3755, opened 2023-01-06
EES-3960 - add superseded warning to data-catalogue & release pages

luke-h1:
This PR adds a superseded warning message on release & data-catalogue pages when a publication has been archived. 

This allows a user to:
* Click a link on the publication page which takes them to the publication that is taking over from the archived publication
* Navigate to data-catalogue step 1 of the publication that is taking over from the archived publication when going through a publication that has been archived

**Relevant changes**:
* updated various content publication view models to expose `supersededById`, `supersededByTitle` & `supersededBySlug`.
* Fixes a bug/code smell in `DatacataloguePage` whereby stale query state was being held onto after the user had navigated back to step 1 of the data catalogue page (and therefore had nothing selected) 
* Added a default value to the `PUBLIC_URL` frontend environment variable (`http://localhost:3000`) for convenience/testing purposes

⚠️ **Deployment notes** ⚠️
* The public `PublicationTree` cache tree will need to be cleared as part of deployment of this work due to changes in publication view models  (see notes left on https://dfedigital.atlassian.net/browse/EES-3973)

- [x] Implement final format (banner or warning text)
- [x] Frontend unit tests
- [x] Amendments to UI tests
- [x] Investigate if possible to redirect to data catalogue step instead of find stats
- [x] Amendments to BE tests

UI test results:
image


src/explore-education-statistics-frontend/src/modules/data-catalogue/DataCataloguePage.tsx
```diff
@@ -144,6 +154,45 @@ const DataCataloguePage: NextPage<Props> = ({
...
+
+    updateState(draft => {
+      draft.query.publication = {
+        id: state.query.publication?.supersededById as string,
+        slug: state.query.publication?.supersededBySlug as string,
+        title: state.query.publication?.supersededByTitle as string,
+        isSuperseded: false,
+        // TODO: Remove 'type' field (see EES-3839) as it is unused on new find stats page
+        // (old-find stats page previously used 'type' to group publications by their type)
+        type: type as PublicationType,
```
benoutram, line 174: Hopefully this is fixed by https://github.com/dfe-analytical-services/explore-education-statistics/pull/3756!

src/explore-education-statistics-frontend/src/modules/find-statistics/__tests__/PublicationReleasePage.test.tsx
```diff
@@ -35,6 +35,42 @@ describe('PublicationReleasePage', () => {
...
+    expect(
+      screen.getByTestId('superseded-warning').querySelector('a'),
+    ).toHaveAttribute('href', '/find-statistics/publication-a');
+
+    expect(
+      screen.getByTestId('superseded-warning').querySelector('a'),
+    ).toHaveTextContent('publication A');
+  });
+
+  test("doesn't render superseded warning text when publication is superseded", () => {
```
ntsim, line 60: Could avoid the contraction here i.e. 'does not render...'

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ViewModels/PublicationTreePublicationViewModel.cs
```diff
@@ -15,6 +15,12 @@ public record PublicationTreePublicationViewModel
 
         public bool IsSuperseded { get; set; }
 
+        public Guid? SupersededById { get; set; }
+        
+        public string? SupersededByTitle { get; set; }
+
+        public string? SupersededBySlug { get; set; }
```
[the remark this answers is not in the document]
luke-h1: Cool, sounds good. Yeah, I wasn't keen on this implementation either as it makes the view model busier than it needs to be amongst other things like you say (talked with Mark & Ben about potentially implementing this type of thing the first time round )

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/Cache/PublicationCacheServiceTests.cs
```diff
@@ -305,6 +305,9 @@ public async Task GetPublicationTree_DataTables_SupersededPublicationWithDataOnL
                     {
                         Title = "Publication A",
                         IsSuperseded = true,
+                        SupersededById = Guid.NewGuid(),
+                        SupersededByTitle = "Publication B",
+                        SupersededBySlug = "publication-b",
```
markhiveit, line 310: I don't think these need to be added. We filter releases that are superseded in `PublicationCacheService#FilterPublicationTreePublication`, which only uses `IsSuperseded`. By adding these extra fields to the test data, you're implying that these are also used by the method under test. 

I think our general pattern with test data is that we only set values that need to be set, such that someone reading the tests can infer what the method under test actually uses. It's possible other devs disagree, however.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/PublicationService.cs
```diff
@@ -182,6 +183,9 @@ public async Task<IList<PublicationTreeThemeViewModel>> GetPublicationTree()
                 : null,
             LatestReleaseId = publication.LatestPublishedReleaseId!.Value,
             IsSuperseded = isSuperseded,
+            SupersededById = publication.SupersededBy?.Id,
+            SupersededByTitle = publication.SupersededBy?.Title,
+            SupersededBySlug = publication.SupersededBy?.Slug,
```
markhiveit, line 188: I think we need some additional testing in `PublicationServiceTests.cs`, that checks that `SupersededById/Title/Slug` are being returned correctly when calling `PublicationService#Get`.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/PublicationService.cs
```diff
@@ -246,6 +250,9 @@ private async Task<PublicationTreePublicationViewModel> BuildPublicationTreePubl
             Title = publication.Title,
             Slug = publication.Slug,
             IsSuperseded = await _publicationRepository.IsSuperseded(publication.Id),
+            SupersededById = publication.SupersededBy?.Id,
+            SupersededByTitle = publication.SupersededBy?.Title,
+            SupersededBySlug = publication.SupersededBy?.Slug,
```
markhiveit, line 255: Also need additional testing for this too - for `PublicationService#GetPublicationTree`.

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -137,8 +140,8 @@ Check that archive-publication subject appears correctly on Data catalogue page
 
     user clicks radio    %{TEST_THEME_NAME}
 
-    user checks page contains    ${PUBLICATION_NAME_ARCHIVE}
-    user checks page does not contain    ${PUBLICATION_NAME_SUPERSEDE}
+    user checks page contains element    //*[@data-testid="Radio item for ${PUBLICATION_NAME_ARCHIVE}"]
+    user checks page does not contain element    //*[@data-testid="Radio item for ${PUBLICATION_NAME_SUPERSEDE}"]
```
markhiveit, line 144: Could use `testid:` here? If so, can make the same change in a couple of other places in this file.

markhiveit (changes requested):

src/explore-education-statistics-frontend/src/modules/data-catalogue/DataCataloguePage.tsx
```diff
@@ -80,6 +82,13 @@ const DataCataloguePage: NextPage<Props> = ({
 
   const handlePublicationStepBack = () => {
     router.push('/data-catalogue', undefined, { shallow: true });
+    updateState(s => {
```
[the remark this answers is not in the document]
luke-h1: Cool, good to know. I've changed this to `draft` instead of `s`. Haven't read much about immer. Assume we typically reach for it when dealing with more complex state that doesn't quite fit in something like `useState` or `useContent`?

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
luke-h1: Yeah I think I meant to clean this up first time round but looks like I may have forgotten to replace `//*[@data-testid]` with the `testid:` shorthand syntax

src/GovUk.Education.ExploreEducationStatistics.Content.Services/PublicationService.cs
luke-h1: Added `Get_IsSuperseded_SupersededPublicationBuildsPublicationSupersededByViewModel` to `PublicationServiceTests.cs`

src/GovUk.Education.ExploreEducationStatistics.Content.Services/PublicationService.cs
luke-h1: Added `Get_PublicationTree_SupersededPublicationBuildsPublicationSupersededByViewModel` to`PublicationServiceTests.cs`. Like the other unit test added to this file, wasn't too sure about the naming (it seems a bit verbose)

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ViewModels/PublicationViewModel.cs
```diff
@@ -16,6 +16,8 @@ public record PublicationViewModel
 
     public bool IsSuperseded { get; init; }
 
+    public PublicationSupersededByViewModel SupersededBy { get; init; } = new();
```
benoutram, line 19: This should be optional since it's set from the `PublicationCacheViewModel` where it's optional.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ViewModels/PublicationViewModel.cs
```diff
@@ -40,11 +42,12 @@ public PublicationViewModel()
         Slug = publication.Slug;
         LatestReleaseId = publication.LatestReleaseId;
         IsSuperseded = publication.IsSuperseded;
+        SupersededBy = publication.SupersededBy!;
```
benoutram, line 45: You won't need the null-forgiving operator `!` here when you make this optional. That was a warning that something was not quite right here.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/PublicationService.cs
```diff
@@ -255,10 +272,11 @@ private async Task<PublicationTreePublicationViewModel> BuildPublicationTreePubl
         };
     }
 
+
```
benoutram, line 275: Minor, but extra blank line added here

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
```diff
@@ -288,6 +288,168 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
...
+                
+                
+                Assert.Equal(1, publicationTree.Count);
+                var topicAPublications = publicationTree[0].Topics[0].Publications;
+                
+                Assert.Equal(publicationA.Id, topicAPublications[0].SupersededBy?.Id);
+                Assert.Equal(publicationA.Title, topicAPublications[0].SupersededBy?.Title);
+                Assert.Equal(publicationA.Slug, topicAPublications[0].SupersededBy?.Slug);
+                
+            }
```
benoutram, line 451: Minor, but we've got some excess blank lines in this block

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
```diff
@@ -288,6 +288,168 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
...
+                            Publications = ListOf(publicationA)
+                        },
+                        new()
+                        {
+                            Title = "Topic B",
+                            Publications = ListOf(publicationB)
+                        }
+                    }
+                }
+            };
```
benoutram, line 427: This test only needs to check that the superseded view model is set so there's no need to set up two different publications. One of the publications isn't even published so isn't in the result for you to check that both have the superseded view model.

I would remove one of the publication / release / and topic. It looks like it's `topicA`, `publicationA`, `releaseA` which is not published and can be removed, allowing you then to rename all the 'B's' when you've only got one of each.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
```diff
@@ -288,6 +288,168 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
...
+                var result = await service.Get(publication.Slug);
+
+                var publicationViewModel = result.AssertRight();
+                
+                Assert.Equal(publication.Id, publicationViewModel.Id);
+                Assert.True(publicationViewModel.IsSuperseded);
+                
+                Assert.Equal(publication.SupersededBy.Title, publicationViewModel.SupersededBy?.Title);
+                Assert.Equal(publication.SupersededBy.Slug, publicationViewModel.SupersededBy?.Slug);
+                Assert.Equal(publication.SupersededBy.Id, publicationViewModel.SupersededBy!.Id);
```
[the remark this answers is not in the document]
luke-h1: I believe `Get_IsSuperseded_SupersedingPublicationHasNoPublishedRelease ` wouldn't be suitable for testing as `isSuperseded` will be set to false and `SupersededBy` will be null since you need to have a minimum of one published release in order for a publication to be superseded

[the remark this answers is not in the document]
benoutram: I see, thanks. I was assuming you were populating the view model based on the presence of `publication.SupersededBy` rather than the value of `IsSuperseded`. The two aren't necessarily in sync because as long as you've loaded `publication.SupersededBy` it can be not-null when `IsSuperseded` is false.

I think you can update `Get_IsSuperseded_SupersedingPublicationHasNoPublishedRelease` to add an assertion like `Assert.Null(publicationViewModel.SupersededBy)` to make sure it's not being populated.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
luke-h1: I've added `PublicationServiceTests#Get_IsSuperseded_SupersedingPublicationBuildsSupersededViewModel` to replace this

src/explore-education-statistics-frontend/src/modules/data-catalogue/DataCataloguePage.tsx
```diff
@@ -144,6 +153,34 @@ const DataCataloguePage: NextPage<Props> = ({
...
+          to="/data-catalogue"
+          shallow
+          testId="superseded-by-link"
+          onClick={() => handleSupersededWarningSubmit()}
+        >
+          {state.query.publication.supersededByTitle}
+        </Link>
+      </WarningMessage>
+    ) : null;
+  };
```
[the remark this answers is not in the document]
luke-h1: Looks like 
```typescript 
useEffect(() => {
    // Update state if the page props change
    updateState(() => initialState);
  }, [initialState, updateState]);
```
causes an infinite loop to occur which is currently causing the unit tests to run out of memory. Everything seems to work when using the FE however this problem will probably cause a memory leak due to the useEffect being fired whenever the dataCataloguePage updates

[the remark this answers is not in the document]
luke-h1: I've changed the `useEffect` dependency array to include the `publicationSlug` and not `updateState` which seems to have stopped the infinite re-renders: 

```typescript
  useEffect(() => {
    updateState(() => initialState)
  // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [router.query.publicationSlug, updateState])
```

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
luke-h1: I've had a look at implementing this but the tests end up throwing `Page should have contained element 'testid:Radio item for UI tests - UI tests - archived publication 20230113-095242' but did not.`  in certain places. I've switched back to using the `//*[@data-testid=""]` syntax for now as it doesn't really seem worth it to pursue getting the `testId` syntax to work simply for the benefit of syntax sugar.

src/explore-education-statistics-frontend/src/modules/find-statistics/__tests__/PublicationReleasePage.test.tsx
```diff
@@ -35,6 +35,42 @@ describe('PublicationReleasePage', () => {
...
+
+    render(<PublicationReleasePage release={testReleaseSuperseded} />);
+
+    expect(
+      screen.getByTestId('superseded-warning').querySelector('a'),
+    ).toHaveAttribute('href', '/find-statistics/publication-a');
+
+    expect(
+      screen.getByTestId('superseded-warning').querySelector('a'),
+    ).toHaveTextContent('publication A');
```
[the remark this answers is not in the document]
ntsim: Looks like this still needs to be done :point_up:

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -187,6 +191,19 @@ Check public archive-publication release page displays correctly
     user waits until h1 is visible    ${PUBLICATION_NAME_ARCHIVE}    %{WAIT_MEDIUM}
     user checks page does not contain    This is the latest data
 
+Check public archive-publication release page displays superseded warning
+    user checks page contains element    testid:superseded-warning
+    user checks page contains element    testid:superseded-by-link
+
```
ntsim, line 197: Shouldn't we check that the link contains the correct publication text?

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -206,11 +223,12 @@ Check data catalogue page contains archive and superseding publication subjects
 
     user clicks radio    %{TEST_THEME_NAME}
 
-    user checks page contains    ${PUBLICATION_NAME_ARCHIVE}
-    user checks page contains    ${PUBLICATION_NAME_SUPERSEDE}
+    user checks page contains element    //*[@data-testid="Radio item for ${PUBLICATION_NAME_ARCHIVE}"]
+    user checks page contains element    //*[@data-testid="Radio item for ${PUBLICATION_NAME_SUPERSEDE}"]
```
ntsim, line 227: Can't we use `user checks page contains radio`?

tests/robot-tests/tests/admin_and_public/bau/archive_publication.robot
```diff
@@ -243,6 +261,62 @@ Check data catalogue page contains archive and superseding publication subjects
...
+    user clicks button    Next step
+
+    user checks page contains element    testid:superseded-warning
+    user checks page contains element    testid:superseded-by-link
+
+    user clicks element    testid:superseded-by-link
+
+    user waits until page contains    Browse our open data    %{WAIT_SMALL}
+
+    user waits until page contains    ${PUBLICATION_NAME_SUPERSEDE}    %{WAIT_SMALL}
```
ntsim, line 318: Might want to check that the page no longer contains the archived publication's name too

src/GovUk.Education.ExploreEducationStatistics.Content.Services/PublicationService.cs
```diff
@@ -224,7 +237,7 @@ private async Task<PublicationTreeTopicViewModel> BuildPublicationTreeTopic(Topi
             .Where(publication => publication.LatestPublishedReleaseId != null)
             .ToAsyncEnumerable()
             .SelectAwait(async publication =>
-                await BuildPublicationTreePublication(publication))
+                await BuildPublicationTreePublication(publication))
```
benoutram, line 240: Minor, but no overall change here, just some whitespace creeping in on L240 and L279 as well.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
```diff
@@ -222,6 +222,11 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
 
                 Assert.Equal(publication.Id, publicationViewModel.Id);
                 Assert.True(publicationViewModel.IsSuperseded);
+                
+                Assert.NotNull(publicationViewModel.SupersededBy);
+                Assert.Equal(publication.SupersededBy.Id, publicationViewModel.SupersededBy?.Id);
+                Assert.Equal(publication.SupersededBy.Title, publicationViewModel.SupersededBy?.Title);
+                Assert.Equal(publication.SupersededBy.Slug, publicationViewModel.SupersededBy?.Slug);
```
benoutram, line 229: I think this should be

```
                Assert.NotNull(publicationViewModel.SupersededBy);
                Assert.Equal(publication.SupersededBy.Id, publicationViewModel.SupersededBy!.Id);
                Assert.Equal(publication.SupersededBy.Title, publicationViewModel.SupersededBy.Title);
                Assert.Equal(publication.SupersededBy.Slug, publicationViewModel.SupersededBy.Slug);
```

where you only need one null-forgiving `!` operator rather than the three null-conditional operators.

I think you should also give the `SupersededBy` publication on L174 a `Title` and a `Slug` otherwise you'll see we are just comparing the default empty string values here if you debug.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
```diff
@@ -285,9 +290,75 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
 
                 Assert.Equal(publication.Id, publicationViewModel.Id);
                 Assert.False(publicationViewModel.IsSuperseded);
+                Assert.Null(publicationViewModel.SupersededBy);
             }
         }
 
+        [Fact]
+        public async Task Get_IsSuperseded_SupersedingPublicationBuildsSupersededViewModel()
```
benoutram, line 298: It confused me for a while when reading the contents of this test but finally worked out it should be named `GetPublicationTree_`... rather than `Get_`... 🙂 

I would also move it much further down the test class to sit below `GetPublicationTree_Superseded` around L1324.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
```diff
@@ -285,9 +290,75 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
...
+            {
+                var service = SetupPublicationService(context);
+
+                var publicationTree = await service.GetPublicationTree();
+
+                var testTopicPublications = publicationTree[0].Topics[0].Publications;
+                
+                Assert.Single(testTopicPublications);
+                
+                var testTopicPublication = testTopicPublications[0];
```
benoutram, line 351: You can combine these two into one by assinging the result e.g.

```
var testTopicPublication = Assert.Single(testTopicPublications);
```

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
```diff
@@ -285,9 +290,75 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
...
+                
+                Assert.True(testTopicPublication.IsSuperseded);
+                
+                Assert.NotNull(testTopicPublication.SupersededBy);
+                Assert.Equal(publication.SupersededBy!.Id, testTopicPublication.SupersededBy!.Id);
```
benoutram, line 356: You don't need the first null-forgiving `!` operator in `publication.SupersededBy!.Id` because it's been declared above with a `SupersededBy` so it's known to be not-null here.

src/explore-education-statistics-frontend/src/modules/find-statistics/__tests__/PublicationReleasePage.test.tsx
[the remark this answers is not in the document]
luke-h1: ah my bad, not sure if I lost this during a force push

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
luke-h1: Ah yeah, good catch! moved below `GetPublicationTree_Superseded` now

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
markhiveit: Looks like this hasn't been changed?

src/GovUk.Education.ExploreEducationStatistics.Content.Services/PublicationService.cs
markhiveit: I suspect you've made these changes and they've got lost somewhere? I'll unresolve any comments where I think the change hasn't been made.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ViewModels/PublicationCacheViewModel.cs
```diff
@@ -25,4 +27,4 @@ public record PublicationCacheViewModel
     public ContactViewModel Contact { get; init; } = null!;
 
     public ExternalMethodologyViewModel? ExternalMethodology { get; init; }
-}
+}
```
markhiveit, line 30: It is a unix convention that text files should end with a new line. In theory, software dealing with text files can expect text files to end with a new line. In practice, I don't think much software will make that assumption because it's broken so regularly.

Although, it's not always followed, we agreed at some point to change our editors to leave a newline at the end of each file. To be honest, it probably doesn't matter.

markhiveit (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
benoutram: I think it's up to date now with the expected changes. I'm going to approve. 🙂

benoutram (approved):

PR #3764, opened 2023-01-13
EES-3283 Prevent creating new releases and methodologies for archived publications + dashboard fixes

benoutram:
This PR makes a change to prevent creating new releases and methodologies for archived publications.

Since the 'Create new release' and 'Create new methodology' buttons in the front-end are already controlled by whether permissions at the publication level allow creating them, this was a simple back-end change to the auth handlers to disallow both if the publication has a superseding publication set.

Prior to making this change I checked with @lauraselby that this should apply as soon as it’s set to be archived rather than waiting for the superseding publication to have a published release.

I also checked that we don't need to prevent managing adopted methodologies, external methodologies or legacy releases, at least for now.

### Other changes

As part of working on this PR I toggled a lot of the back-end publication/release/methodology permissions to understand how their behaviour affects the front-end and found a couple of bugs.

- Contributors looking at the methodologies tab saw this 'Other options' heading with no content since they can't manage external methodologies or adopted methodologies so this is now hidden.

methodologies

- Contributors viewing the legacy releases tab had controls available to create/reorder/edit and delete which all led to permission errors since they have no permission to manage legacy releases. I checked with @cjrace and allowing them to see a read-only view of the legacy releases tab is fine for now, so these additional controls have been removed and tests have been written around that.

legacyreleases

### UI test report

image

\* 27 failures in `Tests.Admin.Bau.Prerelease` not related to this change.


tests/robot-tests/tests/admin/analyst/analyst_absence_permissions.robot
```diff
@@ -32,7 +32,60 @@ Navigate to Pupil absence Publication page
...
+    user checks table cell contains    5    3
+    ...    https://www.gov.uk/government/statistics/pupil-absence-in-schools-in-england-including-pupil-characteristics-academic-year-2010-to-2011
+
+    user checks table cell contains    6    1    0
+    user checks table cell contains    6    2    Academic Year 2009/10
+    user checks table cell contains    6    3
+    ...    https://www.gov.uk/government/statistics/pupil-absence-in-schools-in-england-including-pupil-characteristics-academic-year-2009-to-2010
+
+Check cannot create a legacy release
+    user checks page does not contain button    Create legacy release
```
luke-h1, line 73: Would suggest annotating the test case titles with something such as `Check Analyst1 cannot ...` just to make it clear as to what user we are referring to in these test cases

tests/robot-tests/tests/admin/analyst/role_ui_permissions/publication_approver_ui_permissions.robot
```diff
@@ -14,19 +14,24 @@ Force Tags          Admin    Local    Dev
...
-    Set suite variable    ${PUBLICATION_NAME}    ${PUBLICATION_FOR_RELEASE_APPROVER}
+    Set suite variable    ${PUBLICATION_NAME}    ${PUBLICATION_FOR_PUBLICATION_APPROVER}
 
-Validates release approver publication page is correct
+Validates publication approver publication page is correct
     user navigates to publication page from dashboard    ${PUBLICATION_NAME}
 
     user waits until page contains link    Releases
     user waits until page contains link    Methodologies
-    user waits until page contains link    Legacy releases    # remove as part of EES-3794
```
luke-h1, line 24: Just wondering why we're removing the `EES-3794` comment here and throughout this PR?

luke-h1 (changes requested):

tests/robot-tests/tests/admin/analyst/analyst_absence_permissions.robot
benoutram: Thanks, done that now so it's consistent with the other tests.

tests/robot-tests/tests/admin/analyst/role_ui_permissions/publication_approver_ui_permissions.robot
benoutram: I only realised that the [EES-3794](https://dfedigital.atlassian.net/browse/EES-3794) ticket existed to hide the 'Legacy releases' tab when I saw a reference to this in code while updating UI tests.

The result of my change is that contributors can see legacy releases in the dashboard in a read-only view of them.

I mentioned this to @Cam Race today and he said a read-only view of legacy releases is fine for now.

I've left a comment on EES-3794 to explain and left it open in case we still want to go ahead with removing it completely in future.

This PR's primary change (hiding Create release/methodology buttons for archived publications) has nothing to do with legacy releases, so what started out intending to be a small bugfix while working in the same area of dashboard permissions eventually turned out to be a bigger change than expected. 😢

With hindsight I should have stopped and made it into it's own change, in which case I'd have probably spotted EES-3794 earlier!

markhiveit (approved):

luke-h1 (approved):

PR #3771, opened 2023-01-16
hotfix(frontend): fix incorrect markup for screen-reader users

luke-h1:


src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
```diff
@@ -166,7 +165,7 @@ const FindStatisticsPage: NextPage = () => {
     >
       <InsetText>
         <h2 className="govuk-heading-m">This page has changed</h2>
-        <p className={classNames('govuk-body', 'govuk-warning-text')}>
+        <p className="govuk-body">
```
ntsim, line 168: This doesn't fix the problem. The issue is that we're using a heading here incorrectly.

We should do this instead:
```tsx
<p className="govuk-heading-m">This page has changed.<p>
<p>...</p>
```

Note that we don't need `govuk-body` on `p` elements typically as that styling is applied by default.

src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
```diff
@@ -166,7 +165,7 @@ const FindStatisticsPage: NextPage = () => {
...
           Following user feedback we’ve made some changes to this page to make
           our publications easier to find, if you have any comments on the new
           design please let us know via the{' '}
```
ntsim, line 171: We're also missing a full stop at the end of the sentence (after the link)

ntsim (changes requested):

ntsim (approved):

PR #3776, opened 2023-01-17
EES-3939 add overflow to tables in content

amyb-hiveit:
Tables added in CKEditor could overflow the page on small screens, I've added overflow:auto to their container so that it will scroll instead.

**Before:**
table-before

**After:**
table-after

### DIFF
--- a/src/explore-education-statistics-common/src/styles/_content.scss
+++ b/src/explore-education-statistics-common/src/styles/_content.scss
@@ -18,6 +18,10 @@
     max-width: 100%;
     min-width: 50px;
   }
+
+  .table {
+    overflow: auto;
+  }
 }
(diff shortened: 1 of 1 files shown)


luke-h1 (approved):

duncan-at-hiveit (approved):

src/explore-education-statistics-common/src/styles/_content.scss
```diff
@@ -18,6 +18,10 @@
     max-width: 100%;
     min-width: 50px;
   }
+
```
duncan-at-hiveit, line 21: I don't suppose this might cause any Robot tests to fail due to some table cell no longer being visible?

PR #3780, opened 2023-01-18
EES-4015 Cleanup fetching latest release/subjects in data API

ntsim:
## :warning: Depends on #3761 

This PR performs various cleanup around the data API to prevent errors being thrown if the latest release or subject can't be found (resulting in a 500). Instead, we now correctly return a 404 response instead.

## `OrNotFound` extension

One notable change here is the addition of a new `OrNotFound` extension method for `EitherTasks`. This allows us to conveniently return a `NotFoundResult` (i.e. 404) for tasks that have a nullable result (e.g. `Task<T?>`).

Usage is simply:
```cs
return await someService.SomeNullableResult()
  .OrNotFound() // Return a 404 if null
  .OnSuccess(...); // Do something else if not null
```

The main benefit to this extension is that we can use this to wrap many repository methods that return `Task<T?>`. We no longer need to adapt these methods to use eithers (e.g. `Task<Either<ActionResult, T>>`) to be compatible with the rest of our service methods.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Services/TableBuilderService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Services/TableBuilderService.cs
@@ -66,10 +66,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Services
             ObservationQueryContext queryContext,
             CancellationToken cancellationToken = default)
         {
-            return await _subjectRepository.FindPublicationIdForSubject(queryContext.SubjectId)
-                .OrNotFound()
-                .OnSuccess(publicationId => _releaseRepository.GetLatestPublishedRelease(publicationId))
-                .OnSuccess(release => Query(release.Id, queryContext, cancellationToken));
+            return await FindLatestPublishedReleaseId(queryContext.SubjectId)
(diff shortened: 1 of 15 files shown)


src/GovUk.Education.ExploreEducationStatistics.Data.Services/TableBuilderService.cs
```diff
@@ -143,6 +142,14 @@ private async Task<int> GetMaximumTableCellCount(ObservationQueryContext queryCo
             );
         }
 
+        private async Task<Either<ActionResult, Guid>> FindLatestReleaseId(Guid subjectId)
```
duncan-at-hiveit, line 145: Would you consider renaming this to `FindLatestPublishedReleaseId` to avoid confusion around whether it's trying to find the latest published versus the latest being worked upon in Admin?  Given that this is in a common Services project, it could potentially be getting used by either.

duncan-at-hiveit (approved):

duncan-at-hiveit:
Nice addition to our extensions!

src/GovUk.Education.ExploreEducationStatistics.Data.Services/TableBuilderService.cs
ntsim: Sure, changed!

PR #3783, opened 2023-01-20
EES-3009

N-moh:
Added a missing UI test to publish_release_and_amend suite to visit a cashed admin block (on the content tab) after the data block has been edited in order to assert that the cache has been invalidated.
Also changed few key words from "updated" to "Amended".
Screenshot (51)

### DIFF
--- a/tests/robot-tests/tests/admin_and_public_2/bau/publish_release_and_amend.robot
+++ b/tests/robot-tests/tests/admin_and_public_2/bau/publish_release_and_amend.robot
@@ -170,6 +170,45 @@ Add embedded dashboard to third accordion section
     user waits until h1 is visible    DfE Analytical Services R-Shiny data dashboard template (h1)    90
     unselect frame
 
+User navigates to Data blocks page to edit block
+    [Documentation]    EES-3009
+    user clicks link    Data blocks
+    user waits until h2 is visible    Data blocks    %{WAIT_SMALL}
+
+Edit data block
+    #just updating the table title and source
+    user waits until table is visible
+    user clicks link    Edit block    css:tbody > tr:first-child
+    user waits until h2 is visible    ${DATABLOCK_NAME}
+    user waits until h2 is visible    Data block details
+
+    user enters text into element    id:dataBlockDetailsForm-name    ${DATABLOCK_NAME}
+    user clears element text    id:dataBlockDetailsForm-heading
+    user enters text into element    id:dataBlockDetailsForm-heading    Updated dates table title
+    user clears element text    id:dataBlockDetailsForm-source
+    user enters text into element    id:dataBlockDetailsForm-source    Updated dates source
+
+    user clicks button    Save data block
+    user waits until page contains button    Delete this data block
+
+Navigate to the 'Content' page
+    user clicks link    Content
+    user waits until h2 is visible    ${PUBLICATION_NAME}
+    user waits until page contains button    Add a summary text block    %{WAIT_SMALL}
+
+Verify data block is updated correctly
+    #checking if data block cache has been invalidated by verifying the updates on the block
+    user scrolls to accordion section content    Dates data block    id:releaseMainContent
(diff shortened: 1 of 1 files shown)


tests/robot-tests/tests/admin_and_public_2/bau/publish_release_and_amend.robot
```diff
@@ -170,6 +170,45 @@ Add embedded dashboard to third accordion section
...
+
+Edit data block
+    #just updating the table title and source
+    user waits until table is visible
+    user clicks link    Edit block    css:tbody > tr:first-child
+    user waits until h2 is visible    ${DATABLOCK_NAME}
+    user waits until h2 is visible    Data block details
+
+    user enters text into element    id:dataBlockDetailsForm-name    ${DATABLOCK_NAME}
+    user clears element text    id:dataBlockDetailsForm-heading
```
duncan-at-hiveit, line 186: I don't think you need to use `user clears element text` here (and in other lines added in this PR), as the `user enters text into element` keyword also contains it.

From `common.robot`:

```
user enters text into element
    [Arguments]    ${selector}    ${text}
    user sets focus to element    ${selector}
    user waits until element is visible    ${selector}    %{WAIT_SMALL}
    user clears element text    ${selector}
    press keys    ${selector}    ${text}
    sleep    0.3
```

tests/robot-tests/tests/admin_and_public_2/bau/publish_release_and_amend.robot
```diff
@@ -170,6 +170,45 @@ Add embedded dashboard to third accordion section
...
+
+Verify data block is updated correctly
+    #checking if data block cache has been invalidated by verifying the updates on the block
+    user scrolls to accordion section content    Dates data block    id:releaseMainContent
+    user opens accordion section    Dates data block    id:releaseMainContent
+    ${section}=    user gets accordion section content element    Dates data block    id:releaseMainContent
+
+    user checks chart title contains    ${section}    Updated dates table title
+    user clicks link by visible text    Table    ${section}
+    user waits until parent contains element    ${section}
```
duncan-at-hiveit, line 207: I've seen this written in other places like:

```
    user waits until element contains    testid:dataTableCaption
    ...    Admission Numbers for 'UI test subject' in Nailsea Youngwood and Syon between 2005 and 2017
```

which cuts down a bit of xpath.  Consider using this way instead?

For the future, we should probably have another common keyword for this, like:

`user waits for data table caption`

so we don't have to repeat this pattern all over the place :)

tests/robot-tests/tests/admin_and_public_2/bau/publish_release_and_amend.robot
```diff
@@ -576,13 +615,13 @@ Verify amended Dates data block table has footnotes
...
     user opens accordion section    Test text    id:releaseMainContent
-    user adds content to autosaving accordion section text block    Test text    1    Updated test text!
+    user adds content to autosaving accordion section text block    Test text    1    Amended test text!
     ...    id:releaseMainContent
 
 Update embedded dashboard title
     user updates embedded dashboard in editable accordion section
     ...    Test embedded dashboard section
-    ...    Test embedded dashboard title updated
+    ...    Test embedded dashboard title Amended
```
duncan-at-hiveit, line 624: Very minor thing.  In the same convention where you've added "Amended" to the *start* of an amended bit of text, I'd advise doing the same here:

`Amended Test embedded dashboard title`

and below on the check at line 811.

Just adds a bit of extra consistency to the test style ( I can see that you're just updating the text that was already in the test of course :) )

tests/robot-tests/tests/admin_and_public_2/bau/publish_release_and_amend.robot
duncan-at-hiveit: Not suggesting we need to do this as part of this PR.

duncan-at-hiveit (approved):

PR #3784, opened 2023-01-20
Latest snapshots

benoutram:

The pipeline which runs the snapshots and produces Slack notifications has been failing recently but has just been fixed by https://github.com/dfe-analytical-services/explore-education-statistics/pull/3781 so we should be receiving notifications again soon!


cjrace (approved):

cjrace:
I moved this publication from one theme to another this morning - happy to approve.

PR #3791, opened 2023-01-25
EES-3786 react hook form for FiltersForm

amyb-hiveit:
Replaces Formik with React Hook Form (RHF) on the FiltersForm step of the Table Tool to improve the performance of large checkbox lists.

## Profiling results

Profiling for selecting and unselecting all options for a large filter group (208 options).

RHF is more performant than Formik as it isolates input components so doesn't re-render the entire form unnecessarily. It results in fewer commits and the rendering for each commit is slightly faster.

### Formik
formik-profile
### RHF
rhf-profile

### Typescript

There are some typescript performance issues that have made the implementation trickier than expected, these may be fixed in the next major version, see https://github.com/react-hook-form/react-hook-form/issues/6679 for discussion.

Specifically, I got the error `Type instantiation is excessively deep and possibly infinite` when trying to:
- using `FormProvider` and  `useFormContext` to access form data in nested components. Even just copying an example from the RHF site into a new component resulted in the error.
- passing down RHF useForm methods (e.g. `register` and `control`) to nested components when also passing the FormValues type. I've managed to get round this by not having too much nesting and bodging the types a bit, but it's not ideal.

I tried updating typescript to the latest but it didn't help. 

It's entirely possible I've been doing something that's causing the error, especially in regards to passing down the FormValues type.

### Formik all the way down

Our form components are many layered and very Formik specific so creating RHF alternatives isn't straightforward. I've also had to make RHF versions of some table tool specific components as well as generic form ones.

### Tests

There are a couple of failing tests in `TableToolWizard.test.tx` related to resetting the form and errors. These will need fixed.

## Next steps

Review this PR and decide:
- if it's worth continuing down this path 
- what could be done to improve the implementation of RHF
- should we change to RHF elsewhere? Project-wide would be a mammoth task, but it could be worth considering it for the table tool and/or new forms

### DIFF
--- a/src/explore-education-statistics-common/package-lock.json
+++ b/src/explore-education-statistics-common/package-lock.json
@@ -77,6 +77,7 @@
 				"react-dom": "17.0.2",
 				"react-helmet": "^6.1.0",
 				"react-highlight-words": "^0.16.0",
+				"react-hook-form": "^7.42.1",
 				"react-lazyload": "^3.0.0",
 				"react-leaflet": "^2.7.0",
 				"react-markdown": "^4.3.1",
@@ -16094,6 +16095,22 @@
 			},
 			"peerDependencies": {
 				"react": "^0.14.0 || ^15.0.0 || ^16.0.0-0"
+			}
+		},
+		"node_modules/react-hook-form": {
+			"version": "7.42.1",
+			"resolved": "https://registry.npmjs.org/react-hook-form/-/react-hook-form-7.42.1.tgz",
+			"integrity": "sha512-2UIGqwMZksd5HS55crTT1ATLTr0rAI4jS7yVuqTaoRVDhY2Qc4IyjskCmpnmdYqUNOYFy04vW253tb2JRVh+IQ==",
+			"dev": true,
+			"engines": {
+				"node": ">=12.22.0"
+			},
+			"funding": {
+				"type": "opencollective",
+				"url": "https://opencollective.com/react-hook-form"
+			},
+			"peerDependencies": {
+				"react": "^16.8.0 || ^17 || ^18"
 			}
 		},
 		"node_modules/react-is": {
@@ -33194,6 +33211,13 @@
 				"memoize-one": "^4.0.0",
 				"prop-types": "^15.5.8"
 			}
+		},
+		"react-hook-form": {
+			"version": "7.42.1",
+			"resolved": "https://registry.npmjs.org/react-hook-form/-/react-hook-form-7.42.1.tgz",
+			"integrity": "sha512-2UIGqwMZksd5HS55crTT1ATLTr0rAI4jS7yVuqTaoRVDhY2Qc4IyjskCmpnmdYqUNOYFy04vW253tb2JRVh+IQ==",
+			"dev": true,
+			"requires": {}
 		},
 		"react-is": {
 			"version": "16.13.1",
--- a/src/explore-education-statistics-common/package.json
+++ b/src/explore-education-statistics-common/package.json
@@ -4,6 +4,7 @@
   "description": "",
   "devDependencies": {
     "@hello-pangea/color-picker": "^3.2.2",
+    "@hookform/resolvers": "2.9.10",
     "@microsoft/applicationinsights-web": "^2.5.9",
     "@popperjs/core": "^2.11.2",
     "@reach/auto-id": "^0.16.0",
@@ -71,6 +72,7 @@
     "react-dom": "17.0.2",
     "react-helmet": "^6.1.0",
     "react-highlight-words": "^0.16.0",
+    "react-hook-form": "^7.42.1",
     "react-lazyload": "^3.0.0",
     "react-leaflet": "^2.7.0",
     "react-markdown": "^4.3.1",
--- a/src/explore-education-statistics-common/src/modules/table-tool/components/FiltersForm.tsx
+++ b/src/explore-education-statistics-common/src/modules/table-tool/components/FiltersForm.tsx
@@ -1,43 +1,41 @@
-import CollapsibleList from '@common/components/CollapsibleList';
-import {
-  Form,
-  FormFieldCheckboxSearchSubGroups,
-  FormFieldset,
-  FormGroup,
-} from '@common/components/form';
+import { FormFieldset } from '@common/components/form';
 import SummaryList from '@common/components/SummaryList';
-import SummaryListItem from '@common/components/SummaryListItem';
-import FormCheckboxSelectedCount from '@common/modules/table-tool/components/FormCheckboxSelectedCount';
-import FormFieldCheckboxGroupsMenu from '@common/modules/table-tool/components/FormFieldCheckboxGroupsMenu';
-import ResetFormOnPreviousStep from '@common/modules/table-tool/components/ResetFormOnPreviousStep';
 import TableQueryError from '@common/modules/table-tool/components/TableQueryError';
 import { InjectedWizardProps } from '@common/modules/table-tool/components/Wizard';
-import WizardStepFormActions from '@common/modules/table-tool/components/WizardStepFormActions';
 import WizardStepHeading from '@common/modules/table-tool/components/WizardStepHeading';
 import WizardStepSummary from '@common/modules/table-tool/components/WizardStepSummary';
+import RHFCheckboxSearchSubGroup from '@common/modules/table-tool/components/rhf-filters-form/RHFCheckboxSearchSubGroup';
+import RHFSelectedCount from '@common/modules/table-tool/components/rhf-filters-form/RHFSelectedCount';
+import RHFSelectedIndicators from '@common/modules/table-tool/components/rhf-filters-form/RHFSelectedIndicators';
+import RHFSelectedFilters from '@common/modules/table-tool/components/rhf-filters-form/RHFSelectedFilters';
+import RHFWizardStepFormActions from '@common/modules/table-tool/components/rhf-filters-form/RHFWizardStepFormActions';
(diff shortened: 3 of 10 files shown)


ntsim:
Okay, given this a big look and got a lots of thoughts.

Firstly, seems very promising! As mentioned in Slack, the flamegraph is very nice. To me, it seems logical to try and proceed with this.

## TypeScript issues

The issues with TypeScript can actually be resolved by upgrading! To do this, we need to upgrade each project's `package.json` and the project root's `package.json`. Once this is done, you might want to restart VSCode too.

However, Luke is in the area of upgrading frontend dependencies so it might be better to wait for his changes to land, or we might want to split out the TypeScript upgrade and do it separately. There's been a few issues with getting PNPM working so I'm not sure what will be more prudent right now.

## Not Formik all the way down!

> Our form components are many layered and very Formik specific so creating RHF alternatives isn't straightforward. I've also had to make RHF versions of some table tool specific components as well as generic form ones.

This isn't too true as we've kept most of the form components decoupled from any specific form provider, hence generic `Form*` components and Formik integrated `FormField*` components 😉

Integration with RHF isn't too painful as long as we provide a few more integration points. We definitely don't need to do as much copy-pasta as this PR seems to suggest!

For example, the `RHFCheckboxSearchSubGroups` can simplified to something like this:

```tsx
import { FieldValues } from 'react-hook-form'

export interface RHFCheckboxSearchSubGroupsProps<
  TFormValues extends FieldValues
> extends Omit<FormCheckboxSearchSubGroupsProps, 'name' | 'value'> {
  control: Control<TFormValues>;
  name: Path<TFormValues>;
}

export default function RHFCheckboxSearchSubGroups<
  TFormValues extends FieldValues
>({
  control,
  name,
  inputRef,
  ...props
}: RHFCheckboxSearchSubGroupsProps<TFormValues>) {
  const { field } = useController({ name, control });
  const value = useWatch({ name, control });

  return (
    <FormCheckboxSearchSubGroups
      {...props}
      inputRef={inputRef}
      name={name}
      value={value}
      onAllChange={(event, checked, options) => {
      	// TODO: Implement this properly
        field.onChange(options);
      }}
      onSubGroupAllChange={(event, checked, options) => {
      	// TODO: Implement this properly
        field.onChange(options);
      }}
    />
  );
}
```

Importantly, to integrate with RHF, an `inputRef` needs to be passed through to `FormCheckbox` via the various checkbox group layers. This provides a ref to the underlying checkbox `input` element i.e.

```tsx
export interface FormCheckboxProps {
  inputRef?: Ref<HTMLInputElement>;
}

const FormCheckbox = ({
	inputRef,
}: FormCheckboxProps) => {
  return (
    ...
    <input
      ref={inputRef}
    />
  );
}
```

We could also use `forwardRef` to achieve this, but it's a little simpler to use a prop as `forwardRef` doesn't work too nicely with components with type parameters.

Finally, back in `FiltersForm`, we'll need something like this:

```tsx
const { ref: indicatorRef, ...indicatorsField } = register('indicators');

return (
	<RHFCheckboxSearchSubGroups<FormValues>
	  {...indicatorsField}
	 	control={control}
    error={errors.indicators?.message}
	  id={`${formId}-indicators`}
	  inputRef={indicatorRef}
	  name="indicators"
	  legend={
	    <>
	      Indicators
	      <RHFSelectedCount name="indicators" control={control} />
	    </>
	  }
	  options={orderedIndicators.map(group => ({
	    legend: group.label,
	    options: group.options,
	  }))}
	  legendSize="m"
	  hint="Select at least one indicator below"
	  disabled={isSubmitting}
	/>
);
```

Once these things are in place, we no longer have excessive `FormCheckbox` rendering.

## Create a `FormProvider`

Of course, the above approach has a bunch of issues:

- We have to provide a `control` prop
- Have to destructure variables from `register('indicators')` - this is a faff!
- Still have props like `error`, `id`, etc that were previously just handled by various contexts
- Most importantly - we have to change **lots** of code! 😩

Using a `FormProvider` seems like the right call to get over these problems, especially if we want to deploy RHF more widely across the codebase.

We could define one that looks something like:

```tsx
// components/form/rhf/FormProvider.tsx
import { FormProvider as RHFormProvider } from 'react-hook-form';

interface FormProviderProps<TFormValues extends FieldValues> {
  children: ReactNode | ((form: UseFormReturn<TFormValues>) => ReactNode);
  initialValues?: UseFormProps<TFormValues>['defaultValues'];
  validationSchema?: Schema<TFormValues>;
}

export default function FormProvider<TFormValues extends FieldValues>({
  children,
  initialValues,
  validationSchema,
}: FormProviderProps<TFormValues>) {
  const form = useForm<TFormValues>({
    defaultValues: initialValues,
    resolver: validationSchema ? yupResolver(validationSchema) : undefined,
  });

  return (
    <RHFormProvider {...form}>
      {typeof children === 'function' ? children(form) : children}
    </RHFormProvider>
  );
}
```

The props for this should ideally be very similar to Formik's so that we can make as few changes as possible. We might want to add a few more props here to get feature parity with the current form.

Note that I'm suggesting we store our components in common's `src/components/form/rhf` directory. Ideally, we'd be able to migrate everything all at once, but we'll probably need to opt for a more gradual approach of building up our stock of RHF components to match the `FormField*` components.

## Consuming `FormProvider`

From a consuming component, we would use `FormProvider` like:

```tsx
import FormProvider from '@common/components/form/FormProvider';

const FiltersForm = (...) => {
  const validationSchema = useMemo(
    () =>
      Yup.object<FormValues>({
        indicators: Yup.array()
          .of(Yup.string())
          .required('Select at least one option from indicators'),
        filters: Yup.object(
          mapValues(subjectMeta.filters, filter =>
            Yup.array()
              .of(Yup.string())
              .min(
                1,
                `Select at least one option from ${filter.legend.toLowerCase()}`,
              ),
          ),
        ),
      }),
    [subjectMeta.filters],
  );

	return (
    <FormProvider
      initialValues={initialFormValues}
      validationSchema={validationSchema}
    >
      {form => {
        if (isActive) {
	        return ...;
        }
        
        return ...;
      }}
    </FormProvider>
  );
}
```

This is essentially the same as our usage of `Formik`. We should aim to reduce the number of changes as much as possible. Ideally, it would just be a bunch of import/component name changes.

Our field components should look something like:

```tsx
// components/form/rhf/RHFormFieldCheckboxSearchSubGroupsProps.tsx
import FormCheckboxSearchSubGroups from '../FormCheckboxSearchSubGroups';

export interface RHFormFieldCheckboxSearchSubGroupsProps<
  TFormValues extends FieldValues
> extends Omit<
    FormCheckboxSearchSubGroupsProps,
    'inputRef' | 'name' | 'value' // 👈️ Ignore `inputRef` as a prop
  > {
  name: Path<TFormValues>;
}

export default function RHFormFieldCheckboxSearchSubGroups<
  TFormValues extends FieldValues
>({ name, ...props }: FormFieldCheckboxSearchSubGroupsProps<TFormValues>) {
  const {
    formState: { errors },
    register,
    watch,
  } = useFormContext<TFormValues>();

  const { ref: inputRef, ...field } = useRegister(name, register);

  // We only do the following to avoid the dependency array
  // lint in `handleChange` 😞. We need this for feature 
  // parity with `FormFieldCheckboxSearchSubGroups`.
  const propsOnChange = props.onChange;
  const fieldOnChange = field.onChange;

  const handleChange = useCallback(
    async (event, option) => {
      if (propsOnChange) {
        propsOnChange(event, option);
      }

      if (event.isDefaultPrevented()) {
        return;
      }

      await fieldOnChange(event);
    },
    [propsOnChange, fieldOnChange],
  );

  return (
    <FormCheckboxSearchSubGroups
      {...field}
      {...props}
      error={get(errors, name)?.message as string}
      // ...Other props...
      onChange={handleChange}
    />
  );
}
```

Note that a lot of this implementation is essentially the same as our Formik integrated `FormFieldCheckboxSearchSubGroups`! Ideally, we should try to achieve feature parity where possible.

### The `useRegister` hook

In our `RHFormFieldCheckboxSearchSubGroups` example, we need to create a new `useRegister` hook. This should look like:

```ts
function useRegister<
  TFieldValues extends FieldValues,
  TFieldName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>
>(
  name: TFieldName,
  register: UseFormRegister<TFieldValues>,
): UseFormRegisterReturn<TFieldName> {
  // eslint-disable-next-line react-hooks/exhaustive-deps
  return useMemo(() => register(name), [name]);
}
```

The main purpose of this is to memoize the `register` call. If we don't do this, `register` will return a new object and cause excessive rendering due to dependency changes.

This is quite important and we definitely don't want to skip this step, or we'll lose any performance benefits of moving to RHF!

## Creating an `RHForm` component and `FormIdContext`

Another task will be to create a `RHForm` component that matches our existing `Form` component. Again, this should have feature parity and should essentially work the same.

I'd suggest that we tweak the `FormContext` component (this provides id related functionality to our various components). We don't want it to be confused with our new `FormProvider` component, so should rename `FormContext` to `FormIdContext`.

## The future and migrating forms

Ideally, I'd like us to consolidate back onto one form library for everything. We want to minimise our dependencies and eventually remove Formik.

Whilst it'd probably be possible to do a full migration in a 3-5 days, I doubt this will get prioritised given how wide reaching a change it is.

I think the main objectives will be to:

1. Build up our stockpile of RHF integrated components to match the `FormField` components.
2. Test each of these as much as possible by gradually migrate forms to use them.
3. Gain enough confidence that we migrate any remaining forms all at once in a dedicated ticket.

When we've migrated all forms, the cleanup should simply involve:

- Renaming the `components/form/rhf` directory to `components/form/field`, ensuring all our imports are correct
- Drop any `RH` prefixes from component names. We should avoid having overly specific component names that couple them to a specific implementation.

amyb-hiveit:
Superseded by https://github.com/dfe-analytical-services/explore-education-statistics/pull/3901

PR #3792, opened 2023-01-25
EES-4019 Add initial test data generators

ntsim:
This PR adds the initial implementation of test data generators that can be used in the backend unit tests.

The primary motivation for this is to significantly cut down the amount of boilerplate we produce for test data. We should hopefully be able to re-use these generators for producing test data set CSVs.

The API for the test data generators is heavily influenced by [Bogus library's](https://github.com/bchavez/Bogus) fluent API for test data generation. We have opted to build our own version of this to provide features for improved ergonomics when producing the kinds of test data we expect to need.

To get a better grasp of what test data generators can do, it's advisable to check out the unit tests in `DataFixtureTests` and `GeneratorTests`.

## General usage

Test data generators should be created via a `DataFixture`. This provides a context for generators in the test case and is important for:
- Managing seeds for generating deterministic random data. See section on [Determinism](https://github.com/bchavez/Bogus#determinism) in Bogus docs.
- Keeping track of various indices relating to the numbers of test data instances that have been created.

A typical `DataFixture` would be used like so:

```cs
var fixture = new DataFixture();
var generator = fixture.Generator<Person>();
```

In the above, we have created a `Generator` that produces instances of `Person`. The generator will need to be configured via setter methods to instruct it how to set properties on the instances as it generates them.

At its most basic, a generator may look something like this:

```cs
var fixture = new DataFixture();
var people = fixture.Generator<Person>()
    .ForInstance(s => s.Set(person => person.FirstName, "Jane"))
    .GenerateList(3);
```

This produces 3 instances of `Person`, each with their `FirstName` set to 'Jane'.

Multiple setters can be chained together to set properties across the instance i.e.

```cs
.ForInstance(s => s
    .Set(person => person.FirstName, "Jane")
    .Set(person => person.LastName, "Doe")
    .Set(person => person.Age, "50")
)
```

It is important to note that for setters of the same property, the latest setter will take precedence over previous setters. For example:

```cs
.ForInstance(s => s
    .Set(person => person.FirstName, "Jane")
    .Set(person => person.FirstName, "John")
    .Set(person => person.FirstName, "Jill")
```

The above will set the instance's `FirstName = "Jill"`.

### Instance property setters

Of course, it's unlikely that we want test data to look exactly the same. To vary the test data, we can also provide a setter function to `Set` to change the data in a more dynamic way.

This setter function provides an instance of Bogus' `Faker` class. Like Bogus, this provides a way for us to randomise the test data that we set. 

For example, to set a random first name:

```cs
.ForInstance(s => s
    .Set(person => person.FirstName, faker => faker.Name.FirstName())
)
```

To pick a random value from a set of values:

```cs
.ForInstance(s => s
    .Set(person => person.FirstName, faker => faker.PickRandom("Jane", "John", "Jill"))
)
```

Check out the [Bogus API](https://github.com/bchavez/Bogus#bogus-api-support) for more detail on the available test data methods on `Faker`.

Setter functions also receive parameters for the generated instance, and a context object containing various indices relating to the instance.

### Range setters

One of the main reason we have implemented our own test data generator API is to make it easier to control what is generated at specific ranges in the generated sequence. For example:

```cs
var people = new Generator<Person>()
    .ForRange(..2, s => s
        .Set(p => p.FirstName, "Jane"))
    .ForRange(2.., s => s
        .Set(p => p.FirstName, "John"))    
    .GenerateList(4);
```

In the above, we set the `FirstName = "Jane"` for the first 2 people in the list. However, for the next 2 people, we set `FirstName = "John"` instead.

### Combining instance and range setters

Both instance and range setters can be used interchangeably e.g.

```cs
var people = new Generator<Person>()
    .ForInstance(s => s
        .Set(p => p.FirstName, "Jane"))
    .ForRange(1..3, s => s
        .Set(p => p.FirstName, "John"))    
    .GenerateList(5);
```

The above generat instanaces 2 and 3 with `Firstname = "John"`, whilst instances 1, 4 and 5 will have `FirstName = "Jane"`.

The important things to remember is that range setters take precedence over instance setters. If a range is applicable to the current instance's index, it will override any value previously set by an instance setter.

If the ranges of range setters overlap, they will applied in the order that they were originally registered. For example:

```cs
.ForRange(..2, s => s
    .Set(p => p.FirstName, "John"))    
.ForRange(..2, s => s
    .Set(p => p.FirstName, "Jane"))
```

The first 2 instances will have `FirstName = "Jane"`.

## Extension methods for generators

Generators are designed so that they can be used in as generic a way as possible. However, for things that may express relationships, such as database models, it is useful to provide additional methods to make test data generation easier.

We may also want to just provide methods that set reasonable defaults, or configure the instances in some common way. The aim is to maximize convenience and reduce the amount of boilerplate that we have to write.

For this, we can add extension methods to `InstanceSetters<T>`. This typically looks like:

```cs
public static class PersonGeneratorExtension
{
    public static InstanceSetters<Person> SetDefaults(
        this InstanceSetters<Person> setters)
        => setters
            .Set(p => p.FirstName, f => f.Name.FirstName())
            .Set(p => p.LastName, f=> f.Name.LastName());
}
```

The above creates an extension method called `SetDefaults` on the setter builder that we've seen previously. For example, we can now do this:

```cs
.Generator<Person>()
.ForInstance(s => s.SetDefaults());
```

This will generate any instances of `Person` with random first and last names.

As part of this PR, we have provided a bunch of similar extensions for most of the statistics DB models. These are not meant to be comprehensive and may need further tweaking as we develop the test data generators further.

### Why extension methods on `InstanceSetters<T>`?

After reviewing this code, you may be wondering why not place extension methods on the `Generator`, or even extend the `Generator` class. The main reason is that we can prevent consumers from calling incorrect methods by using the `InstanceSetters` class. We want to avoid prematurely calling methods that may negatively impact on the generation of test data.

Additionally, extending the `Generator` class is a particularly bad approach as we want a fluent builder API (i.e. methods chain together). This is quite awkward to achieve in a nice way due to the limitations of the C# inheritance model. For example, you end up needing to declare the `Generator` class like this:

```cs
public class Generator<T, TGenerator> 
    where T : class
    where TGenerator : Generator<T, TGenerator>
{
    public TGenerator SomeBuilderMethod(...)
    {
        // Do some stuff...

        return This; // 😭 We can't use `this`, we have to use `This`
    }

    private TGenerator This => (TGenerator)this;
}
```
This is pretty ugly as builders and inheritance do not mix very well!


src/GovUk.Education.ExploreEducationStatistics.Data.Model/Observation.cs
```diff
@@ -22,7 +22,7 @@ public class Observation
 
         public Dictionary<Guid, string> Measures { get; set; }
 
-        public ICollection<ObservationFilterItem> FilterItems { get; set; }
+        public List<ObservationFilterItem> FilterItems { get; set; }
```
duncan-at-hiveit, line 25: I'm all for switching to Lists, but just interested to know what the driver was here and in a few other places?

duncan-at-hiveit (approved):
All looks awesome!  Can't see anything in this PR that I'd have any issues using.  I'm sure we'll find some use cases that we need supporting down the line but I reckon we just need to start using this and see how we get on.  Nicely done :)

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Observation.cs
ntsim: Think this might actually be a straggler from my next PR (which depends on this one)!

I've reverted it back to `ICollection`

src/GovUk.Education.ExploreEducationStatistics.Data.Model.Tests/Fixtures/SubjectGeneratorExtensions.cs
```diff
@@ -0,0 +1,106 @@
...
+            f => f.Observations,
+            (_, subject) =>
+            {
+                var list = observations.ToList();
+
+                list.ForEach(observation => observation.Subject = subject);
+
+                return list;
+            }
+        );
```
benoutram, line 80: Few cases here where I think `f` should be `s` 🙂 .

benoutram (approved):
Seems like a great addition and looking forward to trying it out in practice 🙂 .

PR #3796, opened 2023-01-26
EES-4047 show not latest data tag on release page

amyb-hiveit:
Adds the 'This is not the latest data' tag to release pages.
notlatestdata

### DIFF
--- a/src/explore-education-statistics-frontend/src/modules/find-statistics/__tests__/PublicationReleasePage.test.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/find-statistics/__tests__/PublicationReleasePage.test.tsx
@@ -6,22 +6,14 @@ import React from 'react';
 import { testPublication, testRelease } from './__data__/testReleaseData';
 
 describe('PublicationReleasePage', () => {
-  test('renders national statistics image', () => {
-    const { container } = render(
-      <PublicationReleasePage release={testRelease} />,
-    );
-
-    expect(
-      container.querySelector(
-        'img[alt="UK statistics authority quality mark"]',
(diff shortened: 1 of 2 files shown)


luke-h1 (approved):

src/explore-education-statistics-frontend/src/modules/find-statistics/__tests__/PublicationReleasePage.test.tsx
```diff
@@ -224,6 +235,18 @@ describe('PublicationReleasePage', () => {
...
+  test('renders national statistics image', () => {
+    const { container } = render(
+      <PublicationReleasePage release={testRelease} />,
+    );
+
+    expect(
+      container.querySelector(
+        'img[alt="UK statistics authority quality mark"]',
+      ),
+    ).toBeDefined();
```
ntsim, line 247: Shouldn't this be in `toBeTheDocument`? 

Additionally, `toBeDefined` checks if it's `undefined` or not, so this would most likely always pass as `querySelector` returns null if the element doesn't exist.

ntsim (changes requested):

src/explore-education-statistics-frontend/src/modules/find-statistics/__tests__/PublicationReleasePage.test.tsx
amyb-hiveit: Good spot! I just moved this test but didn't really look at what was in it. I've changed it and the test for the image not showing for official stats now.

ntsim (approved):

PR #3797, opened 2023-01-26
fix(frontend): fix dependency hoisting

luke-h1:


.npmrc
```diff
@@ -8,10 +8,10 @@ auto-install-peers=true
...
-# the frontend project needing them present in the relative 
-# 'node_modules' directory in order to build properly.
-hoist-pattern[]=!react/*
-hoist-pattern[]=!react-dom/*
-hoist-pattern[]=!formik/*
-hoist-pattern[]=!applicationinsights/*
\ No newline at end of file
+# exclude react, react-dom, formik & applicationinsights 
+# from hoisting due to the frontend project needing them present 
+# in the relative 'node_modules' directory in order to build properly.
```
ntsim, line 13: Would reword this so we don't need to update it every time we need to change this i.e.

```
# exclude specific dependencies from hoisting as they need to be present
# for the project 'node_modules' directories in order to build properly.
```

PR #3818, opened 2023-01-27
EES-4054 Fix XPath expressions getting publication details when creating the Find Statistics snapshot

benoutram:
This PR fixes the expressions getting the details of each publication for creating/validating the Find Statistics snapshot.

It was incorrectly getting the details from the first publication list item on the page meaning that large parts of the snapshot data are incorrect.

It results in false negative reports that there are no snapshot differences during validation, when there are changes, as seen on 26/01/2023.

Here's an example where release type, published date and theme in the snapshot data are wrong.

Publication data in the snapshot:

image-20230127-090012

Current view of the same publication on the Prod Find Statistics page:

image-20230127-085929

### DIFF
--- a/tests/robot-tests/scripts/create_snapshots.py
+++ b/tests/robot-tests/scripts/create_snapshots.py
@@ -108,15 +108,9 @@ class SnapshotService:
                 lambda driver: driver.find_element(By.XPATH, "//*[@data-testid='publicationsList']")
             )
 
-            publications_list = driver.find_element(By.XPATH, "//*[@data-testid='publicationsList']")
+            publications_list_items = driver.find_elements(By.XPATH, "//*[@data-testid='publicationsList']/li")
 
-            all_publications = publications_list.find_elements(By.CSS_SELECTOR, "li")
-
-            for publication in all_publications:
-                publication.click()
-
-                publication_details = publication.find_element(By.CSS_SELECTOR, "dl")
-
+            for publication in publications_list_items:
                 publication_title = publication.find_element(By.CSS_SELECTOR, "h3").text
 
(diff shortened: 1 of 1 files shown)


tests/robot-tests/scripts/create_snapshots.py
```diff
@@ -108,31 +108,21 @@ def _create_find_statistics_snapshot(self) -> str:
...
 
-            publications_list = driver.find_element(By.XPATH, "//*[@data-testid='publicationsList']")
-
-            all_publications = publications_list.find_elements(By.CSS_SELECTOR, "li")
-
-            for publication in all_publications:
-                publication.click()
-
-                publication_details = publication.find_element(By.CSS_SELECTOR, "dl")
+            publications_list_items = driver.find_elements(By.XPATH, "//*[@data-testid='publicationsList']/li")
```
duncan-at-hiveit, line 111: So I guess previously the old xpath selectors using the double slash prefixes were essentially breaking out of their parent elements and searching from the root of the DOM - and returning the first match they find.

duncan-at-hiveit (approved):

tests/robot-tests/scripts/create_snapshots.py
benoutram: Yes that's correct.

I believe that although `//` is a descendent selector, when it was used at the beginning of the statement
`publication_details.find_element(By.XPATH, "//*[dl]//div//dd[@data-testid='theme']")` the `//` doesn't mean descendent of the `publication_details` element, it's relative from anywhere on the page.

By changing expressions to begin with `./` we now select relative from the current element.

PR #3819, opened 2023-01-27
Latest snapshot changes

benoutram:

There are a lot more changes than we would expect if only 5 new releases were published on 26/01/2023.

This is explained by [EES-4054](https://dfedigital.atlassian.net/browse/EES-4054) fixed by https://github.com/dfe-analytical-services/explore-education-statistics/pull/3818 which meant the snapshot data had previously been incorrect.


cjrace (approved):
Checked through the new script and all looks good! Alphabetical ordering makes the most sense to minimise confusion with future changes to release dates too.

PR #3821, opened 2023-01-27
EES-4056 Tidy up of approving releases ahead of EES-4001

benoutram:
This PR is a tidy-up of code around approving releases in the Admin service, made ahead of raising a PR for [EES-4001](https://dfedigital.atlassian.net/browse/EES-4001) which will add a new option to allow updating the release published date on approval.

This PR should make for a clearer separation of the changes made specifically for EES-4001 and these changes which are not directly related but clean up code in the same area.

- Rename `ReleaseStatusCreateViewModel` to `ReleaseStatusCreateRequest`.
- Rename `LatestInternalReleaseNote` to `InternalReleaseNote` in the back-end `ReleaseStatusCreateRequest` and the front-end form component since `LatestInternalReleaseNote` doesn't make sense at the approval form and request level.
-  Allow `InternalReleaseNote` to be null since it's not a required field on all status transitions.
- Add `ICreatedTimestamp` interface to `ReleaseStatus` so that created date is set automatically.
- Remove unused `Release.DataLastPublished` to avoid any confusion about what dates need to be updated later on.
- Remove `Publication.Published` date also to avoid confusion about what dates need to be updated later. It's ambiguous what this date represented other than being used as a flag to indicate the publication has published releases.
- Migrate `NotifySubscribers` and `NotifiedOn` from the latest 'Approved' release status of the release if one exists, to new fields of `Release`.
- Other changes to make sure back-end / front-end are in agreement about nullability of fields.

Regarding the `NotifySubscribers` change, I can't see any need for this to be currently implemented in `ReleaseStatus` and require fetching all statuses and order them to get the latest in order to know what the effective values should be.

A migration is implemented to migrate values from the latest 'Approved' release status and at a later date we can drop those columns from `ReleaseStatus`.

### Future work to drop unused columns

The following can be dropped in future:

- ReleaseStatus.Approved
- ReleaseStatus.NotifiedOn
- Publication.Published
- Release.DataLastPublished

This work is described by [EES-4058](https://dfedigital.atlassian.net/browse/EES-4058).

### UI test report

image

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
@@ -26,27 +26,23 @@ namespace GovUk.Education.ExploreEducationStatistics.Publisher.Services
 
         public async Task NotifySubscribersIfApplicable(params Guid[] releaseIds)
         {
-            var releasesToNotify = _context.Releases
-                .Include(r => r.ReleaseStatuses)
-                .Where(r => releaseIds.Contains(r.Id))
-                .ToList()
-                .Where(r => r.NotifySubscribers)
-                .ToList();
+            var releasesToNotify = await _context.Releases
+                .Where(r => releaseIds.Contains(r.Id) && r.NotifySubscribers)
+                .ToListAsync();
+
             var messages = await releasesToNotify
                 .ToAsyncEnumerable()
                 .SelectAwait(async release => await BuildPublicationNotificationMessage(release))
                 .ToListAsync();
+
             if (messages.Count > 0)
             {
                 await _storageQueueService.AddMessages(ReleaseNotificationQueue, messages);
                 releasesToNotify
                     .ForEach(release =>
                     {
-                        var status = release.ReleaseStatuses
-                            .OrderBy(rs => rs.Created)
-                            .Last();
-                        _context.Update(status);
-                        status.NotifiedOn = DateTime.UtcNow;
+                        _context.Releases.Update(release);
+                        release.NotifiedOn = DateTime.UtcNow;
                     });
                 await _context.SaveChangesAsync();
             }
--- a/src/explore-education-statistics-admin/src/services/releaseService.ts
+++ b/src/explore-education-statistics-admin/src/services/releaseService.ts
@@ -85,7 +85,7 @@ export interface UpdateReleaseRequest extends BaseReleaseRequest {
 
 export interface CreateReleaseStatusRequest {
   approvalStatus: ReleaseApprovalStatus;
-  latestInternalReleaseNote?: string;
+  internalReleaseNote?: string;
   publishMethod?: 'Scheduled' | 'Immediate';
   publishScheduled?: string;
   nextReleaseDate?: PartialDate;
@@ -179,7 +179,7 @@ export interface ReleaseStatus {
   approvalStatus: ReleaseApprovalStatus;
   notifySubscribers: boolean;
   created: string;
-  createdByEmail: string;
+  createdByEmail?: string;
   releaseVersion: number;
 }
 
--- a/src/explore-education-statistics-admin/src/services/releaseService.ts
+++ b/src/explore-education-statistics-admin/src/services/releaseService.ts
@@ -89,6 +89,7 @@ export interface CreateReleaseStatusRequest {
   publishMethod?: 'Scheduled' | 'Immediate';
   publishScheduled?: string;
   nextReleaseDate?: PartialDate;
+  notifySubscribers?: boolean;
 }
 
 type PublishingStage =
@@ -177,7 +178,6 @@ export interface ReleaseStatus {
   releaseStatusId: string;
   internalReleaseNote: string;
   approvalStatus: ReleaseApprovalStatus;
-  notifySubscribers: boolean;
   created: string;
   createdByEmail?: string;
   releaseVersion: number;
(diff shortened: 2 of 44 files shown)


src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
```diff
@@ -26,27 +26,23 @@ public class NotificationsService : INotificationsService
...
         public async Task NotifySubscribersIfApplicable(params Guid[] releaseIds)
         {
-            var releasesToNotify = _context.Releases
-                .Include(r => r.ReleaseStatuses)
-                .Where(r => releaseIds.Contains(r.Id))
-                .ToList()
-                .Where(r => r.NotifySubscribers)
-                .ToList();
+            var releasesToNotify = await _context.Releases
+                .Where(r => releaseIds.Contains(r.Id) && r.NotifySubscribers && r.NotifiedOn == null)
```
duncan-at-hiveit, line 30: I can't see any harm in testing the NotifiedOn date being null here, but in reality should we ever be able to reach this bit of code when NotifiedOn is *not* null?

Either way, happy for it to stay this way, just curious.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
```diff
@@ -26,27 +26,23 @@ public class NotificationsService : INotificationsService
...
                 await _storageQueueService.AddMessages(ReleaseNotificationQueue, messages);
                 releasesToNotify
                     .ForEach(release =>
                     {
-                        var status = release.ReleaseStatuses
-                            .OrderBy(rs => rs.Created)
-                            .Last();
-                        _context.Update(status);
-                        status.NotifiedOn = DateTime.UtcNow;
+                        _context.Update(release);
```
duncan-at-hiveit, line 44: Maybe `_context.Releases.Update(release);` for the type safety?  I know this code isn't 100% new but I look to point this out if I see any occurrences in PRs now where we're just using the blanked "Update" or "Add" / "AddRange" etc methods rather than the collection-specific ones that provide a bit more safety.

src/explore-education-statistics-admin/src/services/releaseService.ts
```diff
@@ -89,6 +89,7 @@ export interface CreateReleaseStatusRequest {
   publishMethod?: 'Scheduled' | 'Immediate';
   publishScheduled?: string;
   nextReleaseDate?: PartialDate;
+  notifySubscribers?: boolean;
```
duncan-at-hiveit, line 92: Is anything using this new field yet?

duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
benoutram: That's correct, I don't think we should ever be able to reach here with `NotifiedOn` having a not null value.

I was copying the previous behaviour of  the previous get accessor `Release.NotifySubscribers` when it depended on `ReleaseStatsu`:

```
        public bool NotifySubscribers
        {
            get
            {
                var status = ReleaseStatuses
                           .OrderBy(rs => rs.Created)
                           .LastOrDefault();
                return status != null
                       && status.NotifySubscribers
                       && status.NotifiedOn == null;
            }
        }
```

I think we can get rid of this check.

I suppose it's useful to stop subscribers being emailed multiple times if we were to ever force a release through the publishing process more than once rather than amending, but it seems unlikely we'd need to do that and it can still be prevented by altering `NotifySubscribers` to false.

benoutram: > That's correct, I don't think we should ever be able to reach here with `NotifiedOn` having a not null value.
> 
> I was copying the previous behaviour of the previous get accessor `Release.NotifySubscribers` when it depended on `ReleaseStatus`:
> 
> ```
>         public bool NotifySubscribers
>         {
>             get
>             {
>                 var status = ReleaseStatuses
>                            .OrderBy(rs => rs.Created)
>                            .LastOrDefault();
>                 return status != null
>                        && status.NotifySubscribers
>                        && status.NotifiedOn == null;
>             }
>         }
> ```
> 
> I think we can get rid of the `NotifiedOn == null` check.
> 
> I suppose it's useful to stop subscribers being emailed multiple times if we were to ever force a release through the publishing process more than once rather than amending, but it seems unlikely we'd need to do that and it can still be prevented by altering `NotifySubscribers` to false.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
benoutram: Ah yep, made that change now, thanks. 🙂

src/explore-education-statistics-admin/src/services/releaseService.ts
duncan-at-hiveit: Aha OK cool, nice spot!

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
duncan-at-hiveit: Ta

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
duncan-at-hiveit: Sure, happy either way thanks!

PR #3862, opened 2023-02-01
EES-4072 Add new BAU method to allow updating the published date of published releases

benoutram:
This PR adds a new BAU endpoint to allow updating the published date of a release. It's only intended to be used by UI tests and for maintenance support requests.

The new published date/time can be expressed according to the ISO 8601 format in the request body of a HTTP PATCH request to update only the published date of the release.

Example request:

```
PATCH https://{adminUrl}/api/releases/{releaseId}/published

{
  "published": "2022-08-31 08:30:00Z"
}
```

All three of these example values would be treated as the same moment in time when converted to UTC before being saved to the database:

- Date time in UTC: `2022-08-31 08:30:00Z`
- Date time offset 1 hour ahead of UTC e.g. British Summer Time `2022-08-31 09:30:00+01:00`
- Date time assumed to be in local time: `2022-08-31 09:30:00`.

All three requests would be saved to the database as `2022-08-31 08:30:00.0000000`.

It does not allow setting a date in the future since it would have the effect of un-publishing the release.

It does not allow updating an unpublished release since a release must go through the publishing process to have a published date set.

It updates the cached public view of the release so that this doesn’t need to be performed as a separate BAU task.

It updates the publication’s latest cached release if the release being updated is the latest published release of the publication since this is a separate cache entry.

It updates  the interface of `IReleaseCacheService` splitting up `UpdateRelease` into two separate methods, `UpdateRelease` and `UpdateReleaseStaged`. This is to make it clear that `expectedPublishDate` is a required field when updating a staged release and not relevant if not.

Unfortunately the changes in this PR are pretty unwieldy since the Admin now depends on `ReleaseCacheService` to update the cached release, which in turn relies on `Content.Services.ReleaseService` and all of its dependencies to get the public release content.

This is not a new pattern for the Admin, but this change would have been much cleaner if we could rely on invalidating cache entries and the lazy caching mechanism instead of needing to update cache entries, in conjunction with not needing to pre-cache releases into a staging directory during publishing.

### Other changes
- Removes unused `DataLastPublished` from `ReleaseCacheViewModel` missed by [EES-4056](https://dfedigital.atlassian.net/browse/EES-4056) in https://github.com/dfe-analytical-services/explore-education-statistics/pull/3821.

### UI test report

image


src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleasesController.cs
```diff
@@ -214,60 +214,69 @@ public async Task<ActionResult<List<ReleaseViewModel>>> ListScheduledReleases()
...
-        [HttpGet("releases/{releaseId}/checklist")]
+        [HttpGet("releases/{releaseId:guid}/checklist")]
         public async Task<ActionResult<ReleaseChecklistViewModel>> GetChecklist(Guid releaseId)
         {
             return await _releaseChecklistService
                 .GetChecklist(releaseId)
                 .HandleFailuresOrOk();
         }
+
+        [HttpPatch("releases/{releaseId:guid}")]
```
duncan-at-hiveit, line 273: There's no hint of this being an endpoint specifically for updating just one facet of a Release by looking at the URL alone.  SoI might be confused if I saw 2 ways to update a Release using the URL `releases/{releaseId:guid}` - one that operates via a PUT and this one via a PATCH.

Wondering if it's worth being specific that this is an endpoint dealing specifically with Release Published dates, as per the `UpdateReleasePublished` method name, or if we make the `UpdateReleasePublished` method name and the `ReleasePublishedUpdateRequest` class name more reusable for other PATCHable fileds for the future.

Or merge this into the existing PUT on the `releases/{releaseId:guid}` route.

Or maybe we don't need to any of the above, as I'm not too precious about any particular one of them!  I'd possibly be tempted personally to just change the URL slightly to make it obvious that this is specifically something different from what you'd get with the existing PUT.

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/ReleasePublishedUpdateRequest.cs
```diff
@@ -0,0 +1,13 @@
+﻿#nullable enable
+using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Requests;
+
+public class ReleasePublishedUpdateRequest
```
duncan-at-hiveit, line 7: Could this be a `record`?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseService.cs
```diff
@@ -332,6 +337,49 @@ public class ReleaseService : IReleaseService
...
+                    }
+
+                    // Prevent assigning a future date since it would have the effect of un-publishing the release
+                    if (request.Published > DateTime.UtcNow)
+                    {
+                        return ValidationActionResult(ReleasePublishedCannotBeFutureDate);
+                    }
+
+                    _context.Releases.Update(release);
+                    release.Published = request.Published ?? DateTime.UtcNow;
```
duncan-at-hiveit, line 361: We know that `request.Published` is not null by this point, so we should be able to skip this check.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseService.cs
```diff
@@ -332,6 +337,49 @@ public class ReleaseService : IReleaseService
...
+                    await _releaseCacheService.UpdateRelease(
+                        releaseId,
+                        publicationSlug: release.Publication.Slug,
+                        releaseSlug: release.Slug);
+
+                    if (release.Publication.LatestPublishedReleaseId == releaseId)
+                    {
+                        // This is the latest published release so also update the latest cached release
+                        // for the publication which is a separate cache entry
+                        await _releaseCacheService.UpdateRelease(
```
duncan-at-hiveit, line 374: Noice

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/ContentService.cs
```diff
@@ -95,49 +93,70 @@ public async Task DeletePreviousVersionsDownloadFiles(params Guid[] releaseIds)
...
-        private async Task CacheLatestRelease(Publication publication, PublishContext context, params Guid[] includedReleaseIds)
-        {
-            var latestRelease = await _releaseService.GetLatestRelease(publication.Id, includedReleaseIds);
-            await _releaseCacheService.UpdateRelease(context.Staging,
-                context.Published,
-                latestRelease.Id,
-                publication.Slug);
+            foreach (var release in releases)
+            {
+                await _releaseCacheService.UpdateReleaseStaged(
```
duncan-at-hiveit, line 133: Oh yeah I bet you were loving having to put these variations in!

duncan-at-hiveit:
Had a general conversation around testing the timezone behaviour when posting datetimes with no tz info / GMT / UTC etc etc, to see what's best to put in operating instructions.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseService.cs
benoutram: I've made a change now above to begin with this:
`var newPublishedDate = request.Published?.ToUniversalTime() ?? DateTime.UtcNow;`

and then later on here `release.Published = newPublishedDate`.

`request.Published` is optional because that's the only way I could get the `[Required]` validation enforced. If it's not marked as optional and not present in the request body, `Published` will be `0000-00-00`, i.e. the default value of `DateTime` which we don't want to happen.

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/ReleasePublishedUpdateRequest.cs
benoutram: Yes, made this change, thanks.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/ContentService.cs
benoutram: Ha ha, not sure if I've made this better or worse. I think it's a little bit clearer now, at least for understanding when the cache key will use the staging directory, with it now being a separate key!

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleasesController.cs
benoutram: I was thinking that we could have had multiple patch methods in future each with different request types and that the routing would be clever enough to choose the best candidate based on the body, but a quick experiment reveals this is not the case! 😢 

I think merging it into the existing PUT functionality isn't too nice because it requires specifying the other field values in the update and that's why I decided to use PATCH here to make it simple to update a single field.

I've gone with your suggestion for now to change the URL slightly, adding `/publish` to the end to link it with the `UpdateReleasePublished` method name. I don't think this naming is too RESTful though as `published` is not an entity itself which is being updated.

Doing a quick read up it seems like a [JSON Patch](https://jsonpatch.com/) document is the way to go here so that we can offer the ability to patch any number of fields and have a way of specifying which fields should be updated, so as to not accidently set null values for fields not included in the request.

It seems like a bit overkill for now so will leave the current implementation with just the slight change to the URL.

benoutram:
> Had a general conversation around testing the timezone behaviour when posting datetimes with no tz info / GMT / UTC etc etc, to see what's best to put in operating instructions.

Made an additional commit to convert the DateTime to UTC and will update the example in this PR, thanks.

duncan-at-hiveit (approved):

PR #3863, opened 2023-02-01
EES-4074, EES-4017 Update `BlobStorageService.StreamBlob` to use Azure.Storage.Blobs v12

ntsim:
This PR primarily updates the `BlobStorageService.StreamBlob` method to use the updated `Azure.Storage.Blobs` v12 SDK.

## Relevant changes

- Updating the Azure.Storage.Blobs to the latest resolves [EES-4017](https://dfedigital.atlassian.net/browse/EES-4017).
- Refactors `BlobStoragService.DownloadBlobText` to use the recommended `DownloadContentAsync` method to download the blob into memory. See [docs](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-download#download-to-a-string).

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Services/BlobStorageService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Services/BlobStorageService.cs
@@ -366,7 +366,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Services
             {
                 if (exception.Status == 404)
                 {
-                    throw new FileNotFoundException($"Could not find file at {containerName}/{path}");
+                    ThrowFileNotFoundException(containerName, path);
                 }
 
                 throw;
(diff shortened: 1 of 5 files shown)


src/GovUk.Education.ExploreEducationStatistics.Common/Services/BlobStorageService.cs
```diff
@@ -350,43 +350,47 @@ private static async Task<string> UploadToTemporaryFile(IFormFile file)
...
         public async Task<string> DownloadBlobText(IBlobContainer containerName, string path)
         {
             var blob = await GetBlobClient(containerName, path);
 
-            if (!await blob.ExistsAsync())
+            try
             {
-                throw new FileNotFoundException($"Could not find file at {containerName}/{path}");
+                var response = await blob.DownloadContentAsync();
+                return response.Value.Content.ToString();
```
benoutram, line 383: Cool, I'm just wondering what `Value` is here because this differs slightly from the docs linked to in the description?

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Common/Services/BlobStorageService.cs
ntsim: The difference with the docs is that the docs implicitly convert the `Response<BlobDownloadResult>` to a `DownloadResult` by specifying the type.

Without specifying the type, we need to call `Response.Value` to unwrap the `BlobDownloadResult`.

benoutram: Ah yeah, thanks for the explanation about the implicit conversion. 🙂

PR #3864, opened 2023-02-01
EES-4075 Move `CsvUtils` into Common

ntsim:
This PR moves `CsvUtils` into Common so that they can be used in the other projects e.g. `Data.Services` for EES-3777.


src/GovUk.Education.ExploreEducationStatistics.Common/Utils/CsvUtils.cs
```diff
@@ -0,0 +1,329 @@
...
+
+        try
+        {
+            return (GeographicLevel) GeographicLevelLookup.ConvertFromProvider.Invoke(value)!;
+        }
+        catch (ArgumentOutOfRangeException)
+        {
+            throw new InvalidGeographicLevelException(value!);
+        }
+    }
```
benoutram, line 328: I wonder if `GeographicLevelLookup` would be better refactored to being a singleton service in the Importer.

Either that would use `CsvUtils.Value`, or use `CsvUtils.Value` first then `GeographicLevelLookup`.

It would leave CsvUtils a bit purer without referencing a column name here.

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/CsvUtils.cs
```diff
@@ -0,0 +1,329 @@
...
+using GovUk.Education.ExploreEducationStatistics.Common.Model.Data;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Utils;
+
+public static class CsvUtils
+{
+    private static readonly EnumToEnumLabelConverter<GeographicLevel> GeographicLevelLookup = new();
+
+    public const string DefaultFilterGroupLabel = "Default";
+    public const string DefaultFilterItemLabel = "Not specified";
```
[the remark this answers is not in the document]
ntsim: Yep good shout, done!

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/CsvUtils.cs
ntsim: I've opted to move it out into `GeographicLevelCsvUtils` instead of a singleton service. Going for a singleton service felt a bit overkill.

Of course, if you'd prefer to go with that I can make the change, but would personally prefer to keep the PR as simple as possible.

benoutram: Latest change to add `GeographicLevelCsvUtils` seems like a good way to go, thanks!

benoutram (approved):

PR #3871, opened 2023-02-02
EES-3995 - upgrade python dependencies

luke-h1:
This PR upgrades the majority of python dependencies (except from `pyderman`. I've chosen to not upgrade `pyderman` due to the fact that their releases are often not well tested & typically introduce bugs for people on different machines)

image


benoutram (approved):
Tried this in Windows 11, Python 3.10.9, removing virtualenv first and it was all fine, even got a full passing UI test run to match yours! 🙂

PR #3872, opened 2023-02-02
EES-3960 - added BAU cache-clearing endpoint to target publication.js…

duncan-at-hiveit:


src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/BauCacheController.cs
```diff
@@ -147,6 +147,26 @@ public async Task<ActionResult> ClearPublicCachePublications()
...
+        {
+            if (request.Paths.Any())
+            {
+                var pathString = request.Paths.JoinToString('|');
+
+                await _publicBlobStorageService.DeleteBlobs(
+                    BlobContainers.PublicContent,
+                    options: new DeleteBlobsOptions
+                    {
+                        IncludeRegex = new Regex($"^publications/[^/]*/({pathString})$")
```
benoutram, line 162: Maybe we should be a bit more restrictive elsewhere and use `[^/]*` rather than `.*`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/BauCacheController.cs
```diff
@@ -179,6 +199,20 @@ public class ClearPublicCacheReleasePathsViewModel
             public HashSet<string> Paths { get; set; } = new();
         }
 
+        public class ClearPublicCachePublicationPathsViewModel
```
benoutram, line 202: Going with our current naming convention this would be `ClearPublicCachePublicationPathsRequest` or even `DeletePublicCachePublicationPathsRequest` and could be a record type?

PR #3873, opened 2023-02-02
EES-4001 Add option 'Update published date' to Sign off tab when approving an amended release for publication

benoutram:
⚠️ The UI tests of this change depend on [EES-4072](https://dfedigital.atlassian.net/browse/EES-4072) in https://github.com/dfe-analytical-services/explore-education-statistics/pull/3862 which adds a method which the UI tests can use to force a published date to a past date.

This PR adds a new checkbox 'Update published date' to the Admin Sign off tab when approving an **amendment** release for publication.

It allows the analysts to decide whether to update the published date to match the time the publishing process completes.

It applies to both immediate and scheduled publishing methods.

If not checked (default), the behaviour remains the same, i.e. an amendment is given the published date of the previous version.

image

A complication to doing this arises from the publishing process and the way we pre-cache release content for both immediate and scheduled publishing methods. Already, the code to fetch and cache release content has to expect that the release might not be published yet for this pre-caching, and in this case sets an expected date publish date in the view model based on logic around whether it's an amendment or not and when we expect publishing to complete (immediate=now, scheduled=future date).

This happens in separate Azure functions to the post publishing completion step which updates the published date of the release which makes it publicly accessible.

Both need to follow the same logic, i.e. `Content.Services.ReleaseService.GetRelease` used to pre-cache or pre-cache and stage the release, and `Publisher.Services.ReleaseService.SetPublishedDate` used in the post publishing completion step.

These are both tidied up now to use the same `ReleaseRepository.GetPublishedDate` method to resolve the published date based on version and `updatePublishedDate` flag.

However it feels like it would be much cleaner if pre-caching didn't exist, and also if the publish date fields were separated, one to control or audit when a release is published, and one for the display date which can be altered and set ahead of time.

### Other changes

- Remove unused `PublishScheduled` and `NextReleaseDate` from `ReleaseCreateRequest` and check fields of the saved Release in `ReleaseServiceTests` after creating a release. This change is tagged as EES-4056 in the commit history.

### UI test report

image

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/ReleaseStatusCreateRequest.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/ReleaseStatusCreateRequest.cs
@@ -47,4 +47,6 @@ public record ReleaseStatusCreateRequest
     }
 
     [PartialDateValidator] public PartialDate? NextReleaseDate { get; init; }
+
+    public bool? UpdatePublishedDate { get; init; }
 }
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/ReleaseRepository.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/ReleaseRepository.cs
@@ -18,6 +18,39 @@ public class ReleaseRepository : IReleaseRepository
         _contentDbContext = contentDbContext;
     }
 
+    public async Task<DateTime> GetPublishedDate(Guid releaseId,
+        DateTime actualPublishedDate)
+    {
+        var release = await _contentDbContext.Releases
+            .SingleAsync(r => r.Id == releaseId);
+
+        // If the release is already published return the published date
+        if (release.Published.HasValue)
+        {
+            return release.Published.Value;
+        }
+
+        // For the first version of a release or if an update to the published date has been requested
+        // return the actual published date
+        if (release.Version == 0 || release.UpdatePublishedDate)
+        {
+            return actualPublishedDate;
+        }
+
+        // Otherwise, return the published date from the previous version
+        await _contentDbContext.Entry(release)
+            .Reference(r => r.PreviousVersion)
+            .LoadAsync();
+
+        if (!release.PreviousVersion!.Published.HasValue)
+        {
+            throw new ArgumentException(
+                $"Expected Release {release.PreviousVersionId} to have a Published date as the previous version of Release {release.Id}");
+        }
+
+        return release.PreviousVersion.Published.Value;
+    }
+
     public async Task<bool> IsLatestPublishedVersionOfRelease(Guid releaseId)
     {
         var release = await _contentDbContext.Releases
--- a/src/GovUk.Education.ExploreEducationStatistics.Publisher/Functions/PublishImmediateReleaseContentFunction.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Publisher/Functions/PublishImmediateReleaseContentFunction.cs
@@ -39,7 +39,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Publisher.Functions
         /// <param name="executionContext"></param>
         /// <param name="logger"></param>
         /// <returns></returns>
-        [FunctionName("PublishReleaseContent")]
+        [FunctionName("PublishImmediateReleaseContent")]
         // ReSharper disable once UnusedMember.Global
         public async Task PublishImmediateReleaseContent(
             [QueueTrigger(PublishReleaseContentQueue)]
--- a/src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/PublishingCompletionService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/PublishingCompletionService.cs
@@ -17,6 +17,7 @@ public class PublishingCompletionService : IPublishingCompletionService
 {
     private readonly ContentDbContext _contentDbContext;
     private readonly IContentService _contentService;
+    private readonly IMethodologyService _methodologyService;
     private readonly INotificationsService _notificationsService;
     private readonly IReleasePublishingStatusService _releasePublishingStatusService;
     private readonly IPublicationRepository _publicationRepository;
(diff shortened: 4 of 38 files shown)


src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/PublishingCompletionService.cs
```diff
@@ -88,20 +91,27 @@ await releaseIdsToUpdate
...
             .ToAsyncEnumerable()
             .ForEachAwaitAsync(_publicationRepository.UpdateLatestPublishedRelease);
-        
+
+        // Set the published date on any methodologies used by these publications that are now publicly accessible
+        // as a result of releases being published
+        await directlyRelatedPublicationIds
+            .ToAsyncEnumerable()
+            .ForEachAwaitAsync(publicationId =>
+                _methodologyService.SetPublishedDatesByPublication(publicationId));
```
markhiveit, line 104: I assume not, but we don't want to update existing associated methodlogies' published dates as well if an `UpdatePublishedDate` is true?

tests/robot-tests/tests/libs/admin_api.py
```diff
@@ -206,6 +206,13 @@ def user_creates_test_release_via_api(
     return response.json()["id"]
 
 
+def user_updates_release_published_date(release_id: str, published: datetime) -> None:
```
markhiveit, line 209: I think we should add a `_via_api` on to this. I think it is worth making it clear what is done through selenium and what is not.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/ReleaseRepository.cs
```diff
@@ -18,6 +18,39 @@ public ReleaseRepository(ContentDbContext contentDbContext)
...
+        DateTime actualPublishedDate)
+    {
+        var release = await _contentDbContext.Releases
+            .SingleAsync(r => r.Id == releaseId);
+
+        // If the release is already published return the published date
+        if (release.Published.HasValue)
+        {
+            return release.Published.Value;
+        }
```
markhiveit, line 31: I wonder if the overhead of making this release fetch async is worthwhile.

I wonder what's going on under the hood here. I think it'll create another process to get the release, and then immediately pass that information back to the initial thread? My intuition is that that is more costly than than just doing it all on the same thread.

Not saying change this - just thinking out loud :)

src/explore-education-statistics-admin/src/pages/release/components/ReleaseStatusForm.tsx
```diff
@@ -87,7 +88,7 @@ interface Props {
   release: Release;
   statusPermissions: ReleaseStatusPermissions;
   onCancel: () => void;
-  onSubmit: (values: ReleaseStatusFormValues) => void;
+  onSubmit: (values: ReleaseStatusFormValues) => Promise<void> | void;
```
markhiveit, line 91: How come this has changed? I can't figure out why this is necessary.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Functions/PublishImmediateReleaseContentFunction.cs
```diff
@@ -39,7 +39,7 @@ public class PublishImmediateReleaseContentFunction
         /// <param name="executionContext"></param>
         /// <param name="logger"></param>
         /// <returns></returns>
-        [FunctionName("PublishReleaseContent")]
+        [FunctionName("PublishImmediateReleaseContent")]
```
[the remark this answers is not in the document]
benoutram: I've just checked and it looks like every time the publisher functions project is deployed it automatically cleans up any unused functions left behind in the old deployment.

We've also renamed functions fairly recently and not seen any problems. This is one of the renames that got missed.

I spotted it here because as I was developing I ran into some exceptions which mentioned this name in the logs and spent quite a long time trying to track down the error before I realised the function name in the logs was wrong.

src/explore-education-statistics-admin/src/pages/release/components/ReleaseStatusForm.tsx
benoutram: It's not very significant, it's something @ntsim suggested to remove the typescript warning `'await' has no effect on the type of this expression.` within the `useFormSubmit` on L117.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/PublishingCompletionService.cs
benoutram: I assumed that updating release published dates is independent of methodologies and that methodologies continue to get a published date set if they are being published for the first time (either because the publication has it's first release accessible or a release that the methodology depends on is now accessible).

We confirmed at standup on 09/02/23 with @cjrace that this is correct.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/ReleaseRepository.cs
benoutram: I think it's probably a tiny difference especially as it should be called so infrequently, either when publishing a release or lazily caching a release. Maybe we should discuss this at the dev's team meeting to get a better understanding of when to use async and the benefits of it.

I had thought that anything that's potentially going to be waiting for the database is better to use async in case there's any delay it, allowing the calling method to continue execution in the mean time.

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/ReleaseStatusCreateRequest.cs
```diff
@@ -47,4 +47,6 @@ public record ReleaseStatusCreateRequest
     }
 
     [PartialDateValidator] public PartialDate? NextReleaseDate { get; init; }
+
+    public bool? UpdatePublishedDate { get; init; }
```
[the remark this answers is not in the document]
benoutram: I made it optional because the `UpdatePublishedDate` field is conditionally rendered depending on approval status and whether it's an amendment or not.

When moving between Draft -> Higher Review or from any status back to Draft, or when the release is not an amendment, it's not expected in the request.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/PublishingCompletionService.cs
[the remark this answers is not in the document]
benoutram: I've changed it to `SetPublishedDatesIfApplicable` now, thanks.

I thought about other names but decided to go with `IfApplicable` to leave that down to the reader find out what makes it applicable or not, otherwise the name is going to be super long.

ntsim (approved):

PR #3878, opened 2023-02-06
EES-4080 Add `IReleaseFileBlobService` and private/public implementations

ntsim:
This PR adds an `IReleaseFileBlobService` that wraps `IBlobStorageService` to provide methods for working with blobs using a `ReleaseFile` entity.

This removes potential errors from incorrectly calling `IBlobStorageService` methods e.g. with the wrong blob container name or calling `Path` instead of `PublicPath`, etc.

Importantly, this is also necessary for EES-3777 to allow us to create a re-usable `SubjectCsvMetaService` that can work in both the data API and the admin.

## Relevant changes

- Added cancellation token parameters to `DownloadBlobText` and `GetDeserializedJson` methods.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Api/Startup.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Api/Startup.cs
@@ -16,6 +16,8 @@ using GovUk.Education.ExploreEducationStatistics.Common.Utils;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Repository;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Repository.Interfaces;
+using GovUk.Education.ExploreEducationStatistics.Content.Model.Services;
+using GovUk.Education.ExploreEducationStatistics.Content.Model.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Content.Security;
 using GovUk.Education.ExploreEducationStatistics.Content.Security.AuthorizationHandlers;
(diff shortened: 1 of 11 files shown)


benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Startup.cs
```diff
@@ -143,6 +145,7 @@ public void ConfigureServices(IServiceCollection services)
                         new BlobServiceClient(publicStorageConnectionString),
                         provider.GetRequiredService<ILogger<BlobStorageService>>(),
                         new StorageInstanceCreationUtil()));
+            services.AddTransient<IReleaseFileBlobService, PublicReleaseFileBlobService>();
```
benoutram, line 148: Public/private implementations have been configured as services in Admin and Data.Api but look like they are not actually in use yet. Should it be configured in `Content.Api` for consistency?

benoutram (commented):
Just wondering how widely you plan to swap out `IBlobStorageService` for `IReleaseFileBlobService`, e.g. will that happen throughout both the `ReleaseFileService`'s and `ReleaseFileDataService`?

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Startup.cs
ntsim: Good point, I've added it to the `Content.Api` startup as well!

ntsim:
Discussed on Slack with @benoutram - would be good to refactor various services (i.e. `ReleaseFileServices`) to use the new `IReleaseFileBlobService` implementations, but we'll leave that for the future.

However, new work should try and take advantage of it if possible.

PR #3882, opened 2023-02-08
EES-4082 update typescript

amyb-hiveit:
Updates Typescript to 4.9.5. Also, updates @typescript-eslint.

There were a few tsc errors following the update so I've attempted to fix them, let me know if there are better ways to fix these.


src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartLegendConfiguration.tsx
```diff
@@ -168,7 +168,7 @@ const ChartLegendConfiguration = ({
       position: true,
       items: initialValues.items.map(item =>
         mapValues(item, () => true),
-      ) as FormikTouched<LegendItem>[],
+      ) as FormikTouched<LegendItem | unknown>[],
```
ntsim, line 171: Why unknown here?

package.json
```diff
@@ -26,7 +26,7 @@
     "stylelint-config-prettier": "^8.0.2",
     "stylelint-config-sass-guidelines": "^8.0.0",
     "stylelint-prettier": "^1.1.2",
-    "typescript": "^4.4.3"
+    "typescript": "^4.9.5"
```
ntsim, line 29: The `package-lock.json` hasn't been updated (it's probably why the build is failing). An `npm i` should fix it.

src/explore-education-statistics-common/src/hocs/withLazyLoad.tsx
```diff
@@ -6,7 +6,8 @@ interface LazyLoadProps<P> extends OmitStrict<BaseLazyLoadProps, 'children'> {
   placeholder?: ((props: P) => ReactNode) | ReactNode;
 }
 
-export default function withLazyLoad<P>(
+// eslint-disable-next-line @typescript-eslint/ban-types
```
ntsim, line 9: Doesn't seem like we need this. Removed it and there's no lint!

Unfortunately, there's a couple of other lints that do need fixing that are probably breaking the build

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartLegendConfiguration.tsx
amyb-hiveit: It's an attempt to fix the following error, any ideas on a better way?

```error TS2352: Conversion of type '{ dataSet: boolean; label: boolean; colour: boolean; symbol?: boolean | undefined; lineStyle?: boolean | undefined; inlinePosition?: boolean | undefined; }[]' to type 'FormikTouched<LegendItem>[]' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
src/pages/release/datablocks/components/chart/ChartLegendConfiguration.tsx(169,14): error TS2352: Conversion of type '{ dataSet: boolean; label: boolean; colour: boolean; symbol?: boolean | undefined; lineStyle?: boolean | undefined; inlinePosition?: boolean | undefined; }[]' to type 'FormikTouched<LegendItem>[]' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type '{ dataSet: boolean; label: boolean; colour: boolean; symbol?: boolean | undefined; lineStyle?: boolean | undefined; inlinePosition?: boolean | undefined; }' is not comparable to type 'FormikTouched<LegendItem>'.
    Types of property 'dataSet' are incompatible.
      Type 'boolean' is not comparable to type 'FormikTouched<DataSet>'.

Comment (amyb-hiveit) [reply src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartLegendConfiguration.tsx:171]: That gives me the same error.

Comment (ntsim) [reply src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartLegendConfiguration.tsx:171]: Ohhh, it's because `mapValues` will only work on the direct properties of `LegendItem`. However, `LegendItem` also has a `dataSet` property, so to get a correct `FormikTouched` (which can be a deeply nested object with boolean values for each property), we'd also need to map over this as well.

One way to fix this could be the following:

```ts
return {
  position: true,
  items: initialValues.items.map(item => {
    return {
      ...mapValues(item, () => true),
      dataSet: mapValues(item.dataSet, () => true),
    };
  }),
};
```

However, we might be able to just omit the `dataSet` property completely i.e.

```
return {
  position: true,
  items: initialValues.items.map(item =>
    mapValues(omit(item, 'dataSet'), () => true),
  ),
};
```

I'm not sure if there are runtime issues with either approach however, so we'll probably want to test what actually happens if we do either.

Another approach would be the following:

```ts
return {
  position: true,
  items: initialValues.items.map(
    item =>
      mapValues(item, () => true) as FormikTouched<
        Omit<LegendItem, 'dataSet'>
      >,
  ),
};
```

However, I'm not sure if it passes compilation (I'm not on the branch!)

amyb-hiveit: Omiting dataSet seems to wot work fine so I've gone with that. Thanks for your help!

src/explore-education-statistics-common/src/components/Effect.tsx
```diff
@@ -7,7 +7,7 @@ interface Props<T> {
   onChange?: (value: T, previousValue: T) => void;
 }
 
-const Effect = <T extends unknown>({ onChange, onMount, value }: Props<T>) => {
+const Effect = <T,>({ onChange, onMount, value }: Props<T>) => {
```
ntsim, line 10: This is a bit weird, would suggest we change this to `export default function Effect<T>()`

amyb-hiveit: Done

ntsim (approved):

PR #3883, opened 2023-02-09
EES-4091 Change `ITimePeriodService` to use async

benoutram:
This PR changes `ITimePeriodService` to use async where possible.

It was a change made while investigating [EES-3945](https://dfedigital.atlassian.net/browse/EES-3945), looking at why Step 1 of the table builder can be slow, which I thought is worth raising as it's own PR.

There's some changes to make the code a little more concise:

`GetTimePeriodLabels(Guid subjectId)` is able to reuse `GetTimePeriods(subjectId)`.
`GetTimePeriods(Guid subjectId)` is able to reuse `GetDistinctObservationTimePeriods(IQueryable<Observation> observationsQuery)`.

All the methods which previously had ordering are refactored to use a common `OrderTimePeriods` method.

It also adds a little warning about the reason we order time periods in memory rather than in a database query.

I performance tested this change with caching _disabled_. There might be a very slight performance benefit but I used random subject id's and location id's (when filtering) so some slight variance was expected anyway.

 There wasn't any degradation in performance and in any case, these methods are lazily cached.

Areas performance tested were:

- Getting subject meta data
- Filtering subject meta data
- Listing subjects (since time periods are returned in `SubjectViewModel`).

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Services/TimePeriodService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Services/TimePeriodService.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 using GovUk.Education.ExploreEducationStatistics.Common.Model;
 using GovUk.Education.ExploreEducationStatistics.Common.Services;
 using GovUk.Education.ExploreEducationStatistics.Data.Model;
@@ -21,27 +22,22 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Services
             _context = context;
         }
 
-        public IList<(int Year, TimeIdentifier TimeIdentifier)> GetTimePeriods(Guid subjectId)
+        public Task<IList<(int Year, TimeIdentifier TimeIdentifier)>> GetTimePeriods(Guid subjectId)
         {
-            return _context.Observation
-                .AsQueryable()
-                .Where(observation => observation.SubjectId == subjectId)
-                .Select(o => new {o.Year, o.TimeIdentifier})
-                .Distinct()
+            var observationsQuery = _context
+                .Observation
                 .AsNoTracking()
-                .ToList()
-                .OrderBy(tuple => tuple.Year)
-                .ThenBy(tuple => tuple.TimeIdentifier)
-                .Select(tuple => (tuple.Year, tuple.TimeIdentifier))
-                .ToList();
+                .Where(observation => observation.SubjectId == subjectId);
+
+            return GetDistinctObservationTimePeriods(observationsQuery);
         }
 
-        public IList<(int Year, TimeIdentifier TimeIdentifier)> GetTimePeriods(
-            IQueryable<Observation> observations)
+        public Task<IList<(int Year, TimeIdentifier TimeIdentifier)>> GetTimePeriods(
+            IQueryable<Observation> observationsQuery)
         {
-            return GetDistinctObservationTimePeriods(observations);
+            return GetDistinctObservationTimePeriods(observationsQuery);
         }
-        
+
         public IList<(int Year, TimeIdentifier TimeIdentifier)> GetTimePeriodRange(
             IList<Observation> observations)
         {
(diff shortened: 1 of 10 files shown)


src/GovUk.Education.ExploreEducationStatistics.Data.Services/TimePeriodService.cs
```diff
@@ -75,30 +66,35 @@ public TimePeriodLabels GetTimePeriodLabels(Guid subjectId)
...
         }
-        
+
         private static IList<(int Year, TimeIdentifier TimeIdentifier)> GetDistinctObservationTimePeriods(
             IList<Observation> observations)
         {
-            return observations
-                .Select(o => new {o.Year, o.TimeIdentifier})
+            var timePeriods = observations
+                .Select(o => new Tuple<int, TimeIdentifier>(o.Year, o.TimeIdentifier))
```
ntsim, line 84: Without an explicit `Tuple` type, we can define it like:

```cs
.Select(o => (o.Year, o.TimeIdentifier))
```

Same above as well

src/GovUk.Education.ExploreEducationStatistics.Data.Services/TimePeriodService.cs
```diff
@@ -75,30 +66,35 @@ public TimePeriodLabels GetTimePeriodLabels(Guid subjectId)
...
+                .ToList();
+
+            return OrderTimePeriods(timePeriods);
+        }
+
+        private static List<(int Year, TimeIdentifier TimeIdentifier)> OrderTimePeriods(
+            IList<Tuple<int, TimeIdentifier>> timePeriods)
+        {
+            return timePeriods
+                .Select(tuple => ((int Year, TimeIdentifier TimeIdentifier)) tuple.ToValueTuple())
```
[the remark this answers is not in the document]
benoutram: See comment below about why I used `Tuple` here and then converted to `ValueTuple` to name the arguments.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/TimePeriodService.cs
benoutram: I can't do that above because it's the `IQueryable` variant and LINQ has no expression-tree support for `ValueTuple`.

In the expression-tree above we can only use `Tuple` - `.Select(o => new Tuple<int, TimeIdentifier>(o.Year, o.TimeIdentifier))` or an anonymous type - `.Select(o => new { o.Year, o.TimeIdentifier })`.

If we went with an anonymous type we can't pass it as an argument and have a common `OrderTimePeriods` method.

That's why I ended up using `Tuple` for both variants and then wrote `OrderTimePeriods` accepting `IList<Tuple<int, TimeIdentifier>> timePeriods`. However as you've seen, it has a nasty conversion from `Tuple` to `ValueTuple` using `.Select(tuple => ((int Year, TimeIdentifier TimeIdentifier)) tuple.ToValueTuple())` to satisfy the return type and name the arguments.

I've got an alternative solution to incorporate your suggestions and in the `IQueryable` variant project the database result from anonymous type to `ValueTuple`:

```
  var timePeriods = (await observationsQuery
          .Select(o => new { o.Year, o.TimeIdentifier })
          .Distinct()
          .ToListAsync())
      .Select(x => (x.Year, x.TimeIdentifier))
      .ToList();
```

I'll add these changes as a separate commit and if you approve then I'll squash them. Thanks

ntsim: Of course, LINQ expressions strike again :nauseated_face: 

I like the new approach - nice to avoid the conversion!

src/GovUk.Education.ExploreEducationStatistics.Data.Services/TimePeriodService.cs
```diff
@@ -75,30 +66,38 @@ public TimePeriodLabels GetTimePeriodLabels(Guid subjectId)
...
         {
-            return observations
-                .Select(o => new {o.Year, o.TimeIdentifier})
+            var timePeriods = observations
+                .Select(o => (o.Year, o.TimeIdentifier))
                 .Distinct()
-                .ToList()
+                .ToList();
+
+            return OrderTimePeriods(timePeriods);
```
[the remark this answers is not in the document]
benoutram: Good point, I've made that change now. Thanks 🙂

ntsim (approved):

PR #3892, opened 2023-02-13
EES-3976, EES-4081 - migrate to PNPM

luke-h1:
See: https://github.com/dfe-analytical-services/explore-education-statistics/pull/3782 for original PR review

This PR migrates the project from Lerna + NPM to PNPM & PNPM workspaces. 

Useful links:
* https://pnpm.io/installation
* https://github.com/pnpm/pnpm

## Getting started with PNPM 

In order to set up the project & install dependencies for various packages, you would previously have to run two commands: `npm ci && npm run bootstrap`. With PNPM this command now simply turns into `pnpm i`. 
`pnpm i` will install the root dependencies along with all the `workspace` packages defined in `pnpm-workspace.yaml`

## Introduction

Due to the way PNPM symlinks dependencies and the stricter structure it provides out of the box (see https://pnpm.io/symlinked-node-modules-structure), we've had to enable the `shamefully-hoist` option in order to create a flat `node_module` structure.

In addition to this, we've excluded 3 packages from being hoisted to the root of the project. This is purely to ensure the frontend has `react`, `react-dom` and `formik` available in it's relative `node_modules` folder so that it can build properly. 

##  Differences between NPM and PNPM

Package managers such as Yarn or NPM use a flattened `node_modules` structure. The flattened dependency tree results in many issues such as:
 * The algorithm for flattening a dependency tree is complex
* Some packages have to be duplicated inside another project's `node_modules` folder 
* Modules have access to packages they do not depend on (also known as 'phantom' dependencies)

Let's go over these issues with an example. Let's say one project needs `express` as a dependency, and `express` has a dependency on a package named `debug`. This is how the structure of `node_modules` will look like with NPM: 

Screenshot 2023-03-20 at 10 46 27

At this stage, the code can import the debug package i.e. `require('debug')` without any issues whatsoever. We don't explicitly depend on this package via an entry in the `package.json` file. There are two big problems that would arise by doing this:

* Express updates its debug dependency with breaking changes
* Express does not depend on debug anymore 
* Your package now relies on 'phantom' packages making it harder to pinpoint what packages a project relies on

In both cases, our code will fail because it has an implicit dependency on debug. This is a big issue especially when we're working with large monorepos such as this one. It is much harder to track down particular dependencies that a given project uses. I've had to fix up a few issues around this area when migrating the project over to PNPM and it was pretty tedious + time-consuming.

### How does PNPM solve the above problems?
PNPM uses **hard links** and **symlinks** to maintain a **semistrict** `node_modules` structure. Let's take a look at the above example when using PNPM:

Screenshot 2023-03-20 at 10 52 28

The first difference you'll notice is that your code can no longer access the `debug` package because it is not directly under the root level `node_modules` directory. PNPM has created a special folder called `.pnpm` that containers hard links to all the modules.  If you look at `express@4.0.0`, the express module is a hard link to the global `pnpm-store` and a `debug` symbolic link to the `debug` hard link that also links to the global `pnpm-store`. 

Below is an image of the actual `pnpm-store`, which contains the packages. This is where all downloaded dependencies are maintained. When you download a dependency, PNPM first checks whether the dependency is present in this store or not. If it finds the dependency in the store, PNPM fetches it by creating a hard link.

Screenshot 2023-03-20 at 21 57 56

Because of this approach, PNPM re-uses the same packages if they are already installed for another project. 

#### Disk space efficiency 
PNPM uses a content-addressable file system to store the packages and dependencies on disk. This means that the same package will not be duplicated. Even with different versions of the same package, PNPM is intelligent enough to reuse the maximum amount of code. i.e. if package version `1.0.0` has 500 files and version `1.0.1` has just one more file, then PNPM will not write 501 files for version `1.0.1`. Instead, it will create a hard link to the original 500 files and write just the new file. If you compare this with NPM, then version 1.0.0 will also be loaded, thus duplicating the original 500 files. For large monorepo projects, this can make a big difference (and make your `node_modules` folder GBs big 😅). 

#### Improved speed 

PNPM outperforms its peers several times over in most cases. Here's a screenshot taken from PNPM's website  

alotta-files

#### Security 

PNPM like Yarn, has a special file with the checksum of all the installed packages. This ensures that the integrity of all the installed packages is verified before their code is executed. 

In terms of unprivileged access, PNPM also outperforms both NPM & Yarn. In the case of NPM & Yarn, if package `A` depends on package `B`, and `B` depends on `C`, then `A` implicitly gets access to `C` even though `A` has not declared `C` as its dependency. This problem as you can imagine is intensified in a large monorepo setup. PNPM on the other hand, uses a different dependency resolution algorithm and different folder structure of `node_modules` that prevents illegal access to packages 

## PNPM CLI differences

npm command | Yarn command | pnpm equivalent
-- | -- | --
npm install | yarn | pnpm install
npm install [pkg] | yarn add [pkg] | pnpm add [pkg]
npm uninstall [pkg] | yarn remove [pkg] | pnpm remove [pkg]
npm update | yarn upgrade | pnpm update
npm list | yarn list | pnpm list
npm run [scriptName] | yarn [scriptName] | pnpm [scriptName]
npx [command] | yarn dlx [command] | pnpm dlx [command]
npm exec | yarn exec [commandName] | pnpm exec [commandName]
npm init [initializer] | yarn create [initializer] | pnpm create [initializer]

## CI speed improvements 
PNPM is supposedly 2 - 3x times faster than the alternatives (Yarn, NPM, Lerna, etc.). As can be seen in the screenshot below of the build pipeline (left = PNPM + PNPM workspaces, right = current implementation of NPM + Lerna) we're now saving lots of time spent in CI

image (1)

| Step         | NPM + Lerna     | PNPM + PNPM workspaces | difference |
|---------- |-------------- |------------|-----------------|
| Admin            |  11m 40s    |  8m 17s      | -3min 33s ✅
| Frontend       |   22m 42s  |   15m 46s   | -6min 54s ✅

Overall build time difference:
-6min 55seconds ✅

NPM build + PNPM build side by side for future reference (PNPM left, NPM right): 
image (2)

## Infrastructure changes

Due to standard Azure app services not supporting PNPM, we've had to build & deploy the public frontend via Azure's hybrid app-service model. This entails building and pushing a docker image to a container registry. The app service will then boot up the docker container to serve the project. Because of this, we've made some amends to how the public frontend is deployed:

* Created a 'Standard' tier Azure container registry to house docker images
* Added a step in the build pipeline to build + push new docker images on merges
* Adds three new environment variables to the frontend app service related to docker changes

## Relevant changes:

**Updated the build pipeline to Install and use PNPM**
* Since Azure seems to not have a `PNPM` task, I've resorted to just using bash to run relevant PNPM commands.  
* Moved commonly used version specs (for things like Dotnet, Node, etc.) to variables for better organization
* Added appropriate spacing so it's easier to read

**Updated relevant READMEs** 
* Updated public frontend README and removed old information
* Updated main README due to changes in commands 

**General cleanup** 
* Removed check-node-version package as PNPM natively protects against wrong node/PNPM versions being used with the engines field in package.json and engine-strict in .npmrc
* Fixed type errors in `FilterGroupDetails.tsx` and `IndicatorDetails.tsx` that seem to have cropped up
* Fixes flaky (at least for myself locally) `Modal.test.tsx` unit test
* Fixes eslint warning in `MapBlockInternal.tsx`
* Upgrades `postcss-preset-env` to v7+ to avoid a spammy warning in the admin webpack logs: 

**Upgraded admin babel packages** 
* Upgraded admin babel packages to stop a problem occurring in CK editor whereby the editor simply wouldn't work and would crash whenever you navigated to parts of the admin app that rendered an editor. As part of this, amends needed to be made to the Admin's babel config. This was due to log spam in unit tests and when the admin initially started up. See https://stackoverflow.com/questions/67770240/webpacker-babel-spam-removal-message-not-clear-enough for an example of the warnings produced


src/explore-education-statistics-common/src/components/form/FormTextArea.tsx
```diff
@@ -17,6 +17,7 @@ export interface FormTextAreaProps extends FormLabelProps {
   error?: ReactNode | string;
   hint?: string;
   id: string;
+  // eslint-disable-next-line react/no-unused-prop-types
```
ntsim, line 20: Doesn't look like we're using this prop so we can just get rid of it

src/explore-education-statistics-common/src/modules/charts/components/LineChartBlock.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/display-name */
```
ntsim, line 1: Would disable on the specific location rather than on the file level

src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
```diff
@@ -68,6 +69,7 @@ export interface MapBlockProps extends ChartProps {
   id: string;
   legend: LegendConfiguration;
   map?: MapConfig;
+  // eslint-disable-next-line react/no-unused-prop-types
```
ntsim, line 72: This isn't being used so we can just get rid of it

src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
```diff
@@ -59,6 +59,7 @@ export interface MapBlockProps extends ChartProps {
   axes: {
     major: AxisConfiguration;
   };
+  // eslint-disable-next-line react/no-unused-prop-types
```
ntsim, line 62: Don't think we actually need this prop, however, we'll need to adjust `ChartRenderer` props i.e.

```ts
export type ChartRendererProps = {
  source?: string;
} & (
  | ({
      type: 'line';
    } & LineChartProps)
  | ({
      type: 'verticalbar';
    } & VerticalBarProps)
  | ({
      type: 'horizontalbar';
    } & HorizontalBarProps)
  | ({
      type: 'map';
    } & Omit<MapBlockProps, 'id'> & {
        boundaryLevel?: number;
      })
  | ({
      type: 'infographic';
    } & InfographicChartProps)
);
```

src/explore-education-statistics-frontend/src/components/PageSearchFormWithAnalytics.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/destructuring-assignment */
```
ntsim, line 1: Similar to other cases, we should just fix this instead of disable it

ntsim (changes requested):

README.md
```diff
@@ -342,30 +353,28 @@ Examples:
...
 
      ```bash
      # Admin frontend
-     npm run start:admin
+     pnpm start:admin
     
      # Public frontend
-     npm run start:frontend
+     pnpm start:frontend
      ```
```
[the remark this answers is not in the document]
luke-h1: Good point, I've gotten rid of the commands & just refactored it to mention running the frontend via the `run.js` alias

infrastructure/templates/template.json
```diff
@@ -718,6 +718,34 @@
...
+      "metadata": {
+        "description": "The password to log in to the Docker registry."
+      }
+    },
+    "DOCKER_IMAGE_NAME": {
+      "type": "string",
+      "defaultValue": "public-frontend",
+      "metadata": {
+        "description": "The name of the Docker image."
+      }
```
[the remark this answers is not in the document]
luke-h1: Just double checked & I don't think this variable is needed anymore

package.json
```diff
@@ -1,6 +1,15 @@
 {
   "name": "explore-education-statistics",
   "private": true,
+  "packageManager": "pnpm@8.5.1",
+  "dependencies": {
+    "husky": "^7.0.4",
+    "lint-staged": "^10.4.0"
+  },
```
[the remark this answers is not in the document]
luke-h1: The main reason to keep these in `dependencies` is due to the fact that there is a `prepare` script in the package.json. So when we try to run `pnpm --filter=explore-education-statistics-frontend... --prod install`, the command will fail due to husky not being installed

luke-h1:
Waiting on https://dfedigital.atlassian.net/browse/EES-4311 before merging this

README.md
```diff
@@ -300,74 +315,37 @@ If this is not available to you then you will need to use one, or a combination,
...
-     ```
- 
-    - Going into each of the sub-project directories and starting it directly e.g.
-    
-     ```bash
-     cd src/explore-education-statistics-frontend
-     npm start
-     ```
-
+     pnpm start frontend
```
ntsim, line 347: These lines aren't aligned properly so the docblock might not format properly

package.json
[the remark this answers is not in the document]
ntsim: Think we might be able to fix this by appending the `--ignore-scripts` flag to the install command. If that works, we should be able to move them back to `devDependencies`.

If not, would suggest we at least move the `dependencies` block next to the `devDependencies` block, instead of having them separated by `engines`.

src/explore-education-statistics-common/src/components/form/FormFieldCheckboxSearchSubGroups.tsx
```diff
@@ -16,6 +16,7 @@ export type FormFieldCheckboxSearchSubGroupsProps<FormValues> = OmitStrict<
 function FormFieldCheckboxSearchSubGroups<FormValues>(
   props: FormFieldCheckboxSearchSubGroupsProps<FormValues>,
 ) {
+  const { onSubGroupAllChange, onChange } = props;
```
ntsim, line 19: Slight nitpick, but would like to have a line break before any `return` to make the formatting a little nicer so that things aren't crowded.

Would do this for the rest of the files in this PR where we've introduced similar changes.

src/explore-education-statistics-common/src/components/form/FormFieldCheckboxGroup.tsx
```diff
@@ -22,8 +22,8 @@ function FormFieldCheckboxGroup<FormValues>(props: Props<FormValues>) {
           {...field}
           id={id}
           onAllChange={(event, checked) => {
-            if (props.onAllChange) {
-              props.onAllChange(event, checked, options);
+            if (onAllChange) {
+              onAllChange(event, checked, options);
```
ntsim, line 26: Don't have to do this, but we could remove the if block and replace it with:

```ts
onAllChange?.(event, checked, options);
```

A similar thing could be done across the rest of this PR as well.

PR #3897, opened 2023-02-16
EES-2469 Add backend endpoints to support free text key stats

markhiveit:
This PR adds to backend endpoint that the frontend will use to create and update free text key stats.

Deleting and reordering key stats is covered by a generic method for all key stat types.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/KeyStatisticServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/KeyStatisticServiceTests.cs
@@ -347,6 +347,164 @@ public class KeyStatisticServiceTests
     }
 
     [Fact]
+    public async Task CreateKeyStatisticText()
+    {
+        var release = new Release();
+
+        var contextId = Guid.NewGuid().ToString();
+        await using (var context = InMemoryContentDbContext(contextId))
+        {
+            await context.Releases.AddRangeAsync(release);
+            await context.SaveChangesAsync();
+        }
+
+        await using (var context = InMemoryContentDbContext(contextId))
+        {
+            var keyStatisticService = SetupKeyStatisticService(context);
+
+            var result = await keyStatisticService.CreateKeyStatisticText(
+                release.Id,
+                new KeyStatisticTextCreateRequest
+                {
+                    Title = "title",
+                    Statistic = "Over 9000!",
+                    Trend = "trend",
+                    GuidanceTitle = "guidanceTitle",
+                    GuidanceText = "guidanceText",
+                });
+
+            var viewModel = result.AssertRight();
+
+            Assert.Equal("title", viewModel.Title);
+            Assert.Equal("Over 9000!", viewModel.Statistic);
+            Assert.Equal("trend", viewModel.Trend);
+            Assert.Equal("guidanceTitle", viewModel.GuidanceTitle);
+            Assert.Equal("guidanceText", viewModel.GuidanceText);
+            Assert.Equal(0, viewModel.Order);
+            Assert.InRange(DateTime.UtcNow.Subtract(viewModel.Created).Milliseconds, 0, 1500);
+            Assert.Null(viewModel.Updated);
+        }
+
+        await using (var context = InMemoryContentDbContext(contextId))
+        {
+            var keyStatistics = context.KeyStatistics.ToList();
+            var keyStat = Assert.Single(keyStatistics);
+            var keyStatText = Assert.IsType<KeyStatisticText>(keyStat);
+
+            Assert.Equal("title", keyStatText.Title);
+            Assert.Equal("Over 9000!", keyStatText.Statistic);
+            Assert.Equal(release.Id, keyStatText.ReleaseId);
+            Assert.Equal("trend", keyStatText.Trend);
+            Assert.Equal("guidanceTitle", keyStatText.GuidanceTitle);
+            Assert.Equal("guidanceText", keyStatText.GuidanceText);
+            Assert.Equal(0, keyStatText.Order);
+            Assert.Equal(_userId, keyStatText.CreatedById);
+            Assert.Null(keyStatText.UpdatedById);
+            Assert.InRange(DateTime.UtcNow.Subtract(keyStatText.Created).Milliseconds, 0, 1500);
+            Assert.Null(keyStatText.Updated);
+        }
+    }
+
+    [Fact]
+    public async Task CreateKeyStatisticText_NoRelease()
+    {
+        var release = new Release();
+
+        var contextId = Guid.NewGuid().ToString();
+        await using (var context = InMemoryContentDbContext(contextId))
+        {
+            await context.Releases.AddRangeAsync(release);
+            await context.SaveChangesAsync();
+        }
+
+        await using (var context = InMemoryContentDbContext(contextId))
+        {
+            var keyStatisticService = SetupKeyStatisticService(context);
+            var result = await keyStatisticService.CreateKeyStatisticText(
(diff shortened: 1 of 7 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/KeyStatisticServiceTests.cs
```diff
@@ -607,6 +765,257 @@ await using (var context = InMemoryContentDbContext(contextId))
...
+            Assert.Equal(0, viewModel.Order);
+            Assert.Equal(new DateTime(2000, 1, 1), viewModel.Created);
+            Assert.NotNull(viewModel.Updated);
+            Assert.InRange(DateTime.UtcNow.Subtract(viewModel.Updated!.Value).Milliseconds, 0, 1500);
+        }
+
+        await using (var context = InMemoryContentDbContext(contextId))
+        {
+            var keyStatistics = context.KeyStatistics.ToList();
+            var keyStat = Assert.Single(keyStatistics);
```
luke-h1, line 834: Do we need to assign `keyStat` to a variable here?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/KeyStatisticServiceTests.cs
```diff
@@ -346,6 +346,164 @@ await using (var context = InMemoryContentDbContext(contextId))
...
+                new KeyStatisticTextCreateRequest
+                {
+                    Title = "title",
+                    Statistic = "Over 9000!",
+                });
+
+            var viewModel = result.AssertRight();
+
+
+            Assert.Equal("title", viewModel.Title);
```
duncan-at-hiveit, line 470: Probably don't need these other non-Order checks in this particular test method, for brevity.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/KeyStatisticServiceTests.cs
```diff
@@ -346,6 +346,164 @@ await using (var context = InMemoryContentDbContext(contextId))
...
+                .ToList();
+            Assert.Equal(4, keyStatistics.Count);
+
+            Assert.Equal(0, keyStatistics[0].Order);
+            Assert.Equal(1, keyStatistics[1].Order);
+            Assert.Equal(2, keyStatistics[2].Order);
+
+            var keyStatText = Assert.IsType<KeyStatisticText>(keyStatistics[3]);
+
+            Assert.Equal("title", keyStatText.Title);
```
duncan-at-hiveit, line 493: As above- probably only need to be checking its Order here (as well as checking it's the right KeyStat!).

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/KeyStatisticServiceTests.cs
```diff
@@ -607,6 +765,257 @@ await using (var context = InMemoryContentDbContext(contextId))
...
+                new KeyStatisticTextUpdateRequest
+                {
+                    Title = "new title",
+                    Statistic = "new statistic",
+                    Trend = "new trend",
+                    GuidanceTitle = "new guidanceTitle",
+                    GuidanceText = "new guidanceText",
+                });
+
+            result.AssertNotFound();
```
duncan-at-hiveit, line 1014: I was almost expecting an `AssertBadRequest` to be here, but I'm not too fussed either way - it does indeed show that the KeyStatText with that ID does not exist!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/KeyStatisticService.cs
```diff
@@ -82,6 +82,31 @@ await _dataBlockService.IsUnattachedDataBlock(releaseId, dataBlock)
...
+            .OnSuccess(async _ =>
+            {
+                var keyStatisticText = _mapper.Map<KeyStatisticText>(request);
+                keyStatisticText.ReleaseId = releaseId;
+                keyStatisticText.CreatedById = _userService.GetUserId();
+
+                var order = await _context.KeyStatistics
+                    .Where(ks => ks.ReleaseId == releaseId)
+                    .Select(ks => ks.Order)
+                    .MaxAsync(order => (int?)order);
```
duncan-at-hiveit, line 100: What's the need for the cast here?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/KeyStatisticService.cs
```diff
@@ -82,6 +82,31 @@ await _dataBlockService.IsUnattachedDataBlock(releaseId, dataBlock)
...
+                keyStatisticText.Order = order.HasValue ? order.Value + 1 : 0;
```
duncan-at-hiveit, line 101: I'd be tempted to write:

```
keyStatisticText.Order = (await _context
    .KeyStatistics
    .Where(ks => ks.ReleaseId == releaseId)
    .Select(ks => ks.Order)
    .MaxAsync(order => (int?)order)) ?? 0;
```

for brevity.

duncan-at-hiveit: (not much difference to be fair!)

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/KeyStatisticServiceTests.cs
markhiveit: I think this is super minor, to the point where it really doesn't matter.

I think I prefer the way it is: if I remove this variable we'll have this code instead:
```
            Assert.Single(keyStatistics);
            var keyStatText = Assert.IsType<KeyStatisticText>(keyStatistics[0]);
```
I'd prefer to avoid the `keyStatistics[0]`. I also like having the variable nams: `keyStatText` hints that the type has been checked, where `keyStat` implies it hasn't.

I can see that you might want to avoid having a `keyStat` and `keyStatText` variable. But, as I say, this change seems  practically inconsequential to me.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/KeyStatisticServiceTests.cs
markhiveit: I've left it

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/KeyStatisticService.cs
markhiveit: This is a hack that allows us to use Max even if it is possible for it to be passed no elements. Without this, if there are no key stats, it throws an `InvalidOperationException`. With this, `var order` is now nullable and so it's fine.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/KeyStatisticService.cs
markhiveit: I don't think this would be right, as we still need to add the `+ 1` on to the existing max value, if it has one.

duncan-at-hiveit: No worries - I didn't consider the `+1`

duncan-at-hiveit (approved):

PR #3901, opened 2023-02-21
EES-3786 React Hook Form on Filters Form

amyb-hiveit:
Replaces Formik with React Hook Form (RHF) on the FiltersForm step of the Table Tool to improve the performance of large checkbox lists.

Following feedback from @ntsim  on the previous version (https://github.com/dfe-analytical-services/explore-education-statistics/pull/3791), this PR:
- adds a FormProvider to allow child components to access the form context without having to pass it down
- adds a useRegister hook to memoise registering fields
- changes FormContext to FormIdContext to make it clearer what it's for
- creates RHF versions of the following components in a forms/rhf folder:
  - Form
  - FormFieldCheckboxGroupsMenu
  - FormFieldCheckboxSearchSubGroups
  - FormCheckboxSelectedCount
  - CreateErrorHelper

## Profiling results

Profiling for selecting and unselecting all options for a large filter group (208 options).

RHF is more performant than Formik as it isolates input components so doesn't re-render the entire form unnecessarily.

### Formik
formik-profile

### RHF
flame-rhf


src/explore-education-statistics-common/src/components/form/rhf/RHFFormFieldCheckboxGroupsMenu.tsx
```diff
@@ -0,0 +1,38 @@
...
+) {
+  const { name, legend, open = false } = props;
+
+  const {
+    formState: { errors },
+  } = useFormContext();
+
+  return (
+    <DetailsMenu
+      open={open}
```
luke-h1, line 27: Just wondering if we should expose `hiddenText` of the `DetailsMenu` for accessibility?

src/explore-education-statistics-common/src/components/form/rhf/validation/RHFCreateErrorHelper.ts
```diff
@@ -0,0 +1,70 @@
...
+      };
+    }, {});
+  };
+
+  const getError = (name: keyof T | string): string => {
+    const isTouched = get(touched, name, false);
+
+    if (!isTouched) {
+      return '';
+    }
```
ntsim, line 52: Given we have a new flag for including errors for touched/untouched errors, would have expected it to also affect the `getError` function too.

As my above comment, we could add a new parameter to this function to toggle this behaviour:

```ts
getError('something'); // `includeUntounched` defaults to false
getError('something', { includeUntouched: true });
```

src/explore-education-statistics-common/src/components/form/rhf/__tests__/RHFForm.test.tsx
```diff
@@ -0,0 +1,491 @@
...
+          firstName: '',
+          lastName: '',
+        }}
+        validationSchema={Yup.object({
+          firstName: Yup.string().required('First name is required'),
+          lastName: Yup.string().required('Last name is required'),
+        })}
+      >
+        {() => {
+          return (
```
ntsim, line 23: As other places - could spare some lines in this test suite by using the non-render prop variant for the children.

src/explore-education-statistics-common/src/components/form/rhf/__tests__/RHFForm.test.tsx
```diff
@@ -0,0 +1,491 @@
...
+            </RHFForm>
+          );
+        }}
+      </FormProvider>,
+    );
+
+    userEvent.click(screen.getByRole('button', { name: 'Submit' }));
+
+    await waitFor(() => {
+      expect(screen.queryByText('Something went wrong')).toBeInTheDocument();
```
ntsim, line 255: Could use `getByText`

src/explore-education-statistics-common/src/components/form/rhf/__tests__/RHFForm.test.tsx
```diff
@@ -0,0 +1,491 @@
...
+    });
+
+    fireEvent.click(screen.getByText('Reset form'));
```
ntsim, line 311: Can use `userEvent` here

src/explore-education-statistics-common/src/components/form/rhf/__tests__/RHFForm.test.tsx
```diff
@@ -0,0 +1,491 @@
...
+            </RHFForm>
+          );
+        }}
+      </FormProvider>,
+    );
+
+    userEvent.click(screen.getByRole('button', { name: 'Submit' }));
+
+    await waitFor(() => {
+      expect(screen.queryByText('Something went wrong')).toBeInTheDocument();
```
ntsim, line 356: Could use `getByText`

src/explore-education-statistics-common/src/components/form/rhf/__tests__/RHFForm.test.tsx
```diff
@@ -0,0 +1,491 @@
...
+    });
+
+    fireEvent.change(screen.getByLabelText('Firstname'), {
```
ntsim, line 359: Could use a `userEvent.type` to change this field? If we can, we have another `fireEvent.change` we could also get rid of further down (and we can get rid of `fireEvent` as an import).

src/explore-education-statistics-common/src/components/form/rhf/__tests__/RHFForm.test.tsx
```diff
@@ -0,0 +1,491 @@
...
+    const input = screen.getByLabelText('First name');
+
+    input.focus();
+
+    fireEvent.change(input, {
+      target: {
+        value: 'a first name',
+      },
+    });
+    // await userEvent.type(screen.getByLabelText('First name'), 'a first name');
```
ntsim, line 443: As above - is it possible to replace this with `userEvent.type`? If not, then would get rid of the comment.

src/explore-education-statistics-common/src/components/form/rhf/__tests__/RHFForm.test.tsx
```diff
@@ -0,0 +1,491 @@
...
+            </RHFForm>
+          );
+        }}
+      </FormProvider>,
+    );
+
+    userEvent.click(screen.getByRole('button', { name: 'Submit' }));
+
+    await waitFor(() => {
+      expect(screen.getByRole('alert')).toHaveFocus();
```
ntsim, line 399: Using `getByRole` in a `waitFor`. We could check on the error summary text instead, then check focus after the `waitFor`:

```ts
await waitFor(() => {
  expect(screen.getByText('There is a problem')).toBeInTheDocument();
});

expect(screen.getByRole('alert')).toHaveFocus();
```

src/explore-education-statistics-common/src/components/form/rhf/__tests__/RHFForm.test.tsx
```diff
@@ -0,0 +1,491 @@
...
+
+    fireEvent.change(input, {
+      target: {
+        value: 'a first name',
+      },
+    });
+    // await userEvent.type(screen.getByLabelText('First name'), 'a first name');
+
+    await waitFor(() => {
+      expect(screen.getByRole('alert')).not.toHaveFocus();
```
ntsim, line 446: As earlier, using `getByRole` in a `waitFor`. We can optimise this with a `getByText`

src/explore-education-statistics-common/src/components/form/rhf/__tests__/RHFForm.test.tsx
```diff
@@ -0,0 +1,491 @@
...
+
+    screen.getByLabelText('First name').focus();
+    expect(screen.getByRole('alert')).not.toHaveFocus();
+
+    userEvent.click(screen.getByRole('button', { name: 'Submit' }));
+
+    await waitFor(() => {
+      expect(screen.getByRole('alert')).toHaveFocus();
+    });
+  });
```
ntsim, line 490: As earlier, using `getByRole` in `waitFors`. We can optimise this with `getByText`

src/explore-education-statistics-common/src/components/form/rhf/__tests__/RHFFormFieldCheckboxSearchSubGroups.test.tsx
```diff
@@ -0,0 +1,971 @@
...
+describe('FormFieldCheckboxSearchSubGroups', () => {
+  test('renders with correct default ids without form', () => {
+    render(
+      <FormProvider
+        initialValues={{
+          test: [],
+        }}
+      >
+        {() => {
+          return (
```
ntsim, line 20: Similar to other test suites, could spare a few lines by not using the render prop variant to declare the children.

ntsim (changes requested):

src/explore-education-statistics-common/src/components/form/rhf/FormProvider.tsx
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * RHF caches default values, to update them you you have to
+   * reset the form with the new values.
+   */
+  useEffect(() => {
+    if (!isEqual(previousInitialValues.current, initialValues)) {
+      form.reset(initialValues as TFormValues);
+      previousInitialValues.current = initialValues;
+    }
+  }, [form, initialValues]);
```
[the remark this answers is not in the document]
amyb-hiveit: It's needed because of the possibility of going back to a previous step in the table tool and changing the options so the filters/indicators are different. In the Formik version I think this is handled by the `enableReinitialize` prop, RHF doesn't have an equivalent but one of it's developers has recommended this approach in various github discussions.

I'll update the comment to make it clearer and maybe add an optional prop to enable it or not.

src/explore-education-statistics-common/src/components/form/rhf/RHFFormFieldCheckboxGroupsMenu.tsx
```diff
@@ -0,0 +1,38 @@
...
+interface Props<TFormValues extends FieldValues>
+  extends RHFFormFieldCheckboxSearchSubGroupsProps<TFormValues> {
+  legend: string;
+  open?: boolean;
+}
+
+function RHFFormFieldCheckboxGroupsMenu<TFormValues extends FieldValues>(
+  props: Props<TFormValues>,
+) {
+  const { name, legend, open = false } = props;
```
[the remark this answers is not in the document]
amyb-hiveit: Not in this case as we're passing all of props to `RHFFormFieldCheckboxSearchSubGroups`

src/explore-education-statistics-common/src/components/form/rhf/RHFFormFieldCheckboxGroupsMenu.tsx
amyb-hiveit: Yeah, good idea.

src/explore-education-statistics-common/src/components/form/rhf/__tests__/RHFForm.test.tsx
amyb-hiveit: Yep and have realised that this test isn't correct - the alert has gone by the time it checks that it's not focused so will updat it here and in the original Form.test.tsx to make sure the alert is shown again but doesn't have focus.

src/explore-education-statistics-common/src/modules/table-tool/components/TableQueryError.tsx
```diff
@@ -66,32 +74,35 @@ const TableQueryError = ({
...
+              {!releaseId && ` (available when the release is published)`}
+            </Button>
+          </>
+        ) : (
+          <p>{nonDownloadOptionMessage}</p>
+        )}
+      </BaseErrorSummary>
+    );
+  }
+  return null;
```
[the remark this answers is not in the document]
amyb-hiveit: Not sure I'm following you here... when the values haven't changed (the options selected are still what you had when you submitted the form and got the error) then the error shows, when you change the options it's removed.

src/explore-education-statistics-common/src/components/form/rhf/RHFFormFieldCheckboxSearchSubGroups.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const {
+    formState: { errors, submitCount },
+    register,
+    setValue,
+    trigger,
+  } = useFormContext<TFormValues>();
+
+  const { ref: inputRef, ...field } = useRegister(name, register);
+  const { prefixFormId, fieldId } = useFormIdContext();
+  const id = customId ? prefixFormId(customId) : fieldId(name);
```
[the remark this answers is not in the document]
amyb-hiveit: Good idea, done and changed in the various places it's used.

src/explore-education-statistics-common/src/components/form/rhf/validation/RHFCreateErrorHelper.ts
```diff
@@ -0,0 +1,70 @@
+import get from 'lodash/get';
+import {
+  FieldErrors,
+  FieldNamesMarkedBoolean,
+  FieldValues,
+} from 'react-hook-form';
+
+const RHFCreateErrorHelper = <T extends FieldValues>({
```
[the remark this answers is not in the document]
amyb-hiveit: Done and tests added. There were a couple of tests to do with handling undefined values in the error array that I don't think will be needed for the RHF errors so have left those out and added tests for the new isSubmitted option.

src/explore-education-statistics-common/src/components/form/rhf/validation/RHFCreateErrorHelper.ts
```diff
@@ -0,0 +1,70 @@
...
+  touched,
+  errors,
+  showErrorsForUntouchedFields = false,
+}: {
+  touched: Partial<Readonly<FieldNamesMarkedBoolean<T>>>;
+  errors: FieldErrors<T>;
+  showErrorsForUntouchedFields?: boolean;
```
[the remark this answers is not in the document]
amyb-hiveit: It's needed because when a Formik form is submitted it marks all the fields as touched, but RHF form doesn't do this (and has no way to simply set all as touched). To replicate the current error handling we need to know if the form has been submitted and ignore whether fields have been touched or not if it has been, so I've changed this param to `isSubmitted` to make this clearer, added a comment, and use it in getError as well as getAllErrors.

src/explore-education-statistics-common/src/components/form/rhf/FormProvider.tsx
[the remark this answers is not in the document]
ntsim: Cool, yep that makes sense!

src/explore-education-statistics-common/src/components/form/rhf/validation/RHFCreateErrorHelper.ts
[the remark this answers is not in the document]
ntsim: Cool, `isSubmitted` makes a lot more sense!

src/explore-education-statistics-common/src/modules/table-tool/components/TableQueryError.tsx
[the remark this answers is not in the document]
ntsim: Ah, I get you now. I think because the condition used to wrap the component it confused me a little. Makes sense!

ntsim (approved):

PR #3914, opened 2023-02-24
EES-4139 - cleanup performance tests

luke-h1:
This PR: 
* performs a general cleanup in the performance tests. Changes include things such as unnecessary casting, fixing lint issues etc.

### DIFF
--- a/tests/performance-tests/src/tests/public/findStatisticsPage.test.ts
+++ b/tests/performance-tests/src/tests/public/findStatisticsPage.test.ts
@@ -34,7 +34,7 @@ export const getReleaseRequestDuration = new Trend(
 );
 
 const environmentAndUsers = getEnvironmentAndUsersFromFile(
-  __ENV.TEST_ENVIRONMENT as string,
+  __ENV.TEST_ENVIRONMENT,
 );
 
 export function setup() {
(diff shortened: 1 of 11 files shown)


tests/performance-tests/src/tests/public/findStatisticsPage.test.ts
```diff
@@ -55,15 +55,12 @@ const performTest = () => {
     getReleaseFailureCount.add(1);
     errorRate.add(1);
     fail(`Failure to get Find Statistics page - ${JSON.stringify(e)}`);
-    return;
```
duncan-at-hiveit, line 58: I think that we actually want to return early here and not proceed onto the following `check(response` statement, as that will flag up additional failures on top of the ones we've just registered.

If it's linting that's advising us away from having this return statement, can we put the rest of the code in an "else" instead?

luke-h1: There's no need to add a return statement here as it's unreachable. We're in the `catch` and `fail` will immediately throw an error, meaning anything below it will not get executed

duncan-at-hiveit (approved):

PR #3908, opened 2023-02-23
EES-4135 Change time identifier casing to sentence case

ntsim:
This PR changes time identifiers to use sentence casing i.e. 'Academic year' instead of 'Academic Year'. This is in line with the data files as they were originally uploaded and follows [DfE open data standards](https://dfe-analytical-services.github.io/stats-production-guidance-copy/ud.html).

PR #3918, opened 2023-02-27
EES-3777 Add CSV support to tablebuilder and permalink endpoints

ntsim:
> :warning: Dependency on EES-4135 (#3908)

This PR adds support for CSV formatted responses from the `/tablebuilder` and `/permalink` endpoints. As part of this, we've also completely remove the client-side generation of the table CSV and integrated the new CSV endpoints into the table tool frontend.

## Relevant changes

- Added a raft of functionality for working with location CSV columns. This allows us to get things like the location attribute columns e.g. `country_code`, `country_name`, `new_la_code`, etc.
- Added additional formatting for time period years in CSVs (using their time identifier) e.g. academic and fiscal years format to 202021. 
- Added extensions for working with different media types in controller responses. This includes methods for correctly applying the `Content-Disposition` header and checking a request's `Accept` header.

## Other changes

- Added `SnakeCase` string extension to work better with setting `name` defaults for indicators/filter columns. This avoid excessive double-quotes escaping due to spaces in the current default.
- Added new `Single/FirstOrNotFound` extensions to `IQueryable`. Some of the new code originally used these, but it was eventually refactored away. I've kept these in anyway as they're a nice alternative to using `PersistenceHelper`.
- Reduced the logging level of integration tests to warning to avoid excessively spamming the console whilst running.
- Fixed potential intermittent failures when caching setters for the same property in `Generator`. This would previously throw an exception that the same key is already in the cache.


src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/SubjectCsvMetaServiceTests.cs
```diff
@@ -0,0 +1,1528 @@
...
+                SubjectId = releaseSubject.SubjectId,
+                Indicators = indicators.Select(i => i.Id).ToList()
+            };
+
+            var result =
+                await service.GetSubjectCsvMeta(releaseSubject, query, observations);
+
+            var viewModel = result.AssertRight();
+
+            // No release file found. Fallback to a default ordering of headers.
```
duncan-at-hiveit, line 1347: Nice!

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectCsvMetaService.cs
```diff
@@ -0,0 +1,182 @@
...
+    private async Task<Stream?> GetCsvStream(
+        ReleaseSubject releaseSubject,
+        CancellationToken cancellationToken = default)
+    {
+        try
+        {
+            var releaseFile = await _contentDbContext.ReleaseFiles
+                .Include(rf => rf.File)
+                .SingleAsync(
+                    predicate: rf =>
```
duncan-at-hiveit, line 89: Are we needing to use named parameters here?  I think that in general we only use them when we have multiple parameters of the same type that could potentially get mixed up in error e.g. multiple Guid parameters?

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/StringExtensionTests.cs
```diff
@@ -364,6 +364,65 @@ public void UnderscoreSeparatedStringsArePascalCased()
...
+                Assert.Null(input!.SnakeCase());
+            }
+
+            [Fact]
+            public void Empty()
+            {
+                Assert.Equal(string.Empty, string.Empty.SnakeCase());
+            }
+
+            [Theory]
```
duncan-at-hiveit, line 382: Noice!

src/GovUk.Education.ExploreEducationStatistics.Data.Model.Tests/Fixtures/FilterGeneratorExtensions.cs
```diff
@@ -46,7 +47,8 @@ public static InstanceSetters<Filter> SetDefaults(this InstanceSetters<Filter> s
             .SetDefault(f => f.Id)
             .SetDefault(f => f.Label)
             .SetDefault(f => f.Hint)
-            .SetDefault(f => f.Name);
+            .SetDefault(f => f.Name)
+            .Set(f => f.Name, (_, f) => f.Name.SnakeCase());
```
duncan-at-hiveit, line 51: Noice

src/GovUk.Education.ExploreEducationStatistics.Common/Services/TimePeriodLabelFormatter.cs
```diff
@@ -64,7 +71,23 @@ private string FormatYearInternal(int year)
...
         }
 
-        private static TimePeriodLabelFormatter FormatterFor(TimeIdentifier timeIdentifier,
+        private string FormatCsvYearInternal(int year)
+        {
+            if (YearRegex.Match(year.ToString()).Success)
+            {
+                return YearFormat switch
+                {
+                    Academic => $"{year}{(year + 1) % 100:D2}",
```
duncan-at-hiveit, line 80: Can both these cases share the same return statement as they have the same implementation? Might make it easier if we have new year types added.

Also, didn't know about the D2 - nice!

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Controllers/TableBuilderControllerTests.cs
```diff
@@ -130,6 +148,40 @@ public async Task Query()
...
+                    )
+                )
+                .ReturnsAsync(Unit.Instance)
+                .Callback<ObservationQueryContext, Stream, CancellationToken>(
+                    (_, stream, _) => { stream.WriteText("Test csv"); }
+                );
+
+            var client = SetupApp(tableBuilderService: tableBuilderService.Object).CreateClient();
+
+            var response = await client
```
duncan-at-hiveit, line 171: Another nice integration test example is good to have here.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Controllers/TableBuilderController.cs
```diff
@@ -47,26 +49,55 @@ public class TableBuilderController : ControllerBase
...
         [CancellationTokenTimeout(TableBuilderQuery)]
-        public Task<ActionResult<TableBuilderResultViewModel>> Query(
+        public async Task Query(
             [FromBody] ObservationQueryContext query,
             CancellationToken cancellationToken = default)
         {
-            return _tableBuilderService
+            if (Request.AcceptsCsv(exact: true))
+            {
+                Response.ContentDispositionAttachment("text/csv");
```
duncan-at-hiveit, line 60: Would we want to add a suggested filename to this response header at all?  Not too sure what would be good to suggest if we did - maybe the Subject name for example?  Probably not a requirement and not something necessarily to do now but just raising the idea.

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/__snapshots__/TableBuilderServiceTests.QueryToCsvStream_LatestRelease.snap
```diff
@@ -0,0 +1,9 @@
+﻿time_period,time_identifier,geographic_level,country_code,country_name,region_code,region_name,new_la_code,old_la_code,la_name,name_of_filter_0,name_of_filter_1,name_of_indicator_0,name_of_indicator_1,name_of_indicator_2
```
duncan-at-hiveit, line 1: Nice!

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
```diff
@@ -63,8 +65,7 @@ public class PermalinkService : IPermalinkService
                     settings: BuildJsonSerializerSettings());
                 return await BuildViewModel(permalink!);
             }
-            catch (StorageException e)
-                when ((HttpStatusCode)e.RequestInformation.HttpStatusCode == HttpStatusCode.NotFound)
+            catch (FileNotFoundException)
```
duncan-at-hiveit, line 220: Good simplification here

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/EitherTaskExtensionsTests.cs
```diff
@@ -12,31 +12,64 @@ public class OrNotFound
...
 
         [Fact]
         public async Task ReferenceType_NotNull()
         {
-            var result = await Task.FromResult<Test?>(new Test()).OrNotFound();
+            var result = await new Test().OrNotFound();
 
-            var value = result.AssertRight();
-            Assert.Equal(new Test(), value);
+            Assert.Equal(new Test(), result.AssertRight());
```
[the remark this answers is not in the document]
ntsim: Done, and in all other places!

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/QueryableExtensionsTests.cs
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        [Fact]
+        public async Task Found_First()
+        {
+            var result = await ListOf("me", "me")
+                .AsQueryable()
+                .FirstOrNotFound();
+
+            Assert.Equal("me", result.AssertRight());
```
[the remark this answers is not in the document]
ntsim: Changed to use the `result.AssertRight(expected: "me")` extension. Note that we need to use named parameters here or the wrong overload (using the `message` parameter) gets used instead.

I've also just added another test case to cover the first item being returned for multiple items.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/LocationAttribute.cs
```diff
@@ -25,13 +26,48 @@ public virtual string GetCodeOrFallback()
     }
 
     public GeographicLevel GeographicLevel => Enum.Parse<GeographicLevel>(GetType().Name);
+
+    public abstract KeyValuePair<string, string>[] CsvValues { get; }
+
+    // The lower the number, the sooner it should appear in the CSV.
+    public abstract int CsvPriority { get; }
```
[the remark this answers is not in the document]
ntsim: Not too keen on this. The current implementation is using inheritance so that we can get feedback at compile time, ensuring that we're filling out any new implementations of `LocationAttribute` correctly.

Given how anaemic the domain model is anyway, it doesn't really seem like a problem to put the CSV functionality in here.

I think if we need to add more functionality to this in the future, we can reassess this, but I'd err on the side of KISS.

src/GovUk.Education.ExploreEducationStatistics.Data.Model.Tests/Fixtures/FilterGroupGeneratorExtensions.cs
```diff
@@ -17,6 +17,11 @@ public static Generator<FilterGroup> WithDefaults(this Generator<FilterGroup> ge
     public static Generator<FilterGroup> WithFilter(this Generator<FilterGroup> generator, Filter filter)
         => generator.ForInstance(s => s.SetFilter(filter));
 
+    public static Generator<FilterGroup> WithFilterItems(
```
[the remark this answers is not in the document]
ntsim: This method is actually here so that we can do stuff like this in consuming code:

```cs
WithFilterItems(filterItems[..2]); // multiple
WithFilterItems(filterItems[0]); // singular 
```

Main benefit is that we don't need to convert them to lists. Tbh, it's not really a big deal to convert them so have just removed this extension now.

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/SubjectCsvMetaServiceTests.cs
```diff
@@ -0,0 +1,1528 @@
...
+            {
+                Id = releaseSubject.Release.Id,
+            },
+            File = new File
+            {
+                SubjectId = releaseSubject.Subject.Id
+            }
+        };
+
+        var filterItems = _fixture.DefaultFilterItem().GenerateArray(4);
```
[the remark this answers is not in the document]
ntsim: I hadn't added in the shortcut extension for Filters from the read replica PR, but it does make things more concise so updated to use it.

I've also added additional extensions for filter groups and indicator groups so we can also do stuff like:

```cs
_fixture.DefaultFilterGroup(filterItemCount: 2);
_fixture.DefaultIndicatorGroup(indicatorCount: 3);
```

I've not gone with your second suggestion to inline things as we actually need to use most of those variables in various assertions so it doesn't really improve readability.

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/SubjectCsvMetaServiceTests.cs
```diff
@@ -0,0 +1,1528 @@
...
+                    .Generate(1))
+            )
+            .ForRange(1..2, f => f
+                .SetFilterGroups(_fixture.DefaultFilterGroup()
+                    .WithFilterItems(filterItems[2..])
+                    .Generate(1))
+            )
+            .GenerateList(2);
+
+        var indicators = _fixture.DefaultIndicator()
```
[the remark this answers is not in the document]
ntsim: As above - not gone with this suggestion as it doesn't really improve readability.

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/SubjectCsvMetaServiceTests.cs
```diff
@@ -0,0 +1,1528 @@
...
+        var contextId = Guid.NewGuid().ToString();
+
+        await using (var contentDbContext = InMemoryContentDbContext(contextId))
+        await using (var statisticsDbContext = InMemoryStatisticsDbContext(contextId))
+        {
+            await contentDbContext.ReleaseFiles.AddRangeAsync(releaseFile);
+            await contentDbContext.SaveChangesAsync();
+
+            await statisticsDbContext.ReleaseSubject.AddRangeAsync(releaseSubject);
+            await statisticsDbContext.Filter.AddRangeAsync(filters);
```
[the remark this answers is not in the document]
ntsim: In the various tests, the meta is now attached onto the subject as suggested, i.e.

```cs
var releaseSubject = new ReleaseSubject
{
    Release = _fixture.DefaultStatsRelease(),
    Subject = _fixture.DefaultSubject()
        .WithFilters(filters)
        .WithIndicatorGroups(indicatorGroups)
        .WithObservations(observations),
};
```

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectCsvMetaService.cs
ntsim: I don't think there's any hard and fast rules tbh. I like adding named parameters in some cases where the readability is improved and I considered this to be one of them.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkCsvMetaService.cs
```diff
@@ -0,0 +1,271 @@
...
+        return new PermalinkCsvMetaViewModel
+        {
+            Locations = locations,
+            Filters = filters,
+            Indicators = indicators,
+            Headers = headers
+        };
+    }
+
+    private async Task<Stream?> GetCsvStream(
```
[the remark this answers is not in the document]
ntsim: They're essentially the same yes. For now, I'm happy to leave it as-is as we don't have a third case, so I'm going to err on the side of YAGNI.

Also want to avoid any more significant re-working of this PR! :weary:

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Controllers/TableBuilderController.cs
ntsim: I've just given this a filename of `{releaseId}.csv`. Ideally, we might use the subject name, but we don't have it in scope and it's not really worth the effort to get it.

The frontend just sets its own filename anyway so this is a little redundant anyway.

ntsim:
FYI @duncan-at-hiveit I've added commit ea572a6a which fixes an issue with `Generator` creating the wrong number of instances when using the implicit `GenerateX` methods (where we don't provide an explicit count).

Ranges ends are not inclusive e.g. 1..2 only applies for index 1, 2..4 only applies for index 2 and 3, etc. However, when using `ForIndex`, the range it generates is like 1..1.

This previously worked in other test cases, but doesn't work great when using `ForRange` as the max index ends up being 1 greater than it should be.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Location.cs
```diff
@@ -12,6 +15,34 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Model
...
+                    // methods for getting the CSV columns on each type (less verbose and error prone).
+                    .Select(type => (LocationAttribute)FormatterServices.GetUninitializedObject(type))
+                    .OrderBy(attribute => attribute.CsvPriority)
+                    .SelectMany(attribute => attribute.CsvValues.Select(kv => kv.Key))
+                    .Distinct()
+                    .ToImmutableArray();
+            }
+        );
+
+        public static ImmutableArray<string> AllCsvColumns() => _allCsvColumns.Value;
```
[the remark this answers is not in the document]
duncan-at-hiveit: Cool ta!  Wasn't aware of the LocationAttribute CSV-specifics, but I think it's neater separated out, so cheers!

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/QueryableExtensionsTests.cs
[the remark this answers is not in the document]
duncan-at-hiveit: Cool!

duncan-at-hiveit (approved):

PR #3919, opened 2023-02-28
Ees 4002 allow management of pra users for amendments

duncan-at-hiveit:
This PR:
* adds the ability for a user to manage Pre-release Users on Amendments
* adds a warning on the Sign-off page if "Immediately" is selected and Pre-release Users / Invites exist on a Release or Amendment (saying that the users won't get a chance to use any PRA window)
* adds a warning on the Sign-off page if "Scheduled" is selected and Pre-release Users / Invites exist on a Release or Amendment (saying that users will be reviewing the Release in the PRA window)
* stops Pre-release Users / Invites and "Public pre-release list" content from being copied to Amendments - they should start with a clean slate
* provides a migration to clear out Pre-release Users from existing Amendments (as they should start with a clean slate - bear in mind that this user list would never have been visible on Amendments til now, so it won't come as a surprise!)

This PR allows Pre-release Users to be added / invited to Amendments and to therefore get informed about the PRA window for the Amendment.

The ability to see the Amendment in the PRA window comes for free!

We also add 2 warnings to the Sign-off page, but only if there are Pre-release Users or Invites added to a Release or an Amendment:

image

image

:warning: Note that the "Public pre-release list" content has no bearing on any of the warning logic!


tests/robot-tests/tests/admin/bau/prerelease_and_amend.robot
```diff
@@ -0,0 +1,302 @@
...
+${DATABLOCK_NAME}=      UI test table
+${RELEASE_NAME}=        Calendar Year 2000
+
+
+*** Test Cases ***
+Create test publication and release via API
+    ${PUBLICATION_ID}=    user creates test publication via api    ${PUBLICATION_NAME}
+    user creates test release via api    ${PUBLICATION_ID}    CY    2000
+
+Upload subject
```
duncan-at-hiveit, line 27: I think you can probably cut this part out (`Upload subject` and `Add metadata guidance`) as I don't think the subject is being used for the rest of the test?

We can get away with this because other Pre-release tests are doing the testing of Subjects more thoroughly.

This would also make this test run faster, which is always an added bonus :)

tests/robot-tests/tests/admin/bau/prerelease_and_amend.robot
```diff
@@ -0,0 +1,302 @@
...
+    user clicks link    Pre-release access
+    user waits until h2 is visible    Manage pre-release user access
+    user creates public prerelease access list    Initial test public access list
+
+Go to "Sign off" page
+    user clicks link    Sign off
+    user waits until h2 is visible    Sign off
+    user waits until page contains button    Edit release status
+
+Check that there is no Pre-release warning text on the sign off page
```
duncan-at-hiveit, line 68: I'd reword this just slightly from `there is no Pre-release warning text` to `there are no Pre-release warnings` if that makes sense?

tests/robot-tests/tests/admin/bau/prerelease_and_amend.robot
```diff
@@ -0,0 +1,302 @@
...
+    user clicks link    Sign off
+    user waits until h2 is visible    Sign off
+    user waits until page contains button    Edit release status
```
duncan-at-hiveit, line 71: I think that everywhere we use `user clicks button`, we don't actually need to put a `user waits until page contains button` keyword before it.

This is because under the hood, the `user clicks button` keyword itself also uses `user waits until parent contains element` to make sure the button is present on the page before it is clicked.

I appreciate that you probably found this code snippet from another test that was already doing this, but maybe we can remove the `user waits until page contains button` keywords from this particular test script if it's immediately followed by `user clicks button`?

tests/robot-tests/tests/admin/bau/prerelease_and_amend.robot
```diff
@@ -0,0 +1,302 @@
...
+
+Add release note to amendment
+    user clicks button    Add note
+    user enters text into element    id:createReleaseNoteForm-reason    Test release note
+    user clicks button    Save note
+    ${date}=    get current datetime    ${DATE_FORMAT_MEDIUM}
+    user waits until element contains    css:#releaseNotes li:nth-of-type(1) time    ${date}
+    user waits until element contains    css:#releaseNotes li:nth-of-type(1) p    Test release note
+
+Check Pre-release Access page during amendment
```
duncan-at-hiveit, line 154: Maybe to be more explicit about what we're testing here, we could rename this from `Check Pre-release Access page during amendment` to something like `Check that Pre-release users and Public access list are empty in new amendment`

tests/robot-tests/tests/admin/bau/prerelease_and_amend.robot
```diff
@@ -0,0 +1,302 @@
...
+    ${day}=    get current datetime    %d    1
+    ${month}=    get current datetime    %m    1
+    ${year}=    get current datetime    %Y    1
+    ${time_start}=    format uk to local datetime    ${year}-${month}-${day}T00:00:00    %-d %B %Y at %H:%M
+    ${time_end}=    format uk to local datetime    ${year}-${month}-${day}T23:59:00    %-d %B %Y at %H:%M
+    user checks page contains    Pre-release access will be available from ${time_start} until ${time_end}.
+
+Start prerelease
+    user changes to bau1
+    ${day}=    get current datetime    %-d    1
```
duncan-at-hiveit, line 257: I think (not as part of this PR) that what you were suggesting earlier would be really good in terms of looking for things to pull into common keywords.

One candidate for me would be this pattern:

```
    ${day}=    get current datetime    %-d    1
    ${month}=    get current datetime    %-m    1
    ${month_word}=    get current datetime    %B    1
    ${year}=    get current datetime    %Y    1
    .....
    .....
    user enters text into element    id:releaseStatusForm-publishScheduled-day    ${day}
    user enters text into element    id:releaseStatusForm-publishScheduled-month    ${month}
    user enters text into element    id:releaseStatusForm-publishScheduled-year    ${year}
```

which could be boiled down to a single line:

```
    user enters scheduled published date to be X number of days from today   1 
```

It could also be applied to existing keywords like `user approves release for scheduled publication` to make them simpler too.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseService.cs
```diff
@@ -278,16 +290,19 @@ public class ReleaseService : IReleaseService
...
-                .Where(releaseRole =>
-                    !releaseRole.SoftDeleted
-                    && releaseRole.ReleaseId == originalReleaseId)
-                .Select(releaseRole => releaseRole.CopyForAmendment(amendment));
+                // For auditing purposes, we also want to migrate release roles that have Deleted set (when a role is
+                // manually removed from a Release as opposed to SoftDeleted, which is only set when a Release is
+                // deleted)
+                .IgnoreQueryFilters()
+                .Where(releaseRole => releaseRole.ReleaseId == originalReleaseId 
+                                      && releaseRole.Role != ReleaseRole.PrereleaseViewer)
```
markhiveit, line 303: I'm not 100%, but I think we might want to keep the `SoftDeleted` check here. Possible you've already explored this, in which case leave it out - but thought I'd raise it.

I think there may have been a distinction between `SoftDeleted` and `Deleted`. I think `Deleted` implies a user has removed the permission, and we keep an audit trail, but `SoftDeleted` implies an overarching object like the release has been removed. I think that's why the original code filtered them out.

src/explore-education-statistics-admin/src/components/PageErrorBoundary.tsx
```diff
@@ -19,6 +19,16 @@ interface State {
...
+  private unregisterCallback?: () => void;
+
+  private errorPages = {
+    forbidden: () => {
+      this.setState({
+        errorCode: 403,
+      });
+    },
+  };
+
```
markhiveit, line 31: Has this moved for a reason? Not that I think it's an issue if there is no reason - just interested.

src/explore-education-statistics-admin/src/pages/release/pre-release/ReleasePreReleaseAccessPage.tsx
```diff
@@ -39,55 +39,53 @@ const ReleasePreReleaseAccessPage = () => {
     <LoadingSpinner loading={isLoading}>
       {release && (
         <Tabs id="preReleaseAccess">
-          {!release.amendment && (
```
markhiveit, line 42: I bet the realisation that removing this was all that was necessary was a sweet one.

tests/robot-tests/tests/admin/bau/prerelease_and_amend.robot
N-moh: I think it would be good to attach some data , to ensure reproduction of a realistic scenario for when the pre release viewer gets access to pre release content during amendment .

tests/robot-tests/tests/admin/bau/prerelease_and_amend.robot
```diff
@@ -0,0 +1,302 @@
...
+    ...    ${modal}
+    user checks list item contains    testid:invitableList    2    EES-test.ANALYST1@education.gov.uk    ${modal}
+    user clicks button    Confirm
+
+Create public prerelease access list for amendment
+    user clicks link    Pre-release access
+    user waits until h2 is visible    Manage pre-release user access
+    user creates public prerelease access list    Amended test public access list
+
+Validate prerelease has not started for Analyst user during amendment
```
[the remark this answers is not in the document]
N-moh: Good suggestion. I agree.

tests/robot-tests/tests/admin/bau/prerelease_and_amend.robot
N-moh: Noted for next time

tests/robot-tests/tests/admin/bau/prerelease_and_amend.robot
duncan-at-hiveit: My thoughts are that is tested more thoroughly in `prerelease.robot` and so isn't necessary to repeat here, but I could be wrong.  Worth checking?

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20230215154534_EES4002_RemovePreReleaseUsersFromAmendments.sql
```diff
@@ -0,0 +1,10 @@
+-- Delete any Pre-release users from any existing Release amendments, as we require them to initially have blank 
+-- pre-release user lists when an amendment is created.
+DELETE FROM UserReleaseRoles
+WHERE Id IN (
+    SELECT urr.Id
+    FROM UserReleaseRoles urr
+    JOIN Releases ON Releases.Id = urr.ReleaseId
+    WHERE urr.Role = 'PrereleaseViewer'
+    AND Releases.PreviousVersionId IS NOT NULL
```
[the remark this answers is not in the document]
duncan-at-hiveit: I think so yeah, as there would never have been a PRA window for any amendments up til this point (not officially anyway), as there was never a chance to add users to PRA access for Amendments until now.  Any PRA access would've been accidental and limited to the original PRA users for the first version of this Release.

src/explore-education-statistics-admin/src/pages/release/pre-release/ReleasePreReleaseAccessPage.tsx
```diff
@@ -39,55 +39,53 @@ const ReleasePreReleaseAccessPage = () => {
...
                 </p>
+
                 <p>
-                  Pre-release access via Explore Education Statistics is
-                  currently limited to DFE users only, if you need to share your
-                  release with external users you will need to do so outside of
-                  the system.
+                  <UrlContainer
+                    data-testid="prerelease-url"
+                    url={`${window.location.origin}${generatePath<
```
luke-h1, line 72: Aware this isn't part of your changes but I'm thinking we can potentially improve the value we pass to the `url` prop here by replacing `window.location.origin` to `${publicAppUrl}` by using the `useConfig` hook (see `explore-education-statistics-admin/src/pages/release/ReleaseStatusPage.tsx` for an example of its usage)

markhiveit (approved):

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PublicPreReleaseAccessForm.tsx
```diff
@@ -88,9 +88,9 @@ const PublicPreReleaseAccessForm = ({
         </Formik>
       ) : (
         <>
-          {isReleaseLive && (
+          {!canUpdateRelease && (
```
amyb-hiveit, line 91: Just wondering if the check around the inset text above should also be changed from `!isReleaseLive` to `!canUpdateRelease`? (L51 - Github won't let me leave a comment on the line itself)

amyb-hiveit (approved):
Just one question but looks good FE-wise 👍

src/explore-education-statistics-admin/src/pages/release/pre-release/ReleasePreReleaseAccessPage.tsx
duncan-at-hiveit: Ahh would be good, but I think it's actually the Admin URL we're getting here as opposed to the public URL and we don't currently have that in config unfortunately.

src/explore-education-statistics-admin/src/pages/release/pre-release/components/PublicPreReleaseAccessForm.tsx
duncan-at-hiveit: I'm not 100% sure - I was happy changing it here as this `This release has been approved, and can no longer be updated.` is always controlled by the `canUpdateRelease` flag on all other pages, whereas the logic above is governed by slightly different rules.  So I'd be tempted to leave that as is for now as it seems to be deliberately tied to the "liveness" of the Release.

amyb-hiveit: Ok, makes sense 👍

PR #3923, opened 2023-02-28
EES-3913 Add input length validation and fix KeyStatisticDataBlock cascade delete bug

markhiveit:
This PR makes two changes:

### Adds input length validation for key stats. 

Specifically for the title, statistic, trend, and guidance title field. Lengths have been chosen based on a combination of what is currently the max length in use on production and what fits the UI. These values have been run past Cam.

On the backend, the lengths are stored as consts in the `KeyStatistic` class. This was done because originally we intended add these contraints to the database, which means they were reused more times. This is also why I haven't made a similar change on the frontend - it doesn't seem worth creating a central location for them given they only get used two places at most. But obviously happy to change that if others want that.

### Fixes KeyStatisticDataBlock cascade delete bug

Nusrath found a bug where entries from `KeyStatisticsDataBlock` table were being removed without the associated `KeyStatistics` table entry being removed. This was happening due to this code:
```
            modelBuilder.Entity<KeyStatisticDataBlock>()
                .ToTable("KeyStatisticsDataBlock");
```
This code was adding a cascade delete on the `DataBlockId`, if the associated entry in the `ContentBlock` table was removed. Credit to Ben to working out what to add to `ContentDbContext` to remove this cascade delete. This change also meant that a `KeyStatisticDataBlock` must be removed before its associated `DataBlock` object can be removed.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
@@ -539,9 +539,9 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Model.Database
             modelBuilder.Entity<KeyStatisticDataBlock>()
                 .HasOne<DataBlock>(ks => ks.DataBlock)
                 .WithMany()
-                // WARN: This is necessary - otherwise a cascade delete occurs when an associated data block is removed.
-                // This cascade delete _only_ removes the KeyStatisticsDataBlock entry, leaving a KeyStatistics table
-                // entry.
+                // WARN: This is necessary - otherwise an automatically generated cascade delete is added for when an
+                // associated data block is removed. That cascade delete _only_ removes the KeyStatisticsDataBlock
+                // entry, leaving a KeyStatistics table entry, which should never happen.
                 .OnDelete(DeleteBehavior.NoAction);
 
             modelBuilder.Entity<KeyStatisticText>()
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
@@ -535,6 +535,15 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Model.Database
 
             modelBuilder.Entity<KeyStatisticDataBlock>()
                 .ToTable("KeyStatisticsDataBlock");
+
+            modelBuilder.Entity<KeyStatisticDataBlock>()
+                .HasOne<DataBlock>(ks => ks.DataBlock)
+                .WithMany()
+                // WARN: This is necessary - otherwise a cascade delete occurs when an associated data block is removed.
+                // This cascade delete _only_ removes the KeyStatisticsDataBlock entry, leaving a KeyStatistics table
+                // entry.
+                .OnDelete(DeleteBehavior.NoAction);
+
             modelBuilder.Entity<KeyStatisticText>()
                 .ToTable("KeyStatisticsText");
         }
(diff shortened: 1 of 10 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
```diff
@@ -535,6 +535,15 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
                 .ToTable("KeyStatisticsDataBlock");
+
+            modelBuilder.Entity<KeyStatisticDataBlock>()
+                .HasOne<DataBlock>(ks => ks.DataBlock)
+                .WithMany()
+                // WARN: This is necessary - otherwise a cascade delete occurs when an associated data block is removed.
+                // This cascade delete _only_ removes the KeyStatisticsDataBlock entry, leaving a KeyStatistics table
+                // entry.
+                .OnDelete(DeleteBehavior.NoAction);
+
```
benoutram, line 546: `DataBlockService.DeleteDataBlocks` is still saving changes before the `KeyStatisticsDataBlock` is deleted, so I would have thought right now there's still a bug where that fails when saving changes?

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
```diff
@@ -535,6 +535,15 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
 
             modelBuilder.Entity<KeyStatisticDataBlock>()
                 .ToTable("KeyStatisticsDataBlock");
+
+            modelBuilder.Entity<KeyStatisticDataBlock>()
+                .HasOne<DataBlock>(ks => ks.DataBlock)
+                .WithMany()
+                // WARN: This is necessary - otherwise a cascade delete occurs when an associated data block is removed.
+                // This cascade delete _only_ removes the KeyStatisticsDataBlock entry, leaving a KeyStatistics table
+                // entry.
```
benoutram, line 544: I don't understand this comment, because there's no cascade delete now.
Should it be `// That cascade delete _only_ removed`?

src/GovUk.Education.ExploreEducationStatistics.Content.Model/KeyStatistic.cs
```diff
@@ -6,6 +6,11 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Model;
 
 public abstract class KeyStatistic : ICreatedUpdatedTimestamps<DateTime, DateTime?>
 {
+    public const int TitleMaxLength = 24;
+    public const int StatisticMaxLength = 11;
+    public const int TrendMaxLength = 229;
+    public const int GuidanceTitleMaxLength = 65;
```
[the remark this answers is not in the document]
ntsim: Agreed, these numbers are just weird. Would round them off as Ben suggested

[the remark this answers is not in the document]
markhiveit: I ran this by Cam and he said go with 25/12/230/65 - so I'll change it to that.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
markhiveit: I've updated the comment - hopefully clearer now.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
markhiveit: Discussed on Slack - this has been done. I also checked for other locations where this might happen. I believe the only place is in `ContentService#RemoveContentBlock`, but that code should never get run and will get removed in the cleanup ticket EES-3988, so I'm content to leave that.

benoutram (approved):

PR #3924, opened 2023-02-28
EES-4141 - add HTML reporter to k6 tests

luke-h1:
This PR: 
adds a html reporter to the performance tests in order to help with the comparison of stats between stat runs. An example of the report can be found below:

Screenshot 2023-02-28 at 14 19 27


tests/performance-tests/src/tests/public/permalinkPage.test.ts
```diff
@@ -82,11 +83,13 @@ const performTest = () => {
...
-        response.status
-      } response code`,
+      `Failed to get permalink page. Received ${response.status} response code`,
     );
   }
 };
-
+export function handleSummary(data: unknown) {
+  return {
+    'getReleaseApi.html': htmlReport(data),
```
duncan-at-hiveit, line 92: Should be permalinkPage.html here

tests/performance-tests/src/types.d.ts
```diff
@@ -1,3 +1,7 @@
 declare module 'dotenv-json-complex' {
   export default function ({ environment: string });
 }
+
+declare module 'https://raw.githubusercontent.com/benc-uk/k6-reporter/main/dist/bundle.js' {
```
duncan-at-hiveit, line 5: Noice - wouldn't have known to do this.

tests/performance-tests/src/tests/admin/import/import.test.ts
```diff
@@ -3,6 +3,7 @@ import { check, fail } from 'k6';
 import { Counter, Rate, Trend } from 'k6/metrics';
 import { Options } from 'k6/options';
 import exec from 'k6/execution';
+import { htmlReport } from 'https://raw.githubusercontent.com/benc-uk/k6-reporter/main/dist/bundle.js';
```
duncan-at-hiveit, line 6: This'll be cool - will we still get our CLI report too?

tests/performance-tests/src/types.d.ts
luke-h1: Yeah a lot of the k6 libs that we import like this are all in JS so unfortunately we have to duck type a bit here by declaring the types ourselves

tests/performance-tests/src/tests/admin/import/import.test.ts
luke-h1: Yep, the report via the CLI will work as normal

duncan-at-hiveit (approved):

PR #3927, opened 2023-03-02
EES-3988 Key stats clean up

markhiveit:
This PR cleans various things up after the changes made in EES-3828:

### Removes the unused `DataBlock_Summary` column from the `ContentBlock` table.

Previously, `trend`, `guidanceTitle`, and `guidanceText` were stored in JSON in the `DataBlock_Summary` column in the `ContentBlock` table as `dataSummary`, `dataDefinitionTitle` and `dataDefinition` respectively. The summary column also has a `dataKeys` field, which according to `documentation/charts.md` was the legacy name given for indicators. This isn't being used and I don't believe has been used for a long time. I'm not sure why `dataKeys` is an array - it's possible this was used for something else? Or so a single data block could have multiple key stats associated with it?

As key stats now have their own table, `DataBlock_Summary` is now unused. All key stat data contained in this column has been migrated to the new tables in EES-3828, and as the `dataKeys` field is not used and hasn't for a long time, there is no reason to fear losing data here.

### Removes KeyStatistics table `ContentBlockIdTemp`

This column was added in EES-3828 to make migrating data from `DataBlock_Summary` into `KeyStatistics` and `KeyStatisticsDataBlock` easier. Now it has served that purpose, it can be removed.

### Remove documentation/charts.md

This file seems wildly out of date, so I think we can remove it.

### Clean up data in the database and associated code

In the future, any data block that has an associated key stat will have a null `ContentSectionId`, but after the migration there are data blocks that have an associated key stat AND have a `ContentSectionId` set. This is problematic, as if one of these migrated key stats is removed, and the `ContentSectionId` remains set, the data block won't be recognised as unattached and won't be available for selection again. For this reason, code was added in EES-3828 to unset a data block's ContentSectionId when a key stat is removed.

The PR has a migration that will set the ContentSectionId to null for any data block with an associated key stat. This means that the code that sets `ContentSectionId` to null when removing a key stat can be removed.

Once that is done, we can then remove all ContentSections of the 'KeyStatistics' type, as they are no longer used.

A couple of commands should be run before deploying this PR to ensure these database clean up tasks will run smoothly:
```
SELECT * FROM [dbo].[ContentBlock]
WHERE Type = 'DataBlock'
AND Id IN (SELECT DISTINCT DataBlockId From KeyStatisticsDataBlock)
AND ContentSectionId IS NOT NULL
AND ContentSectionId NOT IN (SELECT CS.Id FROM ContentSections CS WHERE CS.Type != 'KeyStatistics');

SELECT * FROM ContentSections
WHERE Type = 'KeyStatistics';
AND Id NOT IN (SELECT CB.ContentSectionId FROM ContentBlock CB);
```
I'll leave a note in the EES-4055 to do that alongside other faffy deploy tasks.

There is some risk with these queries. For example, if not all `KeyStatistics` content sections are removed, public release pages won't be able to be generated. So I intend to keep a close eye on this as it is deployed.

As we're doing this clean up, we can also remove some code - for example, the code that unsets the ContentSectionId of a data block when it's associated key stat is removed, as that code should never be run.


src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20230224161839_CleanUpKeyStatData.sql
```diff
@@ -0,0 +1,9 @@
+﻿UPDATE ContentBlock
+SET ContentSectionId = NULL
+WHERE Type = 'DataBlock'
+  AND Id IN (SELECT DISTINCT DataBlockId FROM KeyStatisticsDataBlock)
+  AND ContentSectionId IS NOT NULL
+  AND ContentSectionId NOT IN (SELECT CS.Id FROM ContentSections CS WHERE CS.Type != 'KeyStatistics'); -- this condition for safety
+
+DELETE ContentSections FROM ContentSections WHERE Type = 'KeyStatistics'
```
benoutram, line 8: `DELETE FROM ContentSections` would be the way to write a regular delete statement.

`DELETE ContentSections FROM ContentSections` would only be needed as a result of joining with another table.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ContentServiceTests.cs
```diff
@@ -622,24 +622,14 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
-                var keyStatisticService = new Mock<IKeyStatisticService>(MockBehavior.Strict);
-                keyStatisticService
-                    .Setup(s =>
-                        s.DeleteAssociatedKeyStatisticDataBlock(dataBlockId))
-                    .Returns(Task.CompletedTask);
-
-                var service = SetupContentService(
-                    contentDbContext: contentDbContext,
-                    keyStatisticService: keyStatisticService.Object);
+                var service = SetupContentService(contentDbContext: contentDbContext);
```
benoutram, line 625: I don't think this test is even relevant anymore. It's named `RemoveContentBlock_AssociatedWithKeyStat` and is setting up a content section with a data block that's also a key statistic data block. That doesn't happen anymore.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/KeyStatisticService.cs
```diff
@@ -165,46 +165,14 @@ await _dataBlockService.IsUnattachedDataBlock(releaseId, dataBlock)
...
-                {
-                    var dataBlock = keyStatDataBlock.DataBlock;
-                    dataBlock.ContentSection = null;
-                    dataBlock.ContentSectionId = null;
-                    _context.ContentBlocks.Update(dataBlock);
-                    await _context.SaveChangesAsync();
-                }
-                return keyStat;
-            })
+                    query.Where(keyStat => keyStat.ReleaseId == release.Id)))
```
benoutram, line 168: I don't think this is going to have any effect because the variable named `query` here is the hydration function, but I might be wrong.

If you write a unit test for this where the key statistic isn't for the release I reckon it would pass currently.

I suggest this is rewritten

```
            .OnSuccessDo(_userService.CheckCanUpdateRelease)
            .OnSuccess(() => _persistenceHelper.CheckEntityExists<KeyStatistic>(q => q
                .Where(ks => ks.Id == keyStatisticId && ks.ReleaseId == releaseId)))
```

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20230224161839_EES3988KeyStatisticCleanUp.cs
```diff
@@ -0,0 +1,42 @@
...
+        private const string MigrationId = "20230224161839";
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropColumn(
+                name: "ContentBlockIdTemp",
+                table: "KeyStatistics");
+
+            migrationBuilder.DropColumn(
+                name: "DataBlock_Summary",
+                table: "ContentBlock");
```
benoutram, line 21: How's the confidence level on charts not relying on anything to do with DataBlock.Summary? 🙂

markhiveit: Pretty high. If it was being used, I would have expected to have seen something go red when I removed it and all associated view models etc. But nothing was using it except key stats. 

If there is a way for me to get caught out over this, I'm not sure what it is. I don't see how anything else could be accessing this column. If we were using javascript, I suspect this could have been a nightmare, but with it being typescript/c#, I think we're fairly safe.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ContentServiceTests.cs
markhiveit: I've removed the test

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/KeyStatisticServiceTests.cs
```diff
@@ -1235,64 +1219,73 @@ await using (var context = InMemoryContentDbContext(contextId))
...
 
             result.AssertNotFound();
         }
+
+        await using (var context = InMemoryContentDbContext(contextId))
+        {
+            var keyStatistics = context.KeyStatistics.ToList();
+            Assert.Equal(4, keyStatistics.Count);
+        }
     }
```
benoutram, line 1289: I think in general the tests in this class are quite bloated with properties and extra data that aren't really required.

I'm not too bothered if you don't want to make this change, but I'll give an example for this particular unit test of how short this could be avoiding setting up all the id's, extra properties, and avoiding the collection of key stats on the releases.

```
var release = new Release();
var otherRelease = new Release();

var keyStatisticText = new KeyStatisticText
{
  Release = release,
  ...<any other attributes that are really necessary>
};


Save the release, otherRelease, and the keyStatisticText.

Call the service delete method with keyStatisticText.Id and otherRelease.Id

Check for the not found error.

Check that the one key stat still exists.
```

markhiveit: I've changed the one test you highlighted. I think you're right it could have been more succinct. I've left the rest of the tests, but will try to do this better from now on.

benoutram (approved):

PR #3953, opened 2023-03-17
Latest snapshots

benoutram:

### DIFF
--- a/tests/robot-tests/tests/snapshots/methodologies_snapshot.json
+++ b/tests/robot-tests/tests/snapshots/methodologies_snapshot.json
@@ -298,6 +298,12 @@
           "Multiplication tables check attainment"
         ],
         "topic_heading": "Key stage 2"
+      },
+      {
+        "methodologies": [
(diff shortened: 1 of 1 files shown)


cjrace:
Is there any reason why these wouldn't have been captured at 9:35 yesterday? Both publications went out at 9:30 as far as I'm aware? @benoutram

benoutram:
> Is there any reason why these wouldn't have been captured at 9:35 yesterday? Both publications went out at 9:30 as far as I'm aware? @benoutram
> 
> I'm not overly worried but wanted to check if there was something that was causing it to not show yesterday

I made a mistake and didn't start from the latest snapshots, so I was including yesterday's changes as well. Please check the diff now where you should see there's only one change, and hopefully it's the change you expected. 🙂 .

I still need to look into why we got 4 file alerts today when only 1 file has changed.

cjrace (approved):
Ah okay, that makes sense. This is the change I was expecting today now 😄

PR #3958, opened 2023-03-20
EES-3791 permalink table endpoint

amyb-hiveit:
Adds a NextJS API endpoint to generate table JSON for permalinks.

The endpoint accepts POST requests with fullTable and tableHeadersConfiguration in the body and returns the JSON for the table and its title.

It expects a PERMALINK_API_SECRET for authorisation, so a shared secret will have to be set up for this to be used by the front and back ends.


src/explore-education-statistics-frontend/src/pages/api/permalink.ts
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * Expects req.body to be:
+ * {
+ *  fullTable: TableDataResponse,
+ *  configuration: {
+ *   tableHeaders: UnmappedTableHeadersConfig
+ *  }
+ * }
+ */
+export default async function handlePermalink(
```
luke-h1, line 24: We might want to call this `permalinkHandler` to fit in with `Next.js` naming conventions of API endpoints

src/explore-education-statistics-frontend/src/pages/api/permalink.ts
```diff
@@ -0,0 +1,72 @@
...
+  req: NextApiRequest,
+  res: NextApiResponse<ResponseData>,
+) {
+  if (
+    process.env.NODE_ENV === 'production' &&
+    req.query.PERMALINK_API_SECRET !== process.env.PERMALINK_API_SECRET
+  ) {
+    res.status(401).send({ error: 'Unauthorised' });
```
luke-h1, line 32: We can make this a bit more succulent and just do `return res.status(401).send({ error: 'Unauthorised' });`

src/explore-education-statistics-frontend/src/pages/api/permalink.ts
```diff
@@ -0,0 +1,72 @@
...
+    return;
+  }
+
+  if (req.method === 'POST') {
```
luke-h1, line 36: Perhaps we could refactor the handling of methods into a higher-order function. Be interesting to hear other poeple's thoughts on this. I've found it just reduces noise inside the function body due to not having `if` statements to handle whether or not the right method is being supplied. 

This function would look something like: 

```typescript
import type { NextApiHandler, NextApiRequest, NextApiResponse } from 'next';

type Method = 'GET' | 'POST' | 'PUT' | 'PATCH' | 'DELETE' | 'OPTIONS' | 'HEAD';

export default function withMethods(methods: Method[], handler: NextApiHandler) {
  // eslint-disable-next-line func-names
  return async function (req: NextApiRequest, res: NextApiResponse) {
    if (!methods.includes(req.method as Method)) {
      return res.status(405).end();
    }

    return handler(req, res);
  };
}

```

and would be used like: 

```typescript 
const handler = async (req: NextApiRequest, res: NextApiResponse) => {
  // ...
};
export default withMethods(['GET'], handler);
```

src/explore-education-statistics-frontend/src/pages/api/permalink.ts
```diff
@@ -0,0 +1,72 @@
...
+    const {
+      body: { configuration, fullTable: unmappedFullTable },
+    } = req;
+
+    if (!unmappedFullTable || !configuration) {
+      res.status(400).send({ error: 'fullTable and configuration required' });
```
luke-h1, line 42: Would suggest `return res.status(400).send({ error: 'fullTable and configuration required' });`

src/explore-education-statistics-frontend/src/pages/api/permalink.ts
```diff
@@ -0,0 +1,72 @@
...
+import type { NextApiRequest, NextApiResponse } from 'next';
+
+interface ResponseData {
+  table?: string;
+  title?: string;
+  error?: unknown;
+}
+
+/**
+ * Endpoint to generate table json and title for permalinks.
```
luke-h1, line 14: Just wondering if we've thought of validating that the body is in this shape with something such as zod?

luke-h1 (changes requested):

src/explore-education-statistics-frontend/src/pages/api/permalink.ts
```diff
@@ -0,0 +1,72 @@
...
+ *  }
+ * }
+ */
+export default async function handlePermalink(
+  req: NextApiRequest,
+  res: NextApiResponse<ResponseData>,
+) {
+  if (
+    process.env.NODE_ENV === 'production' &&
+    req.query.PERMALINK_API_SECRET !== process.env.PERMALINK_API_SECRET
```
luke-h1, line 30: Just wondering how the `API_SECRET` is going to be generated?

ntsim (changes requested):

src/explore-education-statistics-frontend/src/pages/api/permalink.ts
amyb-hiveit: As this is a temporary endpoint that can be removed once the permalinks migration is completed then it seems a bit unnecessary to add middleware for it which will be unused once the endpoint is removed. Unless we expect that we'll be adding other endpoints in future?

ntsim: From what we've discussed previously, I think this endpoint will still be used even after the migration is complete. The flow would look something like this:


```mermaid
sequenceDiagram
    Browser->>Content API: POST /api/permalink
    Content API->>Next.js: POST /api/permalink
    Next.js-->>Content API: Rendered permalink table
    Content API->>Blob Storage: Uploads permalink table
    Content API-->>Browser: Rendered permalink table
```

I think [EES-3792](https://dfedigital.atlassian.net/browse/EES-3792) documented an approach you might be thinking of. We had 3 Amigo's a couple of weeks ago and we changed various things in it to reflect the above.

It'd also be good to have some of this infrastructure in place for the next time we need Next.js endpoints as well.

amyb-hiveit: Ah ok, I didn't realise the plan had changed from when Ben and I originally wrote the tickets (ages ago now!).

src/explore-education-statistics-frontend/src/pages/api/__tests__/permalink.test.ts
```diff
@@ -0,0 +1,291 @@
...
+
+describe('permalink api route', () => {
+  type ApiRequest = NextApiRequest & ReturnType<typeof createRequest>;
+  type ApiResponse = NextApiResponse & ReturnType<typeof createResponse>;
+
+  const originalEnv = process.env;
+
+  describe('unauthorised', () => {
+    beforeEach(() => {
+      jest.resetModules();
```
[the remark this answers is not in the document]
amyb-hiveit: I thought I needed it because of mocking the env values, but could have been wrong. It's gone now anyway as don't need to do that anymore.

src/explore-education-statistics-frontend/src/pages/api/permalink.ts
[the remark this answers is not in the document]
amyb-hiveit: Ok, I've taken it out.  Can bring up whether it's worth doing proper authentication in the next dev meeting and can add a ticket for it then if necessary.

amyb-hiveit: It's been removed now.

src/explore-education-statistics-frontend/src/pages/api/permalink.ts
amyb-hiveit: Middleware added now.

src/explore-education-statistics-frontend/src/modules/api/permalink/__tests__/createPermalinkTable.test.ts
```diff
@@ -0,0 +1,238 @@
...
+import {
+  createApiMocks,
+  createApiRequest,
+  createApiResponse,
+} from '@frontend-test/apiRouteMocks';
+import type { NextApiRequest, NextApiResponse } from 'next';
+
+describe('permalink api route', () => {
+  type ApiRequest = NextApiRequest & ReturnType<typeof createApiRequest>;
+  type ApiResponse = NextApiResponse & ReturnType<typeof createApiResponse>;
```
ntsim, line 11: We don't need these anymore

src/explore-education-statistics-frontend/src/modules/api/permalink/__tests__/createPermalinkTable.test.ts
```diff
@@ -0,0 +1,238 @@
...
+        message: 'fullTable and configuration required',
+        status: 400,
+      });
+    });
+
+    test('returns a 400 when the body does not contain fullTable and configuration', async () => {
+      const {
+        req,
+        res,
+      }: { req: ApiRequest; res: ApiResponse } = createApiMocks({
```
ntsim, line 33: We don't need the explicit type anymore if we use `createApiMocks`. Same applies to other places in this test suite

src/explore-education-statistics-frontend/src/modules/api/permalink/__tests__/createPermalinkTable.test.ts
amyb-hiveit: doh, yes, sorted now.

src/explore-education-statistics-frontend/src/modules/api/permalink/createPermalinkTable.ts
```diff
@@ -0,0 +1,66 @@
+import mapFullTable from '@common/modules/table-tool/utils/mapFullTable';
+import mapTableHeadersConfig from '@common/modules/table-tool/utils/mapTableHeadersConfig';
+import mapTableToJson, {
+  TableJson,
+} from '@common/modules/table-tool/utils/mapTableToJson';
+import generateTableTitle from '@common/modules/table-tool/utils/generateTableTitle';
+import { ErrorBody } from '@frontend/modules/api/types/error';
```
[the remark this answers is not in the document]
amyb-hiveit: It could be used by other endpoints in future so probably better to have in it's own file.

src/explore-education-statistics-frontend/next.config.js
```diff
@@ -15,6 +15,7 @@ const nextConfig = {
     APPINSIGHTS_INSTRUMENTATIONKEY: process.env.APPINSIGHTS_INSTRUMENTATIONKEY,
     GA_TRACKING_ID: process.env.GA_TRACKING_ID,
     PUBLIC_URL: process.env.PUBLIC_URL,
+    PERMALINK_API_SECRET: process.env.PERMALINK_API_SECRET,
```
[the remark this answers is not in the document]
amyb-hiveit: Yep, meant to remove that. It's gone now.

src/explore-education-statistics-frontend/src/modules/api/permalink/createPermalinkTable.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    const title = generateTableTitle(fullTable.subjectMeta);
+
+    if (tableJson && title) {
+      return res.status(200).send({ table: tableJson, title });
+    }
+    return res.status(500).send({ message: 'Cannot build table', status: 500 });
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.error(error);
```
ntsim, line 61: Just realised we should use `logger` here, otherwise the error won't get logged to Application Insights. We can also remove the lint as a side bonus.

amyb-hiveit: Cool, didn't know about that. I've changed it now.

ntsim (approved):

luke-h1 (approved):

PR #3972, opened 2023-03-27
EES-4193 - automate snapshot pull requests

luke-h1:
This PR:
* Automates the creation of snapshot pull requests. Previously, we'd get an alert at 9:30 AM in the `#alerts` channel notifying us that `create_snapshots.py` has detected differences. We'd then have to manually create a branch, push, create a PR, assign the team as reviewers and then notify the `#alerts` channel about it. Now, pull requests will automatically be raised by the `dfe-sdt` bot user. The Action `peter-evans/create-pull-request@v4` will look for changed files inside the `tests/robot-tests/tests/snapshots` directory and if there are changed files, it will raise a pull request targeting dev.

### Relevant changes: 
* Added a fine-grained personal access token `GH_TOKEN` as an action secret to allow the workflow to raise pull requests on behalf of the `dfe-sdt` user
* Added `SLACK_WEBBHOOK_URL` as an action secret
* Set the workflow to run at `9:30AM` every day Monday to Friday (and on workflow dispatch so we can run it at other times if necessary)
* Added dependency caching to reduce wasted time installing python dependencies

### DIFF
--- a/.github/workflows/validate-snapshots.yml
+++ b/.github/workflows/validate-snapshots.yml
@@ -1,0 +1,66 @@
+name: Validate snapshots
+on:
+  workflow_dispatch:
+  schedule: 
+    - cron: '35 9,12,16 * * 1-5'
+env:
+  GH_TOKEN: ${{ secrets.GH_TOKEN }}
+  SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+  PYTHON_VERSION: 3.10.10
+
+jobs:
+  test:
+    name: Validate snapshots
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Install moreutils
+        run: sudo apt install moreutils
+
+      # https://github.com/actions/virtual-environments/issues/1187
+      - name: tune linux network
+        run: sudo ethtool -K eth0 tx off rx off            
+
+      - name: checkout
+        uses: actions/checkout@v3
+
+      - name: Setup python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+          cache: 'pipenv'
+
+      - name: Install pipenv
+        run: curl https://raw.githubusercontent.com/pypa/pipenv/master/get-pipenv.py | python
+
+      - name: pipenv install
+        run: pipenv install
+
+      - name: Validate snapshots
+        working-directory: tests/robot-tests
+        run: pipenv run python scripts/create_snapshots.py --validate --slack-webhook-url ${{ env.SLACK_WEBHOOK_URL }}
+
+      - name: Set output variables
+        id: vars
+        run: |
+          pr_title="chore(tests): Update test snapshots $(date +%d-%m-%Y)"
+          pr_body="This PR was auto-generated on $(date +%d-%m-%Y) \
+            by [create-pull-request](https://github.com/peter-evans/create-pull-request) \
+            due to visual differences identified"
+          echo "pr_title=$pr_title" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


.github/workflows/validate-snapshots.yml
```diff
@@ -0,0 +1,67 @@
+name: Validate snapshots
+on:
+  workflow_dispatch:
+  schedule: 
+    - cron: '30 9 * * 1-5'
+env:
+  PAT_TOKEN: ${{ secrets.PAT_TOKEN }}
```
ntsim, line 7: Not a biggy, but weird name as it would stand for personal access token token? It could also be a little bit more specific.

Perhaps `GITHUB_TOKEN` would be better?

luke-h1: Yep good thinking, that does the trick 😉. I'll get in touch with Cam tomorrow and update the repository secret with a regenerated token

.github/workflows/validate-snapshots.yml
```diff
@@ -0,0 +1,67 @@
+name: Validate snapshots
+on:
+  workflow_dispatch:
+  schedule: 
+    - cron: '30 9 * * 1-5'
```
benoutram, line 5: Could run slightly after 9.30. Publishing isn't guaranteed to happen exactly at 9.30 precisely because the Publisher function app has to initialise and do some processing. I suppose this action takes some time to do it's job too though so maybe it's ok?

Because the published data can change at any time during the day not just at 9.30, depending on what the reviewers think to the possibility of being hit by PR's at different times of the day, this could run say 2 or 3 times during the day?

I think that's a good idea also because this is ultimately a testing tool, and bugs with the data rendered by the site could present themselves at any time during the day, and not just when releases are published/amended, and not limited to weekdays. For that reason I think it could also be beneficial to run at weekends.

.github/workflows/validate-snapshots.yml
```diff
@@ -0,0 +1,67 @@
...
+          commit-message: 'chore(tests): update test snapshots - ${{ env.DATE }}'
+          committer: ${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>
+          branch-suffix: timestamp
+          branch: chore/update-test-snapshots
+          add-paths: |
+            tests/robot-tests/tests/snapshots
+          reviewers: rmbielby, lauraselby, cjrace, chfoster, twilliams24
+          title: ${{ steps.vars.outputs.pr_title }}
+          body: ${{ steps.vars.outputs.pr_body }}          
+          draft: false
```
benoutram, line 67: Do we need to think about what happens if a previous PR has not been merged yet i.e. if there's any delay in the reviews?

Let it stack up multiple PR's / update the existing PR / close an existing PR?

luke-h1: With our current strategy of prepending a time stamp on the end of the commit, a new pull request would be created. Good point though, we can probably get rid of using the time stamp strategy and just have it push to a pre-set branch so that the Action will automatically update the branch and not create a new pull request

.github/workflows/validate-snapshots.yml
luke-h1: Unfortunately due to the caching of dependencies, this workflow kicks off pretty quick. I'll schedule it to run at 9:35AM instead. Yep, sounds good to run it at other times during the day when releases are getting published. Something such as 9:30AM, 12PM, 4PM mon-fri?

benoutram: Cool, the proposed schedule sounds good to me 🙂 .

.github/workflows/validate-snapshots.yml
benoutram: That sounds good, we could do that and change it if there's any problems. 👍

benoutram (approved):

PR #3976, opened 2023-03-30
EES-4193 - fix cron schedule and raising of snapshot PRs

luke-h1:
This PR: 
* Fixes the validate-snapshots action running 1 hour ahead 
* Removes validate option since we always want to write to a file in CI now in order for the pull requestion action to pick up the changed files
* Changes commiter to always be `dfe-sdt`


.github/workflows/validate-snapshots.yml
```diff
@@ -56,7 +58,7 @@ jobs:
         with:
           token: ${{ env.GH_TOKEN }}
           commit-message: 'chore(tests): update test snapshots - ${{ env.DATE }}'
-          committer: ${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>
+          committer: dfe-sdt@users.noreply.github.com>
```
benoutram, line 61: Think the `>` is wrong here

tests/robot-tests/scripts/create_snapshots.py
```diff
@@ -265,13 +266,6 @@ def write_snapshots_to_file(self):
         required=False,
     )
 
-    ap.add_argument(
-        "--validate",
-        dest="validate",
-        action="store_true",
-        help="Compares the latest snapshots to the existing versions and does not write any snapshot files",
-    )
-
```
benoutram, line 274: I think it was me that added this option because I thought it's useful for running locally.
Couldn't we leave this in but just remove `--validate` from the run command like you've already done?

.github/workflows/validate-snapshots.yml
```diff
@@ -2,7 +2,9 @@ name: Validate snapshots
 on:
   workflow_dispatch:
   schedule: 
-    - cron: '35 9,12,16 * * 1-5'
+    # Action runs 1 hour ahead of GMT
```
benoutram, line 5: I think the comment here should say actions are scheduled in UTC.

.github/workflows/validate-snapshots.yml
```diff
@@ -2,7 +2,9 @@ name: Validate snapshots
...
+    # This action runs at 9:35am, 12am, 4pm UK time
+    - cron: '35 8,11,15 * * 1-5'
```
benoutram, line 7: Doesn't look like there's any support currently for timezones and daylight saving time?

Based on the old setting we'd expect the schedule to trigger at:

```
UTC/GMT: 9:35, 12:35, 16:35
BST: 10:35, 13:35, 17:35
```

But the weird thing is according the history of scheduled runs, yesterday it ran at 5:52 GMT (6:52 BST). 😕 

Based on the new setting we'd expect it to be:

```
UTC/GMT: 8:35, 11:35, 15:35
BST: 9:35, 12:35, 16:35
```

Until they add time zone support, to get it to run at 9.35 I think we'd have to change update this config twice a year, swapping it from 9:35 to 8:35 at the beginning of British Summer Time and swapping it back to 9:35 at the end of British Summer Time.

Could add a comment to say this?
Also do you know why it ran at 5:52 GMT (6:52 BST) yesterday? I might be missing something here.

tests/robot-tests/scripts/create_snapshots.py
luke-h1: Ah right ok, yep sounds good

.github/workflows/validate-snapshots.yml
luke-h1: Sounds good, I've added a comment around this. I tried to run the snapshot pipeline manually yesterday but Actions had a major outage yesterday which affected the time the pipeline actually ran at.

benoutram: Ok good to know about the outage, thanks!

tests/robot-tests/scripts/create_snapshots.py
benoutram: Ahh, I think I see the problem you were trying to resolve now. The action needs to both validate and write the files.

benoutram (approved):

PR #3980, opened 2023-03-30
EES-2465 back button behaviour on table tool

amyb-hiveit:
Changes the behaviour when you click the browser back button while using  the table tool.

When I click the back button from:

- a featured table
  - if got there from step 2 > go back to step 2
  - if followed a direct link to the table > normal back button behaviour
- step 2 (subject)
  - if publication step was the initial step > go back to publication step
  - if subject step was the initial step (i.e. came from release page) > normal back button behaviour
- steps 3, 4, 5 & 6 (location, time, filters and final) > go back to step 2

## Notes:
- I ended up with a combination of the two approaches suggested on the ticket - it updates the url when you choose a publication and subject, and intercepts the back click using `beforePopState` to activate the correct step. I initially tried just using `beforePopState` but was ended up with various scenarios where it didn't quite work which adding the publication and subject to the url fixed.
- Currently when you choose a featured table and then go back to step 2 the featured table you chose isn't in the list any more. I've changed this to always show all featured tables after checking with Cam that it wasn't expected behaviour.
- I've removed the hash change on the subject tabs as it was causing problems and doesn't seem necessary.
- This only effects the table tool in the public frontend, not the admin.
- I wasn't sure of the best way to test this so have tagged a couple of basic tests on to an existing UI test. Better ideas are welcome.

## Bonus bug fix

Fixes [EES-4204](https://dfedigital.atlassian.net/browse/EES-4204) as I kept hitting it and it was simple to sort out.

### DIFF
--- a/src/explore-education-statistics-common/src/modules/table-tool/components/TableToolWizard.tsx
+++ b/src/explore-education-statistics-common/src/modules/table-tool/components/TableToolWizard.tsx
@@ -23,7 +23,10 @@ import applyTableHeadersOrder from '@common/modules/table-tool/utils/applyTableH
 import getDefaultTableHeaderConfig from '@common/modules/table-tool/utils/getDefaultTableHeadersConfig';
 import mapFullTable from '@common/modules/table-tool/utils/mapFullTable';
 import parseYearCodeTuple from '@common/modules/table-tool/utils/parseYearCodeTuple';
-import publicationService, { Theme } from '@common/services/publicationService';
+import publicationService, {
+  PublicationTreeSummary,
+  Theme,
+} from '@common/services/publicationService';
 import tableBuilderService, {
   FeaturedTable,
   ReleaseTableDataQuery,
@@ -64,35 +67,47 @@ export interface FinalStepRenderProps {
 }
 
 export interface TableToolWizardProps {
-  themeMeta?: Theme[];
-  initialState?: Partial<InitialTableToolState>;
-  hidePublicationSelectionStage?: boolean;
   finalStep?: (props: FinalStepRenderProps) => ReactElement;
+  hidePublicationStep?: boolean;
+  initialState?: Partial<InitialTableToolState>;
   loadingFastTrack?: boolean;
   renderFeaturedTable?: (featuredTable: FeaturedTable) => ReactNode;
   scrollOnMount?: boolean;
+  showTableQueryErrorDownload?: boolean;
+  themeMeta?: Theme[];
+  currentStep?: number;
+  onPublicationFormSubmit?: (publication: PublicationTreeSummary) => void;
+  onStepChange?: (nextStep: number, previousStep: number) => void;
+  onSubjectFormSubmit?(params: {
+    publication: SelectedPublication;
+    release: SelectedPublication['selectedRelease'];
+    subjectId: string;
+  }): void;
+  onSubjectStepBack?: (publication?: SelectedPublication) => void;
+  onSubmit?: (table: FullTable) => void;
   onTableQueryError?: (
     errorCode: TableQueryErrorCode,
     publicationTitle: string,
     subjectName: string,
   ) => void;
-  showTableQueryErrorDownload?: boolean;
-  onSubmit?: (table: FullTable) => void;
-  onSubjectStepBack?: () => void;
 }
 
 const TableToolWizard = ({
-  themeMeta = [],
-  initialState = {},
-  scrollOnMount,
-  hidePublicationSelectionStage,
-  renderFeaturedTable,
   finalStep,
-  showTableQueryErrorDownload = true,
-  onSubmit,
-  onSubjectStepBack,
-  onTableQueryError,
+  hidePublicationStep,
+  initialState = {},
   loadingFastTrack = false,
+  renderFeaturedTable,
+  scrollOnMount,
+  showTableQueryErrorDownload = true,
+  themeMeta = [],
+  currentStep,
+  onPublicationFormSubmit,
+  onStepChange,
+  onSubjectFormSubmit,
+  onSubjectStepBack,
+  onSubmit,
+  onTableQueryError,
 }: TableToolWizardProps) => {
   const router = useRouter();
   const [state, updateState] = useImmer<TableToolState>({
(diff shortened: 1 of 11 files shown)


luke-h1 (approved):

src/explore-education-statistics-frontend/src/modules/table-tool/TableToolPage.tsx
```diff
@@ -43,7 +44,38 @@ const TableToolPage: NextPage<TableToolPageProps> = ({
   subjectMeta,
   themeMeta,
 }) => {
+  const router = useRouter();
   const [loadingFastTrack, setLoadingFastTrack] = useState(false);
+  const [updateActiveStep, setUpdateActiveStep] = useState<number | undefined>(
+    undefined,
+  );
```
ntsim, line 51: Would suggest renaming this to `currentStep`

src/explore-education-statistics-frontend/src/modules/table-tool/TableToolPage.tsx
```diff
@@ -43,7 +44,38 @@ const TableToolPage: NextPage<TableToolPageProps> = ({
...
+      ) {
+        setUpdateActiveStep(2);
+      }
+
+      return true;
+    });
+
+    return () => {
+      router.beforePopState(() => true);
+    };
```
ntsim, line 77: Do we actually need this? This router API isn't like a typical event handler where we can unregister callbacks, so I'm not sure this is actually doing anything?

src/explore-education-statistics-frontend/src/modules/table-tool/TableToolPage.tsx
```diff
@@ -43,7 +44,38 @@ const TableToolPage: NextPage<TableToolPageProps> = ({
...
+        setUpdateActiveStep(1);
+      }
+
+      // clicking back on any step after step 2 should take you to step 2
+      if (
+        url === '/data-tables/[publicationSlug]' ||
+        url === '/data-tables/[publicationSlug]/[releaseSlug]' ||
+        // when have gone directly to /data-tables/publication-slug the url has
+        // publicationSlug in the query as well so need to account for that.
+        url.startsWith('/data-tables/[publicationSlug]?publicationSlug=')
```
ntsim, line 67: Can this all be simplified to:
```ts
if (url.startsWith('/data-tables/[publicationSlug]')) {
```

ntsim (changes requested):

src/explore-education-statistics-common/src/modules/table-tool/components/TableToolWizard.tsx
```diff
@@ -386,7 +408,9 @@ const TableToolWizard = ({
             scrollOnMount={scrollOnMount}
             initialStep={state.initialStep}
             id="tableToolWizard"
+            updateActiveStep={updateActiveStep}
             onStepChange={async (nextStep, previousStep) => {
+              onStepChange?.();
```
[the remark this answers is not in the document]
amyb-hiveit: I'm not following why we'd want to pass through nextStep and previousStep through when they wouldn't be used?

src/explore-education-statistics-frontend/src/modules/table-tool/TableToolPage.tsx
```diff
@@ -189,6 +192,23 @@ const TableToolPage: NextPage<TableToolPageProps> = ({
...
+        onStepChange={() => setUpdateActiveStep(undefined)}
+        onSubjectFormSubmit={(publicationSlug, releaseSlug, subjectId) => {
+          router.push(
+            `/data-tables/${publicationSlug}/${releaseSlug}/${subjectId}`,
+            undefined,
+            {
+              shallow: true,
+              scroll: false,
+            },
+          );
```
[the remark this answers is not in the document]
amyb-hiveit: Good catch, that's fixed now.

src/explore-education-statistics-frontend/src/modules/table-tool/TableToolPage.tsx
amyb-hiveit: You're right, it's not... I've removed it.

src/explore-education-statistics-common/src/modules/table-tool/components/TableToolWizard.tsx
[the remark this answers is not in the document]
ntsim: Given that `onStepChange` originates from `Wizard`, it makes sense to have a similar signature to this by just passing through the `nextStep` and `previousStep` args.

Whilst we don't have a use case for it currently, it's not difficult to add this functionality and makes sense.

src/explore-education-statistics-common/src/modules/table-tool/components/Wizard.tsx
```diff
@@ -43,6 +43,7 @@ interface Props {
   initialStep?: number;
   id: string;
   scrollOnMount?: boolean;
+  updateActiveStep?: number;
```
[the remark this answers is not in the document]
ntsim: Okay, would still suggest that we rename this to `currentStep` (as suggested in another comment as well)

src/explore-education-statistics-common/src/modules/table-tool/components/TableToolWizard.tsx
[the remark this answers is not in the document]
amyb-hiveit: ok, that's done now.

src/explore-education-statistics-common/src/modules/table-tool/components/Wizard.tsx
```diff
@@ -59,7 +61,7 @@ const Wizard = ({
   const mountedRef = useMountedRef();
 
   const [shouldScroll, setShouldScroll] = useState(scrollOnMount);
-  const [currentStep, setCurrentStepState] = useState(initialStep);
+  const [activeStep, setActiveStepState] = useState(initialStep);
```
ntsim, line 64: Minor, but if we're renaming this to `activeStep`, then the setter could be changed to `setActiveStep` too

amyb-hiveit: ok, renamed now.

ntsim (approved):

PR #3983, opened 2023-03-31
EES-4201 Retrieve and store Permalink statistics

benoutram:
This PR alters a migration originally written for [EES-3751 - Create new content database Permalink entity and store high-level info about Permalinks in database](https://dfedigital.atlassian.net/browse/EES-3751) in #3529 which has already been run successfully in all environments. After that we now track all Permalinks in the database.

This PR adapts that migration code to retrieve and store additional statistics:

- Count of filter items
- Count of footnotes
- Count of indicators
- Count of locations
- Count of observations
- Count of time periods
- Blob content length (bytes)
- Presence of configuration table headers (bool)

This will help us correlate previous information discovered about Permalinks failing to render due to errors and timeouts. These errors and timeouts are thought to be related to their size and the absence of attributes like missing indicators and configuration table headers.

We can use this information in future to exclude certain Permalinks from the migration to Permalink Snapshots planned by [EES-3754](https://dfedigital.atlassian.net/browse/EES-3754).

It will also help us to make an informed decision on if Permalink Snapshots should be stored in a compressed format in blob storage.

This PR also makes a change to store the statistics for every new Permalink that's created in future.

### Running the migration
The migration is triggered by running the secured BAU endpoint `PATCH https://{adminUrl}/api/bau/migrate-permalinks`.

The migration is re-runnable - the new columns can be cleared and the migration can be re-run if necessary.

### Other changes

- Rename `PermalinkCreateViewModel` to `PermalinkCreateRequest`, and move it to the `Data.Api.Requests` package.
- Remove unused `Content.Api` project dependency on `Publisher.Model`.
- Add `TODO`'s highlighting the deprecated `Microsoft.Azure.Storage.Blob` SDK related to [EES-4202](https://dfedigital.atlassian.net/browse/EES-4202).


src/GovUk.Education.ExploreEducationStatistics.Data.Api/GovUk.Education.ExploreEducationStatistics.Data.Api.csproj
```diff
@@ -19,7 +19,6 @@
     <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="7.0.0" />
     <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.18.0" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="6.0.2" />
-    <PackageReference Include="Microsoft.Azure.Storage.Blob" Version="11.2.3" />
```
luke-h1, line 22: Interesting, assume we've never used blob storage directly in the data API project?

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
```diff
@@ -210,7 +213,7 @@ private async Task WriteCsvHeaderRow(CsvWriter csv, PermalinkCsvMetaViewModel me
             try
             {
                 var text = await _blobStorageService.DownloadBlobText(
-                    containerName: Permalinks,
+                    containerName: BlobContainers.Permalinks,
```
luke-h1, line 216: Just wondering why we've removed the static import in favor of specifying the container name as `BlobContainers.Permalinks` ?

luke-h1 (approved):

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
```diff
@@ -258,15 +267,45 @@ private async Task WriteCsvHeaderRow(CsvWriter csv, PermalinkCsvMetaViewModel me
                         permalinkTableResult,
                         request.Query);
 
-                    await _blobStorageService.UploadAsJson(containerName: Permalinks,
+                    var content = JsonConvert.SerializeObject(legacyPermalink,
+                        BuildJsonSerializerSettings());
+
+                    await using var stream = new MemoryStream(Encoding.UTF8.GetBytes(content));
+                    permalink.LegacyContentLength = stream.Length;
```
ntsim, line 274: Interesting approach. Have we done this to avoid making a second request for the blob metadata?

src/GovUk.Education.ExploreEducationStatistics.Data.Api/GovUk.Education.ExploreEducationStatistics.Data.Api.csproj
benoutram: We do, for example storing and retrieving Permalinks, but that code is using the newer v12 SDK `Azure.Storage.Blobs` which is a dependency of `ExploreEducationStatistics.Common` that Data.Api depends on.

I've removed all references to the deprecated `Microsoft.Azure.Storage.Blob` but don't totally get rid of it just yet.

Our `Common.Services.BlobStorageService` has a dependency on `TransferManager` which belongs 
 to `Microsoft.Azure.Storage.DataMovement`. But that in turn depends on the deprecated `Microsoft.Azure.Storage.Blob`.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
benoutram: > Have we done this to avoid making a second request for the blob metadata?

Yes, whether we call `UploadStream`, or `UploadAsJson` which is turned into text and then to a stream, the underlying call used is `blob.UploadAsync` which returns `Response<BlobContentInfo>` but doesn't contain the length of the uploaded content.

> We seem to be okay getting the blob metadata as part of the migration, so just wondering why the inconsistent approach?

I thought it was ok to do it in the migration because we already get the blob and then just one extra call to get the blob properties and with it only being the migration it's code that's only going to run once under supervision and then be thrown away.

Here when creating a Permalink we upload the blob to begin with, then I think we'd need to get the blob again and then a further call to get the blob properties. It seems like that's adding unnecessary steps that could fail and result in us having a blob that's uploaded but not saved in the database.

I can work on trying to make this a bit nicer. I'm open to suggestions. Were you thinking I should change the `BlobStorageService` to try and return some kind of result object with all uploads? 🙂

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
luke-h1: Ah interesting, yeah I do agree it removes the ambiguity of where stuff is getting imported from.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
ntsim: Fair enough, it doesn't seem like a problem to keep it the way it is. 

Personally, I think I would have just gone with two requests - one to upload it first, then another to fetch the properties afterwards. It doesn't really matter though as long as the content length from the stream accurately reflects what gets uploaded to blob storage!

ntsim (approved):

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
benoutram: Thanks for reviewing. I did a check during some testing and the content length was identical to the content length reported in my local storage.

PR #3988, opened 2023-04-03
EES-4084 add text alignment option to ckeditor

amyb-hiveit:
Adds a text alignment option to the CKEditor toolbar.

To avoid stripping out the text-align styles I've had to update santize-html and change its options to allow any styles on any elements as specifying them using `allowedStyles` doesn't work in the browser so analysts wouldn't have been able to tell that the alignment was being applied (the previous setting to only allow `width` on figures worked on the frontend as it's server side rendered but not in the admin).

I've added some extra styling to compensate for it adding `p` tags with the style inside table cells.


src/explore-education-statistics-admin/src/config/ckEditorConfig.ts
```diff
@@ -104,3 +105,5 @@ export const tableContentToolbar: string[] = [
   'tableRow',
   'mergeTableCells',
 ];
+
+export const alignmentOptions: string[] = ['left', 'right', 'center'];
```
luke-h1, line 109: Just wondering if you've thought about potentially narrowing this type? I believe the following is the correct interface: 
```
AlignmentConfig
```

src/explore-education-statistics-admin/src/hooks/useCKEditorConfig.ts
```diff
@@ -176,6 +177,7 @@ const useCKEditorConfig = ({
             waitingTime: 5000,
           }
         : undefined,
+      alignment: { options: alignmentOptions },
```
luke-h1, line 180: Might be good to define this in the `EditorConfig` interface

src/explore-education-statistics-admin/src/config/ckEditorConfig.ts
amyb-hiveit: Good idea done now.

src/explore-education-statistics-admin/src/types/ckeditor.ts
```diff
@@ -34,6 +34,15 @@ export interface AutoSavePluginConfig {
   waitingTime: number;
 }
 
+export interface AlignmentFormat {
+  className?: string;
+  name: 'left' | 'right' | 'center' | 'justify';
+}
+
+export interface AlignmentConfig {
+  options: Array<AlignmentFormat['name'] | AlignmentFormat>;
```
luke-h1, line 43: Small nitpick: We typically specify arrays with the `[]` syntax instead of using `Array`

PR #3992, opened 2023-04-05
EES-4225 Fix exceptions streaming data file blob in `GetCsvStream` methods

benoutram:
This PR fixes exceptions occurring streaming the data file blob in `SubjectCsvMetaService` and `PermalinkCsvMetaService` where within the `GetCsvStream` methods the `ReleaseFile` predicate was not constrained by type `FileType.Data`.

This led to the exception `Sequence contains more than one element.` due to the existence of the other `ReleaseFile` of type `FileType.Metadata`.

There was also an exception occurring streaming the blob using `releaseFile.PublicPath()` due to its reliance on the `releaseFile.Release` being loaded. This is changed to only rely on `releaseFile.ReleaseId`.

### Other changes

- Fix exception logging which was missing the `SubjectId`

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkCsvMetaService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkCsvMetaService.cs
@@ -94,6 +94,7 @@ public class PermalinkCsvMetaService : IPermalinkCsvMetaService
                 .SingleAsync(
                     predicate: rf =>
                         rf.File.SubjectId == releaseSubject.SubjectId
+                        && rf.File.Type == FileType.Data
                         && rf.ReleaseId == releaseSubject.ReleaseId,
                     cancellationToken: cancellationToken
                 );
@@ -105,7 +106,7 @@ public class PermalinkCsvMetaService : IPermalinkCsvMetaService
             _logger.LogError(
                 exception,
                 message: "Could not get file for release subject (ReleaseId = {ReleaseId}, SubjectId = {SubjectId})",
-                releaseSubject.ReleaseId, releaseSubject.Subject);
+                releaseSubject.ReleaseId, releaseSubject.SubjectId);
 
             return null;
         }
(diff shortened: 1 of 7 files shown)


src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkCsvMetaService.cs
```diff
@@ -94,6 +94,7 @@ public class PermalinkCsvMetaService : IPermalinkCsvMetaService
                 .SingleAsync(
                     predicate: rf =>
                         rf.File.SubjectId == releaseSubject.SubjectId
+                        && rf.File.Type == FileType.Data
```
ntsim, line 97: Might be worth having a test with a data and meta file to check that this doesn't regress?

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectCsvMetaService.cs
```diff
@@ -89,6 +89,7 @@ public class SubjectCsvMetaService : ISubjectCsvMetaService
                 .SingleAsync(
                     predicate: rf =>
                         rf.File.SubjectId == releaseSubject.SubjectId
+                        && rf.File.Type == FileType.Data
```
ntsim, line 92: Might be worth having a test with a data and meta file to check that this doesn't regress?

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkCsvMetaService.cs
benoutram: Good point, added one extra test `GetCsvMeta_SubjectExists_MultipleReleaseFilesExist` which sets up the metadata file and also a data file for a different subject.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/SubjectCsvMetaService.cs
benoutram: Added one extra test `GetSubjectCsvMeta_Headers_MultipleReleaseFilesExist`

PR #4002, opened 2023-04-14
EES-3754 Add Permalink snapshot migration endpoint

benoutram:
⚠️ Depends on [EES-3753 - BE work required to support new Permalinks with table snapshots](https://dfedigital.atlassian.net/browse/EES-3753) added by #3966.

This PR adds a Permalink snapshot migration endpoint to the Data.Api's `PermalinkController` and also a Python script allowing this endpoint to be called for a list of Permalinks id's.

### Data API migration endpoint

It has this workflow:

- Finds the legacy permalink in the database with the specified id.
- Ensures the snapshot doesn't exist.
- Downloads and deserializes the legacy permalink JSON from blob storage.
- Generates the universal table JSON for the legacy permalink.
- Generates the CSV metadata for the legacy permalink.
- Uploads the table CSV and universal table JSON to the Permalink snapshots container in blob storage.
- Flags the legacy permalink as having a snapshot so that it's accessible via the new routes and can't be migrated again.

Most of this relies on changes already pushed down to #3966 and services already written for the Permalink snapshots epic.

#### Usage

`PUT /api/permalink/{permalinkId}/snapshot`

#### Returns

`200 Success`  - The permalink snapshot is created successfully.
`400 Bad request` - The legacy permalink has already been migrated to have a snapshot.
`404 Not found` - No permalink found for the specified id or it is not a legacy permalink.

### Python script to migrate all permalinks

This script uses the Data API migration endpoint, reading from a CSV file named `permalinks.csv` and outputs a CSV file named `migration_results.csv`.

#### Usage

`pipenv run python migrate_legacy_permalinks.py`

The input file can be produced by querying the database for the set of legacy permalinks:

```
SELECT LOWER(Id) AS permalink_id,
       IIF(LegacyHasSnapshot = 1, 'True', 'False') AS migrated
FROM Permalinks
WHERE Legacy = 1
ORDER BY Created
```

Full instructions can be found within the script.


src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
```diff
@@ -198,6 +200,43 @@ public class PermalinkService : IPermalinkService
...
+            CancellationToken cancellationToken = default)
+        {
+            return await _contentDbContext.Permalinks
+                .SingleOrNotFoundAsync(
+                    predicate: permalink => permalink.Id == permalinkId && permalink.Legacy,
+                    cancellationToken: cancellationToken)
+                .OnSuccess(async permalink =>
+                {
+                    // Return a bad request result if the Permalink already has a snapshot
+                    if (permalink.LegacyHasSnapshot == true)
```
markhiveit, line 214: Thought I'd caught you doing the old unnecessary `== true` there, until I realised `LegacyHasSnapshot` is nullable :)

useful-scripts/permalink-snapshot-migration/migrate_legacy_permalinks.py
```diff
@@ -0,0 +1,114 @@
...
+Instructions:
+
+1. List all the legacy permalinks by running a query against the Content database to get all the legacy permalinks:
+
+```
+SELECT LOWER(Id) AS permalink_id,
+       IIF(LegacyHasSnapshot = 1, 'True', 'False') AS migrated
+FROM content.dbo.Permalinks
+WHERE Legacy = 1
+ORDER BY Created
```
markhiveit, line 24: Minor point that frankly can probably be ignored, but this will mean we migrate the oldest permalink first? Not sure how long the entire migration will take, but if it'll take a while we could `order by desc` so that we migrate those permalinks most likely to be used first.

markhiveit (approved):

useful-scripts/permalink-snapshot-migration/migrate_legacy_permalinks.py
```diff
@@ -0,0 +1,114 @@
...
+
+Usage: `pipenv run python migrate_legacy_permalinks.py`
+
+Instructions:
+
+1. List all the legacy permalinks by running a query against the Content database to get all the legacy permalinks:
+
+```
+SELECT LOWER(Id) AS permalink_id,
+       IIF(LegacyHasSnapshot = 1, 'True', 'False') AS migrated
```
luke-h1, line 21: Just wondering if this is a typo? `IIF`

useful-scripts/permalink-snapshot-migration/migrate_legacy_permalinks.py
```diff
@@ -0,0 +1,114 @@
...
+        with open("permalinks.csv", "r") as permalinks_csv_file:
+            csv_reader = csv.reader(permalinks_csv_file)
+            permalinks = list(csv_reader)
+
+        self._migrate_permalinks(permalinks, timeout=240, sleep=5)
+
+
+if __name__ == "__main__":
+    ap = argparse.ArgumentParser(
+        prog=f"python {os.path.basename(__file__)}", description="To migrate legacy permalinks to permalink snapshots"
```
luke-h1, line 98: Would suggest prefixing this with `pipenv`

useful-scripts/permalink-snapshot-migration/migrate_legacy_permalinks.py
benoutram: The front-end will have a feature toggle that won't be removed until we've migrated all permalinks and verified that migration by comparing before/after results using both routes.

I don't think there's a need to make sure we migrate those likely to be used first as we are trying to achieve 100% through identifying ones we think are problematic early through other tickets and addressing their issues or deleting them.

When we come to actually run it we might not care about any migration order, or might want to migrate them in size order instead.

The idea is that they will all be migrated by the time the old routes are removed.

useful-scripts/permalink-snapshot-migration/migrate_legacy_permalinks.py
benoutram: It's not a typo, it's shorthand for a case statement, see [IIF](https://learn.microsoft.com/en-us/sql/t-sql/functions/logical-functions-iif-transact-sql?view=sql-server-ver16).

useful-scripts/permalink-snapshot-migration/migrate_legacy_permalinks.py
```diff
@@ -0,0 +1,114 @@
...
+
+"""
+
+
+class MigrateLegacyPermalinks:
+    def __init__(self, data_api_url: str):
+        self.data_api_url = data_api_url
+        self.http_headers = {}
+
+    def _migrate_permalink(self, permalink_id, timeout):
```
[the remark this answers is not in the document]
benoutram: Added type hints and a bunch of other changes in 823bb9d.

PR #4004, opened 2023-04-18
EES-4193, EES-4221 - make improvements to snapshot automation

luke-h1:
This PR makes several improvements to how automation around snapshots tests is managed & performed. 

Relevant changes:
* Changes commit author in `validate-snapshots.yml` so that I don't get included as a commit author
* Adds labels to the pull request created by the bot
* Refactors `_send_slack_notification` to allow a user to pass an optional slack block in order to enable the creation of more richly formatted messages (see screenshot for an example)

Screenshot 2023-04-18 at 11 09 57

Other changes:
* Runs python linter/formatter on all python related files

### DIFF
--- a/tests/robot-tests/scripts/create_snapshots.py
+++ b/tests/robot-tests/scripts/create_snapshots.py
@@ -3,6 +3,7 @@ import base64
 import json
 import math
 import os
+from typing import Optional
 
 import requests
 from bs4 import BeautifulSoup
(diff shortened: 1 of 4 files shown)


tests/robot-tests/scripts/create_snapshots.py
```diff
@@ -219,18 +221,57 @@ def _create_methodologies_snapshot(self) -> str:
...
+                        {
+                            "type": "button",
+                            "style": "danger",
+                            "text": {"type": "plain_text", "text": "View pull request"},
+                            "url": "https://github.com/dfe-analytical-services/explore-education-statistics/pulls/dfe-sdt",
+                        }
+                    ],
+                }
+            )
+        return blocks if len(blocks) > 0 else None
```
markhiveit, line 252: I think this function can be removed, as it is only called once, and `len(messages) > 0` has already been checked when it's called. So instead on L266 you could have this instead:

```
self._send_slack_notification(blocks=[{
  "type": "header",
  ...
},
{
  "type": "actions",
  ...
}]
```

tests/robot-tests/scripts/create_snapshots.py
```diff
@@ -219,18 +221,57 @@ def _create_methodologies_snapshot(self) -> str:
...
         for name in self.snapshots:
             if self._validate_snapshot(name):
                 print(f"Snapshot {name} matches for {self.public_url}")
             else:
                 print(f"Snapshot {name} does not match for {self.public_url}")
-                if self.slack_webhook_url is not None:
-                    self._send_slack_notification(f"Snapshot {name} does not match for {self.public_url}")
-        if self.slack_webhook_url is not None:
-            self._send_slack_notification(f"Snapshot test workflow successfully completed for {self.public_url} ✅")
+                messages.append(f"Snapshot `{name}` does not match for {self.public_url}")
```
markhiveit, line 262: I think `messages` could be a boolean - I don't believe the string appended to the `messages` list is ever being used.

So instead this could be something like
```
failed: bool = False

for name in self.snapshots:
  if self._validate_snapshot(name):
    ...
  else:
    print(...)
    failed = True

if failed and self.slack_webhook_url:
  ...
```

tests/robot-tests/scripts/create_snapshots.py
```diff
@@ -219,18 +221,57 @@ def _create_methodologies_snapshot(self) -> str:
...
+
+        if len(messages) > 0 and self.slack_webhook_url:
+            # At least one snapshot did not match so send a failure message to slack
+            self._send_slack_notification(blocks=self._build_slack_blocks(messages))
 
-    def write_snapshots_to_file(self):
+        else:
```
markhiveit, line 268: This `else` is going to be executed if there are differences, but `self.slack_webhook_url` is `None`. So even though there are failures, it's  going to print "Snapshot script complete. No differences found."

I think I'd do this instead:
```
if len(messages) > 0:
  self._send_slack_notification(blocks=self._build_slack_blocks(messages)) if self.slack_webhook_url else print("waah! snapshots don't match!")
else:
  ...
```

tests/robot-tests/scripts/create_snapshots.py
```diff
@@ -45,27 +46,28 @@ def __init__(self, public_url: str, slack_webhook_url: str):
...
-        else:
-            raise Exception(f"Invalid snapshot name: {name}")
+        switcher = {
+            "data_catalogue_snapshot.json": self._create_data_catalogue_snapshot,
+            "find_statistics_snapshot.json": self._create_find_statistics_snapshot,
+            "methodologies_snapshot.json": self._create_methodologies_snapshot,
+            "table_tool_snapshot.json": self._create_table_tool_snapshot,
+        }
+
+        return switcher.get(name, lambda: "Invalid snapshot name")
```
markhiveit, line 60: I think that this returns a string rather than a function could be confusing when trying to debug this, if someone tries to access a snapshot function that doesn't exist.

I think I'd prefer for this to fail immediately if that happens. Maybe:
```
result = switcher.get(name)
if result is None:
  print(f"waaah! no such snapshot function for '{name}'!")
  os.exit(1)
return result
```

markhiveit (changes requested):

markhiveit (approved):

PR #4005, opened 2023-04-18
EES-3858 custom data groupings

amyb-hiveit:
**Do not merge until https://github.com/dfe-analytical-services/explore-education-statistics/pull/4010 has been merged**

Adds the ability to set data groupings for maps per data set.

For each data set you can set the groupings to:
- a number of equal intervals
- a number of quantiles
- custom groups
- or copy the custom groups from another data set

For custom groups:
- groups cannot overlap
- groups don't have to cover the full range of data
- the unit shows if it's set and the datasets don't have different units
- if you copy custom groups from another data set they are not linked and can be edited independently 

Also:
- We're not planning on migrating existing config for groupings so this is backwards compatible with them.
- Boundary levels are now set on their own tab rather than the same one as data groupings. This is because in future they will also be set per data set which would make having them both in the same tab messy.

customdg1

customdg2


src/explore-education-statistics-common/src/modules/charts/util/getDataSetCategoryConfigs.ts
```diff
@@ -12,30 +12,67 @@ import expandDataSet from '@common/modules/charts/util/expandDataSet';
...
 import keyBy from 'lodash/keyBy';
 import omit from 'lodash/omit';
 import uniqBy from 'lodash/uniqBy';
+import {
+  DataGrouping,
+  DataGroupingConfiguration,
+  DataGroupingType,
+} from '../types/chart';
+
+const defaultDataGrouping: OmitStrict<DataGrouping, 'dataSet'> = {
```
ntsim, line 25: Shouldn't the type just be `DataGroupingConfiguration`?

src/explore-education-statistics-common/src/modules/charts/types/chart.ts
```diff
@@ -72,13 +72,30 @@ export type AxesConfiguration = {
...
 
+export interface DataGroupingConfiguration {
+  type: DataGroupingType;
+  numberOfGroups: number;
+  customGroups: CustomDataGroup[];
+}
+
+export interface DataGrouping extends DataGroupingConfiguration {
+  dataSet: DataSet;
+  unit?: string;
```
ntsim, line 96: Seems like a bit of a potential bug to have this here as we don't actually want to store `unit` in the backend and we only really need this for the chart builder UI components. Consequently, this is a bit leaky and implies the wrong things to consumers of this type.

Would suggest removing this and either:

- Refactoring the type into one of the components and export it from there e.g. `DataGroupingWithUnit`
- Refactor the components to accept a `unit` prop. Looks like we already tried to do this (see `ChartDataGroupingsForm`) so it'd just be a continuation of that.

This means that in cases where we're currently doing `OmitStrict<DataGrouping, 'dataSet'>`, we can just replace this with `DataGroupingConfiguration`

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartDataGroupingsForm.tsx
```diff
@@ -0,0 +1,230 @@
...
+  copyCustomGroups?: string;
+  type: DataGroupingType | 'CopyCustom';
+}
+
+interface Props {
+  dataGrouping: DataGrouping;
+  dataSetsWithCustomGroups: DataGrouping[];
+  id: string;
+  meta: FullTableMeta;
+  unit?: string;
```
ntsim, line 33: This is currently unused (so could be removed), but as mentioned in another comment, we could actually implement this to clean up the `DataGrouping` type (which unnecessarily has `unit` as a property).

For example, one way we could do this:

```tsx
// ChartDataGroupingsConfiguration.tsx

interface DataGroupingWithUnit extends DataGrouping {
  unit: string;
}

interface FormValues {
  dataGroupings: DataGroupingWithUnit[];
}

const ChartDataGroupingsConfiguration = (...) => {
  const [editGrouping, setEditGrouping] = useState<DataGroupingWithUnit>();
  
  return (
    <ChartDataGroupingsForm
      dataGrouping={editGrouping}
      unit={editGrouping.unit} 
    />
  );
}
```

src/explore-education-statistics-common/src/styles/_form.scss
```diff
@@ -6,3 +6,7 @@
 .govuk-textarea[disabled] {
   cursor: not-allowed;
 }
+
+.ReactModalPortal .govuk-radios__conditional--hidden {
+  display: none;
+}
```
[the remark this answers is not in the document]
amyb-hiveit: It's because the hidden class relies on the `js-enabled` being present on the root element, the modal is outside the root element so doesn't inherit the class.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartDataGroupingsConfiguration.tsx
```diff
@@ -0,0 +1,208 @@
...
+                meta={meta}
+                onCancel={() => setEditGrouping(undefined)}
+                onSubmit={values => {
+                  const updated = form.values.dataGroupings.map(group => {
+                    if (isEqual(group.dataSet, values.dataSet)) {
+                      return values;
+                    }
+                    return group;
+                  });
+                  form.setFieldValue('dataGroupings', updated);
```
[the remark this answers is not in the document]
amyb-hiveit: I've had to leave the current defaults in for `dataClassification` and `dataGroups` in MapBlock as couldn't see a simple way to use the `enableCustomDataGroupings` flag there. These mean it always think there's a deprecated classification so don't update the map with the new values. Once it's not under a flag I can remove these.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/__tests__/ChartDataGroupingsConfiguration.test.tsx
```diff
@@ -0,0 +1,1162 @@
...
+    await waitFor(() => {
+      expect(modal.getByText('There is a problem')).toBeInTheDocument();
+    });
+
+    expect(
+      screen.getByRole('link', { name: 'Add one or more custom groups' }),
+    ).toHaveAttribute('href', '#data-groupings-customGroups');
+  });
+
+  test('shows a validation error when do not select a data set to copy custom groups from', async () => {
```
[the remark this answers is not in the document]
ntsim: Doesn't look like this has been changed?

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartLegendConfiguration.tsx
```diff
@@ -354,6 +354,7 @@ const ChartLegendConfiguration = ({
                         <div className="dfe-flex dfe-justify-content--space-between">
                           <div className={styles.labelInput}>
                             <FormFieldTextInput
+                              addOnContainerClassName="dfe-flex-grow--1"
```
ntsim, line 357: Feels like it'd be sensible to add `dfe-flex-grow--1` as one of the conditional classes in `FormBaseInput`:

```tsx
<div
  className={classNames(
    {
      'dfe-flex': !!addOn,
      'dfe-flex-grow--1': !addOn,
    },
    addOnContainerClassName,
  )}
>
```

However, I'm also wondering if we could go further with this and just remove the container div if `addOn` is present or not e.g.

```tsx
// FormBaseInput.tsx

const input = (
  <input
    // rest of props
  />
);

return (
  <>
    // ... some stuff
    {addOn ? (
      <div className={classNames('dfe-flex', addOnContainerClassName)}>
        {input}
        {addOn}
      </div>
    ) : (
      input
    )}
  </>
);

```

ntsim (changes requested):

src/explore-education-statistics-common/src/modules/charts/types/chart.ts
```diff
@@ -91,6 +112,7 @@ export interface ChartProps {
   legend?: LegendConfiguration;
   includeNonNumericData?: boolean;
   showDataLabels?: boolean;
+  map?: MapConfiguration;
```
ntsim, line 115: Wouldn't put this on `ChartProps` as it's specific to maps. We could just modify the props for `MapBlockInternal`:

```ts
export interface MapBlockProps extends ChartProps {
  // ...
  map: MapConfiguration;
}
```

Then in `ChartBuilder` we would just need to fix the `chartProps`:

```ts
const chartProps = useMemo<ChartBuilderChartProps | undefined>(() => {
  // other stuff ...
  switch (definition.type) {
    case 'map':
      return {
        ...(baseProps as MapBlockProps),
        map: map ?? {
          dataSetConfigs: [],
        },
        boundaryLevel: options.boundaryLevel ?? 0,
        type: 'map',
      };
  }
}, [...]);
```

src/explore-education-statistics-common/src/modules/charts/components/MapBlockInternal.tsx
```diff
@@ -60,9 +59,8 @@ export interface MapBlockProps extends ChartProps {
     major: AxisConfiguration;
   };
   boundaryLevel?: number;
-  customDataGroups?: CustomDataGroup[];
   dataGroups?: number;
-  dataClassification?: DataClassification;
+  dataClassification?: DataGroupingType;
```
ntsim, line 63: Would suggest creating a ticket for removing these and migrating over to the new map `dataSetConfigs`.

Should have comments referencing this ticket as well.

markhiveit: I created the ticket to reference on another PR - the description still needs filing out.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartDataGroupingsConfiguration.tsx
```diff
@@ -0,0 +1,216 @@
...
+}: Props) => {
+  const [editDataSetConfig, setEditDataSetConfig] = useState<{
+    dataSetConfig: MapDataSetConfiguration;
+    unit: string;
+  }>();
+  const { updateForm, submitForms } = useChartBuilderFormsContext();
+
+  const handleOnMount = useCallback(
+    values => {
+      return values.dataSetConfigs.length > 0 && map.dataSetConfigs.length === 0
```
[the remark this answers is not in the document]
amyb-hiveit: yeah it's not needed, have removed it.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartDataGroupingsConfiguration.tsx
```diff
@@ -0,0 +1,216 @@
...
+        ? onChange(values.dataSetConfigs)
+        : undefined;
+    },
+    [map.dataSetConfigs?.length, onChange],
+  );
+
+  const initialValues = useMemo<FormValues>(() => {
+    if (map.dataSetConfigs.length) {
+      return { dataSetConfigs: map.dataSetConfigs };
```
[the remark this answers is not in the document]
amyb-hiveit: Fixed now.

amyb-hiveit:
@ntsim this is updated now to address your comments and remove the feature flag.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartDataGroupingsConfiguration.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+  const handleChange = useCallback(
+    (values: FormValues) => onChange(values.dataSetConfigs),
+    [onChange],
+  );
+
+  const handleMount = useCallback(
+    (values: FormValues) => onChange(values.dataSetConfigs),
+    [onChange],
+  );
```
ntsim, line 70: Not a dealbreaker but could just re-use `handleChange` for the Effect's `onMount`

tests/robot-tests/tests/admin/bau/create_data_block_with_chart.robot
```diff
@@ -689,9 +689,53 @@ Validate basic geographic chart preview
 
     user checks map chart indicator tile contains    id:chartBuilderPreview    Admissions in 2005    3,612
 
+Change geographic chart data groupings
```
ntsim, line 692: Wouldn't it be better to have this before `Validate basic geographic chart preview`?

Feels like we'd want to add some extra checks for the legend and data grouping changes to ensure they look correct in the preview.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartBoundaryLevelsConfiguration.tsx
```diff
@@ -0,0 +1,116 @@
...
+    },
+    [normalizeValues, onChange],
+  );
+
+  return (
+    <Formik<FormValues>
+      enableReinitialize
+      initialValues={{ boundaryLevel: options.boundaryLevel }}
+      validateOnMount
+      validationSchema={Yup.object<FormValues>().shape({
```
ntsim, line 63: Don't need the `shape` call, can just be:

```ts
Yup.object<FormValues>({
  ...
});
```

tests/robot-tests/tests/admin/bau/create_data_block_with_chart.robot
amyb-hiveit: I've added tests for checking the data groupings are shown correctly and update now, had to change some of the locations used in the chart because of lack of data for the ones it was using before.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartBoundaryLevelsConfiguration.tsx
```diff
@@ -0,0 +1,122 @@
...
+        onSubmit(normalizeValues(values));
+        await submitForms();
+      }}
+    >
+      {form => (
+        <Form id={formId}>
+          <Effect
+            value={{
+              ...form.values,
+            }}
```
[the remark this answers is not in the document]
duncan-at-hiveit: I noticed that you went with explicitly picking which properties you were carrying across here:

```
          <Effect
            value={{
              formKey: 'boundaryLevels',
              isValid: form.isValid,
              submitCount: form.submitCount,
            }}
```

Was that preferable to the above suggestion of `value={form.values}`?

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartDataGroupingsConfiguration.tsx
```diff
@@ -0,0 +1,208 @@
...
+            </ChartBuilderSaveActions>
+          </Form>
+
+          {editGrouping && (
+            <Modal title="Edit groupings">
+              <ChartDataGroupingsForm
+                dataGrouping={editGrouping}
+                dataSetsWithCustomGroups={form.values.dataGroupings.filter(
+                  grouping => grouping.type === 'Custom',
+                )}
```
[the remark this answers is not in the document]
duncan-at-hiveit: I see `dataSetsWithCustomGroupsOptions` is a memoised version of the above.  Are you happy this comment is addressed with this?

src/explore-education-statistics-common/src/modules/charts/util/getDataSetCategoryConfigs.ts
```diff
@@ -66,15 +103,38 @@ export default function getDataSetCategoryConfigs(
...
+          dataGroupingItemsByDataSet[dataSetKey],
+          'dataSet',
+        );
+
         return {
           config,
           dataSet,
           rawDataSet,
           dataKey: dataSetKey,
+          grouping: getGrouping({ deprecatedGrouping, dataGrouping }),
```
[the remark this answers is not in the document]
duncan-at-hiveit: You decided not to go with the `omit`?

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartDataGroupingsForm.tsx
duncan-at-hiveit: Couldn't easily see if this was resolved with a new interface - was it handled another way?

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartBoundaryLevelsConfiguration.tsx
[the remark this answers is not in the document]
amyb-hiveit: This was referring to the `Effect` above this one, now on line 75. I followed Nick's suggestion and changed it to `<Effect value={form.values} onChange={handleChange} />`. I'm not sure why it was spread initially, it was copied over from the previous version of the component (`ChartMapConfiguration.tsx `)

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartDataGroupingsConfiguration.tsx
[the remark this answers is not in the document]
amyb-hiveit: The feature flag is removed now so this is no longer a problem.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartDataGroupingsConfiguration.tsx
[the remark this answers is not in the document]
amyb-hiveit: It now passes all the `dataSetConfigs` (fka `dataGroupings`) to `ChartDataGroupingForm` which does the filtering and memoizes the result.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartDataGroupingsForm.tsx
amyb-hiveit: yep, I changed it to get the unit in `ChartDataGroupingsConfiguration` and pass it down to `ChartDataGroupingForm` rather than add it to the `DataGrouping` type.

duncan-at-hiveit (approved):

PR #4006, opened 2023-04-19
EES-4083 update to google analytics 4

amyb-hiveit:
Updated to Google Analytics 4:
- replaced react-ga with react-ga4
- removed code to track outbound links as that should be done automatically with GA4
- remove code to log exceptions as it wasn't used and react-ga4 doesn't support it
- updated the CSP
- updated the cookies page

Do not merge - we want to check this on Test before merging to dev.

Will need to update the tracking id for each environment as it's deployed.


src/explore-education-statistics-frontend/src/services/googleAnalyticsService.ts
```diff
@@ -54,37 +53,3 @@ export function logEvent({ category, action, label, value }: AnalyticsEvent) {
...
-      label,
-    },
-
-    function hitCallback() {
-      openLink();
-    },
-  );
-}
-
-export const logException = (description: string, fatal = false) => {
```
luke-h1, line 86: Have you tried doing something such as the following to log exceptions? This is what react-ga sends under the hood I believe so it should just work 

```typescript 
export function logException(description: string, fatal = false) {
  if (initialised) {
    ReactGA.set({
      hitType: 'exception',
      exDescription: description,
      exFatal: fatal,
    });
  }
}

```

amyb-hiveit: It looks like it doesn't support it https://github.com/codler/react-ga4/blob/master/src/ga4.js#L341

luke-h1 (approved):
Looks ok to me. Part of me wonders how the outbound links will show up in the GA console (just thinking how the messaging will differ i.e. how will `Methodology page content link: ${url}` show up now?). Suppose this can be checked via manual testing

benoutram:
Updated branch with the latest from dev.

PR #4010, opened 2023-04-24
EES-4228 Store map chart data groupings by data set on the backend

markhiveit:
Allows new map data groupings from EES-3858 to be saved on the backend.

- I've added `List<ChartDataGroups>?` to the `Chart` object, even though it is only used with maps. This mirrors how the frontend adds it to `ChartDefinition`.
- Changed `ChartLegendItemDataSet` to `ChartItemDataSet` so it can also be used by `ChartDataGroups`. 
- Once EES-1649 is done, I think we could probably remove `ChartItemDataSet` and just have everything use the `ChartDataSet` - I've left a comment in the code.


src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartDataGrouping.cs
```diff
@@ -0,0 +1,25 @@
...
+using Newtonsoft.Json.Converters;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Model.Chart
+{
+    public class ChartDataGrouping
+    {
+        public ChartItemDataSet DataSet;
+        public ChartDataGroupingType Type;
+        public int NumberOfGroups;
+        public List<ChartCustomDataGroup> CustomGroups = new ();
```
ntsim, line 14: Nitpick, but typically we write implicit `new` as `new()`

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartDataSet.cs
```diff
@@ -14,9 +14,18 @@ public class ChartDataSet
         public int? Order;
 
         // TODO EES-1649 Migrate data set configs to legend item configs
+        // After EES-1649 migration, could consider removing ChartItemDataSet?
         public ChartDataSetConfiguration Config;
     }
 
+    public class ChartItemDataSet
```
ntsim, line 21: `ChartItemDataSet` is a bit of weird name. Would suggest `ChartBaseDataSet` and `ChartDataSet` could just also extend from it i.e.

```cs
public class ChartBaseDataSet
{
    public Guid Indicator;
    public List<Guid> Filters = new();
    public ChartDataSetLocation Location;
    public string TimePeriod;
}

public class ChartDataSet : ChartBaseDataSet
{
    public int? Order;

    // TODO EES-1649 Migrate data set configs to legend item configs
    // After EES-1649 migration, could consider removing ChartItemDataSet?
    public ChartDataSetConfiguration Config;
}
```

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReplacementServiceTests.cs
```diff
@@ -3145,6 +3164,15 @@ await using (var statisticsDbContext = InMemoryStatisticsDbContext(statisticsDbC
...
                 Assert.Equal(replacementLocation.Id, chartLegendItem.DataSet.Location!.Value);
 
+                var chartDataSetConfigs = replacedDataBlock.Charts[0].Map;
+                Assert.NotNull(chartDataSetConfigs);
+                var chartDataSetConfig = Assert.Single(chartDataSetConfigs);
+                var filterId = Assert.Single(chartDataSetConfig.DataSet.Filters);
+                Assert.Equal(replacementFilterItem1.Id, filterId);
+                Assert.Equal(replacementIndicator.Id, chartDataSetConfig.DataSet.Indicator);
+                Assert.NotNull(chartLegendItem.DataSet.Location);
+                Assert.Equal(replacementLocation.Id, chartLegendItem.DataSet.Location!.Value);
```
ntsim, line 3174: Why are these assertions here? We're looking at data set configs so these are irrelevant and have already been assert earlier anyway.

It seems like we should actually be asserting on the data set configs instead?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReplacementService.cs
```diff
@@ -784,10 +784,68 @@ private static FilterReplacementViewModel CreateNewlyIntroducedFilterReplacement
...
         }
 
+        private static void ReplaceChartDataSetConfigs(
+            IReadOnlyDictionary<Guid, Guid> filterItemTargets,
+            IReadOnlyDictionary<Guid, Guid> indicatorTargets,
+            IReadOnlyDictionary<Guid, Guid> locationTargets,
+            DataBlock dataBlock,
+            IChart chart)
+        {
+            if (chart.Map == null) { return; }
```
ntsim, line 804: Would suggest we be consistent with how we typically format conditions i.e.

```cs
if (chart.Map == null)
{
    return;
}
```

ntsim (changes requested):
Looks like this has merge conflicts as it was based off `EES-3858` which was rebased with some recent changes. Will need to be rebased back onto the latest `EES-3858`.

However, wouldn't this have been less trouble to just target dev anyway?

markhiveit:
> However, wouldn't this have been less trouble to just target dev anyway?

I wanted to test it alongside the frontend work - I wasn't anticipating many more changes.

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/Charts.cs
```diff
@@ -32,6 +32,7 @@ public interface IChart
         bool IncludeNonNumericData { get; set; }
 
         Dictionary<string, ChartAxisConfiguration>? Axes { get; set; }
+        public List<ChartDataSetConfig>? Map { get; set; }
```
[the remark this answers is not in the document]
markhiveit: I wasn't sure where to put `MapChartConfig`. I put it below `MapChart` - I figured it's too small to be worth creating a new file for it.

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartDataSetConfig.cs
```diff
@@ -0,0 +1,41 @@
...
+    [SuppressMessage("ReSharper", "InconsistentNaming")]
+    [JsonConverter(typeof(StringEnumConverter))]
+    public enum ChartDataGroupingType
+    {
+        EqualIntervals,
+        Quantiles,
+        Custom,
+    }
+
+    public class ChartDataSetConfigDataSet
```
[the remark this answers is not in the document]
markhiveit: I'll make the change. I opted for three different `*DataSet` objects as each have different optional parameters - but it seems that might be wrong? Anyway, just so you're aware, with this change now we may have a data set that requires an indicator but `ChartBaseDataSet` will have optional indicators because of how it's used on something else.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReplacementServiceTests.cs
markhiveit: I copied and pasted from above and missed that I'd not changed these to test on data set config rather than the legend - I do that often enough that I wonder whether copying and pasting isn't worth the risk...

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartDataSet.cs
```diff
@@ -5,16 +5,20 @@
 
 namespace GovUk.Education.ExploreEducationStatistics.Common.Model.Chart
 {
-    public class ChartDataSet
+    public class ChartBaseDataSet
```
ntsim, line 8: Needs an `#enable nullable` at the top of this file

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartDataSetConfig.cs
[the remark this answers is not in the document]
ntsim: Good point! Would suggest we merge that into `ChartBaseDataSet` as well, so the final thing should be:


```cs
public class ChartBaseDataSet
{
    public Guid? Indicator;
    public List<Guid> Filters = new();
    public ChartDataSetLocation? Location;
    public string? TimePeriod;
}
```

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/Charts.cs
```diff
@@ -79,18 +80,24 @@ public class VerticalBarChart : Chart
...
 
+        // Deprecated - Migrate to `DataSetConfigs` planned in EES-4271
         [JsonConverter(typeof(StringEnumConverter))]
         public ChartDataClassification? DataClassification { get; set; }
 
+        // Deprecated - Migration to `DataSetConfigs` planned in EES-4271
         public int? DataGroups { get; set; }
 
-        public List<ChartCustomDataGroup>? CustomDataGroups { get; set; }
+        public MapChartConfig Map { get; set; }
```
benoutram, line 95: Should be `} = null!;` if `Map` is not optional and deliberately being initialised as null.

Could consider whether this should be made optional, or initialised with `new()` instead.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/Charts.cs
```diff
@@ -79,18 +80,24 @@ public class VerticalBarChart : Chart
...
+        public override ChartType Type => ChartType.Map;
 
         // TODO EES-3319 - make mandatory when all Map Charts are migrated to have a Boundary Level set
         public long? BoundaryLevel { get; set; }
 
+        // Deprecated - Migrate to `DataSetConfigs` planned in EES-4271
         [JsonConverter(typeof(StringEnumConverter))]
         public ChartDataClassification? DataClassification { get; set; }
 
+        // Deprecated - Migration to `DataSetConfigs` planned in EES-4271
```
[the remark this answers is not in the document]
markhiveit: Probably overkill, but I've done both :)

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/Charts.cs
markhiveit: I believe it should always be initiated, even if the `List<ChartDataSetConfig>` is empty, so I've gone with `new()`.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReplacementServiceTests.cs
benoutram: Tidied it up in d074f5d34bcf7059385006bb17f3bd2f7448d80c

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReplacementServiceTests.cs
```diff
@@ -2937,9 +2936,671 @@ public async Task Replace()
...
+                Assert.Equal("Original guidance version 2", replacedReleaseSubject.DataGuidance);
+
+                // Check the sequence of filters and indicators remains untouched
+                Assert.Null(replacedReleaseSubject.FilterSequence);
+                Assert.Null(replacedReleaseSubject.IndicatorSequence);
+            }
+        }
+
+        [Fact]
+        public async Task Replace_MapChart() // @MarkFix thin this test case out to be more map specific
```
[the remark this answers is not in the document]
benoutram: Changed `Replace_MapChart` to only test changes to `DataSetConfigs` in d074f5d34bcf7059385006bb17f3bd2f7448d80c.

benoutram (approved):

PR #4013, opened 2023-04-25
Added support for providing arbitrary IdP configuration that will be …

duncan-at-hiveit:
This PR:
- removes the App Registration to Hive's main AD from the codebase
- adds in the ability to drop in any IdP configuration into a .gitignored file `appsettings.Idp.json` so that we can provide configuration away from source control to connect to our new DfE Development AD App Registration

There is also a fix to the Expired Invite UI test where a BST time was being provided to an endpoint that was expecting UTC.

## New config file

The configuration file to connect to the new Azure DfE Development App Registration can be found alongside the other shared passwords under the title "Azure AD IdP configuration".  Its contents can be dropped into a new `appsettings.Idp.json` file in Admin.  Keycloak continues to operate as per usual.

## Tasks after this work has gone in

After everybody has moved to the new AD, the App Registration can be removed from the old AD, the test users removed, and security defaults enabled.

## Future changes and simplifications

Now that we can drop an arbitrary set of IdP configuration into `appsettings.Idp.json`, it's perfectly possible to drop in the Keycloak config into that file as well as the DfE Development Azure config.  Therefore it might no longer be necessary to maintain a `./run.js adminKeycloak`, in favour of just dropping Keycloak config into that file.  This being the case, we could then also do away with the `IdpProviderConfiguration` environment variable and the `BootstrapUsersConfiguration` environment variable too (in favour of another .gitignored file maybe like `users.Idp.json`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,45 +196,46 @@ The service can be started against a set of non-existent database. If no pre-exi
 3. Start the Admin project and this will configure the contained users' permissions via database migrations. The other 
    projects will then be able to be started, using their own contained users to connect to the databases. 
 
-### Running a different identity provider (optional)
+### Setting up an Identity Provider
 
-> For running the project day-to-day as a team member, you can ignore this step.
+The project uses an OpenID Connect Identity Provider to allow login to the Admin service.
 
-Currently, the project defaults to using Active Directory as its identity provider. Typically, this
-will be used by components such as the admin service to allow users to log in.
+> For team members, Azure AD configuration is available alongside other project passwords with the title `Azure AD IdP 
+configuration`. Place the contents of this into [appsettings.Idp.json](
+src/GovUk.Education.ExploreEducationStatistics.Admin\appsettings.Idp.json) and start Admin normally.
 
-If you wish use a different identity provider (e.g. working outside the team), you can use:
+An out-of-the-box IdP is provided for ease of setup which runs [Keycloak](https://www.keycloak.org/) in a Docker container 
+and contains a number of users for different roles. The [appsettings.Keycloak.json](
+src/GovUk.Education.ExploreEducationStatistics.Admin\appsettings.Keycloak.json) configuration file contains the details for 
+connecting Admin to this IdP.
 
-- Our out-of-the-box identity provider called [Keycloak](https://www.keycloak.org/) (as a Docker 
-  container).
-- Any OpenID Connect compatible identity provider e.g. Active Directory. It must have Implicit Flow 
-  enabled and be using the OpenID Connect protocol. It must be set to issue ID Tokens.
+Alternatively, you can provide your own OpenID Connect configuration in the gitignored [appsettings.Idp.json](
+src/GovUk.Education.ExploreEducationStatistics.Admin\appsettings.Idp.json) file (you can use the Keycloak equivalent as a
+template as to how the configuration in the file should be structured).
 
-#### Using Keycloak identity provider with standard seed data users
+#### Using Keycloak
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -196,45 +196,46 @@ The service can be started against a set of non-existent database. If no pre-exi
...
+An out-of-the-box IdP is provided for ease of setup which runs [Keycloak](https://www.keycloak.org/) in a Docker container 
+and contains a number of users for different roles. The [appsettings.Keycloak.json](
+src/GovUk.Education.ExploreEducationStatistics.Admin\appsettings.Keycloak.json) configuration file contains the details for 
+connecting Admin to this IdP.
 
-- Our out-of-the-box identity provider called [Keycloak](https://www.keycloak.org/) (as a Docker 
-  container).
-- Any OpenID Connect compatible identity provider e.g. Active Directory. It must have Implicit Flow 
-  enabled and be using the OpenID Connect protocol. It must be set to issue ID Tokens.
+Alternatively, you can provide your own OpenID Connect configuration in the gitignored [appsettings.Idp.json](
```
luke-h1, line 212:
```suggestion
Alternatively, you can provide your own OpenID Connect configuration in the git ignored [appsettings.Idp.json](
```

README.md
```diff
@@ -249,36 +250,47 @@ cd src/
 docker-compose up --build --force-recreate idp
 ```
 
-#### Using Keycloak identity provider with custom users
+#### Using a different Identity Provider
+
+If you have your own OpenID Connect IdP set up, you can provide its configuration in the gitignored
```
luke-h1, line 255:
```suggestion
If you have your own OpenID Connect IdP set up, you can provide its configuration in the git ignored 
```

src/GovUk.Education.ExploreEducationStatistics.Admin/appsettings.json
```diff
@@ -24,9 +24,6 @@
     "RefreshTokenExpiration": "Absolute"
   },
   "OpenIdConnect": {
-    "ClientId": "97d25f91-bd4b-4b82-843b-2583b8194800",
-    "ClientSecret": "C@R3I_k42?3=15W5-KzV=Nbpg.eBq/:4",
```
ntsim, line 28: Given that details are now firmly engraved into the Git history, are we able to rotate these credentials in our Hive AD once this change has gone in?

duncan-at-hiveit: Yeah, we'll be removing the App Registration altogether, so it'll no longer be a "thing".

README.md
duncan-at-hiveit: Updated the text thanks

README.md
duncan-at-hiveit: Updated the text to be clearer, thanks

README.md
```diff
@@ -249,36 +250,47 @@ cd src/
...
+Then set the environment variables:
 
 ```
 IdpProviderConfiguration={NameOfYourIdentityProvider}
 BootstrapUsersConfiguration={NameOfYourIdentityProvider}BootstrapUsers
 ```
 
-If choosing to provide your own OpenID Connect configuration file, you can use the existing Keycloak configuration file
-as a reference, at [appsettings.Keycloak.json](src/GovUk.Education.ExploreEducationStatistics.Admin\appsettings.Keycloak.json).
+and start up Admin wih these environment variables set. This allows you to more easily switch between different IdP
```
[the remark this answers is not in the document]
duncan-at-hiveit: Updated ta

luke-h1 (approved):

PR #4014, opened 2023-04-25
EES-4231 Backend work to support new featured tables

markhiveit:
This PR provides changes to give the frontend the information it needs to render the new featured tables design.

It seems this requires just two changes:
- Return the `subjectId` alongside featured tables.
- Return filter and indicator labels alongside subjects.

:rotating_light: After this is merged, we will need to clear the public cache `subject.json` files. I'll create a faffy deploy ticket to track this. (Edit: EES-4266) :rotating_light:

I'll also put do not merge on this ticket. In theory, we could merge it immediately - it doesn't affect existing functionality - but it might make sense for these changes to go in alongside the frontend changes for featured tables in EES-4048.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Services/ReleaseService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Services/ReleaseService.cs
@@ -11,11 +11,13 @@ using GovUk.Education.ExploreEducationStatistics.Content.Model;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Content.Security.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Data.Model;
 using GovUk.Education.ExploreEducationStatistics.Data.Model.Database;
 using GovUk.Education.ExploreEducationStatistics.Data.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Data.Services.ViewModels;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.EntityFrameworkCore;
+using Release = GovUk.Education.ExploreEducationStatistics.Content.Model.Release;
 
 namespace GovUk.Education.ExploreEducationStatistics.Data.Services
 {
@@ -88,8 +90,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Services
                                 content: rs.DataGuidance ?? string.Empty,
                                 timePeriods: await _timePeriodService.GetTimePeriodLabels(rs.SubjectId),
                                 geographicLevels: await _dataGuidanceSubjectService.GetGeographicLevels(rs.SubjectId),
-                                filters: await GetFilters(rs.SubjectId),
-                                indicators: await GetIndicators(rs.SubjectId),
+                                filters: await GetFilters(rs.SubjectId, rs.FilterSequence),
+                                indicators: await GetIndicators(rs.SubjectId, rs.IndicatorSequence),
                                 file: releaseFile.ToFileInfo()
                             );
                         }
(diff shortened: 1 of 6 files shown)


src/GovUk.Education.ExploreEducationStatistics.Data.Services/ReleaseService.cs
```diff
@@ -113,11 +135,13 @@ private async Task<List<SubjectViewModel>> GetSubjects(Guid releaseId, List<Guid
...
-                        name: dataBlock.HighlightName ?? string.Empty,
-                        description: dataBlock.HighlightDescription ?? string.Empty
-                    )
+                    dataBlock => new FeaturedTableViewModel
+                    {
+                        Id = dataBlock.Id,
+                        Name =  dataBlock.HighlightName ?? string.Empty,
+                        Description = dataBlock.HighlightDescription ?? string.Empty,
+                        SubjectId = dataBlock.Query.SubjectId,
+                    }
```
benoutram, line 144: I noticed that this `ReleaseService.ListFeaturedTables(Guid releaseId)` is called by two different controllers, one for listing featured tables by release Id and one by publication Id(which in turn works out the latest published release id first before  calling this.

Do you know if this work will make one of these routes redundant? If so we could mark it for removal so we don't forget to tidy up afterwards.

benoutram (commented):
Just wondering if there's anything else in the backend that won't be relevant after this change goes in that we can identify and mark for removal now, so that we don't forget to tidy up later on?

An example might be listing themes in step 1.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ReleaseService.cs
markhiveit: I think this could be covered by the ticket to split featured tables into their own entity, which I'm currently working on, so I will leave a comment for myself to look into this.

markhiveit: Commented added here: https://github.com/dfe-analytical-services/explore-education-statistics/blob/b89c646a3208122c551403f7497cb7b1595bbb43/src/GovUk.Education.ExploreEducationStatistics.Data.Api/Controllers/PublicationController.cs

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ReleaseService.cs
```diff
@@ -97,6 +99,26 @@ private async Task<List<SubjectViewModel>> GetSubjects(Guid releaseId, List<Guid
...
+        private async Task<List<string>> GetIndicators(Guid subjectId)
+        {
+            return await _statisticsDbContext.IndicatorGroup
+                .Include(ig => ig.Indicators)
+                .Where(indicatorGroup => indicatorGroup.SubjectId == subjectId)
+                .SelectMany(indicatorGroup =>
+                    indicatorGroup.Indicators.Select(indicator => indicator.Label))
+                .OrderBy(indicatorLabel => indicatorLabel)
+                .ToListAsync();
+        }
```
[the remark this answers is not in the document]
markhiveit: Also added a couple of unit tests.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ViewModels/FeaturedTableViewModel.cs
```diff
@@ -5,17 +5,12 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Services.ViewModels
 {
     public record FeaturedTableViewModel
     {
-        public Guid Id { get; }
+        public Guid Id { get; set; }
```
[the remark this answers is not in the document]
duncan-at-hiveit: Just so you know, you can also just write this:

```
public record FeaturedTableViewModel(Guid Id, string Name, string Description, Guid SubjectId);
```

duncan-at-hiveit (approved):

PR #4015, opened 2023-04-25
EES-4267 Log errors when submitting Form components that display the error

ntsim:
This PR adds error logging to `Form` and `RHForm` components during submission. This only applies in cases where `showSubmitError` is set to true.


luke-h1 (dismissed):

src/explore-education-statistics-common/src/components/form/rhf/RHFForm.tsx
```diff
@@ -118,12 +119,10 @@ export default function RHFForm<TFormValues extends FieldValues>({
         try {
           await onSubmit(values as TFormValues);
         } catch (error) {
-          if (!error) {
-            return;
-          }
+          if (showSubmitError && isErrorLike(error)) {
```
duncan-at-hiveit, line 122: Do we want to log it even if we choose not to show it actually?

src/explore-education-statistics-common/src/components/form/Form.tsx
```diff
@@ -99,12 +100,10 @@ const Form = ({
       try {
         await submitForm();
       } catch (error) {
-        if (!error) {
-          return;
-        }
+        if (showSubmitError && isErrorLike(error)) {
```
duncan-at-hiveit, line 103: Do we want to log it even if we choose not to show it actually?

src/explore-education-statistics-common/src/components/form/rhf/RHFForm.tsx
ntsim: No, because it gets re-thrown. This will typically get handled by the global error handler, but it's really up to the consuming code to decide what to do with such an error.

duncan-at-hiveit (approved):

PR #4019, opened 2023-04-28
EES-4277 Trim leading and trailing spaces for featured table names and descriptions

markhiveit:
This prevents a bug where by allowing a whitespace featured table name, potentially a featured table with no title could appear on be published and appear on the public site.

This work is done in preparation for splitting featured tables details out of the ContentBlocks table and into their own entity.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
@@ -66,6 +66,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
                 {
                     var dataBlock = _mapper.Map<DataBlock>(dataBlockCreate);
                     dataBlock.Created = DateTime.UtcNow;
+                    dataBlock.HighlightName = dataBlock.HighlightName?.Trim();
+                    dataBlock.HighlightDescription = dataBlock.HighlightDescription?.Trim();
 
                     var added = (await _context.DataBlocks.AddAsync(dataBlock)).Entity;
 
@@ -176,6 +178,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
                             }
 
                             _mapper.Map(dataBlockUpdate, dataBlock);
+                            dataBlock.HighlightName = dataBlock.HighlightName?.Trim();
+                            dataBlock.HighlightDescription = dataBlock.HighlightDescription?.Trim();
 
                             _context.DataBlocks.Update(dataBlock);
                             await _context.SaveChangesAsync();
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
@@ -66,8 +66,6 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
                 {
                     var dataBlock = _mapper.Map<DataBlock>(dataBlockCreate);
                     dataBlock.Created = DateTime.UtcNow;
-                    dataBlock.HighlightName = dataBlock.HighlightName?.Trim();
-                    dataBlock.HighlightDescription = dataBlock.HighlightDescription?.Trim();
 
                     var added = (await _context.DataBlocks.AddAsync(dataBlock)).Entity;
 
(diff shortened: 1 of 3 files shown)


src/explore-education-statistics-admin/src/pages/release/datablocks/components/__tests__/DataBlockDetailsForm.test.tsx
```diff
@@ -77,6 +77,37 @@ describe('DataBlockDetailsForm', () => {
     });
   });
 
+  test('shows validation error if has whitespace featured table name', async () => {
```
luke-h1, line 80: Think this test is a bit redundant since no error will show for this scenario

markhiveit: How do you mean? I think after my changes, if you put a bunch of spaces in the featured table name and try to save it, it'll throw the error "Enter a featured table name".

luke-h1: Why would we want this to be the case? We shouldn't bother the user with this and just trim the string before sending it to the server.

markhiveit: I've only performed the trim on `highlightName` and `highlightDescription` - the other fields still have validation issues, but I think that's outside the remit of the work I'm doing.

I'll create a ticket to test inputs across the service to see what mischief could be achieved, and then we can fix them all at once if that's a priority. The reason I didn't was because I think this is a problem across the codebase I'm not sure how big of a priority this is, as analysts are unlikely to sabotage their own releases in this way. But to be honest, that's for Cam to decide, so I'll create the ticket.

luke-h1: Ah right ok I see, good shout. This wasn't quite the point I was trying to get across in this original comment however. My main point was that we're trimming the string before performing any validation that will surface to the end user. i.e. if we had `Yup.string().trim()` the only validation that will be performed is that the input must be of type `string` (and we trim it after the fact). No error will show up around a string containing whitespace

markhiveit: You mean like if someone input a string like `"    Featured table name    "` which would automatically be trimmed to `"Featured table name"`?

I think in this scenario we should just trim the string and don't need to provide a validation error. I don't think there is a scenario where a user would want whitespace before or after their featured table name, and if we did throw a validation error on strings like this, it just means users will have to trim the whitespace manually. So it makes sense to me to just trim it automatically.

markhiveit: Maybe the confusion is around the name of the unit test you're querying? That test was supposed to cover featured table names that are entirely whitespace - i.e. no characters. The second unit test is to cover the situation where there are leading and/or trailing whitespace. 

I guess you could argue that the validation error should be different based on whether the input name is empty or full of whitespace, but that didn't seem worth the effort to me.

luke-h1: Yep that's right, if someone enters a string with loads of whitespace it'll be automatically trimmed to "Featured table name" without surfacing a validation error. This is what we're doing at the moment in this PR (and across the codebase atm). I see what the test is trying to do now, I think the test title is a bit misleading. Maybe it could be renamed to something such as `shows validation error if the featured table name is entirely a whitespace string` or something along those lines

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
```diff
@@ -66,6 +66,8 @@ public class DataBlockService : IDataBlockService
                 {
                     var dataBlock = _mapper.Map<DataBlock>(dataBlockCreate);
                     dataBlock.Created = DateTime.UtcNow;
+                    dataBlock.HighlightName = dataBlock.HighlightName?.Trim();
+                    dataBlock.HighlightDescription = dataBlock.HighlightDescription?.Trim();
```
benoutram, line 70: I don't think we do this kind of input sanitisation anywhere else on the backend. We leave it to the frontend, which this PR is also doing in `DataBlockDetailsForm.tsx`.

I think it's a good idea for the backend to have sanitisation but don't want this to set a precedent that we should be adding `.Trim()` everywhere. We'd litter the code with `.Trim()` and burden ourselves with spotting places where it's missing in new PR reviews.

Rather we should be coming up with a way to apply sanitisation on the backend globally for all requests with some kind of model binding, also with a way of annotating an exception cases if we think that's going to be required.

markhiveit: As discussed on Slack, I'll add some details to the EES-4279 I've already created so it's clear we're looking for a complete solution to validation on the backend.

Also removed the `.Trims` I've added on the backend.

benoutram (approved):

PR #4021, opened 2023-05-02
EES-3754 Add SSL certificate verification

benoutram:
This PR adds SSL certificate verification, so that we can use a Data API URL with https and not see a `InsecureRequestWarning` repeated in the log output with every request.

>connectionpool.py:1045: InsecureRequestWarning: Unverified HTTPS request is being made to host '...'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings


luke-h1 (approved):
Looks good to me

PR #4024, opened 2023-05-03
EES-4276 - EES-4192 additional enhancements

duncan-at-hiveit:
image

This PR:
- builds upon the work already done for swapping parallel batch imports for sequential ones.
- replaces less efficient CSV lookup strategies (originally using `CsvUtils.BuildType` and `CsvUtils.Value` methods) with more efficient `DataFileReader` and `MetaDataFileReader` classes, which cache column indexes for quick lookup of pertinent information.
- removes `ImporterFilterCache` and `ImporterFilterService` classes, as there is no longer a need to provide a cache which can be shared between various parallel importing jobs.  The quick lookup of FilterItems is retained however in `DataFileReader` as it is still useful.

## Performance benefit

Tests against a local development environment have shown speed advantages overall, more noticeable in particular Subjects.  It's hoped that the benefits will be more exaggerated in a proper Azure environment as well.  See below for some big Subject files uploaded against the dev branch code, the 4192 PR code and this PR code:

### QL01
* Dev - 8 mins 35 secs
* 4192 - 4 mins 0 secs
* 4192 + this - 2 mins 20 secs

### LA01
* Dev - 8 mins 50 secs
* 4192 - 6 mins 10 secs
* 4192 + this - 5 mins 40 secs

### ND01
* Dev - 1 mins 30 secs
* 4192 - 1 mins 24 secs
* 4192 + this - 1 mins 10 secs

### E+T
* Dev - 14 mins 57 secs
* 4192 - 14 mins 17 secs
* 4192 + this - 12 mins 20 secs


src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Models/DataFileReader.cs
```diff
@@ -0,0 +1,425 @@
...
+    
+    private readonly Dictionary<Guid, int>? _indicatorColumnIndexes;
+    private readonly Dictionary<Guid, int>? _filterColumnIndexes;
+    private readonly Dictionary<Guid, int>? _filterGroupColumnIndexes;
+    
+    private readonly Dictionary<string, FilterItem>? _filterItemCache;
+
+    public DataFileReader(
+        List<string> csvHeaders, 
+        SubjectMeta? subjectMeta = null)
```
benoutram, line 151: I'm not sure I follow why `subjectMeta` is optional here and why these dictionaries are optional. The methods which rely on them don't handle them being optional.

For example

```
    public Dictionary<Guid, string> GetMeasures(List<string> rowValues)
    {
        return _indicatorColumnIndexes!
            .ToDictionary(
                indicatorMeta => indicatorMeta.Key,
                indicatorMeta => rowValues[indicatorMeta.Value]);
    }
```

duncan-at-hiveit: Hey Ben - yeah, this is because all of the non-optional stuff in DataFileReader is usable both in Stage 2 of the import (before we have the SubjectMeta) and also in Stage 3 (after we have the SubjectMeta).

I could break the class into 2 reader classes, one called "DataFileSubjectMetaReader" and one called "DataFileIndicatorReader" or something along those lines instead?

benoutram: Yeah, I think splitting it up will make it a bit more obvious as to how it's intended to be used and what's required to construct it and when.

It will break up the particularly long constructor of `DataFileReader` here, and remove the optional from these fields, and also have the benefit of removing the null overriding operators throughout the methods using them, so it seems like a good idea to me to split them up.

duncan-at-hiveit: Cool will do!

PR #4035, opened 2023-05-11
EES-3792 new permalink snapshots under feature flag

amyb-hiveit:
Adds the new permalink creation process and permalink page under a feature flag.

Add the query param `newPermalinks=true` to see the new version:
- to create new permalinks `/data-tables?newPermalinks=true`
- to view new permalinks `/data-tables/permalink<id>?newPermalinks=true`

### DIFF
--- a/src/explore-education-statistics-common/src/services/permalinkSnapshotService.ts
+++ b/src/explore-education-statistics-common/src/services/permalinkSnapshotService.ts
@@ -1,0 +1,53 @@
+import { TableJson } from '@common/modules/table-tool/utils/mapTableToJson';
+import { dataApi } from '@common/services/api';
+import { TableDataQuery } from '@common/services/tableBuilderService';
+import { Footnote } from '@common/services/types/footnotes';
+import { UnmappedTableHeadersConfig } from '@common/services/permalinkService';
+
+export interface PermalinkSnapshot {
+  created: string;
+  dataSetTitle: string;
+  id: string;
+  publicationTitle: string;
+  status:
+    | 'Current'
+    | 'SubjectRemoved'
+    | 'SubjectReplacedOrRemoved'
+    | 'NotForLatestRelease'
+    | 'PublicationSuperseded';
+  table: {
+    caption: string;
+    footnotes: Footnote[];
+    json: TableJson;
+  };
+}
+
+interface CreatePermalink {
+  query: TableDataQuery;
+  configuration: {
+    tableHeaders: UnmappedTableHeadersConfig;
+  };
+}
+
+const permalinkSnapshotService = {
+  createPermalink(query: CreatePermalink): Promise<PermalinkSnapshot> {
+    return dataApi.post(`/permalink-snapshot`, query);
+  },
+  async getPermalink(id: string): Promise<PermalinkSnapshot> {
+    return dataApi.get(`/permalink-snapshot/${id}`, {
+      headers: {
+        Accept: 'application/json',
+      },
+    });
+  },
+  async getPermalinkCsv(id: string): Promise<Blob> {
+    return dataApi.get(`/permalink-snapshot/${id}`, {
+      headers: {
+        Accept: 'text/csv',
+      },
+      responseType: 'blob',
+    });
+  },
+};
+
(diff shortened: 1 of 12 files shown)


benoutram:
Footnotes added in for EES-4292 by https://github.com/dfe-analytical-services/explore-education-statistics/pull/4038.

src/explore-education-statistics-common/src/services/permalinkSnapshotService.ts
```diff
@@ -0,0 +1,51 @@
...
+import { Footnote } from '@common/services/types/footnotes';
+import { UnmappedTableHeadersConfig } from '@common/services/permalinkService';
+
+export interface PermalinkSnapshot {
+  created: string;
+  dataSetTitle: string;
+  id: string;
+  publicationTitle: string;
+  status:
+    | 'Current'
```
luke-h1, line 13: Would suggest creating a common type for this since it's reused in a couple of places

src/explore-education-statistics-common/src/services/permalinkSnapshotService.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+interface CreatePermalink {
+  query: TableDataQuery;
+  configuration: {
+    tableHeaders: UnmappedTableHeadersConfig;
+  };
+}
+
+export default {
```
luke-h1, line 32: Would suggest we make this a named export as It'll soon become a lint error (via EES-3409)

src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
```diff
@@ -42,17 +45,34 @@ const TableToolShare = ({
     }
     setPermalinkLoading(true);
 
-    const { id } = await permalinkService.createPermalink(
-      {
+    let id;
```
luke-h1, line 48: Would suggest adding a type param or default value so it isn't inferred as `any`

src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
```diff
@@ -42,17 +45,34 @@ const TableToolShare = ({
...
+            tableHeaders: mapUnmappedTableHeaders(tableHeaders),
+          },
+        },
+        selectedPublication.selectedRelease.id,
+      );
+      id = response.id;
+    }
+
+    const url = newPermalinks
+      ? `${window.location.origin}/data-tables/permalink/${id}?newPermalinks=true`
```
luke-h1, line 72: Would suggest replacing `window.location.origin` with `process.env.PUBLIC_URL`

luke-h1 (changes requested):

src/explore-education-statistics-common/src/services/permalinkSnapshotService.ts
amyb-hiveit: The only place it's reused is in permalinkService which will be replaced by permalinkSnapshotService when the feature flag is removed so doesn't seem necessary to create a common type for it.

luke-h1: Ah right, cool. Forgot `permalinkService` on both the BE & FE will be deleted in the future

src/explore-education-statistics-frontend/src/modules/permalink/PermalinkPage.tsx
```diff
@@ -141,12 +138,19 @@ const PermalinkPage: NextPage<Props> = ({ data }) => {
 export const getServerSideProps: GetServerSideProps<Props> = async ({
   query,
 }) => {
-  const { permalink } = query;
-  const data = await permalinkService.getPermalink(permalink as string);
+  const { newPermalinks, permalink } = query as Dictionary<string>;
+  // TO DO - EES-4259 remove `newPermalinks` and tidy up
+  let data;
```
luke-h1, line 143: Similar to my other comment, think we should add a type param here so `data` isn't inferred as `any`

luke-h1 (approved):

PR #4037, opened 2023-05-11
EES-4011 handle standard codes in maps

amyb-hiveit:
Allows data containing standard codes (e.g. 'x' or 'c') to be shown in maps. Previously areas with non-numeric values weren't shown at all, now the areas are shown with a transparent background and show the code when selected.

**Previously:**
map-old

**Now:**
map-new2
map-new


src/explore-education-statistics-common/src/modules/charts/util/createDataSetCategories.ts
```diff
@@ -308,19 +308,27 @@ function getSortedDataSetCategoryRange(
...
-  axisConfiguration: AxisConfiguration,
-  results: TableDataResult[],
-  meta: FullTableMeta,
+interface Options {
+  axisConfiguration: AxisConfiguration;
+  data: TableDataResult[];
+  meta: FullTableMeta;
+  includeNonNumericData?: boolean;
+  returnSymbols?: boolean;
+}
```
ntsim, line 317: This should go above the docblock for the function and have a line break after it

src/explore-education-statistics-common/src/modules/charts/util/createDataSetCategories.ts
```diff
@@ -308,19 +308,27 @@ function getSortedDataSetCategoryRange(
...
-export default function createDataSetCategories(
-  axisConfiguration: AxisConfiguration,
-  results: TableDataResult[],
-  meta: FullTableMeta,
+interface Options {
+  axisConfiguration: AxisConfiguration;
+  data: TableDataResult[];
+  meta: FullTableMeta;
+  includeNonNumericData?: boolean;
+  returnSymbols?: boolean;
```
ntsim, line 316: Seems like `returnNonNumbericData` or similar makes more sense

src/explore-education-statistics-common/src/modules/charts/util/createDataSetCategories.ts
```diff
@@ -330,10 +338,11 @@ export default function createDataSetCategories(
...
 
     if (includeNonNumericData && rawValue) {
-      acc.push([childDataSet, Number.isNaN(value) ? null : value] as Pair<
-        ChildDataSet,
-        number
-      >);
+      const nonNumericValue = returnSymbols ? rawValue : null;
+      acc.push([
+        childDataSet,
+        Number.isNaN(value) ? nonNumericValue : value,
```
ntsim, line 344: Would potentially suggest refactoring this a little to combine it with the second `if` block i.e.

```ts
if (!Number.isNaN(value)) {
  acc.push([childDataSet, value]);
} else if (includeNonNumericData && rawValue) {
  acc.push([childDataSet, returnNonNumericData ? rawValue : null]);
}
```

Actually, I'm now confused why we even need the `returnNonNumericData` / `returnSymbols` flag?

Couldn't this just be:

```ts
if (!Number.isNaN(value)) {
  acc.push([childDataSet, value]);
} else if (includeNonNumericData && rawValue) {
  acc.push([childDataSet, rawValue]);
}
```

ntsim (changes requested):

src/explore-education-statistics-common/src/modules/charts/util/createDataSetCategories.ts
amyb-hiveit: It was to not change the current behaviour for charts (the data sets are shown with null values), but I've checked with Cam and he says it's preferable to show the raw value for charts as well so I've got rid of  `returnSymbols` now.

src/explore-education-statistics-common/src/modules/charts/util/createDataSetCategories.ts
amyb-hiveit: This has been removed now.

ntsim (approved):

ntsim (approved):

PR #4038, opened 2023-05-11
EES-4292 Add footnotes to permalink snapshots

benoutram:
This PR is the backend work to add footnotes to permalink snapshots.

EES-3753 in https://github.com/dfe-analytical-services/explore-education-statistics/pull/3966 assumed that the frontend Next.js API would return footnotes but this was not the case. Even if was, it would have meant that the frontend was unnecessarily handling footnotes and responding with them.

Instead of `PermalinkViewModel.Table` being a `dyanamic` type which held the caption and table markup as a json object, we now introduce a new `PermalinkTableViewModel` type which has the caption, json object, and footnotes.

### Other changes

- Add new `FootnotesViewModelBuilder`.
- Rename `FrontEndService` methods `CreateUniversalTable` to `CreateTable`.
- Add realistic frontend responses to unit tests.
- Add unit test snapshots for uploaded snapshot json.


src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Services/FrontEndServiceTests.cs
```diff
@@ -18,8 +18,44 @@ public class FrontEndServiceTests
 {
     private const string HttpClientName = "PublicApp";
 
+    private const string TableJson = @"
```
duncan-at-hiveit, line 21: Nice!

Just wondering if we were to do this again using larger JSON examples, would it be preferable to have the JSON as an actual `.json` file within the `Resources` folder and grabbing it up with ` var jsonPath = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)!,"Resources" + Path.DirectorySeparatorChar + filename);` and reading in the file contents from there?  We do similar in `ProcessorStage3Tests.cs`, for example.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
```diff
@@ -81,10 +81,8 @@ public class PermalinkService : IPermalinkService
                 .OnSuccessCombineWith(_ => DownloadPermalink(permalinkId, cancellationToken))
                 .OnSuccess(async tuple =>
                 {
-                    var (permalink, universalTable) = tuple;
-
-                    PermalinkViewModel viewModel = await BuildViewModel(permalink, universalTable);
-                    return viewModel;
+                    var (permalink, table) = tuple;
```
duncan-at-hiveit, line 84: Wish you could destructure tuples in the lambda expression itself!

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
```diff
@@ -237,26 +241,33 @@ public class PermalinkService : IPermalinkService
...
+                            {
+                                Footnotes = legacyPermalink.FullTable.SubjectMeta.Footnotes
+                            };
 
                             await UploadSnapshot(permalink,
-                                legacyPermalink.FullTable.Results,
-                                csvMeta,
-                                universalTable,
-                                cancellationToken);
+                                observations: legacyPermalink.FullTable.Results,
```
duncan-at-hiveit, line 262: I see you've added all named parameters here, but is it necessary given all the different types?  I'm happy to start adopting this across the board if we figure it's better to do so, so just curious what the driver is here?

src/GovUk.Education.ExploreEducationStatistics.Data.Api/ViewModels/PermalinkTableViewModel.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Collections.Generic;
+using GovUk.Education.ExploreEducationStatistics.Data.Services.ViewModels;
+using Newtonsoft.Json.Linq;
+
+namespace GovUk.Education.ExploreEducationStatistics.Data.Api.ViewModels;
+
+/// <summary>
+/// Permalink table which is generated by the frontend Next.js API and persisted in blob storage.
+/// </summary>
+public record PermalinkTableViewModel
```
duncan-at-hiveit, line 12: Could this be shorthanded somehow, like:

```
public record PermalinkTableViewModel(string Caption, List<FootnoteViewModel> Footnotes, JObject Json)
```

src/GovUk.Education.ExploreEducationStatistics.Data.Api/ViewModels/PermalinkViewModel.cs
```diff
@@ -19,5 +19,5 @@ public record PermalinkViewModel
     [JsonConverter(typeof(StringEnumConverter))]
     public PermalinkStatus Status { get; init; }
 
-    public dynamic Table { get; init; }
+    public PermalinkTableViewModel Table { get; init; } = new();
```
duncan-at-hiveit, line 22: Cool

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
benoutram: Yeah that is a much needed addition! 🙂

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
```diff
@@ -128,25 +128,31 @@ public class PermalinkService : IPermalinkService
                         cancellationToken
                     );
 
-                    await Task.WhenAll(universalTableTask, csvMetaTask);
+                    await Task.WhenAll(frontendTableTask, csvMetaTask);
 
-                    var universalTableResult = universalTableTask.Result;
+                    var frontendTableResult = frontendTableTask.Result;
```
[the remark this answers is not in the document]
benoutram: I tried this at the time of writing the original code but I don't think we can. `TSuccess` is different across the two different tasks.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
```diff
@@ -237,26 +241,33 @@ public class PermalinkService : IPermalinkService
...
                             return _permalinkCsvMetaService.GetCsvMeta(legacyPermalink, cancellationToken);
                         })
                         .OnSuccessVoid(async tuple =>
                         {
-                            var (legacyPermalink, universalTable, csvMeta) = tuple;
+                            var (legacyPermalink, table, csvMeta) = tuple;
+
+                            // To avoid the frontend processing and returning the footnotes unnecessarily,
+                            // add them to the table view model directly
+                            table = table with
```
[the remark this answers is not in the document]
benoutram: Cool, yep, made this change. Thanks.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
benoutram: I've only really started doing this because I saw others doing it more often 😆. I do see some extra benefits of using named params though.

In addition to the protection offered when params are of the same type, the other benefits to me are:

- Readability especially where the argument name doesn't match the parameter name like we see here, `observations` vs `legacyPermalink.FullTable.Results`.
- Makes refactoring easier when parameter order changes as we don't need to reorder all of the arguments.

I suppose it actually makes things less readable if the names are all identical and it also seems a bit overkill if there's only a single param.

We could chat about it at the next Dev's team meeting to see if we want to adopt it?

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Services/FrontEndServiceTests.cs
benoutram: I was already considering doing it so I've made this change now! It also means the same table test json isn't being duplicated across two test classes 🙂 .

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Services/PermalinkService.cs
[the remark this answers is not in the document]
duncan-at-hiveit: Humm... I wonder if we can change `OnSuccessAll` to make it work.  I may give it a go after this code is in, but one for another time.  Thanks!

duncan-at-hiveit (approved):

PR #4043, opened 2023-05-15
EES-4275 Add support to compress and decompress blobs. Confgure Permalink snapshots to use compression

benoutram:
This PR adds support to `BlobStorageService` to compress and decompress blobs, providing two available algorithms, [Gzip](https://learn.microsoft.com/en-us/dotnet/api/system.io.compression.gzipstream?view=net-7.0) or [ZStandard](http://facebook.github.io/zstd/).

Currently this uses the ZstdSharp port of ZStandard to С# but it should be trivial to swap it out for one of the ZStandard C# alternatives.

Methods `BlobStorageService.UploadAsJson` and `BlobStorageService.UploadStream` now take an optional `contentEncoding` parameter. If this is specified (using one of the supported encodings, either `ContentEncodings.Gzip` or `ContentEncodings.Json`) then the content will be compressed on upload.

Methods `BlobStorageService.DownloadBlobText`, `BlobStorageService.DownloadToStream` and `BlobStorageService.GetDeserializedJson` are changed to be aware of compressed blobs. If a blob has property `ContentEncoding` equal to one of the supported encodings, it will be automatically be decompressed.

Method `BlobStorageService.DownloadToStream` is given a `decompress` flag which can be used to ignore compression if there's a reason to stream a blob untouched. In case a blob is known to never be compressed the same flag can be used to avoid the internal API call which occurs to get the blob properties. This extra call is required to find the content encoding.

This PR changes the new permalink snapshot feature to use ZStandard compression using this new feature.
It also alters the migration of all legacy permalinks to use ZStandard compression.

# Other changes
- `BlobStorageService` methods `DownloadBlobText`, `GetDeserializedJson` and `GetDeserializedJson<T>` are changed to return `NotFoundResult` when the blob is not found rather than throwing an exception.
- `BlobStorageService` methods `GetDeserializedJson`, `GetDeserializedJson<T>`, and `UploadAsJson` are changed to accept optional `JsonSerializerSettings`.
- Improve coverage of methods supporting `CancellationToken`.
- Replace occurrences of `CancellationToken? cancellationToken = null` with `CancellationToken cancellationToken = default` , since `CancellationToken` is a value type which can't be null.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseFileService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseFileService.cs
@@ -190,7 +190,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
             Guid releaseId,
             Stream outputStream,
             IEnumerable<Guid>? fileIds = null,
-            CancellationToken? cancellationToken = null)
+            CancellationToken cancellationToken = default)
         {
             return await _persistenceHelper.CheckEntityExists<Release>(
                     releaseId,
@@ -220,14 +220,14 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
             List<ReleaseFile> releaseFiles,
             Release release,
             Stream outputStream,
-            CancellationToken? cancellationToken)
+            CancellationToken cancellationToken)
         {
             using var archive = new ZipArchive(outputStream, ZipArchiveMode.Create);
 
             foreach (var releaseFile in releaseFiles)
             {
                 // Stop immediately if we receive a cancellation request
-                if (cancellationToken?.IsCancellationRequested == true)
+                if (cancellationToken.IsCancellationRequested)
                 {
                     return;
                 }
--- a/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/MockBlobClientExtensions.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/MockBlobClientExtensions.cs
@@ -1,0 +1,70 @@
+﻿#nullable enable
+using System;
+using System.Threading;
+using Azure;
+using Azure.Storage.Blobs;
+using Azure.Storage.Blobs.Models;
+using Moq;
+using Moq.Language.Flow;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Tests.Extensions;
+
+public static class MockBlobClientExtensions
+{
+    public static IReturnsResult<BlobClient> SetupDownloadContentAsync(
+        this Mock<BlobClient> blobClient,
+        string content,
+        string? contentEncoding = null,
+        CancellationToken cancellationToken = default)
+    {
+        var downloadDetails = BlobsModelFactory.BlobDownloadDetails(
+            lastModified: default,
+            metadata: default,
+            contentRange: default,
+            contentEncoding: contentEncoding,
+            cacheControl: default,
+            contentDisposition: default,
+            contentLanguage: default,
+            blobSequenceNumber: default,
+            copyCompletedOn: default,
+            copyStatusDescription: default,
+            copyId: default,
+            copyProgress: default,
+            copySource: default,
+            copyStatus: default,
+            leaseDuration: default,
+            leaseState: default,
+            leaseStatus: default,
+            acceptRanges: default,
+            blobCommittedBlockCount: default,
+            isServerEncrypted: default,
+            encryptionKeySha256: default,
+            encryptionScope: default,
+            blobContentHash: default,
+            tagCount: default,
+            versionId: default,
+            isSealed: default,
+            objectReplicationSourceProperties: default,
+            objectReplicationDestinationPolicy: default
+        );
+
+        var blobDownloadResult = BlobsModelFactory.BlobDownloadResult(
+            content: BinaryData.FromString(content),
+            details: downloadDetails);
+
+        var response = new Mock<Response<BlobDownloadResult>>(MockBehavior.Strict);
+        response.SetupGet(r => r.Value)
+            .Returns(blobDownloadResult);
+
+        return blobClient.Setup(s => s.DownloadContentAsync(cancellationToken))
+            .ReturnsAsync(response.Object);
+    }
+
+    public static IReturnsResult<BlobClient> SetupDownloadContentAsyncNotFound(
+        this Mock<BlobClient> blobClient,
+        CancellationToken cancellationToken = default)
+    {
+        return blobClient.Setup(s => s.DownloadContentAsync(cancellationToken))
+            .ThrowsAsync(new RequestFailedException(status: 404, message: "Blob not found"));
+    }
+}
--- a/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/MockBlobStorageServiceExtensions.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/MockBlobStorageServiceExtensions.cs
@@ -220,5 +220,19 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Tests.Extensions
                     s.GetDeserializedJson(container, path, type, settings, cancellationToken))
                 .ThrowsAsync(exception);
         }
+
+        public static IReturnsResult<IBlobStorageService> SetupUploadAsJson<T>(
+            this Mock<IBlobStorageService> service,
+            IBlobContainer container,
+            string path,
+            T content,
+            string? contentEncoding = null,
+            JsonSerializerSettings? settings = null,
+            CancellationToken cancellationToken = default)
+        {
+            return service.Setup(s =>
+                    s.UploadAsJson(container, path, content, contentEncoding, settings, cancellationToken))
+                .Returns(Task.CompletedTask);
+        }
     }
 }
--- a/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/MockBlobStorageServiceExtensions.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/MockBlobStorageServiceExtensions.cs
@@ -1,15 +1,17 @@
 #nullable enable
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Threading;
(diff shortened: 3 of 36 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseFileService.cs
```diff
@@ -220,14 +220,14 @@ await fileIds.Select(fileId => CheckFileExists(releaseId, fileId, DeletableFileT
...
-            CancellationToken? cancellationToken)
+            CancellationToken cancellationToken)
         {
             using var archive = new ZipArchive(outputStream, ZipArchiveMode.Create);
 
             foreach (var releaseFile in releaseFiles)
             {
                 // Stop immediately if we receive a cancellation request
-                if (cancellationToken?.IsCancellationRequested == true)
+                if (cancellationToken.IsCancellationRequested)
```
duncan-at-hiveit, line 230: Not as part of this PR, but should we be adding parameters to BlobStorageService methods to take optional CancellationTokens as well in the future so we can pass them along?  Presumably the underlying SDK libs that we use will have the ability to accept CancellationTokens too.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/MockBlobClientExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+
+public static class MockBlobClientExtensions
+{
+    public static IReturnsResult<BlobClient> SetupDownloadContentAsync(
+        this Mock<BlobClient> blobClient,
+        string content,
+        string? contentEncoding = null,
+        CancellationToken cancellationToken = default)
+    {
+        var downloadDetails = BlobsModelFactory.BlobDownloadDetails(
```
duncan-at-hiveit, line 20: Yikes!

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/MockBlobStorageServiceExtensions.cs
```diff
@@ -131,7 +152,73 @@ public static class MockBlobStorageServiceExtensions
         {
             return service.Setup(s =>
                     s.DownloadBlobText(container, path, cancellationToken))
-                .ThrowsAsync(new FileNotFoundException("Could not find blob"));
+                .ReturnsAsync(new NotFoundResult());
```
duncan-at-hiveit, line 155: Nice behaviour change

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Services/BlobCacheServiceTests.cs
```diff
@@ -57,8 +50,13 @@ public async Task SetItemAsync()
 
             var blobStorageService = new Mock<IBlobStorageService>(MockBehavior.Strict);
 
-            blobStorageService.Setup(mock =>
-                    mock.UploadAsJson(PublicContent, cacheKey.Key, "test item", null))
+            blobStorageService.Setup(mock => mock.UploadAsJson(
```
duncan-at-hiveit, line 53: Would this be worth adding to MockBlobStorageServiceExtensions?

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/CompressionUtilsTests.cs
```diff
@@ -0,0 +1,173 @@
...
+    {
+        // Compress some content to a stream using the method under test
+        await using var compressedStream = new MemoryStream();
+        await CompressionUtils.CompressToStream(
+            stream: Content.ToStream(),
+            targetStream: compressedStream,
+            contentEncoding: ContentEncodings.Gzip);
+
+        Assert.Equal(0, compressedStream.Position);
+        Assert.True(compressedStream.Length > 0);
```
duncan-at-hiveit, line 40: Worth checking that the compressedStream.Length is less than Content.ToStream().Length at this point?

duncan-at-hiveit: Or at least that it's not still equal to the original Content Stream at this point?

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/CompressionUtilsTests.cs
```diff
@@ -0,0 +1,173 @@
...
+    {
+        // Compress some content to a stream using the method under test
+        await using var compressedStream = new MemoryStream();
+        await CompressionUtils.CompressToStream(
+            stream: Content.ToStream(),
+            targetStream: compressedStream,
+            contentEncoding: ContentEncodings.Zstd);
+
+        Assert.Equal(0, compressedStream.Position);
+        Assert.True(compressedStream.Length > 0);
```
duncan-at-hiveit, line 64: Same here?

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Services/PermalinkServiceLegacyTests.cs
```diff
@@ -666,129 +668,6 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
         }
     }
 
-    [Fact]
-    public async Task GetLegacy_LegacyLocationsFieldIsTransformed()
```
duncan-at-hiveit, line 670: What happened to this test?

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/CompressionUtilsTests.cs
benoutram: I think, depending on content, the compressed stream can be longer than the original content stream. I've gone with checking that it's not equal to the original content stream.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Services/BlobCacheServiceTests.cs
benoutram: Good idea, made this change.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/MockBlobClientExtensions.cs
benoutram: I saw at least 1 other issue raised about a similar factory method not having default methods making using them difficult, but until this one gets the same treatment I think we're stuck with this!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseFileService.cs
benoutram: Yes I think it's a good idea to add these in whenever we touch these methods. The storage and other libs of the SDK that we use supports CancellationTokens where appropriate.

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Services/PermalinkServiceLegacyTests.cs
benoutram: It's removed because of the internal change made to `PermalinkService.DownloadLegacyPermalink`.

Instead of using `BlobStorageService.SetupDownloadBlobText` to download json and then deserialise it to a LegacyPeramlink, it now uses `BlobStorageService.GetDeserializedJson`  to get a LegacyPermalink, passing the `JsonSerializerSettings` settings.

`BlobStorageService` behaviour is set up to return a `LegacyPermalink` which doesn't require the unit test creating any of the json content which we saw here in the test being removed.

We trust that the settings will do the job of transforming legacy locations which should be ok because `PermalinkResultSubjectMetaJsonConverter` has its own unit test coverage for this.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Services/BlobCacheServiceTests.cs
benoutram: Adding the extension method to be used here is fine, and that change is in, but I was just looking at whether this new method could also be used to replace setup behaviour in `PermalinkServiceTests`.

There `path` and `content` are matchers like `It.Is<string>(path => path.EndsWith(".json.zst"))` and `It.IsAny<PermalinkTableViewModel>()`.

It seems like we can't do this, although passing a matcher will happily compile.

The only way I could get this to work would be with an extra extension method, something like this:

```
        public static IReturnsResult<IBlobStorageService> SetupUploadAsJson<T>(
            this Mock<IBlobStorageService> service,
            IBlobContainer container,
            Expression<Func<string, bool>> pathMatcher,
            Expression<Func<T, bool>> contentMatcher,
            string? contentEncoding = null,
            JsonSerializerSettings? settings = null,
        {
            return service.Setup(s =>
                    s.UploadAsJson(container,
                        It.Is(pathMatcher),
                        It.Is(contentMatcher),
                        contentEncoding,
                        settings,
                        cancellationToken))
                .Returns(Task.CompletedTask);
```

This can be used something like this

```
            blobStorageService.SetupUploadAsJson<PermalinkTableViewModel>(
                    container: BlobContainers.PermalinkSnapshots,
                    pathMatcher: path => path.EndsWith(".json.zst"),
                    contentMatcher: content => true,
                    contentEncoding: ContentEncodings.Zstd)
```

This works, but I decided to leave it as it is for now. Just wondering if you have an opinion on this or know of any cleaner way to do this?

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Services/PermalinkServiceLegacyTests.cs
duncan-at-hiveit: Cool ta

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Services/BlobCacheServiceTests.cs
duncan-at-hiveit: As per chat yesterday, seems to be all down to Expression parsing in Moq and not much we can do, so it's a shame but it's all good the way it is for now thanks.

duncan-at-hiveit (approved):

PR #4048, opened 2023-05-22
EES-3516 - added tests to assert that thread exhaustion issues are no longer occurring

duncan-at-hiveit:
This PR:
* Adds in tests to detect that thread exhaustion issues are no long occurring after moving to the new Universal Wrapper-based CacheAspect / CacheAttribute code

There are 2 broad approaches that are being used to test the new behaviour.

The first approach gathers a list of the events that occur when a Cache-annotated controller method is called, and asserts that it is what we expect to see when the aspect is non-blocking.  We do this for both a scenario when there is no previously cached item and a scenario when there is.

The second approach is a brute-force attempt to cause thread escalation and excessive processing time, by generating several concurrent HTTP requests and monitoring the thread pool and total processing time.

Between these 2 approaches, we should be relatively assured that any accidental re-introduction of blocking code in the CacheAspect will be flagged up.

## Making it deliberately fail

By "accidentally" introducing some blocking code into CacheAttribute, we can make these tests fail every time.  Replacing the `CacheAttribute.WrapAsync` method with:

```
        protected override Task<T> WrapAsync<T>(Func<object[], Task<T>> target, object[] args, AspectEventArgs eventArgs)
        {
            var unboxedResultType = typeof(T).GetUnboxedResultTypePath().Last();

            var cacheKey = GetCacheKey(Key, args, eventArgs.Method);

            object? cachedResult = null;
            if (!ForceUpdate)
            {
                cachedResult = GetAsync(cacheKey, unboxedResultType).Result;
            }

            if (cachedResult?.TryBoxToResult(typeof(T), out cachedResult) == true)
            {
                return Task.FromResult((T)cachedResult);
            }

            try
            {
                var result = target(args).Result;

                if (!result.TryUnboxResult(out cachedResult) || cachedResult is null)
                {
                    return Task.FromResult(result);
                }

                Task.WaitAll(SetAsync(cacheKey, cachedResult));

                return Task.FromResult(result);
            }
            catch (Exception exception)
            {
                return OnException<Task<T>>(eventArgs, exception);
            }
        }
``` 

we then run the unit tests, and all 3 tests fail with:

```

  Failed GovUk.Education.ExploreEducationStatistics.Common.Tests.Cache.CacheIntegrationTests.CachingControllerMethod [771 ms]
  Error Message:
   Assert.Equal() Failure
          ↓ (pos 0)
Expected: AsyncActionFilter OnActionExecutionAsync ···
Actual:   Returning no cached item with key TestCac···
          ↑ (pos 0)




[xUnit.net 00:00:02.13]     GovUk.Education.ExploreEducationStatistics.Common.Tests.Cache.CacheIntegrationTests.CachingControllerMethod_ExistingCachedItem [729 ms]
  Error Message:
   Assert.Equal() Failure
          ↓ (pos 0)
Expected: AsyncActionFilter OnActionExecutionAsync ···
Actual:   Returning cached item ControllerResponse ···
          ↑ (pos 0)




  Failed GovUk.Education.ExploreEducationStatistics.Common.Tests.Cache.CacheIntegrationTests.NoThreadExhaustion [1 s]
  Error Message:
   System.Exception : Number of additional concurrent worker threads exceeds limit of 20. 23 more threads are currently in use



```

So here we see that a) the expected order of events is now incorrect, and b) we quickly fail due to too many threads being produced in the thread pool during the running of the NoThreadExhaustion test method.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/CacheAspectThreadManagementIntegrationTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/CacheAspectThreadManagementIntegrationTests.cs
@@ -1,0 +1,587 @@
+#nullable enable
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using GovUk.Education.ExploreEducationStatistics.Common.Cache;
+using GovUk.Education.ExploreEducationStatistics.Common.Cache.Interfaces;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces;
+using GovUk.Education.ExploreEducationStatistics.Common.Tests.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Common.Tests.Fixtures;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.AspNetCore.Mvc.Testing;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Tests.Cache;
+
+[Collection(CacheTestFixture.CollectionName)]
+public class CacheAspectThreadManagementIntegrationTests : 
+    IClassFixture<TestApplicationFactory<TestStartup>>,
+    IClassFixture<CacheTestFixture>,
+    IDisposable
+{
+    // An artificial delay that we will add to dummy async sections of our test setup, including getting
+    // cached items setting cached items and time spent in Controller methods. 
+    public const int AsyncOperationDelayMillis = 250;
+    
+    // A reasonable amount of time for code to reach the next step of the HTTP request processing workflow if
+    // no async delaying action appears between the two steps. 
+    private const int NoAsyncDelayTimeMillis = 50;
+    
+    // An error tolerance in the time it takes for the code after a Task.Delay() to  continue running. Task.Delay() is
+    // not 100% accurate in terms of delay and so it is possible for the underlying mechanisms to wake early by a few
+    // milliseconds.
+    private const int TaskDelayErrorToleranceMillis = 15;
+
+    private readonly WebApplicationFactory<TestStartup> _testApp;
+
+    public CacheAspectThreadManagementIntegrationTests(TestApplicationFactory<TestStartup> testApp)
+    {
+        _testApp = testApp;
+    }
+
+    public void Dispose()
+    {
+        BlobCacheAttribute.ClearServices();
+    }
+
+    /// <summary>
+    /// This test is testing the flow of events is as we would expect to see it when the CacheAspect code is
+    /// non-blocking. It fires off an HTTP request to our TestController which uses the BlobCache attribute on its
+    /// called method, and records the flow of events from there. It compares these with an expected order that would
+    /// be different had the CacheAspect introduced blocking code.
+    ///
+    /// Unfortunately we can't tell if there is further blocking code in one of the multiple continuations that will be
+    /// run as a part of this, but they would get picked up in the <see cref="NoThreadExhaustion"/> test very quickly
+    /// if they existed. 
+    /// </summary>
+    [Fact]
+    public async Task CachingControllerMethod()
+    {
+        // This is a list of events that we will capture in order to verify that the flow of code is showing what
+        // we expect to see, especially in terms of allowing threads to be freed up and not be blocked. 
+        var events = new List<Event>();
+        
+        var blobCacheService = new TestBlobCacheService(events);
+        var app = SetupApp(blobCacheService, events);
+
+        var client = app.CreateClient();
+
+        var expectedCacheKey = new TestCacheKey("myRequest");
+        var expectedControllerResult = new ControllerResponse("myRequest");
+        
+        var response = await client.GetAsync("/test/caching-method?key=myRequest");
+
+        response.AssertOk(expectedControllerResult);
+
+        // The first 4 items in this list of events are the most important orders of Events here. They show that the
+        // CacheAspect is non-blocking in this scenario, as we can see that:
+        //
+        // 1. The AsyncActionFilter.OnActionExecutionAsync() method is called on its way to calling the Controller
+        //    method under test - this is standard MVC behaviour. This work is done in what we'll call the Controller
+        //    Request thread.
+        var asyncActionFilterCalled = 
+            AssertEventMessageEquals(events[0], $"{nameof(AsyncActionFilter)} OnActionExecutionAsync called");
+        
+        // 2. The CacheAspect AOP is encountered before hitting the actual Controller method itself, as we can see from
+        //    the request for any previously existing cached item. This is also part of the original Controller Request
+        //    thread.
+        // 3. The caching code which aims to look up any existing cached items for the given key is encountered. This
+        //    code is asynchronous (as our BlobCacheService code is) and so we are hoping to see evidence at this point
+        //    that this is not causing the Controller Request thread to be blocked.
+        var requestingCachedItem = 
+            AssertEventMessageEquals(events[1], $"Requesting cached item with key {expectedCacheKey}");
+        
+        // We expect this to be called almost immediately after the AsyncActionFilter call.
+        Assert.InRange(
+            requestingCachedItem.Millis, 
+            asyncActionFilterCalled.Millis, 
+            asyncActionFilterCalled.Millis + NoAsyncDelayTimeMillis);
+
+        // 4. We then see this evidence - the AsyncActionFilter.OnActionExecutionAsync() completes, with a
+        //    non-Completed Task as its result. This indicates that there is nothing in the CacheAspect code any more
+        //    that is tying up the Controller Request thread, which is now freed up to be used again.
+        var asyncActionFilterCompleted = 
+            AssertEventMessageEquals(events[2], $"{nameof(AsyncActionFilter)} OnActionExecutionAsync " +
+                                                $"completed - returned Task IsCompleted = False");
+        
+        // We expect this to be called almost immediately after the cached item request call.
+        Assert.InRange(
+            asyncActionFilterCompleted.Millis, 
+            requestingCachedItem.Millis, 
+            requestingCachedItem.Millis + NoAsyncDelayTimeMillis);
+
+        // 5. The flow then continues, showing that the cache finds no existing items with the given cache key, the
+        //    original Controller method is called to get the actual result, this is then cached and the result given to
+        //    the AsyncResultFilter for returning with the HTTP response.
+        //
+        // Previously, when the CacheAspect code was blocking the thread, we would see all of the work up to and
+        // including the setting of the cached item being done *within* the AsyncActionFilter.OnActionExecutionAsync()
+        // method, with the OnActionExecutionAsync method completing just prior to the AsyncResultFilter kicking in to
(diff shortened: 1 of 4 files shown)


src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/CacheAspectThreadManagementIntegrationTests.cs
```diff
@@ -0,0 +1,599 @@
...
+    /// time to process them all and the growth of the Worker Thread pool during the process.
+    ///
+    /// In a scenario where there was blocking code accidentally introduced, the current Worker Thread count would
+    /// immediately increase as all the available threads get blocked. The time taken to complete processing all of
+    /// the requests would also greatly exceed the expected maximum time to complete. This has been proven by
+    /// deliberately adding blocking back into the <see cref="CacheAttribute.WrapAsync{T}"/> implementation and seeing
+    /// these tests fail.
+    /// </summary>
+    /// <exception cref="Exception"></exception>
+    [Fact]
```
ntsim, line 307: Shouldn't this have a timeout to avoid hanging indefinitely if the loop never ends?

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/CacheAspectThreadManagementIntegrationTests.cs
```diff
@@ -0,0 +1,599 @@
...
+        }
+    }
+}
+
+/// <summary>
+/// This TestController has a single async method that has a BlobCache attribute.
+/// This method uses an artificial delay to simulate doing some async work e.g. calling some code that interacts
+/// with the database. 
+/// </summary>
+public class TestController : ControllerBase
```
ntsim, line 466: We've got a bunch of public classes here and below that will be exposed to all testing code unnecessarily. This isn't good practice and will just generally pollute code completions.

These should be marked as `internal` (but will still be exposed to the entire `Common.Tests` namespace) or should just be nested into the test class above.

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/CacheAspectThreadManagementIntegrationTests.cs
duncan-at-hiveit: (that's where the `<exception cref="Exception"></exception> came from in the method doco, which I will also embellish upon)

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/CacheAspectThreadManagementIntegrationTests.cs
duncan-at-hiveit: Good spot - almost surprised these weren't flagged up by Rider.  They are now nested and private.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/CacheAspectThreadManagementIntegrationTests.cs
```diff
@@ -0,0 +1,589 @@
...
+    private const int NoAsyncDelayTimeMillis = 50;
+    
+    // An error tolerance in the time it takes for the code after a Task.Delay() to  continue running. Task.Delay() is
+    // not 100% accurate in terms of delay and so it is possible for the underlying mechanisms to wake early by a few
+    // milliseconds.
+    private const int TaskDelayErrorToleranceMillis = 15;
+
+    // A list of events that we will capture in order to verify that the flow of code is showing what we expect
+    // to see, especially in terms of allowing threads to be freed up and not be blocked. 
+
```
benoutram, line 44: Think the intention is to comment a list here but there's no list 🙂

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/CacheAspectThreadManagementIntegrationTests.cs
```diff
@@ -0,0 +1,589 @@
...
+        public List<Event> GetEvents()
+        {
+            return _events;
+        }
+
+        public void Clear()
+        {
+            _events.Clear();
+        }
+    }
```
benoutram, line 569: Any need to restrict the methods of `EventList` from that of a regular `List`? Just wondering if this would work?

```
    public class EventList : List<Event>
    {
    }
```

duncan-at-hiveit: Hey Ben - good call.  This as with a few other things in this PR had evolved from something simple into something more complex before then going back to something simple again!

I've replaced the whole lot with even simpler `var events = new List<Event>()` all over.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/CacheAspectThreadManagementIntegrationTests.cs
duncan-at-hiveit: Ahh yup, so this was when EventList was a static field here, but I left the comment behind!  I have moved the comment to where the new List<Event> variables are (where they're actually useful to have for the purposes of understanding the relevant tests anyway).

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/CacheAspectThreadManagementIntegrationTests.cs
duncan-at-hiveit: Remembered that the Controller needs to remain `public` and top-level in order for `AddMvc` functionality to pick it up, and therefore a bunch of its dependent classes also need to remain public, but I've nested and internalised what I can.  Interestingly `TestCacheKey` is OK to remain `internal` - I guess as we're only referring to its type rather than attempting to use it as a parameter or return type.

ntsim: Seems like it's possible to make private nested classes work: https://tpodolak.com/blog/2020/06/22/asp-net-core-adding-controllers-directly-integration-tests/

See '3. Handling private controllers'

duncan-at-hiveit: Ooh, nice!  Added that to our existing WebHostBuilderExtensions class and it seems to like it :+1:  Cheers!

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/CacheAspectThreadManagementIntegrationTests.cs
duncan-at-hiveit: All done

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/CacheAspectThreadManagementIntegrationTests.cs
```diff
@@ -0,0 +1,570 @@
...
+        // The first 4 items in this list of events are the most important orders of Events here. They show that the
+        // CacheAspect is non-blocking in this scenario, as we can see that:
+        //
+        // 4. We then see this evidence - the AsyncActionFilter.OnActionExecutionAsync() completes, with a
+        //    non-Completed Task as its result. This indicates that there is nothing in the CacheAspect code any more
+        //    that is tying up the Controller Request thread, which is now freed up to be used again.
+        // Previously, when the CacheAspect code was blocking the thread, we would see all of the work up to and
+        // including the setting of the cached item being done *within* the AsyncActionFilter.OnActionExecutionAsync()
+        // method, with the OnActionExecutionAsync method completing just prior to the AsyncResultFilter kicking in to
+        // return the HTTP response to the caller.
```
ntsim, line 234: Not sure what's gone on here, but have these notes been mangled somehow?

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/CacheAspectThreadManagementIntegrationTests.cs
```diff
@@ -0,0 +1,570 @@
...
+        }
+
+        public Task DeleteCacheFolderAsync(IBlobCacheKey cacheFolderKey)
+        {
+            throw new NotImplementedException();
+        }
+    }
+    
+    // ReSharper disable once ParameterOnlyUsedForPreconditionCheck.Local
+    private static Event AssertEventMessageEquals(Event e, string expectedMessage)
```
ntsim, line 525: Not sure we need this as it's not really adding anything. Would expected that calling `Assert.Equals` would suffice.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/CacheAspectThreadManagementIntegrationTests.cs
```diff
@@ -0,0 +1,570 @@
...
+        // Previously, when the CacheAspect code was blocking the thread, we would see all of the work up to and
+        // including the setting of the cached item being done *within* the AsyncActionFilter.OnActionExecutionAsync()
+        // method, with the OnActionExecutionAsync method completing just prior to the AsyncResultFilter kicking in to
+        // return the HTTP response to the caller.
+        
+        // 1. The AsyncActionFilter.OnActionExecutionAsync() method is called on its way to calling the Controller
+        //    method under test - this is standard MVC behaviour. This work is done in what we'll call the Controller
+        //    Request thread.
+        var asyncActionFilterCalled = 
+            AssertEventMessageEquals(events[0], ($"{nameof(AsyncActionFilter)} OnActionExecutionAsync called"));
```
ntsim, line 240: We seem to have done this here and a bunch of other places later on, but we're wrapping strings with parentheses unnecessarily

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/CacheAspectThreadManagementIntegrationTests.cs
duncan-at-hiveit: I just want to be able to assign them to meaningful variable names without having to take up twice the space (in line with how we do things like `var viewModel = response.AssertRight();` for example.  Otherwise, it's Assert, then assign, then Assert, then assign etc.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/CacheAspectThreadManagementIntegrationTests.cs
duncan-at-hiveit: Ahh yeah, left over from a regex search and replace - well spotted.  Removed, thanks.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/CacheAspectThreadManagementIntegrationTests.cs
duncan-at-hiveit: Yup indeed, reordered now thanks

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Cache/CacheAspectThreadManagementIntegrationTests.cs
ntsim: Okay fair enough! Probably wouldn't suggest making a habit of doing this in other places unless we actually need it though :wink:

PR #4067, opened 2023-06-01
EES-3229 Add ability to replace ancillary files

markhiveit:
This PR adds the ability to replace ancillary files. Previously, to do this, a user would have to remove the ancillary file completely and create a new ancillary file.

No UI tests have been written as part of this PR. They will be created in EES-4315.


luke-h1 (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleaseFileController.cs
```diff
@@ -146,6 +146,19 @@ public async Task<ActionResult<IEnumerable<FileInfo>>> GetAncillaryFiles(Guid re
                 .HandleFailuresOrOk();
         }
 
+        [HttpPost("release/{releaseId}/ancillary/{fileId}/replace")]
```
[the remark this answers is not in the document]
markhiveit: Fixed. I've also made the same change for every controller across the entire codebase. 

I'll leave this change in a separate commit so it's easier to inspect.

src/explore-education-statistics-admin/src/pages/release/data/ReleaseAncillaryFileReplacePage.tsx
```diff
@@ -8,28 +8,30 @@ import useAsyncHandledRetry from '@common/hooks/useAsyncHandledRetry';
...
-const ReleaseAncillaryFileReplacePage = ({
-  history,
-  match: {
-    params: { publicationId, releaseId, fileId },
-  },
-}: RouteComponentProps<ReleaseAncillaryFileReplaceRouteParams>) => {
+const ReleaseAncillaryFileReplacePage = () => {
+  const history = useHistory();
+
+  const { publicationId, releaseId, fileId } = useParams<{
```
luke-h1, line 29: We can just use the `ReleaseAncillaryFileReplaceRouteParams` interface here instead of having to specify a new type

luke-h1 (approved):
Looks ok to me, one small comment & then it's good to go as far as I can see

src/explore-education-statistics-admin/src/pages/release/data/components/AncillaryFileForm.tsx
```diff
@@ -0,0 +1,171 @@
...
+              );
+            },
+          }),
+        summary: Yup.string().required('Enter a summary'),
+        file: Yup.file().minSize(0, 'Choose a file that is not empty'),
+      }).concat(validationSchema ?? Yup.object())}
+    >
+      {form => (
+        <Form id={formId}>
+          <div style={{ position: 'relative' }}>
```
amyb-hiveit, line 120: Is this div needed? I can't see any difference if I remove it.

amyb-hiveit (changes requested):
Just one small thing, otherwise looks good.

src/explore-education-statistics-admin/src/pages/release/data/components/AncillaryFileForm.tsx
ntsim: Removed now. This was just a hangover from previously using an overlay loading spinner on submit.

amyb-hiveit (approved):

amyb-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseFileService.cs
```diff
@@ -342,6 +342,45 @@ await fileIds.Select(fileId => CheckFileExists(releaseId, fileId, DeletableFileT
...
+                        releaseId: releaseId,
+                        filename: newFile.FileName,
+                        contentLength: newFile.Length,
+                        contentType: newFile.ContentType,
+                        type: Ancillary,
+                        createdById: _userService.GetUserId(),
+                        name: oldReleaseFile.Name,
+                        summary: oldReleaseFile.Summary);
+
+                    _contentDbContext.Remove(oldReleaseFile);
```
ntsim, line 371: Seems to me like we should remove the old file as well as long as it is no longer linked to any other releases (i.e. amendments). We already do a similar thing in the `Delete` method.

This will also needs some extra tests to verify this behaviour.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseFileServiceTests.cs
```diff
@@ -2606,6 +2608,120 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
...
+                Assert.Equal("test@test.com", fileInfo.UserName);
+                Assert.InRange(DateTime.UtcNow.Subtract(fileInfo.Created.GetValueOrDefault()).Milliseconds, 0, 1500);
+            }
+
+            await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId))
+            {
+                // Old File shouldn't be removed
+                Assert.Single(await contentDbContext.Files
+                    .Where(f => f.Id == releaseFile.FileId)
+                    .ToListAsync());
```
ntsim, line 2696: Whilst this is probably incorrect behaviour (see other comment on deleting the old file), it'd probably be cleaner to check that something is missing by using an assertion like:

```cs
Assert.NotNull(await contentDbContext.Files
    .FirstOrDefaultAsync(f => f.Id == releaseFile.FileId));
```

ntsim (changes requested):

ntsim (approved):

PR #4072, opened 2023-06-05
EES-3311 Remove ambiguity over which Azure Storage instance is being connected to

markhiveit:
Previously, in various `Startup.cs` files, if we wired up a service that included `IBlobStorageService`, then we would have a "default" would be injected. The "default" would be injected in a `Startup.cs` like this: `services.AddTransient(provider => GetBlobStorageService(provider, "CoreStorage"));`

This has previously caused issues, as someone can easily include `IBlobStorageService` into a new service without having to explicitly state which Azure Storage instance they want.

This PR aims to address this issue with the introduction of two new interfaces, `IPrivateBlobStorageService` and `IPublicBlobStorageService`. Alongside this, we now also don't wire up `IBlobStorageService` in any Startup.cs file, such that the "default" Azure Storage instance for that project won't automatically get injected.

### Other notes:
- `BlobCacheService` still contains an `IBlobStorageService`, but this is no longer injected into `BlobCacheService`. The appropriate Azure Storage instance is now injected explicitly in `Startup.cs` when wiring up `BlobCacheService` itself. We might also want public/private interfaces for `BlobCacheService`, but I didn't want to change that here, as I figure this PR is large enough as it is.
- Some duplication of methods has been necessary in `MockBlobStorageServiceExtensions`. I tried various ways to avoid this, but I couldn't find a better approach.

image


src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/MockBlobStorageServiceExtensions.cs
```diff
@@ -37,27 +37,18 @@ public static class MockBlobStorageServiceExtensions
                 .ReturnsAsync(() => File.OpenRead(filePathToStream));
         }
 
-        public static IReturnsResult<IBlobStorageService> SetupListBlobs(
-            this Mock<IBlobStorageService> service,
+        public static IReturnsResult<IPublicBlobStorageService> SetupCheckBlobExists(
```
duncan-at-hiveit, line 40: If you don't want this repetition, you can use:

```
        public static IReturnsResult<T> SetupCheckBlobExists<T>(
            this Mock<T> service,
            IBlobContainer container,
            string path,
            bool exists)
        where T : class, IBlobStorageService
        {
            return service.Setup(s => s.CheckBlobExists(container, path))
                .ReturnsAsync(exists);
        }
```

src/GovUk.Education.ExploreEducationStatistics.Common/Services/BlobStorageService.cs
```diff
@@ -37,7 +37,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Services
     /// changes that make it difficult to upgrade when relying on lower-level
     /// details e.g. Azure SDK v12 is an entirely new package compared to v11.
     /// </summary>
-    public class BlobStorageService : IBlobStorageService
+    public class BlobStorageService : IPrivateBlobStorageService, IPublicBlobStorageService
```
duncan-at-hiveit, line 40: Aha interesting!  I would've assumed there'd be concrete PrivateBlobStorageService and PublicBlobStorageService classes that extended an abstract `BaseBlobStorageService` but this is a nice way of doing this with less classes and boilerplate :)

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Services/PublicReleaseFileBlobService.cs
```diff
@@ -16,36 +16,36 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Model.Services;
 
 public class PublicReleaseFileBlobService : IReleaseFileBlobService
```
duncan-at-hiveit, line 17: Feels like `PublicReleaseFileBlobService` and `PrivateReleaseFileBlobService` would be nicer handled in the same way as you've done with BlobStorageService, IPublicStorageService and IPrivateStorageService, but one for another PR.  Maybe we should think about breaking out a ticket for this?

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Startup.cs
```diff
@@ -29,19 +29,15 @@ public override void Configure(IFunctionsHostBuilder builder)
...
-                            connectionString,
-                            new BlobServiceClient(connectionString),
-                            provider.GetRequiredService<ILogger<BlobStorageService>>(),
-                            new StorageInstanceCreationUtil()
-                        );
-                        return blobStorageService;
-                    })
+                .AddSingleton<IPrivateBlobStorageService, BlobStorageService>(provider =>
+                {
+                    var privateConnectionString = GetConfigurationValue(provider, "CoreStorage");
```
duncan-at-hiveit, line 34: I guess one of the advantages of having PublicBlobStorageService and PrivateBlobStorageService concrete classes that extend BlobStorageService is that you could have the class itself extract the connection string that relates to it rather than having to wire it manually here.  Then you could write here instead:

```
    .AddSingleton<IPrivateBlobStorageService, PrivateBlobStorageService>
```

and be done with it.  Within PrivateBlobStorageService, it would take the Configuration as a constructor argument and extract the "CoreStorage" string from it.

duncan-at-hiveit: And in this case it would also create its own `new BlobServiceClient(privateConnectionString)` within its constructor too.

src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/Services/PublishingServiceTests.cs
```diff
@@ -407,23 +408,30 @@ public async Task PublishStagedReleaseContent()
...
                 releaseService ?? Mock.Of<IReleaseService>(MockBehavior.Strict),
-                logger ?? Mock.Of<ILogger<PublishingService>>()
+                logger ?? Mock.Of<ILogger<PublishingService>>(),
+                configuration ?? DefaultConfigurationMock().Object
             );
         }
+
+        private static Mock<IConfiguration> DefaultConfigurationMock()
+        {
+            return MockUtils.CreateMockConfiguration(
```
duncan-at-hiveit, line 432: Wondering what this is being used for now?

src/GovUk.Education.ExploreEducationStatistics.Publisher/Startup.cs
```diff
@@ -64,21 +64,34 @@ public override void Configure(IFunctionsHostBuilder builder)
...
+                    return new BlobStorageService(
+                        connectionString: publicConnectionString,
+                        client: new BlobServiceClient(publicConnectionString),
+                        logger: provider.GetRequiredService<ILogger<BlobStorageService>>(),
+                        storageInstanceCreationUtil: new StorageInstanceCreationUtil());
+                })
+                .AddScoped<IPrivateBlobStorageService, BlobStorageService>(provider =>
+                {
+                    var privateConnectionString = GetConfigurationValue(provider, "CoreStorage");
+                    return new BlobStorageService(
```
duncan-at-hiveit, line 80: Again, with a Public and Private concrete class that each extracted their own connection string and created their own BlobServiceClient in their constructor, you wouldn't have to do all of this in-depth wiring here.

src/GovUk.Education.ExploreEducationStatistics.Common/Services/BlobStorageService.cs
duncan-at-hiveit: Although nearer the end of my review, I've done a switcheroo and suggested that maybe having concrete classes would actually be more beneficial and lead to simpler code in `Startup.cs` where you're having to construct these manually each time.

markhiveit: I originally did have classes, but for some reason decided against it. I have now restored the classes and it all seems fine. I ran the UI tests and got no failures.

markhiveit: One thing to note, to achieve this I've had to add an empty constructor to `BlobStorageService` and change `BlobStorageService` variables to `protected` with no `readonly`. But I think this is fine.

src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/Services/PublishingServiceTests.cs
markhiveit: If I switch out `DefaultConfigurationMock` for `Mock.Of<IConfiguration>()` in `BuildPublishingService`, I get several test failures on the `_publicStorageConnectionString = configuration.GetValue<string>("PublicStorage");` line in the `PublishingService` constructor. So I think it is still necessary.

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
```diff
@@ -631,20 +649,13 @@ public void ConfigureServices(IServiceCollection services)
 
             services.AddSingleton<IFileTypeService, FileTypeService>();
             services.AddTransient<IDataArchiveValidationService, DataArchiveValidationService>();
-            services.AddTransient<IBlobCacheService, BlobCacheService>();
+            services.AddTransient<IBlobCacheService, BlobCacheService>(provider =>
```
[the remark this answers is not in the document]
markhiveit: Created EES-4343

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Services/PublicReleaseFileBlobService.cs
markhiveit: Created EES-4344

markhiveit:
Pushed some fixes after rebasing on dev

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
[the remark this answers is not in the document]
duncan-at-hiveit: Noice!

src/GovUk.Education.ExploreEducationStatistics.Common/Services/BlobStorageService.cs
duncan-at-hiveit: Yup sounds good to me!

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Services/PublicReleaseFileBlobService.cs
duncan-at-hiveit: Cool!

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/MockBlobStorageServiceExtensions.cs
duncan-at-hiveit: Sweet!

src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/Services/PublishingServiceTests.cs
duncan-at-hiveit: No worries, cheers for checking it out

src/GovUk.Education.ExploreEducationStatistics.Common/Services/PrivateBlobStorageService.cs
```diff
@@ -0,0 +1,23 @@
+#nullable enable
+using Azure.Storage.Blobs;
+using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces;
+using GovUk.Education.ExploreEducationStatistics.Common.Utils;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Services
+{
+    public class PrivateBlobStorageService : BlobStorageService
```
duncan-at-hiveit, line 10: I would go:

```
public class PrivateBlobStorageService : BlobStorageService, IPrivateBlobStorageService
```
 
here, and not have BlobStorageService implement IPrivateBlobStorageService and IPublicBlobStorageService anymore if I were you?

src/GovUk.Education.ExploreEducationStatistics.Common/Services/PublicBlobStorageService.cs
```diff
@@ -0,0 +1,23 @@
+#nullable enable
+using Azure.Storage.Blobs;
+using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces;
+using GovUk.Education.ExploreEducationStatistics.Common.Utils;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Services
+{
+    public class PublicBlobStorageService : BlobStorageService
```
duncan-at-hiveit, line 10: I would go:

```
public class PublicBlobStorageService : BlobStorageService, IPublicBlobStorageService
```

here, and not have BlobStorageService implement IPrivateBlobStorageService and IPublicBlobStorageService anymore if I were you?

src/GovUk.Education.ExploreEducationStatistics.Common/Services/PrivateBlobStorageService.cs
duncan-at-hiveit: Also, if you wanted to double down, stick all of this in the base constructor of BlobStorageService, and just have this constructor pass in the string "CoreStorage" to the base constructor, and the same for the public version too.

That way, you don't have to make the variables protected anymore.

src/GovUk.Education.ExploreEducationStatistics.Common/Services/BlobStorageService.cs
duncan-at-hiveit: I'd also make this class abstract now so no one can accidentally construct one by accident.

duncan-at-hiveit (approved):

PR #4076, opened 2023-06-08
EES-3940 restrict chart y axis width on mobile

amyb-hiveit:
Fixes the bug rendering horizontal charts on mobile by:
- restricting the y axis width so there's always space for the chart
- positioning tooltips on the left

**Before:**
before

**After:**
now1
now2

### DIFF
--- a/src/explore-education-statistics-common/src/modules/charts/components/HorizontalBarBlock.tsx
+++ b/src/explore-education-statistics-common/src/modules/charts/components/HorizontalBarBlock.tsx
@@ -1,3 +1,4 @@
+import { useMobileMedia } from '@common/hooks/useMedia';
 import ChartContainer from '@common/modules/charts/components/ChartContainer';
 import CustomTooltip from '@common/modules/charts/components/CustomTooltip';
 import useLegend from '@common/modules/charts/components/hooks/useLegend';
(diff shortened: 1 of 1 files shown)


src/explore-education-statistics-common/src/modules/charts/components/HorizontalBarBlock.tsx
```diff
@@ -81,7 +83,13 @@ const HorizontalBarBlock = ({
   const minorDomainTicks = getMinorAxisDomainTicks(chartData, axes.minor);
   const majorDomainTicks = getMajorAxisDomainTicks(chartData, axes.major);
 
-  const yAxisWidth = parseNumber(axes.major.size);
+  // Enforce a max y axis width on mobile as large widths cause
+  // the chart to not be visible.
+  const maxMobileYAxisWidth = 160;
```
ntsim, line 88: Any reason why this particular width has been chosen? Or is it one that just seemed to work?

amyb-hiveit: Just from trying different widths to find one that gave enough space to the chart while trying to not cause too many lines of wrapping on long labels.

ntsim: Cool, guess it will do for now!

PR #4078, opened 2023-06-08
EES-4327 Fix data replacements for data block with map chart with nul…

markhiveit:
…l location in config

This PR fixes a bug where a data replacement will fail if a data block has a map chart attached that has a dataSetConfig that has a dataSet that has a null location.


N-moh (approved):

amyb-hiveit:
Makes sense to me and have tested locally. Will leave to someone backendy to review the code and approve.

benoutram (approved):

PR #4079, opened 2023-06-12
EES-3251 force text wrapping on map key

amyb-hiveit:
Adds a util class for `word-break: break-word` and applies it to the key heading on maps to prevent text overflowing.

### DIFF
--- a/src/explore-education-statistics-common/src/styles/utils/_text.scss
+++ b/src/explore-education-statistics-common/src/styles/utils/_text.scss
@@ -1,3 +1,7 @@
 .dfe-white-space--pre-wrap {
   white-space: pre-wrap;
 }
+
+.dfe-word-break--break-word {
+  word-break: break-word;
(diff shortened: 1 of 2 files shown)


luke-h1 (approved):

src/explore-education-statistics-common/src/styles/utils/_text.scss
```diff
@@ -1,3 +1,7 @@
 .dfe-white-space--pre-wrap {
   white-space: pre-wrap;
 }
+
+.dfe-word-wrap--break-word {
```
ntsim, line 5: Shouldn't this be `dfe-word-break--break-word`?

amyb-hiveit: doh, yes. Sorted now.

ntsim (approved):

PR #4085, opened 2023-06-12
EES-4320 fix permalinks bug and add dedupe code to api

amyb-hiveit:
Does two things:

1. Fixes the broken permalinks reported in EES-4320. This was caused by these permalinks having duplicate locations in `tableHeadersConfig` as well as `fullTable` - most of the old permalinks with duplicate locations don't have this problem so it's not accounted for in the deduplication code and presumably didn't cause an error with the old way of rendering the table.

2. Adds `deduplicatePermalinkLocations` to the `createPermalinkTable` endpoint as I realised this would need to be there for when old permalinks with duplicate locations are migrated.

### DIFF
--- a/src/explore-education-statistics-common/src/modules/table-tool/utils/mapTableToJson.ts
+++ b/src/explore-education-statistics-common/src/modules/table-tool/utils/mapTableToJson.ts
@@ -243,16 +243,21 @@ function mapTableBody(
       }),
     );
 
-    rowsJson.push(...rowsHeaderJson);
+    // TO DO - EES-4259 Can remove this check once the permalinks migration is done.
+    // Check for rowsHeaderJson because of some old permalinks with duplicate
+    // locations in the table headers - EES-4320.
+    if (rowsHeaderJson) {
+      rowsJson.push(...rowsHeaderJson);
 
-    rowsJson.push(
-      ...row.map<TableCellJson>(cell => {
-        return {
-          tag: 'td',
-          text: cell,
-        };
-      }),
-    );
+      rowsJson.push(
+        ...row.map<TableCellJson>(cell => {
+          return {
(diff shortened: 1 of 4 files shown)


src/explore-education-statistics-common/src/modules/table-tool/utils/mapTableToJson.ts
```diff
@@ -243,16 +243,21 @@ function mapTableBody(
...
-      ...row.map<TableCellJson>(cell => {
-        return {
-          tag: 'td',
-          text: cell,
-        };
-      }),
-    );
+    // TO DO - EES-4259 Can remove this check once the permalinks migration is done.
+    // Check for rowsHeaderJson because of some old permalinks with duplicate
+    // locations in the table headers - EES-4320.
```
ntsim, line 248: Not too clear on why we're doing this as it's not obvious why duplicate locations would means that the row headers don't exist. Could you go into a bit more detail?

Also, from having looked at some of the samples, we're getting duplicates because of codes like `E12000001` being used incorrectly across all regions, or weird ones like `:` (meaning 'Not available'). 

We might want to go into more investigation on these ones to figure out if the data is up to scratch. It might be possible we could just disregard these?

Examples:

- ec48be82-8d56-45d9-b82d-da1ce9cbff6d (repeated region code)
- 535c5845-f75a-40b4-ba7f-7cee2794e931 (repeated region code)
- 3a161dba-0116-4bc5-8a3f-72b64ed7f592 (codes like `:`)

amyb-hiveit:
Closing this as we've decided to not fix these permalinks as they have bad data. The change to createPermalinkTable has been done separately in https://github.com/dfe-analytical-services/explore-education-statistics/pull/4090.

PR #4088, opened 2023-06-13
EES-4333 Fix Admin TableBuilderController endpoint returning wrong content type

markhiveit:
This PR fixes an issue where on a Release's Admin Content page, if you previewed the table tool, created a table, and then tried to save that table as a CSV file, the backend would instead return JSON.

As part of this change, I turned the unit test in Admin's `TableBuilderControllerTests` into an integration test. This is the Admin's first integration test so various other code needed to be added such as a new `TestStartup` for Admin.

For this ticket, a hacky solution has been added to `TestStartup` to sidestep authentication/authorisation in Admin integration tests. Long-term, we'll want to emulate non-test authentication/authorisation: a ticket (EES-4338) has been created for that work.

### Other changes

- Changes the route for Admin's TableBuilderController from `api/data/[controller]` to just `api`, and instead explicitly adds `data/tablebuilder/...` to each endpoint's `Http*` attribute. This is to make these endpoints more searchable.
- Remove `stopwatch` logging from the Admin's `TableBuilderController` Query endpoint. I made a judgement call to remove this - happy to readd it if others think it's needed.


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Statistics/TableBuilderControllerTests.cs
```diff
@@ -19,145 +18,97 @@
...
+                $"/api/data/tablebuilder/release/{ReleaseId}",
+                new JsonNetContent(ObservationQueryContext));
 
-            var result = await controller.Query(_releaseId, _query, cancellationToken);
-            VerifyAllMocks(mocks);
+            VerifyAllMocks(tableBuilderService);
 
-            result.AssertOkResult();
+            response.AssertOk(_tableBuilderResults);
         }
```
ntsim, line 113: We should write a test for the CSV variant of this endpoint, otherwise we're not really proving this has been fixed.

See the data API equivalent endpoint(s) for examples.

ntsim (changes requested):

benoutram (approved):

ntsim (approved):

PR #4093, opened 2023-06-14
EES-4315 UI tests to check Ancillary file replacement during Amendment

N-moh:
UI tests to check Ancillary file during amendment in a scenario where we cancel an amendment after replacing the ancillary files.
Screenshot (355)

### DIFF
--- a/tests/robot-tests/tests/admin_and_public_2/bau/publish_amend_and_cancel.robot
+++ b/tests/robot-tests/tests/admin_and_public_2/bau/publish_amend_and_cancel.robot
@@ -235,6 +235,32 @@ Confirm data replacement
     user clicks button    Confirm data replacement
     user waits until h2 is visible    Data replacement complete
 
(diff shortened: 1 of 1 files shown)


tests/robot-tests/tests/admin_and_public_2/bau/publish_amend_and_cancel.robot
```diff
@@ -235,6 +235,32 @@ Confirm data replacement
     user clicks button    Confirm data replacement
     user waits until h2 is visible    Data replacement complete
 
+Edit Ancillary file and replace data
+    [Documentation]    EES-4315
```
markhiveit, line 239: Not saying you need to remove this, but just for future reference I don't think there is a need to do so. We could always work out the ticket it was introduced with from the git history if needed - and additionally I can't imagine a situation where we'll need to know the associated ticket.

markhiveit (approved):

PR #4100, opened 2023-06-19
EES-4339 admin survey banner

amyb-hiveit:
Adds a banner linking to a survey to the admin (yes it's huge because there's a lot of text, but Cam's fine with that). It shows on all pages except pre-release access.

We didn't have anything to manage cookies in the admin so I've added `react-cookie`.

I've added a generic banner component and replaced the front end user testing banner with this as well.

banner2


src/explore-education-statistics-common/src/components/Banner.tsx
```diff
@@ -0,0 +1,47 @@
...
+  onClose: () => void;
+}
+
+export default function Banner({
+  children,
+  showBanner,
+  wide = false,
+  onClose,
+}: Props) {
+  const { isMounted } = useMounted();
```
ntsim, line 20: Wouldn't do an `isMounted` check in here as this is a pretty basic component, so it should delegate it's rendering lifecycle to the parent component.

This is particularly relevant when it comes to SSR as this check is redundant in the admin.

src/explore-education-statistics-common/src/components/Banner.tsx
```diff
@@ -0,0 +1,47 @@
...
+interface Props {
+  children: ReactNode;
+  showBanner: boolean;
+  wide?: boolean;
+  onClose: () => void;
+}
+
+export default function Banner({
+  children,
+  showBanner,
```
ntsim, line 16: What's the benefit of this prop? We can just wrap the banner with a condition to achieve the same thing

src/explore-education-statistics-common/src/components/Banner.tsx
```diff
@@ -0,0 +1,47 @@
...
+import useMounted from '@common/hooks/useMounted';
+import ButtonText from '@common/components/ButtonText';
+import classNames from 'classnames';
+import React, { ReactNode } from 'react';
+
+interface Props {
+  children: ReactNode;
+  showBanner: boolean;
+  wide?: boolean;
+  onClose: () => void;
```
ntsim, line 11: We might want to make this optional as we might need the banner to always be shown in the future?

The relevant UI would also need to not render if this isn't set

src/explore-education-statistics-admin/src/components/Page.tsx
```diff
@@ -40,6 +47,47 @@ const Page = ({
 
       <PageHeader wide={wide} />
 
+      {showBanner && (
```
ntsim, line 50: Should this always be visible? This will also be shown to non-DfE pre-release users, which I suspect isn't intended behaviour.

ntsim (changes requested):

src/explore-education-statistics-common/src/components/Banner.tsx
amyb-hiveit: Yeah good point, I've got rid of it.

src/explore-education-statistics-admin/src/components/Page.tsx
ntsim: Good point, didn't see that!

ntsim (approved):

PR #4108, opened 2023-06-22
EES-4341 screen reader text in table tool forms

amyb-hiveit:
Adds visually hidden text for screen reader users to the table tool location and filters steps:
- select all buttons
- search labels

### DIFF
--- a/src/explore-education-statistics-common/src/components/form/FormCheckboxGroup.tsx
+++ b/src/explore-education-statistics-common/src/components/form/FormCheckboxGroup.tsx
@@ -15,6 +15,7 @@ import React, {
   useMemo,
   useRef,
   Ref,
+  ReactNode,
 } from 'react';
 
 import VisuallyHidden from '../VisuallyHidden';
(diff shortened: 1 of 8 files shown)


src/explore-education-statistics-common/src/components/form/FormCheckboxGroup.tsx
```diff
@@ -40,45 +41,46 @@ export type CheckboxGroupAllChangeEventHandler = (
...
-  hiddenText?: string;
   inputRef?: Ref<HTMLInputElement>;
 }
 
-const getDefaultSelectAllText = (
-  isAllChecked: boolean,
-  opts: CheckboxOption[],
-) => `${isAllChecked ? 'Unselect' : 'Select'} all ${opts.length} options`;
+export const getDefaultSelectAllText = (isAllChecked: boolean, total: number) =>
+  `${isAllChecked ? 'Unselect' : 'Select'} all ${total} options`;
```
luke-h1, line 64: Should we add a check here to not pluralize `options` if the total is only one?

src/explore-education-statistics-common/src/components/form/FormCheckboxSearchSubGroups.tsx
```diff
@@ -187,17 +198,19 @@ const FormCheckboxSearchSubGroups = ({
...
-                selectAllText={(allChecked, opts) =>
-                  `${allChecked ? 'Unselect' : 'Select'} all ${
-                    opts.length
-                  } subgroup options`
-                }
+                selectAllText={(allChecked, opts) => (
+                  <>
+                    {`${allChecked ? 'Unselect' : 'Select'} all ${
+                      opts.length
+                    } subgroup options`}
```
luke-h1, line 205: Similar to my other comment about pluralisation

src/explore-education-statistics-common/src/components/form/FormCheckboxGroup.tsx
amyb-hiveit: I'll check, but I don't think you get the select all button if there's only one option.

amyb-hiveit: I've checked now and the select all button will only show if there's more than one option so this is ok always being plural.

src/explore-education-statistics-common/src/components/form/FormCheckboxSearchSubGroups.tsx
amyb-hiveit: As above, the button only shows if there's more than one option.

src/explore-education-statistics-common/src/components/form/FormCheckboxGroup.tsx
amyb-hiveit: I don't think it's necessary as this is specifically text for selecting all of a list of options. If you were using this where there was only one option you'd have a button that says "Select option" above a single checkbox which would be odd and potentially confusing.

src/explore-education-statistics-common/src/components/form/FormCheckboxSearchSubGroups.tsx
```diff
@@ -32,7 +34,9 @@ export interface FormCheckboxSearchSubGroupsProps
     FormCheckboxGroupProps,
     'onAllChange' | 'selectAll' | 'selectAllText' | 'options'
   > {
+  groupLabel: string;
```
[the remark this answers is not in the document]
amyb-hiveit: Yeah good point, means I don't need all the test changes as well :)

src/explore-education-statistics-common/src/components/form/FormCheckboxGroup.tsx
ntsim: I'd personally fix it because there's no real downside to coding defensively here. It's feasible that this function gets used incorrectly in the future so we could end up having a button that says 'Select/Unselect all 1 options' despite any best intentions here.

Anyway, if you feel particularly strongly about it, I'll approve the PR anyway, but I would suggest just making the change.

ntsim (approved):

src/explore-education-statistics-common/src/components/form/FormCheckboxGroup.tsx
amyb-hiveit: I just feel like defensive coding here would make it more likely that we'd use it wrongly as it would seem like you're meant to show the button when there's only one option. So I'll leave it in for now.

PR #4109, opened 2023-06-23
EES-4049 - add data file info to view datablock page

luke-h1:
This PR:
* amends the data block page to display the data set name of a given data block. 

Other changes:
* Amended UI tests to check for this new field
* Added a new keyword to close the admin survey banner due to it causing intermittent failures in various suites

image

image

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
@@ -117,10 +117,25 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
         {
             return await GetReleaseContentBlock(id)
                 .OnSuccessDo(rcb => _userService.CheckCanViewRelease(rcb.Release))
-                .OnSuccess(CheckIsDataBlock)
-                .OnSuccess(async dataBlock =>
+                .OnSuccessCombineWith(CheckIsDataBlock)
+                .OnSuccess(async tuple =>
                 {
+                    var (releaseContentBlock, dataBlock) = tuple;
+                    var releaseId = releaseContentBlock.ReleaseId;
+
                     var viewModel = _mapper.Map<DataBlockViewModel>(dataBlock);
+
+                    var subjectId = dataBlock.Query.SubjectId;
+
+                    viewModel.DataSetId = subjectId;
+
+                    viewModel.DataSetName = await _context.ReleaseFiles
+                        .Where(rf =>
+                            rf.ReleaseId == releaseId &&
+                            rf.File.SubjectId == subjectId &&
+                            rf.File.Type == FileType.Data)
+                        .Select(rf => rf.Name)
+                        .SingleAsync() ?? "";
 
                     var featuredTable = await _context.FeaturedTables.SingleOrDefaultAsync(
                         ft => ft.DataBlockId == dataBlock.Id);
--- a/tests/robot-tests/tests/admin_and_public/bau/publish_content.robot
+++ b/tests/robot-tests/tests/admin_and_public/bau/publish_content.robot
@@ -59,6 +59,8 @@ Check glossary info icon appears on release preview
     user waits until page contains button    Absence
 
 Click glossary info icon and validate glossary entry
+    user closes admin feedback banner if needed
+
     user clicks button    Absence
     user waits until h2 is visible    Absence
     user checks page contains    When a pupil misses (or is absent from) at least 1 possible school session.
--- a/tests/robot-tests/tests/libs/admin-common.robot
+++ b/tests/robot-tests/tests/libs/admin-common.robot
@@ -855,3 +855,6 @@ user removes key stat
     [Arguments]    ${tile_num}
     user waits until page contains element    xpath://*[@data-testid="keyStat"][${tile_num}]
     user clicks element    xpath://*[@data-testid="keyStat"][${tile_num}]//button[contains(text(), "Remove")]
+
+user closes admin feedback banner if needed
+    user clicks element if exists    //*[@data-testid="admin-survey-banner"]//button[text()="Close"]
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataBlockServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataBlockServiceTests.cs
@@ -35,6 +35,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
         [Fact]
         public async Task Get()
         {
+            var subjectId = Guid.NewGuid();
+            
             var dataBlock = new DataBlock
             {
                 Heading = "Test heading",
@@ -43,6 +45,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                 Order = 5,
                 Query = new ObservationQueryContext
                 {
+                    SubjectId = subjectId,
                     Filters = new List<Guid>
                     {
                         Guid.NewGuid(),
@@ -75,6 +78,22 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                         Width = 500,
                     }
                 },
+                
+            };
+
+            var release = new Release();
+
+            var releaseFile = new ReleaseFile
+            {
+                Name = "test release file",
+                Release = release,
+                File = new File
+                {
+                    Id = Guid.NewGuid(),
+                    SubjectId = subjectId,
+                    Filename = "test filename",
+                    Type = FileType.Data
+                }
             };
 
             var featuredTable = new FeaturedTable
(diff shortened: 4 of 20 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
```diff
@@ -122,9 +122,32 @@ public class DataBlockService : IDataBlockService
                 {
                     var viewModel = _mapper.Map<DataBlockViewModel>(dataBlock);
 
+                    var subjectId = dataBlock.Query.SubjectId;
+
+                    var releaseContentBlock = await _context.ReleaseContentBlocks
+                        .Include(rcb => rcb.Release)
+                        .SingleOrDefaultAsync(rcb => rcb.ContentBlockId == id);
+
+                    var releaseId = releaseContentBlock.ReleaseId;
```
benoutram, line 131: That translates into a huge select statement selecting all the columns of Release joined with all the columns of ReleaseContentBlocks, just to get the release id.

You could rewrite that as

```
var releaseId = await _context.ReleaseContentBlocks
  .Where(rcb => rcb.ContentBlockId == id)
  .Select(rcb => rcb.ReleaseId)
  .SingleAsync();
```

which should result in a simple select statement `SELECT releaseId FROM ReleaseContentBlocks WHERE ContentBlockId = <id>`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
```diff
@@ -122,9 +122,32 @@ public class DataBlockService : IDataBlockService
...
+                            rf.File.SubjectId == subjectId &&
+                            rf.File.Type == FileType.Data)
+                        .Select(rf => rf.File.Id).SingleAsync();
+
+                    viewModel.DataSetName = await _context.ReleaseFiles.Include(rf => rf.File)
+                        .Where(rf =>
+                            rf.ReleaseId == releaseId &&
+                            rf.File.SubjectId == subjectId &&
+                            rf.File.Type == FileType.Data)
+                        .Select(rf => rf.Name).SingleAsync();
```
benoutram, line 145: Even though data files will all have a name, `rf.Name` is optional (because `ReleaseFile` is used for other types of files which don't all have a name), so you need to add a fall back to something like the empty string.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
```diff
@@ -122,9 +122,32 @@ public class DataBlockService : IDataBlockService
                 {
                     var viewModel = _mapper.Map<DataBlockViewModel>(dataBlock);
 
+                    var subjectId = dataBlock.Query.SubjectId;
+
+                    var releaseContentBlock = await _context.ReleaseContentBlocks
+                        .Include(rcb => rcb.Release)
+                        .SingleOrDefaultAsync(rcb => rcb.ContentBlockId == id);
```
benoutram, line 129: All data blocks will have a `ReleaseContentBlock` so this should have been using `Single` rather than `SingleOrDefaultAsync`. Currently you'll have a warning that `releaseContentBlock` can be null due to using `SingleOrDefaultAsync` which your ignoring.

See other comment though as fetching the content block and also including all the fields of the release just to get the id of the release can be simplified.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
```diff
@@ -122,9 +122,32 @@ public class DataBlockService : IDataBlockService
...
+
+                    viewModel.DataSetName = await _context.ReleaseFiles.Include(rf => rf.File)
+                        .Where(rf =>
+                            rf.ReleaseId == releaseId &&
+                            rf.File.SubjectId == subjectId &&
+                            rf.File.Type == FileType.Data)
+                        .Select(rf => rf.Name).SingleAsync();
+
                     var featuredTable = await _context.FeaturedTables.SingleOrDefaultAsync(
                         ft => ft.DataBlockId == dataBlock.Id);
```
benoutram, line 148: You could abstract out the common predicate for getting the `ReleaseFile` as a `Queryable` here which isn't executed yet. You'd end up with this which avoids duplicating the predicate twice:

```
var releaseFile = _context.ReleaseFiles
    .Where(rf =>
        rf.ReleaseId == releaseId &&
        rf.File.SubjectId == subjectId &&
        rf.File.Type == FileType.Data);

viewModel.DataSetId = await releaseFile
    .Select(rf => rf.FileId)
    .SingleAsync();

viewModel.DataSetName = await releaseFile
    .Select(rf => rf.Name)
    .SingleAsync() ?? "";
```

The problem with all this is that including fetching the release id above and getting the data set id and name here, you have 3 queries in total. This optimisation doesn't change that.

If your comfortable with the syntax, you can do everything in one query. This avoids a separate call to get the release id, eliminating what you've got above.

```
var releaseFile = await _context.ReleaseContentBlocks
    .Join(_context.ReleaseFiles,
        rcb => rcb.ReleaseId,
        rf => rf.ReleaseId,
        (rcb, rf) => new { rcb, rf })
    .Where(tuple =>
        tuple.rcb.ContentBlockId == id &&
        tuple.rf.File.SubjectId == subjectId &&
        tuple.rf.File.Type == FileType.Data)
    .Select(t => new
    {
        DataSetId = t.rf.FileId,
        DataSetName = t.rf.Name
    })
    .SingleAsync();

viewModel.DataSetId = releaseFile.DataSetId;
viewModel.DataSetName = releaseFile.DataSetName ?? "";
```

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
```diff
@@ -122,9 +122,34 @@ public class DataBlockService : IDataBlockService
...
 
+                    var subjectId = dataBlock.Query.SubjectId;
+
+                    // TODO luke: consolidate into one query instead of 3 seperate calls
+                    // to the database
+
+                    var releaseId = await _context.ReleaseContentBlocks
+                        .Where(rcb => rcb.ContentBlockId == id)
+                        .Select(rcb => rcb.ReleaseId)
+                        .SingleAsync();
```
markhiveit, line 133: You can get the `releaseId` without a query since we've already called `GetReleaseContentBlock`. 

To get it in this `OnSuccess` block, you can change `OnSuccess(CheckIsDataBlock)` above to `OnSuccessDo(CheckIsDataBlock)`. This means that the `releaseContentBlock` returned from `GetReleaseContentBlock` is passed forward rather than the result of `CheckIsDataBlock`, meaning you can just do `var releaseId = releaseContentBlock.ReleaseId;`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
markhiveit: As we can get the `releaseId` from the result of `GetReleaseContentBlock` - see my other comment - I don't believe this join is necessary.

I think instead we can just do this?

```
                    viewModel.DataSetName = await _context.ReleaseFiles
                        .Where(rf =>
                            rf.ReleaseId == releaseId &&
                            rf.File.SubjectId == subjectId &&
                            rf.File.Type == FileType.Data)
                        .Select(rf => rf.Name)
                        .SingleAsync();
```

markhiveit: As we can get the releaseId from the result of GetReleaseContentBlock - see my other comment - I don't believe this join is necessary.

I think instead we can just do this?

```
                    viewModel.DataSetName = await _context.ReleaseFiles
                        .Where(rf =>
                            rf.ReleaseId == releaseId &&
                            rf.File.SubjectId == subjectId &&
                            rf.File.Type == FileType.Data)
                        .Select(rf => rf.Name)
                        .SingleAsync();
```

luke-h1: Sorted now. As discussed with Ben, it might be worth optimizing the `GetReleaseContentBlock` method to not return the entire release

tests/robot-tests/tests/libs/admin-common.robot
```diff
@@ -855,3 +855,6 @@ user removes key stat
     [Arguments]    ${tile_num}
     user waits until page contains element    xpath://*[@data-testid="keyStat"][${tile_num}]
     user clicks element    xpath://*[@data-testid="keyStat"][${tile_num}]//button[contains(text(), "Remove")]
+
+user closes user testing modal
+    user clicks element    //*[@data-testid="admin-survey-banner"]//button[text()="Close"]
```
markhiveit, line 860: You could change this to `user closes user testing modal if exists` and use `user clicks element if exists` - then tests won't break when this modal gets removed.

tests/robot-tests/tests/admin_and_public/bau/publish_content.robot
```diff
@@ -59,6 +59,8 @@ Check glossary info icon appears on release preview
     user waits until page contains button    Absence
 
 Click glossary info icon and validate glossary entry
+    user closes user testing modal
```
markhiveit, line 62: I assume this modal is temporary. Is there a ticket for removing the user testing modal? If so, I'd leave a comment here and on the ticket that this line needs to be removed from the tests.

markhiveit (approved):

tests/robot-tests/tests/admin_and_public/bau/publish_content.robot
luke-h1: It's temporary but I don't think there is a ticket yet

PR #4110, opened 2023-06-26
EES-4358 Fix table html for filter name related bugs

amyb-hiveit:
The permalinks snapshot testing found two discrepancies in the table html which exposed bugs related to having filters and filter groups with the same name.

After this fix goes in there will still be a difference in the table html as the rowspans were wrong in the before version as well.

### 1. The only filters and filter groups in the column headers have the same names
Permalink id: 4a378c3c-6443-4af8-393f-08db371965b6

Here the `All state-funded mainstream institutions` filter is in filter group `All state-funded mainstream institutions` and `All institutions that are not academies` is in the filter group `All institutions that are not academies`, this caused the scope to be incorrectly set to `colgroup` in the after version (using table Json). In both the before and after versions the `rowspan` was incorrectly set to 2. 

Permalink before and after diff:
snapshotdiff1

With this fix the correct `scope` and `rowspan` is applied:

fixed1

### 2. Column headers have nested filters in different groups with the same names and some of the columns have no data
Permalink id: 1cca2255-b239-465c-832a-07d17e73606a

Here the Male and Female filters are nested under another group with Male and Female filters, and one of the filters for each has no data so the column is excluded. This also caused the scope to be incorrectly set to `colgroup` in the after version (using tableJson). In both the before and after versions the `rowspan` was incorrectly set to 2. 

Permalink before and after diff:
snapshotdiff2

With this fix the correct `scope` and `rowSpan` is applied:

fixed2

### Additional fixes

In the course of this I found that there were multiple issues related to having headers where a whole level of headers contains only merged cells (merged because their parent filter groups have the same label), this results in:
- wrongly setting the scope to colgroup/rowgroup when it's the first or last level of headers
- adding an empty row to column headers when it's a middle level of headers
- having incorrect colspan/rowspan on the merged cells at all levels

I've tried to cover all these cases in unit tests, which has generated a lot of test data. I've split this into 3 files just to make it a bit more manageable.

I've also re-written the existing test for multiple header groups with the same label to tidy up the test data and make it clearer what's being tested. Plus have added a rows version of the test.

### DIFF
--- a/src/explore-education-statistics-common/src/modules/table-tool/utils/__tests__/mapTableToJson.test.ts
+++ b/src/explore-education-statistics-common/src/modules/table-tool/utils/__tests__/mapTableToJson.test.ts
@@ -8,13 +8,46 @@ import {
   testTableWithTwoLevelsOfRowAndOneLevelOfColHeadersConfig,
   testTableWithOneLevelOfRowsAndTwoLevelsOfColHeadersConfig,
   testTableWithMissingTimePeriod,
-  testTableWithThreeLevelsOfColHeadersWithMultipleGroupsWithSameLabelsConfig,
-  testTableWithThreeLevelsOfColHeadersWithMultipleGroupsWithSameLabels,
 } from '@common/modules/table-tool/utils/__data__/testTableData';
+import {
+  testTableWithOnlyMergedCellsInColumnHeadersConfig,
+  testTableWithOnlyMergedCellsInHeaders,
+  testTableWithMergedAndUnMergedCellsInHeaders,
+  testTableWithMergedAndUnMergedCellsInColumnHeadersConfig,
+  testTableWithOnlyMergedCellsInRowHeadersConfig,
+  testTableWithMergedAndUnmergedCellsInRowHeadersConfig,
+  testTableWithOnlyMergedCellsInFirstLevelOfHeaders,
+  testTableWithOnlyMergedCellsInMiddleLevelOfRowsConfig,
+  testTableWithOnlyMergedCellsInFirstLevelOfColumnHeadersConfig,
+  testTableWithMergedAndUnmergedCellsInFirstLevelOfColumnHeadersConfig,
+  testTableWithMergedAndUnmergedCellsInLastLevelOfColumnHeadersConfig,
+  testTableWithOnlyMergedCellsInLastLevelOfColumnHeadersConfig,
+  testTableWithOnlyMergedCellsInFirstLevelOfRowHeadersConfig,
+  testTableWithMergedAndUnmergedCellsInFirstLevelOfHeaders,
+  testTableWithMergedAndUnmergedCellsInFirstLevelOfRowHeadersConfig,
+  testTableWithOnlyMergedCellsInMiddleLevelOfColumnHeadersConfig,
+  testTableWithOnlyMergedCellsInMiddleLevelOfHeaders,
+  testTableWithMergedAndUnmergedCellsInMiddleLevelOfColumnHeadersConfig,
+  testTableWithMergedAndUnmergedCellsInMiddleLevelOfHeaders,
+  testTableWithMergedAndUnmergedCellsInMiddleLevelOfRowHeadersConfig,
+  testTableWithOnlyMergedCellsInLastLevelOfHeaders,
+  testTableWithOnlyMergedCellsInLastLevelOfRowHeadersConfig,
+  testTableWithMergedAndUnmergedCellsInLastLevelOfHeaders,
+  testTableWithMergedAndUnmergedCellsInLastLevelOfRowHeadersConfig,
+} from '@common/modules/table-tool/utils/__data__/testTableDataWithMergedCells';
 import mapTableToJson, {
   TableCellJson,
 } from '@common/modules/table-tool/utils/mapTableToJson';
 import { ReleaseTableDataQuery } from '@common/services/tableBuilderService';
+import {
+  testTableWithDuplicateFilterLabelsInColumnHeadersConfig,
+  testTableWithDuplicateFilterLabels,
+  testTableWithDuplicateFilterLabelsAndMissingData,
+  testTableWithDuplicateFilterLabelsInRowHeadersConfig,
+  testTableWithMultipleGroupsWithSameLabelsInColumnHeadersConfig,
+  testTableWithMultipleGroupsWithSameLabels,
+  testTableWithMultipleGroupsWithSameLabelsInRowHeadersConfig,
+} from '@common/modules/table-tool/utils/__data__/testTableDataWithDuplicateLabels';
 
 describe('mapTableToJson', () => {
   test('returns the correct JSON for a table with one level of row and column headers', () => {
@@ -120,7 +153,7 @@ describe('mapTableToJson', () => {
     ]);
   });
 
-  test('returns the correct JSON for  a table with two levels of col headers and one level of row headers', () => {
+  test('returns the correct JSON for a table with two levels of col headers and one level of row headers', () => {
     const result = mapTableToJson({
       tableHeadersConfig: testTableWithOneLevelOfRowsAndTwoLevelsOfColHeadersConfig,
       subjectMeta: testTableWithOneLevelOfRowAndColHeaders.subjectMeta,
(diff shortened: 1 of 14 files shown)


ntsim (changes requested):

src/explore-education-statistics-common/src/modules/table-tool/utils/mapTableToJson.ts
```diff
@@ -205,11 +205,21 @@ function mapTableHead(
       });
     }
 
+    // If it's the last row of column headers the rowspan should always be 1
+    // and the scope should always be `col`.
+    // There are edge cases that can cause the rowSpan and scope to be wrong if
+    // don't have this check (EES-4358)
```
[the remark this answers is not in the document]
amyb-hiveit: I've removed this comment in the latest version.

src/explore-education-statistics-common/src/modules/table-tool/utils/mapTableToJson.ts
```diff
@@ -205,11 +205,21 @@ function mapTableHead(
...
+    // - the only filters and filter groups in the column headers have the same name
+    // - column headers have nested filters in different groups with the same names
+    // and some of the columns have no data
+    const isLastHeaderRow = rowIndex === expandedColumnHeaders.length - 1;
```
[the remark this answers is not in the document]
amyb-hiveit: Following your comments and our discussions:
- it uses maxCrossSpan to determine the scope for column and row headers
- for column headers it loops over each row as they're added to adjust the rowspan if necessary
- for row headers it checks whether all headers at that level will be merged when it adds the colspan 
- createRowHeaders / createColumnHeaders now return TableCellJson removing the need for additional loops to convert them.
- the tests cover all the variants for row and column headers.

Also, I've updated the PR description to outline the additional issues that have been found and fixed.

src/explore-education-statistics-common/src/modules/table-tool/utils/createExpandedRowHeaders.ts
```diff
@@ -42,21 +46,34 @@ export default function createExpandedRowHeaders(
...
       ) {
-        const isGroup = current.hasChildren();
+        // Check to see if all headers at the current level have only
+        // a single matching child so will be merged.
+        const headersAtCurrentLevel = current.parent
+          ? current.parent.children
+          : rowHeaders;
+        const allHeadersAtCurrentLevelWillBeMerged = headersAtCurrentLevel.every(
+          rowHeader => rowHeader.hasSingleMatchingChild(),
+        );
```
ntsim, line 56: Similarly to `createExpandedColumnHeaders`, this is a little inefficient as we're having to re-calculate this every single time we push a header to the row, so we're potentially repeating ourselves.

Seems like it would be fairly easy to fix this with something like:

```ts
export default function createExpandedRowHeaders(
  rowHeaders: Header[],
): TableCellJson[][] {
  // ...

  // Keep track of levels that can be collapsed (i.e. when each
  // header in the level only has a single matching child)
  let collapsibleLevels: boolean[] = [
    rowHeaders.every(rowHeader => rowHeader.hasSingleMatchingChild()),
  ];

  return rowHeaders.reduce<TableCellJson[][]>((acc, header) => {
    // ...

    // 👇️ Keep track of the depths that have been checked for 
    const depthsChecked = new Set<number>();

    while (stack.length > 0) {
      // ...

      if (
        !matchesPreviousHeader ||
        (matchesPreviousHeader && current.hasSiblings())
      ) {
        const { depth, crossSpan } = current;

        if (collapsibleLevels[depth] === undefined && current.parent) {
          collapsibleLevels[depth] = current.parent.children.every(rowHeader =>
            rowHeader.hasSingleMatchingChild(),
          );
        }

        row.push({
          text: current.text,
          rowSpan: current.span,
          scope: maxDepth > depth + crossSpan ? 'rowgroup' : 'row',
          colSpan: collapsibleLevels[depth] ? 1 : crossSpan,
          tag: 'th',
        });
      }

      // ...
    }

    // 👇️ Need to remember to reset this before moving to next top-level header
    collapsibleLevels = [collapsibleLevels[0]];

    return acc;
  }, []);
}
```

As a side benefit, this actually solves another test case I found by tweaking `createExpandedRowHeaders > returns correct headers with deeply nested rows and multiple identical headers` a little with the following headers:

```ts
const rowHeaders: Header[] = [
  new Header('A', 'A').addChild(
    new Header('A', 'A').addChild(
      new Header('A', 'A').addChild(
        new Header('A', 'A').addChild(new Header('A', 'A')),
      ),
    ),
  ),
  new Header('B', 'B').addChild(
    new Header('B', 'B')
      .addChild(
        new Header('B', 'B').addChild(
          new Header('B', 'B').addChild(new Header('B', 'B')),
        ),
      )
      .addChild(
        new Header('C', 'C').addChild(
          new Header('D', 'D').addChild(new Header('D1', 'D1')),
        ),
      ),
  ),
  new Header('E', 'E').addChild(
    new Header('F', 'F')
      .addChild(
        new Header('F', 'F')
          .addChild(new Header('F', 'F').addChild(new Header('F', 'F')))
          .addChild(new Header('G', 'G').addChild(new Header('G1', 'G1'))),
      )
      .addChild(
        new Header('H', 'H')
          .addChild(new Header('I', 'I').addChild(new Header('I', 'I')))
          .addChild(new Header('J', 'J').addChild(new Header('J', 'J'))),
      ),
  ),
];
```

With the current implementation:

image

With the new level tracking:
image

Probably worth adding this test case in (or changing the existing one).

src/explore-education-statistics-common/src/modules/table-tool/utils/__tests__/mapTableToJson.test.ts
```diff
@@ -515,134 +548,1845 @@ describe('mapTableToJson', () => {
...
 
-  test('correctly adds groups across three levels of column headers with same labels', () => {
-    const result = mapTableToJson({
-      tableHeadersConfig: testTableWithThreeLevelsOfColHeadersWithMultipleGroupsWithSameLabelsConfig,
-      subjectMeta:
-        testTableWithThreeLevelsOfColHeadersWithMultipleGroupsWithSameLabels.subjectMeta,
-      results:
-        testTableWithThreeLevelsOfColHeadersWithMultipleGroupsWithSameLabels.results,
-    }).tableJson;
+  describe('Handles multiple filter groups with the same labels', () => {
```
ntsim, line 551: Probably worth keeping these in now that we have them, but I'd suggest we try and add some extra cases to `createExpandedColumnHeaders` and `createExpandedRowHeaders` test suites to cover these cases as well.

It's also much easier to setup test data for these suites, so it would be good if we could test some more complex variations as well (if possible).

ntsim (changes requested):

src/explore-education-statistics-common/src/modules/table-tool/utils/__tests__/mapTableToJson.test.ts
amyb-hiveit: I've added extra test cases for these to `createExpandedColumnHeaders` and `createExpandedRowHeaders`. I actually find it a bit easier testing the tables in `mapTableToJson` because it's simple to try out the test data in the actual app and see what the tables look like, which is harder to do with just the row/col headers. We could definitely do with a better way to generate test data though as it's a faff either way.

I'm struggling to come up with other valid more complex iterations to test, though I'm sure there must be some! If you have any specific ones you think should be tested let me know and I'll add them.

amyb-hiveit:
Thanks for this @ntsim, I've made the changes you suggested and it's all working well.

src/explore-education-statistics-common/src/modules/table-tool/utils/__tests__/mapTableToJson.test.ts
ntsim: There's probably lots to explore around variations of `returns correct headers with deeply nested rows and multiple identical headers` from `createExpandedColumnHeaders`.

At they very least I'd add this test case to `createExpandedRowHeaders` as well.

We could also write up some random-ish permutations of this test case (for both column and row headers) and label them like `returns correct headers for complex case X` (where X is just the test case number e.g. 1, 2, 3, etc).

It'd be nice if there was a way to test permutation in a more complete way. We could probably  write a utility function to generate complete sets of every header tree permutation (up to a limit), then we'd just need to run them against some snapshots.

amyb-hiveit: I've added tests for 'multiple filter groups with the same labels' for `createExpandedColumnHeaders` and `createExpandedRowHeaders`  now (totally thought I had already, but it was ones for the rows of all merged cells that I added).

I'm not sure there's a lot of value to manually adding more permutations to the tests, there are already a lot of cases covered between the three test files, especially around duplicate groups and merged cells which have been the problem areas.

ntsim (approved):

PR #4112, opened 2023-06-27
EES-4363 Add `GroupCsvColumn` to `Filter` for csv generation

benoutram:
This PR adds a new field `GroupCsvColumn` to `Filter` and makes changes to save the value when importing a data set. The values are taken from a filters configuration in the metadata file.

This new field is being added so that when we generate downloadable table csv files we can use the appropriate csv column header for each filter group column where filters are configured to be grouped.

Currently we don't store this value when data sets are imported so we have no way of creating this column unless we were to invent a new name (which would be different to the underlying data set) or read the metadata file from blob storage each time a csv file is created.

An additional ticket is planned to migrate the values from all dataset csv files from blob storage to the database. See [EES-4364](https://dfedigital.atlassian.net/browse/EES-4364).


src/GovUk.Education.ExploreEducationStatistics.Data.Model/Filter.cs
```diff
@@ -21,12 +22,17 @@ public Filter()
         {
         }
 
-        public Filter(string? hint, string label, string name, Guid subjectId)
+        public Filter(string? hint,
+            string label,
+            string name,
+            string groupCsvColumn,
```
luke-h1, line 28: Just wondering if the type of `groupCsvColumn` should be a nullable string to match the model?

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Filter.cs
```diff
@@ -10,6 +10,7 @@ public class Filter
         public string? Hint { get; set; } 
         public string Label { get; set; } = string.Empty;
         public string Name { get; set; } = string.Empty;
+        public string? GroupCsvColumn { get; set; }
```
luke-h1, line 13: Just wondering if this can be made non-nullable after [EES-4364](https://dfedigital.atlassian.net/browse/EES-4364) ?

benoutram: It will stay nullable because not all Filters have a grouping column.
Many filters don't have a grouping column, so all their filter items get grouped into in a default group labelled 'Default'.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Filter.cs
benoutram: Good point, I think I missed this because `MetaRow` didn't have the nullable annotation context enabled and doing so just now then highlighted this problem and some other related problems. Fixed them in dfe1afc.

luke-h1 (approved):

PR #4116, opened 2023-06-28
EES-4362 add reference lines between two points

amyb-hiveit:
Adds the ability to create a reference line between two data points, for example to explain why there's a gap in the chart:

chart

When adding a reference line on the major axis:
- you can now select 'Between data points' in the Position dropdown
- this reveals dropdowns for start and end points for the line
- you can adjust the other axis position, label and style as normal.

linelabels1

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartAxisConfiguration.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartAxisConfiguration.cs
@@ -73,6 +73,10 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Model.Chart
         public AxisReferenceLineStyle? Style;
 
         public int? OtherAxisPosition;
+
+        public string? OtherAxisStart;
+
+        public string? OtherAxisEnd;
     }
 
     [SuppressMessage("ReSharper", "InconsistentNaming")]
--- a/src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartReferenceLinesConfiguration.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartReferenceLinesConfiguration.tsx
@@ -1,11 +1,14 @@
 import styles from '@admin/pages/release/datablocks/components/chart/ChartReferenceLinesConfiguration.module.scss';
 import Button from '@common/components/Button';
+import { FormFieldset } from '@common/components/form';
 import FormFieldNumberInput from '@common/components/form/FormFieldNumberInput';
 import FormFieldSelect from '@common/components/form/FormFieldSelect';
 import FormFieldTextInput from '@common/components/form/FormFieldTextInput';
 import FormSelect, { SelectOption } from '@common/components/form/FormSelect';
 import Tooltip from '@common/components/Tooltip';
+import VisuallyHidden from '@common/components/VisuallyHidden';
 import {
+  AxisType,
   ChartDefinitionAxis,
   ReferenceLine,
   ReferenceLineStyle,
@@ -15,13 +18,24 @@ import { MinorAxisDomainValues } from '@common/modules/charts/util/domainTicks';
 import { LocationFilter } from '@common/modules/table-tool/types/filters';
 import Yup from '@common/validation/yup';
 import { Formik } from 'formik';
+import produce from 'immer';
 import upperFirst from 'lodash/upperFirst';
 import React, { useMemo } from 'react';
 
+const otherAxisPositionTypes = {
+  default: 'default',
+  betweenDataPoints: 'between-data-points',
+  custom: 'custom',
+} as const;
+type OtherAxisPositionType = keyof typeof otherAxisPositionTypes;
+
 interface AddFormValues {
+  otherAxisEnd?: string;
   label: string;
   otherAxisPosition?: number;
+  otherAxisPositionType?: typeof otherAxisPositionTypes[OtherAxisPositionType];
   position: string | number;
+  otherAxisStart?: string;
   style: ReferenceLineStyle;
 }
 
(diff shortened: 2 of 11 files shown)


src/GovUk.Education.ExploreEducationStatistics.Common/Model/Chart/ChartAxisConfiguration.cs
```diff
@@ -73,6 +73,9 @@ public class AxisReferenceLine
         public AxisReferenceLineStyle? Style;
 
         public int? OtherAxisPosition;
+
+        public string? StartPosition;
```
luke-h1, line 77: These two new fields should follow the same code style as other fields in the model (1 blank line between them)

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartReferenceLinesConfiguration.tsx
```diff
@@ -187,16 +236,59 @@ export default function ChartReferenceLinesConfiguration({
...
-                      hideLabel
-                      placeholder="Choose position"
-                      order={FormSelect.unordered}
-                      options={filteredOptions}
-                    />
+                    <>
+                      <FormFieldSelect
+                        name="position"
+                        id="referenceLines-position"
+                        label="Position"
```
ntsim, line 243: As mentioned in other discussions, I'm not keen on implementing this so that the reference line renders on the opposite axis to where it's actually been added. It's counter-intuitive and is really obvious if the lines are styled (they clearly run at a tangent to the axis the user is working on), and also means that we have add flags all over the code to toggle behaviour (a code smell).

Would suggest that we rejig things so that these bounded reference lines are implemented on the correct axis.

For example, on a line chart, we'd have something like this on the Y axis (minor) tab:

image

The 'Custom' option would work how it does currently, being a percentage along the other axis.

I think ideally we'd also allow none percentage positions to be set as well to really dial in the UX, but we can leave that for another day.

This change would need some modifications to the data structure i.e. `otherAxisStart` and `otherAxisEnd` instead of `startPosition` and `endPosition`.

src/explore-education-statistics-common/src/modules/charts/components/utils/getReferenceLineLabelPosition.ts
```diff
@@ -9,29 +9,37 @@ interface LabelPosition {
 export default function getReferenceLineLabelPosition({
   axis,
   axisType,
+  lineIsBetweenTwoPoints = false,
```
ntsim, line 12: Flags toggling behaviour like this is a bit of a code smell as it adds additional complexity.

I'm not sure we'd need most of the changes in this file if we implement the bounded references lines on the correct axis.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartReferenceLinesConfiguration.tsx
amyb-hiveit: It's moved to the minor axis now.

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartReferenceLinesConfiguration.tsx
```diff
@@ -44,53 +58,125 @@ export default function ChartReferenceLinesConfiguration({
...
+      return schema.shape({
+        otherAxisEnd: Yup.string()
+          .when('otherAxisPositionType', {
+            is: otherAxisPositionTypes.betweenDataPoints,
+            then: Yup.string().required('Enter end point'),
+            otherwise: Yup.string(),
+          })
+          .test(
+            'otherAxisEnd',
+            'End point cannot match start point',
```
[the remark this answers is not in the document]
amyb-hiveit: Yeah I thought about that, but you can group charts by other things so it'd be a bit weird to have an error saying something  like "Maths can't be after English".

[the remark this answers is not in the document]
ntsim: Yeh good point, fine to leave it then!

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartReferenceLinesConfiguration.tsx
```diff
@@ -133,53 +230,42 @@ export default function ChartReferenceLinesConfiguration({
...
+                    delete draft.otherAxisPosition;
+                  });
+                } else {
+                  updatedValues = produce(updatedValues, draft => {
+                    delete draft.otherAxisEnd;
+                    delete draft.otherAxisPosition;
+                    delete draft.otherAxisStart;
+                  });
+                }
+              }
```
ntsim, line 267: We don't need to run `produce` twice, and it'd also be cleaner to do this as a switch:

```ts
const updatedValues = produce(values, draft => {
  switch (draft.otherAxisPositionType) {
    case otherAxisPositionTypes.custom:
      delete draft.otherAxisStart;
      delete draft.otherAxisEnd;
      break;

    case otherAxisPositionTypes.betweenDataPoints:
      delete draft.otherAxisPosition;
      break;

    case otherAxisPositionTypes.default:
      delete draft.otherAxisEnd;
      delete draft.otherAxisPosition;
      delete draft.otherAxisStart;
      break;
  }

  delete draft.otherAxisPositionType;
});

```

It doesn't seem to me like we need the wrapping condition on the 'minor' type either. If the `otherAxisPosition` type field isn't present, then the above switch statement will just ignore it anyway.

ntsim (approved):

PR #4122, opened 2023-06-29
EES-4393 Fix ancillary file replacement issue

markhiveit:
This PR fixes an issue we had with ancillary file replacements. When replacing an ancillary file, the frontend would call two separate endpoints: one to update the `ReleaseFile`s name and summary, and the second to replace the file itself.

The bug we were seeing meant that sometimes, the name and summary wouldn't get updated, even though the file itself would be replaced. We believe this is due to a race condition.

To remove this race condition, we've now made it so only one request is required.

There is already a UI test that replaces an ancillary file, so no additional UI test is required.


src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/ReleaseAncillaryFileReplacementRequest.cs
```diff
@@ -0,0 +1,17 @@
+#nullable enable
+using System.ComponentModel.DataAnnotations;
+using Microsoft.AspNetCore.Http;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Requests;
+
+public class ReleaseAncillaryFileReplacementRequest
```
ntsim, line 7: - Would call this `ReleaseAncillaryFileUpdateRequest`
- This can be a record

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleaseFileController.cs
```diff
@@ -152,10 +153,10 @@ public async Task<ActionResult<IEnumerable<FileInfo>>> GetAncillaryFiles(Guid re
         public async Task<ActionResult<FileInfo>> ReplaceAncillary(
```
ntsim, line 153: Rather than make this a 'replacement', would suggest we change this to be an 'update' across the controller and service

src/explore-education-statistics-admin/src/pages/release/data/ReleaseAncillaryFilePage.tsx
```diff
@@ -42,15 +42,18 @@ export default function ReleaseAncillaryFilePage({
...
-          : undefined,
-      ]);
+          file: newFile,
+        });
+      } else {
+        await releaseAncillaryFileService.updateFile(releaseId, fileId, {
+          title,
+          summary,
+        });
+      }
```
ntsim, line 56: If we fix the semantics from 'replace' to 'update', we can just call a single endpoint for updating the ancillary file.

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleaseFileController.cs
markhiveit: I did this, but didn't like the result. I've pushed what I did to the `EES-4393-2` branch.

I think because the existing update is also being called to update a release data file's title, so to have a single `ReleaseFileService#Update`, meant extra complexity, and the code was quite hard to understand, given that it's being used in different ways across the service. I believe we've had dangerous bugs in the past where someone changes a method not realising it's being called somewhere else, and it breaks a part of the service we wouldn't expect.

So I think I'll start again, and have `ReleaseFileService` have separate `UpdateDataFile` and `UpdateAncillaryFile` methods, and validate that those methods are only being used with files of those types.

markhiveit: I talked this through with Ben, and ended up restoring the status quo and going for a dead simple fix.

The work I did is still on the `EES-4393-2` branch - happy to move ahead with that or something else if you'd prefer.

markhiveit: Discussed this with Nick on a call. I'll restore my initial changes when creating this PR, and then make the following changes:
- Make a single endpoint to update ancillary files that can both update details and replace an ancillary file.
- Rename the replace ancillary methods to update ancillary.
- Don't create a new release file - still create a new file, but just update the existing release file.
- Change the existing update endpoint to only update data file details, as it will only be used to update data files name/summary after these changes.

src/explore-education-statistics-admin/src/services/releaseAncillaryFileService.ts
```diff
@@ -74,13 +80,17 @@ const releaseAncillaryFileService = {
 
     return mapFile(file);
   },
-  async replaceFile(
+  async update(
```
ntsim, line 83: This should be `updateFile` and should replace the existing `updateFile` method

src/explore-education-statistics-admin/src/services/releaseAncillaryFileService.ts
```diff
@@ -28,6 +28,12 @@ export interface UploadAncillaryFileRequest {
   file: File;
 }
 
+export interface ReleaseAncillaryFileUpdateRequest {
+  title: string;
+  summary: string;
+  file?: File | null;
+}
```
ntsim, line 35: We don't need this. We can just update `AncillaryFileUpdateRequest` to have the `file` property.

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/ReleaseFileRequests.cs
```diff
@@ -0,0 +1,35 @@
...
+    public string? Summary { get; set; }
+}
+
+public record ReleaseAncillaryFileUploadRequest
+{
+    [Required]
+    public string Title { get; set; } = null!;
+
+    [Required]
+    public string Summary { get; set; } = null!;
```
ntsim, line 20: Would suggest initialising these as empty strings to be consistent with other places.

Same below for the update request.

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseFileServicePermissionTests.cs
```diff
@@ -228,8 +231,10 @@ await PolicyCheckBuilder<SecurityPolicies>()
             IDataGuidanceFileWriter? dataGuidanceFileWriter = null,
             IUserService? userService = null)
         {
+            contentDbContext ??= Mock.Of<ContentDbContext>();
```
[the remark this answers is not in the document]
markhiveit: My thought is that when we call `SetupReleaseFileService` without a `contentDbContext` argument, that makes `contentDbContext` null. So when we call the `ReleaseFileService` constructor, the `FileRepository` and `ReleaseFileRepository` constructors are being passed null. This creates a warning.

I'll restore it to how it was. This doesn't seem to break anything - just seems wrong to me.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleaseFileController.cs
```diff
@@ -116,13 +116,13 @@ public async Task<ActionResult> Stream(Guid releaseId, Guid fileId)
         }
 
         [HttpPatch("release/{releaseId:guid}/file/{fileId:guid}")]
```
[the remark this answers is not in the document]
markhiveit: Other endpoints in `releaseDataFileService.ts` use `/data/` rather than `/data-file/`, so I've gone with `/data/`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseFileService.cs
```diff
@@ -424,6 +463,8 @@ await fileIds.Select(fileId => CheckFileExists(releaseId, fileId, DeletableFileT
 
         private async Task HydrateReleaseFile(ReleaseFile releaseFile)
         {
+            _contentDbContext.Update(releaseFile);
```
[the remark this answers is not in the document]
markhiveit: I was getting an error that `releaseFile` wasn't tracked by EF and therefore it couldn't include `releaseFile.File`. So I added the `Update` to ensure `releaseFile` is being tracked before those `contentDbContext.Entry(...` calls.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseFileService.cs
```diff
@@ -342,50 +356,75 @@ await fileIds.Select(fileId => CheckFileExists(releaseId, fileId, DeletableFileT
...
+                            contentType: request.File.ContentType,
+                            type: Ancillary,
+                            createdById: _userService.GetUserId());
 
-                    if (!await _releaseFileRepository.FileIsLinkedToOtherReleases(releaseId, oldReleaseFile.FileId))
-                    {
-                        await _privateBlobStorageService.DeleteBlob(
-                            PrivateReleaseFiles,
-                            oldReleaseFile.File.Path());
+                        var updatedReleaseFile = await _releaseFileRepository.Update(
```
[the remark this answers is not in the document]
markhiveit: I've left `UploadChart` alone. I've changed the other two places so they don't use the repository update method anymore.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseFileService.cs
```diff
@@ -22,6 +23,7 @@
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Query.Internal;
```
ntsim, line 26: Doesn't seem like we need this import

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseFileService.cs
```diff
@@ -6,6 +6,7 @@
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using GovUk.Education.ExploreEducationStatistics.Admin.Requests;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.Security;
 using GovUk.Education.ExploreEducationStatistics.Admin.ViewModels;
```
ntsim, line 12: Doesn't seem like we need this import

ntsim (approved):

PR #4123, opened 2023-06-30
EES-4020 - implement automated dependency upgrades

luke-h1:
This PR:
* implements automated dependency upgrade management via [renovate](https://renovatebot.com/). 

For now, we've decided to disable renovate running on the frontend packages until the majority of the frontend dependencies have been upgraded.

At the moment the latest version of `renovate` isn't compatible with the version of node that the project is on so I've decided to not add a pre-commit hook to validate the `renovate.json` file with `renovate-config-validator`. If the config is invalid, an issue will be created via renovate to let us know that something is wrong with the renovate config

⚠️IMPORTANT NOTE⚠️
* It's very important to note that if you push to a given branch that renovate has created, renovate will force push your changes away. This means that if you have to fix up a renovate branch after a dependency has caused breaking changes you will need to first create a branch, point the renovate PR to your desired branch, merge the renovate PR and then start work on your branch (so that renovate doesn't force push your changes away  

Relevant changes:
* Splits PRs dependent on package manager and semver versions of dependencies. Minor and major dependencies will be split up into different PRs to make it easier to manage varying levels of upgrades. 
* Adds a schedule of 1 month for semver minor updates and every 3 months for semver major upgrades (both FE & BE)

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,63 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "dependencyDashboard": true,
+  "extends": ["config:base"],
+  "timezone": "Europe/London",
+  "azure-pipelines": {
+    "enabled": true
+  },
+  "vulnerabilityAlerts": {
+    "groupName": "renovate",
+    "dependencyDashboardApproval": false,
+    "minimumReleaseAge": "0d",
+    "rangeStrategy": "update-lockfile",
+    "commitMessageSuffix": "[SECURITY]",
+    "branchTopic": "{{{datasource}}}-{{{depName}}}-vulnerability",
+    "prCreation": "immediate"
+  },
+  "packageRules": [
+    {
+      "matchDatasources": ["npm"],
+      "groupName": "Minor frontend dependencies",
+      "matchUpdateTypes": ["minor", "patch"],
+      "schedule": ["every month"],
+      "enabled": false
+    },
+    {
+      "matchDatasources": ["npm"],
+      "groupName": "Major frontend dependencies",
+      "matchUpdateTypes": ["major"],
+      "schedule": ["every 3 months"],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -0,0 +1,77 @@
...
+    "commitMessageSuffix": "[SECURITY]",
+    "branchTopic": "{{{datasource}}}-{{{depName}}}-vulnerability",
+    "prCreation": "immediate"
+  },
+  "packageRules": [
+    {
+      "matchDatasources": ["npm"],
+      "groupName": "Minor frontend dependencies",
+      "matchUpdateTypes": ["minor", "patch"],
+      "schedule": ["before 8am every weekday", "on the first day of the month"],
```
markhiveit, line 23: How come using two schedule clauses here and elsewhere? I'd have thought "before 8am every weekday" is alone sufficient?

renovate.json
```diff
@@ -0,0 +1,77 @@
...
+      "groupName": "python dependencies"
+    },
+    {
+      "matchDatasources": ["node-version"],
+      "groupName": "node version updates",
+      "schedule": [
+        "before 8am every weekday",
+        "every 3 months on the first day of the month"
+      ]
+    },
```
markhiveit, line 62: Is this going to suggest node major version updates? I assume we don't want that.

markhiveit (approved):

renovate.json
luke-h1: I've removed this for now as I believe it will try to upgrade to the latest major which is definitely not what we want. We can always keep an eye on the latest LTS releases for node anyways

renovate.json
luke-h1: As discussed I've reduced the granularity of the schedule and have changed the schedule to be `every month` / `every 3 months`

PR #4155, opened 2023-07-06
EES-4400 : Accessibility -Alt-text should not be repeated

N-moh:
Alternative text for the  element has text that duplicates the existing text.

Element Location:
.govuk-header__logotype-crown-fallback-image
GOV.UK
axe

### DIFF
--- a/src/explore-education-statistics-admin/src/components/PageHeader.tsx
+++ b/src/explore-education-statistics-admin/src/components/PageHeader.tsx
@@ -55,7 +55,7 @@ const PageHeader = ({ wide }: Props) => {
             >
               <span className="govuk-header__logotype">
                 <img
-                  alt="GOV.UK"
+                  alt=""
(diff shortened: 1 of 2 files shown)


luke-h1 (approved):
Looks good to me. I was debating whether to leave a comment about removing the alt attribute however upon doing so eslint seemed to raised an issue around image elements defaulting to an empty string for decorative images (full error: `img elements must have an alt prop, either with meaningful text, or an empty string for decorative images.`) so I think this should be fine

luke-h1 (approved):
Looks good to me. I was debating whether to leave a comment about removing the alt attribute however upon doing so eslint seemed to raised an issue around image elements defaulting to an empty string for decorative images (full error: `img elements must have an alt prop, either with meaningful text, or an empty string for decorative images.`) so I think this should be fine

PR #4159, opened 2023-07-07
Ees 4389 api and next caching for 11 july

duncan-at-hiveit:
## Overview

This PR contains a number of quick performance improvements ahead of the July 11th KS2 National Headlines Release.

This PR:
* adds a quick 10 second memory caching to the Publication lists requested from the Find Statistics page.  It caches any and all permutations of filters and sort orders at this stage
* adds a quick 10 second page caching of any Publication or direct Release pages on the public site, using Next.js's `getStaticProps` mechanism.  The behaviour here will be that the first request to a Publication / Release page will generate static HTML for it and subsequent requests will get the static HTML served up too.  After 10 seconds, the next request for that page will return the stale version and in the background a new version will be generated and then it will start to be served up instead.
* enables Live Metrics for the "public" App Insights, so we can see the live traffic as it comes.

### Caching behaviour on environments

Note that locally, all caching (in-memory API results and Next.js page-level caching) is set to 2 seconds, and in Dev is set to 10 seconds so that we can actually see a taste of the behaviour that real users will experience when testing on Dev, but without excessively slowing down UI tests.  Because all caching types are set to consistent values in Local and another set of consistent ones in Dev, `WAIT_MEMORY_CACHE_EXPIRY` has been renamed to simply `WAIT_CACHE_EXPIRY`.

The UI test pipeline has been updated already with this new `WAIT_CACHE_EXPIRY` variable.  When this work goes out, the old `WAIT_MEMORY_CACHE_EXPIRY` variable will be decommissioned.

### Performance impact of these pieces of work

The performance impact of the Publication API caching was such that we could get a scaled-up-and-out environment up to 1,200 RPS for the FInd Statistics page.

The performance impact of the Next.js page caching was that we could get the RPS up to 1,500 RPS for Release pages with only 3 instances of the public app service as opposed to 30!

### UI test run

image

(note that the UI tests will all green lights was achieved by commenting out ancillary file edit checks in `publish_release_and_cancel.robot` as there is an outstanding fix for this at the moment that's unrelated to this work)


tests/robot-tests/tests/libs/public-common.robot
```diff
@@ -60,3 +60,10 @@ user checks methodology note
     [Arguments]    ${number}    ${displayDate}    ${content}
     user waits until element contains    css:#methodologyNotes li:nth-of-type(${number}) time    ${displayDate}
     user waits until element contains    css:#methodologyNotes li:nth-of-type(${number}) p    ${content}
+
+# Next.js will have returned a stale cached page for the last request and will have created a new cached page in
+# the background in the meantime, so request the page again.
+
+user reloads the stale cached page
+    sleep    3
```
ntsim, line 68: Why this particular sleep? Should this be based on `WAIT_CACHE_EXPIRY` or some other variable?

tests/robot-tests/tests/admin_and_public/bau/publish_data.robot
```diff
@@ -505,12 +505,17 @@ Approve release
...
+    user waits for caches to expire
+
     user checks publication is on find statistics page    ${PUBLICATION_NAME}
 
 Navigate to published release page
     user clicks link    ${PUBLICATION_NAME}
     user waits until h1 is visible    ${PUBLICATION_NAME}    %{WAIT_MEDIUM}
 
+    user reloads the stale cached page
+    user waits until h1 is visible    ${PUBLICATION_NAME}    %{WAIT_MEDIUM}
```
ntsim, line 517: Doesn't seem like any of this is necessary as we're not doing an amendment, so the publication and release don't even exist beforehand?

tests/robot-tests/tests/admin_and_public_2/bau/publish_release_and_amend.robot
```diff
@@ -730,11 +730,16 @@ Approve amendment for scheduled release
...
 Verify amendment is on Find Statistics page again
+    user waits for caches to expire
     user checks publication is on find statistics page    ${PUBLICATION_NAME}
 
 Navigate to amendment release page
     user clicks link    ${PUBLICATION_NAME}
     user waits until h1 is visible    ${PUBLICATION_NAME}    %{WAIT_MEDIUM}
+
+    user reloads the stale cached page
+    user waits until h1 is visible    ${PUBLICATION_NAME}    %{WAIT_MEDIUM}
```
ntsim, line 741: Does kind of suck that we have to reload and sit around (similar thing in other places as well)! I wonder if it'd be possible to fire off a Python request to the page in an earlier part of the test before we arrive at this page?

Probably not necessary for this PR though.

tests/robot-tests/tests/admin_and_public/bau/publish_data.robot
duncan-at-hiveit: It's because the Publication slug URL is being hit rather than the direct Release one, hence we need to reload to get the latest Release's content.

tests/robot-tests/tests/admin_and_public_2/bau/publish_release_and_amend.robot
duncan-at-hiveit: I guess hopefully this is just a stop-gap until we have the next version of Next, and then we can invalidate on demand.  At which point, all of these keywords can be mopped up!

tests/robot-tests/tests/libs/public-common.robot
duncan-at-hiveit: No, it's arbitrarily the amount of time that Next needs to regenerate a page in the background.  Until this succeeds, the stale page will continue to be served up.  Could use WAIT_SMALL instead?

tests/robot-tests/tests/admin_and_public/bau/publish_data.robot
ntsim: Not sure I understand. From what I can see in the test suite, neither the publication or release have been published, and we haven't previously visited the publication URL, so why would this even be cached?

tests/robot-tests/tests/libs/public-common.robot
ntsim: `WAIT_SMALL` is 45 seconds by default though? That seems way too long for an explicit sleep.

Maybe it needs another env variable like `WAIT_NEXT_CACHE_EXPIRY` to set this more exactly for your environment.

tests/robot-tests/tests/admin_and_public/bau/publish_data.robot
duncan-at-hiveit: Hehe let me change it and give it a whirl!  I may have assumed too much here!

tests/robot-tests/tests/admin_and_public_2/bau/publish_release_and_amend.robot
ntsim: Yeh good point, hopefully that should be in fairly soon with the frontend deps upgrade.

tests/robot-tests/tests/admin_and_public/bau/publish_data.robot
duncan-at-hiveit: By 'eck you were right!  Removed this line, ta

ntsim (dismissed):

tests/robot-tests/tests/libs/public-common.robot
duncan-at-hiveit: Done - WAIT_NEXT_ISR, as discussed ta

ntsim (approved):

PR #4165, opened 2023-07-10
EES-4411 Add memory caching for Find Stats and Release pages

ntsim:
This PR implements memory caching for Find Statistics and Release pages.

We have also remove the caching via `getStaticProps` as this is potentially producing excessive CPU usage.


src/explore-education-statistics-frontend/server.js
```diff
@@ -112,6 +113,19 @@ async function startServer(port = process.env.PORT || 3000) {
...
   }
 
+  const cacheTtl = process.env.APP_ENV === 'Local' ? 2 : 10;
+
+  server.get(
+    '/find-statistics',
+    routeCache.cacheSeconds(cacheTtl),
+    (req, res) => handleRequest(req, res),
+  );
+  server.get(
```
duncan-at-hiveit, line 123: We'll need the plain /find-statistics/:publication path too, to catch views to the latest Release for a Publication.

ntsim: This route already handles that, it's using an optional `release` path segment.

luke-h1 (dismissed):

PR #4178, opened 2023-07-11
EES-4385 Add methodology status history

markhiveit:
This PR adds a new Methodology status history table to the Methodology sign off page:
image

### Notes
- The existing `MethodologyStatus` has been renamed to `MethodologyApprovalStatus`, matching `ReleaseApprovalStatus`.
- Existing `MethodologyVersion.InternalReleaseNote`s are migrated to `MethodologyStatus.InternalReleaseNote`s.

### Other changes
- Where applicable, we now call `BuildMethodologyVersionViewModel` instead of `GetMethodology` - saving a query and a permission check. Methods that used to call `GetMethodology` already had their own permission checks.
- Renamed some variables from `methodology` to `methodologyVersion` to make the type clearer.
- Removed an unnecessary query in `ValidateMethodologySlugUniqueForUpdate`.
- Remove internal release note comparison check in `IsStatusUpdateRequired`. I don't think we want a methodology status update if just an internal release note changes - I double checked with Laura.

There is further potential cleanup that could be done around the updating of methodologies, but it would take extra time to unpick. The difficulties of making these changes revolves around the unit tests - specifically how they interact with Entity Framework's tracking of entities and the additional mocking that would have to take place.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyApprovalServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyApprovalServiceTests.cs
@@ -1,6 +1,7 @@
 #nullable enable
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.Methodologies;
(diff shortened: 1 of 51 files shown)


luke-h1 (changes requested):

src/explore-education-statistics-admin/src/queries/methodologyQueries.ts
```diff
@@ -0,0 +1,13 @@
+import { createQueryKeys } from '@lukemorales/query-key-factory';
+import methodologyService from '@admin/services/methodologyService';
+
+const methodologyQueries = createQueryKeys('methodology', {
+  getMethodologyStatuses(methodologyId: string) {
+    return {
+      queryKey: ['getMethodologyStatuses', methodologyId],
```
ntsim, line 7: You don't need the `getMethodologyStatuses` part. If this has been copied from `releasePermissionQueries`, would suggest updating them to remove their `listReleaseBlah` strings as well

src/explore-education-statistics-admin/src/pages/methodology/edit-methodology/status/MethodologyStatusPage.tsx
```diff
@@ -159,6 +167,58 @@ const MethodologyStatusPage = () => {
...
+                        <thead>
+                          <tr>
+                            <th scope="col">Date</th>
+                            <th scope="col">Status</th>
+                            <th scope="col">Internal note</th>
+                            <th scope="col">Methodology version</th>
+                            <th scope="col">By user</th>
+                          </tr>
+                        </thead>
+                        {methodologyStatuses && (
```
luke-h1, line 188: We've already checked that `methodologyStatuses` isn't falsy on line 171 so shouldn't need this check

src/explore-education-statistics-admin/src/pages/methodology/edit-methodology/status/MethodologyStatusPage.tsx
```diff
@@ -40,6 +44,10 @@ const MethodologyStatusPage = () => {
 
   const [isEditing, toggleForm] = useToggle(false);
 
+  const { data: methodologyStatuses } = useQuery(
```
luke-h1, line 47: I imagine you'll need to wrap any tests of this page in a query client provider now

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyApprovalServiceTests.cs
```diff
@@ -256,6 +257,136 @@ await using (var context = InMemoryApplicationDbContext(contentDbContextId))
...
+            {
+                var service = SetupService(contentDbContext: context,
+                    methodologyContentService: contentService.Object,
+                    methodologyVersionRepository: methodologyVersionRepository.Object);
+
+                var updatedMethodologyVersion = (await service.UpdateApprovalStatus(methodologyVersion.Id, request)).AssertRight();
+
+                VerifyAllMocks(contentService, methodologyVersionRepository);
+
+                Assert.Equal(methodologyVersion.Id, updatedMethodologyVersion.Id);
```
duncan-at-hiveit, line 313: I'd be tempted to cut out all of these assertions that aren't related to checking the status history from this test, as they're tested elsewhere in other test methods.  Either that, or don't have the status history check in its own specific test method but rather bundle into a generic "approval" test method.  I'm aware there may be other test methods in this file that test perhaps more than they need to though!

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyApprovalServiceTests.cs
```diff
@@ -256,6 +257,136 @@ await using (var context = InMemoryApplicationDbContext(contentDbContextId))
...
+                await context.SaveChangesAsync();
+            }
+
+            await using (var context = InMemoryApplicationDbContext(contentDbContextId))
+            {
+                var service = SetupService(contentDbContext: context);
+
+                var updatedMethodologyVersion = (await service.UpdateApprovalStatus(methodologyVersion.Id, request)).AssertRight();
+
+                Assert.Equal(methodologyVersion.Id, updatedMethodologyVersion.Id);
```
duncan-at-hiveit, line 371: Same as above comment.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
```diff
@@ -2024,6 +2024,201 @@ await context.Methodologies.AsQueryable()
             }
         }
 
+        [Fact]
+        public async Task GetMethodologyStatuses()
+        {
+            var methodologyId = Guid.NewGuid();
+            var unrelatedMethodologyId = Guid.NewGuid();
+
+            var methodology = new Methodology
```
duncan-at-hiveit, line 2033: Looks like we're crying out for a Methodology test data generator :heart_eyes:

src/explore-education-statistics-admin/src/pages/methodology/edit-methodology/status/MethodologyStatusPage.tsx
markhiveit: Also removed a similar line in `ReleaseStatusPage.tsx`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20230706081603_MigrateMethodologyInternalReleaseNotes.sql
```diff
@@ -0,0 +1,11 @@
+﻿INSERT INTO [dbo].[MethodologyStatus] (Id, MethodologyVersionId, InternalReleaseNote,
+                                       ApprovalStatus, Created, CreatedById)
+SELECT NEWID() AS Id,
+       MV.Id AS MethodologyVersionId,
+       MV.InternalReleaseNote AS InternalReleaseNote,
```
duncan-at-hiveit, line 5: Will there be a follow-up piece of work to remove MethodologyVersion.InternalReleaseNote?

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20230706081603_MigrateMethodologyInternalReleaseNotes.sql
```diff
@@ -0,0 +1,11 @@
...
+       MV.Status AS ApprovalStatus,
+       MV.Published AS Created,
+       NULL AS CreatedById
+FROM [dbo].[MethodologyVersions] MV
+WHERE MV.Published IS NOT NULL;
```
duncan-at-hiveit, line 10: How come they need to be published already?  Do we just want to create ones for status histories that made the Methodology public?

tests/robot-tests/tests/admin_and_public/bau/publish_methodology.robot
```diff
@@ -381,6 +427,17 @@ Schedule a methodology amendment to be published with a release amendment
     ...    publishing_strategy=WithRelease
     ...    with_release=${PUBLICATION_NAME} - ${RELEASE_NAME}
 
+Check methodology status history table contains new row after approving amendment
```
duncan-at-hiveit, line 430: Do we check in any tests that we're not accidentally creating more status history items if we're *not* approving it, out of interest?  E.g. Draft => Draft, or Approved => Draft?

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20230706081603_MigrateMethodologyInternalReleaseNotes.sql
markhiveit: I had considered this and initially thought we should leave it - but having given it a closer look, I think we should. I think it'll just involve a small change to `BuildMethodologyVersionViewModel` and a little unit test juggling.

I'll create a subtask.

markhiveit: There is some question of how this should be implemented though. Right now, `InternalReleaseNote` only gets set when a release is approved. But a `MethodologyStatus` may contains a note when the methodology is switched to draft or higher review. So if we were to grab the latest `MethodologyStatus`, we will be changing the behaviour, I believe.

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20230706081603_MigrateMethodologyInternalReleaseNotes.sql
markhiveit: My thought was to try and reduce any potential confusion around what MethodologyStatus.Created datetime we set. It seems like a battle between MethodologyVersion.Published and MethodologyVersion.Updated - and I thought the published date is more reliable.

If we migrate `MethodologyVersion.Updated` into `MethodologyStatus.Created` instead, we could use `WHERE MV.Status = 'Approved'` instead.

Maybe that is marginally better, but I'm not fussed about it. Happy to switch to that or something different if you'd prefer.

tests/robot-tests/tests/admin_and_public/bau/publish_methodology.robot
markhiveit: I would expect it to create a new status history entry if it was moved from Approved => Draft, but not if it was moved from Draft => Draft or Approved => Approved, and with my manual testing (I believe) that is how it currently works. This is why I removed the InternalReleaseNote check from `IsStatusUpdateRequired`.

I don't think there are any tests around this at the moment. But this is deliberate as I didn't want to write a load of tests but then have to rewrite them once the Higher Review (EES-4376) work is done. I believe the risk in the mean time is minimal - I'm confident it works, but even if it didn't it would be a super easily fix, if we even considered a priority to fix.

All that said, happy to write more tests if you'd prefer.

PR #4180, opened 2023-07-12
EES-4420 - fix build number environment variable not being set

luke-h1:
This PR: 
* fixes the build number not being displayed on the public frontend. This stemmed from it not being explicitly set when the docker image is built

### DIFF
--- a/azure-pipelines.dfe.yml
+++ b/azure-pipelines.dfe.yml
@@ -275,6 +275,7 @@ jobs:
           Dockerfile: 'docker/Dockerfile.public-frontend'
           buildContext: '$(System.DefaultWorkingDirectory)'
           tags: $(Build.BuildNumber)
+          arguments: '--build-arg BUILD_BUILDNUMBER=$(Build.BuildNumber)'
 
   - job: 'MiscellaneousArtifacts'
     pool:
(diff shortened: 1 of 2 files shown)


benoutram (approved):

azure-pipelines.dfe.yml
```diff
@@ -275,6 +275,7 @@ jobs:
           Dockerfile: 'docker/Dockerfile.public-frontend'
           buildContext: '$(System.DefaultWorkingDirectory)'
           tags: $(Build.BuildNumber)
+          arguments: '--build-arg BUILD_BUILDNUMBER=$(Build.BuildNumber)'
```
ntsim, line 278: Is this actually sufficient? We need to convert the build argument into an environment variable inside of the image as well

luke-h1: Ah good point, missed that part of the article. I've added this to the first stage of the docker image. Given that Next.js will transform this variable to `BUILD_NUMBER` at build time I think this should be sufficient (and we shouldn't need to include it in the second stage)

ntsim: Cool, yep this looks better now

PR #4193, opened 2023-07-17
EES-4425 Prevent duplicate methodologies appearing on public Methodologies page

markhiveit:
This PR fixes a bug where methodologies could appear multiple times on the public Methodologies page if a single methodology was associated with multiple publications under the same theme.

I went with the simplest fix for this issue. I did consider updating the  `GET /methodology-themes` call in the backend, since I believe we only use this endpoint for the Methodologies page, and since we now only group methodologies by theme, there is no reason for the backend to return methodologies grouped by theme -> topic -> publication.

I discussed this with Dunc. This endpoint will be changed if/when remove topics (as part of EES-4260) so I believe it's ok to leave this for now. If others disagree, I'm happy to discuss / change the approach.

### DIFF
--- a/src/explore-education-statistics-frontend/src/modules/methodologies/MethodologyIndexPage.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/methodologies/MethodologyIndexPage.tsx
@@ -13,7 +13,7 @@ import orderBy from 'lodash/orderBy';
 import { GetServerSideProps, NextPage } from 'next';
 import React from 'react';
 import { MethodologySummary } from '@common/services/types/methodology';
-import { uniqWith } from 'lodash';
+import uniqBy from 'lodash/uniqBy';
 
 interface Props {
   themes: MethodologyTheme[];
@@ -116,9 +116,5 @@ function getMethodologiesForTopics(topics: MethodologyTopic[]) {
   const methodologies = topics.flatMap(topic =>
     topic.publications.flatMap(pub => pub.methodologies),
   );
-  return uniqWith(
-    methodologies,
-    (methodology1: MethodologySummary, methodology2: MethodologySummary) =>
-      methodology1.id === methodology2.id,
-  );
+  return uniqBy(methodologies, methodology => methodology.id);
 }
--- a/src/explore-education-statistics-frontend/src/modules/methodologies/MethodologyIndexPage.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/methodologies/MethodologyIndexPage.tsx
@@ -12,6 +12,8 @@ import PageSearchFormWithAnalytics from '@frontend/components/PageSearchFormWith
 import orderBy from 'lodash/orderBy';
 import { GetServerSideProps, NextPage } from 'next';
 import React from 'react';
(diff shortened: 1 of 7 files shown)


src/explore-education-statistics-frontend/src/modules/methodologies/MethodologyIndexPage.tsx
```diff
@@ -12,6 +12,8 @@ import PageSearchFormWithAnalytics from '@frontend/components/PageSearchFormWith
 import orderBy from 'lodash/orderBy';
 import { GetServerSideProps, NextPage } from 'next';
 import React from 'react';
+import { MethodologySummary } from '@common/services/types/methodology';
+import { uniqWith } from 'lodash';
```
ntsim, line 16: When using Lodash, we need use the fully qualified import i.e. 

```ts
import uniqWith from 'lodash/uniqWith';
```

ntsim (changes requested):

src/explore-education-statistics-frontend/src/modules/methodologies/__tests__/MethodologyIndexPage.test.tsx
```diff
@@ -101,6 +101,41 @@ const testThemes: MethodologyTheme[] = [
       },
     ],
   },
+  {
+    id: 'theme-3',
+    summary: 'Theme 3 summary',
+    title: 'Theme 3',
+    topics: [
```
ntsim, line 108: Personally, I probably would have leaned towards doing this in the backend by removing topics as we're in the area anyway.

src/explore-education-statistics-frontend/src/modules/methodologies/MethodologyIndexPage.tsx
```diff
@@ -111,7 +113,12 @@ export const getServerSideProps: GetServerSideProps<Props> = async () => {
 export default MethodologyIndexPage;
 
 function getMethodologiesForTopics(topics: MethodologyTopic[]) {
-  return topics.flatMap(topic =>
+  const methodologies = topics.flatMap(topic =>
```
ntsim, line 116: Personally, I would have leaned towards de-duping in the backend by removing topics as we're in the area anyway.

markhiveit: The backend isn't duplicating, really: it's just it provides the categories theme -> topic -> publication -> methodology - so if a methodology is in more than one topic or methodology under one theme, it will appear multiple times in the flattened results. So if we were to update the backend, we'd have to change the categories provided in the response to theme -> methodology - which is all the frontend requires, but I thought wasn't worth the effort.

src/explore-education-statistics-frontend/src/modules/methodologies/MethodologyIndexPage.tsx
markhiveit: There were a few other instances in the codebase where we did this, so I changed those as well.

ntsim (approved):

PR #4196, opened 2023-07-19
EES-3752 Add release checklist errors for empty Summary and Related dashboards section content blocks

benoutram:
This PR adds new checklist errors to the publishing checklist on the release sign off tab when the Summary or Related dashboards sections have an empty content block.

### Other changes

- Optimise the database queries checking for empty generic content sections and empty generic content section content blocks so that they no longer need to fetch all of the content sections and content.
- Change the UI tests to also test for checklist errors when there's an empty generic content section or a generic content section with an empty content block.

### Test report

image

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseChecklistService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseChecklistService.cs
@@ -121,7 +121,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
                     ValidationErrorMessages.GenericSectionsContainEmptyHtmlBlock));
             }
 
-            if (!await ReleaseHasKeyStatistic(release.Id) && !await ReleaseHasNonEmptyHeadlineBlock(release.Id))
+            if (!(await ReleaseHasKeyStatistic(release.Id) ||
+                  await ReleaseSectionHasNonEmptyHtmlBlock(release.Id, ContentSectionType.Headlines)))
             {
                 errors.Add(new ReleaseChecklistIssue(
                     ValidationErrorMessages.ReleaseMustContainKeyStatOrNonEmptyHeadlineBlock));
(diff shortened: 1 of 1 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseChecklistService.cs
```diff
@@ -143,28 +149,31 @@ private async Task<bool> ReleaseHasNonEmptyHeadlineBlock(Guid releaseId)
...
-            return releaseGenericContentBlocks
-                .Any(block => block is HtmlBlock htmlBlock && htmlBlock.Body.IsNullOrEmpty());
+        private async Task<bool> ReleaseRelatedDashboardsSectionContainsEmptyContentBlock(Guid releaseId)
+        {
+            return await _contentDbContext.ContentBlocks
+                .Where(cb =>
+                    cb.ContentSection!.Type == ContentSectionType.RelatedDashboards && 
+                    cb.ContentSection.Release.ReleaseId == releaseId)
+                .OfType<HtmlBlock>()
+                .AnyAsync(htmlBlock => string.IsNullOrEmpty(htmlBlock.Body));
```
markhiveit, line 176: Not particularly fussed about this, but you could accept a `ContentSectionType` as an argument and merge these two methods into one.

markhiveit: I think maybe `ReleaseHasNonEmptyHeadlineBlock` could also be removed as part of this - possibly not because I think that also ensures all headlines blocks are `HtmlBlock`s.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseChecklistService.cs
```diff
@@ -143,28 +149,31 @@ private async Task<bool> ReleaseHasNonEmptyHeadlineBlock(Guid releaseId)
...
-                    && rcs.ContentSection.Type == ContentSectionType.Generic)
-                .AnyAsync(rcs => rcs.ContentSection.Content.Count == 0);
+            return await _contentDbContext.ContentSections
+                .Where(cs =>
+                    cs.Type == ContentSectionType.Generic && 
+                    cs.Release.ReleaseId == releaseId)
+                .AnyAsync(cs => cs.Content.Count == 0);
         }
 
         private async Task<bool> ReleaseGenericContentSectionsContainEmptyContentBlock(Guid releaseId)
```
markhiveit, line 159: I think it was me that originally named this - so I'm criticising myself here - but I'd be tempted to rename these methods to use `HtmlBlock` instead of `ContentBlock`. Same for the new related dashboards method.

benoutram: Replaced them with a single method `ReleaseSectionHasEmptyHtmlBlock`.

benoutram:
@markhiveit mentioned to me on Slack that we don't check that the summary section hasn't been given an empty content block.

It was pretty quick to add that in following so the same logic as the related dashboards check so I've made that additional change now.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseChecklistService.cs
benoutram: The error produced by `ReleaseHasNonEmptyHeadlineBlock` needs to check there's a html block with content as opposed to checking there's no empty html blocks. I've added a new method `ReleaseSectionHasNonEmptyHtmlBlock` to do that and changed 

```
(!await ReleaseHasKeyStatistic(release.Id) &&
 !await ReleaseHasNonEmptyHeadlineBlock(release.Id))
```

to

```
if (!(await ReleaseHasKeyStatistic(release.Id) ||
      await ReleaseSectionHasNonEmptyHtmlBlock(release.Id, ContentSectionType.Headlines)))
```

markhiveit (approved):

PR #4212, opened 2023-07-27
Revert EES-4358

amyb-hiveit:
Reverts changes to tables made for EES-4358 because of problems with table rendering, see https://dfedigital.atlassian.net/browse/EES-4450


benoutram:
Our current thinking is that we won't revert EES-4358 using this PR, but fix forward by developing a new hotfix for EES-4450.

This is because there's no urgency to revert EES-4358 currently on Prod as it's not public due to the DNS switchover not taking place as scheduled yesterday.

PR #4217, opened 2023-08-02
Updating csv file size

N-moh:
Fixing UI test failure in general_public suite.
Changed 37 MB to 37 Mb


benoutram (approved):
Wait until you've confirmed the UI tests are passing locally with this change and the local db changes together, before merging

PR #4218, opened 2023-08-03
EES-2701 glossary plugin

amyb-hiveit:
Adds a CKEditor plugin to insert glossary entries from the available list.

The plugin itself is very simple, it calls a function in the admin app which displays a modal to select a glossary entry and then inserts it as a link. Having the modal in the app instead of within CKEditor makes it much simpler to have a searchable list and means we can make changes to it more easily.

Once added the glossasry links behave like normal links in CKEditor. As currently, glossary links in the release preview and published releases have icons and when clicked show the entry in a modal.

Also:
- Fixed a bug which caused an error when removed a link in CK Editor using the unlink option.
- Added a height restriction to the glossary modal as long entries could be taller than the viewport.

Toolbar button:
glossary-button
Modal:
glossary-modal


src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseBlock.tsx
```diff
@@ -7,6 +7,8 @@ import { Block } from '@common/services/types/blocks';
 import React from 'react';
 import useReleaseImageAttributeTransformer from '@common/modules/release/hooks/useReleaseImageAttributeTransformer';
 import Gate from '@common/components/Gate';
+import { useQuery } from '@tanstack/react-query';
+import glossaryQueries from 'src/queries/glossaryQueries';
```
ntsim, line 11: Unneeded imports?

src/explore-education-statistics-admin/src/components/editable/EditableContentForm.tsx
```diff
@@ -1,7 +1,7 @@
 import { useCommentsContext } from '@admin/contexts/CommentsContext';
 import styles from '@admin/components/editable/EditableContentForm.module.scss';
 import FormFieldEditor from '@admin/components/form/FormFieldEditor';
-import { Element } from '@admin/types/ckeditor';
+import { Element, GlossaryItem } from '@admin/types/ckeditor';
```
ntsim, line 4: Unneeded import?

src/explore-education-statistics-admin/src/components/form/FormEditor.tsx
```diff
@@ -306,15 +313,30 @@ const FormEditor = ({
...
+              />
+              <Modal
+                title="Insert glossary link"
+                open={showGlossaryModal}
+                onExit={toggleGlossaryModal.off}
+              >
+                <InsertGlossaryItemForm
+                  onCancel={toggleGlossaryModal.off}
+                  onSubmit={item => {
+                    glossaryPlugin.current?.addGlossaryItem(item);
```
ntsim, line 334: Upon adding a glossary link, we can see the link and interact with it in the read-only content block:

image

It might be a good idea to render this like the actual preview where `GlossaryEntryButton` is rendered instead for parity i.e.

image

ntsim (changes requested):

src/explore-education-statistics-admin/src/queries/glossaryQueries.ts
```diff
@@ -0,0 +1,13 @@
+import glossaryService from '@admin/services/glossaryService';
+import { createQueryKeys } from '@lukemorales/query-key-factory';
+
+const glossaryQueries = createQueryKeys('glossary', {
+  list() {
+    return {
+      queryKey: ['glossary'],
```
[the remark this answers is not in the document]
amyb-hiveit: I'm not quite sure what you're recommending here - `queryKey` is required,  I can't just have:
```
const glossaryQueries = createQueryKeys('glossary', {
  list() {
    return {
      queryFn: () => glossaryService.listEntries(),
    };
  },
});
```

Do you mean I should just have an empty string as the queryKey (`queryKey: ['']`) or that I don't need to use 'createQueryKeys' at all?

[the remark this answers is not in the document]
ntsim: Realised that it's not actually super clear how you define argument-less query keys.

Think this should potentially work:

```ts
const glossaryQueries = createQueryKeys('glossary', {
  list: {
    queryKey: null,
    queryFn: () => glossaryService.listEntries(),
  },
});
```

Note that it's no longer a callable query key anymore as it doesn't accept any args, so we'd just use it like:

```
useQuery(glossaryQueries.list);
```

src/explore-education-statistics-admin/src/components/form/FormEditor.tsx
amyb-hiveit: Clicking on the read only blocks is meant to open the editor, but currently when you click on a link it'll open that link instead, which probably isn't what you'd want as it will take you away from editing the release. Adding `GlossaryEntryButton` means that it'll flash up the modal briefly then open the editor. 

So I've done this:
- prevented links from opening in read only blocks, so if you click on a link it'll open the editor like when you click anywhere else in the block.
- added `GlossaryEntryButton` so the icon shows but disabled the modal from opening, so when you click on them it'll open the editor. 

What do you think?

ntsim: I think there's some accessibility problems with this approach. 

Screen reader users will still be able to tab to `GlossaryEntryButton` and the screen reader will announce that it's a button that you can show a glossary term definition for. This could be quite confusing when it actually does something completely different by changing the block to editing mode.

A similar thing will happen with links in that the screen reader will announce them, but the behaviour will not be what they're expecting.

There's two options from what I can see:

1. Leave interactive elements (buttons, links, etc) to work normally and don't toggle to editing mode if they directly interacted with one of these
2. Make any interactive elements within the preview container non-interactive

Option 2 might be preferable if we want to streamline the editing experience a bit more. There's a whole bunch of techniques for implementing it, but the simplest seems to be to add the `inert` attribute to a parent element. Seems like support is pretty good across browsers nowadays.

amyb-hiveit: Yeah good point. I've gone for option 2 using `inert` (with a workaround because of https://github.com/facebook/react/issues/17157)

src/explore-education-statistics-admin/src/components/editable/EditableContentBlock.tsx
```diff
@@ -195,10 +196,15 @@ const EditableContentBlock = ({
               ref={ref}
               onClick={isEditable ? onEditing : undefined}
             >
-              <ContentHtml
-                html={content || '<p>This section is empty</p>'}
-                sanitizeOptions={sanitizeOptions}
-              />
+              {/* Workaround to use inert, see https://github.com/facebook/react/issues/17157 */}
+              {/* eslint-disable-next-line react/jsx-props-no-spreading */}
+              <div {...{ inert: '' }}>
```
ntsim, line 201: It'd be better to do this via ambient declarations so we can have access to `inert` globally:


```ts
// types/react.d.ts

declare module 'react' {
  // eslint-disable-next-line @typescript-eslint/no-unused-vars
  interface HTMLAttributes<T> {
    inert?: '' | undefined;
  }
}

declare global {
  namespace JSX {
    interface IntrinsicAttributes {
      inert?: '' | undefined;
    }
  }
}

export {};
```

Also remember to re-export in `types.index.d.ts`:

```diff
export * from './govuk-frontend';
export * from './memoizee';
+ export * from './react';
export * from './react-app-env';
export * from './recharts';
export * from './util';
```

amyb-hiveit: ok, that's done now.

ntsim (approved):

PR #4220, opened 2023-08-03
EES-4358-4450 revert table rendering changes

amyb-hiveit:
Just in case we need it, this PR reverts table rendering to before the fixes for EES-4358 & EES-4450.


amyb-hiveit:
Closing as not needed now.

PR #4229, opened 2023-08-07
EES-4446 - removed Admin Banner and cookies usage from Admin project

duncan-at-hiveit:
This PR:
* removes the "Pl-EES-e" Admin Banner from the Admin site :sob: 
* retains the breaking out of the reusable `Banner.tsx` component, rather than reverting back to having the explicit UserTestingBanner component

image


src/explore-education-statistics-admin/package.json
```diff
@@ -103,7 +103,6 @@
     "react": "17.0.2",
     "react-app-polyfill": "^2.0.0",
     "react-beautiful-dnd": "^13.0.0",
-    "react-cookie": "^4.1.0",
```
amyb-hiveit, line 106: react-cookie is used in the `useCookie` hook added in https://github.com/dfe-analytical-services/explore-education-statistics/pull/4100, so we either need to remove that as well or keep both in case we need to use cookies in the admin again in future.

amyb-hiveit (changes requested):

amyb-hiveit (approved):

PR #4230, opened 2023-08-07
EES-4470 - debug loading of env vars

luke-h1:
This PR:
* debugs loading of environment variables on development after a rename of frontend environment variables

### DIFF
--- a/src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
@@ -169,6 +169,12 @@ const FindStatisticsPage: NextPage = () => {
             to help you understand and analyse our range of statistics.
           </p>
         </div>
+        <pre>
+          content API: {process.env.NEXT_PUBLIC_CONTENT_API_BASE_URL}
+          data API: {process.env.NEXT_PUBLIC_DATA_API_BASE_URL}
+          notification API: {process.env.NEXT_PUBLIC_NOTIFICATION_API_BASE_URL}
+        </pre>
+
         <div className="govuk-grid-column-one-third">
           <RelatedInformation heading="Related information">
             <ul className="govuk-list">


src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
```diff
@@ -169,6 +169,12 @@ const FindStatisticsPage: NextPage = () => {
             to help you understand and analyse our range of statistics.
           </p>
         </div>
+        <pre>
+          content API: {process.env.NEXT_PUBLIC_CONTENT_API_BASE_URL}
+          data API: {process.env.NEXT_PUBLIC_DATA_API_BASE_URL}
+          notification API: {process.env.NEXT_PUBLIC_NOTIFICATION_API_BASE_URL}
+        </pre>
```
ntsim, line 176: We don't really need this because we can see that we're getting `localhost:` base URLs in the frontend network requests

luke-h1: > We don't really need this because we can see that we're getting the default `localhost:5xxx` base URLs in the frontend network requests
> 
> This means we're not actually injecting overrides at all. This could be an ARM template or infrastructure deploy problem i.e. redeploy the infrastructure.

I looked into the app settings yesterday and can see they're been set which is why I wanted to try to see if they're available client side: 

image

luke-h1: > We don't really need this because we can see that we're getting the default `localhost:5xxx` base URLs in the frontend network requests
> 
> This means we're not actually injecting overrides at all. This could be an ARM template or infrastructure deploy problem i.e. redeploy the infrastructure.

I looked into the app settings yesterday and can see they're being set. This is the main reason I wanted to see if the environment variables are available client side 
Screenshot 2023-08-08 at 10 21 00

ntsim: They're not available client-side because you can simply browse to Find Statistics and see it in the network requests.

No need to dump this out onto the page.

PR #4232, opened 2023-08-08
EES-4470 - downgrade node

luke-h1:
This PR:
* Attempts to fix the recent issues related to environment variables not being available client side after https://dfedigital.atlassian.net/browse/EES-3409. This is potentially related to https://github.com/nodejs/node/issues/46167


ntsim (approved):
Environment variable loading doesn't seem likely to be related to the issue you quoted.

Not really sure why the Node version would have anything to do with this, but happy to give it a shot anyway.

PR #4233, opened 2023-08-08
EES-4460 New test data sets around removing filter/filter item

markhiveit:
Adding a couple of test data sets I created around replacing data:
- One replacement removes a filter found in the first data set
- The other replacement removes a filter item found in the first data set

### DIFF
--- a/tests/test-data/files/small-files/replace-data/remove_filter_before.meta.csv
+++ b/tests/test-data/files/small-files/replace-data/remove_filter_before.meta.csv
@@ -1,0 +1,6 @@
+col_name,col_type,label,indicator_grouping,indicator_unit,indicator_dp,filter_hint,filter_grouping_column
+pupil_sen_status,Filter,SEN provision,,,,EHC plan/Statement of SEN or SEN support/School Action/School Action plus,
+primary_need,Filter,Primary type of need,,,,"(ASD, HI, VI…)",
+number_of_pupils,Indicator,Headcount,Number of pupils,,0,,
+pupil_gender_boys,Indicator,Boys,Gender,,0,,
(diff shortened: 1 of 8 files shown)


tests/test-data/files/small-files/replace-data/remove_filter_before.meta.csv
```diff
@@ -0,0 +1,6 @@
+col_name,col_type,label,indicator_grouping,indicator_unit,indicator_dp,filter_hint,filter_grouping_column
+pupil_sen_status,Filter,SEN provision,,,,EHC plan/Statement of SEN or SEN support/School Action/School Action plus,
+primary_need,Filter,Primary type of need,,,,"(ASD, HI, VI�)",
```
benoutram, line 3: Is there a problem with the encoding of these files? Do they need to be saved as utf-8?
Same comment in `tests/test-data/files/small-files/replace-data/remove_filter_item_after_visual_impair.meta.csv`

markhiveit: Strictly, I think you're right. I created these files by copying and editing existing test data sets we have, so they'll also have this issue.

But practically, I don't think it's worth the effort to fix this - the files import and work and serve their purpose for testing. I don't think we actually gain anything by fixing this. If it becomes an issue in the future, we could fix it then.

benoutram: Cool, will approve it. It would have been nice not to have this problem in the data though which proliferates with copying and pasting 🙂 .

benoutram (approved):

PR #4234, opened 2023-08-08
EES-4447 adjust key stat widths

amyb-hiveit:
Changes Key Statistics so the widths are:
- when there's only one Key Stat, it's full width with centred text (I'll check Cam is ok with this before merging)
- when there are two or four Key Stats, they're half width
- any other number of Key Stats, they're one third width

Also, I renamed `column` / `hasColumn` to `wrapper` / `includeWrapper` as it's a wrapper around a single Key Stat rather that around a column of them.

ks1

ks2

ks3

ks4

### DIFF
--- a/src/explore-education-statistics-common/src/modules/find-statistics/components/KeyStat.module.scss
+++ b/src/explore-education-statistics-common/src/modules/find-statistics/components/KeyStat.module.scss
@@ -12,6 +12,7 @@
   margin: 0;
   padding: 0.2rem 0 0.2rem 0.5rem;
   position: relative;
+  text-align: left;
   width: 100%;
 
   &[open] {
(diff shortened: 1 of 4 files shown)


src/explore-education-statistics-common/src/modules/find-statistics/components/KeyStat.module.scss
```diff
@@ -63,13 +64,27 @@
   @include govuk-media-query($from: tablet, $until: desktop) {
     flex: 0 1 50%;
     max-width: 50%;
+
+    // If there is only one key stat it should be full width
+    &:only-child {
+      flex: 1;
+      max-width: 100%;
+      text-align: center;
+    }
```
ntsim, line 73: Could put this after the media queries to avoid repeating this for each media query?

amyb-hiveit: Yep, cheers.

PR #4246, opened 2023-08-10
EES-4413 alt text validation

amyb-hiveit:
- Fixes the validation for images without alt text
- In addition to the standard error, it shows a warning message with a link to more info about writing alt text
- When editing a block, images without alt text are outline in red
- Adds a UI test to check the validation works

Also, fixes the bug where large images can overflow content blocks.

alttext

### DIFF
--- a/src/explore-education-statistics-admin/src/components/editable/EditableContentBlock.tsx
+++ b/src/explore-education-statistics-admin/src/components/editable/EditableContentBlock.tsx
@@ -1,7 +1,9 @@
 import EditableBlockLockedMessage from '@admin/components/editable/EditableBlockLockedMessage';
 import { useCommentsContext } from '@admin/contexts/CommentsContext';
 import EditableBlockWrapper from '@admin/components/editable/EditableBlockWrapper';
-import EditableContentForm from '@admin/components/editable/EditableContentForm';
+import EditableContentForm, {
+  AltTextWarningMessage,
+} from '@admin/components/editable/EditableContentForm';
 import styles from '@admin/components/editable/EditableContentBlock.module.scss';
 import glossaryService from '@admin/services/glossaryService';
 import { UserDetails } from '@admin/services/types/user';
@@ -13,6 +15,7 @@ import toHtml from '@admin/utils/markdown/toHtml';
 import Button from '@common/components/Button';
 import ContentHtml from '@common/components/ContentHtml';
 import Tooltip from '@common/components/Tooltip';
+import useToggle from '@common/hooks/useToggle';
 import { Dictionary } from '@common/types';
 import sanitizeHtml, {
   defaultSanitizeOptions,
@@ -21,7 +24,7 @@ import sanitizeHtml, {
 } from '@common/utils/sanitizeHtml';
 import classNames from 'classnames';
 import mapValues from 'lodash/mapValues';
-import React, { useMemo } from 'react';
+import React, { useEffect, useMemo, useRef } from 'react';
 
 interface EditableContentBlockProps {
   actionThrottle?: number;
@@ -79,6 +82,8 @@ const EditableContentBlock = ({
   onSubmit,
 }: EditableContentBlockProps) => {
   const { comments } = useCommentsContext();
+  const [showAltTextMessage, toggleAltTextMessage] = useToggle(false);
+  const contentRef = useRef<HTMLDivElement>(null);
 
   const content = useMemo(
     () => (useMarkdown ? toHtml(value) : value),
@@ -119,6 +124,17 @@ const EditableContentBlock = ({
       },
     };
   }, [comments, transformImageAttributes]);
+
+  // Check content for images without alt text.
+  useEffect(() => {
+    if (content && !isEditing) {
+      const hasInvalidImage = !!contentRef.current?.querySelector(
+        'img:not([alt]), img[alt=""]',
+      );
+
+      toggleAltTextMessage(hasInvalidImage);
+    }
+  }, [content, isEditing, toggleAltTextMessage]);
 
   if (isEditing && !lockedBy) {
     return (
@@ -174,6 +190,8 @@ const EditableContentBlock = ({
         </div>
       )}
 
+      {showAltTextMessage && <AltTextWarningMessage />}
+
       {locked && lockedBy && (
         <EditableBlockLockedMessage locked={locked} lockedBy={lockedBy} />
       )}
@@ -196,7 +214,7 @@ const EditableContentBlock = ({
               ref={ref}
               onClick={isEditable ? onEditing : undefined}
             >
-              <div inert="">
+              <div inert="" ref={contentRef}>
                 <ContentHtml
                   getGlossaryEntry={glossaryService.getEntry}
                   html={content || '<p>This section is empty</p>'}
--- a/src/explore-education-statistics-admin/src/components/editable/EditableContentForm.tsx
+++ b/src/explore-education-statistics-admin/src/components/editable/EditableContentForm.tsx
@@ -1,7 +1,7 @@
 import { useCommentsContext } from '@admin/contexts/CommentsContext';
 import styles from '@admin/components/editable/EditableContentForm.module.scss';
 import FormFieldEditor from '@admin/components/form/FormFieldEditor';
-import { Element } from '@admin/types/ckeditor';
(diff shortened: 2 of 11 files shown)


src/explore-education-statistics-admin/src/components/editable/EditableContentForm.tsx
```diff
@@ -76,20 +78,29 @@ const EditableContentForm = ({
...
-    });
-
-    return error;
-  }, []);
+  function isInvalidImage(element: Node) {
+    return (
+      (element.name === 'imageBlock' || element.name === 'imageInline') &&
+      !element.getAttribute('alt')
+    );
+  }
```
ntsim, line 86: Normally we declare handlers inside of a function using arrow function syntax.

I think this function could probably be extracted to be outside of the component though.

src/explore-education-statistics-admin/src/components/editable/EditableContentForm.tsx
```diff
@@ -76,20 +78,29 @@ const EditableContentForm = ({
...
+
+  const validateElements = useCallback(
```
ntsim, line 88: Whilst it's valid to pop this error during `onBlur` validation, I wonder if there's anything we can do to re-validate when the image changes i.e. alt text is added?

Users could potentially find it confusing that adding alt text doesn't immediately make the error go away.

src/explore-education-statistics-admin/src/components/editable/EditableContentForm.tsx
```diff
@@ -152,6 +163,20 @@ const EditableContentForm = ({
 
             return (
               <Form id={`${id}-form`}>
+                {showAltTextMessage && (
+                  <WarningMessage>
```
ntsim, line 167: Given my other comment on adding outlines in content block previews, it might be a good idea to show this warning in the preview mode as well.

We could probably implement this via a DOM selector when the preview renders to trigger the message. The selector would just look for any images with no alt text of course.

ntsim (changes requested):

src/explore-education-statistics-admin/src/components/editable/EditableContentForm.tsx
```diff
@@ -76,20 +78,29 @@ const EditableContentForm = ({
...
+      const hasInvalidImage = elements.some(element =>
+        isInvalidImage(element)
+          ? true
+          : Array.from(element.getChildren()).some(child =>
+              isInvalidImage(child),
+            ),
+      );
+      toggleAltTextMessage(hasInvalidImage);
+      return hasInvalidImage
+        ? 'All images must have alternative (alt) text'
```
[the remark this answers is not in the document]
amyb-hiveit: The `(alt)` was already there on the previous version of the validation that didn't work, but you're right it's not necessary so I've removed it. 

For the error summary I've added a `screenReaderOnlyErrorSummary` option to the `Form` component which will visually hide the error summary but still announce the message to screen readers. When this is set the summary doesn't get focus and it lists the error messages without links.

src/explore-education-statistics-admin/src/components/editable/EditableContentForm.tsx
amyb-hiveit: Yep, I've moved it out.

src/explore-education-statistics-admin/src/components/editable/EditableContentForm.tsx
amyb-hiveit: The form was set to not revalidate on change, I couldn't see any reason why this is necessary and have checked the UI tests, so have removed it. Now it'll remove this and the 'Enter content' errors when they no longer apply.

src/explore-education-statistics-admin/src/components/editable/EditableContentForm.tsx
amyb-hiveit: Yep, showing the message and the red border now.

src/explore-education-statistics-admin/src/components/editable/EditableContentBlock.tsx
```diff
@@ -120,6 +125,43 @@ const EditableContentBlock = ({
...
+          return childrenArray.some(
+            (child: JSX.Element | string) =>
+              typeof child !== 'string' &&
+              child.type === 'img' &&
+              !child.props.alt,
+          );
+        }
+
+        return false;
+      });
```
ntsim, line 159: A much simpler way to do this is to use the DOM e.g.

```ts
const hasInvalidImage = !!contentRef.current?.querySelector(
  'img:not([alt]), img[alt=""]',
);

toggleAltTextMessage(hasInvalidImage);
```

This just needs a `contentRef` setup somewhere e.g. on the `inert` wrapper div:
```tsx
const contentRef = useRef<HTMLDivElement>(null);

<div inert="" ref={contentRef}>
  <ContentHtml ... />
</div>
```

Alternatively, could do it by adding using a `forwardRef` on `ContentHtml`

src/explore-education-statistics-admin/package.json
```diff
@@ -40,6 +40,7 @@
     "fs-extra": "^9.1.0",
     "geojson": "^0.5.0",
     "govuk-frontend": "^4.7.0",
+    "html-react-parser": "^1.3.0",
```
ntsim, line 43: We shouldn't need this if we just use the DOM to detect invalid images (see other comment)

src/explore-education-statistics-admin/src/components/editable/EditableContentBlock.tsx
```diff
@@ -150,6 +192,7 @@ const EditableContentBlock = ({
 
   return (
     <>
+      {showAltTextMessage && <AltTextWarningMessage />}
```
ntsim, line 195: Relatively minor, but would just have this next to the `EditableBlockLockedMessage` to make it more obvious

src/explore-education-statistics-common/src/components/ErrorSummary.tsx
```diff
@@ -87,13 +93,21 @@ const ErrorSummary = ({
...
           <li key={error.id}>
-            <a href={`#${error.id}`} onClick={onErrorClick}>
-              {error.message}
-            </a>
+            {!screenReaderOnly ? (
+              <a href={`#${error.id}`} onClick={onErrorClick}>
+                {error.message}
+              </a>
+            ) : (
+              <>{error.message}</>
```
ntsim, line 109: Don't think we need the wrapping fragment here?

src/explore-education-statistics-common/src/components/ErrorSummary.tsx
```diff
@@ -14,6 +15,7 @@ export interface ErrorSummaryMessage {
 
 interface BaseErrorSummaryProps {
   children: ReactNode;
+  screenReaderOnly?: boolean;
```
ntsim, line 18: Fairly fine as-is, but it's a little verbose so might be tempted by other options:

- `srOnly` (shorter, pretty common in various other libraries)
- `visuallyHidden` (more closely aligns with the name of the style)

src/explore-education-statistics-common/src/components/form/Form.tsx
```diff
@@ -23,6 +23,7 @@ import React, {
 interface Props {
   children: ReactNode;
   id: string;
+  screenReaderOnlyErrorSummary?: boolean;
```
ntsim, line 26: We should also add this for `RHFForm` as well

src/explore-education-statistics-admin/src/components/editable/EditableContentForm.tsx
```diff
@@ -20,6 +20,7 @@ import classNames from 'classnames';
 import { Formik, FormikHelpers } from 'formik';
 import React, { useCallback, useRef } from 'react';
 import { useIdleTimer } from 'react-idle-timer';
+import WarningMessage from '@common/components/WarningMessage';
```
ntsim, line 23: Nitpick, but this import could be alphabetized

ntsim (changes requested):

src/explore-education-statistics-common/src/components/ErrorSummary.tsx
amyb-hiveit: I've gone for visuallyHidden

ntsim (approved):

PR #4248, opened 2023-08-16
EES-4480 Use `SELECT INTO` rather that `INSERT INTO` in stored procedure `RebuildIndexes`

benoutram:
This PR follows up one of the recommendations made in [EES-3664](https://dfedigital.atlassian.net/browse/EES-3664) to use `SELECT INTO` rather than `INSERT INTO` in the stored procedure `RebuildIndexes`.


Dancemammal (approved):
Looks ok to me

PR #4254, opened 2023-08-21
EES-4444 Swap pyderman for webdriver-manager

ntsim:
This PR changes the downloading of Chromedriver to use [webdriver-manager](https://github.com/SergeyPirogov/webdriver_manager) instead of pyderman. This was motivated by a recent change to how Chromedriver downloads are published, meaning that JSON endpoints should be called for correct versions: https://github.com/GoogleChromeLabs/chrome-for-testing#json-api-endpoints

Currently, webdriver-manager supports this, but pyderman does not. See:
- https://github.com/SergeyPirogov/webdriver_manager/issues/580
- https://github.com/shadowmoose/pyderman/issues/35

## Webdriver download directory

Note that a major difference with pyderman is that Webdrivers are now installed globally to your home directory in the `~/.wdm` directory. The current `webdriver` directory in the project itself is now redundant and will not actually be used.


tests/robot-tests/scripts/get_webdriver.py
```diff
@@ -1,23 +1,16 @@
...
-    pyderman.install(
-        file_directory="./webdriver/",
-        filename=chromedriver_filename,
-        verbose=True,
-        chmod=True,
-        overwrite=False,
-        version=version,
-    )
+def get_webdriver(version: Optional[str] = None) -> None:
+    driver_path = ChromeDriverManager(driver_version=version).install()
```
markhiveit, line 9: Where is chromedriver getting saved?

Having it downloaded to the webdrivers directory previously was useful to allow you to manually download a different version and use that rather than what the script downloads, which has happened on occasion.

markhiveit (approved):

benoutram (approved):

tests/robot-tests/scripts/get_webdriver.py
ntsim: As mentioned in the PR summary, all webdrivers get downloaded to your home directory in `~/.wdm`. Each version downloaded gets its own directory, so you can just delete these as required.

However, I _think_ WDM by default will just try and figure out what Chrome you're using and get the correct version based on that. 

If you want to download a specific version, you should probably be using the `--chromedriver` flag instead when running the tests.

PR #4264, opened 2023-08-25
EES-4376 Don't show unapproved methodologies on prerelease and admin …

markhiveit:
…release content pages

This PR fixes a bug where unapproved methodologies would appear in release content under Quick links in admin and prerelease . Now only approved methodologies appear here.

NOTE: Filtering out methodologies that are approved isn't perfect, as it's still possible for a methodology to appear on prerelease, but for the methodology to not be live when the release is published, if the methodology is scheduled to be published with a different release - but this is such an edge cause I hesitated to even write this paragraph.

### Other changes

Also added a `isPreRelease` variable to the `ReleaseContent` component. This is so the link to a methodology under Quick links can be correct. 

NOTE: There is an existing issue where the link in Admin release content doesn't actually work. I've left this for now. I expect the `isPreRelease` variable could be used in other places in `ReleaseContent` and that `ReleaseContent` generally also needs a review to ensure it's aligned with the public release page.


src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ManageContent/ManageContentPageViewModel.cs
```diff
@@ -87,7 +87,7 @@ public class PublicationViewModel
 
             public Contact Contact { get; set; }
 
-            public List<IdTitleViewModel> Methodologies { get; set; }
+            public List<Content.Services.ViewModels.MethodologyVersionSummaryViewModel> Methodologies { get; set; }
```
luke-h1, line 90: Minor nitpick: just wondering if this could be shortened down to:

```cSharp
            public List<MethodologyVersionSummaryViewModel> Methodologies { get; set; }
```

src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseContent.tsx
```diff
@@ -87,11 +91,15 @@ const ReleaseContent = () => {
...
-    }));
+    const methodologies = publication.methodologies
+      .filter(methodology => methodology.status === 'Approved')
+      .map(methodology => ({
+        key: methodology.id,
+        title: methodology.title,
+        url: isPreRelease
+          ? `/publication/${publication.id}/release/${release.id}/prerelease/methodologies/${methodology.id}`
+          : `/methodology/${methodology.id}/summary`,
+      }));
```
benoutram, line 102: If we only ever list methodologies which are in status Approved, I'm wondering why we didn't just filter them in the back end, in method `GetManageContentPageViewModel`?

It doesn't seem great to expose unapproved methodologies in the back end response especially to pre-release users who maybe shouldn't even be aware of them.

src/explore-education-statistics-common/src/services/types/methodology.ts
```diff
@@ -1,7 +1,13 @@
+export type MethodologyApprovalStatus =
+  | 'Draft'
+  | 'HigherLevelReview'
+  | 'Approved';
```
ntsim, line 4: We shouldn't need to this in common as it's admin specific. I think this correlates with another comment by @benoutram about exposing these details to the public frontend.

Would suggest we revert all the changes around moving the status into common.

src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseContent.tsx
markhiveit: I think you're right - looks like I've just had an epic brain fart.

I'll close this PR and reopen a new one.

markhiveit:
Going to completely rework this PR

PR #4269, opened 2023-08-31
EES-4462 fix missing table header border style

amyb-hiveit:
Fixes a styling issue where the right border is missing on table header cells when a row above has a merged cell at the end.

I've gone for a slightly hacky CSS fix as it seemed the simplest way. It is possible using JS but would involve checking previous rows for merged cells at the end and adding a class to the affected cell to show the border, which seemed less efficient. Happy to discuss though.

**Before**
missingborder

**After**
fixed

### DIFF
--- a/src/explore-education-statistics-common/src/modules/table-tool/components/MultiHeaderTable.module.scss
+++ b/src/explore-education-statistics-common/src/modules/table-tool/components/MultiHeaderTable.module.scss
@@ -47,34 +47,39 @@ $border-width: 1px;
 
 .tableHead {
   td {
-    background: #fff;
+    background: govuk-colour('white');
     position: relative;
     z-index: 3;
   }
 
-  .emptyColumnHeaderCell {
-    background: govuk-colour('light-grey');
-    height: 46px;
-    z-index: 1;
+  // Uses the spacer cell to cover the left border
+  // of header cells. We can't just remove the
+  // left border on the first cell in header rows
+  // because this causes inconsistent styling when there
+  // are merged cells which span down through rows.
+  td::after {
+    background: govuk-colour('white');
+    bottom: 0;
+    content: '';
+    position: absolute;
+    right: -1px;
+    top: 0;
+    width: 1px;
   }
 
   th {
     background: govuk-colour('light-grey');
     border-bottom: 1px solid $govuk-border-colour;
-    border-right: 1px solid $govuk-border-colour;
+    border-left: 1px solid $govuk-border-colour;
     padding: govuk-spacing(2);
     position: relative;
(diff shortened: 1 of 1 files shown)


src/explore-education-statistics-common/src/modules/table-tool/components/MultiHeaderTable.module.scss
```diff
@@ -47,34 +47,39 @@ $border-width: 1px;
...
+    right: -1px;
+    top: 0;
+    width: 1px;
   }
 
   th {
     background: govuk-colour('light-grey');
     border-bottom: 1px solid $govuk-border-colour;
-    border-right: 1px solid $govuk-border-colour;
+    border-left: 1px solid $govuk-border-colour;
```
luke-h1, line 73: Super minor comment but I'm just wondering if we perhaps want a snapshot test for this in order to be an early warning system when someone might end up changing these styles in the future?

ntsim: A snapshot test won't work here unless we're talking about visual regression. That's a lot more work than just this ticket though.

src/explore-education-statistics-common/src/modules/table-tool/components/MultiHeaderTable.module.scss
```diff
@@ -47,34 +47,39 @@ $border-width: 1px;
...
-  .emptyColumnHeaderCell {
-    background: govuk-colour('light-grey');
-    height: 46px;
-    z-index: 1;
+  // Uses the spacer cell to cover the left border
+  // of header cells. We can't just remove the
+  // left border on the first cell in header rows
+  // because this causes inconsistent styling when there
+  // are merged cells which span down through rows.
+  td::after {
```
ntsim, line 60: Not a problem if you prefer it like this, but just wondering if having a left border is really a problem. Part of me wonders if we could just leave it in rather than do this?

amyb-hiveit: Not sure it looks quite right with a left border:
left-border
Maybe if had a right one as well:
borders

Either way, we should check with Cam before making a style change on the pubic site, so for now I'd prefer to keep the style as is.

ntsim: I was also wondering if it'd look better if we just shaded all of `thead` as well to make it look more symmetrical.

Anyway, fine to keep it as-is for now.

ntsim (approved):

PR #4272, opened 2023-09-04
EES-4507 fix data block save bug

amyb-hiveit:
Fixes a bug where when you edit an existing data block and save it the table and lists of selected filters etc revert to how they were before your changes. If you refreshed the changes had actually been saved.

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/datablocks/components/DataBlockPageTabs.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/datablocks/components/DataBlockPageTabs.tsx
@@ -377,7 +377,6 @@ const DataBlockPageTabs = ({
           <TabsSection title="Data source" id="dataBlockTabs-dataSource">
             {!isLoading && tableState && (
               <DataBlockSourceWizard
-                key={saveNumber}
                 dataBlock={dataBlock}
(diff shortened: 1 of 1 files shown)


src/explore-education-statistics-admin/src/pages/release/datablocks/components/DataBlockPageTabs.tsx
```diff
@@ -377,7 +377,6 @@ const DataBlockPageTabs = ({
           <TabsSection title="Data source" id="dataBlockTabs-dataSource">
             {!isLoading && tableState && (
               <DataBlockSourceWizard
-                key={saveNumber}
```
ntsim, line 380: We still use `saveNumber` for the chart tab. Do we still need this?

amyb-hiveit: It's used to prevent an error when you have a chart and change the filters etc. Though I have just found another bug while testing that.... I'll raise a separate ticket for it.

ntsim: Cool!

PR #4282, opened 2023-09-06
EES-4278 Use `releaseId` in call to create permalink if available and improve error handling

benoutram:
This PR makes a frontend change to use releaseId in call to create permalink if it's available which will be the case if navigating to the table tool from a release page.

This fixes a bug where a user can be viewing a table for a data set which is not part of the latest published release (in the time series) and without including the release id, no `ReleaseSubject` is found for the subject id of the query and the latest published release.

It also makes an improvement to the error handling so that the user is aware if an error has occurred when clicking 'Generate shareable link'.

### Other changes

- Change the backend to return 201 Created responses instead of 200 OK using `CreatedAtAction`.

### DIFF
--- a/src/explore-education-statistics-common/src/services/permalinkSnapshotService.ts
+++ b/src/explore-education-statistics-common/src/services/permalinkSnapshotService.ts
@@ -1,6 +1,6 @@
 import { TableJson } from '@common/modules/table-tool/utils/mapTableToJson';
 import { dataApi } from '@common/services/api';
-import { TableDataQuery } from '@common/services/tableBuilderService';
+import { ReleaseTableDataQuery } from '@common/services/tableBuilderService';
 import { Footnote } from '@common/services/types/footnotes';
 
 export type TableHeader =
@@ -40,15 +40,25 @@ export interface PermalinkSnapshot {
 }
 
 interface CreatePermalink {
-  query: TableDataQuery;
+  query: ReleaseTableDataQuery;
   configuration: {
     tableHeaders: UnmappedTableHeadersConfig;
   };
 }
 
 const permalinkSnapshotService = {
-  createPermalink(query: CreatePermalink): Promise<PermalinkSnapshot> {
-    return dataApi.post(`/permalink-snapshot`, query);
+  createPermalink(permalink: CreatePermalink): Promise<PermalinkSnapshot> {
+    const {
+      query: { releaseId, ...query },
+      configuration,
+    } = permalink;
+    const data = {
+      query,
+      configuration,
+    };
+    return releaseId
+      ? dataApi.post(`/permalink-snapshot/release/${releaseId}`, data)
+      : dataApi.post(`/permalink-snapshot`, data);
   },
   async getPermalink(id: string): Promise<PermalinkSnapshot> {
     return dataApi.get(`/permalink-snapshot/${id}`, {
--- a/src/explore-education-statistics-common/src/services/permalinkSnapshotService.ts
+++ b/src/explore-education-statistics-common/src/services/permalinkSnapshotService.ts
@@ -48,17 +48,10 @@ interface CreatePermalink {
 
 const permalinkSnapshotService = {
   createPermalink(permalink: CreatePermalink): Promise<PermalinkSnapshot> {
-    const {
-      query: { releaseId, ...query },
-      configuration,
-    } = permalink;
-    const data = {
-      query,
-      configuration,
-    };
+    const { releaseId } = permalink.query;
     return releaseId
-      ? dataApi.post(`/permalink-snapshot/release/${releaseId}`, data)
-      : dataApi.post(`/permalink-snapshot`, data);
+      ? dataApi.post(`/permalink-snapshot/release/${releaseId}`, permalink)
+      : dataApi.post(`/permalink-snapshot`, permalink);
   },
   async getPermalink(id: string): Promise<PermalinkSnapshot> {
     return dataApi.get(`/permalink-snapshot/${id}`, {
--- a/src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
@@ -7,7 +7,7 @@ import UrlContainer from '@common/components/UrlContainer';
 import { TableHeadersConfig } from '@common/modules/table-tool/types/tableHeaders';
 import mapUnmappedTableHeaders from '@common/modules/table-tool/utils/mapUnmappedTableHeaders';
 import permalinkSnapshotService from '@common/services/permalinkSnapshotService';
-import { TableDataQuery } from '@common/services/tableBuilderService';
+import { ReleaseTableDataQuery } from '@common/services/tableBuilderService';
 import ButtonLink from '@frontend/components/ButtonLink';
 import React, { useEffect, useState } from 'react';
 
@@ -16,7 +16,7 @@ const linkInstructions =
 
 interface Props {
   tableHeaders?: TableHeadersConfig;
-  query: TableDataQuery;
+  query: ReleaseTableDataQuery;
 }
 
 const TableToolShare = ({ tableHeaders, query }: Props) => {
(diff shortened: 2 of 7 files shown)


src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
```diff
@@ -16,13 +19,14 @@ const linkInstructions =
...
   tableHeaders?: TableHeadersConfig;
-  query: TableDataQuery;
+  query: ReleaseTableDataQuery;
 }
 
 const TableToolShare = ({ tableHeaders, query }: Props) => {
   const [permalinkUrl, setPermalinkUrl] = useState('');
   const [permalinkLoading, setPermalinkLoading] = useState<boolean>(false);
   const [screenReaderMessage, setScreenReaderMessage] = useState('');
+  const [permalinkError, togglePermalinkError] = useToggle(false);
```
ntsim, line 29: Would put this next to the other permalink states for better organisation

src/explore-education-statistics-common/src/services/permalinkSnapshotService.ts
```diff
@@ -40,25 +40,28 @@ export interface PermalinkSnapshot {
...
 
 interface CreatePermalink {
-  query: TableDataQuery;
+  query: ReleaseTableDataQuery;
   configuration: {
     tableHeaders: UnmappedTableHeadersConfig;
   };
 }
 
 const permalinkSnapshotService = {
```
ntsim, line 49: We can just call this `permalinkService` again too

src/explore-education-statistics-common/src/services/permalinkSnapshotService.ts
```diff
@@ -40,25 +40,28 @@ export interface PermalinkSnapshot {
...
-  createPermalink(query: CreatePermalink): Promise<PermalinkSnapshot> {
-    return dataApi.post(`/permalink-snapshot`, query);
+  createPermalink(permalink: CreatePermalink): Promise<PermalinkSnapshot> {
+    const { releaseId } = permalink.query;
+    return releaseId
+      ? dataApi.post(`/permalink/release/${releaseId}`, permalink)
+      : dataApi.post('/permalink', permalink);
```
ntsim, line 54: Not super necessary to change this in the hotfix, but not sure why we opted for a separate endpoints to attach a release?

We could just have a single `POST /api/permalink` endpoint which takes an optional `releaseId` in the request body. This would probably allow us to simplify the service code to a single method as well as part of this.

Would change this as part of this PR if it's not too much trouble, or consider changing it in a non-hotfix PR at least.

src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
```diff
@@ -34,17 +38,26 @@ const TableToolShare = ({ tableHeaders, query }: Props) => {
...
+        },
+      });
+
+      setPermalinkUrl(`${process.env.PUBLIC_URL}data-tables/permalink/${id}`);
+      setPermalinkLoading(false);
+
+      setScreenReaderMessage(`Shareable link generated. ${linkInstructions}`);
+    } catch (err) {
+      logger.error(err);
+      togglePermalinkError();
```
ntsim, line 55: Would suggest using normal state for the permalink error so we can make the error message consistent e.g.

```tsx
const [permalinkError, setPermalinkError] = useState<string>();

// Later...

const errorMessage = 'There was a problem generating the share link.'

setPermalinkError(errorMessage);
setScreenReaderMessage(`Error: ${errorMessage}`);

// Then in the render...

{permalinkError ? (
  <ErrorMessage>
    {permalinkError}
  </ErrorMessage>
) : (
  <ButtonText onClick={handlePermalinkClick}>
    Generate shareable link
  </ButtonText>
)}
```

src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
```diff
@@ -34,17 +38,26 @@ const TableToolShare = ({ tableHeaders, query }: Props) => {
     }
     setPermalinkLoading(true);
```
ntsim, line 40: We might also want to unset the permalink error here so that it gets removed if we try to generate the link again

src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
```diff
@@ -34,17 +38,26 @@ const TableToolShare = ({ tableHeaders, query }: Props) => {
...
+      setPermalinkLoading(false);
+
+      setScreenReaderMessage(`Shareable link generated. ${linkInstructions}`);
+    } catch (err) {
+      logger.error(err);
+      togglePermalinkError();
+      setScreenReaderMessage(
+        `Error: There was a problem generating the share link.`,
+      );
+      setPermalinkLoading(false);
```
ntsim, line 59: This can go in a `finally`

src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
```diff
@@ -34,17 +38,26 @@ const TableToolShare = ({ tableHeaders, query }: Props) => {
...
+    try {
+      const { id } = await permalinkSnapshotService.createPermalink({
+        query,
+        configuration: {
+          tableHeaders: mapUnmappedTableHeaders(tableHeaders),
+        },
+      });
+
+      setPermalinkUrl(`${process.env.PUBLIC_URL}data-tables/permalink/${id}`);
+      setPermalinkLoading(false);
```
ntsim, line 50: This can go in a `finally`

ntsim (changes requested):

src/explore-education-statistics-frontend/src/modules/table-tool/components/__tests__/TableToolShare.test.tsx
```diff
@@ -9,17 +10,23 @@ import _permalinkSnapshotService, {
 import { render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import React from 'react';
+import { ReleaseTableDataQuery } from '@common/services/tableBuilderService';
 
 jest.mock('@common/services/permalinkSnapshotService');
 
 const permalinkSnapshotService = _permalinkSnapshotService as jest.Mocked<
```
ntsim, line 17: As per other comments, this can be changed to `permalinkService`

src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
```diff
@@ -34,17 +38,26 @@ const TableToolShare = ({ tableHeaders, query }: Props) => {
...
-        tableHeaders: mapUnmappedTableHeaders(tableHeaders),
-      },
-    });
-
-    setPermalinkUrl(`${process.env.PUBLIC_URL}data-tables/permalink/${id}`);
-    setPermalinkLoading(false);
-
-    setScreenReaderMessage(`Shareable link generated. ${linkInstructions}`);
+    try {
+      const { id } = await permalinkSnapshotService.createPermalink({
```
ntsim, line 42: As per other comments, this can be changed to `permalinkService`

src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
benoutram: Done in combination with change to use normal state rather than toggle.

src/explore-education-statistics-common/src/services/permalinkSnapshotService.ts
benoutram: Renamed now, thanks. I didn't go as far as renaming `PermalinkSnapshot` to `Permalink`. Would you also recommend making that change?

src/explore-education-statistics-frontend/src/modules/table-tool/components/TableToolShare.tsx
benoutram: I've made this change now, thanks.

src/explore-education-statistics-common/src/services/permalinkSnapshotService.ts
benoutram: I'm not sure why that's the case, it's been that way since release amendments were introduced.

I initially thought about making this change but decided not to because ideally I wanted to combine it with making a  change to the `TableBuilderService.Query` to have optional `releaseId` which would clean things up by avoiding having to working out the release id in `PermalinkService`.

I'd also considered using `ReleaseSubjectService.Find` to find the ReleaseSubject from an optional releaseId, but looking a the code it doesn't quite have the same behaviour as the way we determine the releaseId in `PermalinkService` and `TableBuilderService`.

I've made the change you suggested on this PR now but still work out `releaseId` in `PermalinkService`. That's needed because we store set it on the permalink entity that's saved to the database for reference.

ntsim (approved):

PR #4285, opened 2023-09-07
EES-4513 sanitize data set summaries

amyb-hiveit:
Sanitizes data set summaries on save to prevent unwanted html, such as tables, which have been inserted via copy and paste.

As per the ticket acceptance criteria, this won't effect existing summaries.

PR #4286, opened 2023-09-07
EES-4512 allow links in footnotes

amyb-hiveit:
Adds the ability to put links in footnotes by changing it to use a very cut down CKEditor and making sure that any other attributes are stripped out.

footnoteform

### DIFF
--- a/src/explore-education-statistics-admin/src/components/form/FormEditor.tsx
+++ b/src/explore-education-statistics-admin/src/components/form/FormEditor.tsx
@@ -7,6 +7,8 @@ import {
   Editor as EditorType,
   Element,
   GlossaryPlugin,
+  PluginName,
+  ToolbarOption,
 } from '@admin/types/ckeditor';
 import { defaultAllowedHeadings } from '@admin/config/ckEditorConfig';
 import useCKEditorConfig from '@admin/hooks/useCKEditorConfig';
@@ -43,9 +45,10 @@ export interface FormEditorProps {
   hideLabel?: boolean;
   hint?: string;
   id: string;
+  includePlugins?: ReadonlySet<PluginName> | Set<PluginName>;
   label: string;
   testId?: string;
-  toolbarConfig?: string[];
+  toolbarConfig?: ReadonlyArray<ToolbarOption> | Array<ToolbarOption>;
   value: string;
   onAutoSave?: (values: string) => void;
   onBlur?: () => void;
@@ -66,6 +69,7 @@ const FormEditor = ({
   hideLabel,
   hint,
   id,
+  includePlugins,
   label,
   testId,
   toolbarConfig,
@@ -91,6 +95,7 @@ const FormEditor = ({
     allowComments,
     allowedHeadings,
     editorInstance,
+    includePlugins,
     toolbarConfig,
     onAutoSave,
     onCancelComment,
@@ -208,8 +213,14 @@ const FormEditor = ({
       );
 
       editorInstance.current = editor;
-      commentsPlugin.current = editor.plugins.get<CommentsPlugin>('Comments');
-      glossaryPlugin.current = editor.plugins.get<GlossaryPlugin>('Glossary');
+
+      if (editor.plugins.has<CommentsPlugin>('Comments')) {
+        commentsPlugin.current = editor.plugins.get<CommentsPlugin>('Comments');
+      }
+
+      if (editor.plugins.has<GlossaryPlugin>('Glossary')) {
+        glossaryPlugin.current = editor.plugins.get<GlossaryPlugin>('Glossary');
+      }
 
       setMarkersOrder(getMarkersOrder([...editor.model.markers]));
     },
--- a/src/explore-education-statistics-admin/src/config/ckEditorConfig.ts
+++ b/src/explore-education-statistics-admin/src/config/ckEditorConfig.ts
@@ -1,45 +1,57 @@
 import {
   AlignmentConfig,
   HeadingOption,
+  PluginName,
   ResizeOption,
+  ToolbarOption,
 } from '@admin/types/ckeditor';
-import { Dictionary } from '@admin/types';
 
-export const toolbarConfigs: Dictionary<string[]> = {
-  full: [
-    'heading',
-    '|',
-    'bold',
-    'italic',
-    'link',
-    '|',
-    'bulletedList',
-    'numberedList',
-    '|',
-    'blockQuote',
-    'insertTable',
-    'toggleTableCaption',
-    'imageUpload',
-    'alignment',
-    '|',
-    'redo',
-    'undo',
-    '|',
-    'comment',
-    'glossary',
-  ],
-  simple: [
-    'bold',
-    'italic',
-    'link',
-    '|',
-    'bulletedList',
-    'numberedList',
-    '|',
-    'redo',
-    'undo',
-  ],
-};
+export const toolbarConfigFull: ReadonlyArray<ToolbarOption> = [
+  'heading',
+  '|',
+  'bold',
+  'italic',
+  'link',
+  '|',
+  'bulletedList',
+  'numberedList',
+  '|',
+  'blockQuote',
+  'insertTable',
+  'toggleTableCaption',
+  'imageUpload',
+  'alignment',
+  '|',
+  'redo',
+  'undo',
+  '|',
+  'comment',
+  'glossary',
+];
+export const toolbarConfigSimple: ReadonlyArray<ToolbarOption> = [
+  'bold',
+  'italic',
+  'link',
+  '|',
+  'bulletedList',
+  'numberedList',
+  '|',
+  'redo',
+  'undo',
+];
+export const toolbarConfigLinkOnly: ReadonlyArray<ToolbarOption> = ['link'];
+
+export const corePlugins: ReadonlySet<PluginName> = new Set<PluginName>([
+  'Essentials',
+  'Paragraph',
+]);
+
+export const pluginsConfigLinksOnly: ReadonlySet<PluginName> =
+  new Set<PluginName>(['Link']);
+
+export const pluginsConfigSimple: ReadonlySet<PluginName> = new Set<PluginName>(
+  ['Bold', 'Italic', 'Link', 'List'],
+);
 
 export const defaultAllowedHeadings: string[] = ['h3', 'h4', 'h5'];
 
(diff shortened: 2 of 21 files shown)


ntsim (approved):

src/explore-education-statistics-admin/src/pages/release/footnotes/components/FootnoteForm.tsx
```diff
@@ -124,7 +129,11 @@ const FootnoteForm = ({
             to.
           </p>
 
-          <FormFieldTextArea<BaseFootnote> name="content" label="Footnote" />
+          <FormFieldEditor<BaseFootnote>
+            name="content"
+            label="Footnote"
+            toolbarConfig={['link']}
```
ntsim, line 135: I noticed this in #4285, but users will now be able to copy and paste in all sorts of elements from office docs:

image

Ideally, we should try and prevent users from doing this, although I can appreciate we're already letting this happen in some other places as well.

To do this, we have two options:

1. Manually disable plugins with the `removePlugins` config option (see [docs](https://ckeditor.com/docs/ckeditor5/latest/installation/getting-started/configuration.html#removing-features))
2. Create custom CKEditor classes which don't include any of these features in the first place.

**Option 1** is fairly straightforward, but the code isn't too great because we have to do stuff like this:

```ts
import Editor from 'explore-education-statistics-ckeditor';

// later in config...

removePlugins: Editor.builtinPlugins.filter(plugin => plugin.pluginName !== 'Link'),
```

Of course, we'd need to make it more flexible for different use-cases. I think it also breaks existing functionality where we expect the plugin to be installed e.g. comments. Nothing that can't necessarily be fixed though.

**Option 2** would be nicer, but requires a major re-think of how we create the CKEditor editor instances. 

I've also given everything a look and it's currently not looking too good, particularly as we managed to mess up the dependency upgrades and break the CKEditor build by trying to upgrade everything to bleeding edge. 

I gave downgrading a quick test, but there seems to be some other issues with that (probably because we've moved onto Node v18).

I've created [EES-4526](https://dfedigital.atlassian.net/browse/EES-4526) to handle upgrading CKEditor and fixing the build.

amyb-hiveit: I've gone with option 1 - you can configure which plugins to include. I've applied this to all the cases where we use the simple toolbar as well, so this will also fix [EES-4513](https://dfedigital.atlassian.net/browse/EES-4513) (I'll close the PR).

It seems that the 'Paragraph' plugin is needed for the editor to work so I'm still sanitising the footnote text as we don't want paragraphs in there.

ntsim (changes requested):

src/explore-education-statistics-admin/src/config/ckEditorConfig.ts
```diff
@@ -39,6 +39,12 @@ export const toolbarConfigs: Dictionary<string[]> = {
     'redo',
     'undo',
   ],
+  linkOnly: ['link'],
+};
+
+export const pluginsConfig: Dictionary<string[]> = {
+  linkOnly: ['Link', 'Essentials', 'Paragraph'],
```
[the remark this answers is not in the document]
amyb-hiveit: They are, the editor doesn't work without them, I've included them separately as core plugins.

src/explore-education-statistics-admin/src/components/form/FormEditor.tsx
```diff
@@ -208,9 +211,12 @@ const FormEditor = ({
       );
 
       editorInstance.current = editor;
-      commentsPlugin.current = editor.plugins.get<CommentsPlugin>('Comments');
-      glossaryPlugin.current = editor.plugins.get<GlossaryPlugin>('Glossary');
-
+      if (commentsPlugin.current) {
+        commentsPlugin.current = editor.plugins.get<CommentsPlugin>('Comments');
```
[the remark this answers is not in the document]
amyb-hiveit: I've changed this, in both cases, to check if the editor has the plugin before getting it. The get errors otherwise.

src/explore-education-statistics-admin/src/config/ckEditorConfig.ts
```diff
@@ -39,6 +39,12 @@ export const toolbarConfigs: Dictionary<string[]> = {
     'redo',
     'undo',
   ],
+  linkOnly: ['link'],
+};
+
+export const pluginsConfig: Dictionary<string[]> = {
```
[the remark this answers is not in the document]
amyb-hiveit: I've gone for separate imports as you suggest.

amyb-hiveit:
@ntsim I've responded to your comments and also realised that footnotes are in the ODS file download so have handled links there as well. I looked at having the links be clickable in the ODS file, but this seems to be only possible at the cell or all cell text level, so wouldn't work with multiple links in the same footnote. Instead I've appended the link url after the link text, e.g. "Link text (url)", using sanitize-html in a slightly hacky way as it doesn't provide the text in `transformTags`, see https://github.com/apostrophecms/sanitize-html/issues/581.

src/explore-education-statistics-common/src/modules/table-tool/components/utils/downloadTableOdsFile.ts
```diff
@@ -120,7 +122,26 @@ export function appendFootnotes(
...
+          // tags and append the urls after the link text.
+          v: `(${index + 1}) ${sanitizeHtml(footnote.label, {
+            allowedTags: [],
+            transformTags: {
+              a: (tagName, attribs) => {
+                linkHref = attribs.href;
+                return {
+                  tagName,
+                  attribs,
+                };
```
ntsim, line 135: What an awkward API. It would have been nice of them to provide the text in the callback params.

src/explore-education-statistics-common/src/modules/table-tool/components/utils/downloadTableOdsFile.ts
```diff
@@ -108,6 +109,7 @@ export function appendFootnotes(
   sheet: WorkSheet,
   footnotes: FullTableMeta['footnotes'],
 ): WorkSheet {
+  let linkHref: string;
```
ntsim, line 112: - Minor, but would suggest adding a line beneath to space things out
- Would add a comment explaining that this is a bit of a workaround to get around sanitize-html's API

src/explore-education-statistics-admin/src/utils/ckeditor/customUploadAdapterPlugin.ts
```diff
@@ -9,8 +9,15 @@ export default function customUploadAdapterPlugin(
...
-  return class CustomUploadAdapterPlugin {
-    constructor(editor: Editor) {
+  return class CustomUploadAdapterPlugin implements Plugin {
+    static get pluginName(): PluginName {
+      return 'CustomUploadAdapter';
+    }
+
+    constructor(
+      editor: Editor,
+      readonly pluginName: PluginName,
```
ntsim, line 19: Ah, apologies, looks we need to change the types a bit more:

```ts
// types/ckeditor.ts

export interface PluginClass {
  get pluginName(): PluginName;
}

// eslint-disable-next-line @typescript-eslint/no-empty-interface
export interface Plugin {}
```

Then we can use `PluginClass` for the adapter so we don't need to provide this as a property e.g.

```ts
return class CustomUploadAdapterPlugin implements PluginClass {
  get pluginName(): PluginName {
    return 'CustomUploadAdapter';
  }

  constructor(editor: Editor) {
    const fileRepository =
      editor.plugins.get<FileRepository>('FileRepository');

    fileRepository.createUploadAdapter = loader => {
      return new CustomUploadAdapter(loader, onUpload, onCancel);
    };
  }
};
```

We also need to change `CustomEditor` to use this as well:

```ts
// types/ckeditor-ambient.d.ts

interface CustomEditor extends EditorClass {
    builtinPlugins?: PluginClass[];
}
```

src/explore-education-statistics-common/src/modules/table-tool/components/utils/downloadTableOdsFile.ts
amyb-hiveit: yep!

src/explore-education-statistics-admin/src/utils/ckeditor/customUploadAdapterPlugin.ts
amyb-hiveit: Cool, sorted now,

src/explore-education-statistics-admin/src/components/form/FormEditor.tsx
```diff
@@ -43,9 +45,10 @@ export interface FormEditorProps {
   hideLabel?: boolean;
   hint?: string;
   id: string;
+  includePlugins?: ReadonlySet<PluginName>;
```
ntsim, line 48: We probably want to make this a little more flexible so it's less awkward to provide configs for very bespoke cases:

```ts
includePlugins?: ReadonlySet<PluginName> | Set<PluginName>;
```

src/explore-education-statistics-admin/src/components/form/FormEditor.tsx
```diff
@@ -43,9 +45,10 @@ export interface FormEditorProps {
...
   label: string;
   testId?: string;
-  toolbarConfig?: string[];
+  toolbarConfig?: readonly ToolbarOption[];
```
ntsim, line 51: Similar to above:

```ts
toolbarConfig?: ReadonlyArray<ToolbarOption> | Array<ToolbarOption>;
```

src/explore-education-statistics-admin/src/hooks/useCKEditorConfig.ts
```diff
@@ -29,13 +37,14 @@ const useCKEditorConfig = ({
...
-  editorInstance?: MutableRefObject<Editor | undefined>;
+  editorInstance?: MutableRefObject<EditorType | undefined>;
   glossaryItems?: {
     title: string;
     slug: string;
     body: string;
   }[];
-  toolbarConfig?: string[];
+  includePlugins?: ReadonlySet<PluginName>;
+  toolbarConfig?: readonly ToolbarOption[];
```
ntsim, line 47: Similar to `FormEditor`, we could make these types more flexible by allowing the non-readonly types as well

ntsim (approved):

PR #4287, opened 2023-09-08
chore: Remove ignored Mime dependency from Renovate and tidy up config

benoutram:
This PR makes changes to the Renovate configuration.

- ~~It adds `ignoreDeps` configuration to ignore Moq.~~ We decided as a team to continue to receive Moq version updates.
- It removes the ignore on the Mime dependency which is not required after #4283.
- It switches ignored Mime-Detective to use `ignoreDeps` as a shorter way of ignoring dependencies than using `matchPackagePatterns` combined with `enabled`.
- It removes the ignore on ckeditor because all frontend dependencies are ignored.
- It removes the following config because that doesn't appear to be supported in the docs:

```
    {
      "packageNames": ["node"],
      "enabled": false
    },
```

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -10,6 +10,7 @@
   "python": {
     "enabled": false
   },
+  "ignoreDeps": ["Mime-Detective"],
   "vulnerabilityAlerts": {
     "groupName": "renovate",
     "dependencyDashboardApproval": false,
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -10,6 +10,7 @@
   "python": {
     "enabled": false
   },
+  "ignoreDeps": ["Mime", "Mime-Detective", "Moq"],
```
ntsim, line 13: I'd suspect we can get rid of Mime now. #4283 should now bring Mime's native libs into the build correctly, so version pinning isn't a concern anymore.

ntsim (approved):

PR #4295, opened 2023-09-11
EES-4530 Change Renovate config to ignore upgrading Automapper

benoutram:
This PR changes the Renovate config to restrict the versions of `Automapper` and `AutoMapper.Extensions.Microsoft.DependencyInjection` due to the versions of `Microsoft.AspNetCore.ApiAuthorization.IdentityServer:3.1.32`  -> `IdentityServer4.EntityFramework.Storage:3.0.0` -> that we use  in Admin depending on `AutoMapper` 9.0.0.

It's possible that if we did some additional work to upgrade `IdentityServer`, `IdentityServer4.EntityFramework` and `IdentityServer4.EntityFramework.Storage` to 3.1.4 then we might be able to use more recent versions (but still old) i.e. `AutoMapper` 10.1.1 and `AutoMapper.Extensions.Microsoft.DependencyInjection` 8.1.1.

However when I attempted this update I consistently got the following error:

```
IdentityServer4.Hosting.IdentityServerMiddleware[0]
      Unhandled exception: Method not found: '!!0 AutoMapper.IMapper.Map(System.Object)'.
      System.MissingMethodException: Method not found: '!!0 AutoMapper.IMapper.Map(System.Object)'.
         at IdentityServer4.EntityFramework.Mappers.PersistedGrantMappers.ToEntity(PersistedGrant model)
```

Using an outdated version of IdentityServer is the problem here and it's going to cause us a major problem if we want to update to .NET 7 or the upcoming LTS release, .NET 8.

We can't update to the latest version of IdentityServer without a licence and IdentityServer4 already reached end-of-support on 13/12/22. IdentityServer4 won't be receiving any updates like we'd need to upgrade to future versions of .NET or to fix this issue with AutoMapper.

### Other changes
- Remove `IgnoreMap` from `EmbedBlockLink`. It shouldn't be needed because `EmbedBlock` is not a field of the target type `EmbedBlockLinkViewModel`. `IgnoreMap` is removed by AutoMapper v11.
- Declare `AutoMapper.Extensions.Microsoft.DependencyInjection` as a dependency only in projects where it's required.
- Add a TODO to remind us that `AutoMapper.Extensions.Microsoft.DependencyInjection` is being removed with `AutoMapper` 13 (it's getting merged into `AutoMapper`).

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -74,6 +74,18 @@
       ],
       "matchUpdateTypes": ["major"],
       "enabled": false
+    },
+    {
+      "description": "Restrict AutoMapper due to incompatibility with IdentityServer4.EntityFramework.Storage",
+      "matchDatasources": ["nuget"],
+      "matchPackageNames": ["AutoMapper"],
+      "allowedVersions": "<=9.0.0"
+    },
+    {
+      "description": "Restrict AutoMapper.Extensions.Microsoft.DependencyInjection to correspond with AutoMapper version",
+      "matchDatasources": ["nuget"],
+      "matchPackageNames": ["AutoMapper.Extensions.Microsoft.DependencyInjection"],
+      "allowedVersions": "<=7.0.0"
     }
   ]
 }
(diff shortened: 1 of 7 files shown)


renovate.json
```diff
@@ -74,6 +74,18 @@
       ],
       "matchUpdateTypes": ["major"],
       "enabled": false
+    },
+    {
+      "description": "Restrict AutoMapper due to incompatibility with IdentityServer4.EntityFramework.Storage",
```
ntsim, line 79: Part of me thinks it might be easier to just get rid of AutoMapper completely. This is not to say we shouldn't upgrade Identity Server though.

Personally, I'd be all for just hand-rolling our mappings, but if not, there are obviously other libraries we could use (e.g. Mapperly maybe).

ntsim (approved):

renovate.json
benoutram: I agree, I'm up for stopping using it. I would already replace simple mappings right now and probably just leave the complicated ones where we save tons of lines by using a mapper, until we reach a decision going forward.

Without upgrading IdentityServer we are also exposed to old transitive dependencies (like we'd still be including an old version of AutoMapper indirectly).

PR #4299, opened 2023-09-12
EES-4509 remove publication info from release summary tab

amyb-hiveit:
Removes the publication details from the release summary tab in the admin.

Also, removes an unused import from `ChartAxisConfiguration.tsx`


src/explore-education-statistics-admin/src/pages/release/ReleaseSummaryPage.tsx
```diff
@@ -41,20 +41,6 @@ const ReleaseSummaryPage = () => {
...
-            </SummaryListItem>
-            <SummaryListItem term="Publication summary">
-              {release.publicationSummary}
-            </SummaryListItem>
-            <SummaryListItem term="Lead statistician">
-              {release.contact?.contactName}
-            </SummaryListItem>
-          </SummaryList>
-
-          <h3>Release summary</h3>
```
benoutram, line 57: Looks good. Just a quick observation, do you know if this leaves any of these fields unused?

If so it might be possible that we can tidy the following up from the back end.

https://github.com/dfe-analytical-services/explore-education-statistics/blob/dfdb485381861537ccbae4114dc330db84268872/src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ReleaseViewModels.cs#L24-L26

https://github.com/dfe-analytical-services/explore-education-statistics/blob/dfdb485381861537ccbae4114dc330db84268872/src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ReleaseViewModels.cs#L55

amyb-hiveit: Yep looks like they aren't used anywhere else so I've removed them from the front end and back end  (hopefully not got the BE part too wrong!)

benoutram: Cool 😃 . After your backend change to stop populating it in the `ReleaseViewModel`, we no longer need to loading the Contact from the database at the same time as we retrieve the Release. That's in method `ReleaseService.HydrateRelease`. 

I've just made this change and also tidied up a bit of the test data which was setting up contacts to satisfy `HydrateRelease`.

benoutram (approved):

PR #4305, opened 2023-09-14
EES-4531 Fix unpublished adopted methodology bug

markhiveit:
This fixes an issue that arose of the work to introduce the `Methodology.LatestPublishedVersionId` column in EES-4504.

We missed the scenario where a unpublished methodology could be published if it is adopted by a publication that is already live.

Before EES-4504, this would have meant that the methodology version's files wouldn't have been published and it's published date wouldn't have been set, but as whether it was public was determined by IsPubiclyAccessible (now IsToBePublished), they would appear on the public page. After the ticket, the methodology version won't appear to be published at all, since now the service mostly checks the Methodology.LatestPublishedVersionId column to determine if something is live.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using AutoMapper;
 using GovUk.Education.ExploreEducationStatistics.Admin.Models;
+using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.Methodologies;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.Security;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Util;
@@ -35,6 +36,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services.Methodologie
         private readonly IMethodologyImageService _methodologyImageService;
         private readonly IMethodologyApprovalService _methodologyApprovalService;
         private readonly IMethodologyCacheService _methodologyCacheService;
+        private readonly IPublishingService _publishingService;
         private readonly IUserService _userService;
 
         public MethodologyService(
@@ -46,6 +48,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services.Methodologie
             IMethodologyImageService methodologyImageService,
             IMethodologyApprovalService methodologyApprovalService,
             IMethodologyCacheService methodologyCacheService,
+            IPublishingService publishingService,
             IUserService userService)
         {
             _persistenceHelper = persistenceHelper;
@@ -56,6 +59,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services.Methodologie
             _methodologyImageService = methodologyImageService;
             _methodologyApprovalService = methodologyApprovalService;
             _methodologyCacheService = methodologyCacheService;
+            _publishingService = publishingService;
             _userService = userService;
         }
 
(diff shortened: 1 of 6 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -79,6 +85,19 @@ public class MethodologyService : IMethodologyService
                         Owner = false
                     });
 
+                    // If the adopted methodology is unpublished, it may require publishing,
+                    // if the adopting publication is live
+                    var versionToBePublishedId = await _methodologyVersionRepository.IsToBePublished(methodology);
```
duncan-at-hiveit, line 90: I suppose this method should be called `GetMethodologyVersionToBePublished()`?

duncan-at-hiveit: I wonder if this whole block of code (which sets the published date and publishes the files) should be moved into MethodologyApprovalService and called from here.  See the already- similar logic from MethodologyApprovalService.UpdateStatus():

```
                        // We cannot rely on Methodology.LatestPublishedVersionId, as it may now be incorrect,
                        // if we are approving and publishing this methodology version.
                        var isToBePublished = await _methodologyVersionRepository.IsToBePublished(methodologyVersion);

                        if (isToBePublished)
                        {
                            methodologyVersion.Published = DateTime.UtcNow;
                            methodologyVersion.Methodology.LatestPublishedVersionId = methodologyVersion.Id;

                            await _publishingService.PublishMethodologyFiles(methodologyVersion.Id);
                        }
```

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/MethodologyVersionRepository.cs
```diff
@@ -133,15 +133,15 @@ await _contentDbContext.Entry(methodology)
...
-            {
-                methodology.OwningPublicationTitle = updatedTitle;
-
-                if (slugChanged && methodology.Slug == originalSlug &&
-                    methodology.LatestPublishedVersionId == null)
                 {
-                    methodology.Slug = updatedSlug;
-                }
-            });
+                    methodology.OwningPublicationTitle = updatedTitle;
```
duncan-at-hiveit, line 137: Has anything changed here?

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/MethodologyVersionRepository.cs
```diff
@@ -189,6 +189,24 @@ await IsVersionScheduledForPublishingWithPublishedRelease(nextVersion))
             return await IsVersionScheduledForPublishingWithPublishedRelease(methodologyVersion);
         }
 
+        public async Task<Guid?> IsToBePublished(Methodology methodology)
```
duncan-at-hiveit, line 192: See other comment about renaming.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/MethodologyVersionRepository.cs
markhiveit: Nope. Rider autoindented this.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -79,6 +85,20 @@ public class MethodologyService : IMethodologyService
...
+                    // If the adopted methodology is unpublished, it may require publishing,
+                    // if the adopting publication is live
+                    var versionToBePublishedId = await _methodologyVersionRepository
+                        .GetMethodologyVersionToBePublished(methodology);
+                    if (versionToBePublishedId != null)
+                    {
+                        var versionToBePublished = _context.MethodologyVersions
+                            .Single(mv => mv.Id == versionToBePublishedId);
+                        versionToBePublished.Published = DateTime.UtcNow;
+                        versionToBePublished.Methodology.LatestPublishedVersionId = versionToBePublished.Id;
```
benoutram, line 97: Not really relevant to this PR, but relevant in the sense that it might be something else not working, similar code to this was added in to the Publisher in #4267 

https://github.com/dfe-analytical-services/explore-education-statistics/blob/c31ab4b4ae662407ac405af4eb64552ca1bd9e49/src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/MethodologyService.cs#L64-L76

This code is missing an `Update` on the entity being updated, so it's either not getting updated, or it's relying on something else already calling update elsewhere.

It also replaced a method called `SetPublishedDatesIfApplicable` which has been left unused in the code.

benoutram (changes requested):

markhiveit:
Closing this PR as it is going to be reworked.

We're changing tack, and will instead not allow unpublished methodologies to be adopted.

PR #4306, opened 2023-09-14
EES-4510 release type modal on release page

amyb-hiveit:
Moves the release type from a tag at the top to the summary list. Clicking it will open a modal containing information about the release type.

I've done a bit of tidying to reduce inconsistencies between the admin and public versions of the release pages:
- added a `BasicReleaseSummary` component for use in both and removed the admin only one
- combined both the `ReleaseHelpAndSupportSection` components into one 

Also, I've made a change to the `Modal` component to make it simpler to have the content be scrollable and have a close button. This is used on the Glossary and Find Stats - Filters modals as well as the new one for release type.

releasetype-pr


src/explore-education-statistics-common/src/modules/release/components/BasicReleaseSummary.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+interface Props {
+  isEditing?: boolean;
+  lastUpdated?: Date;
+  release: Release;
+  releaseDate: string;
+  renderReleaseNotes: ReactNode;
+  renderStatusTags: ReactNode;
+  renderSubscribeLink: ReactNode;
+  onToggleReleaseTypeModal?: () => void;
```
ntsim, line 43: `onShowReleaseTypeModal` might be better. `onToggle` implies the event should fire when the modal closes too.

src/explore-education-statistics-common/src/modules/release/components/BasicReleaseSummary.tsx
```diff
@@ -0,0 +1,127 @@
...
+          </ButtonText>
+        </SummaryListItem>
+
+        <SummaryListItem term="Receive updates">
+          {renderSubscribeLink}
+        </SummaryListItem>
+      </SummaryList>
+
+      <Modal
+        className="govuk-!-width-one-half"
```
ntsim, line 116: Not sure this responds too great on smaller desktops / tablets: 

image

It doesn't seem like we really need this at all personally.

Without the width:

image

With the width:

image

src/explore-education-statistics-common/src/services/types/releaseType.ts
```diff
@@ -7,4 +13,12 @@ export const releaseTypes = {
   ManagementInformation: 'Management information',
 } as const;
 
+export const releaseTypeComponents = {
+  NationalStatistics: NationalStatisticsSection,
+  OfficialStatistics: OfficialStatisticsSection,
+  ExperimentalStatistics: ExperimentalStatisticsSection,
+  AdHocStatistics: AdHocStatisticsSection,
+  ManagementInformation: ManagementInformationSection,
+};
```
ntsim, line 22: Doesn't seem appropriate to put this in here. This is the `services` directory, so we shouldn't need to put React components in here. 

These are also quite specific to the section components for the release pages, which doesn't make a lot of sense for a file that will be used more generally.

Would suggest we just create a dedicated `ReleaseTypeSection` component that takes a `type` prop and wraps these other sections. We might even want to just move all the other section components into the same file for simplicity (up to you).

ntsim (changes requested):

src/explore-education-statistics-common/src/components/Modal.tsx
```diff
@@ -50,6 +55,18 @@ const Modal = ({
     };
   }, [fullScreen]);
 
+  const content = scrollableContent ? (
+    <div
+      className={styles.content}
+      // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+      tabIndex={0}
```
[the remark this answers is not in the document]
amyb-hiveit: The `tabIndex` here is for the scrollable area inside the modal, not the modal itself (the screenshot is of how it works currently). If we made the whole modal scrollable (is that what you're suggesting?) I don't think it would work as well as the close button wouldn't be visible until you scroll to the bottom.
modal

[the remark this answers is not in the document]
ntsim: Fair enough, we can keep the `tabIndex` if we want to always show the Close button. However, it's still be a good idea to lock the scrolling on the main window viewport as it's a bit weird to have two scrollbars. 

I wonder if it'd be possible if we could scroll the modal content when scrolling outside the modal i.e. on the overlay. Not super necessary though.

src/explore-education-statistics-common/src/components/Modal.module.scss
```diff
@@ -46,3 +46,13 @@
     outline-offset: -$govuk-focus-width;
   }
 }
+
+.content {
+  margin-bottom: govuk-spacing(5);
+  max-height: 50vh;
```
ntsim, line 52: Would it be possible to bump this up some more? Something like 70vh could be better.

src/explore-education-statistics-common/src/components/Modal.tsx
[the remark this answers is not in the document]
amyb-hiveit: Scrolling - I've actually changed my mind now after testing them out on a few more screen sizes and have changed it to scroll the whole modal instead of just the content section. 

I've started switching over to Radix but it's going to be a lot of changes so I'll do that in a follow up PR instead of this one (https://dfedigital.atlassian.net/browse/EES-4567). That'll sort the scroll locking issue as well.

src/explore-education-statistics-common/src/modules/find-statistics/components/AdHocStatisticsSection.tsx
```diff
@@ -1,8 +1,12 @@
 import React from 'react';
 
-const AdHocOfficialStatisticsSection = () => (
+const AdHocStatisticsSection = ({
```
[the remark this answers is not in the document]
amyb-hiveit: I've moved these and will do a tidy up PR for the rest later.

src/explore-education-statistics-common/src/components/Modal.module.scss
amyb-hiveit: This has been removed now.

ntsim (approved):

PR #4307, opened 2023-09-14
EES-4547 Fix snapshot generation for methodologies page

markhiveit:
This fixes an issue caused when the methodologies page was reworked to remove topics and update to use the new GOV.UK accordions.

As we've added new id tags to elements to generate the methodologies' page snapshot, the snapshots for production can be updated once this has been deployed to production.

### DIFF
--- a/src/explore-education-statistics-frontend/src/modules/methodologies/MethodologyIndexPage.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/methodologies/MethodologyIndexPage.tsx
@@ -80,7 +80,7 @@ const MethodologyIndexPage: NextPage<Props> = ({ themes = [] }) => {
                 <ul className="govuk-!-margin-top-0">
                   {orderBy(getMethodologiesForTopics(topics), 'title').map(
                     methodology => (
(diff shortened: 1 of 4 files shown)


src/explore-education-statistics-frontend/src/modules/methodologies/MethodologyIndexPage.tsx
```diff
@@ -80,7 +80,7 @@ const MethodologyIndexPage: NextPage<Props> = ({ themes = [] }) => {
                 <ul className="govuk-!-margin-top-0">
                   {orderBy(getMethodologiesForTopics(topics), 'title').map(
                     methodology => (
-                      <li key={methodology.id}>
+                      <li key={methodology.id} id="methodology-heading">
```
ntsim, line 83: Similar thing to other comment, ids should be unique. `methodology-heading` is also a weird name as well.

This really doesn't need any kind of id at all, we can just use a selector on `li` within the accordion section

ntsim (changes requested):

ntsim (approved):

PR #4309, opened 2023-09-15
EES-4426 Remove MethodologyVersion.InternalReleaseNote

markhiveit:
This PR removes the InternalReleaseNote column from the MethodologyVersion table. This work was opened up by the creation of the MethodologyStatus table in EES-4385. Now that internal notes are stored in methodology statuses, there was no reason to also store them on methodology versions.

The only change to note is adding the `GetLatestMethodologyStatus` method which is called in `BuildMethodologyVersionViewModel`. `GetLatestMethodologyStatus` gets the latest `MethodologyStatus` for the specific version passed to it, rather than getting the latest status for the entire methodology. I reviewed the surrounding code and think this is the correct choice - although I suspect if we did pass the latest status for the entire methodology we would probably get the same result regardless.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyApprovalServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyApprovalServiceTests.cs
@@ -120,7 +120,6 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.Method
                 VerifyAllMocks(contentService, methodologyVersionRepository);
 
                 Assert.Equal(methodologyVersion.Id, updatedMethodologyVersion.Id);
-                Assert.Equal("Test approval", updatedMethodologyVersion.InternalReleaseNote);
                 Assert.Null(updatedMethodologyVersion.Published);
                 Assert.Equal(Immediately, updatedMethodologyVersion.PublishingStrategy);
                 Assert.Null(updatedMethodologyVersion.ScheduledWithRelease);
@@ -139,6 +138,17 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.Method
                 Assert.Equal(Immediately, updatedMethodology.PublishingStrategy);
                 Assert.True(updatedMethodology.Updated.HasValue);
                 Assert.InRange(DateTime.UtcNow.Subtract(updatedMethodology.Updated!.Value).Milliseconds, 0, 1500);
+
+                var statusList = await context
+                    .MethodologyStatus
+                    .Where(ms => ms.MethodologyVersionId == methodologyVersion.Id)
+                    .ToListAsync();
+                var status = Assert.Single(statusList);
+                Assert.Equal(methodologyVersion.Id, status.MethodologyVersionId);
+                Assert.Equal("Test approval", status.InternalReleaseNote);
+                Assert.Equal(Approved, status.ApprovalStatus);
+                Assert.InRange(DateTime.UtcNow.Subtract(status.Created!.Value).Milliseconds, 0, 1500);
+                Assert.Equal(UserId, status.CreatedById);
             }
         }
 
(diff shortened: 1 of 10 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyApprovalServiceTests.cs
```diff
@@ -139,6 +138,17 @@ await using (var context = InMemoryApplicationDbContext(contentDbContextId))
...
+                var statusList = await context
+                    .MethodologyStatus
+                    .Where(ms => ms.MethodologyVersionId == methodologyVersion.Id)
+                    .ToListAsync();
+                var status = Assert.Single(statusList);
+                Assert.Equal(methodologyVersion.Id, status.MethodologyVersionId);
+                Assert.Equal("Test approval", status.InternalReleaseNote);
+                Assert.Equal(Approved, status.ApprovalStatus);
+                Assert.InRange(DateTime.UtcNow.Subtract(status.Created!.Value).Milliseconds, 0, 1500);
+                Assert.Equal(UserId, status.CreatedById);
```
benoutram, line 151: I would say these lines are not necessary because this is a variant of the unit test specific to methodology images and the status is already covered by `UpdateApprovalStatus_ApprovingMethodologyAddsMethodologyStatus`. This is just duplication of that.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyApprovalServiceTests.cs
```diff
@@ -256,6 +265,17 @@ await using (var context = InMemoryApplicationDbContext(contentDbContextId))
...
```
benoutram, line 278: Same comment about this being unecessary as it's already covered by `UpdateApprovalStatus_ApprovingMethodologyAddsMethodologyStatus`.

I would go through all the other variations of this test and determine whether checking the status is adding any beneft. I can see the benefit of checking in `UpdateApprovalStatus_NoStatusUpdateRequiredNoMethodologyStatusAdded` but not sure about the rest.

If they are kept, I think:

```
var status = Assert.Single(context.MethodologyStatus);
```

or this:

```
var status = await context
      .MethodologyStatus
      .Where(ms => ms.MethodologyVersionId == methodologyVersion.Id)
      .Single();
```

could be a nicer alternative to this:

```
var statusList = await context
    .MethodologyStatus
    .Where(ms => ms.MethodologyVersionId == methodologyVersion.Id)
    .ToListAsync();
var status = Assert.Single(statusList);
```

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
```diff
@@ -807,13 +812,20 @@ public async Task GetMethodology()
                 AlternativeTitle = "Alternative title"
             };
 
+            var methodologyStatus = new MethodologyStatus
+            {
+                MethodologyVersion = methodologyVersion,
+                InternalReleaseNote = "Test approval",
+            };
```
benoutram, line 819: Missing `ApprovalStatus` here?

markhiveit: I've added it, but I don't think this is strictly necessary as it gets the methodology version's status from the `MethodologyVersion` object, not from the `MethodologyStatus`. The `MethodologyStatus.ApprovalStatus` isn't unused by `GetMethodology`.

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
markhiveit: > I see that now that we have MethodologyStatus table the MethodologyVersion.Status can differ

I did have the thought that given that we're removing `InternalReleaseNote` from `MethodologyVersion` because it's now stored in `MethodologyStatus`, you could make an argument to do the same for approval status. 

With that said, I'm not personally of the opinion that we should remove `MethodologyVersion.Status` - I think the additional confusion wouldn't make it worth it.

PR #4314, opened 2023-09-20
EES-4444 Fix UI test failures and assert page loading state using network requests

ntsim:
This PR deals primarily deals with UI test failures revolving around pages not being ready and network requests not having completed.

## Page loading

In some cases, we're not waiting sufficiently long enough for the page to complete loading. This is leading to the DOM not being ready to actually receive things like click events. This can lead to the intermittent failures where a link was clicked, but no navigation occurred.

Unfortunately, there isn't really any good way to figure out when React has fully prepared the DOM and we just have to rely on a combination of implicit and explicit waits.

## Network requests

In a number of cases, network requests will be initiated by an action (e.g. clicking some button), but the tests won't wait for these to finish before moving on to validate that the action has completed.

We have tried to rectify this by implementing additional hooks on `Client` so that we now monitor network activity:

- When there are any network requests in progress, we add a `data-network-activity="active"` state to the `body`
- When the network goes idle, we toggle this state to `data-network-activity="idle"`

We can then watch for this using Robot Framework to wait for inflight requests to complete.

**Note** - This only hooks into requests we make using `Client` and not raw XHR, fetch or any other type of HTTP client.

## Changes to loading keywords

- Renamed `user waits for page to finish loading` to the more consistent `user waits until page finishes loading`. This now handles waiting for the network to become idle.
- Removed the `user waits for page to not contain loading spinner` in favour of `user waits until page finishes loading`. The loading spinner keyword is redundant as we're only ever really interested in the loading state of the entire page (not just a spinner).

## Other changes

- Added `composeProviders` HOC that allows us to flatten down the 'provider tree of doom' in our application entrypoints.
- Cleaned up various UI tests to use `user clicks button/link` keywords
- Updated `user clicks button` keywords to search for text in nested elements

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -874,6 +874,9 @@ importers:
       typescript:
         specifier: ^5.1.6
         version: 5.1.6
+      xhr-mock:
+        specifier: ^2.5.1
+        version: 2.5.1
 
   src/explore-education-statistics-frontend:
     dependencies:
@@ -1944,8 +1947,8 @@ packages:
       '@babel/core': ^7.0.0-0
     dependencies:
       '@babel/core': 7.22.8
-      '@babel/helper-environment-visitor': 7.22.5
-      '@babel/helper-plugin-utils': 7.22.5
+      '@babel/helper-environment-visitor': 7.18.9
+      '@babel/helper-plugin-utils': 7.20.2
       '@babel/helper-remap-async-to-generator': 7.18.9(@babel/core@7.22.8)
       '@babel/plugin-syntax-async-generators': 7.8.4(@babel/core@7.22.8)
     transitivePeerDependencies:
@@ -1959,8 +1962,8 @@ packages:
       '@babel/core': ^7.0.0-0
     dependencies:
       '@babel/core': 7.9.0
-      '@babel/helper-environment-visitor': 7.22.5
-      '@babel/helper-plugin-utils': 7.22.5
+      '@babel/helper-environment-visitor': 7.18.9
+      '@babel/helper-plugin-utils': 7.20.2
       '@babel/helper-remap-async-to-generator': 7.18.9(@babel/core@7.9.0)
       '@babel/plugin-syntax-async-generators': 7.8.4(@babel/core@7.9.0)
     transitivePeerDependencies:
@@ -3068,13 +3071,13 @@ packages:
     dependencies:
       '@babel/core': 7.22.8
       '@babel/helper-annotate-as-pure': 7.18.6
-      '@babel/helper-compilation-targets': 7.22.6(@babel/core@7.22.8)
-      '@babel/helper-environment-visitor': 7.22.5
-      '@babel/helper-function-name': 7.22.5
+      '@babel/helper-compilation-targets': 7.20.7(@babel/core@7.22.8)
+      '@babel/helper-environment-visitor': 7.18.9
+      '@babel/helper-function-name': 7.19.0
       '@babel/helper-optimise-call-expression': 7.18.6
-      '@babel/helper-plugin-utils': 7.22.5
+      '@babel/helper-plugin-utils': 7.20.2
       '@babel/helper-replace-supers': 7.20.7
-      '@babel/helper-split-export-declaration': 7.22.6
+      '@babel/helper-split-export-declaration': 7.18.6
       globals: 11.12.0
     transitivePeerDependencies:
       - supports-color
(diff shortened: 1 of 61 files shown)


duncan-at-hiveit:
I did get one failure while running this branch locally.

Seems to me a strange 3 steps that we do when saving the Pre-release access list in 2 places:

```
    user enters text into element    id:publicPreReleaseAccessForm-preReleaseAccessList    ${content}
    user clicks button    Save access list
    user waits until element contains    id:publicPreReleaseAccessForm-preReleaseAccessList    ${content}
```

whereas in another we do:

```
    user clicks button    Save access list
    user waits until element contains    css:[data-testid="publicPreReleaseAccessListPreview"]
    ...    Test public access list
```

which seems to make more sense, in that the former example is checking to make sure a form that was on the page prior to clicking "Save access list" is present again (presumably after a save has completed), whereas the latter example is waiting for its readonly preview mode to be present, which feels like the right thing to do.

I've just changed `user creates public prerelease access list` and `user updates public prerelease access list` to follow this latter approach and will see how it goes. 

Here's the failure by the way:

image

src/explore-education-statistics-common/src/contexts/NetworkActivityContext.tsx
```diff
@@ -0,0 +1,176 @@
+import useMountedRef from '@common/hooks/useMountedRef';
```
duncan-at-hiveit, line 1: Very cool

tests/robot-tests/tests/libs/admin-common.robot
```diff
@@ -73,9 +73,9 @@ user selects dashboard theme and topic if possible
     [Arguments]
```
duncan-at-hiveit, line 73: Would also suggest changing the final lines of `user creates public prerelease access list` and `user updates public prerelease access list` to:

```
    user clicks button    Save access list
    user waits until element contains    css:[data-testid="publicPreReleaseAccessListPreview"]
    ...    ${content}
```

as the previous waiting after the button click was but this seems to be correct.

src/explore-education-statistics-common/src/contexts/NetworkActivityContext.tsx
```diff
@@ -0,0 +1,176 @@
...
+const NetworkActivityContext = createContext<NetworkActivityState | undefined>(
+  undefined,
+);
+
+export interface NetworkActivityContextProviderProps {
+  children?: ReactNode;
+  idleTimeout?: number;
+}
+
+export function NetworkActivityContextProvider({
```
amyb-hiveit, line 40: It might be useful to have a comment here explaining why we're doing this as it's not immediately obvious why you'd want to have a data attribute for network activity.

tests/robot-tests/tests/admin/analyst/role_ui_permissions/release_and_publication_role_end_to_end_permissions.robot
```diff
@@ -41,7 +41,7 @@ Check publication owner can update methodology summary
 
 Check publication owner cannot approve methodology for publication
     user clicks link    Sign off
-    user waits until page does not contain loading spinner
+    user waits until page finishes loading
```
duncan-at-hiveit, line 44: Makes me wonder if moving forwards with this, if we should building this in as a compulsory step before any step, or before certain subsets of our steps.

Or flip it and include it *after* any of our steps that expect the page to update based upon an action e.g. all "user clicks link/button", "user goes to url", etc etc.

Not suggesting we do this as part of this PR, but I'd say that I'm seeing this step appearing very regularly right after certain other steps have been called.

To start this, we could look at everywhere where we have used this step `user waits until page finishes loading` and see what the preceding steps are.  If we see enough repetition with certain steps, we can build it into them (and maybe provide an optional way to suppress waiting for the page to finish loading if we really don't want to wait).

We might be concerned that we could pay a performance penalty for this, but I'd say that that's easily tested with a before-and-after set of timings.

tests/robot-tests/tests/libs/admin-common.robot
duncan-at-hiveit: I've actually included this in the 4448 PR as I was having a lot of failures around it, so no need to change it here.

pnpm-lock.yaml
```diff
@@ -1944,8 +1947,8 @@ packages:
       '@babel/core': ^7.0.0-0
     dependencies:
       '@babel/core': 7.22.8
-      '@babel/helper-environment-visitor': 7.22.5
```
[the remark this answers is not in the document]
duncan-at-hiveit: Cool just checking thanks!

tests/robot-tests/tests/admin/analyst/role_ui_permissions/release_and_publication_role_end_to_end_permissions.robot
duncan-at-hiveit: Cool no worries, thanks for checking it out.

duncan-at-hiveit (approved):

PR #4315, opened 2023-09-20
EES-4557 Show all methodologies associated with release on Admin Cont…

markhiveit:
This PR ensures that the correct methodologies are displayed on Admin prerelease pages and on Admin release content pages.

This fixes the bug introduced by EES-4376.

The rules for displaying methodologies are now:
- On prerelease pages, only methodologies with a latest published version be displayed.
- On admin content pages, a methodology with any version will be displayed.

Other problems surrounding this will be dealt with in EES-4565 and EES-4566. For example, this now allows prerelease users to potentially see draft methodologies by utilising the `GetManageContentPageData` endpoint. They were seeing unpublished methodologies up until recently anyway, and this is next in lined to be fixed.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ManageContent/ManageContentPageServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ManageContent/ManageContentPageServiceTests.cs
@@ -148,19 +148,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.Manage
                 {
                     Id = Guid.NewGuid(),
                     AlternativeTitle = "Methodology 2 title",
-                    Status = MethodologyApprovalStatus.Approved,
-                },
-                new MethodologyVersion
-                {
-                    Id = Guid.NewGuid(),
-                    AlternativeTitle = "Methodology should not appear 1",
                     Status = MethodologyApprovalStatus.Draft,
-                },
-                new MethodologyVersion
-                {
-                    Id = Guid.NewGuid(),
-                    AlternativeTitle = "Methodology should not appear 2",
-                    Status = MethodologyApprovalStatus.HigherLevelReview,
                 }
             );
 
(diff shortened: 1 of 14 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/ManageContentPageService.cs
```diff
@@ -62,17 +62,22 @@ public class ManageContentPageService : IManageContentPageService
...
+                    {
+                        methodologyVersions = await _methodologyVersionRepository
+                            .GetLatestVersionByPublication(release.PublicationId);
+                    }
 
                     var releaseViewModel = _mapper.Map<ManageContentPageViewModel.ReleaseViewModel>(release);
                     releaseViewModel.DownloadFiles = files.ToList();
                     releaseViewModel.Publication.Methodologies =
-                        _mapper.Map<List<IdTitleViewModel>>(approvedMethodologyVersions);
+                        _mapper.Map<List<IdTitleViewModel>>(methodologyVersions);
```
benoutram, line 80: If the version id set in the view model is important here, e.g. because pre-release users can click through to the content of it, then I think they are only ever going to see the content of the latest published version, which would exclude the  version of a methodology which is about to be published with this release.

I think pre-release users should see the methodologies list accurate to what it will be when the release is published. Might be one to check with Cam, on how this should work.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ManageContent/ManageContentPageServiceTests.cs
```diff
@@ -367,6 +355,146 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+                    dataBlockService: dataBlockService.Object,
+                    methodologyVersionRepository: methodologyVersionRepository.Object,
+                    releaseFileService: releaseFileService.Object);
+
+                var result = await service.GetManageContentPageViewModel(
+                    release.Id, isPrerelease: true);
+
+                Assert.True(result.IsRight);
+
+                var contentRelease = result.Right.Release;
```
benoutram, line 482: You can use `var contentRelease = result.AssertRight()` here.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ManageContent/ManageContentPageServiceTests.cs
```diff
@@ -367,6 +355,146 @@ await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContex
...
+                Publication = publication,
+                PublishScheduled = DateTime.Parse("2020-09-08T23:00:00.00Z", styles: DateTimeStyles.AdjustToUniversal),
+                Published = null,
+                ReleaseName = "2020",
+                RelatedInformation = new List<Link>(),
+                Slug = "2020-21",
+                TimePeriodCoverage = AcademicYear,
+                Type = ReleaseType.OfficialStatistics,
+                Updates = new List<Update>(),
+                KeyStatistics = new List<KeyStatistic>(),
```
benoutram, line 387: Some of the fields setup on this test can probably be slimmed down if they aren't relevant to the test.

Here's an example of two which are already initialised by default in `Release`.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ManageContent/ManageContentPageServiceTests.cs
markhiveit: I've removed all other `.Right`s from this test suite.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/ManageContentPageService.cs
markhiveit: Discussed this with Ben and Cam. So now this is the logic that determines what methodologies are displayed around Admin:
- On the Admin content pages: the latest methodology version is displayed, regardless of status.
- On Prerelease content and prerelease methodologies tab: the latest methodology version is displayed if it is approved, but if it is not approved, the previous version is displayed.

This should ensure that any methodology set to go out with a release will be displayed in prerelease. 

It's possible that a methodology set to be released with a different release may also appear on prerelease, but I think the time investment to ensure that is handled correctly, alongside the likelyhood of that scenario happening, makes that not worth it - especially as we have been displaying draft methodologies on prerelease for a long time and nobody complained..

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Methodologies/MethodologyController.cs
```diff
@@ -78,10 +78,10 @@ public async Task<ActionResult<MethodologyVersionViewModel>> GetMethodology(Guid
         }
 
         [HttpGet("publication/{publicationId:guid}/methodologies")]
-        public async Task<ActionResult<List<MethodologyVersionSummaryViewModel>>> ListLatestMethodologyVersions(Guid publicationId)
+        public async Task<ActionResult<List<MethodologyVersionSummaryViewModel>>> ListLatestApprovedMethodologyVersions(Guid publicationId)
         {
             return await _methodologyService
-                .ListLatestMethodologyVersions(publicationId)
+                .ListLatestApprovedMethodologyVersions(publicationId)
                 .HandleFailuresOrOk();
```
benoutram, line 85: Hang on, won't this introduce a new bug where an analyst that should be able to see draft methodologies on the 'Manage methodologies' tab won't see them now?

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -151,38 +151,59 @@ await _methodologyVersionRepository.GetLatestPublishedVersion(methodology.Id))
...
+                                if (methodologyVersion.PreviousVersion != null)
+                                {
+                                    // If there is a previous version, it must be approved, because cannot
+                                    // create an amendment for an unpublished version
+                                    methodologyVersion = methodologyVersion.PreviousVersion;
+                                }
+                                else
+                                {
+                                    return null;
+                                }
```
benoutram, line 185: Inverting this would be shorter and easier to read in my opinion

```
  // Get latest approved versions
  if (methodologyVersion.PreviousVersion == null)
  {
    return null;
  }

  // Use the previous version which must be approved, because cannot
  // create an amendment for an unpublished version
  methodologyVersion = methodologyVersion.PreviousVersion
```

benoutram (approved):

benoutram (approved):

PR #4317, opened 2023-09-20
EES-4448 - responding to PR comments regarding improvements to integration tests

duncan-at-hiveit:
This PR:
- disables additional Startup configuration (Swagger, IdentityServer, Application Insights) during integration tests, and swaps out the use of flags in favour of using an "IntegrationTests" Host Environment instead.  Extension methods have been added to keep environment checking consistent with OOTB `IsDevelopment()`, `IsProduction()` etc.
- reduces the configuration necessary for running Integration Tests to simply whatever is included in a project's `appsettings.IntegrationTests.json` file.
- removes manual instantiation of TableStorageService instances in various Startup classes in favour of looking them up from the service collection.  We now follow the pattern as laid out with PrivateBlobStorageService / PublicBlobStorageService, pushing the lookup of connection strings into the implementations themselves and further reducing the amount of configuration necessary to start the integration tests.
- simplifies a bunch of service wiring in Startup.cs that originally would have been forced to be longhand due to needing to target particular table storage connection strings (e.g. ReleasePublishingStatusRepository).
- simplifies a bunch of unrelated service wiring in Startup.cs that were unnecessarily being performed manually e.g. PublicationService.  There seemed to be no reason to wire them up manually in the first place, and so I suspect these were left over cases from old piece of work and were just missed.


ntsim (approved):

src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/HostEnvironmentExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Hosting;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+
+public static class HostEnvironmentExtensions
+{
+    public const string IntegrationTestEnvironment = "IntegrationTests";
+
+    public static bool IsIntegrationTests(this IHostEnvironment? hostEnvironment) =>
```
ntsim, line 12: Not a biggy, but would it be better to make this singular i.e. `IsIntegrationTest` and `IntegrationTest`?

duncan-at-hiveit: No probs, I've updated that now ta

PR #4320, opened 2023-09-22
EES-4477 Make contact telephone number optional

markhiveit:
This PR makes a Contact telephone number optional. It also adds a migration to remove the DfE Enquiries telephone number from existing contacts, and frontend validation to prevent the DfE Enquiries number being added as a contact tel no in the future.

:rotating_light: NOTE: After the migration, to update the caches, we need to hit the `DELETE admin/api/bau/public-cache/publications` endpoint. :rotating_light: 

### Notes

- I checked the Contact tel numbers currently in the Prod DB, and I wrote my migration to cover all cases, as the number was spaced differently in different rows.
- The frontend validation regex is designed to ensure nobody can get around the DfE Enquiries number validation by using spacing.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Requests/ContactSaveRequestTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Requests/ContactSaveRequestTests.cs
@@ -3,94 +3,57 @@ using System.ComponentModel.DataAnnotations;
 using GovUk.Education.ExploreEducationStatistics.Admin.Requests;
 using Xunit;
 
-namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Requests
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Requests;
+
+public class ContactSaveRequestTests
 {
-    public class ContactSaveRequestTests
+    [Fact]
+    public void ContactSaveRequest_ContactTelNo()
     {
-        [Fact]
-        public void ContactSaveRequest_ContactTelNo()
+        var request = new ContactSaveRequest
         {
-            var request = new ContactSaveRequest
-            {
-                ContactTelNo = "12345678",
-            };
+            ContactTelNo = "01234567",
+        };
 
-            var validationContext = new ValidationContext(request);
+        var validationContext = new ValidationContext(request);
 
-            var results = request.Validate(validationContext);
+        var results = request.Validate(validationContext);
 
-            Assert.Empty(results);
-        }
+        Assert.Empty(results);
+    }
 
-        [Fact]
-        public void ContactSaveRequest_ContactTelNo_Null()
+    [Fact]
+    public void ContactSaveRequest_ContactTelNo_Null()
+    {
+        var request = new ContactSaveRequest
         {
-            var request = new ContactSaveRequest
-            {
-                ContactTelNo = null,
-            };
+            ContactTelNo = null,
+        };
 
-            var validationContext = new ValidationContext(request);
+        var validationContext = new ValidationContext(request);
 
-            var results = request.Validate(validationContext);
+        var results = request.Validate(validationContext);
 
-            Assert.Empty(results);
-        }
+        Assert.Empty(results);
+    }
 
-        [Fact]
-        public void ContactSaveRequest_ContactTelNo_TooShort()
+    [Theory]
+    [InlineData(" 0 3 7 0 0 0 0 2 2 8 8 ")]
+    [InlineData("03700002288")]
+    [InlineData("    0370 000 2288")]
+    [InlineData("037 0000 2288    ")]
+    public void ContactSaveRequest_ContactTelNo_NotDfEEnquiriesNum(string contactTelNo)
+    {
+        var request = new ContactSaveRequest
         {
-            var request = new ContactSaveRequest
-            {
-                ContactTelNo = "1234567",
-            };
+            ContactTelNo = contactTelNo,
+        };
 
-            var validationContext = new ValidationContext(request);
+        var validationContext = new ValidationContext(request);
 
-            var results = request.Validate(validationContext);
+        var results = request.Validate(validationContext);
 
-            var validationResult = Assert.Single(results);
-            Assert.Equal("Contact telephone must have a minimum length of '8'.", validationResult.ErrorMessage);
-        }
-
-        [Theory]
-        [InlineData("abcdefghijk")]
-        [InlineData(@"!&$\*^!\*&)%!")]
-        [InlineData("a12345678")]
-        [InlineData("12345678a")]
-        public void ContactSaveRequest_ContactTelNo_NotNumeric(string contactTelNo)
-        {
-            var request = new ContactSaveRequest
-            {
-                ContactTelNo = contactTelNo,
-            };
-
-            var validationContext = new ValidationContext(request);
-
-            var results = request.Validate(validationContext);
-
-            var validationResult = Assert.Single(results);
-            Assert.Equal("Contact telephone must only contain numbers and spaces", validationResult.ErrorMessage);
-        }
-
-        [Theory]
-        [InlineData(" 0 3 7 0 0 0 0 2 2 8 8 ")]
-        [InlineData("03700002288")]
-        [InlineData("    0370 000 2288")]
-        [InlineData("037 0000 2288    ")]
-        public void ContactSaveRequest_ContactTelNo_NotDfEEnquiriesNum(string contactTelNo)
-        {
-            var request = new ContactSaveRequest
-            {
-                ContactTelNo = contactTelNo,
-            };
-
-            var validationContext = new ValidationContext(request);
-
-            var results = request.Validate(validationContext);
-
-            var validationResult = Assert.Single(results);
-            Assert.Equal("Contact telephone cannot be DfE Enquiries number", validationResult.ErrorMessage);
-        }
+        var validationResult = Assert.Single(results);
+        Assert.Equal("Contact telephone cannot be DfE Enquiries number", validationResult.ErrorMessage);
     }
 }
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Requests/ContactSaveRequestTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Requests/ContactSaveRequestTests.cs
@@ -1,0 +1,96 @@
+#nullable enable
+using System.ComponentModel.DataAnnotations;
+using GovUk.Education.ExploreEducationStatistics.Admin.Requests;
+using Xunit;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Requests
+{
+    public class ContactSaveRequestTests
+    {
+        [Fact]
+        public void ContactSaveRequest_ContactTelNo()
+        {
+            var request = new ContactSaveRequest
+            {
+                ContactTelNo = "12345678",
+            };
+
+            var validationContext = new ValidationContext(request);
+
+            var results = request.Validate(validationContext);
+
+            Assert.Empty(results);
+        }
+
+        [Fact]
+        public void ContactSaveRequest_ContactTelNo_Null()
+        {
+            var request = new ContactSaveRequest
+            {
+                ContactTelNo = null,
+            };
+
+            var validationContext = new ValidationContext(request);
+
+            var results = request.Validate(validationContext);
+
+            Assert.Empty(results);
+        }
+
+        [Fact]
+        public void ContactSaveRequest_ContactTelNo_TooShort()
+        {
+            var request = new ContactSaveRequest
+            {
+                ContactTelNo = "1234567",
+            };
+
+            var validationContext = new ValidationContext(request);
+
+            var results = request.Validate(validationContext);
+
+            var validationResult = Assert.Single(results);
+            Assert.Equal("Contact telephone must have a minimum length of '8'.", validationResult.ErrorMessage);
+        }
+
+        [Theory]
+        [InlineData("abcdefghijk")]
+        [InlineData(@"!&$\*^!\*&)%!")]
+        [InlineData("a12345678")]
+        [InlineData("12345678a")]
+        public void ContactSaveRequest_ContactTelNo_NotNumeric(string contactTelNo)
+        {
+            var request = new ContactSaveRequest
+            {
+                ContactTelNo = contactTelNo,
+            };
+
+            var validationContext = new ValidationContext(request);
+
+            var results = request.Validate(validationContext);
+
+            var validationResult = Assert.Single(results);
+            Assert.Equal("Contact telephone must only contain numbers and spaces", validationResult.ErrorMessage);
+        }
+
+        [Theory]
+        [InlineData(" 0 3 7 0 0 0 0 2 2 8 8 ")]
+        [InlineData("03700002288")]
+        [InlineData("    0370 000 2288")]
+        [InlineData("037 0000 2288    ")]
+        public void ContactSaveRequest_ContactTelNo_NotDfEEnquiriesNum(string contactTelNo)
+        {
+            var request = new ContactSaveRequest
+            {
+                ContactTelNo = contactTelNo,
+            };
+
+            var validationContext = new ValidationContext(request);
+
+            var results = request.Validate(validationContext);
+
+            var validationResult = Assert.Single(results);
+            Assert.Equal("Contact telephone cannot be DfE Enquiries number", validationResult.ErrorMessage);
+        }
+    }
+}
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
@@ -26,7 +26,8 @@ using static GovUk.Education.ExploreEducationStatistics.Admin.Validators.Validat
 using static GovUk.Education.ExploreEducationStatistics.Common.Services.CollectionUtils;
 using static GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils.MockUtils;
 using static Moq.MockBehavior;
-using IPublicationRepository = GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.IPublicationRepository;
+using IPublicationRepository =
+    GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.IPublicationRepository;
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
 {
@@ -623,7 +624,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                 var publicationService = BuildPublicationService(context,
                     userService: userService.Object);
 
-                var result = (await publicationService.GetPublication(publication.Id, includePermissions: true)).AssertRight();
+                var result = (await publicationService.GetPublication(publication.Id, includePermissions: true))
+                    .AssertRight();
 
                 Assert.Equal(publication.Id, result.Id);
 
@@ -707,7 +709,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
             {
                 var publicationService = BuildPublicationService(context);
 
-                var result = (await publicationService.GetPublication(publication.Id, includePermissions: true)).AssertRight();
+                var result = (await publicationService.GetPublication(publication.Id, includePermissions: true))
+                    .AssertRight();
 
                 Assert.Equal(publication.Id, result.Id);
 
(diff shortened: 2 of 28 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Model/Contact.cs
```diff
@@ -1,23 +1,18 @@
...
-        [Required]
-        public string ContactName { get; set; }
+    [Required] public string ContactName { get; set; } = null!;
 
-        [Required]
-        public string ContactTelNo { get; set; }
-    }
-}
\ No newline at end of file
+    public string? ContactTelNo { get; set; }
```
benoutram, line 17: My first thought is that none of these DataAnnotations should be required, since EF will recognise the `Id` field is the key and nullability of the fields is handled automatically too now that you've added the nullable context to this class.

I see that there's one request `PublicationController.UpdateContact` using this type in the request body.

I think I would look at making that use the `ContactSaveViewModel` and then strip all the annotations from this class.

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20230920145213_EES4477_MakeContactTelNosNullable.cs
```diff
@@ -0,0 +1,40 @@
...
+                nullable: true,
+                oldClrType: typeof(string),
+                oldType: "nvarchar(max)");
+
+            migrationBuilder.Sql(
+                @"UPDATE [dbo].[Contacts]
+                  SET ContactTelNo = NULL
+                  WHERE TRIM(ContactTelNo) = '03700002288'
+                  OR TRIM(ContactTelNo) = '0370 000 2288'
+                  OR TRIM(ContactTelNo) = '037 0000 2288';");
```
benoutram, line 24: Was thinking `TRIM` is probably not necessary here, as strings are padded when compared for equality (try out `SELECT 'Equal!' WHERE '03700002288    ' = '03700002288'` but then I realised that it's only the case for trailing space, not leading space so you are probably right to include the trim here.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
```diff
@@ -855,6 +855,60 @@ await using (var context = InMemoryApplicationDbContext(contextId))
...
+                var publicationViewModel = result.AssertRight();
+                Assert.Equal("Test publication", publicationViewModel.Title);
+
+                Assert.Equal("John Smith", publicationViewModel.Contact.ContactName);
+                Assert.Null(publicationViewModel.Contact.ContactTelNo);
+
+                var createdPublication = await context.Publications.FindAsync(publicationViewModel.Id);
+                Assert.NotNull(createdPublication);
+                Assert.Equal("John Smith", createdPublication.Contact.ContactName);
+                Assert.Null(createdPublication.Contact.ContactTelNo);
```
benoutram, line 908: I think your missing an `Include` on `Contact` here. I would also use `First` not `FindAsync` because then the test will fail due to the exception if it's not found, avoiding the extra `NotNull` assertion, and I'm wary of using `FindAsync` because it uses cached results.

markhiveit: Interesting - I wonder if it was because `FindAsync` is cached that it passed despite the missing `Include`.

I've made the same change in the test I copied this from.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Contact.cs
markhiveit: I've removed the attributes from `Contact`, but I've switched to use `ContactViewModel`. I don't think we need the attributes on the class when returning the view model. And `ContactSaveViewModel` seems to represent a request, so I've also renamed and moved `ContactSaveViewModel` to be `ContactSaveRequest`.

markhiveit: I also created a dummy migration to make sure this hasn't changed anything, and it was blank, so all looks good.

src/explore-education-statistics-admin/src/pages/publication/components/PublicationContactForm.tsx
```diff
@@ -36,16 +36,17 @@ const PublicationContactForm = ({
...
-          return false;
-        }
-        return true;
-      },
-    }),
+    contactTelNo: Yup.string()
+      .trim()
+      .matches(
+        /^[0-9 \t]*$/,
+        'The telephone number should only contain numeric characters',
```
ntsim, line 43: This message is inconsistent with other validation messages and the recommended message style. It doesn't directly reference the field as labelled and doesn't mention that whitespace is allowed either.

`Contact telephone must only contain numeric or whitespace characters` might be better.

src/explore-education-statistics-admin/src/pages/publication/components/PublicationContactForm.tsx
```diff
@@ -36,16 +36,17 @@ const PublicationContactForm = ({
...
-        if (!value) {
-          return false;
-        }
-        return true;
-      },
-    }),
+    contactTelNo: Yup.string()
+      .trim()
+      .matches(
+        /^[0-9 \t]*$/,
```
ntsim, line 42: It'd be more suitable to use `\s` as covers any type of whitespace (and is more succinct). ` \t` only covers space and tabs.

src/explore-education-statistics-admin/src/pages/publication/components/PublicationContactForm.tsx
```diff
@@ -36,16 +36,17 @@ const PublicationContactForm = ({
...
+        'The telephone number should only contain numeric characters',
+      )
+      .matches(
+        /^(?!^0[ \t]*3[ \t]*7[ \t]*0[ \t]*0[ \t]*0[ \t]*0[ \t]*2[ \t]*2[ \t]*8[ \t]*8$)/,
```
ntsim, line 46: Same as above, `\s` would be better than ` \t` as it covers any whitespace.

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/ContactSaveRequest.cs
benoutram, line 1: Missing `#nullable enable` on `ContactSaveRequest` after latest PR changes.

src/explore-education-statistics-admin/src/pages/publication/components/PublicationForm.tsx
```diff
@@ -186,7 +196,7 @@ const PublicationForm = ({
 
               <FormFieldTextInput<FormValues>
                 name="contactTelNo"
-                label="Contact telephone number"
+                label="Contact telephone number (optional)"
```
ntsim, line 199: Just noticed this label is actually inconsistent with `PublicationContactForm`. Think we should make this consistent.

Would suggest `Contact telephone` for brevity.

markhiveit:
Added some UI test fixes

src/explore-education-statistics-admin/src/pages/publication/components/PublicationContactForm.tsx
```diff
@@ -36,16 +36,17 @@ const PublicationContactForm = ({
...
+        'The DfE enquiries number is not suitable for use on statistics publications',
```
ntsim, line 47: Similar to above, this style of message is not consistent with other validation messages.

`Contact telephone cannot be the DfE enquiries number` would probably be better.

src/explore-education-statistics-admin/src/pages/publication/components/PublicationContactForm.tsx
```diff
@@ -36,16 +36,17 @@ const PublicationContactForm = ({
...
+      )
+      .min(8, 'The telephone number must be at least 8 characters long'),
```
ntsim, line 49: Similar to above - `Contact phone must be 8 characters or more` would probably be better

src/explore-education-statistics-admin/src/pages/publication/components/PublicationForm.tsx
```diff
@@ -96,7 +96,17 @@ const PublicationForm = ({
...
+        .trim()
+        .matches(
+          /^[0-9 \t]*$/,
+          'The telephone number should only contain numeric characters',
+        )
+        .matches(
+          /^(?!^0[ \t]*3[ \t]*7[ \t]*0[ \t]*0[ \t]*0[ \t]*0[ \t]*2[ \t]*2[ \t]*8[ \t]*8$)/,
+          'The DfE enquiries number is not suitable for use on statistics publications',
+        )
+        .min(8, 'The telephone number must be at least 8 characters long'),
```
ntsim, line 109: Validation changes to `PublicaitonContactForm` should also be applied here

src/explore-education-statistics-admin/src/pages/publication/__tests__/PublicationContactPage.test.tsx
```diff
@@ -173,10 +173,42 @@ describe('PublicationContactPage', () => {
         }),
       ).toBeInTheDocument();
 
+      // NOTE: Contact telephone number is optional, so no validation
+    });
+  });
+
+  test('show validation error when contact tel no is not valid', async () => {
```
ntsim, line 180: It would be a good idea to create more test cases for various permutations. This can be achieved succinctly by using `test.each` to provide test data e.g.

```ts
test.each([
  [' 03700002288 '],
  ['03700002288'],
  ['0370 000 2288'],
  [' 0 3 7 0 0 0 0 2 2 8 8 '],
])(
  'show validation error when contact tel no (%s) is DfE enquiries number',
  async telNo => {
    // the test case...
  }
);
```

Would suggest we also create test cases for:

- `show validation error when contact tel no is less than 8 characters`
- `show validation error when contact tel no is not only numeric or whitespace characters` (use `test.each` for multiple permutations of this)

src/explore-education-statistics-admin/src/pages/publication/components/__tests__/PublicationForm.test.tsx
```diff
@@ -148,10 +150,33 @@ describe('PublicationForm', () => {
         }),
       ).toBeInTheDocument();
 
+      // NOTE: Contact telelphone number is optional, so no validation error
+    });
+  });
+
+  test('show validation error when contact tel no is not valid', async () => {
```
ntsim, line 157: Same as `PublicationContactForm`, we should create more test cases to cover the various validation permutations.

benoutram (approved):
Reviewed backend changes only. I think @ntsim is also going to take a look at this PR.

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
```diff
@@ -836,7 +836,10 @@ await using (var context = InMemoryApplicationDbContext(contextId))
                 Assert.Equal(topic.Theme.Title, publicationViewModel.Theme.Title);
 
                 // Do an in depth check of the saved release
-                var createdPublication = await context.Publications.FindAsync(publicationViewModel.Id);
+                var createdPublication = await context.Publications
+                    .Include(p => p.Contact)
+                    .Include(p => p.Topic)
+                    .FirstAsync(p => p.Id == publicationViewModel.Id);
```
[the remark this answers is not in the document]
markhiveit: Fixed - and I've added new `await using (var context = InMemoryApplicationDbContext(contextId))` blocks for the same scenario across all test cases in this file.

I had to add a couple of `.Include(p => p.Topic)` lines elsewhere, so we were being hit by caching in test cases across this file. But the tests otherwise passed so all looks fine.

I also took this opportunity to remove any other `FindAsync` calls in the file.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -115,7 +115,9 @@ public class PublicationService : IPublicationService
                     var contact = await _context.Contacts.AddAsync(new Contact
                     {
                         ContactName = publication.Contact.ContactName,
-                        ContactTelNo = publication.Contact.ContactTelNo,
+                        ContactTelNo = String.IsNullOrWhiteSpace(publication.Contact.ContactTelNo)
```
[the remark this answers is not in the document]
markhiveit: Also made the same change in multiple places across the codebase.

src/explore-education-statistics-admin/src/pages/publication/components/PublicationContactForm.tsx
markhiveit: I've made the change. I ran this by Cam because the original I used was from the ticket's acceptance criteria.

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/ContactSaveRequest.cs
```diff
@@ -0,0 +1,14 @@
...
+
+public class ContactSaveRequest
+{
+    [Required] public string TeamName { get; set; } = string.Empty;
+
+    [Required, EmailAddress] public string TeamEmail { get; set; } = string.Empty;
+
+    [Required] public string ContactName { get; set; } = string.Empty;
+
+    public string? ContactTelNo { get; set; }
```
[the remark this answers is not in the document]
ntsim: You mean frontend-only validation right? 

Whilst it's not super likely to see anyone trying to circumvent this, it's not really good practice to skip server-side valiation. 

It's fairly straightforward to create custom validations so I'd suggest we don't skip our belts and braces. 

We should at least create a new `PhoneNumber` validation attribute, and add the built-in `MinLength` validation. I'd probably say we don't desperately need the validation for the DfE phone number, but up to you.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseChecklistService.cs
```diff
@@ -159,7 +160,7 @@ private async Task<bool> ReleaseSectionHasEmptyHtmlBlock(Guid releaseId, Content
                     cb.ContentSection!.Release.ReleaseId == releaseId &&
                     cb.ContentSection.Type == sectionType)
                 .OfType<HtmlBlock>()
-                .AnyAsync(htmlBlock => string.IsNullOrEmpty(htmlBlock.Body));
+                .AnyAsync(htmlBlock => htmlBlock.Body.IsNullOrEmpty());
```
benoutram, line 163: `StringExtensions.IsNullOrEmpty` can't be translated to SQL. I've not tested it but I believe this change will introduce a new bug that no analyst will be able to approve a release as this will throw an exception when calculating the checklist.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseChecklistService.cs
```diff
@@ -169,7 +170,7 @@ private async Task<bool> ReleaseSectionHasNonEmptyHtmlBlock(Guid releaseId, Cont
                     cb.ContentSection!.Release.ReleaseId == releaseId &&
                     cb.ContentSection.Type == sectionType)
                 .OfType<HtmlBlock>()
-                .AnyAsync(htmlBlock => !string.IsNullOrEmpty(htmlBlock.Body));
```
benoutram, line 172: Same comment as above, I think this will throw an exception producing the release checklist. I suggest we don't make this change and have a chat to consider whether `StringExtensions.IsNullOrEmpty` is worth any value, given that it can't be translated to SQL out-of-the-box.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseChecklistService.cs
benoutram: Tested this now by running this branch locally, and I'm afraid it's bad news 😢 

image

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseChecklistService.cs
ntsim: Yeh fair point, it's unfortunate that it's a bit of a LINQ footgun.

Personally, I'm not against having this, as a similar thing applies for most extension methods we might want to use in a LINQ context. Kinda feels a bit unfair to single this out specifically.

If people feel particularly strongly about it, happy for us to go with whatever though.

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/ContactSaveRequest.cs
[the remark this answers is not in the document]
markhiveit: > You mean frontend-only validation right?

Sorry yeah frontend-only validation.

I'm not keen on implementing this just because I don't feel like it's worth the time adding. If an analyst were to get around the validation, it would be simple to change after the fact. So even if an analyst got around the validation, I don't think it's a big deal.

If you're keen on it, maybe take it up with Cam?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseChecklistService.cs
markhiveit: I'll just restore them all.

markhiveit: Done

benoutram: Ok cool, will approve. Fyi, I was only aware of this being a problem in `ReleaseChecklistService`. It was probably fine in the other cases. The difference is whether or not the expression needs to be translated by LINQ to SQL.

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/ContactSaveRequest.cs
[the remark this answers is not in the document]
markhiveit: Added validation after our discussion with Cam.

I've not used a new attribute, but `ContactSaveRequest` now inherits `IValidatableObject`. I didn't use `MinLength` as `ContactTelNo` can be null - I just put all the validation into the `Validate` method. Also added unit tests in `ContactSaveRequestTests.cs`.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Requests/ContactSaveRequestTests.cs
```diff
@@ -0,0 +1,96 @@
...
+                ContactTelNo = contactTelNo,
+            };
+
+            var validationContext = new ValidationContext(request);
+
+            var results = request.Validate(validationContext);
+
+            var validationResult = Assert.Single(results);
+            Assert.Equal("Contact telephone must only contain numbers and spaces", validationResult.ErrorMessage);
+        }
```
ntsim, line 74: Don't need these tests if we have a `UKPhone` attribute as described in another comment.

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/publication/components/PublicationContactForm.tsx
```diff
@@ -36,16 +36,17 @@ const PublicationContactForm = ({
...
-          return false;
-        }
-        return true;
-      },
-    }),
+    contactTelNo: Yup.string()
+      .trim()
+      .matches(
+        /^[0-9\s]*$/,
+        'Contact telephone must only contain numeric or whitespace characters',
```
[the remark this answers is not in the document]
markhiveit: I just ran this past Cam to make sure he's happy with it, and he said he's happy for us to add an extra validation that telephone numbers begin with a zero, so I'll add that.

I don't believe a number beginning with a zero ensures that it is a UK number - I believe it's `00` for international calls? - so I'll choose a new attribute and validation message with that in mind.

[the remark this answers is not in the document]
ntsim: It would be a UK number in the sense that you could dial it in the UK and it would be legitimate. See: https://en.wikipedia.org/wiki/Telephone_numbers_in_the_United_Kingdom

> All mobile numbers, nearly all geographic numbers, and nearly all non-geographic numbers have ten national (significant) digits after the [trunk code](https://en.wikipedia.org/wiki/Trunk_prefix) 0.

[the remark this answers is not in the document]
markhiveit: Should all be sorted.

I've named the attribute `PhoneNumber` - happy to change it if you'd prefer something else.

ntsim (approved):

PR #4321, opened 2023-09-22
EES-4486 New Index

Dancemammal:
Explain the **motivation** for making this change. What existing problem does the pull request solve?

You might also want to think about the following:

- How did you solve complex parts of the problem? Other contributors may benefit from knowing more about your solution.
- Are there any potential tradeoffs that need mentioning, or edge cases that have not been covered?
- Have you added a reusable piece of functionality? Consider including some code snippets to better illustrate its usage.

## Related changes

Are there any changes that were closely related to this change but not directly part of the main work e.g. refactoring, fixes, etc?

## Screenshots

Are there any screenshots that help illustrate the change made?


src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20230906093324_EES4486_AlterIndexDataImportsFileId.cs
benoutram, line 1: There's a few different problems here so I will try and keep this all in one comment.

- The PR doesn't have a description of the change being made - I think you've accidently included the Github issue template without making any changes to it when raising the PR.
- The build is failing for a few different reasons at compile time:
  - As well as this class, the `dotnet ef migrations add` cli command also generates a partial class ending `.Designer.cs` containing metadata which hasn't been added to git here.
  - `MigrationConstants.MigrationPath` should be `MigrationConstants.ContentMigrationsPath`.
  - `PreviousRebuildIndexesMigrationId` has a compile error because `InitialCreate_Custom` hasn't been imported but it isn't needed here anyway as we don't have have a previous version of a migration which added this index on its own in a definition that we can rollback to. If we want to have a version to rollback to it would be best to bundle it with this migration Id with a file like `20230906093324_Index_IXDataImportsFileId.previous.sql` and execute that instead in the `Down` method.
-  The `Down` method is currently going to execute `Current_Index_IXDataImportsFileId.sql`. The intention of our `Current` files is to contain the current object state so that when we change objects we can easily see a diff between its old and new state. This 'Current' version wouldn't be the version we want to rollback to.
 
However, before you make any of these changes...

The EES-4486 ticket description says

> We should try and make this change with EF configuration if possible. 

This is because if we modify this index directly by DDL update with SQL outside of entity frameworks control, then entity framework will not be aware of this model change. The configuration of the new index isn't tracked and if someone was to make further changes to the table or index configuration in entity framework, the code generated for the migration to apply the diff wouldn't be based on the current state of the index in the database, but the last known state based on the current model in code.

I did a bit of research and it looks like it's pretty easy to make a change to the existing nonclustered index to include this field, which _should_ result in the same changes under the hood as the one you are making here.

In `GovUk.Education.ExploreEducationStatistics.Content.Model\Database\ContentDbContext.cs` around L110, if you add the following

```
modelBuilder.Entity<DataImport>()
    .HasIndex(import => import.FileId)
    .IncludeProperties(
        import => new { import.Status });
```

Then generate a new migration, .e.g. `dotnet ef migrations add EES4486_AlterIndexDataImportsFileId`.

Then have a look at the generated code you should see this...

```
protected override void Up(MigrationBuilder migrationBuilder)
{
    migrationBuilder.DropIndex(
        name: "IX_DataImports_FileId",
        table: "DataImports");

    migrationBuilder.CreateIndex(
        name: "IX_DataImports_FileId",
        table: "DataImports",
        column: "FileId",
        unique: true)
        .Annotation("SqlServer:Include", new[] { "Status" });
}

protected override void Down(MigrationBuilder migrationBuilder)
{
    migrationBuilder.DropIndex(
        name: "IX_DataImports_FileId",
        table: "DataImports");

    migrationBuilder.CreateIndex(
        name: "IX_DataImports_FileId",
        table: "DataImports",
        column: "FileId",
        unique: true);
}
```

I tried this and then started the Admin to test the migration and saw that the existing `FileId` index on the table was migrated to be a nonclustered index including `Status`, and this was the only code required.

benoutram (changes requested):

PR #4322, opened 2023-09-25
EES-3849 Lowercase API request paths and add Trim to string model binding

benoutram:
This PR fixes a bug where backend API requests that use the blob cache internally could create additional cache entries for any resources which use string parameters as database and cache keys, by appending trailing spaces and/or by changing the case of those keys in the request path.

These keys are typically captured in route parameters.

E.g. for the public `GetRelease` endpoint `publications/{publication-slug}/releases/{release-slug}` the following variations will all fetch the same resource for the publication slug `pupil-attendance-in-schools` and release slug `2023-week-29`:

```
/api/publications/pupil-attendance-in-schools/releases/2023-week-29%20%20
/api/publications/pupil-attendance-in-schools%20%20/releases/2023-week-29
/api/publications/pupil-attendance-in-schools/releases/2023-Week-29
/api/publications/Pupil-Attendance-In-Schools/releases/2023-Week-29
```

This was bad for the following reasons:

* Unlimited variants of the same resource's URL could be requested by appending a differing number of trailing spaces to the URL and/or by using combinations of mixed case in the URL. Each new unique URL evades the original intended cache entry of the resource. After a cache miss for the new key, the resource is fetched from the database in potentially heavy query(s) that caching was meant to mitigate against.

* Each 'variant' key is added as a new cache entry. As well as wasting storage in the cache, the entries are never invalidated when changes to that resource are made. As an example, a cached release contains publication data which includes the latest published release and other release versions. These are immediately out of date as soon as a new release is published so the cache entry is updated at publishing time. However these 'variant' cache entries now become stale but are still accessible with incorrect data. Public URL's responsible for these 'variants' may have been shared, published on other websites, bookmarked, and indexed by search engines so users end up seeing old data. This was seen in [EES-4496](https://dfedigital.atlassian.net/browse/EES-4496).

This happens because of behaviour in the database when fetching resources after a cache miss.

After a miss on the cache key, the resource is located in the database using string keys as query parameters. These queries find results even though the parameters have trailing space and don't match by case. The SQL Server database collation we use is case-insensitive so there's no exact case matching on string equality. Trailing space doesn't effect string comparison because the ANSI SQL standard for string comparison right-pads strings to be the same length. Try the query `SELECT 'Equal!' WHERE 'abc    ' = 'ABC'` will return one result.

### Considerations for fixing this

* I avoided tackling trailing space in comparisons at a database level because that's an ANSI SQL standard. We could have done something like change every comparison to also compare length, or append some text to the end of both strings but both felt hacky.

* I avoided changing the database collation or some other change like adding an explicit collation to every query because that wouldn't be a solution here either. Adding case sensitivity would prevent all requests using mixed or upper case.
Using upper case in a request path should be ok, we just don't want resources found using mixed case keys to be cached uniquely.

* I considered making a change in `CacheAttribute.GetCacheKey` when constructing a `CacheKey` to lower case all parameters like slugs but it would affect all string parameters used in a `CacheKey` constructor, not just those used by the `CacheKey.Key`.

* I considered making a change in `BlobCacheService` `GetItemAsync` and `SetItemAsync` methods to lowercase `CacheKey.Key` which is used for the blob path but that would mean blob cache key's could never be case-sensitive in any circumstances even if we wanted them to be.

### How this was fixed

* Added new `LowercasePathRule` which rewrites all request paths to lowercase (bar the query string and certain exceptions such as static resources excluded by regex) which guarantees all route parameters captured in paths are lowercase.

* Added new `TrimStringModelBinder` as a replacement for binding primitive type `string` by `SimpleTypeModelBinder`.  `TrimStringModelBinder` does the same thing as `SimpleTypeModelBinder` except for trimming whitespace on the result of model binding.

* I made an additional change to helper methods of `FileStoragePathUtils` which produce all the paths based on slugs to make them safe for whitespace and always return lowercase paths no matter what data source the paths are constructed from.

### Other changes

- Remove unnecessary `GetKey` methods in `PublicationTreeCacheKey` and `AllMethodologiesCacheKey`.
- Add new sets of unit tests in classes `FileStoragePathUtilsTests.PublicPathTests` and `FileStoragePathUtilsTests.PrivatePathTests`.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/TestStartup.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/TestStartup.cs
@@ -1,7 +1,9 @@
 #nullable enable
-using GovUk.Education.ExploreEducationStatistics.Common.ModelBinding;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Common.Rules;
 using GovUk.Education.ExploreEducationStatistics.Common.Tests.Fixtures;
 using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Rewrite;
 using Microsoft.Extensions.DependencyInjection;
 using Newtonsoft.Json;
 
@@ -23,16 +25,18 @@ public class TestStartup
                 options => { options.SerializerSettings.NullValueHandling = NullValueHandling.Ignore; }
             );
 
-        services.AddControllers(
-            options =>
+        services.AddControllers(options =>
             {
-                options.ModelBinderProviders.Insert(0, new SeparatedQueryModelBinderProvider(","));
+                options.AddCommaSeparatedQueryModelBinderProvider();
+                options.AddTrimStringBinderProvider();
             }
         );
     }
 
     public void Configure(IApplicationBuilder app)
     {
+        app.UseRewriter(new RewriteOptions()
+            .Add(new LowercasePathRule()));
         app.UseMvc();
     }
 }
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Rules/LowercasePathRule.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Rules/LowercasePathRule.cs
@@ -1,0 +1,45 @@
+﻿#nullable enable
+using System.Linq;
+using System.Text.RegularExpressions;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Rewrite;
+using Microsoft.Net.Http.Headers;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Rules;
+
+public class LowercasePathRule : IRule
+{
+    private const string ExcludedPathsRegex = "^/(_configuration|identity|static).*$";
+
+    public void ApplyRule(RewriteContext context)
+    {
+        var request = context.HttpContext.Request;
+
+        var scheme = request.Scheme;
+        var pathBase = request.PathBase;
+        var path = request.Path;
+        var host = request.Host;
+        var query = request.QueryString;
+
+        if (IsApplicable(path))
+        {
+            var location = $"{scheme}://{host.Value}{pathBase.Value}{path.Value}".ToLower() + query;
+
+            var response = context.HttpContext.Response;
+            response.StatusCode = StatusCodes.Status308PermanentRedirect;
+            response.Headers[HeaderNames.Location] = location;
+            context.Result = RuleResult.EndResponse;
+        }
+        else
+        {
+            context.Result = RuleResult.ContinueRules;
+        }
+    }
+
+    private static bool IsApplicable(PathString path)
+    {
+        return path.HasValue
+               && !Regex.IsMatch(path, ExcludedPathsRegex, RegexOptions.IgnoreCase)
+               && path.Value.Any(char.IsUpper);
+    }
+}
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Rules/LowercasePathRule.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Rules/LowercasePathRule.cs
@@ -7,6 +7,22 @@ using Microsoft.Net.Http.Headers;
 
 namespace GovUk.Education.ExploreEducationStatistics.Common.Rules;
 
+/// <summary>
+/// Rule which is used to redirect any request paths that contain uppercase characters to lowercase.
+/// This excludes any paths that match the regex defined in <see cref="ExcludedPathsRegex"/>.
+/// Examples of requests that we might want to exclude:
+/// <list type="bullet">
+/// <item>
+/// <description>Oidc configuration</description>
+/// </item>
+/// <item>
+/// <description>Identity endpoints</description>
+/// </item>
+/// <item>
+/// <description>Static assets</description>
+/// </item>
+/// </list>
+/// </summary>
 public class LowercasePathRule : IRule
 {
     private const string ExcludedPathsRegex = "^/(_configuration|identity|static).*$";
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/TestStartup.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/TestStartup.cs
@@ -1,5 +1,6 @@
 #nullable enable
-using GovUk.Education.ExploreEducationStatistics.Common.ModelBinding;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Common.Rules;
 using GovUk.Education.ExploreEducationStatistics.Common.Tests.Fixtures;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Tests.Utils;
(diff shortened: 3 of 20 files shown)


src/GovUk.Education.ExploreEducationStatistics.Common/Rules/LowercasePathRule.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        var scheme = request.Scheme;
+        var pathBase = request.PathBase;
+        var path = request.Path;
+        var host = request.Host;
+        var query = request.QueryString;
+
+        if (IsApplicable(path))
+        {
+            var location = $"{scheme}://{host.Value}{pathBase.Value}{path.Value}".ToLower() + query;
```
ntsim, line 26: Just wondering, but did we consider just trimming the spacing here? Feels like it'd be simpler than adding all the additional model binding code?

benoutram: Yes I did consider this but trimming `path.Value` here wouldn't trim whitespace from parameters where a route template contains the parameters mid-way through it rather than just at the end, e.g. `/api/publications/{publication-slug}/releases/latest` or `/api/publications/{publication-slug}/releases/{release-slug}`.

Did you mean trying to remove whitespace part way through the path string? I didn't think that would be safe as I couldn't think of a way of just targeting the route parameters here without knowing them in advance to trim them.

The model binder solution for trimming comes with the bonus of trimming parameters that come from value providers for query string values and form values as well as route values (see `TrimStringModelBinderIntegrationTests` for examples).

Aside from the problem with caching I'm trying to solve here, I really wanted to implement a model binder that would trim body values too for complex types using `[FromBody]`. Due to the way values are read from the message body using a JSON media-type formatter and deserialised, a one-fits-all solution to trimming wasn't possible so I gave up with that.

ntsim: Couldn't we just strip out any whitespace with a regex? Spaces are not actually allowed in URLs at all, so feels like we should be able to do this.

In fact, a similar problem exists with the model binding implementation as we're not accounting for permutations of things where the whitespace is within the path segment e.g. `some-path fragment`

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/TestStartup.cs
```diff
@@ -30,9 +32,11 @@ public void ConfigureServices(IServiceCollection services)
                 options => { options.SerializerSettings.NullValueHandling = NullValueHandling.Ignore; }
             );
 
-        services.AddControllers(
-                options => { options.ModelBinderProviders.Insert(0, new SeparatedQueryModelBinderProvider(",")); }
-            )
+        services.AddControllers(options =>
+            {
+                options.AddCommaSeparatedQueryModelBinderProvider();
+                options.AddTrimStringBinderProvider();
```
markhiveit, line 38: I was wondering whether we definitely want to trim spaces? I wonder if we want to instead error if there is spaces in query params?

src/GovUk.Education.ExploreEducationStatistics.Common/Rules/LowercasePathRule.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Text.RegularExpressions;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Rewrite;
+using Microsoft.Net.Http.Headers;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Rules;
+
+public class LowercasePathRule : IRule
+{
+    private const string ExcludedPathsRegex = "^/(_configuration|identity|static).*$";
```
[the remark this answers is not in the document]
benoutram: Cool, I've added comment to LowercasePathRule now.

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/TestStartup.cs
benoutram: Model validation occurs after model binding.
If a required field bound to a query parameter is empty after the trimming done by the model binder, it will error.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/TestStartup.cs
```diff
@@ -23,16 +25,18 @@ public void ConfigureServices(IServiceCollection services)
...
 
-        services.AddControllers(
-            options =>
+        services.AddControllers(options =>
             {
-                options.ModelBinderProviders.Insert(0, new SeparatedQueryModelBinderProvider(","));
+                options.AddCommaSeparatedQueryModelBinderProvider();
+                options.AddTrimStringBinderProvider();
             }
         );
```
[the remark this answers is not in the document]
benoutram: I didn't have to make any changes to the Admin `TestStartup` due to a recent change made by @duncan-at-hiveit where it now inherits the configuration of the read Admin `Startup`. I think this is the way to go for the Content and Data Api's `TestStartup` config but don't think that's a change to be made here. I will create a new ticket for it if there's not one already.

markhiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Common/Rules/LowercasePathRule.cs
benoutram: I agree URL's should not contain spaces but spaces can exist in URL's in URL encoded notation (decoded in the request prior to model binding) and I think it's safer to only target parameters that are being bound to the model and sanitise those by trimming leading and trailing whitespace from them instead of removing spaces from a URL indiscriminately.

The model binder implemented here never intends to remove whitespace from _within_ parameters,  e.g. your example `some-path segment` would be left alone, just as I would expect it to be after sanitisation if it's a string parameter bound to the model.

Also spaces _within_ parameters that are used for cache keys don't cause the problem with caching that I'm trying to address here. Parameters which go on to be used as keys in a database lookup after a cache-miss will find a result only if there's an exact match just as we'd expect, unlike if they contain trailing whitespace where they still find a result.

It could be argued that spaces within route parameters or other parts of the URL which shouldn't be there (e.g. introduced in to parts of the URL by something other than not sanitising user inputs from leading/trailing whitespace) should be fixed at source and return 404.

I agree we could make the change you suggest here, changing the rule to remove all spaces as well as lowercasing like it does now and that it would have the advantage of returning a clean redirect the client.

I guess we'd have to tread carefully around removing spaces from the query string, which could genuinely contain spaces, so we'd only want to remove all spaces before the query string, and then either leave the query string untouched (same as with the lowercasing), or go on to just trim individual query string parameters without removing spaces indiscriminately. As I mentioned before, the model binder implemented here will already handle trimming parameters from query sources and form sources too.

Apart from the redirect I'm not convinced that there's any benefit to doing this with a rule over what I've already implemented with the model binder, or that it's worth the effort changing it.

ntsim: If we did make this change here, you're right that it'd be best to only do it on the main part of the URL and not the query string.

If we're worried about it being a bit too aggressive of a measure, happy for us not to do this. As long as we get the intended behaviour for malformed URLs (404s) instead of a cached result, it's fine. From what you're saying, this seems to be the case, so happy to go with it.

I was thinking we'd be able to simplify the code a little bit without the model binders, but as you've said, the model binders also trim query and form input. That alone seems worth it to keep them in!

benoutram: Ok thanks for your time spent considering the options here. It's been interesting understanding how controller actions and model binding work and weighing up the different options for doing this.

If it turns out to not work or not do what we want then we can switch plan! 🙂

PR #4323, opened 2023-09-25
EES-4568 Update data guidance text file to allow links

ntsim:
This PR adds HTML to text conversion for links in the data guidance text file. These are converted to something like `the link text (https://the-link)`.

As part of this, we've significantly refactored the HTML to converter code so that:

- Elements are now categorised as block, inline and table elements
- Spacing is calculated based on the type of element and the elements in its adjacent context
- We no longer have to cover every single HTML element in a big switch statement.


src/GovUk.Education.ExploreEducationStatistics.Common/Utils/Html/HtmlToTextConverter.cs
```diff
@@ -148,40 +170,58 @@ private void ParseTextNode(INode node)
                 return;
             }
 
-            // As there are potentially inline elements such as span/strong/em,
-            // we should try and add an extra space to avoid the text bunching up.
+            var prev = node.PreviousNonWhitespaceSibling();
+
+            if (prev is null)
+            {
+                if (node.Parent is IElement parentElement && parentElement.IsInlineType())
```
duncan-at-hiveit, line 177: Might be worth a few comments around the usage of "prev", "prevElement" and "parentElement" for the uninitialted if they come to this down the line.  I'm guessing here that we're saying for instance that if this is the first child nested within a parent element, then we need to be breaking out of the parent and seeing what the parent's previous sibling was in order to determine if whitespace is necessary or not.  Clear to me now after I've studied it a bit, but I had to think about it :slightly_smiling_face:

duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/Html/HtmlToTextConverter.cs
ntsim: Good point! I've updated the comments in this area of code and significantly bumped up the level of detail and context, if it looks good to you now?

duncan-at-hiveit: Yup, that's perfect, thanks!

duncan-at-hiveit (approved):

PR #4324, opened 2023-09-27
EES-4575 fix PRA email validation

amyb-hiveit:
Replaces the custom email validation for PRA users with Yup email validation so that `gov.wales` and other valid TLDs are not rejected.

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
@@ -61,10 +61,6 @@ const PreReleaseUserAccessForm = ({
 
   const [invitePlan, setInvitePlan] = useState<PreReleaseInvitePlan>();
 
-  const isValidEmail = (input: string) => {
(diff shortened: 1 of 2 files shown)


src/explore-education-statistics-admin/src/pages/release/pre-release/components/PreReleaseUserAccessForm.tsx
```diff
@@ -141,8 +137,9 @@ const PreReleaseUserAccessForm = ({
                 test(value?: string) {
                   if (value) {
                     const emails = splitAndTrimLines(value);
+                    const schema = Yup.string().email();
                     const indexOfFirstInvalid = emails.findIndex(
-                      email => !isValidEmail(email),
+                      email => !schema.isValidSync(email),
```
ntsim, line 142: Just checking, but assuming the previous validation was more restrictive on purpose, are we now making this too lax?

Have we checked this with Cam?

amyb-hiveit: I've checked with Cam, we don't think the restriction was on purpose and he's fine with it using the Yup validation instead. This is the ticket https://dfedigital.atlassian.net/browse/EES-3582 the regex was added for.

ntsim (approved):

PR #4325, opened 2023-09-27
EES 4486

Dancemammal:
Motivation

This change updates a current index to include a field in the results that will save time on a key lookup and reduce the query execution path thus speeding up queries that rely on this index.

The change came from identifying why a SQL query was running slowly, I used tools that helped me to identify the query execution plan and with that, I was able to identify how to speed up the query. 

## Related changes

No related changes

## Screenshots

SQL Code only, no need for screenshots


benoutram (approved):
I ran this migration and observed the `IX_DataImports_FileId` index change to include the `Status` column.

PR #4328, opened 2023-09-28
EES-4577 convert some more forms to react hook form

amyb-hiveit:
Replaces Formik with React Hook Form in the following places:

- table headers reordering form in (table tool in public and admin apps (preview & pre-release), create / edit data blocks)
- download form (table tool final step in public and admin apps (preview & pre-release), permalink page)
- add and edit comments forms
- create publication form (I've tidied up this a bit as it had some unused stuff from when the same for was used to edit publications)
- publication details form
- publication contact details form

## Server validation error mapping for RHF

Previously, server validation errors would be converted to field errors, however this implementation was specific to Formik. 

We've now changed this to be non-library specific, providing a `mapServerFieldErrors` function that outputs a list of generic field errors. The field errors can then be iterated over and the error messages added to the form fields for both Formik and RHF.

## Changes to display server errors and submit errors

We've made some substantial changes to the way that server validation errors and submit errors are displayed in the form's error summary. Most logic around this has been moved to the respective components:

- Formik - logic now in `useFormSubmit`
- RHF - logic now in `FormProvider`

Consolidating this logic into these areas allows us to display submit errors when the form submission is programmatically triggered outside of a `form` element's submit event. This is particularly useful for cases where we need the user to press a confirmation button before the form is actually submitted.

Previously, a user would be able to trigger an RHF / Formik submit, but this would not result in any handling of submit errors.

### Server validation errors

Server validation errors were previously being added to form state indiscriminately. This meant that upon form submission, this could result in phantom errors that don't relate to any specific field.

We now check each server validation error to see if has a matching form field (via the form's values at the time of submission). If there is no matching form field, we just don't display these errors anymore.

If there are no server validation errors that map to form fields, a default error message ('The form submission is invalid and could not be processed'). If we want to change the message, we can use the `fallbackServerValidationError` option.

### Submit errors

Submit errors are any error that isn't a server validation error. These can typically be thrown unexpectedly whilst a form submit is in progress.

Previously, if the form component did not have a `showSubmitError` prop set, it would most likely result in an error page. Showing the submit error would also be likely to display user unfriendly messages.

We've now improved this so that submit errors can be more precisely controlled by either:

- Using the `fallbackSubmitError` option (as a general fallback)
- Throwing a `SubmitError` during the form submission

Throwing a `SubmitError` allows a more specific error message to be displayed. Additionally, we can even specify a form field that it should apply to (this changes the link).

As part of this, we've now removed the `showSubmitError` prop as it's no longer necessary.

## Improved `as` prop for `RHFFormField`

We've now completely removed the render props from `RHFFormField` in favour of the `as` prop. 

The `as` prop has also been improved to result in type errors more easily if we try to use an incompatible form component. It was previously too permissive and would allow invalid components to be used.

## Other changes

- Various cleanup of testids and labels on form elements e.g. unnecessary `comment-textarea` on all textareas.
- Added call to error control's `handleError` in during RHF form submits (this previously wasn't being called).


mmyoungman (approved):

mmyoungman:
For the future, I've not reviewed this - Amy and Nick have reviewed each other's work here.

PR #4333, opened 2023-10-02
EES-4585 Include Contact name in Contact Us section

markhiveit:
This PR includes a publication contact's contact name in the Contact Us sections across the service. It previously wouldn't be present if the contact telephone wasn't present. We've also changed the layout of the contact details:

image

### Other notes
- I've used `<a href="tel:[...]">` for other phone numbers in the Contact Us section and used the `Telephone: <telNo>` format so it is aligned with how a publication contact's details are displayed. 
- I also fixed a backend validation issue with `PhoneNumberAttribute`. The validation was failing if the frontend provided an empty string or whitespace - we want to succeed in those cases because the contact telephone is now optional.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Validators/PhoneNumberAttributeTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Validators/PhoneNumberAttributeTests.cs
@@ -11,8 +11,6 @@ public class PhoneNumberAttributeTests
     [InlineData("01234567")]
     [InlineData("0   123 4567")]
     [InlineData("0 1 2 3   4 5 6 7      ")]
-    [InlineData("")]
-    [InlineData("    ")]
     public void AllowedValues_Valid(string value)
     {
         var attribute = new PhoneNumberAttribute();
@@ -41,14 +39,27 @@ public class PhoneNumberAttributeTests
         Assert.Equal("A phone number must be a string", result?.ErrorMessage);
     }
 
-    [Fact]
-    public void DisallowedValues_TooShort()
+    [Theory]
+    [InlineData("0123456        ")]
+    [InlineData("0         ")]
+    [InlineData("     0123456")]
+    public void DisallowedValues_TooShort(string value)
     {
         var attribute = new PhoneNumberAttribute();
-        var value = "0123456";
         var result = attribute.GetValidationResult(value, new ValidationContext(value));
 
         Assert.Equal("The value must have a minimum length of 8.", result?.ErrorMessage);
+    }
+
+    [Theory]
+    [InlineData("")]
+    [InlineData("    ")]
+    public void DisallowedValues_EmptyStringOrWhitespace(string value)
+    {
+        var attribute = new PhoneNumberAttribute();
+        var result = attribute.GetValidationResult(value, new ValidationContext(value));
+
+        Assert.Equal("A phone number cannot be an empty string or whitespace", result?.ErrorMessage);
     }
 
     [Theory]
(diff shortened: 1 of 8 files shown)


N-moh (approved):

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Validators/PhoneNumberAttributeTests.cs
```diff
@@ -11,6 +11,8 @@ public class PhoneNumberAttributeTests
     [InlineData("01234567")]
     [InlineData("0   123 4567")]
     [InlineData("0 1 2 3   4 5 6 7      ")]
+    [InlineData("")]
+    [InlineData("    ")]
```
benoutram, line 15: I tried adding this as a test case which I think should fail but it passes:

`    [InlineData("0 1 2 3   4 5 6        ")]`

markhiveit: That should pass. The minimum characters includes spaces. I didn't consider it worth the effort to cover that case.

benoutram: I appreciate the phone number might be formatted with spaces like `01234 567890` or `0123 456 789`. I'm just a bit surprised that you can pass the validation by appending whitespace.

These would have been better examples than in my original comment of passing test cases which I think we would expect to fail:

```
[InlineData("0       ")]
[InlineData("0123456 ")]
```

src/GovUk.Education.ExploreEducationStatistics.Common/Validators/PhoneNumberAttribute.cs
```diff
@@ -20,6 +21,11 @@ public class PhoneNumberAttribute : ValidationAttribute
             return new ValidationResult("A phone number must be a string");
         }
 
+        if (telNo.IsNullOrWhitespace())
+        {
+            return ValidationResult.Success;
+        }
```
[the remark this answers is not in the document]
markhiveit: I'll make the change, but not sure I consider it a big deal if the backend is more permissive and accepts whitespace. There is already backend code to coerce whitespace and empty strings into null for the database.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Validators/PhoneNumberAttributeTests.cs
markhiveit: Those would fail. It's whitespace between characters that causes an issue. So '0                0' passes when you would probably expect it to fail.

benoutram: I only left this comment because they were passing when I tried them. If they fail now after recent changes, that's good. 🙂

ntsim (approved):
Approved from my perspective, but might want to let @benoutram approve too before proceeding

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Validators/PhoneNumberAttributeTests.cs
markhiveit: Okay - actually you're correct - those will pass on the backend. I'll add a fix for that.

markhiveit: I've pushed a fix.

benoutram (approved):

PR #4343, opened 2023-10-05
EES-4566 Add AddTransient line for ViewRelease auth handler

mmyoungman:
As simple as the title says.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Security/StartupSecurityConfiguration.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Security/StartupSecurityConfiguration.cs
@@ -211,6 +211,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Security
             /**
              * Release management
              */
+            services.AddTransient<IAuthorizationHandler, ViewReleaseAuthorizationHandler>();
             services.AddTransient<IAuthorizationHandler, ViewSpecificReleaseAuthorizationHandler>();
             services.AddTransient<IAuthorizationHandler, UpdateSpecificReleaseAuthorizationHandler>();
             services.AddTransient<IAuthorizationHandler, DeleteSpecificReleaseAuthorizationHandler>();


benoutram (commented):
I've read the Jira description and comment and it's not clear on what was actually broken here before, if anything.

It sounds as though there was a misunderstanding that calls to `CheckCanViewRelease` should have led to `ViewReleaseAuthorizationHandler` being invoked, and that because it wasn't that there must have been some gap in permissions checking.

I see that this PR configures it so it will now be invoked, but I would say it was probably correct before. 

`Content.Security.AuthorizationHandlers.ViewReleaseAuthorizationHandler` is used by the Content and Data API's to control public access to published releases.

`Admin.Security.AuthorizationHandlers.ViewSpecificReleaseAuthorizationHandler` is used by the Admin to control private access to all releases.

If there was a problem with access to releases in the Admin I would have first looked at `ViewSpecificReleaseAuthorizationHandler` to check if that needs a change making to it first.

If there is a problem where some area of Admin is not allowing access to published releases when it should do, this possibly is the right change to make. However I think that adding it here if that's not the case is probably a mistake.

Presumably both `ViewReleaseAuthorizationHandler` and `ViewSpecificReleaseAuthorizationHandler` are now invoked on every `CheckCanViewRelease` call, and `ViewReleaseAuthorizationHandler` will pull back all releases of the publication just to check that it's the latest.

If it turns out the published check is required, we could rewrite `ViewSpecificReleaseAuthorizationHandler` to not be a `CompoundAuthorizationHandler` (which we would like to do anyway), i.e. by putting all the logic into a single `HandleRequirementAsync` method. When doing that we could pay special attention to the order of the different success conditions. For example, any that require little to no processing like the claim check could be ordered first, followed by the other conditions in order that they are most likely to be relevant/succeed.

`ViewReleaseRequirement` existing within the same file `ViewReleaseAuthorizationHandler.cs` but applied to both the admin and public handler is probably what's caused a big part of the confusion here.

PR #4347, opened 2023-10-06
EES-4599 Fix `HtmlToTextConverter` outputting OS-dependent line endings

ntsim:
This changes `HtmlToTextConverter` to only output Windows CRLF line endings. We're specifically only using CRLF as this will work best with the most users (who use Windows).

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Utils/Html/HtmlToTextConverter.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Utils/Html/HtmlToTextConverter.cs
@@ -36,15 +36,18 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Utils.Html
         {
             var rawLines = _builder.ToString().ToLines();
 
-            // Run post-filtering on lines to remove any extraneous
-            // whitespace from the end of lines. This is mostly for
-            // better compat with editors/IDEs, where whitespace may
-            // be automatically removed upon file save.
-            var filteredBuilder = new StringBuilder();
+            // Run post-filtering on lines to:
+            // 1. Remove any extraneous whitespace from the end of lines. This is
+            // mostly for better compatibility with formatters / editors / IDEs, where
+            // trailing whitespace may be automatically removed upon file save.
+            // 2. Ensure we have CRLF (\r\n) line endings for better compatibility with
+            // our Windows users (the majority of users). Otherwise, line endings
+            // would be server OS dependent and this can lead to inconsistent files.
+            var postFilterBuilder = new StringBuilder();
 
-            rawLines.ForEach(line => filteredBuilder.AppendLine(line.TrimEnd()));
+            rawLines.ForEach(line => postFilterBuilder.AppendCrlfLine(line.TrimEnd()));
 
-            return filteredBuilder.ToString().TrimEnd();
+            return postFilterBuilder.ToString().TrimEnd();
         }
 
         private void ParseChildren(INode node)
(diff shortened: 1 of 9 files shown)


src/GovUk.Education.ExploreEducationStatistics.Common/Utils/Html/HtmlToTextConverter.cs
```diff
@@ -145,9 +148,13 @@ private void AddSpacing(IElement element)
...
-                // If last two characters are new line characters, then don't
-                // append any more spacing as this would be excessive.
-                if (_builder[^2] == '\n' && _builder[^1] == '\n')
+                // Get last 4 chars to check their line endings. Windows uses CRLF line endings (\r\n),
+                // so there may be up to 4 chars, instead of 2 for LF (\n) when using Linux / Mac.
+                var currentEnding = _builder.Substring(^4..);
+
+                // If already have two line endings, then don't
+                // append more spacing as this would be excessive.
+                if (currentEnding.EndsWith($"{Environment.NewLine}{Environment.NewLine}"))
```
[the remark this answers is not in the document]
ntsim: I've improved the spacing behaviour a bit so that it's more consistent to avoid appending too many lines as you've suggested.

As it's a bit smarter about how it appends lines, I've also applied it more generally in `ParseElement`. A nice improvement as lines don't need to be appended on a case-by-case basis anymore.

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/Html/HtmlToTextConverter.cs
```diff
@@ -127,32 +127,38 @@ private void ParseElement(IElement element)
...
                 {
                     return;
                 }
 
-                _builder.AppendLine();
+                // Don't a trailing line ending currently, so it should
```
sambiramairelogic, line 154: Super important comment I know but there's a slight typo here in `Don't a trailing...` :P

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/Html/HtmlToTextConverter.cs
```diff
@@ -36,15 +36,18 @@ private string Output()
...
+            // mostly for better compatibility with formatters / editors / IDEs, where
+            // trailing whitespace may be automatically removed upon file save.
+            // 2. Ensure we have CRLF (\r\n) line endings for better compatibility with
+            // our Windows users (the majority of users). Otherwise, line endings
+            // would be server OS dependent and this can lead to inconsistent files.
+            var postFilterBuilder = new StringBuilder();
+
+            rawLines.ForEach(line => postFilterBuilder.AppendCrlfLine(line.TrimEnd()));
+
+            return postFilterBuilder.ToString().TrimEnd();
```
sambiramairelogic, line 50: Not that it does any harm, but shouldn't this `TrimEnd()` be unnecessary since it's been called on each line anyway?

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/Html/HtmlToTextConverter.cs
ntsim: Good point! Fixed :wink:

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/Html/HtmlToTextConverter.cs
ntsim: It's necessary because we `TrimEnd` the contents of each line, before re-appending it with `AppendCrlfLine`.

This means without the final trim, we end up with an additional line break right at the end of the string, which isn't necessary. If we remove this, all the tests basically fail!

sambiramairelogic (approved):
Looks good to me (& most importantly runs on my Windows machine 😛 

Good work!

PR #4348, opened 2023-10-10
EES-4602 Validate slug is unique when creating new methodology

mmyoungman:
This PR adds slug validation when creating a new methodology. It no longer allows a methodology to be created if the slug is already being used.

### Notes
- As the `Create new methodology` button is a button and not a form, I'm not aware of an easy way to feed back the `SlugNotUnique` error from the backend to the user. So if someone did try to create a methodology with a duplicate slug, it now results in "Sorry, something went wrong" page - I figure this is better than a duplicate slug. Ultimately, this bug has been present for a long time and nobody has created a duplicate slug and the redirect work will fix this. So unless there is a quick fix, I'm inclined to leave this as it is for now.
- The code change to `MethodologyService#UpdateDetails`, whereby we only validate a slug if it has changed, cannot currently be hit. This is because currently the title and slug can only be updated in tandem. But I've added this now so it work correctly in theory. I've removed the `Slug` - which was generated via `NamingUtils.SlugFromTitle`, *not* the user - property from `MethodologyUpdateRequest` to make this clearer.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
@@ -364,6 +364,10 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.Method
                     .Setup(s => s.CreateMethodologyForPublication(publication.Id, User.Id))
                     .ReturnsAsync(createdMethodology);
 
+                methodologyVersionRepository
+                    .Setup(mock => mock.GetLatestPublishedVersionBySlug(publication.Slug))
+                    .ReturnsAsync((MethodologyVersion?)null);
+
                 context.Attach(createdMethodology);
 
                 var viewModel = (await service.CreateMethodology(publication.Id)).AssertRight();
@@ -382,6 +386,62 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.Method
                 Assert.Equal(publication.Id, viewModel.OwningPublication.Id);
                 Assert.Equal("Test publication", viewModel.OwningPublication.Title);
                 Assert.Empty(viewModel.OtherPublications);
+            }
+        }
+
+        [Fact]
+        public async Task CreateMethodology_SlugNotUnique()
+        {
+            var publication = new Publication
+            {
+                Title = "Test publication"
+            };
+
+            var contentDbContextId = Guid.NewGuid().ToString();
+            await using (var context = InMemoryApplicationDbContext(contentDbContextId))
+            {
+                await context.Publications.AddAsync(publication);
+                await context.SaveChangesAsync();
+            }
+
+            await using (var context = InMemoryApplicationDbContext(contentDbContextId))
+            {
+                var methodologyVersionRepository = new Mock<IMethodologyVersionRepository>(Strict);
+                methodologyVersionRepository
+                    .Setup(mock => mock.GetLatestPublishedVersionBySlug(publication.Slug))
+                    .ReturnsAsync(new MethodologyVersion());
+
+                var service = SetupMethodologyService(
+                    contentDbContext: context,
+                    methodologyVersionRepository: methodologyVersionRepository.Object);
+
+                var createdMethodology = new MethodologyVersion
+                {
+                    Id = Guid.NewGuid(),
+                    Methodology = new Methodology
+                    {
+                        Id = Guid.NewGuid(),
+                        Slug = "test-publication",
+                        OwningPublicationTitle = publication.Title,
+                        Publications = new List<PublicationMethodology>
+                        {
+                            new()
+                            {
+                                Owner = true,
+                                Publication = publication
+                            }
+                        }
+                    },
+                    Status = MethodologyApprovalStatus.Draft
+                };
+
+                context.Attach(createdMethodology);
+
+                var actionResult = (await service.CreateMethodology(publication.Id)).AssertLeft();
+
+                VerifyAllMocks(methodologyVersionRepository);
+
+                actionResult.AssertBadRequest(SlugNotUnique);
             }
         }
 
@@ -1574,7 +1634,13 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.Method
 
             await using (var context = InMemoryApplicationDbContext(contentDbContextId))
             {
-                var service = SetupMethodologyService(context);
+                var methodologyVersionRepository = new Mock<IMethodologyVersionRepository>(Strict);
+                methodologyVersionRepository
+                    .Setup(mock => mock.GetLatestPublishedVersionBySlug("updated-methodology-title"))
+                    .ReturnsAsync((MethodologyVersion?)null);
+
+                var service = SetupMethodologyService(context,
+                    methodologyVersionRepository: methodologyVersionRepository.Object);
 
                 var viewModel = (await service.UpdateMethodology(methodologyVersion.Id, request)).AssertRight();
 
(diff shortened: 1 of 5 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -511,18 +514,16 @@ private static IdTitleViewModel BuildPublicationViewModel(PublicationMethodology
             return new IdTitleViewModel(publication.Id, publication.Title);
         }
 
-        private async Task<Either<ActionResult, Unit>> ValidateMethodologySlugUniqueForUpdate(
-            MethodologyVersion methodologyVersion, string slug)
+        private async Task<Either<ActionResult, Unit>> ValidateMethodologySlugUnused(string slug)
```
ntsim, line 521: This name doesn't make a lot of sense. `ValidateMethodologySlugUnique` would be a more appropriate name.

ntsim (dismissed):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -511,18 +514,16 @@ private static IdTitleViewModel BuildPublicationViewModel(PublicationMethodology
...
         {
-            if (await _context
-                    .Methodologies
-                    .AsQueryable()
-                    .AnyAsync(p => p.Slug == slug && p.Id != methodologyVersion.MethodologyId))
+            var methodologyVersion = await _methodologyVersionRepository.GetLatestPublishedVersionBySlug(slug);
+
+            if (methodologyVersion != null)
             {
                 return ValidationActionResult(SlugNotUnique);
```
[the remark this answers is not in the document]
mmyoungman: Maybe I'm missing something obvious, but as we create the new methodology on `PublicationMethodologyPagej.tsx` via a button rather than a form, there is no `useFormSubmit` to pass the error mapper to.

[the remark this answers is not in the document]
ntsim: As discussed in Slack, realised that methodology creation is done via a button click so this currently isn't applicable.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -323,13 +325,18 @@ await _context.Entry(loadedMethodologyVersion.ScheduledWithRelease)
...
-                // another Methodology has previously set its AlternativeTitle (and Slug) to something specific and then
-                // this Methodology attempts to set its AlternativeTitle (and Slug) to the same value.  Whilst an
-                // unlikely scenario, it's entirely possible.
-                .OnSuccessDo(methodologyVersion =>
-                    ValidateMethodologySlugUniqueForUpdate(methodologyVersion, request.Slug))
+                .OnSuccessDo(async _ => {
+                    if (newSlug != methodologyVersionToUpdate.Methodology.Slug)
+                    {
+                        // Only need to validate if the slug is changing
+                        return await ValidateMethodologySlugUnique(newSlug);
```
sambiramairelogic, line 335: I think you could move the `newSlug != {existing}.Slug` check into this method. I appreciate we want to avoid making a call to the repository if we don't have to, but you can still do that inside the method. At the moment it feels like we're doing half the validation inside the method and half outside.

I also feel like this could be a bit more readable - it's surprising that a method named `ValidateFoo` should return an ActionResult or a Unit. I'd usually have expected it to return an object representing validation errors, or just a boolean flag for valid or not. Returning the ActionResult feels like we're merging our HTTP logic with our validation logic, and doing two things in one method.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/MethodologyVersionRepository.cs
```diff
@@ -82,6 +82,17 @@ await _contentDbContext.Entry(methodology)
                 .ToList();
         }
 
+        public async Task<MethodologyVersion?> GetLatestPublishedVersionBySlug(string slug)
+        {
+            return await _contentDbContext
+                .MethodologyVersions
+                .Include(mv => mv.Methodology)
+                .Where(mv =>
+                    mv.Slug == slug
```
sambiramairelogic, line 91: Is it worth making this equivalence check case invariant, or do we have that implicitly by this point?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
```diff
@@ -385,6 +389,62 @@ await using (var context = InMemoryApplicationDbContext(contentDbContextId))
...
+                    Status = MethodologyApprovalStatus.Draft
+                };
+
+                context.Attach(createdMethodology);
+
+                var actionResult = (await service.CreateMethodology(publication.Id)).AssertLeft();
+
+                VerifyAllMocks(methodologyVersionRepository);
+
+                actionResult.AssertBadRequest(SlugNotUnique);
```
sambiramairelogic, line 444: I don't think this should be a 400, since there isn't actually anything wrong with the request syntax or headers. I think a 409 (Conflict) would be more appropriate here, or 422 (Processable Entity)

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/MethodologyVersionRepository.cs
mmyoungman: I believe we have it implicitly.

The check is already case insensitive as in our databases:
```
SELECT 1
WHERE 'test-slug' = 'TEST-SLUG'
```
returns 1. But (in case you're not aware) the fun doesn't stop there... Our databases also don't take into account trailing whitespace in comparisons too. So:
```
SELECT 1
WHERE 'test-slug' = 'TEST-SLUG            '
```
also returns 1.

I think we have looked into changing the MSSQL defaults on this - both of these cases have stung us in the past - but for some reason buried in the past it didn't happen. Might be one worth exploring again at a dev meeting.

Other than that, I think we assume slugs are already lowercased in many places across the codebase. My understanding is that all publication and methodology slugs are currently generated via `NamingUtils.SlugFromTitle`. As far as I'm aware, the user cannot manually enter a publication/methodology slug - they can only change it indirectly by changing a title - and therefore we can currently assume that all slugs are already lowercase. Ben also did some work recently to ensure that slugs fetched from the URL are also lowercase: https://github.com/dfe-analytical-services/explore-education-statistics/pull/4322

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyServiceTests.cs
mmyoungman: What you're saying seems right to me, but I'm no HTTP status code expert, so I'll take your word for it.

I think we return 400 in many in other places where there isn't anything wrong with the request syntax or headers, so I believe this problem is spread throughout the codebase. If we were going to fix this, I'd be inclined to fix it everywhere.  And if we're going to do that, I'd prefer to arrive at a solution with the other devs first rather than jam something in for this PR for a single particular status code. If it was quick to change just this instance, I would, but I think it's going to need a fair chunk of code the way we're currently set up.

So if it's cool with you, I'm inclined to leave this PR as is and raise this at the next dev meeting - where we can agree a holistic solution and get it ticketed and prioritised. If you feel strongly about it getting it fixed now, we can canvas other's opinions - if others agree with you I'm happy to fix this instance as part of this PR.

ntsim: @sambiramairelogic by default, .NET outputs validation errors as 400. This arguably falls into the category of validation errors so it seems fine to me as a status code.

However, as you've said, a more strict interpretation of spec would perhaps indicate 409 or even 422 as being more appropriate. Unfortunately, I think because there's so much overlap in the definitions it's pretty unclear if there's really a true answer to this.

I'd personally lean towards 400 to keep things simple and consistent, rather than have to be too precise. Our private APIs aren't consumed publicly so adding more detail here isn't something to really sweat about.

In the frontend, any codes in the 400 range are basically treated as client errors so the precise semantics of the status code don't really matter that much.

PR #4362, opened 2023-10-17
EES-4607 fix table tool publication form error message

amyb-hiveit:
Fixes a bug on the first step of a table tool where when you clicked 'Next step' without selecting anything you'd get an error message of 'themeId cannot be null'. This has been changed to 'Choose a theme'.


sambiramairelogic (approved):
Good stuff!

PR #4369, opened 2023-10-23
EES-4574 Add new `LocalSkillsImprovementPlanArea` location attribute

benoutram:
This PR adds a new `LocalSkillsImprovementPlanArea` location attribute and allows importing LSIP level statistical data.

The `GeographicLevel` enum value `LocalSkillsImprovementPlanArea` was already added by https://github.com/dfe-analytical-services/explore-education-statistics/pull/4346 which also added geometry data for this level.

### Other changes

- Remove obsolete Locations field from `ObservationQueryContext` and delete `LocationQuery`.
- Add new test data files `LSIP.csv` and `LSIP.meta.csv` containing LSIP level data.
- Change test data file `all_geographics.csv` to include an LSIP level example.
- Attempt to tidy up some of the other test data files in the `small-files` directory where they are using all the location attribute column names unnecessarily.
- Remove the 4 files `tests/test-data/files/admin-api-*` - I couldn't really see what they are used for except being related to old Postman tests.


src/GovUk.Education.ExploreEducationStatistics.Data.Model/LocationAttribute.cs
```diff
@@ -182,6 +196,7 @@ public record PlanningArea(string? Code, string? Name) : LocationAttribute(Code,
 
 public record Provider(string? Code, string? Name) : LocationAttribute(Code, Name)
 {
+    [SuppressMessage("ReSharper", "StringLiteralTypo")]
```
sambiramairelogic, line 199: Perhaps we could exempt the `.idea/dictionaries` folder from the gitignore, and commit a project-level shared dictionary to avoid having to suppress typos like this?

sambiramairelogic (approved):
Looks good to me - as far as I can tell without much domain knowledge! 
Have just left one comment with an entirely optional improvement if you agree. Good work!

benoutram:
> Looks good to me - as far as I can tell without much domain knowledge! Have just left one comment with an entirely optional improvement if you agree. Good work!

Hi @sambiramairelogic, thanks for reviewing! Just left a reply to your comment about the spelling. Interested to know if you would also have done it with ReSpeller. Thanks.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/LocationAttribute.cs
sambiramairelogic: Awesome stuff! 
Hrm that's a shame it wasn't as straight forward as I'd hoped, but it sounds like the Dotsettings file is a better way to go anyway :)

PR #4370, opened 2023-10-23
EES-4392 Attempted fix for SlugNotUnique intermittent UI test failure

mmyoungman:
This attempts to fix a rare intermittent UI test failure we've had with the error "SlugNotUnique".

We believe this is due to the `run_tests.py` script being invoked multiple times at the same time, and two of these runs coincidentally generating the same `runIdentifer` value. If this is correct, this fix should prevent this by adding additional random characters to the end of the `runIdentifier` variable.

### DIFF
--- a/tests/robot-tests/run_tests.py
+++ b/tests/robot-tests/run_tests.py
@@ -10,7 +10,9 @@ import argparse
 import datetime
 import json
 import os
+import random
 import shutil
+import string
 from pathlib import Path
 
 import requests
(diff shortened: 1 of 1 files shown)


tests/robot-tests/run_tests.py
```diff
@@ -342,7 +344,8 @@ def setup_authentication(clear_existing=False):
 
     # NOTE(mark): Tests that alter data only occur on local and dev environments
     if args.env in ["local", "dev"]:
-        runIdentifier = datetime.datetime.utcnow().strftime("%Y%m%d-%H%M%S")
+        randomStr = "".join([random.choice(string.ascii_letters + string.digits) for n in range(12)])
```
ntsim, line 347: 12 chars is quite long, could we do like 6?

ntsim (approved):

PR #4376, opened 2023-10-29
Ees 4467 part 1

duncan-at-hiveit:
This PR:
- Adds a convenience extension method for checking recently created dates.
- Adds tests for AssertExtensions.
- Adds a facility to ignore certain fields when using the AssertDeepEqualsTo extension method.

## Convenience extension method for checking recently created dates

This PR adds a shorthand way to check for a recently created date, and also handles null checking too. Rather than having to write for example:

```
Assert.NotNull(methodologyStatus.Created);
Assert.InRange(DateTime.UtcNow.Subtract(methodologyStatus.Created!.Value).Milliseconds, 0, 1500);
```

we can simply now just write:

```
methodologyStatus.Created.AssertUtcNow(); 
```

The other thing to point out was that the original pattern was incorrect, in that it should have been using "TotalMilliseconds" rather than Milliseconds, as Milliseconds only gives the Milliseconds portion of the DateTime (0 to 999), and hence would *never* had exceeded the "1500" limit that was put in place all over the unit tests. This is now corrected and tested.

## Facility to ignore certain fields when using AssertDeepEqualsTo

This will be used in future work within EES-4467 and serves to help test the cloned Release tree more concisely and more thoroughly during Amendment unit tests, by allowing all but the fields tat we expect to differ to be checked for equality.

This will probably be replaced in the near future with other equality checking approaches, but is very useful right now for the purposes of testing EES-4467.


src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/AssertExtensions.cs
```diff
@@ -10,24 +14,58 @@ public static class AssertExtensions
...
             var compareLogic = new CompareLogic();
+            ignoreProperties?.ForEach(compareLogic.Config.IgnoreProperty);
             var comparison = compareLogic.Compare(actual, expected);
             return comparison.AreEqual;
         }
+
+        /// <summary>
+        /// Assert that the given DateTime is effectively "now", within a given tolerance of milliseconds.
+        /// </summary>
+        public static void AssertUtcNow(this DateTime dateTime, int withinMillis = 1500)
```
ntsim, line 57: Is the tolerance a bit on the high side? Guess we want to avoid false negatives due to timing issues, but being 1.5 seconds out is potentially a bit much? 

Maybe 1 second at most would be more appropriate?

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/AssertExtensionsTests.cs
```diff
@@ -0,0 +1,83 @@
...
+namespace GovUk.Education.ExploreEducationStatistics.Common.Tests.Extensions;
+
+public class AssertExtensionsTests
+{
+    private record Company(string Name, List<Person> Employees);
+
+    private record Person(string Name, int Age);
+
+    [Fact]
+    public void AssertDeepEqualsTo_Simple()
```
ntsim, line 17: Could group methods into nested classes, then we don't need to prefix each method with the method name in a repetitive way e.g.

```cs
public class AssertDeepEqualsToTests
{
  [Fact]
  public void Simple() { ... }
  
  [Fact]
  public void Simple_NotEqual() { ... } 

  // etc...
}
```

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/AssertExtensionsTests.cs
```diff
@@ -0,0 +1,83 @@
...
+
+    [Fact]
+    public void AssertUtcNow_DateTimesOutsideTolerance()
+    {
+        Assert.Throws<InRangeException>(() => DateTime.UtcNow.AddDays(-1501).AssertUtcNow());
+        Assert.Throws<InRangeException>(() => DateTime.UtcNow.AddMilliseconds(-500).AssertUtcNow(withinMillis: 100));
+    }
+
+    [Fact]
+    public void AssertDeepEqualsTo_IgnoreField()
```
ntsim, line 75: Shouldn't this be grouped with the other `AssertDeepEqualsTo` tests?

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/AssertExtensions.cs
duncan-at-hiveit: Fine with this, yeah.  I only kept it at 1500 because that was the example used everywhere around the codebase already.  I've pushed the boat out and set it to 500ms, which should be plenty.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/AssertExtensionsTests.cs
duncan-at-hiveit: Good idea - done!

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/AssertExtensionsTests.cs
duncan-at-hiveit: Done ta

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/AssertExtensions.cs
duncan-at-hiveit: I've also added in additional tests for the nullable DateTime AssertUtcNow override.

mmyoungman (approved):

PR #4381, opened 2023-10-30
EES-4634 fix text area character count

amyb-hiveit:
The update to react hook forms for the publication details forms broke the character count so it no longer updated. This fixes it.

### DIFF
--- a/src/explore-education-statistics-common/src/components/form/FormBaseTextArea.tsx
+++ b/src/explore-education-statistics-common/src/components/form/FormBaseTextArea.tsx
@@ -1,0 +1,88 @@
+import ErrorMessage from '@common/components/ErrorMessage';
+import FormLabel, { FormLabelProps } from '@common/components/form/FormLabel';
+import classNames from 'classnames';
+import React, {
+  ChangeEventHandler,
+  FocusEventHandler,
+  KeyboardEventHandler,
+  MouseEventHandler,
+  ReactNode,
+  Ref,
+  memo,
+} from 'react';
+
+export interface FormTextAreaProps
+  extends Pick<FormLabelProps, 'hideLabel' | 'label'> {
+  className?: string;
+  disabled?: boolean;
+  error?: ReactNode | string;
+  hint?: string;
+  id: string;
+  inputRef?: Ref<HTMLTextAreaElement>;
+  maxLength?: number;
+  name: string;
+  rows?: number;
+  value?: string;
+  onBlur?: FocusEventHandler<HTMLTextAreaElement>;
+  onChange?: ChangeEventHandler<HTMLTextAreaElement>;
+  onClick?: MouseEventHandler<HTMLTextAreaElement>;
+  onKeyPress?: KeyboardEventHandler<HTMLTextAreaElement>;
+}
+
+function FormBaseTextArea({
+  className,
+  disabled,
+  error,
+  hint,
+  id,
+  inputRef,
+  hideLabel,
+  label,
+  maxLength,
+  name,
+  rows = 5,
+  value,
+  onBlur,
+  onChange,
+  onClick,
+  onKeyPress,
+}: FormTextAreaProps) {
+  return (
+    <>
+      <FormLabel id={id} label={label} hideLabel={hideLabel} />
+
+      {hint && (
+        <div id={`${id}-hint`} className="govuk-hint">
+          {hint}
+        </div>
+      )}
+
+      {error && <ErrorMessage id={`${id}-error`}>{error}</ErrorMessage>}
+
+      <textarea
+        aria-describedby={
+          classNames({
+            [`${id}-error`]: !!error,
+            [`${id}-hint`]: !!hint,
+            [`${id}-info`]: !!maxLength,
+          }) || undefined
+        }
+        className={classNames('govuk-textarea', className, {
+          'govuk-textarea--error': !!error,
+        })}
+        disabled={disabled}
+        id={id}
+        name={name}
+        ref={inputRef}
+        onBlur={onBlur}
+        onChange={onChange}
+        onClick={onClick}
+        onKeyPress={onKeyPress}
+        rows={rows}
+        value={value}
+      />
+    </>
+  );
+}
+
+export default memo(FormBaseTextArea);
--- a/src/explore-education-statistics-common/src/components/form/rhf/RHFFormField.tsx
+++ b/src/explore-education-statistics-common/src/components/form/rhf/RHFFormField.tsx
@@ -71,6 +71,7 @@ export default function RHFFormField<
     onBlur: fieldOnBlur,
     onChange: fieldOnChange,
   } = useRegister(name, register);
+
   const { fieldId } = useFormIdContext();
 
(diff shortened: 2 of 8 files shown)


sambiramairelogic (approved):

src/explore-education-statistics-common/src/components/form/rhf/RHFFormField.tsx
```diff
@@ -85,6 +88,7 @@ export default function RHFFormField<
         id={fieldId(name, customId)}
         name={name}
         inputRef={mergeRefs(fieldRef, inputRef)}
+        value={value}
```
ntsim, line 91: Doesn't this leave the component controlled?

Not sure this is great performance-wise as RHF typically updates the value of DOM elements directly via refs

amyb-hiveit: @ntsim as discussed, I've changed this to move getting the value to a new `RHFFormTextArea` component and created new `FormBaseTextArea` and `FormCharacterCount` components that are used by both the RHF and Formik versions.

src/explore-education-statistics-common/src/components/form/FormBaseTextArea.tsx
```diff
@@ -0,0 +1,90 @@
...
+      <textarea
+        aria-describedby={
+          classNames({
+            [`${id}-error`]: !!error,
+            [`${id}-hint`]: !!hint,
+            [`${id}-info`]: !!maxLength,
+          }) || undefined
+        }
+        className={classNames('govuk-textarea', className, {
+          'govuk-js-character-count govuk-textarea--error': hasMaxLengthError,
```
ntsim, line 73: Doesn't seem like we actually need the `govuk-js-character-count` class? If this is the case, we don't really need the `hasMaxLengthError` prop

The current code also doesn't seem right as we need to apply `govuk-textarea--error` if `error` is truthy as well.

src/explore-education-statistics-common/src/components/form/FormTextArea.tsx
```diff
@@ -1,116 +1,36 @@
...
-        onClick={onClick}
-        onKeyPress={onKeyPress}
-        rows={rows}
-        value={value}
-      />
-    </>
-  );
+  const hasMaxLengthError = maxLength
+    ? (value?.length ?? 0) > maxLength
+    : false;
```
ntsim, line 16: Would be cleaner to place this within the conditional so it only gets ran if necessary

src/explore-education-statistics-common/src/components/form/rhf/RHFFormTextArea.tsx
```diff
@@ -0,0 +1,41 @@
...
+  id,
+  maxLength,
+  name,
+  ...props
+}: FormTextAreaProps) {
+  const value = useWatch({ name });
+
+  const hasMaxLengthError = maxLength
+    ? (value?.length ?? 0) > maxLength
+    : false;
```
ntsim, line 19: Similar to other comment, would be better to place this in the conditional to only run it if necessary

src/explore-education-statistics-common/src/components/form/rhf/__tests__/RHFFormTextArea.test.tsx
```diff
@@ -0,0 +1,243 @@
+import RHFFormTextArea from '@common/components/form/rhf/RHFFormTextArea';
+import FormProvider from '@common/components/form/rhf/FormProvider';
+import { render } from '@testing-library/react';
+import noop from 'lodash/noop';
+import React from 'react';
+
+describe('RHFFormTextArea', () => {
+  test('renders correctly with required props', () => {
+    const { container, getByLabelText } = render(
```
ntsim, line 9: We could use the newer `screen` selector instead i.e. `screen.getByLabelText`

Think we can refactor the rest of this suite to use this instead as well.

src/explore-education-statistics-common/src/components/form/FormTextArea.tsx
```diff
@@ -1,116 +1,36 @@
...
-                remaining !== -1 ? 's' : ''
-              } too many`}
-        </div>
+        <FormGroup>
+          <FormBaseTextArea
+            hasMaxLengthError={hasMaxLengthError}
+            id={id}
+            maxLength={maxLength}
+            value={value}
+            {...props}
```
ntsim, line 27: Would place the spread as the first prop to avoid accidentally overriding any explicit props in the future i.e.

```tsx
<FormBaseTextArea
  {...props}
  id={id}
  // etc...
/>
```

src/explore-education-statistics-common/src/components/form/FormTextArea.tsx
```diff
@@ -1,116 +1,36 @@
...
+          />
+          <FormCharacterCount id={id} maxLength={maxLength} value={value} />
+        </FormGroup>
       </div>
     );
   }
 
-  return textArea;
+  return <FormBaseTextArea id={id} value={value} {...props} />;
```
ntsim, line 35: Same as above - spread first

src/explore-education-statistics-common/src/components/form/rhf/RHFFormTextArea.tsx
```diff
@@ -0,0 +1,41 @@
...
+  if (!!maxLength && maxLength > 0) {
+    return (
+      <div className="govuk-character-count">
+        <FormGroup>
+          <FormBaseTextArea
+            hasMaxLengthError={hasMaxLengthError}
+            id={id}
+            maxLength={maxLength}
+            name={name}
+            {...props}
```
ntsim, line 30: Same as `FormTextArea` - would spread first

src/explore-education-statistics-common/src/components/form/rhf/RHFFormTextArea.tsx
```diff
@@ -0,0 +1,41 @@
...
+          />
+        </FormGroup>
+        <FormCharacterCount id={id} maxLength={maxLength} value={value} />
+      </div>
+    );
+  }
+
+  return (
+    <FormBaseTextArea id={id} maxLength={maxLength} name={name} {...props} />
```
ntsim, line 39: Same as above - spread first

ntsim (changes requested):

src/explore-education-statistics-common/src/components/form/FormBaseTextArea.tsx
amyb-hiveit: yeah good spot, that's sorted now.

src/explore-education-statistics-common/src/components/form/FormTextArea.tsx
amyb-hiveit: It's removed now.

src/explore-education-statistics-common/src/components/form/rhf/RHFFormTextArea.tsx
amyb-hiveit: It's removed now.

ntsim (approved):

PR #4382, opened 2023-10-31
EES-4618 frontend redirects

amyb-hiveit:
Adds middleware to redirect methodology and publication pages when their slugs have changed.

- currently this is commented out as the endpoint doesn't exist yet, it'll be uncommented as part of EES-4482
- the middleware is only run on methodology and publication pages
- there's some very basic caching of the list of redirects as suggested on the discussion on https://dfedigital.atlassian.net/browse/EES-3948, there might be a better way to do this - I'm open to suggestions
- it uses fetch for the request as Axios can't be used in middleware
- the hostname will need to be added for the other environments.


src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
```diff
@@ -0,0 +1,62 @@
...
+
+interface StoredRedirects {
+  redirects: Redirects;
+  fetchedAt: number;
+}
+
+let storedRedirects: StoredRedirects | undefined;
+
+const redirectPaths = {
+  methodologyRedirects: '/methodology',
```
duncan-at-hiveit, line 16: Given the owning object is called `redirectPaths`, these attributes can probably just be called `methodologies` and `publications`.

src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
```diff
@@ -0,0 +1,62 @@
...
+  publicationRedirects: '/find-statistics',
+};
+
+const cacheTime = 300000;
```
duncan-at-hiveit, line 20: Is this 5 mins?  Might be worth just adding a comment to that effect.  Might potentially be a little high as well - but Cam would probably be best placed to make the call on that, if this didn't come from him.

src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export default async function redirectPages(request: NextRequest) {
+  const shouldRefetch =
```
duncan-at-hiveit, line 23: Do you actually need that first `|| (storedRedirects?.fetchedAt` check if storedRedirects is defined - is there a way for fetchedAt to be falsy at that point?

src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
```diff
@@ -0,0 +1,62 @@
...
+      storedRedirects.fetchedAt + cacheTime < Date.now());
+
+  if (shouldRefetch) {
+    storedRedirects = {
+      redirects: await redirectService.list(),
+      fetchedAt: Date.now(),
+    };
+  }
+
+  let redirectUrl: string | undefined;
```
duncan-at-hiveit, line 35: Rather than neding this `let`, could you instead just replace the assignment of `redirectUrl` with:

```
const redirectUrl = pathNameSections
          .map(section =>
            section === rewriteRule?.fromSlug ? rewriteRule?.toSlug : section,
          )
          .join('/');
return NextResponse.redirect(new URL(redirectUrl, request.url))
```

No biggie either way, but I've always had it drummed into me to try to avoid lets where possible!

src/explore-education-statistics-frontend/src/middleware.ts
```diff
@@ -0,0 +1,18 @@
...
+  // Uncomment this and remove `return NextResponse.next();` to enable redirects.
+  // Also remove the line disabling linting no-unused-vars.
+  // return redirectPages(request);
+
+  return NextResponse.next();
+}
+
+// Restrict to release and methodology pages.
+export const config = {
+  matcher: ['/find-statistics/:path/:path*', '/methodology/:path*'],
```
duncan-at-hiveit, line 17: I'm guessing this could make use of the `redirectPaths` as defined in redirectPages.ts so we only need to change in one place if we're reconfiguring the paths?

src/explore-education-statistics-frontend/src/services/redirectService.ts
```diff
@@ -0,0 +1,21 @@
+export interface Redirects {
+  methodologyRedirects: Redirect[];
```
duncan-at-hiveit, line 2: Again, given that these attributes are within something called `Redirects`, can they just be plain `methodologies` and `publications`?  If you think the readability's better the way it is though, I've not got a strong opinion on it.

src/explore-education-statistics-frontend/.env
```diff
@@ -4,3 +4,4 @@ NOTIFICATION_API_BASE_URL=http://localhost:7073/api
 GA_TRACKING_ID=
 PUBLIC_URL=http://localhost:3000/
 APP_ENV=Local
+HOSTNAME=localhost
```
duncan-at-hiveit, line 7: Does this have any impact when deploying to Azure?  Do we need to set HOSTNAME in the real environments?

src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
mmyoungman: This is my naming inherited from the backend. I'll change the name in the backend to `methodologies`.

src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
mmyoungman, line 20: I'm wondering if this delay, even if reduced, could cause problems in UI test runs. Can we make this easily changable or different for local + dev environments? It might make sense to always have it shorter on our local environments anyway.

src/explore-education-statistics-frontend/.env
ntsim: Pretty sure we don't need this as we can use `URL` on the `PUBLIC_URL` variable i.e.

```ts
const url = new URL(process.env.PUBLIC_URL);

url.hostname // outputs localhost if PUBLIC_URL is http://localhost:3000/
```

src/explore-education-statistics-frontend/src/middleware.ts
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import redirectPages from '@frontend/middleware/pages/redirectPages';
```
ntsim, line 2: We don't need this currently, so could remove it as we're just bundling more code than we need

src/explore-education-statistics-frontend/src/middleware.ts
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
ntsim, line 1: Would suggest disabling the specific line rather than disabling at file level

src/explore-education-statistics-frontend/src/middleware/pages/__tests__/redirectPages.test.ts
```diff
@@ -0,0 +1,275 @@
...
+    ],
+    publicationRedirects: [
+      { fromSlug: 'original-slug-3', toSlug: 'updated-slug-3' },
+      { fromSlug: 'original-slug-4', toSlug: 'updated-slug-4' },
+    ],
+  };
+
+  afterEach(() => {
+    redirectSpy.mockReset();
+  });
```
ntsim, line 29: Don't think we need this. The project's been configured to automatically reset mocks after each test case

src/explore-education-statistics-frontend/src/middleware/pages/__tests__/redirectPages.test.ts
```diff
@@ -0,0 +1,275 @@
...
+
+  test('does not re-request the list of redirects once it has been fetched', async () => {
+    redirectService.list.mockResolvedValue(testRedirects);
+    const req = new NextRequest(
+      new Request('https://my-env/methodology/original-slug'),
+      {},
```
ntsim, line 35: Seems like this arg is optional and can be removed. Same thing for similar cases in this test suite

src/explore-education-statistics-frontend/src/middleware/pages/__tests__/redirectPages.test.ts
```diff
@@ -0,0 +1,275 @@
...
+      redirectService.list.mockResolvedValue(testRedirects);
+      const req = new NextRequest(
+        new Request('https://my-env/methodology/original-slug-1'),
+        {},
+      );
+      await redirectPages(req);
+
+      expect(redirectSpy).toHaveBeenCalledTimes(1);
+      expect(redirectSpy).toHaveBeenCalledWith(
+        new URL('/methodology/updated-slug-1', req.url),
```
ntsim, line 61: To make this more explicit and clearer, could we use a fully qualified URL instead of the `req.url` as a base? i.e.

```ts
new URL('https://my-env/methodology/updated-slug-1')
```

Same for other similar cases in this test suite

src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
ntsim: We can also format the number a little nicer too i.e. `300_000`

src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
```diff
@@ -0,0 +1,62 @@
...
+} from '@frontend/services/redirectService';
+import type { NextRequest } from 'next/server';
+import { NextResponse } from 'next/server';
+
+interface StoredRedirects {
+  redirects: Redirects;
+  fetchedAt: number;
+}
+
+let storedRedirects: StoredRedirects | undefined;
```
ntsim, line 13: Nitpick, but maybe `cachedRedirects` might be better (and `CachedRedirects` for the type)

src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
```diff
@@ -0,0 +1,62 @@
...
+      redirects: await redirectService.list(),
+      fetchedAt: Date.now(),
+    };
+  }
+
+  let redirectUrl: string | undefined;
+
+  Object.keys(redirectPaths).forEach(key => {
+    if (
+      request.nextUrl.pathname.startsWith(redirectPaths[key as RedirectType])
```
ntsim, line 39: Could extract `key` to a `redirectType` variable to avoid doing `as RedirectType` twice

src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
```diff
@@ -0,0 +1,62 @@
...
+    ) {
+      const pathNameSections = request.nextUrl.pathname.split('/');
```
ntsim, line 41: Not a biggy, but 'segment' is usually more appropriate for URLs i.e. `pathSegments`

src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
```diff
@@ -0,0 +1,62 @@
...
+
+      const rewriteRule = storedRedirects?.redirects[key as RedirectType]?.find(
+        ({ fromSlug }) => pathNameSections[2] === fromSlug,
+      );
+
+      if (rewriteRule) {
+        redirectUrl = pathNameSections
+          .map(section =>
```
ntsim, line 49: As above - could be `segment`

src/explore-education-statistics-frontend/test/setupTests.js
```diff
@@ -13,3 +13,6 @@ if (typeof window !== 'undefined') {
   // fetch polyfill for making API calls.
   require('cross-fetch');
 }
+
+global.Request = jest.requireActual('node-fetch').Request;
+global.Response = jest.requireActual('node-fetch').Response;
```
ntsim, line 18: Not sure we need this as Fetch API landed in Node in v18 (which we're currently on).

src/explore-education-statistics-frontend/src/middleware.ts
amyb-hiveit: Annoyingly we can't - "The matcher values need to be constants so they can be statically analyzed at build-time. Dynamic values such as variables will be ignored." https://nextjs.org/docs/app/building-your-application/routing/middleware#matcher

src/explore-education-statistics-frontend/test/setupTests.js
amyb-hiveit: Without it the tests fail to run because of `ReferenceError: Request is not defined`. I've checked that my local node version is 18.

src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
amyb-hiveit: Changed to `methodologies` and `publications` following the back end change.

src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
amyb-hiveit: That didn't quire work but I've refactored it to use a reduce instead so that a `let` isn't needed.

src/explore-education-statistics-frontend/.env
amyb-hiveit: Oh yeah If course it's not needed. Removed now, thanks.

src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
amyb-hiveit: Good point, I've set it to zero for the test environment and added an optional param so that it can be overridden in the unit test to test the caching. It's only 1 min on other envs now so should be ok for dev, but we can always adjust it later if it causes problems.

src/explore-education-statistics-frontend/test/setupTests.js
ntsim: Hmm weird, the API documentation seems to suggest it should be there! Guess this is fine anyway

src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
duncan-at-hiveit: Nice one, thanks!

src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
duncan-at-hiveit: Haha appreciated - hopefully worth the effort :)

src/explore-education-statistics-frontend/src/middleware.ts
duncan-at-hiveit: Cool ta!

duncan-at-hiveit: Past Amy will hopefully be heaped in praise!

duncan-at-hiveit (approved):

PR #4383, opened 2023-10-31
EES-4482 Introduce Methodology Redirects

mmyoungman:
This PR introduces methodology redirects. This allows a methodology to change a slug for a live methodology, with users visiting URLs with the old slug being redirected to the latest slug.

This piece of work was spawned off EES-3789. When attempting to introduce publication redirects, it became apparent that methodology redirects would have to be done simultaneously, and then it further became apparent that we could do methodology redirects on their own, as a stepping stone to introducing publication redirects.

### Notes

- The frontend work for this work will be done separately in EES-4618. This PR requires the frontend work, hence the Draft PR and `do not merge` label. I'll change this once I've rebased off the EES-4618 work.

- In an attempt to keep this work simple, we have implemented these changes to mostly avoid data changes during publishing. To achieve this, we store `MethodologyVersionId`s in redirects instead of `MethodologyId`s. This allows a redirect to automatically become live when it's associated methodology version is published.

- Despite live publication's still not being able to change their slugs, this PR includes many changes that aim in this direction. Most notably in `PublicationTitleOrSlugChanged`, which should already be able to handle publication slug changes for a live methodology, even if that functionality isn't currently being used.

- There are various other fixes that will happen after this work is deployed. These are covered by EES-4613 and EES-4627.


src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/MethodologyVersionRepository.cs
```diff
@@ -87,74 +88,39 @@ await _contentDbContext.Entry(methodology)
             return await _contentDbContext
                 .MethodologyVersions
                 .Where(mv =>
-                    mv.Methodology.Slug == slug
-                    && mv.Methodology.LatestPublishedVersionId == mv.Id)
+                    mv.Methodology.LatestPublishedVersionId == mv.Id
```
duncan-at-hiveit, line 91: The alternative, which we've used before for getting around restrictions with JSON columns is to do something like:

```
    return (await _contentDbContext
        .MethodologyVersions
        .Where(mv => mv.Methodology.LatestPublishedVersionId == mv.Id)
        // Pull into memory
        .ToListAsync())
        .SingleOrDefault(mv => slug == mv.Slug);
```

Not sure if there's an argument for one or other approach here tbh - if mv.Slug had a more complex logic then perhaps a stronger argument for this method, but hey!

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Methodology.cs
```diff
@@ -13,11 +13,12 @@ public class Methodology
 
         public List<PublicationMethodology> Publications { get; set; } = new();
 
-        [Required]
-        public string OwningPublicationTitle { get; set; }
+        [Required] public string OwningPublicationTitle { get; set; } = null!;
```
duncan-at-hiveit, line 16: Which formatting are we favouring?  Inline attributes or the other?

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Fixtures/MethodologyGeneratorExtensions.cs
```diff
@@ -17,7 +17,7 @@ public static Generator<Methodology> WithDefaults(this Generator<Methodology> ge
     public static InstanceSetters<Methodology> SetDefaults(this InstanceSetters<Methodology> setters)
         => setters
             .SetDefault(p => p.Id)
-            .SetDefault(p => p.Slug);
+            .SetDefault(p => p.OwningPublicationSlug);
```
duncan-at-hiveit, line 20: Should we add .SetDefault(p => p.OwningPublicationTitle) here as well?

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Repository/MethodologyVersionRepositoryTests.cs
```diff
@@ -854,6 +854,152 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
...
+
+                var result = await service
+                    .GetLatestPublishedVersionBySlug("slug");
+
+                VerifyAllMocks(methodologyRepository);
+
+                Assert.NotNull(result);
+                Assert.Equal(latestPublishedVersionId, result.Id);
+                Assert.Null(result.AlternativeSlug);
+                // doesn't return result.Methodology, so cannot check result.Slug/result.Methodology.OwningPublicationSlug
```
duncan-at-hiveit, line 954: Typically in this case where we get back just a view model or a partially loaded entity as a result, we'd do one more separate contentDbContext block below, like:

```
await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId))
{
    // retrieve same entity as "result" above direct from the dbContext. 
    // do checks for result.Slug/result.Methodology.OwningPublicationSlug here
}    
```

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Repository/MethodologyVersionRepositoryTests.cs
```diff
@@ -854,6 +854,152 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
...
+            }
+        }
+
+        [Fact]
+        public async Task GetLatestPublishedVersionBySlug_UnpublishedMethodology()
+        {
+            var methodology = new Methodology
+            {
+                LatestPublishedVersionId = null,
+                OwningPublicationSlug = "slug",
```
duncan-at-hiveit, line 964: All of these being "slug", I'm not 100% what we're proving here :)  Do any of these need changing to provide more proof that the mechanism is working or are you happy it's OK as it is?

duncan-at-hiveit: Actually I see the result is null, but I'm guessing that in real life this data setup is not possible as we'd need to publish the first version before creating the second version?

src/GovUk.Education.ExploreEducationStatistics.Content.Services/MethodologyService.cs
```diff
@@ -22,29 +23,34 @@ public class MethodologyService : IMethodologyService
...
         public async Task<Either<ActionResult, MethodologyVersionViewModel>> GetLatestMethodologyBySlug(string slug)
         {
             return await _persistenceHelper
                 .CheckEntityExists<Methodology>(
                     query => query
-                        .Where(mp => mp.Slug == slug))
+                        .Include(m => m.LatestPublishedVersion)
+                        .Where(m =>
+                            m.LatestPublishedVersion != null
+                            && slug == (m.LatestPublishedVersion.AlternativeSlug ?? m.OwningPublicationSlug))) // slug == mv.Slug doesn't translate
```
duncan-at-hiveit, line 49: Same as per my other comment regarding pulling into memory.  Either that, or maybe transpose your other comment here?

src/GovUk.Education.ExploreEducationStatistics.Content.Services/MethodologyService.cs
```diff
@@ -125,5 +131,82 @@ private async Task<List<MethodologyVersionSummaryViewModel>> BuildMethodologiesF
...
+
+            var ownedMethodology = await _contentDbContext
+                .PublicationMethodologies
+                .Include(pm => pm.Methodology.LatestPublishedVersion)
+                .Include(pm => pm.Methodology.Versions)
+                .Where(pm => pm.PublicationId == publicationId && pm.Owner)
+                .Select(pm => pm.Methodology)
+                .SingleOrDefaultAsync();
+
+            if (ownedMethodology == null)
```
duncan-at-hiveit, line 151: I can't tell if this is actually a bad request or not at this point?  Do you think this is a valid method call if the supplied parametes produce no owned Methodology, or should we opt for either an ArgumentException or the returning of an Either<ActionResult, Unit> and do a `.SingleOrDefaultAsync().OrNotFound()`?

src/GovUk.Education.ExploreEducationStatistics.Content.Services/MethodologyService.cs
```diff
@@ -125,5 +131,82 @@ private async Task<List<MethodologyVersionSummaryViewModel>> BuildMethodologiesF
...
+
+            ownedMethodology.OwningPublicationTitle = updatedTitle;
+            ownedMethodology.OwningPublicationSlug = updatedSlug;
+
+            _contentDbContext.Methodologies.Update(ownedMethodology);
+
+            if (slugChanged)
+            {
+                // A redirect is only needed for the LatestPublishedVersion.
+                // Unpublished methodology's don't need a redirect - they're not live.
```
duncan-at-hiveit, line 164: methodology's => methodologies

src/GovUk.Education.ExploreEducationStatistics.Content.Services/MethodologyService.cs
```diff
@@ -125,5 +131,82 @@ private async Task<List<MethodologyVersionSummaryViewModel>> BuildMethodologiesF
...
+                // An unpublished amendment doesn't need a redirect because:
+                // - if it uses OwningPublicationSlug, it is covered by the LatestPublishedVersion redirect created here
+                // - if it uses AlternativeSlug, a redirect would have been created at the time the AlternativeSlug
+                //   was set (and that redirect will become active when that version is published).
+                if (ownedMethodology.LatestPublishedVersion is { AlternativeSlug: null }
```
duncan-at-hiveit, line 169: Nice pattern matching here :)  I should use it more!

src/GovUk.Education.ExploreEducationStatistics.Content.Services/MethodologyService.cs
```diff
@@ -125,5 +131,82 @@ private async Task<List<MethodologyVersionSummaryViewModel>> BuildMethodologiesF
...
+                    _contentDbContext.MethodologyRedirects.Add(redirect);
+
+                    // It's possible we now have two redirects from the same slug. This happens if:
+                    // - An unpublished amendment sets an AlternativeSlug
+                    // - Then the OwningPublicationSlug changes when the LatestPublishedVersion is
+                    //   inheriting it.
+                    // We must delete the unpublished amendment's redirect, as otherwise if the unpublished version
+                    // changes its slug again, no redirect will be created for the methodology's new slug. (If updating
+                    // an unpublished version's slug multiple times, a redirect is only created the first time - see
+                    // MethodologyUpdate code). If this doesn't make any sense, see the associated unit test
```
duncan-at-hiveit, line 190: Instead of `If this doesn't make any sense, see the associated unit test`, I would instead write `See MethodologyServiceTests.PublicationTitleOrSlugChanged_NastyScenario() for an example.`

src/GovUk.Education.ExploreEducationStatistics.Content.Services/RedirectsService.cs
```diff
@@ -0,0 +1,65 @@
...
+    private readonly ContentDbContext _contentDbContext;
+
+    public RedirectsService(ContentDbContext contentDbContext)
+    {
+        _contentDbContext = contentDbContext;
+    }
+
+    public async Task<Either<ActionResult, RedirectsViewModel>> List()
+    {
+        // A redirect for a methodologyVersion that isn't published shouldn't appear in the list. A redirect becomes
```
duncan-at-hiveit, line 25: Uppercase Ms are needed in these comments :)

src/GovUk.Education.ExploreEducationStatistics.Content.Services/RedirectsService.cs
```diff
@@ -0,0 +1,65 @@
...
+        // active once the associated methodologyVersion is published. It remains active if subsequent methodology
+        // amendment's are published. We establish this by checking against each methodologyVersion's Version.
```
duncan-at-hiveit, line 27: No need for apostrophe in `amendment's`

src/GovUk.Education.ExploreEducationStatistics.Content.Services/RedirectsService.cs
```diff
@@ -0,0 +1,65 @@
...
+        var methodologyRedirects = await _contentDbContext.MethodologyRedirects
+            .Where(mr => mr.MethodologyVersion.Methodology.LatestPublishedVersion != null
+                         && mr.MethodologyVersion.Methodology.LatestPublishedVersion.Version >=
+                         mr.MethodologyVersion.Version
+            )
+            .Select(mr => new
+            {
+                RedirectSlug = mr.Slug,
+                // MethodologyVersion.Slug cannot be translated into SQL, so we do this...
+                LatestPublishedSlug = mr.MethodologyVersion.Methodology.LatestPublishedVersion!.AlternativeSlug
```
duncan-at-hiveit, line 37: Same as previous comments here

src/GovUk.Education.ExploreEducationStatistics.Content.Services/RedirectsService.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                if (mr.RedirectSlug == mr.LatestPublishedSlug)
+                {
+                    // It is possible for a redirect to point to the currently active slug. This can happen as a
+                    // methodology can change to a slug even if there is a redirect for that slug,
+                    // *if* the redirect is for the same methodology - i.e. the methodology's slug is changing
+                    // back to a previous slug it used. But the slug may not change back immediately, if the change
+                    // is on an unpublished amendment, so we cannot remove the redirect immediately when changing the
+                    // slug back; redirect needs to be active until the amendment is published.
+                    return null;
```
duncan-at-hiveit, line 53: Should we instead be maintaining a "Published" DateTime on MethodologyRedirects itself that is set and unset by the Publisher to make it clearer which redirects are active at any given time, or are you happy with the way that we have it here?

src/GovUk.Education.ExploreEducationStatistics.Content.Services/RedirectsService.cs
```diff
@@ -0,0 +1,65 @@
...
+                }
+
+                return new MethodologyRedirectViewModel(
```
duncan-at-hiveit, line 56: Playing devil's advocate here as I don't really mind one way or the other, but you could probably do this in AutoMapper if you wanted.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/RedirectsServiceTests.cs
```diff
@@ -0,0 +1,403 @@
...
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Services.Tests
+{
+    public class RedirectsServiceTests
+    {
+        [Fact]
+        public async Task List_MethodologyRedirect()
+        {
+            var latestPublishedVersionId = Guid.NewGuid();
+            var methodology = new Methodology
```
duncan-at-hiveit, line 19: Might be good to get some practice in using the DataFixtures soon.  I don't mind if that's the *next* PR though :)

src/explore-education-statistics-admin/src/pages/methodology/components/MethodologySummaryForm.tsx
```diff
@@ -19,7 +19,8 @@ const errorMappings = [
   mapFieldErrors<MethodologySummaryFormValues>({
     target: 'title',
     messages: {
-      SlugNotUnique: 'Choose a unique title',
+      SlugNotUnique: 'Used by other methodology. Choose a unique title',
```
duncan-at-hiveit, line 22: Have these been run my Cam?  Not sure "Used by methodology redirect" would mean much to an Analyst, although it is definitely more useful than just a plain "Choose a unique title" I guess :)

tests/robot-tests/tests/admin_and_public/bau/publish_methodology.robot
```diff
@@ -415,7 +415,9 @@ Verify that the amended methodology is visible on the public methodologies page
     user scrolls down    400
     user checks page contains link with text and url
     ...    ${PUBLICATION_NAME} - Amended methodology
-    ...    ${PUBLIC_METHODOLOGY_URL_ENDING}
+    ...    ${PUBLIC_METHODOLOGY_URL_ENDING}-amended-methodology    # Slug has changed
+
+# @MarkFix visit old slug and get redirected to methodology
```
duncan-at-hiveit, line 420: Does this need addressing once Amy's frontend work is available?  Does this need a ticket ref?

duncan-at-hiveit (commented):
A fine piece of work here!  I've just added a few comments and suggestions, ta.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Methodologies/MethodologyService.cs
```diff
@@ -476,7 +504,12 @@ await _context.Entry(loadedMethodologyVersion.ScheduledWithRelease)
                 .OnSuccess(() => _methodologyImageService.DeleteAll(methodologyVersion.Id, forceDelete))
                 .OnSuccessVoid(async () =>
                 {
+                    _context.MethodologyRedirects.RemoveRange(
```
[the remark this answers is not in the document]
mmyoungman: The unit tests pass if I remove it. Generally, I've tended not to rely on the cascades and always remove things manually regardless. I'll remove it.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/MethodologyVersionRepository.cs
mmyoungman: I guess this alternative isn't going to be as efficient, but then if someone changes `Slug` in the future it'll be clear that this also needs to be changed.

I think I'll leave it - if we pull it into memory we need the DB, it seems a bit wasteful to return every `LatestPublishedVersion` including all fields for those methodology versions.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Methodology.cs
mmyoungman: I imagine I only changed this because Rider did it for me. I've changed them all to be inline just so they match - doesn't matter to me either way.

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Repository/MethodologyVersionRepositoryTests.cs
mmyoungman: I can add it, but I don't think it's worth it in this instance because `GetLatestPublishedVersionBySlug` doesn't make any DB changes, so we'd just be checking the test data. I'll leave it for now.

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Repository/MethodologyVersionRepositoryTests.cs
mmyoungman: I've updated the test data so it's more realistic

src/GovUk.Education.ExploreEducationStatistics.Content.Services/MethodologyService.cs
mmyoungman: I left the other instances alone so I'll leave this one alone too - prioritising avoiding having to pull every latest published methodology version out of the database.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/MethodologyService.cs
mmyoungman: I think it is correct. `PublicationTitleOrSlugChanged` is called when the publication's title or slug is changed. It's valid that the publication may not own a methodology.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/MethodologyService.cs
mmyoungman: I can't get that right at the moment it seems...

src/GovUk.Education.ExploreEducationStatistics.Content.Services/MethodologyService.cs
mmyoungman: I only used it because Rider suggested it, to be honest :smile:

src/GovUk.Education.ExploreEducationStatistics.Content.Services/RedirectsService.cs
mmyoungman: I left the other instances alone so I'll leave this one alone too - prioritising avoiding having to pull every latest published methodology version out of the database.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/RedirectsService.cs
mmyoungman: I think I'm happy without it. My approach was to attempt to avoid having to make changes during publishing wherever possible. So the logic for whether a methodology redirect is active lives solely in `RedirectsService#List`. My general thought was that makes the feature more contained, as you just have to create the redirects, not worry about when they should be published. If the publisher is refactored, you only need to worry about updating the cache.

I'm inclined to think that was the simpler approach, but honestly not sure if that is right. One of those instances where it would be nice to see both implemented to more easily compare them.

Not sure how much work it would be to switch. I don't think it would be a lot of code - most of the time would probably be spent making sure I'd not broken anything. If you feel it would be worth it, I'm happy to give it a go.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/RedirectsService.cs
mmyoungman: I'll leave it - I'm not a fan of AutoMapper generally.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/RedirectsServiceTests.cs
mmyoungman: I'll use it next PR. I've been mostly avoiding it cause I feel like it'll take me a while to get my head round it, and I'd prefer to remain productive. But I'm going to have to learn it at some point.

src/explore-education-statistics-admin/src/pages/methodology/components/MethodologySummaryForm.tsx
mmyoungman: I haven't - I'll do that.

My main thought in including the "Used by methodology redirect" error was mainly for our benefit if we're debugging an issue.

tests/robot-tests/tests/admin_and_public/bau/publish_methodology.robot
mmyoungman: My plan is to do this after Amy's PR gets merged and I can rebase onto it. There will be a few other changes when I do that as well.

mmyoungman: This UI test has been written and commited now. So when rebasing on Amy's branch we'll just need to enable her work per the `middleware.ts` comment.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/RedirectsService.cs
duncan-at-hiveit: Haha ok!

tests/robot-tests/tests/admin_and_public/bau/publish_methodology.robot
```diff
@@ -415,7 +415,12 @@ Verify that the amended methodology is visible on the public methodologies page
     user scrolls down    400
     user checks page contains link with text and url
     ...    ${PUBLICATION_NAME} - Amended methodology
-    ...    ${PUBLIC_METHODOLOGY_URL_ENDING}
+    ...    ${PUBLIC_METHODOLOGY_URL_ENDING}-amended-methodology    # Slug has changed
+
+Validate methodology redirect works
```
amyb-hiveit, line 420: @mmyoungman this test will need to wait for the redirect cache to expire. It's set to 10 seconds on local and dev so we might want to add WAIT_REDIRECT_CACHE_EXPIRY variable and a keyword for it.

duncan-at-hiveit (approved):

duncan-at-hiveit (approved):

PR #4384, opened 2023-10-31
EES-4638 fix server error handling on methodology form

amyb-hiveit:
Fixes the server error handling on the methodology summary form. Previously if you tried to change the title to one that is already taken you'd get an error page instead of the standard form error. I've converted the form to React Hook Form as part of this work.

err

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/methodology/components/MethodologySummaryForm.tsx
+++ b/src/explore-education-statistics-admin/src/pages/methodology/components/MethodologySummaryForm.tsx
@@ -1,22 +1,22 @@
 import Button from '@common/components/Button';
 import ButtonGroup from '@common/components/ButtonGroup';
 import ButtonText from '@common/components/ButtonText';
-import Form from '@common/components/form/Form';
-import FormFieldTextInput from '@common/components/form/FormFieldTextInput';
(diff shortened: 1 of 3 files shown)


src/explore-education-statistics-admin/src/pages/methodology/components/MethodologySummaryForm.tsx
```diff
@@ -25,81 +25,77 @@ const errorMappings = [
...
 interface Props {
-  id: string;
-  initialValues?: MethodologySummaryFormValues;
   defaultTitle: string;
+  id: string;
+  initialValues: FormValues;
   submitText: string;
   onCancel: () => void;
-  onSubmit: (values: MethodologySummaryFormValues) => void;
+  onSubmit: (title: string) => Promise<void>;
```
ntsim, line 33: Could make this a little more flexible by using `Promise<void> | void`. This way it doesn't need to be async callback and `noop` would work in the tests

amyb-hiveit: Cool, I've changed it now.

duncan-at-hiveit (approved):

PR #4388, opened 2023-11-01
EES-4643 fixing Prod UI test failures in 'general_public' suite

N-moh:
A few UI tests failed due to a mismatch of data from the live environment compared to the seed data for the UI tests. Additionally, a few xpaths are not valid anymore so they got replaced with ID selectors instead.

Screenshot (564)

### DIFF
--- a/tests/robot-tests/tests/general_public/notifications_absence.robot
+++ b/tests/robot-tests/tests/general_public/notifications_absence.robot
@@ -35,5 +35,6 @@ Sign up for email alerts
     user clicks button    Subscribe
 
     # EES-1265
-    user waits until h1 is visible    Subscribed    %{WAIT_LONG}
+    #user waits until h1 is visible    Subscribed    %{WAIT_LONG}
+    user waits until page contains    Subscribed    %{WAIT_SMALL}
     user checks page contains    Thank you. Check your email to confirm your subscription.
--- a/tests/robot-tests/tests/general_public/notifications_absence.robot
+++ b/tests/robot-tests/tests/general_public/notifications_absence.robot
@@ -35,6 +35,5 @@ Sign up for email alerts
     user clicks button    Subscribe
 
     # EES-1265
-    #user waits until h1 is visible    Subscribed    %{WAIT_LONG}
-    user waits until page contains    Subscribed    %{WAIT_SMALL}
+    user waits until page contains    Subscribed    %{WAIT_LONG}
     user checks page contains    Thank you. Check your email to confirm your subscription.
--- a/tests/robot-tests/tests/general_public/permalink_prod_1.robot
+++ b/tests/robot-tests/tests/general_public/permalink_prod_1.robot
@@ -43,7 +43,7 @@ Validate table
     user checks table cell contains    1    3    148,820
 
 Validate footnotes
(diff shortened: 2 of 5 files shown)


tests/robot-tests/tests/general_public/permalink_prod_1.robot
```diff
@@ -43,10 +43,10 @@ Validate table
     user checks table cell contains    1    3    148,820
 
 Validate footnotes
-    user checks page contains element
-    ...    xpath://h3[text()="Footnotes"]/../ol/li[text()="State-funded secondary schools include city technology colleges and all secondary academies, including all-through academies and free schools."]
-    user checks page contains element
-    ...    xpath://h3[text()="Footnotes"]/../ol/li[text()="x - 1 or 2 pupils, or a percentage based on 1 or 2."]
+    user checks page contains element    //h3[@class="govuk-heading-m"]
```
mmyoungman, line 46: I think a better selector here would be `testid:footnotes`. The same change can be made in`permalink_prod_2` and `permalink_prod_3`.

While `//h3[@class="govuk-heading-m"]` it is unique on the page, there are other instances of tags with the class "govuk-heading-m", and if one of those turned into a h3 tag, then this test would always pass. 

Also, if additional class attribute were added to the h3 tag - i.e. `<h3 class="govuk-heading-m something-else">` - then this would also fail. If you want to selecting on a class, it's normally easier to use a CSS selector, like `css:h3.govuk-heading-m`. This is better because if another class was added, the CSS selector would still select work, while an xpath selector just compares the whole class string. 

You could do `//h3[contains(@class, 'govuk-heading-m')]`, but that has the issue that if someone adds a new tag to the service that looks like `<h3 class="govuk-heading-m-bold">` the selector would also find that element. So if you want to select on class, CSS is generally more robust.

mmyoungman (changes requested):

duncan-at-hiveit:
Just to say, I've run this against local and Dev, and all working well!  Happy to approve and merge in once Mark's comment is looked at.

tests/robot-tests/tests/general_public/notifications_absence.robot
```diff
@@ -35,5 +35,6 @@ Sign up for email alerts
     user clicks button    Subscribe
 
     # EES-1265
-    user waits until h1 is visible    Subscribed    %{WAIT_LONG}
+    #user waits until h1 is visible    Subscribed    %{WAIT_LONG}
```
mmyoungman, line 38: I think this commented out line can be removed?

tests/robot-tests/tests/general_public/notifications_absence.robot
```diff
@@ -35,5 +35,6 @@ Sign up for email alerts
...
+    user waits until page contains    Subscribed    %{WAIT_SMALL}
```
mmyoungman, line 39: I think we need to keep the `WAIT_LONG` here. The ticket EES-1265 suggests that if nobody has used notify for a while, the response can be very slow, so to ensure this tests passes if the response from notify does take a long time, we need a longer wait.

mmyoungman (changes requested):

mmyoungman (approved):

PR #4396, opened 2023-11-06
EES-4623 add OSR line to footer

amyb-hiveit:
Adds a line about the OSR to the footer.

footer

### DIFF
--- a/src/explore-education-statistics-admin/src/components/PageFooter.tsx
+++ b/src/explore-education-statistics-admin/src/components/PageFooter.tsx
@@ -41,6 +41,16 @@ const PageFooter = ({ wide }: Props) => {
                 </Link>
               </li>
             </ul>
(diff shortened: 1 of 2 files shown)


jack-hive (approved):
Looks good! :)

Do we add robot tests for this sort of thing at all? To check that the link is there and that it does what you expect? Or do we only tend to add robot tests for more meaty changes? (This is more just me asking for my own knowledge of our processes)

amyb-hiveit:
> Looks good! :)
> 
> Do we add robot tests for this sort of thing at all? To check that the link is there and that it does what you expect? Or do we only tend to add robot tests for more meaty changes? (This is more just me asking for my own knowledge of our processes)

There aren't any definitive rules, though that's something the styleguide meetings may address. My opinion is that robot tests are for meatier things related to the functionality of the site and it would be overkill to check all the static content on the site with them.

PR #4400, opened 2023-11-08
EES-4584 trim whitespace from text inputs

amyb-hiveit:
Automatically trims whitespace in the React Hook Form and Formik versions of the `FormFieldTextInput` and `FormFieldTextArea` components .

I've ended up with different implementations for RHF and Formik which isn't ideal but I think the RHF version is better and we'll hopefully be rid of Formik soonish.

RHF:
- uses `setValueAs` on `register` to trim the input, this runs before validation when the form is submitted, see https://react-hook-form.com/docs/useform/register
- it's controlled by an optional `trimInput` param 
- `RHFFormFieldTextInput` and `RHFFormFieldTextArea` set `trimInput` to true by default but it can be overriden

Formik:
- trims the inputs on blur
- it's controlled by an optional `trimInput` param 
- `FormFieldTextInput` and `FormFieldTextArea` set `trimInput` to true by default but it can be overriden
- for inputs it also trims (by blurring the field) when you press Enter, without this the trim won't happen if you submit the form with Enter. This doesn't happen for textareas as Enter adds a new line rather than submitting the form.

Also:
I've added some manual trimming for `FormSearchBar` as it doesn't use the standard `FormFieldX` components.

### DIFF
--- a/src/explore-education-statistics-common/src/components/form/FormField.tsx
+++ b/src/explore-education-statistics-common/src/components/form/FormField.tsx
@@ -5,6 +5,7 @@ import {
   FieldInputProps,
   FieldMetaProps,
   useField,
+  useFormikContext,
 } from 'formik';
 import React, {
   ChangeEvent,
@@ -28,6 +29,8 @@ export interface FormFieldProps<FormValues = unknown> {
   formGroupClass?: string;
   name: FormValues extends Record<string, unknown> ? keyof FormValues : string;
   showError?: boolean;
+  trimInput?: boolean;
+  trimInputOnEnter?: boolean;
 }
 
 interface FormFieldInputProps<Value> extends FieldInputProps<Value> {
@@ -57,6 +60,8 @@ function FormField<Value, Props = Record<string, unknown>>({
   id: customId,
   name,
   showError = true,
+  trimInput = false,
+  trimInputOnEnter = false,
   type,
   ...props
 }: FormFieldProps & InternalFormFieldProps<Props, Value>) {
@@ -67,6 +72,8 @@ function FormField<Value, Props = Record<string, unknown>>({
     name,
     type,
   });
+
+  const { setFieldValue } = useFormikContext();
 
   const error = showError && meta.error && meta.touched ? meta.error : '';
 
@@ -95,6 +102,12 @@ function FormField<Value, Props = Record<string, unknown>>({
             }
           }}
           onBlur={(event: FocusEvent) => {
+            if (trimInput) {
+              setFieldValue(
+                name,
+                (event.target as HTMLInputElement).value.trim(),
+              );
+            }
             if (typedProps.onBlur) {
               typedProps.onBlur(event);
             }
@@ -103,6 +116,11 @@ function FormField<Value, Props = Record<string, unknown>>({
               field.onBlur(event);
             }
           }}
+          onKeyPress={(event: KeyboardEvent) =>
+            trimInputOnEnter &&
+            event.key === 'Enter' &&
+            (event.currentTarget as HTMLInputElement)?.blur()
+          }
         />
       );
     }
@@ -130,6 +148,9 @@ function FormField<Value, Props = Record<string, unknown>>({
     meta,
     name,
     props,
+    setFieldValue,
+    trimInput,
+    trimInputOnEnter,
   ]);
 
   return formGroup ? (
--- a/src/explore-education-statistics-common/src/components/form/FormSearchBar.tsx
+++ b/src/explore-education-statistics-common/src/components/form/FormSearchBar.tsx
@@ -43,10 +43,11 @@ const FormSearchBar = ({
   }, [initialValue]);
 
   const handleSubmit = () => {
-    if (searchTerm.length >= min) {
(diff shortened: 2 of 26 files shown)


src/explore-education-statistics-common/src/components/form/rhf/hooks/useRegister.ts
```diff
@@ -16,7 +16,14 @@ export default function useRegister<
...
   name: TFieldName,
   register: UseFormRegister<TFieldValues>,
+  trimInput = false,
 ): UseFormRegisterReturn<TFieldName> {
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  return useMemo(() => register(name), [name]);
+  return useMemo(
+    () =>
+      register(name, {
+        setValueAs: value => (trimInput ? value.trim() : value),
```
ntsim, line 24: This is unsafe and could throw an error for non-string values. We'd want to check that the value is actually a string before we try and trim it.

However, we don't actually need this if we just do the trimming in the `FormTextArea` and `FormTextInput` (as mentioned in other places).

src/explore-education-statistics-common/src/components/form/__tests__/FormFieldTextArea.test.tsx
```diff
@@ -0,0 +1,318 @@
...
+    expect(screen.getByLabelText('Test field')).toHaveAttribute(
+      'aria-describedby',
+      'customId-hint',
+    );
+
+    expect(screen.getByText('Test hint')).toHaveAttribute(
+      'id',
+      'customId-hint',
+    );
+  });
```
ntsim, line 155: Bit torn about a lot of these tests as they're really repetitive across a lot of the field components. Appreciate I probably wrote the first components that did this type of checking, but regretting it a little now!

In these tests, we're just checking that a correct `id` has been passed to the component rendered by `FormField`. We (usually) check hint and error ids align correctly to the `id` prop in the `Form*` (non-field) component tests too, so this extra checking is basically redundant.

We don't have to do it in this PR, but perhaps we could do a follow up to refactor the Formik and RHF component tests and move any relevant tests into `FormField` and `RHFFormField` test suites respectively.

Tests would be something like the following:

```tsx
describe('FormField', () => {
  const TestInput = ({
    error,
    id,
    label,
    name,
    ...props
  }: { error?: string; label: string } & HTMLAttributes<HTMLInputElement>) => {
    return (
      <>
        <label htmlFor={id}>{label}</label>

        {error && <p>{error}</p>}

        <input
          // eslint-disable-next-line react/jsx-props-no-spreading
          {...props}
          name={name}
          id={id}
        />
      </>
    );
  };

  test('renders field with default id', () => {
    render(
      <Formik initialValues={{}} onSubmit={noop}>
        <FormField name="testField" label="Test field" as={TestInput} />
      </Formik>,
    );

    expect(screen.getByLabelText('Test field')).toHaveAttribute(
      'id',
      'testField',
    );
  });

  test('renders field with default id with form', () => ...);

  test('renders field with default id without form', () => ...);

  test('renders field with custom id with form', () => ...); 

  test('renders field with custom id without form', () => ...);

  test('renders error message', () => ...);

  // etc, etc...
});
```

Note that this means we don't really need to check stuff like hints anymore as we can leave this to the base `Form*` component tests instead. 

Would probably suggest keeping tests more relevant to proving integration or behaviours like changing / blurring the field in the specific `FormField*` and `RHFFormField*` components

Any thoughts? We can also just keep doing what we're doing. That's fine too if we don't mind the repetitiveness of it.

ntsim (changes requested):

src/explore-education-statistics-common/src/components/form/FormField.tsx
```diff
@@ -95,6 +102,12 @@ function FormField<Value, Props = Record<string, unknown>>({
             }
           }}
           onBlur={(event: FocusEvent) => {
+            if (trimInput) {
+              setFieldValue(
```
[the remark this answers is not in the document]
amyb-hiveit: Yeah that'd be simpler! I'll do that and open a new PR for it.

src/explore-education-statistics-common/src/components/form/__tests__/FormFieldTextArea.test.tsx
amyb-hiveit: TBH I only added them because I thought you'd say they should be there to be consistent with the other FormField* tests 😆 
I'll add a note to the RHF form ticket to look at improving the tests as part of that.

src/explore-education-statistics-common/src/components/form/FormSearchBar.tsx
```diff
@@ -43,10 +43,11 @@ const FormSearchBar = ({
   }, [initialValue]);
 
   const handleSubmit = () => {
-    if (searchTerm.length >= min) {
+    const trimmed = searchTerm.trim();
```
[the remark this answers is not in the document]
amyb-hiveit: I've left this out of the new PR for this.

amyb-hiveit:
Closing this in favour of https://github.com/dfe-analytical-services/explore-education-statistics/pull/4405

PR #4409, opened 2023-11-15
EES-4674 fix table alignment bug

amyb-hiveit:
Fixes a new table alignment bug caused by missing data and merged cells. The previous fix didn't work in this case as it didn't account for there being a mix of rows with different merged and unmerged cells in within a row group.

Before:
broken

After:
fixed

### DIFF
--- a/src/explore-education-statistics-common/src/modules/table-tool/utils/createExpandedRowHeaders.ts
+++ b/src/explore-education-statistics-common/src/modules/table-tool/utils/createExpandedRowHeaders.ts
@@ -91,31 +91,31 @@ function getRowHeadersInfo(rowHeaders: Header[]): {
 
   rowHeaders.forEach(header => {
     const { maxCrossSpan } = header;
+    const stack = [header];
 
-    let current: Header | undefined = header;
+    while (stack.length > 0) {
+      const current = stack.shift();
 
-    while (current) {
-      const { depth } = current;
-
-      if (current.parent) {
-        const isCollapsibleRowHeaderLevel =
-          current.hasSiblings() &&
-          current.parent.children.every(child =>
-            child.hasSingleMatchingChild(),
-          );
-
-        collapsibleLevels[depth] =
-          (collapsibleLevels[depth] ?? true) && isCollapsibleRowHeaderLevel;
+      if (!current) {
+        break;
       }
 
-      current = current.getFirstChild();
-    }
+      if (current.hasChildren()) {
+        const { depth } = current;
 
+        if (current.parent && collapsibleLevels[depth] !== false) {
+          collapsibleLevels[depth] =
+            current.hasSiblings() &&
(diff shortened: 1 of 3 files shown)


src/explore-education-statistics-common/src/modules/table-tool/utils/createExpandedRowHeaders.ts
```diff
@@ -91,31 +91,33 @@ function getRowHeadersInfo(rowHeaders: Header[]): {
...
     const { maxCrossSpan } = header;
+    const stack = [header];
 
-    let current: Header | undefined = header;
+    while (stack.length > 0) {
+      const current = stack.shift();
+
+      if (!current) {
+        break;
+      }
```
ntsim, line 102: Think we might be able to do a further optimisation here by checking if `!current.hasChildren`? e.g.

```ts
if (!current.hasChildren()) {
  continue;
}
```

Means we ignore the edge nodes, which never have a single matching child (and are hence always false for the collapsible level check).

src/explore-education-statistics-common/src/modules/table-tool/utils/createExpandedRowHeaders.ts
```diff
@@ -91,31 +91,33 @@ function getRowHeadersInfo(rowHeaders: Header[]): {
...
             child.hasSingleMatchingChild(),
           );
 
-        collapsibleLevels[depth] =
-          (collapsibleLevels[depth] ?? true) && isCollapsibleRowHeaderLevel;
+        collapsibleLevels[depth] = isCollapsibleRowHeaderLevel;
       }
 
-      current = current.getFirstChild();
+      if (current.hasChildren()) {
```
ntsim, line 115: If we do the above optimisation around `!current.hasChildren`, we don't need this condition anymore

src/explore-education-statistics-common/src/modules/table-tool/utils/createExpandedRowHeaders.ts
```diff
@@ -91,31 +91,33 @@ function getRowHeadersInfo(rowHeaders: Header[]): {
 
   rowHeaders.forEach(header => {
     const { maxCrossSpan } = header;
+    const stack = [header];
```
ntsim, line 94: Nice use of DFS. I did wonder if a BFS might be more beneficial somehow? Haven't look too hard into it, but maybe something to consider as well.

src/explore-education-statistics-common/src/modules/table-tool/utils/createExpandedRowHeaders.ts
amyb-hiveit: There's an eslint rule against using `continue`, it recommends using if statements instead so I've done that.

ntsim (approved):

PR #4413, opened 2023-11-17
EES-4680 - moved SendPreReleaseUserInviteEmails() from public method …

duncan-at-hiveit:
## UI test run update - 17/11/23

image

This PR:
- resolves a bug whereby a permission check was failing when attempting to send Pre-Release emails, causing the approval of the Release to silently fail to send the emails.

There were 2 main things to sort out in this PR, and an additional bonus one:
* Stop the permissions issue from occurring.
* Stop silently ignoring errors during Release approval.
* Bonus - prevent the release approval completing if any failures are encountered along the way. 

## Stop the permissions issue from occurring
Upon looking further into it, and given that ReleaseApprovalService was the only caller for SendPreReleaseUserInviteEmails(), I chose to bring that method into ReleaseApprovalService itself, and strip out the permission check as it's already been done in the calling method.

## Stop silently ignoring errors during Release approval
For this, I re-jigged the code to always take into account the return values from sending emails.  I also noticed that we were ignoring potential failures from informing the Publisher of the Release change as well, so I got it to take into account the return values from this call as well.

## Prevent approval from completing if failures occur
It seemed dangerous to allow the Release approval to partially go through if either Pre-Release emails couldn't be sent or the Publisher was not informed of the Release change.  Therefore, I added a basic rollback of the Release updates and the new Release Status that is created just prior to the Publisher being notified.  Removing unused files is NOT rolled back, but it seemed that that was a low priority thing to attempt to roll back.  

I thought initially that we could just delay the updating of the Release and the creation of the new Release Status until after the Publisher was notified of the change, so we didn't need t have anything to roll back, but I was surprised that this actually caused failures.  Given notifying the Publisher was simply a queue message drop, I would have thought there would be no race condition likely between then moving on to saving the Release changes, and the queue message being dealt with, but locally at least, there was!

## Other misc things
* Removed an old unused SecurityPolicy.
* Replaced PersistenceHelper with plain DbContext lookups 
* Raised ticket EES-4681 to investigate why sometimes we're marking emails as sent and sometimes we're not.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServicePermissionTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServicePermissionTests.cs
@@ -27,7 +27,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
         private readonly Release _release = new()
         {
             Id = Guid.NewGuid(),
-            Publication= new Publication(),
+            Publication = new Publication(),
             Published = DateTime.Now,
             TimePeriodCoverage = TimeIdentifier.April
         };
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServicePermissionTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServicePermissionTests.cs
@@ -9,7 +9,6 @@ using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.Manag
 using GovUk.Education.ExploreEducationStatistics.Common.Model;
 using GovUk.Education.ExploreEducationStatistics.Common.Services;
 using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces.Security;
-using GovUk.Education.ExploreEducationStatistics.Common.Utils;
 using GovUk.Education.ExploreEducationStatistics.Content.Model;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Repository.Interfaces;
@@ -17,7 +16,7 @@ using Microsoft.Extensions.Options;
 using Moq;
 using Xunit;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Security.SecurityPolicies;
-using static GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils.MockUtils;
+using static GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.DbUtils;
 using static GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils.PermissionTestUtils;
 using IReleaseRepository = GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.IReleaseRepository;
 
@@ -25,15 +24,10 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
 {
     public class ReleaseApprovalServicePermissionTests
     {
-        private static readonly Publication Publication = new()
-        {
-            Id = Guid.NewGuid()
-        };
-
         private readonly Release _release = new()
         {
             Id = Guid.NewGuid(),
-            PublicationId = Publication.Id,
+            Publication= new Publication(),
             Published = DateTime.Now,
             TimePeriodCoverage = TimeIdentifier.April
         };
@@ -42,26 +36,43 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
         public async Task GetReleaseStatuses()
         {
             await PolicyCheckBuilder<SecurityPolicies>()
-                .SetupResourceCheckToFail(_release, CanViewReleaseStatusHistory)
-                .AssertForbidden(
-                    userService =>
+                .SetupResourceCheckToFailWithMatcher<Release>(r => r.Id == _release.Id, CanViewReleaseStatusHistory)
+                .AssertForbidden(async userService =>
+                {
+                    var contentDbContextId = Guid.NewGuid().ToString();
+                    await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContextId))
                     {
-                        var service = BuildService(userService.Object);
-                        return service.GetReleaseStatuses(_release.Id);
+                        await contentDbContext.AddRangeAsync(_release);
+                        await contentDbContext.SaveChangesAsync();
                     }
-                );
+
+                    await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContextId))
+                    {
+                        var service = BuildService(contentDbContext: contentDbContext, userService.Object);
+                        return await service.GetReleaseStatuses(_release.Id);
+                    }
+                });
         }
 
         [Fact]
         public async Task UpdateReleaseStatus_Draft()
         {
             await PolicyCheckBuilder<SecurityPolicies>()
-                .SetupResourceCheckToFail(_release, CanMarkSpecificReleaseAsDraft)
-                .AssertForbidden(
-                    userService =>
+                .SetupResourceCheckToFailWithMatcher<Release>(r => r.Id == _release.Id, CanMarkSpecificReleaseAsDraft)
+                .AssertForbidden(async userService =>
+                {
+                    var contentDbContextId = Guid.NewGuid().ToString();
+                    await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContextId))
                     {
-                        var service = BuildService(userService.Object);
-                        return service.CreateReleaseStatus(
+                        await contentDbContext.AddRangeAsync(_release);
+                        await contentDbContext.SaveChangesAsync();
+                    }
+
+                    await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContextId))
+                    {
+                        var service = BuildService(contentDbContext: contentDbContext, userService.Object);
+
+                        return await service.CreateReleaseStatus(
                             _release.Id,
                             new ReleaseStatusCreateRequest
                             {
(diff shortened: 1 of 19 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseApprovalService.cs
```diff
@@ -113,92 +113,155 @@ IEmailTemplateService emailTemplateService
...
-                            }
-
-                            switch (request.ApprovalStatus)
+                        })
+                        .OnSuccess(() =>
+                            SendEmailNotificationsAndInvites(request, releaseToUpdate)
+                            .OnSuccess(() => NotifyPublisher(releaseId, request, oldStatus, releaseStatus))
+                            // If a failure occurs in either the sending of the emails or the Publisher notification,
+                            // roll back changes to the Release and remove the added Release Status.
+                            .OnFailureDo(async _ =>
```
benoutram, line 170: Is there anything extra that could be mentioned here to explain why we aren't just delaying the saving until after both of those are successful? Without reading the PR description first, if someone hit this bit of code they might be tempted to try undoing this.

Is notifying the publisher likely to fail given that it's just a queue message drop? It looks like our code within that method would currently only fail gracefully if the release isn't found.

I think delaying the save until the notifications have been sent, then doing the save, and then notifying the publisher could have been another approach here, but happy to go with this!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseApprovalService.cs
```diff
@@ -113,92 +113,155 @@ IEmailTemplateService emailTemplateService
             Guid releaseId,
             ReleaseStatusCreateRequest request)
         {
-            return await _persistenceHelper
-                .CheckEntityExists<Release>(releaseId, ReleaseChecklistService.HydrateReleaseForChecklist)
+            return await _context
+                .Releases
+                .HydrateReleaseForChecklist()
```
benoutram, line 118: It seems a bit strange what we do this here, when I would have thought the `ReleaseCheckListService` it should be responsible for checking the release is hydrated. I think I would change things to not pass the release around in the first place, and pass the unique identifier of the release instead.

Probably not one for this PR though.

I found the extension method for `HydrateReleaseForChecklist` a little strange in the way that its scope now makes this method accessible to anywhere in the Admin, seen when checking available methods of `_context.Releases`.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServicePermissionTests.cs
```diff
@@ -9,31 +9,25 @@
...
-        private static readonly Publication Publication = new()
-        {
-            Id = Guid.NewGuid()
-        };
-
         private readonly Release _release = new()
         {
             Id = Guid.NewGuid(),
-            PublicationId = Publication.Id,
+            Publication= new Publication(),
```
benoutram, line 30: Looks like a formatting problem here

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServiceTests.cs
```diff
@@ -1280,6 +1359,240 @@ await using (var context = InMemoryApplicationDbContext(contextId))
...
+                    .HydrateReleaseForChecklist()
+                    .Include(r => r.ReleaseStatuses)
+                    .SingleAsync(r => r.Id == release.Id);
+
+                // Assert that the failure to send emails prevented the Release from completing approval.
+                // The Release should remain unchanged from the original Release, as changes to it were rolled back.
+                // Additionally, no new ReleaseStatus entries were added.
+                saved.AssertDeepEqualTo(release);
+
+                // Futhermore, we have proven that the Publisher was not informed of the Release change, as it
```
benoutram, line 1475: There's a typo here

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseApprovalService.cs
```diff
@@ -113,92 +113,155 @@ IEmailTemplateService emailTemplateService
...
+            var emailResults = await unsentUserReleaseInvites
+                .ToAsyncEnumerable()
+                .SelectAwait(async invite =>
+                {
+                    var user = await _userRepository.FindByEmail(invite.Email);
+                    return await _preReleaseUserService.SendPreReleaseInviteEmail(
+                            release,
+                            invite.Email.ToLower(),
+                            isNewUser: user == null)
+                        .OnSuccessDo(() => _preReleaseUserService.MarkInviteEmailAsSent(invite));
```
benoutram, line 258: If `invite` is passed to `SendPreReleaseInviteEmail` then it can be responsible for marking as sent making this logic more encapsulated, but I'm guessing you already attempted this change and then realised that not all usages of `SendPreReleaseInviteEmail` are marking as sent currently?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServicePermissionTests.cs
```diff
@@ -42,52 +36,79 @@ public class ReleaseApprovalServicePermissionTests
         public async Task GetReleaseStatuses()
         {
             await PolicyCheckBuilder<SecurityPolicies>()
-                .SetupResourceCheckToFail(_release, CanViewReleaseStatusHistory)
-                .AssertForbidden(
-                    userService =>
+                .SetupResourceCheckToFailWithMatcher<Release>(r => r.Id == _release.Id, CanViewReleaseStatusHistory)
```
benoutram, line 39: Just wondering if there's any benefit to using this over `.SetupResourceCheckToFail(_release, CanViewReleaseStatusHistory)` given that we have `_release` available here?

src/GovUk.Education.ExploreEducationStatistics.Common/Model/Either.cs
```diff
@@ -82,6 +86,18 @@ public static class EitherExtensions
...
+        /// failure.
+        /// </summary>
+        public static Either<TFailure, Unit> OnSuccessAllReturnVoid<TFailure, TSuccess>(
+            this IEnumerable<Either<TFailure, TSuccess>> items)
+        {
+            return items
+                .OnSuccessAll()
+                .OnSuccessVoid();
+        }
+
```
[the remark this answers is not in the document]
duncan-at-hiveit: :smile:

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseApprovalService.cs
```diff
@@ -75,16 +72,19 @@ IEmailTemplateService emailTemplateService
...
-            return await _persistenceHelper
-                .CheckEntityExists<Release>(releaseId, release =>
-                    release.Include(r => r.ReleaseStatuses)
-                        .ThenInclude(rs => rs.CreatedBy))
+            return await _context
+                .Releases
+                .Include(r => r.ReleaseStatuses)
+                .ThenInclude(rs => rs.CreatedBy)
+                .SingleOrDefaultAsync(r => r.Id == releaseId)
+                .OrNotFound()
```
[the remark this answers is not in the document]
duncan-at-hiveit: Oh cool I had no idea :) Thanks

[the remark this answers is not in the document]
duncan-at-hiveit: Done across the board, thanks.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseApprovalService.cs
```diff
@@ -113,92 +113,155 @@ IEmailTemplateService emailTemplateService
...
+                }
+            }
+
+            return Unit.Instance;
+        }
+
+        private async Task<Either<ActionResult, Unit>> SendPreReleaseUserInviteEmails(Release release)
+        {
+            var unsentUserReleaseInvites = await _context.UserReleaseInvites
+                .AsQueryable()
```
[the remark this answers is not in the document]
duncan-at-hiveit: Yup, just a hangover from the original method.  Will remove and check that it's still happy thanks.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseApprovalService.cs
```diff
@@ -113,92 +113,155 @@ IEmailTemplateService emailTemplateService
...
+
+                if (notifyHigherReviewers)
+                {
+                    return userReleaseRoles
+                        .Select(urr => urr.User.Email)
+                        .Concat(userPublicationRoles.Select(upr => upr.User.Email))
+                        .Distinct()
+                        .Select(email => _emailTemplateService.SendReleaseHigherReviewEmail(email, release))
+                        .OnSuccessAllReturnVoid();
+                }
```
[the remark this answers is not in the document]
duncan-at-hiveit: Great idea.  Done!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseApprovalService.cs
duncan-at-hiveit: Thanks for pushing on this.

Turns out the root problem was that the ReleaseStatus, if being saved after the NotifyChange message was sent, would not have been assigned an Id yet and so the message to the Publisher contained an Guid.Empty ReleaseStatusId.

The fix is to assign the Id manually before sending the Publisher message.

As it turns out, the "rollback" tests can continue to assert that nothing's changed as they were before.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseApprovalService.cs
duncan-at-hiveit: Yup that's interesting.  I'll leave it for now, but it's definitely a sign I guess that we're maybe not using the ReleaseChecklistService in the best way.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServicePermissionTests.cs
duncan-at-hiveit: Sorted ta

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServiceTests.cs
duncan-at-hiveit: Hehe took me about a minute to see it - sorted ta.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseApprovalService.cs
duncan-at-hiveit: Yeah exactly - I don't want to change any existing behaviour as part of this ticket.  Rather, I've suggested an investigation and implementation in ticket EES-4681.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServicePermissionTests.cs
duncan-at-hiveit: Because we're not using PersistanceHelper anymore, we're not getting back the same instance of Release that is _release anymore, but rather the instance that is pulled out of the contentDbContext in the method being checked.  Therefore we can't rely on the instance check anymore.

We could start throwing Equals methods around in our domain objects, but I'd personally prefer to do that if the production code needs it rather than helping the tests out.

duncan-at-hiveit: If I'm understanding you, that is?  :slightly_smiling_face:

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseApprovalService.cs
benoutram: Cool, it was great to spot that one!

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServicePermissionTests.cs
benoutram: Yes that's what I meant thanks, I understand now.

I was making a similar change today and I think the reason why I didn't run into needing this change is that I saved `_release` outside this block just before `await PolicyCheckBuilder<SecurityPolicies>()`.

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseApprovalServicePermissionTests.cs
duncan-at-hiveit: :+1:  I think it'd be worth finding the best pattern to use here given we'll be doing a lot of this as we phase out PersistenceHelper

PR #4417, opened 2023-11-17
EES-4697 Stale element reference fix

saicharan2789:
This PR fixes the Stale element reference issue which was happening in data catalogue for 'Choose Publication' Test Case.


duncan-at-hiveit:
I'm suggesting raising this as a bug ticket, as the behaviour that currently breaks this test is manually reproducible and will affect end users if they're quick enough or if our search results are slow enough.

https://www.awesomescreenshot.com/video/22552319?key=216154244fdf85e85725f525df40041e

When tackling this bug ticket, we can also revisit this test and hopefully remove the page reload.

duncan-at-hiveit (approved):

PR #4420, opened 2023-11-20
EES-4660 Part 1 - Refactor Data Guidance to use file id's and rename terminology to use 'data set'

benoutram:
This PR is being raised prior to the main change for EES-4660 which will begin setting the data guidance content for data sets onto `ReleaseFile` and migrate existing data guidance from `ReleaseSubject` to `ReleaseFile`.

In preparation for that change this PR changes all the data guidance code, frontend and backend, to use file id's instead of subject id's.

This will make more sense in future for the request updating data guidance, where the entities being updated are of type `ReleaseFile` rather than a `ReleaseSubject`, when no part of the subject will be updated in future.

It also simplifies the interaction on the backend between the `DataGuidanceService` and the `DataGuidanceDataSetService` for listing data sets. Previously we began with a release, looked up the data set files attached to that release, got a list of the subject id's from those files, before handing off to `DataGuidanceDataSetService` which needed to find the file id's again for the subjects to build the guidance for each data set.

After this change we will work with file id's everywhere except for needing the subject id to fetch the meta data to populate the rest of the data guidance.

This PR also changes all occurrences of the term 'subject' used throughout the data guidance code to be 'data set'.

## Other changes

- Replace backend view models used for updating data guidance with new request types.
- Validates file id's are attached to the release when updating data guidance
- Fixes a bug with `DoZipFilesToStream` methods writing data guidance to the text file for files which are not in the zip file because they weren't found in blob storage. This was previously missed by the unit test which should have covered this scenario but didn't because the files were not set up with subject id's and were not considered to be data files because of that.
- Make use of `CancellationToken` where available.

## UI test report

image


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataGuidanceService.cs
```diff
@@ -1,150 +1,157 @@
...
 
-        public async Task<Either<ActionResult, DataGuidanceViewModel>> Get(Guid releaseId)
+        public async Task<Either<ActionResult, DataGuidanceViewModel>> GetDataGuidance(Guid releaseId,
+            CancellationToken cancellationToken = default)
         {
-            return await _contentPersistenceHelper.CheckEntityExists<Release>(releaseId)
-                .OnSuccessDo(release => _userService.CheckCanViewRelease(release))
-                .OnSuccess(BuildViewModel);
+            return await _contentDbContext.Releases
+                .FirstOrNotFoundAsync(release => release.Id == releaseId, cancellationToken)
```
mmyoungman, line 51: Just interested in why you're doing this instead of using `CheckEntityExists`? I assume because of the cancellation token?

mmyoungman (approved):

benoutram:
Added 'Do Not Merge' label because it looks like there's some UI test failures which I need to fix.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataGuidanceService.cs
benoutram: It's because we've recently been talking about phasing out our use of `PersistenceHelper` due to it not adding extra value over what can already be achieved using Entity Framework ootb methods like `FirstOrDefaultAsync` combined with our own extensions like `FirstOrNotFoundAsync`. This combines `FirstOrDefaultAsync` and our own `NotFound`. 

Using `PersistenceHelper` also requires wiring an extra dependency.

We have a note about it already raised for the back end style guide meeting.

PR #4423, opened 2023-11-21
Set .python-version to 3.10.11

ntsim:
This was previously broken as 3.10 is not a valid `.python-version`.


benoutram (commented):
A few of us are having an issue updating pyenv to use any 3.10.x version greater than 3.10.11.

```
$ pyenv update
:: [Info] ::  Mirror: https://www.python.org/ftp/python
:: [Info] ::  Scanned 202 pages and found 662 installers.
```

```
$ pyenv install 3.10.13
:: [Info] ::  Mirror: https://www.python.org/ftp/python
pyenv-install: definition not found: 3.10.13
```

Just mentioning this while we investigate further.

benoutram:
The problem appears to be that the Windows executables haven't been published here for versions 3.10.12 and 3.10.13, unlike 3.10.11.

Compare https://www.python.org/ftp/python/3.10.13/ with https://www.python.org/ftp/python/3.10.11/.

benoutram (approved):

PR #4424, opened 2023-11-22
EES-4660 Part 2 - Begin setting Data Guidance on `ReleaseFile.Summary`

benoutram:
This PR follows on from https://github.com/dfe-analytical-services/explore-education-statistics/pull/4420.

It begins setting data guidance in `ReleaseFile.Summary` for each data set updated in  `DataGuidanceService.UpdateDataGuidance`.

It also changes `ReplacementService` to begin copying data guidance from the original to the replacement `ReleaseFile` during the replace operation.

The current behaviour of setting and replacing data guidance in its original location in `ReleaseSubject.DataGuidance` is retained for now.

This PR will be followed up with a further change to migrate data guidance for all existing data sets from `ReleaseSubject` to `ReleaseFile`.

### Other changes

- Fixes an issue in `ReplacementService.RemoveOriginalSubjectAndFileFromRelease` which would ignore an error if the replacement file is not found.
- Adds a test to ensure `DataGuidanceDataSetService.ListDataSets` ignores data replacements in progress.
- Adds some 'TODO's' in places requiring changes in EES-4661 when `ReleaseSubject.DataGuidance` will be dropped.
- Changes the behaviour of the methods in `ReplacementService` to return not found if the file id's are not `FileType.Data`, removing `ValidationErrorMessages.ReplacementFileTypesMustBeData`.
- Remove setting `ReleaseSubject.DataGuidance` values on tests where its not relevant to the test.

## UI test report

image

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataGuidanceServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataGuidanceServiceTests.cs
@@ -203,7 +203,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                 {
                     Filename = "file1.csv",
                     Type = FileType.Data
-                }
+                },
+                Summary = "Data set 1 guidance"
             };
 
             var releaseFile2 = new ReleaseFile
@@ -213,7 +214,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                 {
                     Filename = "file2.csv",
                     Type = FileType.Data
-                }
+                },
+                Summary = "Data set 2 guidance"
             };
 
             var contentDbContextId = Guid.NewGuid().ToString();
@@ -255,11 +257,23 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
 
             await using (var contentDbContext = InMemoryApplicationDbContext(contentDbContextId))
             {
-                // Assert no changes have been made
+                // Assert no changes have been made to the release or any of the data sets
                 var actualRelease = await contentDbContext.Releases
                     .FirstAsync(r => r.Id == release1.Id);
 
                 Assert.Equal("Release 1 guidance", actualRelease.DataGuidance);
+
+                var actualReleaseFile1 = await contentDbContext.ReleaseFiles
+                    .FirstAsync(rf => rf.ReleaseId == release1.Id
+                                      && rf.FileId == releaseFile1.FileId);
+
+                Assert.Equal("Data set 1 guidance", actualReleaseFile1.Summary);
+
+                var actualReleaseFile2 = await contentDbContext.ReleaseFiles
+                    .FirstAsync(rf => rf.ReleaseId == release2.Id
+                                      && rf.FileId == releaseFile2.FileId);
+
+                Assert.Equal("Data set 2 guidance", actualReleaseFile2.Summary);
             }
         }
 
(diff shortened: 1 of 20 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataGuidanceService.cs
```diff
@@ -132,6 +133,15 @@ public class DataGuidanceService : IDataGuidanceService
             {
                 var content = updateRequestsByFileId[releaseFile.FileId].Content;
 
+                _contentDbContext.Update(releaseFile);
+                releaseFile.Summary = content;
```
mmyoungman, line 137: The `Summary` column was already on `ReleaseFile`? So I guess it wasn't being used?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReplacementService.cs
```diff
@@ -141,29 +135,50 @@ await plan.Footnotes
...
+            return await _contentDbContext.ReleaseFiles
+                .Include(rf => rf.File)
+                .FirstOrNotFoundAsync(rf => rf.ReleaseId == releaseId
+                                            && rf.FileId == originalFileId
+                                            && rf.File.Type == FileType.Data)
+                .OnSuccessCombineWith(async _ => await _contentDbContext.ReleaseFiles
+                    .Include(rf => rf.File)
+                    .FirstOrNotFoundAsync(rf => rf.ReleaseId == releaseId
+                                                && rf.FileId == replacementFileId
+                                                && rf.File.Type == FileType.Data))
```
mmyoungman, line 163: I was looking at this and wondering whether it could be a single query - not sure it's worth the bother, cause I think it require extra code below and performance isn't a concern here. Honestly, wouldn't be surprised if you tried one query first and then switched to this :)

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataGuidanceServiceTests.cs
mmyoungman, line 1: I was surprised there was no conflicts with this PR and Dunc's amendment changes. At first I was tempted to suggest rebasing onto dev, to get Dunc's changes, to ensure nothing untoward has gone on. But looking at it further, I think it's all fine as it looks like you've got amendments for free with this work. So I think it's fine.

Still leaving this comment in case you see something I don't.

mmyoungman (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataGuidanceServiceTests.cs
benoutram: I need to run the UI tests again anyway, so I will rebase from dev. Thanks.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataGuidanceService.cs
benoutram: Yep that's correct, it's not used for ReleaseFiles where FileType is 'Data'.

We've had some discussions recently about this not being ideal due to it mixing html and non-html text in the `Summary` field. `Summary` is already being used for other types of files like ancillary where it's not html.

Ideally we'd refactor files of type 'Data' and 'Metadata' to be a single DataSet entity, and then we might have something like a DataSetVersion or a ReleaseDataSet where the guidance would belong.

We said that this refactoring would be out of scope for this ticket.

There's also a plan to strip html from the data guidance in the near future and tidy up any guidance for data sets that's only repeating what's already found in the metadata, so that the guidance is more like a summary. We need cleaner guidance to be displayed as summaries in the data catalogue.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReplacementService.cs
benoutram: I already considered it but decided against it because I figured we would have to handle the case where the resultset of the database query contained zero items or only 1 item.

Even in the case where both items are returned as expected we would still have needed to pick out the original and replacement items by file id to return them from this method in a consistent order since that wouldn't be guaranteed by the database query.

PR #4425, opened 2023-11-22
Ees 4622/bugfix/fix release type not appearing on featured tables

sambiramairelogic:
For Featured Tables on public releases, the new section displaying Release Type wasn't displaying correctly. 
This was ultimately caused by a value holding a runtime value that would be invalid in the type system; the `ReleaseType` which has the following type:
image
held the value `'0'` at runtime, which obviously isn't a value it should ever hold. 
The cause of this was the ViewModel holding an `Enum` property and the `get` method in `Client.ts` returning an unknown type, which allowed for the Dotnet representation of the ReleaseType enum to translate straight into Javascript. 

To solve this, I'm now returning a string from the view model and casting the Enum. 

I've also changed the text "Office for Statistics Regulation" into an href to the appropriate url.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Api/ViewModels/FastTrackViewModel.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Api/ViewModels/FastTrackViewModel.cs
@@ -3,6 +3,7 @@ using GovUk.Education.ExploreEducationStatistics.Common.Model.Data;
 using GovUk.Education.ExploreEducationStatistics.Content.Model;
 using GovUk.Education.ExploreEducationStatistics.Data.Services.ViewModels;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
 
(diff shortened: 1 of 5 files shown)


src/GovUk.Education.ExploreEducationStatistics.Data.Api/ViewModels/FastTrackViewModel.cs
```diff
@@ -26,7 +26,7 @@ public record FastTrackViewModel
 
         public string ReleaseSlug { get; set; }
         
-        public ReleaseType ReleaseType { get; set; }
+        public string ReleaseType { get; set; }
```
benoutram, line 29: I think you should be able to use the annotation `[JsonConverter(typeof(StringEnumConverter))]` rather than making this change from `ReleaseType` to `string`. There's lots of other examples of view models in the codebase using this.

benoutram (approved):
Approved from a backend point of view.

I think @amyb-hiveit is going to approve the front end changes.

🙂

PR #4428, opened 2023-11-23
EES-4715 suppress jsdom css errors

amyb-hiveit:
JSDOM doesn't handle CSS well and seems to be particularly objecting to CKEditor, which is causing a ridiculous amount of errors in the log when running the admin tests. ~~As far as I can tell it shouldn't be trying to parse it anyway as all CSS files should be going through the transform we have configured in jest.config.js, but it is so I've just supressed this error.~~

As suggested by @ntsim below, I've stubbed out CKEditor to prevent errors from its CSS build polluting the test output.

### DIFF
--- a/src/explore-education-statistics-admin/jest.config.js
+++ b/src/explore-education-statistics-admin/jest.config.js
@@ -32,6 +32,7 @@ const config = {
       '<rootDir>/../explore-education-statistics-common/test/$1',
     '^axios$': '<rootDir>/node_modules/axios/dist/axios.js',
     '^marked$': '<rootDir>/node_modules/marked/lib/marked.umd.js',
+    '^explore-education-statistics-ckeditor$': '<rootDir>/test/EditorStub.js',
   },
   moduleFileExtensions: [
     'web.js',
--- a/src/explore-education-statistics-admin/test/EditorStub.js
+++ b/src/explore-education-statistics-admin/test/EditorStub.js
@@ -1,0 +1,3 @@
+// Stub out CKEditor to prevent errors from its
+// CSS build polluting the test output.
+export default class EditorStub {}


src/explore-education-statistics-admin/test/setupTests.js
```diff
@@ -10,3 +10,11 @@ if (typeof window !== 'undefined') {
...
   require('cross-fetch');
 }
+
+const originalConsoleError = console.error;
+const jsDomCssError = 'Error: Could not parse CSS stylesheet';
+console.error = (...params) => {
+  if (!params.find(p => p.toString().includes(jsDomCssError))) {
+    originalConsoleError(...params);
+  }
+};
```
ntsim, line 20: Looked into the stylesheet a bit more and it seems like there's a few syntax errors which suggests something's potentially wrong with the built CSS. Most of the build config around this is from `@ckeditor/ckeditor5-dev-utils`, so maybe that's not quite configured correctly, or they may even have errors in their source CSS.

Unfortunately, the browser will just ignore stylesheet errors, so I guess this is slipping under the radar right now (one point for JSDOM). Hopefully they'll fix it in a later version of CKEditor.

Anyway, regardless of this issue, I think we can do a little better than just suppress the error. I figured out two options below:

## Option 1 - Swap in VirtualTestEditor

This is a more complicated approach and is based on [this issue](https://github.com/ckeditor/ckeditor5-react/issues/225#issuecomment-795236892), using Jest's `moduleNameMapper` we can stub out CKEditor (which doesn't really work in Node) with a `VirtualTestEditor` implementation. This can be found in `ckeditor5-core`, but isn't published as part of the package (check out [source file](https://github.com/ckeditor/ckeditor5/blob/b6c56e5ea7d31456590d9aeb5e4833810197294a/packages/ckeditor5-core/tests/_utils/virtualtesteditor.js)).

Testing this out, I created a test build of `explore-education-statistics-ckeditor` which exports a copy-paste of the `VirtualTestEditor`:

```js
// explore-education-statistics-ckeditor/src/ckeditor-test.js

import { DataApiMixin, Editor } from '@ckeditor/ckeditor5-core';

export default class VirtualTestEditor extends DataApiMixin(Editor) {
  constructor(config) {
    super(config);

    // Create the ("main") root element of the model tree.
    this.model.document.createRoot();
  }

  static create(config = {}) {
    return new Promise(resolve => {
      const editor = new this(config);

      resolve(
        editor
          .initPlugins()
          .then(() => editor.data.init(config.initialData || ''))
          .then(() => {
            editor.fire('ready');
            return editor;
          }),
      );
    });
  }
}
```

This just needs a few tweaks to the Webpack config:

```diff
// explore-education-statistics-ckeditor/webpack.config.js

+  entry: {
+    ckeditor: path.resolve(__dirname, 'src', 'ckeditor.js'),
+    'ckeditor-test': path.resolve(__dirname, 'src', 'ckeditor-test.js'),
+  },

   output: {
     // The name under which the editor will be exported.
     library: 'ClassicEditor',
     path: path.resolve(__dirname, 'build'),
+    filename: '[name].js',
     libraryTarget: 'umd',
     libraryExport: 'default',
   },
```

Once rebuilt, there's a `build/ckeditor-test.js` we can use to stub out the implementation. We can just swap this in through a `moduleNameMapper` in admin:

```js
moduleNameMapper: {
  // ...
  '^explore-education-statistics-ckeditor$': '<rootDir>/../explore-education-statistics-ckeditor/build/ckeditor-test.js',
}
```

## Option 2 - Stub out CKEditor

Iterating on option 1, we can actually go further and just do something like this:

```js
// explore-education-statistics-admin/test/EditorStub.js

export default class EditorStub {}
```

Then just configure the `moduleNameMapper` appropriately:

```js
moduleNameMapper: {
  // ...
  '^explore-education-statistics-ckeditor$': '<rootDir>/test/EditorStub.js',
}
```

This is a simpler approach and has some advantages:

- Practically no code needs to be built (option 1 outputs ~300 KB of JS)
- Basically nothing needs to be parsed during test runs

Giving it a light test, I didn't really see much of a difference in test performance on my machine. However, the above advantages might be more noticeable on lower end machines.

The main downside to this approach is that if in the future, we want to use a 'real' implementation of CKEditor in the tests, the stub obviously won't cut it. 

The `VirtualTestEditor` implementation actually runs CKEditor and would allow actual methods to be called on it (etc). The main difference is that it doesn't try and render anything to the DOM (hence works great in Node).

For the timebeing, this isn't really necessary as we just swap out the `CKEditor` component (in `FormEditor`) with a textarea.

Personally, I'd go with option 2 and we can swap in option 1 if needed in the future.

amyb-hiveit: Thanks @ntsim, I've changed it to use option 2 now.

PR #4429, opened 2023-11-23
EES-4660 Part 3 - Migrate existing data guidance to `ReleaseFile.Summary`

benoutram:
⚠️ This has an associated faffy deploy task [EES-4724](https://dfedigital.atlassian.net/browse/EES-4724) ⚠️ 

This PR follows on from https://github.com/dfe-analytical-services/explore-education-statistics/pull/4424.

It adds a migration endpoint which copies data guidance from the statistics database to the content database.

This needs to be run before we drop `ReleaseSubject.DataGuidance` in [EES-4661](https://dfedigital.atlassian.net/browse/EES-4661).

For every content database `ReleaseFile` of type `FileType.Data` which is also not a file replacement currently in progress, it copies `DataGuidance` from the corresponding `ReleaseSubject` found in the stattistics database.

### Usage
Make a request to the secured BAU endpoint `PATCH /bau/migrate-data-guidance?dryRun={true|false}`.

When option `dryRun=true` no update will be made to the content database. The default value is `true`.

### Report
A report is returned in the response containing the following values:

- `dryRun`: The value of the `dryRun` query parameter indicating whether a dry run took place or not.
- `releaseDataFilesExcludingReplacementsCount`: The count of release data files found in the content database excluding any replacement files for data replacements currently in progress.
- `fileIdsWithNoMatchingSubject`: Set of content database `File` id's where release files were found with no corresponding subject in the statistics database.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataGuidanceMigrationService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataGuidanceMigrationService.cs
@@ -1,0 +1,83 @@
+﻿#nullable enable
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces;
+using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.Security;
+using GovUk.Education.ExploreEducationStatistics.Admin.ViewModels;
+using GovUk.Education.ExploreEducationStatistics.Common.Model;
+using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces.Security;
+using GovUk.Education.ExploreEducationStatistics.Content.Model;
+using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
+using GovUk.Education.ExploreEducationStatistics.Data.Model.Database;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Services;
+
+/// <summary>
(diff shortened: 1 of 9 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataGuidanceMigrationService.cs
```diff
@@ -0,0 +1,83 @@
...
+    public DataGuidanceMigrationService(ContentDbContext contentDbContext,
+        StatisticsDbContext statisticsDbContext,
+        IUserService userService)
+    {
+        _contentDbContext = contentDbContext;
+        _statisticsDbContext = statisticsDbContext;
+        _userService = userService;
+    }
+
+    public async Task<Either<ActionResult, DataGuidanceMigrationReport>> MigrateDataGuidance(bool dryRun = true,
```
mmyoungman, line 38: Since this is ultimately going to be removed, I'd have put this all in the controller file, then you'd save yourself having to create the service, the interface, and adding the Startup.cs line. I'd have defined `DataGuidanceMigrationReport` in the controller file too.

Not saying change this now it's done - I think this is a super minor point - but I think it would have saved effort both in creating the migration endpoint and in removing it - and less chance something gets missed when removing it.

Although maybe it would have been more difficult to unit test...

mmyoungman (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataGuidanceMigrationService.cs
benoutram: I already saw that you did this in `MethodologyMigrationController` but it was no big deal to create the extra classes / interfaces and I've got comments on stuff to help identity what to remove later.

Part of the reason I did it in this way was to roll out our standard pattern of unit testing the service behaviour and adding a permissions test too.

PR #4431, opened 2023-11-23
EES-4627 Various clean up tasks associated with MethodologyRedirects

mmyoungman:
This ticket cleans up a couple of things after MethodologyRedirects have been deployed:
- We remove Methodology.Slug column. This was kept so that we could undo the MethodologyRedirect deploy more easily, but now it is successfully deployed it can be removed.
- Removed MethodologyMIgrationController#MigrateMethodologyLatestPublishedVersionIds
- Investigate a comment that referenced the non-existent ticket EES-2672. We weren't sure what EES-2672 was about, but did investigate whether the `Collection(...).LoadAsync()` call in the method refreshed `Methodology.Versions` if it had already been loaded. Via debugging, we established that it was being refreshed, and so we believe the comment can be safely removed.

### Other notes
- I've added an `Include` to the query in `GetLatestPublishedVersionBySlug`. I believe this not being done is a dormant bug, that would have been revealed should anyone call `result.Slug` on the result of that method. 
- The migration to remove Methodology.Slug also included other changes unrelated to this PR's code changes. I think these are probably associated with EES-4497, but I've only took a cursory glance. Since this PR is so small, I've just included them here rather than created a separate PR.


src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/MethodologyVersionRepository.cs
```diff
@@ -87,6 +86,7 @@ await _contentDbContext.Entry(methodology)
         {
             return await _contentDbContext
                 .MethodologyVersions
+                .Include(mv => mv.Methodology) // for MethodologyVersion.Slug
```
duncan-at-hiveit, line 89: I don't think this is necessary only because of the comment below, stating:

```
// EF cannot translate mv.Slug into a Queryable, so we have to do this...
```

PR #4446, opened 2023-11-30
EES-4732 fix slow locations rendering for schools

amyb-hiveit:
Fixes a bug in the table tool where loading school data in the locations step was very slow and could cause the browser tab to crash. 

School data is very large, 28,000+ schools, so we don't render the checkboxes for them until you search and only show a maximum of 500. Or so we thought... it turns out that there's an `!isMounted` check in the component which renders the checkboxes without search that was kicking in first. I'm assuming was added so the component will work without JS enabled (though I'm dubious if you can actually get to anywhere it's used on the site without JS).

This fix adds a check for the `searchOnly` flag to prevent this happening.

### DIFF
--- a/src/explore-education-statistics-common/src/components/form/FormCheckboxSearchGroup.tsx
+++ b/src/explore-education-statistics-common/src/components/form/FormCheckboxSearchGroup.tsx
@@ -20,13 +20,17 @@ export interface FormCheckboxSearchGroupProps extends FormCheckboxGroupProps {
 
 const FormCheckboxSearchGroup = ({
   searchLabel,
+  searchOnly = false,
   ...props
 }: FormCheckboxSearchGroupProps) => {
   const { isMounted } = useMounted();
 
   const [searchTerm, setSearchTerm] = useState('');
 
-  if (!isMounted) {
+  // searchOnly is used when we expect a very large number
+  // of options which would cause very slow rendering
+  // if they were all rendered.
+  if (!isMounted && !searchOnly) {
     return <FormCheckboxGroup {...props} />;
   }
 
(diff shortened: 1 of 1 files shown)


src/explore-education-statistics-common/src/components/form/FormCheckboxSearchGroup.tsx
```diff
@@ -20,13 +20,17 @@ export interface FormCheckboxSearchGroupProps extends FormCheckboxGroupProps {
...
 }: FormCheckboxSearchGroupProps) => {
   const { isMounted } = useMounted();
 
   const [searchTerm, setSearchTerm] = useState('');
 
-  if (!isMounted) {
+  // searchOnly is used when we expect a very large number
+  // of options which would cause very slow rendering
+  // if they were all rendered.
+  if (!isMounted && !searchOnly) {
```
duncan-at-hiveit, line 33: So do we not have any tests that would potentially catch this issue should it arise in the future?  E.g. a test that sets searchOnly to true and asserts that there are no checkboxes initially rendered?

src/explore-education-statistics-common/src/components/form/FormCheckboxSearchGroup.tsx
```diff
@@ -20,13 +20,17 @@ export interface FormCheckboxSearchGroupProps extends FormCheckboxGroupProps {
 
 const FormCheckboxSearchGroup = ({
```
duncan-at-hiveit, line 21: Do we want to have this as a hotfix against master?

src/explore-education-statistics-common/src/components/form/FormCheckboxSearchGroup.tsx
amyb-hiveit: We've already got tests for this, but they wouldn't pick up this issue as the tests run with JS so the component would be mounted almost instantly and the checkboxes would never be visible.

duncan-at-hiveit (approved):

PR #4462, opened 2023-12-06
README: Fix section anchors, update/clarify setup stages.

tomjonesdev:
Update the setup guide:
- corrected anchors to section headings/files/URLs
- add clarity to IdP, storage emulator, and seed data steps

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,11 +73,12 @@ To run applications in this service you will require the following:
 **Note** - Both .NET 6 and 8 are currently required as not all projects are using the latest SDK version.
 
 To run the databases:
-   - [Docker and Docker Compose](https://docs.docker.com/) - see [Setting up the database](#setting-up-the-database-and-storage-emulator)
+   - [Docker and Docker Compose](https://docs.docker.com/) - see [Set up your database](#set-up-your-database)
 
 To emulate Azure storage services (blobs, tables and queues) you will require one of the following options.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -209,22 +210,26 @@ The service can be started against a set of non-existent database. If no pre-exi
...
+    pnpm start admin
+```
+2. Directly via the start script using pnpm start dataStorage
+```bash
+    pnpm start dataStorage
+```
+3. Directly via Docker Compose
+```bash
+    docker-compose up data-storage
+```
```
ntsim, line 226: Formatting around this is a little messy:

image

Would expect it to look like this:

image

- Code blocks don't align correctly with the list items 
- No need to start code block contents with an indentation - code should as far left as possible
- Could space out the list items so the raw Markdown reads nicer

Not going to quote any code directly as it won't output correctly in GitHub, but something like this would be nicer:

image

README.md
```diff
@@ -209,22 +210,26 @@ The service can be started against a set of non-existent database. If no pre-exi
 4. Start the Admin project and this will configure the contained users' permissions via database migrations. 
    The other projects will then be able to be started, using their own contained users to connect to the databases. 
 
+### Setting up the storage emulator
+The Azurite Docker container can be started by one of the following methods:
+1. Indirectly by starting the admin via the start script
+```bash
+    pnpm start admin
+```
+2. Directly via the start script using pnpm start dataStorage
```
ntsim, line 219: No need to repeat the command here. Can simplify this to:

2. Directly via the `start` script using:

ntsim (approved):

PR #4474, opened 2023-12-13
EES-4748 Add API versioning with Swashbuckle integration

ntsim:
This PR:

- Adds API versioning in the form of the [ASP.NET API Versioning](https://github.com/dotnet/aspnet-api-versioning) package.
- Integrates the API versioning with Swashbuckle.
- Reconfigures Swashbuckle to gather documentation from XML docblocks. This is better for formatting summaries and descriptions in a nice way.

## Note

These changes splits out all Swashbuckle config into `Program`, instead of `Startup`. 

This is necessary to get access to some of the extensions on `WebApplication` (in the newer minimal API format), but also streamlines the configuration in `Startup` and avoids having this configuration in integration tests.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Config/SwaggerConfig.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Config/SwaggerConfig.cs
@@ -1,0 +1,103 @@
+using System.Text.Json;
+using Asp.Versioning.ApiExplorer;
+using Microsoft.AspNetCore.Mvc.ApiExplorer;
+using Microsoft.Extensions.Options;
+using Microsoft.OpenApi.Models;
+using Swashbuckle.AspNetCore.SwaggerGen;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Config;
+
+public class SwaggerConfig(IApiVersionDescriptionProvider provider) : IConfigureOptions<SwaggerGenOptions>
+{
+    public void Configure(SwaggerGenOptions options)
+    {
+        options.OperationFilter<SwaggerDefaultValues>();
+
+        var fileName = typeof(Program).Assembly.GetName().Name + ".xml";
+        var filePath = Path.Combine(AppContext.BaseDirectory, fileName);
+
+        options.IncludeXmlComments(filePath);
+        options.UseAllOfForInheritance();
+        options.UseOneOfForPolymorphism();
+        options.CustomOperationIds(apiDesc =>
+            {
+                var actionDescriptor = apiDesc.ActionDescriptor;
+
+                return actionDescriptor.AttributeRouteInfo?.Name
+                       ?? actionDescriptor.EndpointMetadata
+                               .OfType<IEndpointNameMetadata>()
+                               .LastOrDefault()
+                               ?.EndpointName
+                       ?? (apiDesc.TryGetMethodInfo(out var methodInfo) ? methodInfo.Name : null);
+            }
+        );
+
+        foreach (var description in provider.ApiVersionDescriptions)
+        {
+            options.SwaggerDoc(
+                description.GroupName,
+                new OpenApiInfo
+                {
+                    Title = "Explore education statistics - Public Data API",
+                    Version = description.ApiVersion.ToString(),
+                    Contact = new OpenApiContact
+                    {
+                        Name = "Explore education statistics",
+                        Email = "explore.statistics@education.gov.uk",
+                        Url = new Uri("https://explore-education-statistics.service.gov.uk")
+                    },
+                }
+            );
+        }
+    }
+
+    private class SwaggerDefaultValues : IOperationFilter
+    {
+        public void Apply(OpenApiOperation operation, OperationFilterContext context)
+        {
+            var apiDescription = context.ApiDescription;
+
+            operation.Deprecated |= apiDescription.IsDeprecated();
+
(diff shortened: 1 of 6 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Controllers/HelloWorldController.cs
```diff
@@ -1,22 +1,75 @@
...
 {
-    [HttpGet]
-    public async Task<ActionResult<string>> Get()
+    /// <summary>
+    /// List v1
+    /// </summary>
+    /// <remarks>
+    /// Test the API can list with `GET` in v1.
+    /// </remarks>
+    [HttpGet(Name = "List")]
```
jack-hive, line 24: Could we use `nameof` here? (same for other actions)

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Controllers/HelloWorldController.cs
```diff
@@ -1,22 +1,75 @@
...
+[Route("api/v{version:apiVersion}/HelloWorld")]
 public class HelloWorldController(PublicDataDbContext publicDataDbContext) : ControllerBase
 {
-    [HttpGet]
-    public async Task<ActionResult<string>> Get()
+    /// <summary>
+    /// List v1
+    /// </summary>
+    /// <remarks>
+    /// Test the API can list with `GET` in v1.
```
jack-hive, line 22: this would be for v2 too wouldn't it?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.csproj
```diff
@@ -5,9 +5,17 @@
         <Nullable>enable</Nullable>
         <ImplicitUsings>enable</ImplicitUsings>
         <RootNamespace>GovUk.Education.ExploreEducationStatistics.Public.Data.Api</RootNamespace>
+
+        <IncludeOpenAPIAnalyzers>true</IncludeOpenAPIAnalyzers>
+        <GenerateDocumentationFile>true</GenerateDocumentationFile>
+        
+        <!-- Disable warnings for undocumented methods -->
+        <NoWarn>1591</NoWarn>
```
jack-hive, line 13: Do we want this? If it applies to only controller actions then maybe we want the warning if we have missed the documentation?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Config/SwaggerConfig.cs
```diff
@@ -0,0 +1,92 @@
+using System.Text.Json;
+using Asp.Versioning.ApiExplorer;
+using Microsoft.AspNetCore.Mvc.ApiExplorer;
+using Microsoft.Extensions.Options;
+using Microsoft.OpenApi.Models;
+using Swashbuckle.AspNetCore.SwaggerGen;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Config;
+
+public class SwaggerConfig(IApiVersionDescriptionProvider provider) : IConfigureOptions<SwaggerGenOptions>
```
jack-hive, line 10: how does this provider get injected in?

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.csproj
ntsim: We need this otherwise the build logs start getting spammed with warnings like the following for undocumented methods and classes:

```
/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/YourClass.cs(11,14): warning CS1591: Missing XML comment for publicly visible type or member 'YourMethod'
```

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Controllers/HelloWorldController.cs
ntsim: This was just for illustrating how to set the endpoint's documented operation ID. However, I've changed this behaviour so that it:

1. Tries to use route name defined by `Http*` or `Route` attribute (i.e. the above)
2. If no route name, fallback to the method name itself

This means that in most cases, the method name by itself will suffice (no need for the name in an attribute)

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Controllers/HelloWorldController.cs
ntsim: Nope, it gets superceded in v2 by the `ListV2` method!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Config/SwaggerConfig.cs
ntsim: This is registered in the DI container as:

```cs
builder.Services.AddTransient<IConfigureOptions<SwaggerGenOptions>, SwaggerConfig>();
```

So it automatically wires it in for us :smile:

PR #4478, opened 2023-12-14
EES-3789 Introducing publication redirects

mmyoungman:
This PR allows users to change live publication slugs, and creates `PublicationRedirects` for those slugs that were previously live.

This builds on the work of EES-4482 which introduced methodology redirects. The two pieces of work are strongly linked due to a methodology's ability to inherit a publication's slug. There also complications due to a methodology slugs not being changed immediately, but slugs being attached to versions and going live when that particular version is published - although they are changed immediately by necessity if a methodology inherits a publication slug.

The work has been done such that while everything should still work if there are redundant redirects - i.e. there is a redirect to the currently used slug, or duplicate redirects to the same slug - the redirect database entries shouldn't get into this state. If they do, then I think we should treat it as symptomatic of a bug. 

### To person from the future

The approach I took with both this work and the methodology redirects work was to avoid making changes to the database at publishing time around redirects. I put my money on that being the better approach, but now the implementation is complete, I'm no longer so sure. In an ideal world, I'd try switching to that approach, but alas time isn't an infinite resource. While I've attempted to simplify the code as much as possible, I still think this code will be hard to adapt or work around in the future if we need to. If you're reading this now, person in the future, and that's what you're trying to do, then you may consider adapting the methodology redirects code so that the `MethodologyRedirects` table is updated on publishing - although it's possible that that approach has it's own problems that aren't immediately obvious.

This PR also allows publications and methodologies to reuse a slug that they previously used - i.e. a methodology could change it's slug from `methodology-a` to `methodology-b` and then back to `methodology-a`. Another potential simplification would be to not allow this.

### Other changes

- The `PublicationTitleOrSlugChanged` method has moved into Admin's `MethodologyService`. It previously lived in Content's `MethodologyService`.
- I've created validation errors `MethodologySlugNotUnique` and `PublicationSlugNotUnique`. This is for use with `ValidatePublicationSlug` - the previous validation error "SlugNotUnique` would have been ambiguous.
- Increase `AssertUtcNow`'s `withinMillis` from 500 to 750. At 500, it failed once for me while running the unit tests locally.
- I've fixed a bug in the previous MethodologyRedirect work: previously, if a Methodology which inherits a publication title is published, and then has an amendment created with an `AlternativeSlug`, if someone then changes the inherited publication title, when the amendment is published there will be no redirect for the new inherited methodology slug.


N-moh (commented):
Is GA covered with this?

src/explore-education-statistics-admin/src/pages/publication/components/PublicationForm.tsx
```diff
@@ -25,7 +25,12 @@ const errorMappings = [
   mapFieldErrors<FormValues>({
     target: 'title',
     messages: {
-      SlugNotUnique: 'Choose a unique title',
+      PublicationSlugNotUnique: 'Choose a unique title',
+      PublicationSlugUsedByRedirect: 'Choose a previously unused title',
+      MethodologySlugNotUnique:
+        "As this publication's owned methodology inherits this publication's title, this action changes the owned methodology. Choose a unique methodology title",
```
duncan-at-hiveit, line 31: Just wondering if these messages need any vetting from Cam?  I'd be tempted to try something a little less wordy, like "This title also changed the methodology title to one used by another methodology. Choose a unique methodology title".  Same for the one below.  Not a big deal if we don't change it though!

duncan-at-hiveit:
Nice PR!  Great work in a very tricky area.  Only other thing to say is that we could do with a UI test to cover this, as we do for methodologies in `publish_methodology.robot`.

src/explore-education-statistics-admin/src/pages/publication/components/PublicationForm.tsx
duncan-at-hiveit: FYI I spoke with @cjrace and we confirmed wording should read:

"This title also changed the methodology title to one used by an existing methodology. Choose a unique methodology title"

duncan-at-hiveit:
> Nice PR! Great work in a very tricky area. Only other thing to say is that we could do with a UI test to cover this, as we do for methodologies in `publish_methodology.robot`.

FYI I spoke to @saicharan2789 and @N-moh about this and Sai is on with adding a redirect UI test case for Publications.

src/explore-education-statistics-admin/src/pages/publication/components/PublicationForm.tsx
mmyoungman: I changed that and another message, so I ran them all by Cam again and he gave it the :+1:

duncan-at-hiveit: Nice, ta!

duncan-at-hiveit (approved):

PR #4482, opened 2023-12-15
EES-4227 Add file name length validation to file uploads

tomjonesdev:
Limit the number of characters file names can have when uploading individual files and archives, and provide feedback to the user if their files exceed this limit.

This protects end users from hitting errors from having file names that are too long, for example there have been examples in the past where users can’t unzip files as the path is too long.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataArchiveValidationService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataArchiveValidationService.cs
@@ -19,6 +19,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
     {
         private readonly IFileTypeService _fileTypeService;
 
+        private const int MaxFilenameSize = 150;
+
         private static readonly Dictionary<FileType, IEnumerable<Regex>> AllowedMimeTypesByFileType =
             new()
             {
@@ -56,6 +58,16 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
             var dataFile = file1.Name.Contains(".meta.") ? file2 : file1;
             var metaFile = file1.Name.Contains(".meta.") ? file1 : file2;
 
+            var filenamesValid = ValidateFilenameLengths(
+                zipFile.FileName.Length,
+                dataFile.Name.Length,
+                metaFile.Name.Length);
+
+            if (filenamesValid.IsLeft)
+            {
+                return filenamesValid.Left;
+            }
+
             return new DataArchiveFile(dataFile: dataFile, metaFile: metaFile);
         }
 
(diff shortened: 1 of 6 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataArchiveValidationService.cs
```diff
@@ -56,6 +58,17 @@ public DataArchiveValidationService(IFileTypeService fileTypeService)
             var dataFile = file1.Name.Contains(".meta.") ? file2 : file1;
             var metaFile = file1.Name.Contains(".meta.") ? file1 : file2;
 
+            var filenamesValid = ValidateFilenameLengths(
+                zipFile.FileName.Length,
+                dataFile.Name.Length,
+                metaFile.Name.Length,
+                out var isValid);
+
+            if (!isValid)
```
mmyoungman, line 67: Not a big deal, but you can use `filenamesValid.IsLeft` here instead of `!isValid`, and so remove the need for an out param.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataArchiveValidationService.cs
```diff
@@ -72,5 +85,37 @@ private async Task<bool> IsZipFile(IFormFile file)
                    )
                    && _fileTypeService.HasMatchingEncodingType(file, ZipEncodingTypes);
         }
+
+        private static Either<ActionResult, Unit> ValidateFilenameLengths(
+            int zipFilenameLength,
```
mmyoungman, line 90: I'd be inclined to make one instance of `ValidateFilenameLengths` and make `zipFilenameLength` an optional argument that defaults to null, rather than having two instances of it. Then we can also have one instance of `MaxFilenameSize` and avoid any mishaps with someone changing one without realising they need to change the other.

It's possible there is something I don't understand here with one `ValidateFilenameLengths` method returning a list of errors while the other only returns a single error.

If we do only have one instance of `ValidateFilenameLengths`, there is the question of where it should live. I'm not fussed - in either service is fine by me - but others might have an opinion on it. Could argue it should be in another service so it can be unit tested separately - that seems like overkill to me.

src/explore-education-statistics-admin/src/pages/release/data/components/DataFileUploadForm.tsx
```diff
@@ -37,6 +39,10 @@ function baseErrorMappings<FormValues extends DataFileUploadFormValues>(
           DataZipFileDoesNotContainCsvFiles:
             'ZIP file does not contain any CSV files',
           DataFilenameNotUnique: 'Choose a unique ZIP data file name',
+          DataZipFilenameTooLong: `Maximum ZIP data filename cannot exceed ${MAX_FILENAME_SIZE} characters`,
+          DataFilenameTooLong: `Maximum data filename cannot exceed ${MAX_FILENAME_SIZE} characters`,
+          MetaFilenameTooLong: `Maximum metadata filename cannot exceed ${MAX_FILENAME_SIZE} characters`,
+          DataZipContentFilenamesTooLong: `Maximum data and metadata filenames cannot exceed ${MAX_FILENAME_SIZE} characters`,
```
mmyoungman, line 45: You collect all errors in one `ValidateFilenameLengths` method but then create `DataZipCOntentFilenamesTooLong` for the other. Is there a reason why, or maybe `DataZipContentFilenamesTooLong` can be removed?

mmyoungman, line 45: I think it might be possible to validate this on the frontend as well via Formik's `validationSchema`? I'm not sure though, and to be honest I'm fine if we leave it as is.

mmyoungman (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataArchiveValidationService.cs
tomjonesdev: Since the zip file is uploaded from a single form field, it can only display a single error message - there's currently no logic to handle a list on the front-end (e.g. combine and comma separate) - whereas the individual files can return an error message for each field.

I tried a few variations to combine the logic into a single method, but they weren't as readable and decided against it.

src/explore-education-statistics-admin/src/pages/release/data/components/DataFileUploadForm.tsx
tomjonesdev: The zip file field can only handle a single error message response (any other items returned in a list are ignored), but the error may relate to both files contained within the archive (data and metadata).

`DataZipContentFilenamesTooLong` combines validation for both content files and ensures the user only has to resubmit the form once if both filenames are too long (as opposed to being told the data filename is too long, submitting, then being told the meta filename is too long).

tomjonesdev: There's a task being worked on at the moment to gradually replace Formik across the codebase, so on this occasion it's better to leave as-is to avoid wasted effort.

mmyoungman (approved):

PR #4483, opened 2023-12-15
EES-4194 : Add playwright test scripts to 'Glossary page'

saicharan2789:
This PR focusses on converting the 'Glossary page' test suite exist in the 'Robot framework' into playwright tests.

Following is the JIRA link:
[[https://dfedigital.atlassian.net/browse/EES-4194?atlOrigin=eyJpIjoiNDBjZTBjMmIyNTYzNDFiM2IyZjdmOGUzNDVmNjBhOGIiLCJwIjoiamlyYS1zbGFjay1pbnQifQ]]


tests/playwright-tests/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
```
duncan-at-hiveit, line 1: As we perform the rest of our builds and unit test runs in Azure DevOps, it'd be nice to have these incorporated in DevOps too.  I'm wondering what the effort is of creating a YAML pipeline for doing the same as this but in an Azure DevOps-compatible YAML file?

See `azure-pipelines.dfe.yml` and search for `pnpm test:ci` for an example of where we're running unit tests against a TS project.

Not suggesting that we need to do this as part of this PR, just interested to know, and if we adopt Playwright going forwards, Azure DevOps pipelines would be how we'd want to support them.

tests/playwright-tests/README.md
```diff
@@ -0,0 +1,45 @@
+**How to install Playwright and set up VS Code?**
```
duncan-at-hiveit, line 1: Moving forwards, we'll also want to incorporate our other widely-used IDEs in here too e.g. Rider. But someone who uses them mostly can pick that part up :slightly_smiling_face:

tests/playwright-tests/general-public/pages/glossary-page.ts
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable import/prefer-default-export */
+/* eslint-disable lines-between-class-members */
+import { Locator, Page } from '@playwright/test';
+
+// Glossary page
+export class GlossaryPage {
+  readonly page: Page;
+  readonly page_search_box: Locator;
```
duncan-at-hiveit, line 8: I don't know if there are established variable name styles that are specific to Playwright or if there's a convention of snakecase variables where you've worked before?

If not, we tend to use camelCase variable names in Typescript projects.

So for instance, `page_search_box` would be `pageSearchBox` instead.

tests/playwright-tests/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "playwright-tests",
+  "version": "1.0.0",
+  "description": "**How to install Playwright and set up VS Code?**",
```
duncan-at-hiveit, line 4: Is this the right description?

tests/playwright-tests/playwright.config.ts
```diff
@@ -0,0 +1,78 @@
...
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  /* Opt out of parallel tests on CI. */
+  workers: process.env.CI ? 1 : undefined,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter: 'html',
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    /* Base URL to use in actions like `await page.goto('/')`. */
+    baseURL: 'https://dev.explore-education-statistics.service.gov.uk/',
```
duncan-at-hiveit, line 28: Do you override this in a .env file if you want to run locally?  If so, it would be worth adding some details in the README :slightly_smiling_face:

tests/playwright-tests/general-public/pages/glossary-page.ts
saicharan2789: Sure I will update that accordingly

tests/playwright-tests/playwright.config.ts
saicharan2789: Apologies, I haven't used this one , as I understand sensitive information shouldn't exposed towards outside world. Will maintain everything via .env file

tests/playwright-tests/README.md
N-moh: I managed to set it up in Pycharm . 
Happy to update the Readme as a separate PR after this is merged.
However I could not yet setup a Playwright test runner.

tests/playwright-tests/playwright.config.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
+    trace: 'on',
```
N-moh, line 31: The trace feature is awesome in pointing out the exact failure
image

N-moh (commented):
Impressed with the speed of runs and the interactive debug mode. The setup in Pycharm was fairly simple. 
The page object modal may lead to creation of too many files covering thousands of test cases in hundreds of user journeys which could lead to a complex test result report in the end.
But it works really well.
Well done!

tests/playwright-tests/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
...
+on:
+  push:
+    branches: [ main, master ]
+  pull_request:
+    branches: [ main, master ]
+jobs:
+  test:
+    timeout-minutes: 60
```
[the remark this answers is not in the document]
saicharan2789: Thanks @N-moh  for your  insights. As part of this PR, I have focussed only on the scripting work.. However, I will make changed to 2 minutes or anything else based on your inputs.

tests/playwright-tests/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 18
+    - name: Install dependencies
+      run: npm ci
```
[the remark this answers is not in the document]
saicharan2789: we can take a look into this  once, when we create new PR while we focus on .yml file

tests/playwright-tests/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
...
+    - name: Install Playwright Browsers
+      run: npx playwright install --with-deps
+    - name: Run Playwright tests
+      run: npx playwright test
+    - uses: actions/upload-artifact@v3
+      if: always()
+      with:
+        name: playwright-report
+        path: playwright-report/
+        retention-days: 30
```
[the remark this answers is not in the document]
saicharan2789: Commented in the above

tests/playwright-tests/README.md
```diff
@@ -0,0 +1,47 @@
...
+* VS Code Editor: Download and Install VS code from 
+https://code.visualstudio.com/download
+* Playwright VS Code plugin
+* ES-lint extension
+* Playwright test runner
+
+
+# Installation
+
+Step 1: Navigate to 'Playwright-tests' folder
```
duncan-at-hiveit, line 17: Would suggest making use of Markdown's numbered list feature here.

Rather than needing to manually say "Step 1:", "Step 2:" etc, if you replace every "Step X:" with "1. " instead, so it would look like:

```
1. Navigate to 'Playwright-tests' folder
1. Ensure the NodeJS version is above 14 from the command prompt.  Blah blah.
    node -v
1. Launch the VS code
etc etc
```

Note all of these lines begin with "1. ", not "1. ", "2. ", "3. " etc etc.  This is so that you can slot in new items anywhere in the list but you don't have to spend ages updating the numbers.  By putting them all as "1"s, it automatically spits out a correctly numbered list.

Also, where we include script snippets in READMEs, we tend to put them in a code block to show it's a script or a command to run e.g:

node -v

would become:

~~~
```bash
node -v
```
~~~

which in the README, would then look like:

```bash
node -v
```

tests/playwright-tests/utils/env.ts
```diff
@@ -0,0 +1,7 @@
+import dotenv from 'dotenv';
```
duncan-at-hiveit, line 1: This is great - I would just add a mention of having to add a .env file to the README if it's mandatory to do so.

We generally check in a `.env.example` file with local settings in so that people can just copy that to a new `.env` file - we then add something like the following to `.gitignore`:

```
.env
.env.*
!.env.example
```

For Robot, for example, we have these lines in `.gitignore` in the root of the project:

```
/tests/robot-tests/.env.*
!/tests/robot-tests/.env.example
```

tests/playwright-tests/utils/env.ts
```diff
@@ -0,0 +1,7 @@
...
+
+dotenv.config();
+// eslint-disable-next-line import/prefer-default-export
+export class environment {
+  public static BASE_URL = process.env.PUBLICURL;
```
duncan-at-hiveit, line 6: Tiny thing, but generally we use underscores between words in these files, so that `PUBLICURL` becomes `PUBLIC_URL`, much in the same way that you have `BASE_URL` already.

tests/playwright-tests/general-public/pages/glossary-page.ts
```diff
@@ -0,0 +1,39 @@
...
+    this.accordionSectionC = page.locator('//span[text()="C"]');
+    this.accordionSectionD = page.locator('//span[text()="D"]');
+    this.accordionSectionE = page.locator('//span[text()="E"]');
+    this.accordionSectionZ = page.locator('//span[text()="Z"]');
+    this.voluntaryRepaymentSection = page.locator('//h3[text()="Voluntary repayment"]');
+    this.voluntaryRepaymentSectionText = page.locator('//div[@id="voluntary-repayment"]//p[1]');
+    this.educationStatisticsMethodology = page.locator('//a[text()="Education statistics: methodology"]');
+  }
+  // standard text
+  readonly voluntary_repayment_section_paragraph_text = 'A borrower can at any time choose to repay some or all of their loan balance early, in addition to any repayments they are liable to make based on their income';
```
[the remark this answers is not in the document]
saicharan2789: Sorry, its my mistake, Will update that accordingly.

duncan-at-hiveit:
> Impressed with the speed of runs and the interactive debug mode. The setup in Pycharm was fairly simple. The page object modal may lead to creation of too many files covering thousands of test cases in hundreds of user journeys which could lead to a complex test result report in the end. But it works really well. Well done!

I think that the idea of the POM is that we'd reuse the page objects to interact with various UI elements on each page.  So we'd hopefully just end up with 1 file per page (or thereabouts).

Where we have specific variables / methods here to support a single test case, we'd probably find another way of supporting that as this was used more and more often I'd expect, and these particular expectation or test-specific selectors would actually disappear form here and be put more in control of the calling test itself.

tests/playwright-tests/utils/env.ts
duncan-at-hiveit: Also, I suspect it'll currently fail in the pipeline because it doesn't have a .env file from which to construct its BASE_URL, so maybe have a default fallback here like:

```
public static BASE_URL = process.env.PUBLIC_URL ?? "https://dev.explore-education-statistics.service.gov.uk/";
```

duncan-at-hiveit: Although this'll probably fail right now too because of basic auth.  We'd probably need to set up some pipeline variables which aren't checked in that you could then combine with the URL above to get past Dev's basic auth.

tests/playwright-tests/utils/env.ts
saicharan2789: Hi @duncan-at-hiveit - Thanks for spotting this one:).  Could you please help me if you would like me to put any variables inside the '.env.example' file

duncan-at-hiveit (approved):

PR #4484, opened 2023-12-15
EES-4730 Send release published notification emails to superseded pub…

mmyoungman:
…lication subscribers

This PR makes changes to who is emailed when a release is published. We now send emails to subscribers of any publications superseded by the publication of the release being published.

### Notes

- There is some faffy deploy steps that are covered by ticket EES-4764 - specifically adding the new template IDs to the infra release pipeline's override template parameters before this is deployed. This cannot be done in advance, as the infra release pipeline would fail until this work is deployed.

- I tested this by sending 10,000 fake emails locally  - I didn't find any issues. This was mainly done to ensure there are no issues around the `sentToEmailAddresses` HashSet if it grew large.

- While publication's on the service aren't expected to supersede more than one other publication, the service supports having multiple publications supersede a publication. This is why we need to pass the array `SupersededPublicationIds ` to `Notifier`.

### DIFF
--- a/infrastructure/templates/template.json
+++ b/infrastructure/templates/template.json
@@ -325,6 +325,20 @@
         "description": "Gov UK Notify service template Id for amended release notification"
       }
     },
+    "releaseEmailSupersededSubscribersTemplateId": {
+      "type": "securestring",
+      "defaultValue": "",
+      "metadata": {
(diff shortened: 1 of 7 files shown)


mmyoungman:
This PR isn't ready yet

PR #4486, opened 2023-12-20
EES-4751 Various public data API schema updates

ntsim:
This PR makes various tweaks to the `PublicDataDbContext` used by the public data API:

- New `PrivateId` and `PublicatId` properties added to location and filter option meta for use in querying Parquet data more efficiently
- `DataSetVersion.MetaSummary` is now required
- New `TotalResults` column added to `DataSetVersion`
- Renamed connection string name to `PublicDataDb`


jack-hive (approved):
Looks good! Already discussed the questions I had around the new fields :)

PR #4489, opened 2023-12-21
EES-4751 Refactor entity configs into separate classes

ntsim:
:warning: Depends on #4486 

This PR refactors `PublicDataDbContext` entity configs  into nested config classes on the respective entities. This change follows the style guide meeting where we discussed this as a good middle ground between using attributes and the fluent API.


jack-hive (approved):
Looks good! :)

PR #4494, opened 2024-01-02
EES-4730 On publishing, also email superseded publication subscribers

mmyoungman:
This PR changes who is emailed when a release is published. We now send additional emails to subscribers of any publications superseded by the publication of the release being published.

### Notes
- There is some faffy deploy steps that are covered by ticket EES-4764 - specifically adding the new template IDs to the infra release pipeline's override template parameters before this is deployed. This cannot be done in advance, as the infra release pipeline would fail until this work is deployed.

- The `Notifier.Tests` project has been created as part of this work.

- To allow testing of Notifier, multiple parts of the code have been pulled out into their own service, allowing the tests to inject mocks where necessary.

- I also manually tested this by sending 10,000 fake emails locally, using a GOV.UK Notify fake API key - I didn't see any issues.

- While publication's on the service aren't expected to supersede more than one other publication, the service supports having multiple publications supersede a publication. This is why we need to pass the array `SupersededPublicationIds` to Notifier.

- `Microsoft.WindowsAzure.Storage.Table` is deprecated, so I've moved to `Microsoft.Azure.Cosmos.Table`. No other code changes were required except to change the import.


src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/ReleaseNotifier.cs
```diff
@@ -0,0 +1,184 @@
...
+                    }
+                } while (token != null);
+
+                logger.LogInformation("Emailed {NumEmailsSent} subscribers from a superseded publication",
+                    numSupersedeSubscriberEmailsSent);
+            }
+
+            logger.LogInformation("Sent {NumEmailsSent} emails in total to subscribers",
+                sentToEmailAddresses.Count);
+        }
```
benoutram, line 182: I'm not keen on this function because I think it has too many responsibilities now. It's got quite long to read.

I've not looked at it too closely but feels like there's a common method to be extracted between the regular and superseded functionality.

I think when we discussed the approach for this I mentioned my plan would be to have separate functions for the different cases, so there would be a function for sending the publication emails, and a function for sending the superseded emails. Each function would know about its own template id and have the responsibility for querying subscribers for the publication given in the message and building the appropriate email template values for that template. I think there would still be some common code reused between them for running the query.

I think it makes it a bit cleaner with it being clear which template each function is using and we know why each function has been invoked. It should be easier to unit test each function rather than one big one. If we also had variants of the functions for amendments we'd get rid of the problem where we set the `update_note` variable when it's not always relevant.

I don't have a preference on doing this though and it could be argued that it puts too much responsibility on the publisher to queue the correct type and number of messages but I don't see that as a big overhead.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
```diff
@@ -67,6 +67,10 @@ await _context.Entry(release)
                 latestUpdateNoteReason = latestUpdateNote.Reason;
             }
 
+            var supersededPublications = _context.Publications
+                .Where(p => p.SupersededById == release.PublicationId)
+                .Select(p => new { p.Id, p.Title })
+                .ToList();
```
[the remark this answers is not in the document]
mmyoungman: I'm not sure what benefit async gives us in this kind of situation - to be honest, I don't think it's much if anything, and may even be a disadvantage in performance. I've added it.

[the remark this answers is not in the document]
mmyoungman: I don't think so. A publication can only have one `SupersededById` set. There is a chance that it could fetch multiple different publications - i.e. a publication supersedes multiple publications - but that case is handled.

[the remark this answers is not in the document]
benoutram: Ah yep, I see now, they would all be different publications if there's more than one result. 👍

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/NotificationsService.cs
```diff
@@ -77,6 +81,8 @@ await _context.Entry(release)
                 ReleaseSlug = release.Slug,
                 Amendment = release.Version > 0,
                 UpdateNote = latestUpdateNoteReason,
+                SupersededPublicationIds = supersededPublications.Select(p => p.Id).ToList(),
+                SupersededPublicationTitles = supersededPublications.Select(p => p.Title).ToList(),
```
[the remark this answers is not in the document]
mmyoungman: I've used `IdTitleViewModel` - which I've moved into Common.

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Utils/NotifierTestUtils.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Threading.Tasks;
+using Moq;
+using Moq.Protected;
+using Notify.Client;
+
+namespace GovUk.Education.ExploreEducationStatistics.Notifier.Tests.Utils;
+
+public class NotifierTestUtils
+{
+    private const string notificationKey =
```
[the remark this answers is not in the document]
mmyoungman: Fixed the casing.

The key does need to be like that. When generating the mock NotificationClient, it does do some validation on the notify key, so it needs to be `text-guid-guid` to be valid, I believe.

src/GovUk.Education.ExploreEducationStatistics.Notifier/Subscription.cs
```diff
@@ -1,12 +1,13 @@
+#nullable enable
 using System;
-using Microsoft.WindowsAzure.Storage.Table;
+using Microsoft.Azure.Cosmos.Table;
```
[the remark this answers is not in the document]
mmyoungman: I've updated EES-4771 to now be to update to using `Azure.Data.Tables`

src/GovUk.Education.ExploreEducationStatistics.Notifier/Services/IStorageTableService.cs
```diff
@@ -1,5 +1,6 @@
+#nullable enable
```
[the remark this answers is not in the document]
mmyoungman: I tried it and it was relatively painless, so I've made the change. The only real changes I needed to make were to `SubscriptionManager.cs` - and that was replacing `.Result` with `await`s, which we should be doing anyway.

src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/ReleaseNotifier.cs
```diff
@@ -0,0 +1,184 @@
...
+                            },
+                            { "unsubscribe_link", $"{baseUrl}{supersededPublicationId}/unsubscribe/{unsubscribeToken}" },
+                            { "superseded_publication_title", supersededPublicationTitle },
+                        };
+
+                        _emailService.SendEmail(notificationClient, email, newReleaseSupersededSubscribersTemplateId, values);
+                    }
+                } while (token != null);
+
+                logger.LogInformation("Emailed {NumEmailsSent} subscribers from a superseded publication",
```
luke-h1, line 176: Just leaving a minor comment as I was thinking in general about future improvements to monitoring: Would be great to alert / trend data based on either:
*  expectations were thrown on failures around sending notifications (from the EES notifier side)
* upstream errors calling the notify service happened (i.e. the gov uk notify client was in the middle of an incident and failed to deal with EES's API request)

src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/ReleaseNotifier.cs
```diff
@@ -0,0 +1,184 @@
...
+                                // NOTE: update_note not used by original release email template
+                                "update_note", msg.UpdateNote
+                            },
+                            { "unsubscribe_link", $"{baseUrl}{supersededPublicationId}/unsubscribe/{unsubscribeToken}" },
+                            { "superseded_publication_title", supersededPublicationTitle },
+                        };
+
+                        _emailService.SendEmail(notificationClient, email, newReleaseSupersededSubscribersTemplateId, values);
+                    }
+                } while (token != null);
```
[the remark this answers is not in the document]
mmyoungman: It's not my code - for whatever reason, it's got confused about the file moving directory and is treating it all as changed.

Hope you're well :)

src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/ReleaseNotifier.cs
mmyoungman: I think Ben has raised before that we currently don't wait for a response from Notify on whether the email was sent successfully - we just trust it to do so. Although I just looked if we had a ticket for it - I suspected we would - but couldn't find it.

A lot of this Notifier code just hasn't been touched in ages - and I think that is because despite it's problems, it's worked well enough, so improving it hasn't been a priority.

src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/ReleaseNotifier.cs
```diff
@@ -0,0 +1,178 @@
...
+
+        TableContinuationToken? token = null;
+        do
+        {
+            var resultSegment =
+                await table.ExecuteQuerySegmentedAsync(query, token);
+            token = resultSegment.ContinuationToken;
+
+            logger.LogInformation("Emailing {SubscriberCount} subscribers",
+                resultSegment.Results.Count);
```
benoutram, line 168: You don't need this here so this method probably doesn't neeed `logger` at all anymore.

src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/ReleaseNotifier.cs
```diff
@@ -0,0 +1,178 @@
...
+                },
+                { "unsubscribe_link", $"{baseUrl}{msg.PublicationId}/unsubscribe/{unsubscribeToken}" },
+            };
+
+            _emailService.SendEmail(notificationClient, email, newReleaseTemplateId, values);
+        }
+        var sentEmails = releaseSubscriberEmails.ToHashSet();
+
+        // Send emails to subscribers of any associated superseded publication
+        var newReleaseSupersededSubscribersEmailTemplateId = msg.Amendment
```
benoutram, line 101: Just wondering why this is prefixed with `new`?  My mind is trying to process if it's a temporary name that's accidently been left in because it's a 'new' variable, or if it's because it's a 'new release' or because it's a 'new template id'. 😆

mmyoungman: It was for "new release" -  not sure if this was me originally or not - but I'll remove it either way :)

mmyoungman:
Been discussing this in Slack with Ben. Ben raised that there might be a default retry if an exception is thrown in `ReleaseNotifierFunc`. I tested it and he was right. So I've added `"maxDequeueCount": 1` to Notifier's `host.json` to ensure we don't send duplicate emails in the case of an exception being thrown.

src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/ReleaseNotifier.cs
mmyoungman: There is now a ticket to pretty much rewrite the Notifier project - although you cannot see it as you're not working on this project anymore :stuck_out_tongue:

src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/ReleaseNotifier.cs
mmyoungman: Changes have been made - most notable created methods `GetSubscriberEmails`/ `SendSubsciberEmail`/`SendSupersededSubscriberEmail`

src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/ReleaseNotifier.cs
```diff
@@ -0,0 +1,204 @@
...
+            { "publication_name", msg.PublicationName },
+            { "release_name", msg.ReleaseName },
+            {
+                "release_link",
+                $"{webApplicationBaseUrl}find-statistics/{msg.PublicationSlug}/{msg.ReleaseSlug}"
+            },
+            {
+                // NOTE: update_note not used by original release email template - only with amendments
+                "update_note", msg.UpdateNote
+            },
```
benoutram, line 157: Would be a quick improvement to only add this to the dictionary if the amendment flag is true, removing the need for the comment.

There's one other case of this as well.

src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/ReleaseNotifier.cs
```diff
@@ -0,0 +1,204 @@
...
+                subscribersTable, releaseSupersededPubSubsQuery);
+
+            var numSupersededSubscriberEmailsSent = 0;
+
+            foreach (var email in supersededPublicationSubscriberEmails)
+            {
+                if (sentEmails.Contains(email))
+                {
+                    continue;
+                }
```
benoutram, line 97: Do you know if all subscriber email addresses are held in a certain case, because this check will be case sensitive resulting in subscribers potentially getting more than one email if so.

src/GovUk.Education.ExploreEducationStatistics.Notifier.Model/GovUk.Education.ExploreEducationStatistics.Notifier.Model.csproj
```diff
@@ -8,4 +8,8 @@
     <ItemGroup>
         <PackageReference Include="IsExternalInit" Version="1.0.3" PrivateAssets="all" />
     </ItemGroup>
+
+    <ItemGroup>
+      <ProjectReference Include="..\GovUk.Education.ExploreEducationStatistics.Common\GovUk.Education.ExploreEducationStatistics.Common.csproj" />
+    </ItemGroup>
```
benoutram, line 14: Would be interesting to know if this has pumped up the size of the deployment package significantly or not.

This does have a lot of dependencies that won't be relevant to the Notifier, for the sake of sharing the `IdTitleViewModel`.

We will probably benefit from using more shared code in future too so I don't think it's an issue but just asking more out of curiosity.

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/ReleaseNotifier.cs
mmyoungman: Good catch.

Discovered you can just do this: `new HashSet<string>(StringComparer.OrdinalIgnoreCase);`

mmyoungman:
Final push was to resolve conflict with Gov.UK.EES.Blah.sln file.

PR #4497, opened 2024-01-02
EES-3503: Increase max file upload size to 2GB

tomjonesdev:
The file limit of 100mb was found too restrictive for certain users when uploading ancillary data. These changes increase the limit to 2GB, and provide helpful messaging for information and validation.

- Client-side validation halts submission of files exceeding the limit before any transfer takes place
- If client validation were to fail, a file upload which exceeds the web server limit would result in a HTTP 400 (Bad Request), due to the endpoint restriction of int.MaxValue, which is ~2GB in bytes
- If the web server limit is increased (and the controller attributes removed, or the limit increased), and a file larger than 2GB is uploaded, the custom server-side validation will kick in and provide similar error feedback to the UI.

image

File streaming and chunking were considered as options, but due to the infrequency of large file uploads, were deemed excessive as the work would require quite substantial refactoring of various validation mechanisms.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/FileUploadsValidatorService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/FileUploadsValidatorService.cs
@@ -27,6 +27,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
         private readonly ContentDbContext _context;
 
         private const int MaxFilenameSize = 150;
+        private const int MaxFileSize = int.MaxValue; // 2GB
 
         public FileUploadsValidatorService(IFileTypeService fileTypeService,
             ContentDbContext context)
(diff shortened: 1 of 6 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/FileUploadsValidatorService.cs
```diff
@@ -74,6 +74,11 @@ public class FileUploadsValidatorService : IFileUploadsValidatorService
             {
                 return ValidationActionResult(FileCannotBeEmpty);
             }
+            
+            if (file.Length > 2147483648)
```
duncan-at-hiveit, line 78: Don't know if it's perhaps worth sticking this in as a constant to make its choice a bit more obvious perhaps?  Like `TWO_GB" or similar?

src/explore-education-statistics-admin/src/pages/release/data/components/AncillaryFileForm.tsx
```diff
@@ -113,6 +114,7 @@ export default function AncillaryFileForm({
         summary: Yup.string().required('Enter a summary'),
         file: Yup.file()
           .minSize(0, 'Choose a file that is not empty')
+          .maxSize(2147483648, 'Choose a file that is under 2GB')
```
duncan-at-hiveit, line 117: I would have said a similar comment to the C# code and putting this in a named constant, but given that it's right next to the error message that tells you what's going wrong, I think we're good here :slightly_smiling_face:

duncan-at-hiveit (approved):

PR #4504, opened 2024-01-05
EES-4772 - Rewrite admin Robot test in playwright

saicharan2789:
New implementation of playwright script which focusses on one of the admin specific flow:

- Rewritten the creation of themes specific robot TC into playwright one.
-  For login specific stuff -  implemented using Azure AD


tests/playwright-tests/admin/azpage/AzLoginPage.ts
```diff
@@ -0,0 +1,35 @@
+import { Locator, Page } from '@playwright/test';
+import  environment  from '../../utils/env';
+
+// AzureLogin Page
```
ntsim, line 4: Could remove this comment? The class already says what it is, so this doesn't add anything?

We could get rid of these comments on the other page classes as well!

tests/playwright-tests/admin/azpage/AzLoginPage.ts
```diff
@@ -0,0 +1,35 @@
...
+export default class AzureLoginPage {
```
ntsim, line 5: Given this class name, would suggest we name the file `AzureLoginPage.ts` to reflect this as well

tests/playwright-tests/utils/env.ts
```diff
@@ -1,7 +1,13 @@
...
-  public static BASE_URL = process.env.PUBLIC_URL;
+
+const environment =  {
+  PUBLIC_URL : process.env.PUBLIC_URL ?? "" ,
+  ADMIN_URL : process.env.ADMIN_URL ?? "",
+  ADMIN_EMAILADDR : process.env.ADMIN_EMAIL ?? "",
+  ADMIN_PASSWORD : process.env.ADMIN_PASSWORD ?? "",
 }
+
+// Export the function
```
ntsim, line 12: This comment doesn't really add anything so would suggest we remove it :smile:

tests/playwright-tests/utils/generateUITheme.ts
```diff
@@ -0,0 +1,10 @@
+
+function generateUIThemeName(): string {
+    const date = new Date();
+    const utcFormat = date.toISOString().replace("T", " ").substring(0, 19);
+    const uithemename = 'UI Test'.concat(utcFormat);
+    return uithemename;
+}
+
+// Export the function
+export default generateUIThemeName;
```
ntsim, line 10: We can roll up default function exports into:
```ts
export default function generateUIThemeName(): string
```

tests/playwright-tests/utils/generateUITheme.ts
```diff
@@ -0,0 +1,10 @@
+
+function generateUIThemeName(): string {
+    const date = new Date();
+    const utcFormat = date.toISOString().replace("T", " ").substring(0, 19);
+    const uithemename = 'UI Test'.concat(utcFormat);
+    return uithemename;
+}
+
+// Export the function
```
ntsim, line 9: Similar to other places, no need for this comment

tests/playwright-tests/admin/azpage/AzLoginPage.ts
```diff
@@ -0,0 +1,35 @@
...
+    await this.password.waitFor({state:'visible'})
+    await this.password.fill(environment.ADMIN_PASSWORD);
+    await this.signInButton.click();
+    await this.noButton.waitFor({ state: 'visible'});
+    await this.noButton.click();
+  }
+}
+
+
+
```
ntsim, line 35: We'll need to run Prettier to make sure everything is formatted correctly. You can do this via `pnpm run format`.

tests/playwright-tests/tests/themes.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+
+        await adminPage.clickSignIn();
+        await azPage.doSignIn();
+    });
+
+test("Verify that themes are being created and displayed in the themes home screen", async () => {
+    await adminPage.manageThemesTopicLink.click();
+    await themesPage.createThemeLink.click();
+
+    const uiThemeName: string = generateUIThemeName();
```
duncan-at-hiveit, line 30: Can probably remove the `:string` castings - I expect we'd all know they were strings :slightly_smiling_face:

tests/playwright-tests/utils/generateUITheme.ts
```diff
@@ -0,0 +1,10 @@
+
+function generateUIThemeName(): string {
+    const date = new Date();
+    const utcFormat = date.toISOString().replace("T", " ").substring(0, 19);
+    const uithemename = 'UI Test'.concat(utcFormat);
```
duncan-at-hiveit, line 5: To avoid any warnings about unnecessary variable assignments, might just wanna change the last 2 lines to just be:

`return 'UI Test'.concat(utcFormat);`

Also I don't know about general TS conventions, but where we've done string concatenations elsewhere in the project, I think we've generally favoured using string literals like:

```return `UI Test${utcFormat}`;```

File: tests/playwright-tests/tests/themes.spec.ts [hunk 9]
```diff
@@ -0,0 +1,37 @@
...
+    const uiThemeTile: string = 'Title'.concat(uiThemeName);
+    const uiThemeSummary: string = 'Summary'.concat(uiThemeName);
+
+    await createThemePage.doCreateTheme(uiThemeTile,uiThemeSummary);
```
duncan-at-hiveit, line 34: I like that the test cases themselves are nice and neat and the use of the page object - very nice.

tests/playwright-tests/tests/themes.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+    await themesPage.checkThemeIsDisplayed(uiThemeTile);
```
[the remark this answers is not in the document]
saicharan2789: yeah we can do that - since the ticket is to pick up few test from 'Themes and topics.robot' I have chosen this way. Would you want me to implement the edit themes and topics as part of this ticket? I'm happy to do. FYI

tests/playwright-tests/tests/themes.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+        await adminPage.clickSignIn();
+        await azPage.doSignIn();
+    });
+
+test("Verify that themes are being created and displayed in the themes home screen", async () => {
+    
+    await adminPage.manageThemesTopicLink.click();
+    await themesPage.createThemeLink.click();
+
+    const uiThemeName = generateUIThemeName();
```
duncan-at-hiveit, line 31: I'd be tempted to call `generateUIThemeName` `generateRunIdentifier` instead, as it's not actually generating the theme name.  That's done below, on line 32.  (also I think you have a typo there - uiThemeTile should be uiThemeTitle).

In the existing UI tests, we use that unique "run identifier" in other places as well so it's not just tied to being used for the theme name.  For that reason as well, I'd suggest the renaming.

If you were going to rename this method, I'd also rename the file that it's in as well :slightly_smiling_face:

tests/playwright-tests/utils/env.ts
```diff
@@ -1,7 +1,12 @@
...
 dotenv.config();
-// eslint-disable-next-line import/prefer-default-export
-export class environment {
-  public static BASE_URL = process.env.PUBLIC_URL;
-}
+
+const environment = {
+  PUBLIC_URL: process.env.PUBLIC_URL ?? '',
+  ADMIN_URL: process.env.ADMIN_URL ?? '',
+  ADMIN_EMAILADDR: process.env.ADMIN_EMAIL ?? '',
```
ntsim, line 8: Wouldn't `ADMIN_EMAIL` be sufficient? No need for the `ADDR`

tests/playwright-tests/admin/azpage/AzureLoginPage.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export default class AzureLoginPage {
+  readonly page: Page;
+  readonly emailAddress: Locator;
+  readonly nextButton: Locator;
+  readonly password: Locator;
+  readonly signInButton: Locator;
+  readonly noButton: Locator;
+
+  constructor(page) {
```
ntsim, line 12: Think to avoid any warnings from the compiler or IDE, it'd be good to have an explicit type here i.e.

```ts
constructor(page: Page)
```

Would suggest doing this for the other page classes as well

tests/playwright-tests/tests/themes.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+    await adminPage.clickSignIn();
+    await azPage.doSignIn();
+  });
+
+  test('Verify that themes are being created and displayed in the themes home screen', async () => {
+    await adminPage.manageThemesTopicLink.click();
+    await themesPage.createThemeLink.click();
+
+    const uiThemeName = generateRunIdentifier();
+    const uiThemeTitle = `Theme${uiThemeName}`;
```
ntsim, line 31: Wouldn't this generate something like the following? `ThemeUI Test2024-01-17 12:00:00`

This doesn't look right for a few reasons:

- It looks like inconsistent from current conventions for naming high level things like themes in the UI tests
- There isn't sufficient spacing between the parts due to how the interpolations are arranged in the string

Would suggest we do something like the following instead: `UI Test - Test theme - 2024-01-17 12:00:00`

This is neater and follows our naming convention better.

Check out my later comment in `generateRunTimeIdentifier.ts` on how we could abstract this.

tests/playwright-tests/tests/themes.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+    const uiThemeSummary = `Summary${uiThemeName}`;
```
ntsim, line 32: Similar thing to above, this could be `UI Test - Test summary - 2024-01-17 12:00:00`

tests/playwright-tests/utils/generateRunIdentifier.ts
```diff
@@ -0,0 +1,5 @@
+export default function generateRunIdentifier() {
+  const date = new Date();
+  const utcFormat = date.toISOString().replace('T', ' ').substring(0, 19);
+  return `UI Test${utcFormat}`;
+}
```
ntsim, line 5: The generation of this run identifier will be dependent on when the function is called. However, this isn't a good idea as this will lead to the generated run identifier being different at different times in the tests.

This means it'll be difficult to identify what things belong to the same test run (which is the point of the run identifier).

In the current Robot tests, we set up a `RUN_IDENTIFIER` environment variable once when the tests start. We don't generate this again later.

Would suggest we do something similar i.e.

```ts
// env.ts

const environment = {
  // ...
  RUN_IDENTIFIER: generateRunIdentifier(),
};
```

Then we can use this in our test cases like:

```ts
const theme = `UI Test - Test theme - ${environment.RUN_IDENTIFIER}`
```

We don't have to do this next part, but perhaps if we want to abstract this a little so that it's easier to use (and more consistent), we could create another util:

```ts
// uiTestString.ts

import environment from './environment';

export default function uiTestString(str: string): string {
  return `UI Test - ${str} - ${environment.RUN_IDENTIFIER}`;
}
```

This could then be used like:

```ts
const theme = uiTestString('Test theme'); // Creates 'UI Test - Test Theme - 2024-01-17 12:00:00'
```

tests/playwright-tests/utils/generateRunIdentifier.ts
```diff
@@ -0,0 +1,5 @@
+export default function generateRunIdentifier() {
+  const date = new Date();
+  const utcFormat = date.toISOString().replace('T', ' ').substring(0, 19);
+  return `UI Test${utcFormat}`;
```
ntsim, line 4: For the run identifier, we normally append an additional random string that's 6 characters in length. 

This provides some uniqueness to prevent collisions with other test runs that are running in parallel. This avoids multiple tests run all having the same run identifier.

Would suggest we do something similar here.

The best way to do this would be to have some util like:

```ts
export default function generateRandom(length: number): string {
  let random = '';

  // Avoid empty string being created if `Math.random` creates 0
  while (!current) {
    random = Math.random().toString(36).slice(2, 2 + length);
  }

  return random;
}
```

This would then be appended to the run identifier string i.e.

```ts
export default function generateRunIdentifier(): string {
  // ...
  return `${utcFormat} ${generateRandom(6)}`
}
```

ntsim (approved):

PR #4506, opened 2024-01-08
EES-4553: Add exception handling and Slack notifications to snapshot creation script

tomjonesdev:
**Motivation**
When the create_snapshots.py python file was being run, it occasionally failed to complete, however there was no process in place to pick up these failures and notify relevant parties.

**Changes**
Wrap script functionality in an exception handling block in order to catch any errors raised during execution, and send notifications to the appropriate Slack channel.

The exception notification is sent using the webhook parameter, which is an _optional_ parameter in the local development environment, however this parameter is always present during execution of the pipeline.

**Note for reviewers**
None of the code within the try block should have changed (it has just been indented)

### DIFF
--- a/tests/robot-tests/scripts/create_snapshots.py
+++ b/tests/robot-tests/scripts/create_snapshots.py
@@ -266,98 +266,133 @@ class SnapshotService:
 
 
 if __name__ == "__main__":
-    from get_webdriver import get_webdriver
+    def send_alert(message: str) -> None:
+        if args.slack_webhook_url == None or args.slack_webhook_url == "":
+            return
 
-    ap = argparse.ArgumentParser(
-        prog=f"pipenv run python {os.path.basename(__file__)}",
-        description="To create snapshots of specific public frontend pages",
-    )
+        snapshot_service = SnapshotService(public_url=args.public_url, slack_webhook_url=args.slack_webhook_url)
+        snapshot_service._send_slack_notification(text="fallback", blocks=[
+            {
+                "type": "header",
+                "text": {
+                    "type": "plain_text",
+                    "text": ":x: Snapshot process failed",
+                }
+            },
+            {
+                "type": "section",
+                "text": {
+                    "type": "plain_text",
+                    "text": f"{message}",
+                }
+            },
+        ])
 
-    ap.add_argument(
-        "--public-url",
-        dest="public_url",
-        default="https://explore-education-statistics.service.gov.uk",
-        nargs="?",
-        help="URL of public frontend you wish to create snapshots for",
-        type=str,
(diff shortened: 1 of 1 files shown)


tests/robot-tests/scripts/create_snapshots.py
```diff
@@ -266,98 +266,133 @@ def write_snapshots_to_file(self) -> None:
...
-    if args.ci:
-        snapshot_service.validate_snapshots()
-        snapshot_service.write_snapshots_to_file()
-
-    if args.validate:
-        snapshot_service.validate_snapshots()
-
-    driver.quit()
+    def send_alert(message: str) -> None:
+        if args.slack_webhook_url == None or args.slack_webhook_url == "":
```
mmyoungman, line 270: Doesn't really matter, but you could do `if not args.slack_webhook_url:` here, as both `None` and `""` evaluate to `False`.

tests/robot-tests/scripts/create_snapshots.py
```diff
@@ -266,98 +266,133 @@ def write_snapshots_to_file(self) -> None:
...
+        driver.quit()
+    except Exception as ex:
+        message = ""
+        
+        if hasattr(ex, 'message'):
+            message = ex.message
+        elif hasattr(ex, 'msg'):
+            message = ex.msg
+        else:
+            message = ex
```
mmyoungman, line 395: Wondering where you got this from? I can see people complaining about the message either being `ex.message` or just `ex`, but couldn't find a mention about `ex.msg`.

mmyoungman (approved):

tests/robot-tests/scripts/create_snapshots.py
tomjonesdev: Ah ok, similar "falsey" behaviour to JS then. I'll leave as-is on this occasion.

tests/robot-tests/scripts/create_snapshots.py
tomjonesdev: From what I gathered, `message` was the most common property of `Exception`-type classes, and those which derive from them. Unfortunately there doesn't seem to be anything definitive as there's no interface/implementation contract like something a .NET class might have (e.g. if it derives from `Exception`, it _must_ have a `message` property).

`msg` is the equivalent property for Selenium's `NoSuchElementException` - again, there's nothing forcing them to use `message`. It's essentially a free-for-all what the error summary property is called, which is why I added a catch-all `ex` for exceptions which don't have either of those two properties. The problem here is that the full `ex` object is likely to include stack trace information, but other conditions can easily be added to this block in the future if necessary.

PR #4507, opened 2024-01-08
EES-4705 removing content api and content services dependencies

jack-hive:
These changes are to allow the new Public.Data.Api project to not have the Content.Api and Content.Services projects as dependencies.

When we add the Publications endpoints to the Public.Data.Api project, it will need to reference the Request objects of the Content.Api project and the View Model objects of the Content.Services project.

Moving these into their own assembly allows us to stop directly referencing those existing projects.

**Note:**

For each of the request objects and view models, we also removed redundant nullable directives + removed unused using statements + converted all block-scoped namespaces into file-scoped namespaces. The latter change is why it appears like there are lots of adds and deletes rather than changed files


src/GovUk.Education.ExploreEducationStatistics.Content.Requests/GovUk.Education.ExploreEducationStatistics.Content.Requests.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
```
ntsim, line 6: This enables nullable for the entire subproject, so we can just get rid of `#nullable enable` from the request.

src/GovUk.Education.ExploreEducationStatistics.Content.Requests/GovUk.Education.ExploreEducationStatistics.Content.Requests.csproj
```diff
@@ -0,0 +1,14 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\GovUk.Education.ExploreEducationStatistics.Content.Model\GovUk.Education.ExploreEducationStatistics.Content.Model.csproj" />
+    <ProjectReference Include="..\GovUk.Education.ExploreEducationStatistics.Content.Services\GovUk.Education.ExploreEducationStatistics.Content.Services.csproj" />
```
ntsim, line 11: Shouldn't need a dependency on `Content.Services`. We can remove this by moving the `PublicationsSortBy` enum into this subproject.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/MethodologyServiceTests.cs
```diff
@@ -10,14 +10,14 @@
...
 using GovUk.Education.ExploreEducationStatistics.Content.Services.Tests.Mappings;
-using GovUk.Education.ExploreEducationStatistics.Content.Services.ViewModels;
 using Moq;
 using Xunit;
 using static GovUk.Education.ExploreEducationStatistics.Common.Services.CollectionUtils;
 using static GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils.MockUtils;
 using static GovUk.Education.ExploreEducationStatistics.Content.Model.MethodologyPublishingStrategy;
 using static GovUk.Education.ExploreEducationStatistics.Content.Model.MethodologyApprovalStatus;
 using static GovUk.Education.ExploreEducationStatistics.Content.Model.Tests.Utils.ContentDbUtils;
+using GovUk.Education.ExploreEducationStatistics.Content.ViewModels;
```
ntsim, line 20: As other comments - import ordering could be retained from before

src/GovUk.Education.ExploreEducationStatistics.Content.ViewModels/GovUk.Education.ExploreEducationStatistics.Content.ViewModels.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
```
ntsim, line 6: We don't have to, but these settings mean we could remove any unnecessary `#nullable enable` and unneeded `System` imports from all the view models (which is basically all of them).

It's a little bit extra, but might be nice to do given we're here.

If you're not inclined to do it as part of this work, then would suggest we still keep these settings enabled anyway as they're sensible defaults for new subprojects.

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Requests/GovUk.Education.ExploreEducationStatistics.Content.Requests.csproj
jack-hive: Good spot

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/MethodologyServiceTests.cs
jack-hive: Sorted :)

src/GovUk.Education.ExploreEducationStatistics.Content.ViewModels/GovUk.Education.ExploreEducationStatistics.Content.ViewModels.csproj
jack-hive: Good spot :) Sorted it as part of this work

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Methodologies/MethodologyApprovalServiceTests.cs
```diff
@@ -28,6 +27,7 @@
 using static GovUk.Education.ExploreEducationStatistics.Content.Model.MethodologyPublishingStrategy;
 using static GovUk.Education.ExploreEducationStatistics.Content.Model.MethodologyApprovalStatus;
 using static Moq.MockBehavior;
+using GovUk.Education.ExploreEducationStatistics.Content.ViewModels;
```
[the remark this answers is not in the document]
jack-hive: Sorted :)

src/GovUk.Education.ExploreEducationStatistics.Content.Requests/GovUk.Education.ExploreEducationStatistics.Content.Requests.csproj
jack-hive: Good spot, I've switched the dependency around

src/GovUk.Education.ExploreEducationStatistics.Content.ViewModels/GovUk.Education.ExploreEducationStatistics.Content.ViewModels.csproj
```diff
@@ -0,0 +1,13 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\GovUk.Education.ExploreEducationStatistics.Data.Services\GovUk.Education.ExploreEducationStatistics.Data.Services.csproj" />
```
ntsim, line 10: Shouldn't need this dependency. Think this can be replaced with a reference to `Content.Model` instead

jack-hive: Refactored so that this no longer depends on `Data.Services`. I've pulled the view models out of `Data.Services` into their own project `Data.ViewModels`

PR #4509, opened 2024-01-10
EES-4683 Add public API infrastructure

Dancemammal:
Explain the **motivation** for making this change. What existing problem does the pull request solve?

This pull request creates the IaC Azure Architecture for the public API.
The IaC is built out of reusable components that can be utilised in other projects

## Related changes
No related Changes

## Screenshots
All documentation is included as part of the Technical Architecture Document


infrastructure/templates/public-api/application/etlFunctionApp.bicep
```diff
@@ -0,0 +1,140 @@
...
+@description('Specifies the location for all resources.')
+param location string
+
+//Specific parameters for the resources
+@description('Application : Insights name')
+param applicationInsightsName string
+
+@description('Function App Plan operating system')
+@allowed([
+  'Windows'
```
jack-hive, line 13: Why don't we just choose Windows?

infrastructure/templates/public-api/application/etlFunctionApp.bicep
```diff
@@ -0,0 +1,140 @@
...
+  'Linux'
+])
+param appServiceplanOS string = 'Linux'
+
+@description('Function App : Runtime Language')
+@allowed([
+  'dotnet'
```
jack-hive, line 20: I believe we just need `dotnet`. I think it's a DFE requirement? @ntsim can confirm

infrastructure/templates/public-api/main.bicep
```diff
@@ -0,0 +1,372 @@
...
+    storageAccountName: storageAccountModule.outputs.storageAccountName
+    //tags
+  }
+  dependsOn: [
+    storageAccountModule
+  ]
+}
+
+//Deploy Function blob store
+module blobStoreModule 'components/fileShares.bicep' = {
```
jack-hive, line 190: should this be `components/blobStore.bicep`?

jack-hive (changes requested):

infrastructure/templates/public-api/main.bicep
```diff
@@ -0,0 +1,372 @@
...
+
+
+//outputs
+output crLoginServer string = containerRegistryModule.outputs.crLoginServer
+output crName string = containerRegistryModule.name
+output metadataDatabaseRef string = databaseModule.outputs.metadataDatabaseRef
+
+
+
+
```
jack-hive, line 372: empty lines

infrastructure/templates/public-api/main.bicep
```diff
@@ -0,0 +1,372 @@
...
+  params: {
+    storageAccountName: storageAccountModule.outputs.storageAccountName
+    //tags
+  }
+  dependsOn: [
+    storageAccountModule
+  ]
+}
+
+
```
jack-hive, line 201: Sometimes we have 1 line between blocks of code, sometimes 2 and sometimes more. Might me nice to make them consistent?

infrastructure/templates/public-api/main.bicepparam
```diff
@@ -0,0 +1,116 @@
+using './main.bicep'
```
jack-hive, line 1: the stuff in this file **ALMOST** (but not quite) looks like a duplicate of the stuff at the top of `main.bicep`. Do we need both?

jack-hive (changes requested):

infrastructure/templates/public-api/application/etlFunctionApp.bicep
Dancemammal: Linux is smaller, faster and cheaper than running a windows vm

infrastructure/templates/public-api/application/etlFunctionApp.bicep
Dancemammal: Agreed, but options left in for training purposes

infrastructure/templates/public-api/main.bicep
Dancemammal: Yes correct, it should be

infrastructure/templates/public-api/main.bicep
Dancemammal: Done

infrastructure/templates/public-api/main.bicep
Dancemammal: Agree

infrastructure/templates/public-api/application/etlFunctionApp.bicep
jack-hive: Have we chosen which one we want to use for the public API project? Just wondered why we'd want 2 options rather than choosing just 1?

infrastructure/templates/public-api/main.bicepparam
jack-hive: I'm not entirely sure what this file is used for so this is quite possibly just a lack of knowledge on my part (apologies if so!)

Dancemammal: Its a useful tool to have, I have kept most items as default, but I wanted to include the ability to make changes to the parameter file whilst leaving the main.bicep untouched. It is also best practice (MS)

jack-hive: I think this was simply me not understanding how these files were used.

Is there a reason it doesn't match the ones in `main.bicep` exactly though?

infrastructure/templates/public-api/main.bicep
```diff
@@ -0,0 +1,362 @@
...
+//ServiceBus Queue Params -------------------------------------------------------------------
+@description('Name of the Service Bus namespace')
+param namespaceName string = 'etlnamespace'
+
+@description('Name of the Queue')
+param queueName string = 'etlfunctionqueue'
+
+//ETL Function Paramenters ------------------------------------------------------------------
+@description('Application : Insights name')
+param applicationInsightsName string = 'etlFunctionInsights'
```
ntsim, line 103: Don't think we need this. As per another comment, names for Application Insights should just be derived from the app that it's attached to

infrastructure/templates/public-api/main.bicep
```diff
@@ -0,0 +1,362 @@
...
+module blobStoreModule 'components/blobStore.bicep' = {
+  name: 'blobStoreDeploy'
+  params: {
+    storageAccountName: storageAccountModule.outputs.storageAccountName
+    //tags
+  }
+  dependsOn: [
+    storageAccountModule
+  ]
+}
```
ntsim, line 198: As mentioned earlier, doesn't look like this gets used?

infrastructure/templates/public-api/main.bicep
```diff
@@ -0,0 +1,362 @@
...
+
+//Deploy Key Vault
+module keyVaultModule 'components/keyvault.bicep' = {
```
ntsim, line 201: The casing for this filename string is wrong - should be `keyVault.bicep`. 

We should also correct this for the actual file itself.

infrastructure/templates/public-api/main.bicep
```diff
@@ -0,0 +1,362 @@
...
+    serviceOwnerName: serviceOwnerName
+    dateProvisioned: dateProvisioned
+    createdBy: createdBy
+    deploymentRepo: deploymentRepo
+    deploymentScript: 'network.bicep'
+  }
+}
+
+//Deploy Storage Account
+module storageAccountModule 'components/storageAccount.bicep' = {
```
ntsim, line 148: Whilst the casing of the filename in this string, it's not actually right in the project files. It's currently `storageaccount.bicep`, so we should correct this

infrastructure/templates/public-api/main.bicepparam
```diff
@@ -0,0 +1,116 @@
...
+
+@description('Function App : Runtime Language')
+@allowed([
+  'dotnet'
+  'node'
+  'python'
+  'java'
+])
+param functionAppRuntime = 'python'
+
```
ntsim, line 116: Would get rid of all of these descriptions. They're defined in `main.bicep` so there's no need to redeclare them. Otherwise, we're going to quickly see discrepancies between the two as people change things.

infrastructure/templates/public-api/main.bicepparam
ntsim: Jack's correct, this doesn't look right at all.

I think this should only really contain parameters that are overriding parameters in  `main.bicep`. Parameters that are using the default shouldn't need to be re-declared here!

We've done this with the current infrastructure where we have different `.parameters.json` files for each respective environment. We then deploy the infrastructure for each environment by specifying the parameter file to use.

We should do a similar thing here, but using `bicepparam` files instead. We can do a similar thing to the current infrastructure and create a new `parameters` directory under `public-api` that contains these files.

infrastructure/templates/public-api/components/network.bicep
```diff
@@ -0,0 +1,165 @@
...
+param adminSubnetPrefix string = '10.0.0.0/24'
+
+@description('Importer Subnet Address Prefix')
+param ImporterSubnetPrefix string = '10.0.1.0/24'
+
+@description('Publisher Subnet Address Prefix')
+param publisherSubnetPrefix string = '10.0.2.0/24'
+
+@description('Content Subnet Address Prefix')
+param contentSubnetPrefix string = '10.0.3.0/24'
```
ntsim, line 22: As mentioned later, we shouldn't need any of these parameters if we just use existing resource references.

infrastructure/templates/public-api/components/postgresqlDatabase.bicep
```diff
@@ -0,0 +1,153 @@
...
+module StoreADOConnectionStringToKeyVault './keyVaultSecret.bicep' = {
+  name: 'ado-kv-connectionString'
+  params: {
+    KeyVaultName: KeyVaultName
+    IsEnabled: true
+    SecretValue: adoNetDbConnectionString 
+    ContentType: 'text/plain'
+    SecretName: adoNetSecretName
+  }
+}
```
ntsim, line 139: As discussed, we probably shouldn't override the secret if one hasn't already been deployed to Key Vault.

To do the initial secret creation, we would need to pass in an generated random string for the admin password as a pipeline parameter. There'd be two ways of doing this:

1. Generating it a new one on each deployment
2. Just set a generated password in manually

With option 1, we'd need to be careful that this doesn't try and change the database's admin password every time there's a deployment.

With option 2, we'd need to remove the password once it's been made available to Key Vault.

Thinking about it some more though, I think we might want to consider some alternatives...

## Deploying Key Vaults with 'seed' secrets

This would involve creating a separate Bicep deployment that only deploys the Key Vaults and any 'seed' secrets, such as the admin password.

These Key Vaults could then be referenced in this deployment to extract the admin password and use it for the database and generating the connection string secret.

## Using Managed Identities

I think this is arguably the best option. Passwordless authentication using Managed Identities is pretty ideal and would mean we don't need to go around the houses with this.

I'm sure we'll probably run into permission issues in the DfE Azure, but I think this is probably the way to go.

Given that we already have to make a managed identity for any Container App, this doesn't feel like too much more on top!

With this approach, we'll also need to configure the Processor (ETL) with an appropriate managed identity.

infrastructure/templates/public-api/application/etlFunctionApp.bicep
```diff
@@ -0,0 +1,140 @@
...
+var appServicePlanName = '${subscription}-apsp-${buildNumber}'
+
+
+//---------------------------------------------------------------------------------------------------------------
+// All resources via modules
+//---------------------------------------------------------------------------------------------------------------
+
+//Application Insights Deployment
+module applicationInsightsModule '../components/appInsights.bicep' = {
+  name: 'appInsightsDeploy-${buildNumber}'
```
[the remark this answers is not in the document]
Dancemammal: Changed to use more human friendly names

infrastructure/templates/public-api/application/etlFunctionApp.bicep
```diff
@@ -0,0 +1,140 @@
...
+    databaseConnectionString: databaseConnectionStringURI
+    functionAppName: functionAppModule.outputs.functionAppName
+    functionAppRuntime: functionAppRuntime
+    storageAccountConnectionString: storageAccountConnectionString
+    serviceBusConnectionString: serviceBusConnectionString
+  }
+  dependsOn: [
+    functionAppModule
+  ]
+}
```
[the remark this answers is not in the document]
Dancemammal: done

infrastructure/templates/public-api/components/containerApp.bicep
```diff
@@ -0,0 +1,286 @@
...
+param dateProvisioned string = utcNow('u')
+param createdBy string = 'Unknown'
+param deploymentRepo string = 'N/A'
+param deploymentScript string = 'N/A'
+
+
+//Variables 
+var ContainerEnvName = '${subscription}-env-${containerAppEnvName}'
+var containerImageName = '${acrLoginServer}/${acrHostedImageName}'
+var ContainerAppName = '${subscription}-app-${containerAppName}'
```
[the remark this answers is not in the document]
Dancemammal: resource prefix will not match how the names are created in the template.yaml, it may be easier to follow that naming convention

infrastructure/templates/public-api/components/containerApp.bicep
```diff
@@ -0,0 +1,286 @@
...
+var UserIdentityName = '${subscription}-id-${containerAppName}'
+var ContainerLogName = '${subscription}-log-${containerAppLogAnalyticsName}'
+var ContainerImportName = '${subscription}importContainerImage'
+
+//var revisionSuffix = uniqueString(containerImage, containerAppName)
+//var sanitizedRevisionSuffix = substring(revisionSuffix, 0, 10)
+
+var acrPullRole = resourceId('Microsoft.Authorization/roleDefinitions', '7f951dda-4ed3-4680-a7ca-43fe172d538d')
```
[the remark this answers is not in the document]
Dancemammal: Yes, its the ID for ACRPull

infrastructure/templates/public-api/components/containerApp.bicep
```diff
@@ -0,0 +1,286 @@
...
+    DateProvisioned: dateProvisioned
+    CreatedBy: createdBy
+    DeploymentRepo: deploymentRepo
+    DeploymentScript: deploymentScript
+  }
+}
+
+//Registry Seeder
+@description('This module seeds the ACR with the public version of the app')
+module acrImportImage 'br/public:deployment-scripts/import-acr:3.0.1' = if (seedRegistry) {
```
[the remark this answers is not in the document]
Dancemammal: We cannot build an AppContainer without an image in the ACR, this is a work around, and is recommended by many

infrastructure/templates/public-api/components/network.bicep
```diff
@@ -0,0 +1,165 @@
...
+// Variables and created data
+var vNetName = '${subscription}-vnet-${environment}'
+var adminSubnetName = '${subscription}-snet-${environment}-admin'
+var importerSubnetName = '${subscription}-snet-${environment}-importer'
+var publisherSubnetName = '${subscription}-snet-${environment}-publisher'
+var contentSubnetName = '${subscription}-snet-${environment}-content'
+var databaseSubnetName = '${subscription}-snet-${environment}-database'
+
+//Resources 
+resource virtualnetwork 'Microsoft.Network/virtualNetworks@2021-05-01' = if (deploySubnets) {
```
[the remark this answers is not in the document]
Dancemammal: This will create a mirror network for testing, when we move we can use the flag deploySubnets to use the existing network

infrastructure/templates/public-api/components/postgresqlDatabase.bicep
```diff
@@ -0,0 +1,153 @@
...
+param subscriptionName string = 'Unknown'
+param costCentre string = 'Unknown'
+param serviceOwnerName string = 'Unknown'
+param dateProvisioned string = utcNow('u')
+param createdBy string = 'Unknown'
+param deploymentRepo string = 'N/A'
+param deploymentScript string = 'N/A'
+
+// Variables and created data
+var metadataDatabaseName = '${subscription}-postgre-${serverName}'
```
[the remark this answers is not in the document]
Dancemammal: Unfortunately that will not match with how it is written in the template
${subscription}-psql-${environment}-${serverName}

infrastructure/templates/public-api/components/postgresqlDatabase.bicep
```diff
@@ -0,0 +1,153 @@
...
+var pythonDbConnectionString = 'database=<database>, user=${administratorLoginName}@${metadataDatabase.name}, host=${metadataDatabase.name}${environment().suffixes.sqlServerHostname}, password=${administratorLoginPassword}, port=5432'
+var adoNetDbConnectionString = 'Server=${metadataDatabase.name}${environment().suffixes.sqlServerHostname};Database=<database>;Port=5432;User Id=${administratorLoginName}@${metadataDatabase.name};Password=${administratorLoginPassword};'
```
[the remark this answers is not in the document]
Dancemammal: In Bicep, strings are marked with singled quotes, and must be declared on a single line.
Using concat throws an error with the linter.

Use string interpolation instead of the concat function.
The following example fails this test because the concat function is used.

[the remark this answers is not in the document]
ntsim: You could disable the linter in this situation: https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/linter#silencing-false-positives

infrastructure/templates/public-api/components/storageaccount.bicep
```diff
@@ -0,0 +1,103 @@
+@description('Specifies the Subscription to be used.')
+param subscription string
+
+@description('Specifies the location for all resources.')
+param location string
+
+//Specific parameters for the resources
+@description('Storage : Storage Account Name')
+param storageAccountName string = 'saeespublicapi'
```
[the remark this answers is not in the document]
Dancemammal: Linter issues with 2 char, moved to -sacc- to remove linting errors

infrastructure/templates/public-api/main.bicep
Dancemammal: done

infrastructure/templates/public-api/components/containerRegistry.bicep
```diff
@@ -0,0 +1,75 @@
...
+    Environment: environmentName
+    Subscription: subscriptionName
+    CostCentre: costCentre
+    ServiceOwner: serviceOwnerName
+    DateProvisioned: dateProvisioned
+    CreatedBy: createdBy
+    DeploymentRepo: deploymentRepo
+    DeploymentScript: deploymentScript
+  }
+}
```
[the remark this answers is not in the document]
Dancemammal: Made it so you can create or use an existing registry

PR #4522, opened 2024-01-16
EES-4760 Remove associated Azure Storage ReleaseStatus rows when remo…

mmyoungman:
…ving topic

This PR ensures that entries to the Azure Storage Publisher `ReleaseStatus` table are removed as part of removing a theme or topic. This was necessary because we were seeing alerts like this against the dev environment:
image

After investigating the reason for this alert, it was found that if a release is scheduled for publication, but then removed as part of the UI tests test data teardown, then the `ReleaseStatus` entry remained. Because we had this entry for a scheduled release that no longer exists in the Content or Statistics DBs, when it was staged as part of the midnight publishing step, it caused an alert.

As we now remove these entries from `ReleaseStatus`, this should no longer happen.

# Additional notes

- With `Microsoft.Azure.Cosmos.Table`, it isn't possible to batch delete table entities, and it also is necessary to get entities before removing them. Hence why I fetch all relevant entries first, and then remove them one by one. This can be done more efficiently when we migrate to Azure.Data.Tables as part of EES-4771: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/tables/Azure.Data.Tables/MigrationGuide.md#delete-table-entities

- This new code gets called in `TopicService#DeleteTopic` as part of the UI tests test data teardown, and `ThemeService#DeleteTheme` which is used by the UI test suite `themes_and_topics.robot`

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleasePublishingStatusRepository.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleasePublishingStatusRepository.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Publisher.Model;
 using Microsoft.Azure.Cosmos.Table;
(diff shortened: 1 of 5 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleasePublishingStatusRepository.cs
```diff
@@ -47,5 +48,38 @@ public Task<IEnumerable<ReleasePublishingStatus>> GetAllByOverallStage(Guid rele
             var query = new TableQuery<ReleasePublishingStatus>().Where(filter);
             return _publisherTableStorageService.ExecuteQueryAsync(PublisherReleaseStatusTableName, query);
         }
+
+        public async Task RemovePublisherReleaseStatuses(List<Guid> releaseIds)
+        {
+            if (releaseIds.IsNullOrEmpty())
+            {
+                // If filter is string.Empty, the query returns all table entities
+                return;
```
tomjonesdev, line 57: If an empty filter is considered an error, should we feed something back to the caller so that it can be handled accordingly, such as an `ArgumentNullException(nameof(releaseIds))`, or a failed/cancelled task?

tomjonesdev (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleasePublishingStatusRepository.cs
mmyoungman: I think it is correct as it is. It's possible for `DeleteReleasesForPublications` to be called on publications that have no releases. In those cases, we'd expect `RemovePublisherReleaseStatuses` to be called with an empty array and to handle that correctly i.e. do nothing.

I've updated the comment to (hopefully) make it clearer the reason why this initial if statement is necessary.

PR #4525, opened 2024-01-19
EES-4522 Use RIGHT JOIN in Observation SQL query

mmyoungman:
This PR implements a performance optimisation found by Paul. See Jira ticket for more details.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Services/ObservationService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Services/ObservationService.cs
@@ -117,7 +117,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Services
                     INSERT INTO #{nameof(MatchedObservation)} 
                     SELECT o.id FROM Observation o " +
                     (locationIdsTempTableName != null
-                        ? $"RIGHT JOIN {locationIdsTempTableName} ON {locationIdsTempTableName}.Id = o.LocationId "
+                        ? $"JOIN {locationIdsTempTableName} ON {locationIdsTempTableName}.Id = o.LocationId "
                         : "") +
                     "WHERE o.SubjectId = @subjectId " +
                     (timePeriodQuery != null ? $"AND ({GetTimePeriodsClause(timePeriodQuery)}) " : "") +
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Services/ObservationService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Services/ObservationService.cs
@@ -105,9 +105,9 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Services
             {
                 await TempTableCreator.CreateTemporaryTable<MatchedObservation>(context, cancellationToken);
 
-                var (locationIdsClause, locationIdsTempTable) = !locationIds.IsNullOrEmpty()
-                    ? await GetLocationsClause(context, locationIds!, cancellationToken)
-                    : default;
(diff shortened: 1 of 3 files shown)


src/GovUk.Education.ExploreEducationStatistics.Data.Services/ObservationService.cs
```diff
@@ -105,36 +105,38 @@ public class MatchingObservationsQueryGenerator : IMatchingObservationsQueryGene
...
                     ? await GetSelectedFilterItemIdsClause(context, subjectId, filterItemIds!, cancellationToken)
                     : default;
 
                 var sql = @$"
                     INSERT INTO #{nameof(MatchedObservation)} 
-                    SELECT o.id FROM Observation o
-                    WHERE o.SubjectId = @subjectId " +
+                    SELECT o.id FROM Observation o " +
+                    (locationIdsTempTableName != null
+                        ? $"RIGHT JOIN {locationIdsTempTableName} ON {locationIdsTempTableName}.Id = o.LocationId "
```
duncan-at-hiveit, line 120: This all looks good.  I'm curious however as to why this RIGHT JOIN would behave different from a default inner join - might be good to test the 2 out side by side in a local db to see how each performs?

I think that they should behave the same if we're guaranteed that every Location Id in the #LocationTempTable will have at least one match in the Observation table, but if it doesn't (let's say that we have one Location Id of `11111111-5dc5-4cf0-fa53-08dc12e7fedd` that has no Observation matches, we'll end up with null Observation Ids coming out of this query because the RIGHT join is returning rows from #LocationTempTable that have no match in the Observation table too.

duncan-at-hiveit: I guess the 2 questions here are:
1. Was there a particular reason why a RIGHT join outperformed an INNER join?
2. Are we happy that all of the Location Ids supplied at this point are guaranteed to have at least one match in the Observation table?  If not, we'll end up with null Observation Ids in the #MatchedObservation table.

duncan-at-hiveit: Note that to test this I did the following in my local Stats DB:

```sql
Create TABLE #MatchedObservation (Id UNIQUEIDENTIFIER PRIMARY KEY NOT NULL);

CREATE TABLE #LocationTempTable (Id UNIQUEIDENTIFIER PRIMARY KEY NOT NULL);

INSERT INTO #LocationTempTable
VALUES 
    ('d8587137-e930-4ee0-f97b-08dc12e7fedd'), -- a real Location Id referenced from the Observation table
    ('e54d308b-3543-4299-f9e0-08dc12e7fedd'), -- a real Location Id referenced from the Observation table
    ('8942215f-5dc5-4cf0-fa53-08dc12e7fedd'), -- a real Location Id referenced from the Observation table
    ('11111111-5dc5-4cf0-fa53-08dc12e7fedd') -- a fake Location Id!
;

SELECT DISTINCT (o.Id) FROM Observation o
JOIN #LocationTempTable ON #LocationTempTable.Id = o.LocationId; -- returns 82 rows

SELECT DISTINCT (o.Id) FROM Observation o
RIGHT JOIN #LocationTempTable ON #LocationTempTable.Id = o.LocationId;  -- returns 83 rows
```

mmyoungman: Changed it to just use `JOIN`.

I will test it on dev - it's an easy change to revert and there is already a lot of test data on dev.

duncan-at-hiveit (approved):

PR #4530, opened 2024-01-22
EES-4798 Customise validation problem responses from APIs

ntsim:
This PR updates the validation problem responses across the various project APIs to include a custom `errors` property. This contains the list of validation errors, and provides more detail about each in the format of:

```json
{
  "path": "theFieldPath",
  "code": "TheErrorCode",
  "message": "A default error message",
  "details": {}
}
```

Where `details` is an optional free-form object that can provide additional context about the validation error.

With our standard validation using `ValidationResult` or `ValidationActionResult` utils (throughout the service layer), we should typically expect really basic errors like:

```cs
{
  "code": "DataFileCannotBeEmpty",
  "message": ""
}
```

In the future, it'd be nice to fill these errors out more. For now, these essentially work the same as current and will be mapped into any form fields if mappings have been configured.

Importantly, this work now enables richer error responses using FluentValidation, which provides richer error details out of the box. We're hoping that this will work well with validating public API endpoints, particularly for querying.

Unfortunately, it's unlikely we'll be able to achieve a similar thing using attribute-based validation as these only provide a `path` and `message`. We should ideally switch to FluentValidation entirely, although this is unlikely to happen any time soon.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Config/FluentValidationActionFilterTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Config/FluentValidationActionFilterTests.cs
@@ -1,0 +1,471 @@
+#nullable enable
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Net.Http.Json;
+using System.Text.Json;
+using System.Threading.Tasks;
+using FluentValidation;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Common.Tests.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Common.Tests.Fixtures;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Testing;
+using Microsoft.AspNetCore.WebUtilities;
+using Xunit;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Tests.Config;
+
+public class FluentValidationActionFilterTests : IntegrationTest<TestStartup>
+{
+    private const string NotEmptyCode = "NotEmpty";
+    private const string NotEmptyMessage = "Must not be empty.";
+
+    private FluentValidationActionFilterTests(TestApplicationFactory<TestStartup> testApp) : base(testApp)
+    {
+    }
+
+    public class TestClassTests : FluentValidationActionFilterTests
+    {
+        public TestClassTests(TestApplicationFactory<TestStartup> testApp) : base(testApp)
+        {
+        }
+
+        [Fact]
+        public async Task Body_Valid()
+        {
+            var body = new TestClass { Name = "Test name" };
+
+            var client = BuildApp().CreateClient();
+            var response = await client.PostAsJsonAsync(nameof(TestController.TestClassBody), body);
+
+            response.AssertOk();
+        }
+
+        [Fact]
+        public async Task Body_Invalid()
+        {
+            var body = new TestClass { Name = "" };
+
+            var client = BuildApp().CreateClient();
+            var response = await client.PostAsJsonAsync(nameof(TestController.TestClassBody), body);
+
+            var details = response.AssertValidationProblem();
+
+            var errors = details.Errors;
+
+            Assert.Single(errors);
+
+            Assert.Equal("name", errors[0].Path);
+            Assert.Equal(NotEmptyCode, errors[0].Code);
+            Assert.Equal(NotEmptyMessage, errors[0].Message);
+            Assert.Null(errors[0].Detail);
+        }
+
+        [Fact]
+        public async Task BodyNested_Valid()
+        {
+            var body = new TestClass
+            {
+                Name = "Test name",
+                TestAddresses = new List<TestAddress>
+                {
+                    new()
+                    {
+                        Line1 = "Test line 1"
+                    }
+                }
+            };
+
+            var client = BuildApp().CreateClient();
+            var response = await client.PostAsJsonAsync(nameof(TestController.TestClassBody), body);
+
+            response.AssertOk();
+        }
+
+        [Fact]
+        public async Task BodyNested_Invalid()
+        {
+            var body = new TestClass
+            {
+                Name = "",
+                TestAddresses = new List<TestAddress>
+                {
+                    new()
+                    {
+                        Line1 = ""
+                    }
+                }
+            };
+
+            var client = BuildApp().CreateClient();
+            var response = await client.PostAsJsonAsync(nameof(TestController.TestClassBody), body);
+
+            var details = response.AssertValidationProblem();
+
+            var errors = details.Errors;
+
+            Assert.Equal(2, errors.Count);
+
+            Assert.Equal("name", errors[0].Path);
+            Assert.Equal(NotEmptyCode, errors[0].Code);
+            Assert.Equal(NotEmptyMessage, errors[0].Message);
+            Assert.Null(errors[0].Detail);
+
+            Assert.Equal("testAddresses[0].line1", errors[1].Path);
+            Assert.Equal("MinimumLength", errors[1].Code);
+            Assert.Equal("Must be at least 10 characters (was 0).", errors[1].Message);
+
+            // Assert.Equal(3, errors[1].Detail!.Count);
+            // var minLength = Assert.IsType<JsonElement>(errors[1].Detail!["MinLength"]);
+            // Assert.Equal(10, minLength.GetInt32());
+        }
+
+        [Fact]
+        public async Task Form_Valid()
+        {
+            var form = new FormUrlEncodedContent(
+                new Dictionary<string, string>
+                {
+                    { "name", "Test name" },
+                }
+            );
+
+            var client = BuildApp().CreateClient();
+            var response = await client.PostAsync(nameof(TestController.TestClassForm), form);
+
+            response.AssertOk();
+        }
+
+        [Fact]
+        public async Task Form_Invalid()
+        {
+            var form = new FormUrlEncodedContent(
+                new Dictionary<string, string>
+                {
+                    { "name", "" },
+                }
+            );
+
+            var client = BuildApp().CreateClient();
+            var response = await client.PostAsync(nameof(TestController.TestClassForm), form);
+
+            var details = response.AssertValidationProblem();
+
+            var errors = details.Errors;
+
+            Assert.Single(errors);
+
+            Assert.Equal("name", errors[0].Path);
+            Assert.Equal(NotEmptyCode, errors[0].Code);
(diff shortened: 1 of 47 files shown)


src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/HttpResponseMessageTestExtensions.cs
```diff
@@ -78,6 +82,46 @@ public static void AssertNotModified(this HttpResponseMessage message)
...
+    public static Task<ProblemDetails> AssertInternalServerError(this HttpResponseMessage message)
+    {
+        Assert.Equal(InternalServerError, message.StatusCode);
+
+        return message.AssertBodyIsProblemDetails();
+    }
+
+    public static async Task<ProblemDetails> AssertBodyIsProblemDetails(this HttpResponseMessage message)
+    {
+        var details = await HttpContentJsonExtensions.ReadFromJsonAsync<ProblemDetails>(message.Content);
```
jack-hive, line 108: could just do `message.Content.ReadFromJsonAsync<ProblemDetails>()`

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/HttpResponseMessageTestExtensions.cs
```diff
@@ -78,6 +82,46 @@ public static void AssertNotModified(this HttpResponseMessage message)
...
+
+        Assert.NotNull(details);
+
+        return details;
+    }
+
+    public static async Task<ValidationProblemViewModel> AssertBodyIsValidationProblem(
+        this HttpResponseMessage message)
+    {
+        var details = await HttpContentJsonExtensions.ReadFromJsonAsync<ValidationProblemViewModel>(message.Content);
```
jack-hive, line 118: could just do `message.Content.ReadFromJsonAsync<ValidationProblemViewModel>()`

src/GovUk.Education.ExploreEducationStatistics.Common/ViewModels/ValidationProblemViewModel.cs
```diff
@@ -0,0 +1,54 @@
...
+public class ValidationProblemViewModel : ValidationProblemDetails
+{
+    /// <summary>
+    /// Additional errors that are related to the problem.
+    /// </summary>
+    public new IReadOnlyList<ErrorViewModel> Errors { get; set; } = new List<ErrorViewModel>();
+
+    [JsonIgnore]
+    [Newtonsoft.Json.JsonIgnore]
+    public ValidationProblemDetails? OriginalDetails { get; init; }
```
jack-hive, line 20: I'm assuming we just want this for debugging purposes?

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Config/FluentValidationActionFilterTests.cs
```diff
@@ -0,0 +1,389 @@
...
+        [Fact]
+        public async Task BodyNested_Valid()
+        {
+            var body = new TestClass { Name = "Test name" };
+
+            var client = BuildApp().CreateClient();
+            var response = await client.PostAsJsonAsync(nameof(TestController.TestClassBody), body);
+
+            response.AssertOk();
+        }
```
jack-hive, line 74: I think this is a duplicate of `Body_Valid`

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Config/FluentValidationActionFilterTests.cs
```diff
@@ -0,0 +1,389 @@
...
+
+        [Fact]
+        public async Task Body_Invalid()
```
jack-hive, line 45: This looks like a valid scenario to me? Maybe just need to change the name of the test

jack-hive (changes requested):

src/explore-education-statistics-common/src/services/types/problemDetails.ts
```diff
@@ -0,0 +1,17 @@
+export interface ProblemDetails {
+  title: string;
+  type: string;
+  status: number;
```
jack-hive, line 4: Do we want to include `detail` on here for completeness?

jack-hive: Or are we purposefully leaving it off for succinctness since we don't use it?

src/explore-education-statistics-common/src/components/form/rhf/__tests__/RHFForm.test.tsx
```diff
@@ -590,8 +592,14 @@ describe('Form', () => {
...
 
+      expect(
+        screen.getByText(
+          'The form submission is invalid and could not be processed',
+        ),
+      ).toBeInTheDocument();
       expect(screen.queryByText('Global error')).not.toBeInTheDocument();
-      expect(screen.queryByText('Invalid field 1')).not.toBeInTheDocument();
+      expect(screen.queryByText('UnmappedCode')).not.toBeInTheDocument();
+      expect(screen.queryByText('Invalid field 2')).not.toBeInTheDocument();
```
jack-hive, line 602: `Invalid field 1`

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Common/Validators/ValidationUtils.cs
```diff
@@ -1,36 +1,43 @@
...
     {
-        public static ActionResult ValidationResult(params Enum[] errors)
-        {
-            return ValidationResult(errors.Select(error => error.ToString()).ToArray());
-        }
-        
-        public static ActionResult ValidationResult(params string[] errors)
-        {
-            var problemDetails = new ValidationProblemDetails(
-                errors.ToDictionary(
```
jack-hive, line 20: I think this might have been broken before? Adding the same empty key if you supplied multiple errors in the method parameters

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Common/Validators/ValidationUtils.cs
```diff
@@ -1,36 +1,43 @@
...
+    public static BadRequestObjectResult ValidationResult<TEnum>(params TEnum[] codes) where TEnum : Enum
     {
-        public static ActionResult ValidationResult(params Enum[] errors)
-        {
-            return ValidationResult(errors.Select(error => error.ToString()).ToArray());
-        }
-        
-        public static ActionResult ValidationResult(params string[] errors)
-        {
-            var problemDetails = new ValidationProblemDetails(
```
jack-hive, line 19: I think this might have been broken before? Adding the same empty key if you supplied multiple errors in the method parameters

src/explore-education-statistics-common/src/services/types/problemDetails.ts
```diff
@@ -0,0 +1,17 @@
+export interface ProblemDetails {
```
amyb-hiveit, line 1: `ProblemDetails` only seems to be used to be extended for `ValidationProblemDetails`, so does it need to exist separately and be exported?

amyb-hiveit (commented):
I've had a look over the frontend parts of this, just one query, otherwise seems fine 👍

src/explore-education-statistics-common/src/services/types/problemDetails.ts
ntsim: This mirrors how errors are modelled in the back end where `ProblemDetails` is the base error response type for .NET APIs.

Whilst we're not currently using this elsewhere, it's likely this will have some utility in some future case so it's a good idea to export it.

src/explore-education-statistics-common/src/services/types/problemDetails.ts
ntsim: I left it out for succinctness originally, but I've included it now.

To make things clearer for any consumers, I've also added docblocks for all the properties as well.

src/GovUk.Education.ExploreEducationStatistics.Common/Config/FluentValidationLanguageManager.cs
```diff
@@ -0,0 +1,91 @@
+using System.Collections.Generic;
+using FluentValidation.Resources;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Config;
+
+public class FluentValidationLanguageManager : LanguageManager
+{
+    public FluentValidationLanguageManager()
```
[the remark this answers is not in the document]
jack-hive: Sounds useful then! :)

src/GovUk.Education.ExploreEducationStatistics.Common/ViewModels/ValidationProblemViewModel.cs
```diff
@@ -0,0 +1,60 @@
...
+
+    public static ValidationProblemViewModel Create(
+        ValidationProblemDetails problemDetails,
+        IEnumerable<ErrorViewModel> errors)
+    {
+        var viewModel = new ValidationProblemViewModel
+        {
+            Title = problemDetails.Title,
+            Type = problemDetails.Type,
+            Status = problemDetails.Status ?? default,
```
jack-hive, line 49: I think this will end up defaulting the `Status` to `0` if `problemDetails.Status` is `null`. I'd just remove the line completely and let the default `400` status you've set with the `new` keyword take over

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Common/Config/ProblemDetailsResultFilter.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            return;
+        }
+
+        if (problemDetails is ValidationProblemViewModel validationProblemViewModel)
+        {
+            if (validationProblemViewModel.OriginalDetails is null)
+            {
+                ApplyValidationProblemDefaults(context, validationProblemViewModel);
+            }
```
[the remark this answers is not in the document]
jack-hive: ah yes. I can see that now

src/GovUk.Education.ExploreEducationStatistics.Common/ViewModels/ValidationProblemViewModel.cs
jack-hive: I'm with you. I can see this better now with a fresh pair of eyes

jack-hive (approved):
looks good! :)

jack-hive (approved):

PR #4543, opened 2024-01-29
EES-3986 Extend PRA into Publish Day until Release is actually Published

sambiramairelogic:
When scheduling a Release, the user is prompted to enter a date sans time, which is saved as its floor (midnight beginning of Publish Day). 
The Pre-Release Access period is calculated using config values, but typically runs from 1440 minutes before that midnight (= the previous midnight) through to 1 minute before. 
However the Release itself isn't available until 9:30 that day, and attempting to access it between midnight and 9:30 will yield a message saying the PRA period has ended. This PR aims to extend the PRA such that Releases are still viewable up to their actual release. 

The simpler route might be to simply change the calculation for the PRA end time. This is fraught with danger however as these are existing config values, and this period might be used for other calculations elsewhere in the codebase. To maintain safety I propose to introduce the new concept of the `Publish Day PRA Lenience Period`, calculated by the delta between midnight and when the Cron job occurs to complete the Release. 

This is further complicated by when those Cron jobs occur - they have been normalised to always occur at 9:30 local time. This means we have to calculate a different offset during GMT and BST respectively.

This diagram is intended as a visual aid to the status quo, and the proposed change:
PreRelease Dates

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PreReleaseServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PreReleaseServiceTests.cs
@@ -30,13 +30,6 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
             }
         }));
 
-        private readonly Release _testRelease = new()
-        {
-            PublishScheduled = DefaultScheduledPublishDate,
-            Published = null,
-            ApprovalStatus = ReleaseApprovalStatus.Approved
-        };
-
 
         [Theory]
         [InlineData("2003-11-10T00:00:00.00Z", false, PreReleaseAccess.Before)]
@@ -68,9 +61,16 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
         }
 
         [Fact]
-        public void CalculatesAllAccessRangeFieldsAsExpected()
+        public void CalculatesStartAndScheduledPublishDatesAsExpected()
         {
-            var result = _sut.GetPreReleaseWindowStatus(_testRelease, DateTime.Parse("2003-11-14T10:24:00.00Z", styles: DateTimeStyles.AdjustToUniversal));
+            var testRelease = new Release()
+            {
+                PublishScheduled = DefaultScheduledPublishDate,
+                Published = null,
+                ApprovalStatus = ReleaseApprovalStatus.Approved
+            };
+
+            var result = _sut.GetPreReleaseWindowStatus(testRelease, DateTime.Parse("2003-11-14T10:24:00.00Z", styles: DateTimeStyles.AdjustToUniversal));
 
             Assert.Equal(PreReleaseAccess.Within, result.Access);
             Assert.Equal(DateTime.Parse("2003-11-14T00:00:00.00Z", styles: DateTimeStyles.AdjustToUniversal), result.Start);
@@ -80,28 +80,28 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
         [Fact]
         public void AccessIsNoneSetIfPublishedScheduleIsStillInReview()
         {
-            var releaseWithNoInformation = new Release()
+            var testReleaseInReview = new Release()
             {
                 PublishScheduled = DefaultScheduledPublishDate,
                 Published = null,
                 ApprovalStatus = ReleaseApprovalStatus.HigherLevelReview
             };
 
-            var preReleaseWindowStatus = _sut.GetPreReleaseWindowStatus(releaseWithNoInformation, DateWithinPreReleaseAccessWindow);
+            var preReleaseWindowStatus = _sut.GetPreReleaseWindowStatus(testReleaseInReview, DateWithinPreReleaseAccessWindow);
             Assert.Equal(PreReleaseAccess.NoneSet, preReleaseWindowStatus.Access);
         }
 
         [Fact]
         public void AccessIsNoneSetIfPublishedScheduleIsStillInDraft()
         {
-            var releaseWithNoInformation = new Release()
+            var testReleaseInDraft = new Release()
             {
                 PublishScheduled = DefaultScheduledPublishDate,
                 Published = null,
                 ApprovalStatus = ReleaseApprovalStatus.Draft
             };
 
-            var preReleaseWindowStatus = _sut.GetPreReleaseWindowStatus(releaseWithNoInformation, DateWithinPreReleaseAccessWindow);
+            var preReleaseWindowStatus = _sut.GetPreReleaseWindowStatus(testReleaseInDraft, DateWithinPreReleaseAccessWindow);
             Assert.Equal(PreReleaseAccess.NoneSet, preReleaseWindowStatus.Access);
         }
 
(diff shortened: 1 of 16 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PreReleaseService.cs
```diff
@@ -71,28 +66,52 @@ private DateTime GetEndTime(DateTime publishScheduled)
             return publishScheduled.AddMinutes(-_preReleaseOptions.MinutesBeforeReleaseTimeEnd);
         }
 
+        private DateTime GetPublishDayLenienceDeadline(DateTime publishScheduled)
```
duncan-at-hiveit, line 69: This'll probably work OK so long as each environment has the correct configuration provided.  On Prod and Pre-Prod, this'd be 570, on Dev it would probably be 120, and locally it would probably be 2.

What I was saying on our call thoug is that perhaps we can work this out dynamically based upon the first time the "930 job" runs on the given day that we're working out the lenience for.

An example of where we use this time is in `ReleaseService.ValidatePublishDate`.

duncan-at-hiveit: If we're doing the "subtract 60 minutes" logic on an environment like local or Dev, you'll probably end up with negative numbers too, FYI.

duncan-at-hiveit: The other super simple other approach could be to say that if you're already past midnight and the Release is due out today, simply allow PRA access up until the Release is published.  Literally "Is it past midnight (or the original PRA access window end) and has it been published yet?  No?  Then they can still see it!"

sambiramairelogic: Precisely - I'm hoping I've compensated for that with: 
`Math.Max(_preReleaseOptions.MinutesIntoPublishDayByWhichPublishingHasOccurred - 60, 0)`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PreReleaseService.cs
```diff
@@ -51,13 +52,12 @@ public PreReleaseWindowStatus GetPreReleaseWindowStatus(Release release, DateTim
...
             var startTime = GetStartTime(publishScheduled);
-            var endTime = GetEndTime(publishScheduled);
 
             return new PreReleaseWindowStatus
             {
-                Start = startTime,
-                End = endTime,
-                Access = GetAccess(release, startTime, endTime, referenceTime)
+                Start = GetStartTime(publishScheduled),
+                ScheduledPublishDate = publishScheduled,
```
duncan-at-hiveit, line 59: Is it worth keeping the variable names consistent and renaming this to `PublishScheduledDate`?

src/explore-education-statistics-admin/src/pages/release/pre-release/__tests__/PreReleasePageContainer.test.tsx
```diff
@@ -130,6 +130,47 @@ describe('PreReleasePageContainer', () => {
...
+
+      expect(banner.getByText('Test team:')).toBeInTheDocument();
+      expect(
+        banner.getByRole('link', { name: 'test@test.com' }),
+      ).toBeInTheDocument();
+
+      expect(screen.getByRole('link', { name: 'Content' })).toBeInTheDocument();
+
+      expect(
+        screen.queryByRole('heading', {
```
duncan-at-hiveit, line 162: Hey @amyb-hiveit do we need to adhere to some conventions about queryByRole usage in waitFors? Some performance implications?

src/explore-education-statistics-admin/src/services/permissionService.ts
```diff
@@ -33,7 +33,7 @@ export type PreReleaseAccess = 'Before' | 'After' | 'Within' | 'NoneSet';
 export interface PreReleaseWindowStatus {
   access: PreReleaseAccess;
   start: Date;
-  end: Date;
+  scheduledPublishDate: Date;
```
duncan-at-hiveit, line 36: Again to keep consistent with current naming, would it be worth renaming this to `publishScheduled`?

tests/robot-tests/tests/admin/bau/prerelease.robot
```diff
@@ -247,12 +248,13 @@ Validate prerelease has not started for Analyst user
     user checks nth breadcrumb contains    1    Home
     user checks nth breadcrumb contains    2    Pre-release access
 
-    ${day}=    get current datetime    %d    1
-    ${month}=    get current datetime    %m    1
-    ${year}=    get current datetime    %Y    1
-    ${time_start}=    format uk to local datetime    ${year}-${month}-${day}T00:00:00    %-d %B %Y at %H:%M
-    ${time_end}=    format uk to local datetime    ${year}-${month}-${day}T23:59:00    %-d %B %Y at %H:%M
-    user checks page contains    Pre-release access will be available from ${time_start} until ${time_end}.
+    ${tomorrow}=    get current datetime    %Y-%m-%dT00:00:00    1
```
duncan-at-hiveit, line 251: Nice!  Very nice to have it boiled down to 1 line :)

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PreReleaseServiceTests.cs
```diff
@@ -0,0 +1,122 @@
...
+using Xunit;
+using GovUk.Education.ExploreEducationStatistics.Content.Model;
+using System.Globalization;
+using GovUk.Education.ExploreEducationStatistics.Admin.Models;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
+{
+    public class PreReleaseServiceTests
+    {
+        private static readonly DateTime DefaultScheduledPublishDate =
```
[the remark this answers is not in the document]
sambiramairelogic: Aha, fair enough. Will change! 
I often choose whatever constructor or parse method accepts an ISO string because it's language agnostic - but that's just a hangover from dealing with Javascript's native `Date` object which indexes *some* fields from 0 and others from 1... amongst other madness. Should of course have checked EES' conventions!

src/explore-education-statistics-admin/src/services/permissionService.ts
sambiramairelogic: I did debate this a bit, but eventually chose to rename this for increased readability. 
To me, `publishScheduled` sounds like a boolean denoting whether a publish is scheduled or not, and `ScheduledPublishDate` is a bit more descriptive. 

I appreciate it sort of implicitly *is* actually a boolean because we hinge logic off of whether it's null or not, but I had to read a fair bit of code to check that assumption. 

Certainly happy to revert the name though in the cause of consistency if you prefer?

src/explore-education-statistics-admin/src/pages/release/pre-release/__tests__/PreReleasePageContainer.test.tsx
sambiramairelogic: There's no reason this assertion should actually need to be within a `waitFor`. It should be possible to do the awaiting once to give it time to render, then make these assertions synchronously. I was just copying from the tests above in this case, so might give them the same treatment too :)

duncan-at-hiveit (approved):

PR #4544, opened 2024-01-29
EES-4807 Add `Public.Data.Processor` project with HelloWorldFunction

ntsim:
This PR creates a new `Public.Data.Processor` function project (with a basic `HelloWorldFunction`). 

We need this to enable @Dancemammal to integrate this project into the build and deployment pipeline as part of EES-4807.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/HelloWorldFunction.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/HelloWorldFunction.cs
@@ -1,0 +1,27 @@
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Extensions.Logging;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Functions;
+
+public class HelloWorldFunction
+{
+    private readonly ILogger<HelloWorldFunction> _logger;
+
+    public HelloWorldFunction(ILogger<HelloWorldFunction> logger)
+    {
+        _logger = logger;
+    }
+
+    [Function("HelloWorld")]
+    public IActionResult Run(
+        [HttpTrigger(AuthorizationLevel.Function, "get", "post")] HttpRequest req,
+        FunctionContext executionContext)
+    {
+        _logger.LogInformation("C# HTTP trigger function processed a request.");
+
+        return new OkObjectResult("Welcome to Azure Functions!");
+
+    }
+}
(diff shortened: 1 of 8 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Program.cs
```diff
@@ -0,0 +1,7 @@
+using Microsoft.Extensions.Hosting;
+
+var host = new HostBuilder()
+    .ConfigureFunctionsWorkerDefaults()
```
jack-hive, line 4: Since we're upgrading to use the isolated process model, I wonder if we should also use the [ASP.NET Core integration](https://learn.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide?tabs=windows#aspnet-core-integration)? 

We would just need to change this line to `.ConfigureFunctionsWebApplication()`, and include the package reference for `Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore`

It then means we can use types from ASP.NET Core like `HttpRequest`, `HttpResponse` and `IActionResult` which would make it more consistent with our other projects.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/HelloWorldFunction.cs
```diff
@@ -0,0 +1,32 @@
...
+    private readonly ILogger _logger;
+
+    public HelloWorldFunction(ILoggerFactory loggerFactory)
+    {
+        _logger = loggerFactory.CreateLogger<HelloWorldFunction>();
+    }
+
+    [Function("HelloWorld")]
+    public HttpResponseData Run(
+        [HttpTrigger(AuthorizationLevel.Function, "get", "post")] HttpRequestData req,
```
jack-hive, line 19: Then we can change `HttpRequestData` here to `HttpRequest`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/HelloWorldFunction.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    private readonly ILogger _logger;
+
+    public HelloWorldFunction(ILoggerFactory loggerFactory)
+    {
+        _logger = loggerFactory.CreateLogger<HelloWorldFunction>();
+    }
+
+    [Function("HelloWorld")]
+    public HttpResponseData Run(
```
jack-hive, line 18: And `HttpResponseData` here to `IActionResult`

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Program.cs
ntsim: Nice idea! I've added those in now and its much more ergonomic to use the standard .NET Core APIs!

jack-hive (approved):
Looks good! :)

PR #4550, opened 2024-01-30
EES-4817 Add test data generator extensions for public API data model

ntsim:
This PR creates the various generator extension methods needed for the public API data model in upcoming work. We've primarily focused around data sets, versions and their meta. We haven't included change sets as these will be introduced later on and aren't necessary right now.

## Related changes

- Changed the format for default strings generated for properties to `{ClassType} :: {PropertyName}` e.g. `Person 0 :: FirstName`. This is simpler and more clear than the previous format.
- Added `Generator.GenerateRandom` methods to allow generating of random numbers of instances. The number generated is then limited by a maximum number of instances parameter.

## Usage

There's lots of different ways to use these generators, but sensible defaults have been set for most things to help you on your way.

For the most part, you only need to fill in the parts that are typically relevant. The setter extensions try and handle setting any related fields.

- `DataSet.LatestVersion` will also set `DataSet.Versions`
- `DataSet.Versions` will also set `DataSet.LatestVersion` (if applicable)
- `DataSetVersion.Meta` will also set `DataSetVersion.MetaSummary`
- Any bi-directional relationships will be configured

### Example controlling specific version numbers

In this example, we control the specific versions generated and their meta. We also only publish two of the versions.

```cs
DataSet dataSet = _dataFixture.DefaultDataSet()
    .WithStatusPublished()
    .WithVersions(
        _dataFixture.DefaultDataSetVersion()
            .ForIndex(0, s => s
                .SetVersionNumber(1, 0)
                .SetStatusPublished()
                .SetMeta(_dataFixture.DefaultDataSetMeta(
                    filters: 5, locations: 3, timePeriods: 10, indicators: 6)))
            .ForIndex(1, s => s
                .SetVersionNumber(2, 0)
                .SetStatusPublished()
                .SetMeta(_dataFixture.DefaultDataSetMeta(
                    filters: 6, locations: 3, timePeriods: 15, indicators: 6, maxLocationOptions: 15)))
            .ForIndex(2, s => s
                .SetVersionNumber(3, 0)
                .SetMeta(_dataFixture.DefaultDataSetMeta(
                    filters: 7, 
                    locations: 4, 
                    timePeriods: 20, 
                    indicators: 7,
                    maxLocationOptions: 20,
                    maxFilterOptions: 15)))
            .GenerateList()
    );
```

Note that we have a bunch of parameters for `DefaultDataSetMeta` that allow us to quickly spin up however much meta of each type that we need.

Note that we also have `maxLocationOptions` and `maxFilterOptions` too. These will deterministically create a random number of options for each filter and location meta, limited by the respective `max*` parameter.

### Example using only defaults

If you want to go even further, we've configured the default methods to pass through all the default meta parameters, so things can be written really tersely as:

```cs
 DataSet dataSet = _dataFixture.DefaultDataSet()
     .WithStatusPublished()
     .WithVersions(
         _dataFixture.DefaultDataSetVersion(filters: 5, locations: 3, timePeriods: 10, indicators: 6)
             .WithStatusPublished()
             .Generate(3)
     );
```

In this example, we lose control over things like the version number, but this defaults to incrementing each version as a minor bump instead.

We also lose the individual unpublished version that we had in the previous example. Depending on your situation, this is probably acceptable for your needs.

Remember that you can always use `ForIndex` and `ForRange` to set specific properties on specific instances within the sequence being generated.


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetMeta.cs
```diff
@@ -28,6 +28,21 @@ public class DataSetMeta : ICreatedUpdatedTimestamps<DateTimeOffset, DateTimeOff
...
+    {
+        return new DataSetVersionMetaSummary
+        {
+            Filters = Filters.Select(f => f.Label).ToList(),
+            Indicators = Indicators.Select(i => i.Label).ToList(),
+            GeographicLevels = GeographicLevels,
+            TimePeriodRange = new TimePeriodRange
+            {
+                Start = TimePeriods.First(),
+                End = TimePeriods.Last()
```
jack-hive, line 41: Oooo is this risky? Might not be guaranteed

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/DataSetMetaGeneratorExtensions.cs
```diff
@@ -0,0 +1,115 @@
...
+        int filters,
+        int indicators,
+        int locations,
+        int timePeriods,
+        int maxFilterOptions = 10,
+        int maxLocationOptions = 10)
+    {
+        var locationMeta = fixture.DefaultLocationMeta()
+            .WithOptions(() => fixture.DefaultLocationOptionMeta().GenerateRandomList(maxLocationOptions))
+            .GenerateList(locations);
```
jack-hive, line 22: Might as well put this inline below like with the filters. Or, assign the filters, indiciators and geographic levels a variable too to be consistent? Don't feel strongly either way though

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/FilterMetaGeneratorExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+        this Generator<FilterMeta> generator,
+        Func<IEnumerable<FilterOptionMeta>> options)
+        => generator.ForInstance(s => s.SetOptions(options));
+
+    public static InstanceSetters<FilterMeta> SetDefaults(this InstanceSetters<FilterMeta> setters)
+        => setters
+            .SetDefault(m => m.Identifier)
+            .SetDefault(m => m.Label)
+            .SetDefault(m => m.Hint)
+            .Set(m => m.Options, []);
```
jack-hive, line 42: do we not want to set any default options?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/LocationMetaGeneratorExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+        => generator.ForInstance(s => s.SetOptions(options));
+
+    public static InstanceSetters<LocationMeta> SetDefaults(this InstanceSetters<LocationMeta> setters)
+        => setters
+            .Set(
+                m => m.Level,
+                (_, _, context) =>
+                    GeographicLevelUtils.Levels[context.Index % GeographicLevelUtils.Levels.Length]
+            )
+            .Set(m => m.Options, []);
```
jack-hive, line 40: same here, do we think it would be more helpful to leave this empty?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/TimePeriodMetaGeneratorExtensions.cs
```diff
@@ -0,0 +1,41 @@
+using GovUk.Education.ExploreEducationStatistics.Common.Model;
+using GovUk.Education.ExploreEducationStatistics.Common.Tests.Fixtures;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests.Fixtures;
+
+public static class TimePeriodMetaGeneratorExtensions
+{
+    private const int DefaultStartYear = 2015;
```
jack-hive, line 8: maybe start at `2000` like in the other place? I.e. `DataSetVersionMetaSummaryGeneratorExtensions`

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/DataSetMetaGeneratorExtensions.cs
ntsim: We've deliberately set this as separate variable because we want to only generate a single list of location meta. We're then using this twice, once in `WithGeographicLevels` and `WithLocations`.

If we inlined it, we'd end up generating two lists that are potentially different and would  mean the geographic levels and locations aren't synced up properly.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/FilterMetaGeneratorExtensions.cs
ntsim: Think for this case, it's fine for it to be empty as it's the base `SetDefaults`. We need to be able to generate a default without its relationships populated as you're not necessarily going to need them in lots of tests.

However, for where we do want to use generated defaults, we'd use the `DataFixture.DefaultFilterMeta` extension instead e.g.

```cs
var filterMeta = DataFixture.DefaultFilterMeta(options: 5);
```

I've actually just removed the line you commented on and replaced it with a default on the `FilterMeta` entity itself

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/LocationMetaGeneratorExtensions.cs
ntsim: Similar to `FilterMeta`, we need a base `SetDefaults` that doesn't set relationships.

I've also added in the empty list as a default to `LocationMeta` itself

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/TimePeriodMetaGeneratorExtensions.cs
ntsim: Yep, fair point!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/DataSetMetaGeneratorExtensions.cs
jack-hive: didn't realise, it's good where it is in that case :)

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetMeta.cs
ntsim: Discussed in person, but essentially the assumption is that time periods should be ordered by the code that saves the `DataSetMeta` into the database.

Ordering around time periods is also quite a tricky topic so it's better being dealt outside of this mapping function anyway.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/FilterMetaGeneratorExtensions.cs
jack-hive: cool

jack-hive (approved):

PR #4555, opened 2024-01-31
Add Notifier README file

tomjonesdev:
Add local development details for the Notifier Azure function.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Notifier/README.md
+++ b/src/GovUk.Education.ExploreEducationStatistics.Notifier/README.md
@@ -1,0 +1,17 @@
+# Notifier Functions
+
+This project contains Azure Functions responsible for managing subscriptions to publications, and notifying users of changes to their subscribed publications.
+
+The [GOV.UK Notify](https://www.notifications.service.gov.uk) service is used for sending notifications by e-mail.
+
+## Local development and testing
+
+Create a copy of the `appsettings.Local.json.example` file found in the root of the project directory and call it `appsettings.Local.json`.
+
+*The new file should be automatically ignored by source control as it will contain sensitive information.*
+
+To obtain a `NotifyApiKey`, a member of the EES team will need to invite you to the Gov.UK Notify service.
+
+Once invited, you will also gain access to the various email templates used by the service, each with its own ID, which can be pasted into the `...TemplateId` sections marked `"change-me"` in the example JSON.
+
+*IMPORTANT: The templates should **not** be edited for testing purposes, as they are used by the production application. To test a new or updated template, create a new template and reference it's ID in the JSON above.*
--- a/src/GovUk.Education.ExploreEducationStatistics.Notifier/README.md
+++ b/src/GovUk.Education.ExploreEducationStatistics.Notifier/README.md
@@ -2,16 +2,36 @@
 
 This project contains Azure Functions responsible for managing subscriptions to publications, and notifying users of changes to their subscribed publications.
 
-The [GOV.UK Notify](https://www.notifications.service.gov.uk) service is used for sending notifications by e-mail.
+It uses [GOV.UK Notify](https://www.notifications.service.gov.uk) to send notifications by email.
 
 ## Local development and testing
 
-Create a copy of the `appsettings.Local.json.example` file found in the root of the project directory and call it `appsettings.Local.json`.
+To test sending emails when running the Azure Functions app locally, you will need to configure the GOV.UK Notify API key and the email template ids. These are kept as secrets and should not be committed to the source code repository.
 
-*The new file should be automatically ignored by source control as it will contain sensitive information.*
+Create a copy of the `appsettings.Local.json.example` as `appsettings.Local.json`.
 
-To obtain a `NotifyApiKey`, a member of the EES team will need to invite you to the Gov.UK Notify service.
+```bash
+cp src/GovUk.Education.ExploreEducationStatistics.Notifier/appsettings.Local.json.example src/GovUk.Education.ExploreEducationStatistics.Notifier/appsettings.Local.json
+```
 
-Once invited, you will also gain access to the various email templates used by the service, each with its own ID, which can be pasted into the `...TemplateId` sections marked `"change-me"` in the example JSON.
+`appsettings.Local.json` is already in the `.gitignore` file, so it should remain as untracked.
 
-*IMPORTANT: The templates should **not** be edited for testing purposes, as they are used by the production application. To test a new or updated template, create a new template and reference it's ID in the JSON above.*
+Update the `GovUkNotify` section of the `appsettings.Local.json` file by substituting values `change-me` with values for the `ApiKey` and each of the email templates ids.
+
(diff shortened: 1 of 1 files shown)


src/GovUk.Education.ExploreEducationStatistics.Notifier/local.settings.json
benoutram, line 1: I think it would be a good idea to revert this and move it to `local.settings.json.example` instead.
That way it's easier to maintain and keep it formatted, and taking a copy to get started is as easy as duplicating the file without the `.example` extension.

You could mention in the instructions to copy `local.settings.json.example` to `local.settings.json`.

.gitignore
```diff
@@ -90,3 +90,6 @@ src/GovUk.Education.ExploreEducationStatistics.Admin/appsettings.IdpBootstrapUse
 
 ## IdentityServer Developer Signing key
 **/*/tempkey.rsa
+
+## Local Gov.UK Notify integration settings
+/src/GovUk.Education.ExploreEducationStatistics.Notifier/local.settings.json
```
benoutram, line 95: In the comment I would call this `Notifier Local settings file` because it contains more than just the Gov.uk Notify integration settings.

You could also put this in a new `.gitignore` file in the root of the Notifier project. We have mutliple `.gitignore` files so there's already precedent for doing this.

src/GovUk.Education.ExploreEducationStatistics.Notifier/README.md
```diff
@@ -0,0 +1,44 @@
+# Notifier Functions
+
+This project contains Azure Functions responsible for notifying users of changes to their subscribed publications using the Gov.UK Notify service.
```
benoutram, line 3: Something along the lines of `responsible for managing subscriptions to publications and notifying users of changes to their subscribed publications using the Gov.UK Notify service` would be more appropriate because it also contains functions that allow users to subscribe and unsubscribe to the publications.

src/GovUk.Education.ExploreEducationStatistics.Notifier/README.md
```diff
@@ -0,0 +1,44 @@
...
+  "ConnectionStrings": {
+  },
+  "Host": {
+    "CORS": "*"
+  }
+}
+
+```
+
+*This file should be automatically ignored by source control as it will contain sensitive information.*
```
benoutram, line 36: See comment on the deleted file about this potentially existing as an `.example` file instead of within the README to make it easier to maintain and copy.

src/GovUk.Education.ExploreEducationStatistics.Notifier/GovUk.Education.ExploreEducationStatistics.Notifier.csproj
```diff
@@ -28,9 +28,5 @@
     <None Update="host.json">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
-    <None Update="local.settings.json">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-      <CopyToPublishDirectory>Never</CopyToPublishDirectory>
-    </None>
```
ntsim, line 34: Guess you might not have noticed, but in our other function apps, we have startups that load in `appsettings.Local.json` files that are .gitignored and won't be checked into source control.

We can enable loading of this config file by simply adding:

```cs
public class Startup : FunctionsStartup
{
    public override void ConfigureAppConfiguration(IFunctionsConfigurationBuilder builder)
    {
        var binDir = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
        var rootDir = Path.GetFullPath(Path.Combine(binDir!, ".."));
        builder.ConfigurationBuilder
            .AddJsonFile($"{rootDir}/appsettings.Local.json", optional: true, reloadOnChange: true);
    }
}
```

I think this should get merged together with `local.settings.json` so it should get the behaviour you're looking for.

Would suggest trying this out before going with the proposed approach. It's not ideal to add additional set up steps as this complicates things for anyone coming onto the project.

For most of the project, we've adopted sensible defaults so that things usually work out of the box after the main set up steps in the project README have been completed.

benoutram: @ntsim In #4673 I've adopted the Options pattern for our custom settings.

### local.settings.json

I've kept `local.settings.json` as this seems to be a standard for running functions locally for all example functions projects I've seen, and this has the core properties required to configure a function app.

`local.settings.json` now only has additional properties for `BaseUrl`, `TableStorageConnectionString`, `TokenSecretKey`, `PublicAppUrl`. These are all defaults which should never need to be changed when running locally.

```
{
  "IsEncrypted": false,
  "Values": {
    "AzureWebJobsStorage": "UseDevelopmentStorage=true;DevelopmentStorageProxyUri=http://data-storage",
    "FUNCTIONS_WORKER_RUNTIME": "dotnet-isolated",
    "AppSettings:BaseUrl": "http://localhost:7073/api",
    "AppSettings:TableStorageConnectionString": "AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;DefaultEndpointsProtocol=http;QueueEndpoint=http://data-storage:10001/devstoreaccount1;TableEndpoint=http://data-storage:10002/devstoreaccount1;",
    "AppSettings:TokenSecretKey": "email-token-security-signing-key",
    "AppSettings:PublicAppUrl": "http://localhost:3000"
  },
  "ConnectionStrings": {
  },
  "Host": {
    "LocalHttpPort": 7073,
    "CORS": "*"
  }
}
```

### appsettings.Local.json

I've added additional config during Notifier startup to load the gitignored `appsettings.Local.json` like you mentioned and this is now consistent with our other function apps.

`appsettings.Local.json` can now be used for all the gitignored GovUkNotify config and I've added a `appsettings.Local.json.example` file as follows:

```
{
  "GovUkNotify" : {
    "ApiKey": "",
    "EmailTemplates" : {
      "ReleaseAmendmentPublishedId": "",
      "ReleaseAmendmentPublishedSupersededSubscribersId": "",
      "ReleasePublishedId": "",
      "ReleasePublishedSupersededSubscribersId": "",
      "SubscriptionConfirmationId": "",
      "SubscriptionVerificationId": ""
    }
  }
}
```

If you agree with this then the changes on this PR except for the README could be scrapped when #4673 is merged? I can alter the README here to explain setting up `appsettings.Local.json`.

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Notifier/GovUk.Education.ExploreEducationStatistics.Notifier.csproj
benoutram: Chatted with @ntsim yesterday and there's a slight change to what I proposed above. Each function app now has an `appsettings.json` for our own app variables which don't need to change and has an `appsettings.Local.json` which is gitignored for overrides and secrets.

PR #4558, opened 2024-02-01
EES-4666 Introduce new parent-child relationship with releases by adding `ReleaseParent`

benoutram:
This PR introduces a new `ReleaseParent` entity to create a new parent-child relationship with collections of release versions which are all derived from the same base version.

With this PR, every release version 0 that gets created when creating a new release will be associated with a new parent entity that is assigned its own new unique id. When new release version is created through the process of amending an existing version it will now inherit the same parent as the version its derived from.

A migration is also run to populate a new `ReleaseParent` for every collection of existing release versions and assign `ReleaseParentId` to each of those versions.

This work is based from a proposal originally put together by @Dancemammal in [EES-4487](https://dfedigital.atlassian.net/browse/EES-4487).

The purpose of adding this  parent-child relationship for release versions is to help simplify the way we work out if release versions are related to each other and find the latest release versions.

We commonly want to:

* Determine if a release version is the latest version (or latest published version) within it's release, given a release version id.
* List all the latest versions (or latest published versions) within a publication, given a publication id.
* Get the latest version (or latest published version) from all release versions in reverse chronological order that are associated with a publication, given a publication id.

As mentioned in https://github.com/dfe-analytical-services/explore-education-statistics/pull/4541, these methods rely on loading all release versions for a publication (or from a releases own publication if starting from a release) and then iterating through the releases to eliminate all but those which have no other release version with `PreviousVersionId` equal to their own id. This results in finding the latest release version or versions, depending on whether we begin with a release or a publication.

This PR only adds the new parent-child relationship. It doesn't make any improvements to the efficiency of these methods which is planned for the next bit of work.

With the addition of this relationship, in future we should be able to work out the latest release version (or latest published version) from a release either by querying the top version from the parent's versions, or from querying version id's persisted to the parent entity, depending on how we decide to implement it.

One slight change that I did make in this PR to take advantage of `ReleaseParent` straight away has been to rewrite `ReleaseApprovalService.GetReleaseStatuses` as `ListReleaseStatuses`. This previously used `ReleaseRepository.GetAllReleaseVersionIds` which ran database queries to fetch each previous version by `PreviousVersionId` one by one.  This now becomes a simple query using `ReleaseParentId`.

### Other changes

- Added `ReleaseStatusGeneratorExtensions`.
- Rewrote `ReleaseRepository.CreateStatisticsDbReleaseAndSubjectHierarchy` to not depend on AutoMapper since it only copies `PublicationId`. Also rewrote its unit tests.
- Removed `Versioned` abstract class which wasn't adopted by any other classes except for `Release` and change `Release` to implement `ICreatedTimestamp` instead.
- Removed unused interface `IPublicationService`.
- Made additions to the team dictionary.


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseRepository.cs
```diff
@@ -75,46 +71,35 @@ public class ReleaseRepository : IReleaseRepository
...
+                    Id = releaseId,
+                    PublicationId = release.PublicationId
+                });
             }
             else
             {
-                _mapper.Map(release, existingRelease);
-                _statisticsDbContext.Release.Update(existingRelease);
+                existingStatsRelease.PublicationId = release.PublicationId;
+                _statisticsDbContext.Release.Update(existingStatsRelease);
```
mmyoungman, line 90: I know you didn't add this, but I don't understand why this `else` is necessary. I think it can be removed.

`CreateStatisticsDbReleaseAndSubjectHierarchy` looks like it is only called when uploading a new file or zip. If the Statistics DB Release entry already exists, we shouldn't need to update it, I think?

There is no way to change the publication for a release, so the publication shouldn't ever change.

mmyoungman (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseRepository.cs
benoutram: I was also confused about this. We shouldn't need to update the publication so I've removed it. Thanks.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseApprovalService.cs
```diff
@@ -69,42 +66,35 @@ public class ReleaseApprovalService : IReleaseApprovalService
...
                         .Select(rs =>
                             new ReleaseStatusViewModel
                             {
                                 ReleaseStatusId = rs.Id,
                                 InternalReleaseNote = rs.InternalReleaseNote,
                                 ApprovalStatus = rs.ApprovalStatus,
                                 Created = rs.Created,
-                                CreatedByEmail = rs.CreatedBy?.Email,
-                                ReleaseVersion = rs.Release.Version,
+                                CreatedByEmail = rs.CreatedBy == null ? null : rs.CreatedBy.Email,
```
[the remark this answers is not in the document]
benoutram: There is a bit of a difference here...

Previously it looked like this:

```csharp
                    var allReleaseVersionIds = await _releaseRepository.GetAllReleaseVersionIds(release);
                    return _context.ReleaseStatus
                        .Include(rs => rs.Release)
                        .Include(rs => rs.CreatedBy)
                        .Where(rs => allReleaseVersionIds.Contains(rs.ReleaseId))
                        .ToList()
                        .Select(rs =>
                            new ReleaseStatusViewModel
                            {
                                ReleaseStatusId = rs.Id,
                                InternalReleaseNote = rs.InternalReleaseNote,
                                ApprovalStatus = rs.ApprovalStatus,
                                Created = rs.Created,
                                CreatedByEmail = rs.CreatedBy?.Email,
                                ReleaseVersion = rs.Release.Version,
                            })
                        .OrderByDescending(vm => vm.Created)
                        .ToList();
```

and now it's this:

```csharp
                    return await _context.ReleaseStatus
                        .Where(rs => rs.Release.ReleaseParentId == release.ReleaseParentId)
                        .OrderByDescending(rs => rs.Created)
                        .Select(rs =>
                            new ReleaseStatusViewModel
                            {
                                ReleaseStatusId = rs.Id,
                                InternalReleaseNote = rs.InternalReleaseNote,
                                ApprovalStatus = rs.ApprovalStatus,
                                Created = rs.Created,
                                CreatedByEmail = rs.CreatedBy == null ? null : rs.CreatedBy.Email,
                                ReleaseVersion = rs.Release.Version
                            })
                        .ToListAsync();
```

In the original version the database query will contain all the columns of ReleaseStatus, joined with all the columns of Release and columns of User, into memory to build the view model, so very inefficient.

In the new version only the columns of `ReleaseStatusViewModel` are in the database query projection so it should be efficient as possible. Note there's no need for any `.Include`'s.

In the original version,  the call is `List<ReleaseStatus>.Select(Func<ReleaseStatus,ReleaseStatusViewModel> selector)`  where `selector` is a delegate function. That function can contain the conditional operator so we could do `rs.CreatedBy?.Email`.

In the latter version, the call is `IQueryable<ReleaseStatus>.Select(Expression<Func<ReleaseStatus,ReleaseStatusViewModel>> selector)` where selector is a lambda expression used for the LINQ query translation of ReleaseStatus to ReleaseStatusViewModel.

A lambda expression cannot contain the conditional operator so we can't do `rs.CreatedBy?.Email`.

mmyoungman (approved):

PR #4561, opened 2024-02-02
EES-4847 Add Swashbuckle schema filter to handle JsonConverter

ntsim:
This PR adds a new `JsonConverterSchemaFilter` to allow Swashbuckle to handle specific types of `JsonConverter`. Without this, Swashbuckle is unable to figure out what the schema should be after a conversion has taken place.

For example:

- `JsonStringEnumConverter` - the name of the enum (not integer)
- `EnumToEnumLabelJsonConverter` - the enum label (configured with attribute)
- `EnumToEnumValueJsonConverter` - the enum value (configured with attribute)

As part of this, we've also added `EnumToEnumValueJsonConverter` for `System.Text.Json` to provide more complete conversion functionality for enums.

## Other changes

- Moved `SwaggerConfig` out of `Config` into `Swagger` namespace
- Added `UseAllOfToExtendReferenceSchemas` to allow property docblocks to override reference type docblocks


jack-hive (commented):
Maybe the `SwaggerConfig.cs` class should be moved to the `.Swagger` namespace?

jack-hive (changes requested):
I tested this with:

```cs
[JsonConverter(typeof(EnumToEnumLabelJsonConverter<GeographicLevel>))]
public GeographicLevel GeographicLevel { get; init; }
```

And it seemed to spit out integers in the Open API schema. Not sure if I configured this wrong or not?

PR #4568, opened 2024-02-07
EES-4667 Add new `ReleaseParent` test data generator options and add unit tests for `ReleaseRepository`

benoutram:
This PR makes 4 changes:

### (1) Enable nullable reference types in `GovUk.Education.ExploreEducationStatistics.Content.Model.Tests`

We've now reached a stage where the majority of files in the `Content.Model.Tests` project have the `#nullable enable` pragma, and the few that don't have no nullable warnings. This PR removes the `#nullable enable` pragma from the files.

### (2) Add new `ReleaseParent` test data generator options

This PR introduces a new override of `ReleaseParentGeneratorExtensions.DefaultReleaseParent` providing options for generating a `ReleaseParent` with a specified number of published release versions, and a draft version. Year can also be provided to override the generated year that's otherwise applied to every version.

Typically we see unit test data setup as follows with variations in the naming e.g. `originalRelease/amendmentRelease` , `previousVersion/latestVersion` etc...

```csharp
        var previousVersion = new Release
        {
            Id = Guid.NewGuid(),
            Publication = publication,
            Published = DateTime.UtcNow
        };

        var latestVersion = new Release
        {
            Id = Guid.NewGuid(),
            Publication = publication,
            PreviousVersionId = previousVersion.Id,
            ApprovalStatus = Draft
        };
```

In this type of setup, one version needs to be declared first so that the following version can reference the id to set `PreviousVersionId`, and both versions might be given identical years to differentiate them from versions belonging to other releases.

If this code was swapped to be generated with the test release data generator using `ReleaseGeneratorExtensions.DefaultRelease`, then out-of-the-box they would have different generated years which would be wrong, since year is normally always derived from the previous version. They would also all have `Version = 0` which is wrong.

With the new extension override in this PR this kind of code can be swapped out with the following to generate a release parent with any number of published release versions, followed optionally by a draft version.

```csharp
    ReleaseParent releaseParent = _dataFixture.DefaultReleaseParent(publishedVersions: 1, draftVersion: true);
```

* Published release versions will be approved by default, and have a default published date later than any preceding published version.
* All versions will automatically have `PreviousVersionId` set to their preceding version's id.
* All versions in the same release parent will have a consistent generated year.
* All versions in the same release parent will have a generated `Version` sequence.

Year can be overridden where we want to make the year declaration explicit, e.g.

```csharp
    ReleaseParent releaseParent = _dataFixture.DefaultReleaseParent(publishedVersions: 2, draftVersion: true, year: 2022)
```

### (3) Add new `WithReleaseParents` method to publication test data generator

Although `Publication` has no direct relationship with `ReleaseParents` (yet!), I've added this to the publication test data generator to allow declaring a publication with release parents and their release versions.

```csharp
    Publication publication = _dataFixture
        .DefaultPublication()
        .WithReleaseParents(_dataFixture
            .DefaultReleaseParent(publishedVersions: 1)
            .Generate(1));
```

This code will automatically set the bidirectional relationship between the publication and release versions.

Given a list of release parents it will also automatically set `Publication.LatestPublishedRelease` to be the latest published version taken from all versions in reverse chronological order of the releases.

This avoids another pattern we typically see a lot in the unit test code where a release is declared separately to a publication to set `LatestPublishedRelease`:

```csharp
    var releaseA = new Release
    {
        Id = Guid.NewGuid()
    };

    var publicationA = new()
    {
        Title = "Publication A",
        LatestPublishedRelease = releaseA
    };
```

### (4) Add unit tests for `ReleaseRepsository`

Before we start making any changes to `ReleaseRepsository` in EES-4667 I wanted to add unit tests for this class. After #4541 it's being relied on for anything related to working out latest release versions. It's already tested indirectly via other tests but they won't necessarily be covering all permutations such as no existing data and I won't be able to make the changes I want to without exhaustive tests.

The tests make use of the new test data generator methods in this PR.

I deliberately didn't test `ListLatestPublishedReleaseVersionIds()` which is marked for removal by EES-4336.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Extensions/FileExtensionTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Extensions/FileExtensionTests.cs
@@ -1,5 +1,4 @@
-﻿#nullable enable
-using System;
+﻿using System;
 using GovUk.Education.ExploreEducationStatistics.Common.Model;
 using GovUk.Education.ExploreEducationStatistics.Common.Utils;
 using Xunit;
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Extensions/MethodologyFileExtensionTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Extensions/MethodologyFileExtensionTests.cs
@@ -1,5 +1,4 @@
-﻿#nullable enable
-using System;
+﻿using System;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Extensions;
 using Xunit;
 using static GovUk.Education.ExploreEducationStatistics.Common.Model.FileType;
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Extensions/ReleaseExtensionTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Extensions/ReleaseExtensionTests.cs
@@ -1,5 +1,4 @@
-﻿#nullable enable
-using System;
+﻿using System;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Extensions;
 using Xunit;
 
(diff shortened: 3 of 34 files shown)


mmyoungman (approved):

mmyoungman:
Put Do not merge on this due to dev being merge into master sometime today.

mmyoungman:
...and removing it again. Ben pointed out these are just backend unit test changes so wouldn't influence a deploy.

PR #4581, opened 2024-02-15
EES-4714 Adding endpoint to retrieve a paginated list of data set versions for a particular data set

jack-hive:
Adding an endpoint for retrieving a paginated list of data set versions for a particular data set.

The endpoint lives at `api/v1/data-sets/{dataSetId}/versions`.

The endpoint returns the following status codes:
- 200
- 400
- 404

And returns the following JSON response when the response is a 200 OK:

```json
{
  "paging": {
    "page": 0,
    "pageSize": 0,
    "totalResults": 0,
    "totalPages": 0
  },
  "results": [
    {
      "number": "string",
      "type": "string",
      "status": "string",
      "published": "string(datetime)",
      "unpublished": "string(datetime)",
      "notes": "string",
      "totalResults": 0,
      "timePeriods": {
        "start": "string",
        "end": "string"
      },
      "geographicLevels": [
        "string"
      ],
      "filters": [
        "string"
      ],
      "indicators": [
        "string"
      ]
    }
  ]
}
```


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetsControllerTests.cs
```diff
@@ -113,4 +115,408 @@ private async Task<HttpResponseMessage> GetDataSet(Guid dataSetId)
...
+            Assert.Equal(2, content.Paging.Page);
+            Assert.Equal(5, content.Paging.PageSize);
+            Assert.Equal(12, content.Paging.TotalResults);
+            Assert.Equal(5, content.Results.Count);
+
+            Assert.Equal("3.0", content.Results[0].Number);
+            Assert.Equal("2.2", content.Results[1].Number);
+            Assert.Equal("2.1", content.Results[2].Number);
+            Assert.Equal("2.0", content.Results[3].Number);
+            Assert.Equal("1.2", content.Results[4].Number);
```
ntsim, line 229: Given we're generating multiple pages of versions, we might want to just check it orders correctly across all the pages as well. We could do this as successive requests for each page.

However, would consider bringing down the number of versions generated to shorten the test as 12 items is a bit overkill.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetsControllerTests.cs
```diff
@@ -113,4 +115,408 @@ private async Task<HttpResponseMessage> GetDataSet(Guid dataSetId)
...
+                    dataSetVersion.MetaSummary.TimePeriodRange.End.Year,
+                    dataSetVersion.MetaSummary.TimePeriodRange.End.Code),
+                result.TimePeriods.End);
+            Assert.Equal(dataSetVersion.MetaSummary.GeographicLevels, result.GeographicLevels);
+            Assert.Equal(dataSetVersion.MetaSummary.Filters, result.Filters);
+            Assert.Equal(dataSetVersion.MetaSummary.Indicators, result.Indicators);
+        }
+
+        [Fact]
+        public async Task DataSetVersionAvailableForOtherDataSet_Returns200_OnlyVersionForRequestedDataSet()
```
ntsim, line 303: Similar to other places, could be simpified to `AvailableVersionForOtherDataSet_Returns200_OnlyVersionForRequestedDataSet`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Utils/Extensions/DataSetVersionExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+    public static DataSetVersionType VersionType(this DataSetVersion dataSetVersion)
+    {
+        if (dataSetVersion.VersionMinor == 0)
+        {
+            return DataSetVersionType.Major;
+        }
+
+        return DataSetVersionType.Minor;
+    }
+}
```
ntsim, line 17: Wouldn't do this as an extension. It makes more sense being on the `DataSetVersion` model itself as a getter method so we can just call it like:

```cs
dataSetVersion.VersionType;
```

The `VersionType` may be useful in other situations later on, so it'd be better for it to be part of the domain model.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Utils/Extensions/DataSetVersionExtensionsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            DataSetVersion dataSetVersion = DataFixture
+                .DefaultDataSetVersion()
+                .WithVersionNumber(majorVersion, minorVersion);
+
+            var versionType = dataSetVersion.VersionType();
+
+            Assert.Equal(DataSetVersionType.Minor, versionType);
+        }
+    }
+}
```
ntsim, line 45: As mentioned in other comment, we don't need to do this as an extension so we should refactor this into a `DataSetVersionTests` class in `Public.Data.Model.Tests`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/ViewModels/DataSetViewModels.cs
```diff
@@ -111,3 +111,106 @@ public DataSetPaginatedListViewModel(List<DataSetViewModel> results, PagingViewM
     {
     }
 }
+
+public enum DataSetVersionType
+{
+    Major,
+    Minor
+}
```
ntsim, line 119: Similar to other comments, would move this into `Public.Data.Model` and make this part of the data model. It's feasible that this would be useful in other places so it makes sense to go there.

ntsim (changes requested):

ntsim:
Some minor things, but looking good otherwise :+1: 

Things are getting into a really good rhythm!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/DataSetVersionGeneratorExtensions.cs
```diff
@@ -170,9 +195,9 @@ public static InstanceSetters<DataSetVersion> SetStatusStaged(this InstanceSette
     public static InstanceSetters<DataSetVersion> SetStatusUnpublished(
         this InstanceSetters<DataSetVersion> instanceSetter)
         => instanceSetter
-            .SetStatus(DataSetVersionStatus.Published)
+            .SetStatus(DataSetVersionStatus.Unpublished)
             .SetUnpublished(DateTimeOffset.UtcNow)
-            .SetPublished(null);
+            .SetPublished(DateTimeOffset.UtcNow);
```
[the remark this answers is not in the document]
jack-hive: I thought we spoke about this and agreed that we'd keep the published date around when we unpublish? Hence why I populated it here (technically I should probably set it to be before the unpublished date).

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetsControllerTests.cs
jack-hive: The reason I had 12 items before was to make sure that the page I requested had some overlap into each major version. And I wanted to make sure each major version had at least 3 minor versions to make sure that things were being ordered correctly (could have been 2 I guess). Then I was just testing the middle page.

But I guess it would be more thorough to test each page. And if we do, then there is less need for 12 items. So I'll change it :)

jack-hive: I'm going to still make enough items so that some pages have major version overlaps though as I think that's somewhat nice to test

jack-hive: I've also removed the checks for the page properties as we're already testing this in the other pagination tests

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Utils/Extensions/DataSetVersionExtensions.cs
jack-hive: I was actually going to bring this up with you to come to an agreement.

The problem with this is that `DataSetVersionType` is in `GovUk.Education.ExploreEducationStatistics.Public.Data.Api.ViewModels` as it's part of the view model. So the `Models` project would need a circular reference back to the `Api` project in this case. Which is why I added it as an extension rather than on the model itself.

An alternative would be to put `DataSetVersionType` in `Models` which means I could have this method on the model without adding that circular dependency. So if we're happy with `DataSetVersionType` being a **domain model** rather than a **view model** then this is the route to go down.

What are your thoughts?

jack-hive: agreed to move to `Models` :)

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/DataSetVersionGeneratorExtensions.cs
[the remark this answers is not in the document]
ntsim: Yep, we should keep it, but if we've already called `SetPublished` before we call `SetStatusUnpublished` e.g.

```cs
.SetPublished(myCustomPublishedDate)
.SetStatusUnpublished()
```

The above will override `Published` and always set it to the `UtcNow`, which doesn't  seem correct as we'd expect `myCustomPublishedDate` to be set still.

Basically we need something to check that `Published` is already set or not before trying to default it to something. As you've said, also might want to make this date earlier than the `Unpublished` date

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetsControllerTests.cs
ntsim: Nice one, looks good! :+1:

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/DataSetVersionGeneratorExtensions.cs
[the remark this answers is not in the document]
jack-hive: I've implemented your suggestion :)

ntsim (approved):

PR #4585, opened 2024-02-20
EES-4770: Allow analysts to customise order of other releases in this series as shown on the publication page

tomjonesdev:
Publications listed EES releases chronologically, followed by legacy releases, where only the latter could be reordered manually. These changes allow both types of release to be combined and ordered in whichever way the publishers think make the most sense to EES users.

A `ReleaseOrder` property has been added to publications, which contains the list of associated `Release` and `LegacyRelease` items, along with their order when displayed in two locations:
1. the list of releases in the series on the publications page (public website), and
2. the Legacy Releases view (admin website), where they can be reordered.

This property also includes additional metadata to indicate the type of release, it's draft status, and whether or not it's an amendment of an existing release.

`ReleaseOrder` properties are created, updated, or deleted in the following scenarios:
- **create release**: created, and added to the top of the list, with a tag indicating that it's in draft status. The release can be reordered prior to publishing if required
- **create release amendment**: created, and added to the same position as the original release, with a tag indicating that it's an amendment which is in draft status. Draft amendments can be reordered, and the order value is carried over to the original release so that live releases can still be reordered whilst an amendment is still pending. This has the added benefit of retaining an assigned `Order` even if the amendment is subsequently cancelled
- **delete release (cancel amendment)**: deleted
- **publish release**: updated/deleted. If this is a new release, the `ReleaseOrder` draft status is updated. If the release is an amendment, the original `ReleaseOrder` is deleted, and the new one is updated to remove the draft amendment status
- **create legacy release**: created, and added to the top of the list
- **delete legacy release**: deleted
- **reorder combined releases**: updated, each `Order` position is reassigned based on the position in the list.

Release order values count upward from `1`, the higher the number, the higher up the list the release is shown. When a `ReleaseOrder` is deleted, these order values are reset to ensure consistent numbering. This reset happens as part of:
- delete release (cancel amendment)
- publish release (the original release's `ReleaseOrder` is deleted).
- delete legacy release

Tags on the releases screen indicate the following statuses:
- DRAFT: a release has been created, but not yet published
- DRAFT AMENDMENT: a release amendment has been created, but not yet published
- LATEST: this is the most recent published release

A migration endpoint has also been added for administrative users to trigger the population of `ReleaseOrder` columns on existing publications. Ordering for releases matches the existing behaviour of reverse chronological, with the most recent being at the top of the list, and is based on a release's year and time period. Legacy releases are appended, and the ordering will be matched with how they were prior to migration (highest first).

Draft releases and amendments will also have ordering applied by this migration. Drafts are placed at the top, in reverse chronological order if there are multiple, and draft amendments will match and "overlay" the order of their previous version.

Changes to a draft amendment's order will also update the original release's order. This allows analysts to change the position of a live release whilst an amendment is still unpublished. When the amendment is published, the order set originally will be carried over, so no further changes are required. If an amendment is cancelled, the current order is retained by the original.


src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/ContentDbContextModelSnapshot.cs
benoutram, line 1: Curious to find out if all the changes in this file are genuine or not as they don't look related to your changes.

Please could you run `dotnet ef --version` to see what version of .NET EF CLI tools you are running?

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
```diff
@@ -407,10 +407,10 @@ public virtual void ConfigureServices(IServiceCollection services)
 
             services.Configure<PreReleaseOptions>(Configuration);
             services.Configure<LocationsOptions>(Configuration.GetSection(LocationsOptions.Locations));
-            services.Configure<ReleaseApprovalOptions>(
-                Configuration.GetSection(ReleaseApprovalOptions.ReleaseApproval));
+            services.Configure<ReleaseApprovalOptions>(Configuration.GetSection(ReleaseApprovalOptions.ReleaseApproval));
```
benoutram, line 410: As a team I think we've been using line wrapping set up at 120 characters which is why this is formatted the way it is, so I think you should revert this change and set your line wrapping  at 120.

EES-4595 will enforce this standard when it comes in.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationReleaseOrderService.cs
```diff
@@ -0,0 +1,226 @@
...
+        {
+            var publication = await GetPublication(publicationId);
+
+            if (releaseId == Guid.Empty)
+            {
+                throw new ArgumentNullException(nameof(releaseId));
+            }
+
+            var legacyRelease = await _context.LegacyReleases.FindAsync(releaseId)
+                ?? throw new KeyNotFoundException($"No matching {nameof(LegacyRelease)} found with ID {releaseId}");
```
benoutram, line 53: I would replace this with

```csharp
var legacyRelease = await _context.LegacyReleases.FirstAsync(lr => lr.Id == legacyReleaseId);
```

It will throw an exception if it's not found which saves you having to handle it in this custom way.

It would be worth making this change elsewhere too as I can see this style is used quite a lot.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationReleaseOrderService.cs
```diff
@@ -0,0 +1,226 @@
...
+                ReleaseId = releaseId,
+                IsLegacy = true,
+                IsDraft = false,
+                Order = publication.ReleaseOrders.Count + 1 // Add to the top of the list rather than presume the intended position
+            });
+
+            _context.Update(publication);
+        }
+
+        public async Task DeleteForDeleteLegacyRelease(Guid publicationId, Guid releaseId)
```
benoutram, line 43: You don't really need publication Id here I don't think. Legacy releases can only belong to one publication. If you reorder getting the legacy release before the publication you can use `legacyRelease.PublicationId` to get the publication.

benoutram, line 43: I would rename `releaseId` here to `legacyReleaseId` to make it a bit clearer what this is since legacy releases are not the same thing as releases. Throughout the codebase `releaseId` refers to the unique identifier of a `Release`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationReleaseOrderService.cs
```diff
@@ -0,0 +1,226 @@
...
+        {
+            var publication = await GetPublication(publicationId);
+
+            if (releaseId == Guid.Empty)
+            {
+                throw new ArgumentNullException(nameof(releaseId));
+            }
```
benoutram, line 50: Add the `#nullable enable` pragma to this class and then it will be clear that the id argument is a non-nullable reference type.

I think it's overkill to check for a `null` or `Guid.Empty` value here and we don't do this level of argument checking elsewhere.

src/GovUk.Education.ExploreEducationStatistics.Admin/appsettings.json
```diff
@@ -45,5 +45,8 @@
         "LocalAuthority"
       ]
     }
+  },
+  "Environment": {
+    "BaseUrl": "https://explore-education-statistics.service.gov.uk"
```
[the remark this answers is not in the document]
tomjonesdev: Ah ok, didn't realise there was one already - the piece of code I was looking at at the time used a hardcoded domain so I assumed there wasn't a class configured yet.

Maybe these options are configured differently than I expected, but appsettings.json typically applies to production so it's expected to find a production URL here, I just didn't add similar values to appsettings.Development.json (etc.) just in case it wasn't wanted.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/ReleaseOrderMigrationController.cs
```diff
@@ -0,0 +1,92 @@
...
+
+                foreach (var publication in publications)
+                {
+                    await _publicationCacheService.UpdatePublication(publication.Slug);
+                }
+
+                return Ok($"Database updated successfully. {publicationsUpdated} publications updated, containing a total of\n- {eesReleasesOrdered} EES releases ordered, and\n- {legacyReleasesReordered} legacy releases reordered.");
+            }
+
+            return Ok($"Dry run completed. {publicationsUpdated} publications updated, containing a total of\n- {eesReleasesOrdered} EES releases ordered, and\n- {legacyReleasesReordered} legacy releases reordered.");
```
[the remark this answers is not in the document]
tomjonesdev: Yeah, this seemed simpler and more useful to the caller. The code will be deleted on a subsequent ticket anyway so I didn't want to spend too long fussing over the feedback method.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/LegacyReleasesControllerTests.cs
```diff
@@ -0,0 +1,161 @@
...
+using GovUk.Education.ExploreEducationStatistics.Common.Model;
+using GovUk.Education.ExploreEducationStatistics.Common.Tests.Extensions;
+using Microsoft.AspNetCore.Mvc;
+using Moq;
+using Xunit;
+using static GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils.MockUtils;
+using static Moq.MockBehavior;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Controllers.Api
+{
```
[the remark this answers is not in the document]
tomjonesdev: Can you give an example of what you mean by this?

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20240115170541_EES4770_AddReleaseOrderColumn.cs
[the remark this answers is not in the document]
tomjonesdev: There were a couple of migrations in the lead-up the final decision. Adding a `ReleaseOrder` column to the `Release` table was one of them, which was reversed in a subsequent migration.

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/ContentDbContextModelSnapshot.cs
tomjonesdev: 8.0.0. - I'll see if I can revert/condense the migration files into a single one representing the end result.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/LegacyReleasesControllerTests.cs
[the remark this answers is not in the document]
benoutram: Yep sure, there's an explanation of file-scoped namespaces here https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/file-scoped-namespaces.

In our codebase in most of the older code you will see a namespace declaration that looks a bit like this:

```csharp
// usings
using System;
...

namespace GovUk.Education.ExploreEducationStatistics.Project
{
    // types
    public class MyClass
    {
    }
}
```

But this has unnecessary indentation. Since c#10 you can now write this as:

```csharp
// usings
using System;
...

namespace GovUk.Education.ExploreEducationStatistics.Project;

public class MyClass
{
}
```

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationReleaseOrderService.cs
tomjonesdev: I opted for the more descriptive throws to help identify an exact cause if any errors arise which need to be investigated. These methods manage different entities and rely on several keys, so it might not be immediately obvious (from a log message) when a generic message is returned.

I can simplify to use the normal approach if preferred, but just thought I should point out the intention before I undo it.

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/ContentDbContextModelSnapshot.cs
benoutram: So I tried the following:

* Delete `EES4770_AddReleaseOrderColumn` cs files.
* Delete `EES4770_AddPublicationReleaseOrdersColumn` cs files
* Revert all changes from `ContentDbContextModelSnapshot.cs` so that they match the dev branch.

Generate a new 'EES4770_AddPublicationReleaseOrdersColumn' migration.

The only change I get to `ContentDbContextModelSnapshot.cs` is the following inserted on L829 which is all that I'd expect to change:

```csharp
b.Property<string>("ReleaseOrders")
    .IsRequired()
    .HasColumnType("nvarchar(max)");
```

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationReleaseOrderService.cs
benoutram: Cool, I see that, thanks. Maybe keep some of them in then where you are finding items within `ReleaseOrders` if that's going to help identify problems.

I don't see much of a need for it when getting resources from the database like Publication, Release, or in the example above, LegacyRelease. We don't do anything like that elsewhere in the codebase.

Personally I prefer to keep the code concise which is easier to read, and rely on tests for things like this. We can identify problems from line numbers in stack traces too.

src/GovUk.Education.ExploreEducationStatistics.Admin/appsettings.json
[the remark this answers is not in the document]
benoutram: `appsettings.json` is just the default settings file for keys not overridden elsewhere in configuration sources that take a higher precedence like environment variables and `appsettings.Development.json`.

We should probably have default or empty values for all configuration params  like `PublicAppUrl: ""`in `appsettings.json` so that it's obvious what the total set of configuration keys the application expects is.

`PublicAppUrl` is in `appsettings.Development.json` for our local environment and defined as an environment variable in the infra config for all Azure environments.

If you still need the base URL then I would defer to using `PublicAppUrl` or at least do some refactoring so we don't end up with two variables that are the same thing. For reference, [ConfigurationController.cs](https://github.com/dfe-analytical-services/explore-education-statistics/blob/88b29f9798dbfcc47b2063b4fbc148dd435dd22c/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ConfigurationController.cs) is using it and note that the were recent changes to it in https://github.com/dfe-analytical-services/explore-education-statistics/pull/4575.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/LegacyReleaseService.cs
```diff
@@ -56,16 +67,59 @@ public class LegacyReleaseService : ILegacyReleaseService
...
+                        });
+                    }
+
+                    foreach (var eesRelease in eesReleases)
+                    {
+                        combinedReleasesViewModels.Add(new()
+                        {
+                            Description = eesRelease.Title,
+                            Id = eesRelease.Id,
+                            Url = $"{_options.BaseUrl}/find-statistics/{publication.Slug}/{eesRelease.Slug}",
```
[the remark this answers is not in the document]
benoutram: The frontend has access to the environment base URL. Search for `.tsx` files containing `config.publicAppUrl` and you should see examples like 
```tsx
href={`${config.PublicAppUrl}/find-statistics/${publicationSlug}/${releaseSlug}`}
```

Personally I don't think you need the link for non-legacy releases and there could be other problems with those links.

-  First version releases that are not published will have a link that's not found.
- Unpublished amendments will have a link that goes to the latest published version.
 
On the 'Releases' tab you won't see links to the public releases either. The standard way to get the public link to a release is to go to the Sign off tab of that release where there's a link under the title `The public release will be accessible at:`...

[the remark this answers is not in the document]
tomjonesdev: I'll update this so that first version drafts don't show a URL, since it will be broken.

Draft amendments will retain the original release URL as this was the intended behaviour. The draft "overlays" the original release on the reordering page until it is published, so it makes sense for the URL to be for the original release.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationReleaseOrderService.cs
```diff
@@ -0,0 +1,226 @@
...
+                }
+
+                var releaseOrder = publication.ReleaseOrders.Find(ro => ro.ReleaseId == update.Id)
+                    ?? throw new KeyNotFoundException($"No matching ReleaseOrder found for {(update.IsLegacy ? nameof(LegacyRelease) : nameof(Release))} with ID {update.Id}");
+
+                releaseOrder.Order = update.Order;
+
+                _context.Update(publication);
+            }
+        }
```
benoutram, line 200: I think this is quite complex and I would have tried to avoid any attributes like `IsDraft` which are stateful from being in the ReleaseOrder's.

We should be able to find out from the release itself what state it is currently in rather than relying on `IsDraft`.

A 'draft' release in this service means unpublished whereas draft in release terminology is just one of the unpublished states that a release can be in so that's also slightly confusing.

Duplicating the state adds the possibility that it could become out of sync and we see that there's a change required to the Publisher function to maintain this state which has added to the complexity of the Publisher, as well as put a dependency on the Admin project. 

`IsDraft` is also not applicable to legacy releases so it's not so nice that it appears in every persisted entry.

`ReleaseOrder.IsLegacy` could be avoided by giving `ReleaseOrder` separate id attributes like `Guid? ReleaseId` and `Guid? LegacyReleaseId`. There can then be a computed property like `public bool IsLegacy => LegacyReleaseId.HasValue` if necessary.

Assuming that ReleaseOrder needs no other attributes and that null values get omitted when serialising to Json, I imagine it's possible to get this working with the persisted value looking something like this:

```
[{"ReleaseId":"76447fab-ca79-4918-acde-08dc3394e05a"},
 {"ReleaseId":"5f047886-4193-40c8-acdf-08dc3394e05a"},
 {"LegacyReleaseId":"02ea926b-c95a-4e80-81a3-8035adaea191"}]
```

compared to what you have now which looks like this

```
[{"ReleaseId":"76447fab-ca79-4918-acde-08dc3394e05a","IsLegacy":false,"IsDraft":true,"IsAmendment":false,"Order":1},
 {"ReleaseId":"5f047886-4193-40c8-acdf-08dc3394e05a","IsLegacy":false,"IsDraft":true,"IsAmendment":false,"Order":2},
 {"ReleaseId":"02ea926b-c95a-4e80-81a3-8035adaea191","IsLegacy":true,"IsDraft":false,"IsAmendment":false,"Order":3}]
```

To me, we shouldn't really care whether a release is draft or an amendment. Say we've got an ordered list of release id's for a publication. Then in the content API when producing the public view model we get the list of published releases, can't we simply order that list of published releases against the order they are observed in the Publication? It won't matter that the Publication's list of ordered releases has draft or amended release id's in it as they will be ignored.

To make this simpler, you could consider swapping the release id's for `ReleaseParent` id's.

All releases derived from the same initial version have the same `ReleaseParent`. It never changes when amending a release. In future I intend to rename `ReleaseParent -> Release` and `Release -> ReleaseVersion`.

If you use `ReleaseParent` as the id to track the order:

- You won't need to ever delete an entry because they will always exist once the first version is created. The first version can never be removed (for now) so a parent exists forever.
- You won't need to do anything on publishing to update draft state.
- You won't need to do create new entries in the amendment service when a new release version is created.

Any remaining logic could be done elsewhere and you might not need this service at all.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/ReleaseOrder.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        public Guid ReleaseId { get; set; }
+
+        public bool IsLegacy { get; set; }
+
+        public bool IsDraft { get; set; }
+
+        public bool IsAmendment { get; set; }
+
+        public int Order { get; set; }
```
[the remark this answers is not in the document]
tomjonesdev: Amendments use the same `Order` value as their original release, so ordering by position won't work for these situations.

src/explore-education-statistics-frontend/src/modules/find-statistics/PublicationReleasePage.tsx
```diff
@@ -334,21 +362,18 @@ const PublicationReleasePage: NextPage<Props> = ({ release }) => {
...
-                              <Link
-                                to={`/find-statistics/${release.publication.slug}/${slug}`}
-                              >
-                                {title}
-                              </Link>
-                            </li>
-                          ),
-                        ),
-                        ...release.publication.legacyReleases.map(
-                          ({ id, description, url }) => (
```
mmyoungman, line 349: There is code like this in Admin's `ReleaseContent.tsx` that also needs updating.

I think probably some legacy release stuff that can be removed from view models when this is done soon, like from `publicationService`'s `interface Publication`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/LegacyReleaseController.cs
```diff
@@ -29,6 +29,14 @@ public async Task<ActionResult<List<LegacyReleaseViewModel>>> ListLegacyReleases
                 .HandleFailuresOrOk();
         }
 
+        [HttpGet("publications/{publicationId:guid}/combined-releases")]
+        public async Task<ActionResult<List<CombinedReleaseViewModel>>> ListCombinedReleases(Guid publicationId)
```
mmyoungman, line 33: I think naming variables `CombinedReleases` is confusing, because it's not immediately clear what has been combined and to what purpose.

I'd be inclined to rename `CombinedReleases` to `ReleaseSeries` and things like `CombinedRelease` to `ReleaseSeriesItem`.

Again, I realise this is a pain to change, but I think the change will be worth it in the long run.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationReleaseOrderService.cs
mmyoungman: Ben said: `ReleaseOrder.IsLegacy could be avoided by giving ReleaseOrder separate id attributes like Guid? ReleaseId and Guid? LegacyReleaseId. There can then be a computed property like public bool IsLegacy => LegacyReleaseId.HasValue if necessary.`

This is a larger change and perhaps should be a different ticket, but I was thinking there is even a possibility to remove the `LegacyReleases` table entirely, since they're just a description and a URL, and so could be stored in the `ReleaseOrders` JSON - no ID is required, the order is implied by it's place in the JSON array, and it's in the `Publications` table so no need to keep a `PublicationId` either.

This would be a big simplification and remove a huge amount of code, because then practically all of `LegacyReleaseController` could be deleted.

I think it's definitely worth exploring how much effort this would be now, as I think it would be a big win. But also happy to push this to another ticket if it's too much.

mmyoungman (changes requested):

mmyoungman:
Just for the record, I'm now picking this

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Publication.cs
```diff
@@ -44,5 +44,7 @@ public class Publication
         public Guid? LatestPublishedReleaseId { get; set; }
 
         public Release? LatestPublishedRelease { get; set; }
+
+        public List<ReleaseOrder> ReleaseOrders { get; set; } = new();
```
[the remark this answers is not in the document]
mmyoungman: Done

src/explore-education-statistics-admin/src/services/legacyReleaseService.ts
```diff
@@ -24,6 +41,9 @@ const legacyReleaseService = {
   listLegacyReleases(publicationId: string): Promise<LegacyRelease[]> {
     return client.get(`/publications/${publicationId}/legacy-releases`);
   },
```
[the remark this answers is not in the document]
mmyoungman: Done

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/LegacyReleaseController.cs
mmyoungman: Done

mmyoungman:
This PR already has loads of comments and stuff on it - and I'm pushing changes and probably will get a load of comments when I've finished making my changes, so I'm going to close this and reopen a new PR once I've finished

PR #4586, opened 2024-02-20
EES-4743 Adding endpoint to retrieve a single Data Set Version

jack-hive:
Adding an endpoint for retrieving a single Data Set Version.

The endpoint lives at `api/v1/data-sets/{dataSetId}/versions/{version}`.

The endpoint returns the following status codes:
- 200
- 404

And returns the following JSON response when the response is a 200 OK:

```json
{
  "number": "string",
  "type": "string",
  "status": "string",
  "published": "string(datetime)",
  "unpublished": "string(datetime)",
  "notes": "string",
  "totalResults": 0,
  "timePeriods": {
    "start": "string",
    "end": "string"
  },
  "geographicLevels": [
    "string"
  ],
  "filters": [
    "string"
  ],
  "indicators": [
    "string"
  ]
}
```

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetsControllerTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetsControllerTests.cs
@@ -308,7 +308,7 @@ public abstract class DataSetsControllerTests : IntegrationTestFixture
                 .DefaultDataSet()
                 .WithStatusPublished();
 
-            await TestApp.AddTestData<PublicDataDbContext>(context => 
+            await TestApp.AddTestData<PublicDataDbContext>(context =>
                 context.DataSets.AddRange(dataSet1, dataSet2));
 
             DataSetVersion dataSet1Version = DataFixture
@@ -331,7 +331,7 @@ public abstract class DataSetsControllerTests : IntegrationTestFixture
                 .WithVersionNumber(2, 2)
                 .WithDataSetId(dataSet2.Id);
 
-            await TestApp.AddTestData<PublicDataDbContext>(context => 
+            await TestApp.AddTestData<PublicDataDbContext>(context =>
                 context.DataSetVersions.AddRange(dataSet1Version, dataSet2Version));
 
             var response = await ListVersions(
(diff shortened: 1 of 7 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetsControllerTests.cs
```diff
@@ -517,4 +517,171 @@ public async Task InvalidDataSetId_Returns404()
...
+                    indicators: 1,
+                    locations: 1,
+                    timePeriods: 3)
+                .WithStatus(dataSetVersionStatus)
+                .WithPublished(DateTimeOffset.UtcNow)
+                .WithDataSetId(dataSet.Id);
+
+            await TestApp.AddTestData<PublicDataDbContext>(context => context.DataSetVersions.Add(dataSetVersion));
+
+            var response = await GetVersion(dataSet.Id, dataSetVersion.Version);
```
ntsim, line 551: Not a necessity, but might be worth using named parameters. Similar could apply in any of the new tests too

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Controllers/DataSetsController.cs
```diff
@@ -38,6 +38,25 @@ public DataSetsController(IDataSetService dataSetService)
             .HandleFailuresOrOk();
     }
 
+    /// <summary>
+    /// Get a data set version
+    /// </summary>
+    /// <remarks>
+    /// Get a data set version, including a full list of its changes.
```
ntsim, line 45: Looks like this description is wrong. We don't get the full list of changes here anymore as we'll have a separate endpoint for this.

We could adjust this to:

```
Get a data set version's summary details.
```

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetService.cs
```diff
@@ -99,6 +107,23 @@ private static DataSetViewModel MapDataSet(DataSet dataSet)
...
+            .Where(dsv => dsv.DataSetId == dataSetId)
+            .Where(ds => ds.Status == DataSetVersionStatus.Published
+                || ds.Status == DataSetVersionStatus.Unpublished
+                || ds.Status == DataSetVersionStatus.Deprecated)
+            .ToListAsync();
+
+        return await dataSetVersions
+            .Where(dsv => dsv.Version == dataSetVersion)
+            .SingleOrDefault()
+            .OrNotFound();
```
ntsim, line 124: Not sure this is right. We're bringing back all the data set's versions into memory and checking them there, rather than just directly querying them in the database?

Think this can just be:

```cs
return await _publicDataDbContext.DataSetVersions
    .Where(dsv => dsv.DataSetId == dataSetId)
    .Where(ds => ds.Status == DataSetVersionStatus.Published
        || ds.Status == DataSetVersionStatus.Unpublished
        || ds.Status == DataSetVersionStatus.Deprecated)
    .Where(dsv => dsv.Version == dataSetVersion)
    .SingleOrNotFound();
```

Plus we get to use `SingleOrNotFound` as well

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Controllers/DataSetsController.cs
```diff
@@ -38,6 +38,25 @@ public DataSetsController(IDataSetService dataSetService)
...
+    /// </remarks>
+    [HttpGet("{dataSetId:guid}/versions/{dataSetVersion}")]
+    [Produces("application/json")]
+    [SwaggerResponse(200, "The requested data set version", type: typeof(DataSetVersionViewModel))]
+    [SwaggerResponse(404)]
+    public async Task<ActionResult<DataSetVersionViewModel>> GetVersion(
+        [SwaggerParameter("The ID of the data set.")] Guid dataSetId,
+        [SwaggerParameter("The data set version.")] string dataSetVersion)
```
ntsim, line 53: Might be worth throwing some little examples into the description e.g. `The data set version e.g. 1.0, 1.1, 2.0, etc`

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetService.cs
jack-hive: Can't do that as it can't convert `dsv.Version == dataSetVersion` into an expression translatable into SWL :/

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetsControllerTests.cs
jack-hive: I think we agreed to use them when we had 3 or more parameters (under any circumstances), and always use them when we have 2 or more parameters of the same type?

I'm happy to adjust this to use named parameters but just want to be consistent with the coding standards so we're consistent across the codebase

ntsim: Yeh I don't think it's a necessity. As you've said, it's not breaking any style guide rules. 

I'd personally have written them as named parameters just for better readability, but happy to leave it if you'd prefer it as-is!

ntsim (approved):

PR #4592, opened 2024-02-21
EES-4287 Update to React 18

amyb-hiveit:
Updates the admin, common and frontend packages to use React and React DOM 18.

When you checkout this branch you'll need to do an install and it's probably best to clean out your node modules first to ensure everything is using the correct version of React:
```
pnpm clean
pnpm i
```

### Other packages have been upgraded to be compatible:

- react-leaflet:
  - This had a major rewrite for version 3 which meant changing our implementation, see [this commit](https://github.com/dfe-analytical-services/explore-education-statistics/commit/ab115490666fd35ae44b0011d6ceadd7756d1c9b). 
  - In version 4 it has changed to use ES modules which means we can't test it with our current Jest set up. I've added a ticket, [EES-4902](https://dfedigital.atlassian.net/browse/EES-4902),  to look into this. 
  - The version of Leaflet has also been updated.
- react-beautiful-dnd: this isn't maintained any more so I've switched to use a fork, [@hello-pangea/dnd](https://github.com/hello-pangea/dnd),  that's been updated to work with React 18
- react-markdown: updated to v8, there is a v9 but that caused problems
- react-lazyload: this now targets a specific commit as it's been updated to work with React 18 but the version hasn't been published. We should look for a better maintained alternative to this in future.
- react-hook-form and react-query: I've pinned both of these to specific versions as there are type changes on minor updates that weren't a quick fix to deal with (I had to delete and rebuild my pnpm lock file at one point and found that these 'minor' updates broke things)

### Tests 
 @testing-library/react and its associated packages have been updated - the majority of the changes are due to this.
  - `userEvent` must now be awaited (I tried to only update to v13 to avoid polluting this PR with so many changes but it's very buggy so had to go to v14).
  - testing-library/react-hooks has been subsumed into testing-library/react but it hasn't been a complete port:
    - wrappers when testing context hooks don't get the `initialProps`, they have to be passed directly to the wrapper, which means we can't test changing `initialProps` on `rerender`
    - `renderHook` doesn't have `waitforNextUpdate` which causes problems testing intermediary states that change quickly such as `isLoading` props
  - Jest `useFakeTimers` doesn't play nicely with `userEvent` now. I found that we were using it unnecessarily in a few places when we could just wait for the DOM to change so have updated those tests. I eventually found out that it is possible to still use it if you use the new userEvent setup with one of the following options:
   `const user = userEvent.setup({ delay: null });`
   `const user = userEvent.setup({ advanceTimers: jest.advanceTimersByTime });`
   Note: `userEvent.setup` is the recommended way to handle user events now, but I haven't changed it in many places as the old way still works for now and this PR was big enough already. See https://testing-library.com/docs/user-event/intro#writing-tests-with-userevent

### UI tests

I've had to make some small adjustments to a few tests to get them to work - things like scrolling down a bit or changing what's clicked. I'm not really sure why this was necessary as nothing seems to have changed that would cause them to fail and I couldn't recreate any of the failures manually. 

I'm very open to better solutions here!

### Table tool weirdness
The Locations step on the table tool behaved oddly after the updated:
- some groups of checkboxes weren't clickable - splitting the component up so it's not trying to process the locations before they're fetched has fixed this
- the error behaviour was slightly different - I've added an `initialTouched` param to `RHFForm` which is used to show the error summary for untouched fields, this is needed for when you go back to the step and unselect all the options.


src/explore-education-statistics-common/src/modules/table-tool/components/FiltersForm.tsx
```diff
@@ -259,6 +259,7 @@ export default function FiltersForm({
                               ariaControls="filterGroups"
                               className="govuk-!-margin-bottom-2"
                               onClick={() => {
+                                console.log('CLICK', allFiltersOpen);
```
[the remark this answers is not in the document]
amyb-hiveit: Doh , yes :) It's gone now.

src/explore-education-statistics-admin/src/index.tsx
```diff
@@ -1,13 +1,16 @@
...
 import React from 'react';
-import ReactDOM from 'react-dom';
 import * as serviceWorker from './serviceWorker';
 
 configureAxios();
 
 import('./App').then(({ default: App }) => {
-  ReactDOM.render(<App />, document.getElementById('root'));
+  const container = document.getElementById('root');
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
sambiramairelogic, line 11: Purely a suggestion, not a blocking comment - but since we're adding a line here anyway to disable the es-lint rule, I wonder if a "if not container throw error" line would be preferable, for the sake of failing fast?

src/explore-education-statistics-admin/src/pages/release/data/components/__tests__/DataFileReplacementPlan.test.tsx
```diff
@@ -894,16 +898,16 @@ describe('DataReplacementPlan', () => {
...
+    await userEvent.click(screen.getByRole('button', { name: 'Confirm' }));
 
-    await waitFor(() => {
+    await waitFor(() =>
       expect(
         screen.queryByText(/Are you sure you want to delete/),
-      ).toBeInTheDocument();
-    });
+      ).not.toBeInTheDocument(),
+    );
```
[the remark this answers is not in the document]
sambiramairelogic: (Just tried this locally and it still passes for me without the waitFor)

src/explore-education-statistics-common/src/modules/charts/components/utils/__tests__/generateFeaturesAndDataGroups.test.ts
```diff
@@ -0,0 +1,499 @@
...
+        min: '£12,833',
+        max: '£16,388',
+        minRaw: 12833,
+        maxRaw: 16388,
+      },
+    ];
+
+    expect(result.dataGroups).toEqual(expectedDataGroups);
+
+    // expect(legendColours[0].style.backgroundColor).toBe(
```
sambiramairelogic, line 485: Were these left in by mistake? :)

src/explore-education-statistics-common/src/modules/charts/components/hooks/useLegend.ts
```diff
@@ -1,20 +1,20 @@
...
-    nextProps => {
+    (nextProps: DefaultLegendContentProps) => {
       const nextLegendProps = omit(nextProps, 'content');
 
       // Need to do a deep comparison of the props to
       // avoid falling into an infinite rendering loop.
       if (JSON.stringify(legendProps) !== JSON.stringify(nextLegendProps)) {
         setTimeout(() => {
-          setLegendProps(nextLegendProps);
+          setLegendProps(nextLegendProps as LegendProps);
```
sambiramairelogic, line 17: Again not a blocking comment, just a personal preference:

I try to avoid `as` casts wherever possible, and would personally opt for changing the above lines to: 
`      const nextLegendProps: LegendProps = {
        ...omit(nextProps, 'content'),
        width: nextProps.width ? Number(nextProps.width) : undefined,
        height: nextProps.height ? Number(nextProps.height) : undefined,
      };`
  But I appreciate the probability of height/weight containing non-numeric strings is pretty low :D

src/explore-education-statistics-admin/src/pages/release/pre-release/components/__tests__/PreReleaseUserAccessForm.test.tsx
```diff
@@ -180,11 +182,17 @@ describe('PreReleaseUserAccessForm', () => {
...
-      await userEvent.type(
+      await user.type(
         screen.getByLabelText('Invite new users by email'),
-        'test@test.com{enter}invalid-1{enter}invalid-2',
+        'test@test.com{Enter}invalid-1{Enter}invalid-2',
       );
-      userEvent.tab();
+      await user.tab();
+
+      // await userEvent.type(
```
sambiramairelogic, line 191: Can this commented block go now?

src/explore-education-statistics-common/src/components/Details.tsx
```diff
@@ -100,6 +100,7 @@ const Details = ({
         aria-expanded={onMounted(open)}
         className="govuk-details__summary"
         role={onMounted('button')}
+        // role="button"
```
sambiramairelogic, line 103: Is this left in deliberately? 

(Also just for my own curiousity - why don't we want this to have the button role until it's fully rendered?)

src/explore-education-statistics-common/src/components/__tests__/Button.test.tsx
```diff
@@ -97,46 +78,30 @@ describe('Button', () => {
...
-    jest.useRealTimers();
   });
 
   test('enabled once the current `onClick` handler has finished', async () => {
-    jest.useFakeTimers();
-
-    const handleClick = jest.fn(
-      () => new Promise(resolve => setTimeout(resolve, 2000)),
-    );
+    const handleClick = jest.fn(async () => delay(100));
```
sambiramairelogic, line 91: Can't help but notice that this method in the file below is `... => await delay(100));` whereas the same method appears here without awaiting the delay. Presumably delay should be always be awaited?

src/explore-education-statistics-common/src/components/__tests__/ModalConfirm.test.tsx
```diff
@@ -27,10 +25,16 @@ describe('ModalConfirm', () => {
...
       );
 
-      userEvent.click(screen.getByRole('button', { name: 'Confirm' }));
+      expect(
+        await screen.findByRole('button', { name: 'Confirm' }),
+      ).toBeInTheDocument();
 
+      await userEvent.click(screen.getByRole('button', { name: 'Confirm' }));
+
+      // await waitFor(() => {
```
sambiramairelogic, line 34: Can these commented lines go now?

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/PublicationForm.test.tsx
```diff
@@ -389,28 +395,26 @@ describe('PublicationForm', () => {
...
       <PublicationForm {...wizardProps} themes={testThemes} onSubmit={noop} />,
     );
 
-    await userEvent.type(screen.getByLabelText('Search publications'), '[');
+    await userEvent.type(screen.getByLabelText('Search publications'), '[[');
 
-    expect(() => {
-      jest.runOnlyPendingTimers();
-    }).not.toThrow();
+    expect(() => {}).not.toThrow();
```
sambiramairelogic, line 417: Does this line serve a purpose?

src/explore-education-statistics-admin/src/index.tsx
amyb-hiveit: Good idea, I've done that now.

src/explore-education-statistics-common/src/modules/charts/components/utils/__tests__/generateFeaturesAndDataGroups.test.ts
amyb-hiveit: Yep, removed now.

src/explore-education-statistics-common/src/modules/charts/components/hooks/useLegend.ts
amyb-hiveit: Yeah that seems like a better solution, I've changed it as you suggest.

src/explore-education-statistics-admin/src/pages/release/content/components/__tests__/ReleaseEditableBlock.test.tsx
```diff
@@ -614,7 +614,7 @@ describe('ReleaseEditableBlock', () => {
     ).not.toBeInTheDocument();
   });
 
-  test('re-renders locked state when current user interacts with block before expiry', async () => {
+  test.skip('re-renders locked state when current user interacts with block before expiry', async () => {
```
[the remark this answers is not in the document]
amyb-hiveit: We should add comments with a ticket reference, which I've done now for all the ones I had to skip :)

src/explore-education-statistics-admin/src/pages/release/pre-release/components/__tests__/PreReleaseUserAccessForm.test.tsx
amyb-hiveit: yep!

src/explore-education-statistics-common/src/components/Details.tsx
amyb-hiveit: No, it was left over from when I was wrangling with v13 of react testing library which had loads of problems clicking buttons!  I ended up upgrading to v14 instead.

I think it only gets the role on mount so that it works as a standard html details element when javascript is disabled. But I could be wrong, this component has been here longer than me.

src/explore-education-statistics-common/src/components/__tests__/Button.test.tsx
amyb-hiveit: That's just a style difference, if you're returning it directly you don't need to await it and will get an eslint error if you do - `Redundant use of `await` on a return value`. I've updated it in `ButtonText.test.tsx` to be like it is here as it's more concise.

src/explore-education-statistics-common/src/components/__tests__/ModalConfirm.test.tsx
amyb-hiveit: yep

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/PublicationForm.test.tsx
amyb-hiveit: Good spot :) I think I updated this test before I finally found out how to get Jest fake timers to work with userEvent. I've updated it now so it's like the previous version of the test.

src/explore-education-statistics-frontend/src/modules/data-catalogue/__tests__/DataCataloguePage.test.tsx
```diff
@@ -1584,15 +1600,17 @@ describe('DataCataloguePage', () => {
...
             name: 'Remove filter: Release Release title 1',
           }),
         );
 
-        await waitFor(() => {
-          expect(screen.getByText('30 data sets')).toBeInTheDocument();
-        });
+        // await waitFor(() => {
+        //   expect(screen.getByText('30 data sets')).toBeInTheDocument();
+        // });
```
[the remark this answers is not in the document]
amyb-hiveit: yep, gone!

PR #4595, opened 2024-02-22
EES-4860 Updates public API database schema to

ntsim:
This PR updates the public API's database schema (`PublicDataDbContext`) by:

- Updating the location meta model with specialised sub-types
- Replacing the denormalised JSONB columns used for the data set meta with standard SQL tables

## Location meta model changes

Previously, the location meta model was too simplistic and didn't consider different types of locations correctly. Not all location types have a single 'universal' code that can be used and this needed to be accounted for.

For example:

- Schools have URN and LAESTAB codes (and multiple others)
- Providers have UKPRN codes (and multiple others)
- Local authorities have 'old' codes (mostly an internal DfE code) and 'new' codes (used more widely by other departments e.g. ONS)
- RSC regions don't have any code at all!

We've now modelled this correctly by including these details in the location meta model. This comes in the form of subtypes of `LocationOptionMeta` e.g. `LocationLocalAuthorityOptionMeta`.

This has been implemented so that `LocationOptionMeta` still has references to all the columns used by its sub-types, allowing us to configure a unique index across all the table's main columns.

The unique index incurs additional storage cost (essentially doubling the table size), but allows us to avoid duplicates in the table (see later).

## Normalised meta tables

After some initial performance testing, it was clear the JSONB approach didn't perform as well as originally expected. This is particularly problematic when trying to SELECT options, where query times could range from 40 - 70 ms. 

When SELECTing, we'd typically have to convert the JSONB array into table rows so they can be returned by the query. This is imprecise and means Postgres has to potentially read and transform megabytes of JSONB (e.g. schools, qualifications, etc).

The normalised table approach is much more performant as the meta is actually very tabular in nature. This means we can have efficient indices on the columns, making SELECTs very fast (e.g. going from ~50ms to ~0.25ms).

Additionally, we have implemented many-to-many link tables with immutable option tables that should significantly reduce the storage overhead. This will most likely save space compared to JSONB as we won't store nearly as much duplicated / redundant data.

## Test data seeder

Alongside the database schema changes, we've had to re-work larger parts of the test data seeder. There were some challenges in:

- Inserting new filter and locations options into respective option tables without duplicating 
- Creating the subsequent links back to the respective meta tables

We have now accomplished this by using the MERGE command to add new the new filter and location options without duplication. This only inserts new options to the table by checking across all the columns as it inserts.

The options are then linked to their respective meta tables by using a 'row key' technique which allows us to quickly find the required options that we want to use. 

Other approaches were tested, but typically created the links much more slowly due to the use of complicated WHERE clauses checking every column.


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/LocationMeta.cs
```diff
@@ -1,10 +1,52 @@
...
 {
-    public required GeographicLevel Level { get; set; }
+    public int Id { get; set; }
 
-    public required List<LocationOptionMeta> Options { get; set; } = [];
+    public required Guid DataSetVersionId { get; set; }
+
+    public DataSetVersion DataSetVersion { get; set; } = null!;
+
+    public GeographicLevel Level { get; set; }
```
jack-hive, line 17: should this not be `required`?

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/FilterMetaGeneratorExtensions.cs
```diff
@@ -34,14 +37,26 @@ public static Generator<FilterMeta> WithHint(this Generator<FilterMeta> generato
...
 
+    public static Generator<FilterMeta> WithOptionLinks(
+        this Generator<FilterMeta> generator,
+        Func<IEnumerable<FilterOptionMetaLink>> links)
+        => generator.ForInstance(s => s.SetOptionLinks(links));
+
     public static InstanceSetters<FilterMeta> SetDefaults(this InstanceSetters<FilterMeta> setters)
         => setters
-            .SetDefault(m => m.Identifier)
+            .SetDefault(m => m.PublicId)
```
jack-hive, line 47: Do we need to set the `Id` too?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/FilterOptionMetaGeneratorExtensions.cs
```diff
@@ -26,21 +20,9 @@ public static Generator<FilterOptionMeta> WithLabel(this Generator<FilterOptionM
 
     public static InstanceSetters<FilterOptionMeta> SetDefaults(this InstanceSetters<FilterOptionMeta> setters)
         => setters
-            .SetDefault(m => m.PublicId)
-            .SetDefault(m => m.PrivateId)
             .SetDefault(m => m.Label)
```
jack-hive, line 23: Need to set `Id` I think

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/IndicatorMetaGeneratorExtensions.cs
```diff
@@ -25,11 +27,12 @@ public static Generator<IndicatorMeta> WithUnit(this Generator<IndicatorMeta> ge
         byte? decimalPlaces)
         => generator.ForInstance(s => s.SetDecimalPlaces(decimalPlaces));
 
+
     public static InstanceSetters<IndicatorMeta> SetDefaults(this InstanceSetters<IndicatorMeta> setters)
         => setters
-            .SetDefault(m => m.Identifier)
             .SetDefault(m => m.Label)
```
jack-hive, line 33: Maybe need to set Id too

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/FilterMetaGeneratorExtensions.cs
jack-hive: And `Created`/`Updated`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/IndicatorMetaGeneratorExtensions.cs
```diff
@@ -25,11 +27,12 @@ public static Generator<IndicatorMeta> WithUnit(this Generator<IndicatorMeta> ge
...
             .SetDefault(m => m.Unit)
+            .SetDefault(m => m.PublicId)
```
jack-hive, line 35: And `Created`/`Updated`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/LocationCodedOptionMetaGeneratorExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+
+    public static Generator<LocationCodedOptionMeta> WithCode(
+        this Generator<LocationCodedOptionMeta> generator,
+        string code)
+        => generator.ForInstance(s => s.SetCode(code));
+
+    public static InstanceSetters<LocationCodedOptionMeta> SetDefaults(
+        this InstanceSetters<LocationCodedOptionMeta> setters)
+        => setters
+            .SetDefault(m => m.Label)
```
jack-hive, line 21: `Id`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/LocationLocalAuthorityOptionMetaGeneratorExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+
+    public static Generator<LocationLocalAuthorityOptionMeta> WithOldCode(
+        this Generator<LocationLocalAuthorityOptionMeta> generator,
+        string oldCode)
+        => generator.ForInstance(s => s.SetOldCode(oldCode));
+
+    public static InstanceSetters<LocationLocalAuthorityOptionMeta> SetDefaults(
+        this InstanceSetters<LocationLocalAuthorityOptionMeta> setters)
+        => setters
+            .SetDefault(m => m.Label)
```
jack-hive, line 28: `Id`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/LocationMetaGeneratorExtensions.cs
```diff
@@ -30,14 +45,25 @@ public static Generator<LocationMeta> WithLevel(this Generator<LocationMeta> gen
...
+        Func<IEnumerable<LocationOptionMetaLink>> links)
+        => generator.ForInstance(s => s.SetOptionLinks(links));
+
     public static InstanceSetters<LocationMeta> SetDefaults(this InstanceSetters<LocationMeta> setters)
         => setters
             .Set(
                 m => m.Level,
-                (_, _, context) =>
-                    GeographicLevelUtils.Levels[context.Index % GeographicLevelUtils.Levels.Length]
+                (_, _, context) => GeographicLevelUtils.Levels[context.Index % GeographicLevelUtils.Levels.Length]
```
jack-hive, line 57: `Id` And `Created`/`Updated`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/LocationProviderOptionMetaGeneratorExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+
+    public static Generator<LocationProviderOptionMeta> WithUkprn(
+        this Generator<LocationProviderOptionMeta> generator,
+        string ukprn)
+        => generator.ForInstance(s => s.SetUkprn(ukprn));
+
+    public static InstanceSetters<LocationProviderOptionMeta> SetDefaults(
+        this InstanceSetters<LocationProviderOptionMeta> setters)
+        => setters
+            .SetDefault(m => m.Label)
```
jack-hive, line 23: `Id`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/LocationSchoolOptionMetaGeneratorExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+        => generator.ForInstance(s => s.SetUrn(ukprn));
+
+    public static Generator<LocationSchoolOptionMeta> WithLaEstab(
+        this Generator<LocationSchoolOptionMeta> generator,
+        string laEstab)
+        => generator.ForInstance(s => s.SetLaEstab(laEstab));
+    public static InstanceSetters<LocationSchoolOptionMeta> SetDefaults(
+        this InstanceSetters<LocationSchoolOptionMeta> setters)
+        => setters
+            .SetDefault(m => m.Label)
```
jack-hive, line 27: `Id`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/LocationSchoolOptionMetaGeneratorExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+            .SetDefault(m => m.Urn)
+            .SetDefault(m => m.LaEstab);
+
+    public static InstanceSetters<LocationSchoolOptionMeta> SetUrn(
+        this InstanceSetters<LocationSchoolOptionMeta> setters,
+        string ukprn)
```
jack-hive, line 33: rename to `urn`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/LocationSchoolOptionMetaGeneratorExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+        this DataFixture fixture)
+        => fixture.Generator<LocationSchoolOptionMeta>().WithDefaults();
+
+    public static Generator<LocationSchoolOptionMeta> WithDefaults(
+        this Generator<LocationSchoolOptionMeta> generator)
+        => generator.ForInstance(s => s.SetDefaults());
+
+    public static Generator<LocationSchoolOptionMeta> WithUkprn(
+        this Generator<LocationSchoolOptionMeta> generator,
+        string ukprn)
```
jack-hive, line 17: `urn`

And method should be `WithUrn`

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/LocationMeta.cs
```diff
@@ -1,10 +1,52 @@
...
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
 
 namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Model;
 
-public class LocationMeta
+public class LocationMeta : ICreatedUpdatedTimestamps<DateTimeOffset, DateTimeOffset?>
 {
-    public required GeographicLevel Level { get; set; }
+    public int Id { get; set; }
```
jack-hive, line 11: I think we said we'd stick to using `Guids`?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/LocationOptionMeta.cs
```diff
@@ -1,12 +1,190 @@
...
+
 namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Model;
 
-public class LocationOptionMeta
+public abstract class LocationOptionMeta
 {
-    public required string PublicId { get; set; }
-
-    public required int PrivateId { get; set; }
+    public int Id { get; set; }
```
jack-hive, line 9: `Guid`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/IndicatorMeta.cs
```diff
@@ -1,14 +1,41 @@
...
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
 
 namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Model;
 
-public class IndicatorMeta
+public class IndicatorMeta : ICreatedUpdatedTimestamps<DateTimeOffset, DateTimeOffset?>
 {
-    public required string Identifier { get; set; }
+    public int Id { get; set; }
```
jack-hive, line 10: `Guid`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/IndicatorMeta.cs
```diff
@@ -1,14 +1,41 @@
...
+
+    public required Guid DataSetVersionId { get; set; }
+
+    public DataSetVersion DataSetVersion { get; set; } = null!;
+
+    public required string PublicId { get; set; }
```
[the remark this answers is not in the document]
jack-hive: If it isn't unique, then should we create a unique composite index on the `PublicId`/`DataSetVersionId` pair?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/FilterOptionMetaLink.cs
```diff
@@ -0,0 +1,28 @@
...
+    public FilterOptionMeta Option { get; set; } = null!;
+
+    internal class Config : IEntityTypeConfiguration<FilterOptionMetaLink>
+    {
+        public void Configure(EntityTypeBuilder<FilterOptionMetaLink> builder)
+        {
+            builder.Property(l => l.PublicId)
+                .ValueGeneratedOnAdd();
+
+            builder.HasIndex(l => l.PublicId);
```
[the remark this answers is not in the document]
jack-hive: On the other hand, if `PublicId` is meant to be unique, then I don't think we need this composite key I mentioned.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/FilterOptionMetaLink.cs
```diff
@@ -0,0 +1,28 @@
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Model;
+
+public class FilterOptionMetaLink
+{
+    public int PublicId { get; set; }
```
jack-hive, line 8: `Guid`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/FilterOptionMeta.cs
```diff
@@ -2,11 +2,13 @@ namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Model;
 
 public class FilterOptionMeta
 {
-    public required string PublicId { get; set; }
-
-    public required int PrivateId { get; set; }
+    public int Id { get; set; }
```
jack-hive, line 5: `Guid`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/FilterMeta.cs
```diff
@@ -1,12 +1,53 @@
...
+    public DateTimeOffset? Updated { get; set; }
+
+    internal class Config : IEntityTypeConfiguration<FilterMeta>
+    {
+        public void Configure(EntityTypeBuilder<FilterMeta> builder)
+        {
+            builder.Property(m => m.PublicId)
+                .HasMaxLength(100);
+
+            builder.HasIndex(m => new { m.DataSetVersionId, m.PublicId })
```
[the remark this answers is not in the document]
jack-hive: Coming back to this...

I think I've figured out that `PublicIds` are unique when it comes to the options, but are not unique when it comes to the base filter. **(Please confirm/deny)**

But leaving this thread here just in case.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/FilterMeta.cs
```diff
@@ -1,12 +1,53 @@
...
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+
 namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Model;
 
-public class FilterMeta
+public class FilterMeta : ICreatedUpdatedTimestamps<DateTimeOffset, DateTimeOffset?>
 {
-    public required string Identifier { get; set; }
+    public int Id { get; set; }
```
jack-hive, line 9: `Guid`

jack-hive (changes requested):

jack-hive (changes requested):
I've taken a look through everything, but haven't quite 100% got my head around it all.

I plan on going back over the `SeedDataCommand` properly on Monday - I spent a fair bit of time on it but couldn't quite convince myself that I understood it completely. So I want to spend a bit more time on it. Either going through it by myself, or with you.

Overall looks really good! I just have a few outstanding things I wanted to clarify with you when you're back. It might be good to just pair to finish off any questions I have if you're OK with that?
- I want to clarify which types of `PublicId` are unique, and which are re-used (I think I know the answer know but want to be sure)
- I don't understand the intricacies of why we need the `linq2db` library. I saw one of your comments explaining why we need it but I'm still not completely clear on it. I'm thinking maybe one of the main reasons we need it is because EF doesn't provide a similar `merge` functionality which we need for the seeder (and importer when that comes along)? I'd like to confirm this with you though.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
```diff
@@ -79,8 +79,8 @@ public void ConfigureServices(IServiceCollection services)
         // cause the data source builder to throw a host exception.
         if (!_hostEnvironment.IsIntegrationTest())
         {
-            var dataSourceBuilder = new NpgsqlDataSourceBuilder(_configuration.GetConnectionString("PublicDataDb"));
-            dataSourceBuilder.MapEnum<GeographicLevel>();
```
jack-hive, line 83: After removing this line, I'm now seeing this again in the DB. Are we sure we don't need this line? 
image

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/FilterOptionMeta.cs
```diff
@@ -2,11 +2,13 @@ namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Model;
...
 
     public required string Label { get; set; }
 
-    public bool? IsAggregate { get; set; } = null!;
+    public bool? IsAggregate { get; set; }
+
+    public List<FilterMeta> Metas { get; set; } = [];
+
+    public List<FilterOptionMetaLink> MetaLinks { get; set; } = [];
```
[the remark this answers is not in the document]
jack-hive: **Discussed:**

We actually DO WANT to be able to have the same option used across a version. For example, the 'Total' option.

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/GeographicLevelUtils.cs
```diff
@@ -17,128 +17,120 @@ public static class GeographicLevelUtils
...
+                    Codes: new[] { "region_code" },
                     Name: "region_name"
                 )
             },
             {
                 GeographicLevel.RscRegion,
                 new GeographicCsvColumns(
-                    Code: "rsc_region_code",
-                    Name: "rsc_region_name"
+                    Codes: Array.Empty<string>(),
```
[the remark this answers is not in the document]
ntsim: Think this code was originally ported over from the prototype, but it's incorrect and think it was based on some older code in the project that doesn't exist anymore. 

Clarified with Cam and `rsc_region_lead_name` is the only column relevant to RSC Regions.

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/GeographicLevelUtils.cs
```diff
@@ -17,128 +17,120 @@ public static class GeographicLevelUtils
...
+                    Name: "rsc_region_lead_name"
```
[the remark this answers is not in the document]
ntsim: Yep, as above, the only column relevant to RSC Regions is `rsc_region_lead_name`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/FilterMeta.cs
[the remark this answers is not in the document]
jack-hive: **Discussed:**

`PublicId` is NOT UNIQUE across filters/versions for the `FilterMeta`, nor is it unique for the options either.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/FilterOptionMetaLink.cs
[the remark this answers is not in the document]
jack-hive: **Edit:**

Looks like this composite key is being set up already with other config.

So I think this comment can be resolved

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/FilterMeta.cs
```diff
@@ -1,12 +1,53 @@
...
+
+    public required Guid DataSetVersionId { get; set; }
+
+    public DataSetVersion DataSetVersion { get; set; } = null!;
+
+    public required string PublicId { get; set; }
```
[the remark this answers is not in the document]
jack-hive: **Discussed:**

Not unique

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Scripts/Commands/SeedDataCommand.cs
```diff
@@ -281,144 +287,321 @@ public async Task Generate()
...
+                                o.OldCode + ',' +
+                                o.Urn + ',' +
+                                o.LaEstab + ',' +
+                                o.Ukprn
+                        )
+                        .ToHashSet();
+
+                    var links = await optionTable
+                        .Where(
+                            o => o.Type == batch[0].Type &&
```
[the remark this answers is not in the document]
jack-hive: **Discussed:**

each batch will have the same `Type`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/LocationMeta.cs
jack-hive: **Discussed:**

Only using `Guids` for publically available IDs. Using `int` for strictly internally available IDs as it is more performant.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/IndicatorMeta.cs
jack-hive: **Discussed:**

Only using `Guids` for publicly available IDs. Using `int` for strictly internally available IDs as it is more performant.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/LocationOptionMeta.cs
jack-hive: **Discussed:**

Only using `Guids` for publicly available IDs. Using `int` for strictly internally available IDs as it is more performant.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/FilterOptionMeta.cs
jack-hive: **Discussed:**

Only using `Guids` for publicly available IDs. Using `int` for strictly internally available IDs as it is more performant.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/FilterMeta.cs
jack-hive: **Discussed:**

Only using `Guids` for publicly available IDs. Using `int` for strictly internally available IDs as it is more performant.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/FilterOptionMetaLink.cs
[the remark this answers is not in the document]
jack-hive: **Discussed:**

`PublicId` is not unique. We want to ability to be able to insert a duplicate value, but also let it append automatically on adding a new option

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/LocationOptionMeta.cs
```diff
@@ -1,12 +1,190 @@
...
+            builder.HasIndex(o => o.LaEstab);
+            builder.HasIndex(o => o.Ukprn);
+
+            // This index is primarily for providing a uniqueness constraint
+            // preventing duplicates making it into the table by accident.
+            // Unfortunately, it comes at the cost of essentially doubling the table's
+            // size, but it's worth having for now (can remove it later if needed).
+            builder.HasIndex(o => new {
+                    o.Type,
+                    o.Label,
```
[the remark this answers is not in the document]
ntsim: Good catch! The ones below don't need to be renamed at all. It's actually just a hangover from a previous attempt at modelling this.

I've changed it to just use the standard property names

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetVersion.cs
```diff
@@ -25,11 +25,17 @@ public class DataSetVersion : ICreatedUpdatedTimestamps<DateTimeOffset, DateTime
 
     public required string Notes { get; set; }
 
-    public required DataSetVersionMetaSummary MetaSummary { get; set; }
+    public long TotalResults { get; set; }
 
-    public DataSetMeta Meta { get; set; } = null!;
+    public DataSetVersionMetaSummary? MetaSummary { get; set; }
```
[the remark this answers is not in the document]
ntsim: Think I changed this because of some problems I was having with the generators, but you're right that this can still be required. Reverted!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/LocationOptionMeta.cs
```diff
@@ -1,12 +1,190 @@
...
+                .Property(o => o.Urn)
+                .HasColumnName("Urn");
+
+            builder
+                .Property(o => o.LaEstab)
+                .HasColumnName("LaEstab");
+
+            builder
+                .Property(o => o.Ukprn)
+                .HasColumnName("Ukprn");
```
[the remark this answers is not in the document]
jack-hive: I'm guessing you removed them, but then had to re-add them back in with the `IsRequired` because of the empty string => null change you made?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/TimePeriodMeta.cs
```diff
@@ -1,10 +1,36 @@
...
+
+    public DateTimeOffset? Updated { get; set; }
+
+    internal class Config : IEntityTypeConfiguration<TimePeriodMeta>
+    {
+        public void Configure(EntityTypeBuilder<TimePeriodMeta> builder)
+        {
+            builder.Property(o => o.Code)
+                .HasConversion(new EnumToEnumValueConverter<TimeIdentifier>())
+                .HasMaxLength(10);
```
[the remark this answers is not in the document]
jack-hive: Yeah sounds like a good idea!

Only reason I originally suggested using the same length as for `GeographicLevel` is because I thought the length we chose was arbitrary.

But sounds like it has a decent potential of climbing to longer lengths so I'm happy with 10 :)

jack-hive (approved):
Looks good! :D

PR #4598, opened 2024-02-23
EES-4814 - fixing Robot test run issues around logout under Entra ID.  Making logout under Keycloak more consistent.

duncan-at-hiveit:
:warning: Keycloak Docker containers and images need to be dropped and recreated when this PR goes in.

:warning: All App Registrations for Admin SPAs across each environment will need to be configured to return the optional "login_hint" Claim to enable silent logout.

# Overview

This PR aims to fix issues that some team members are experiencing around inconsistent logout behaviour when running under Entra ID.

It also brings Keycloak logout behaviour more in line with Entra ID's logout by supplying a logout endpoint in the authority metadata so that a full logout redirect flow can be supported which results in the user being logged out of Keycloak as well as the app, thus ensuring they log in again instead of just hitting "Sign in" and being magically signed in again!

I've added a new login and logout Robot test as well, for testing various logout cases.

# Other changes

## Silent logout

I've configured the App Registration for our local development Admin SPA to support silent logout so that we don't have to select an account when logging out.  By adding the "login_hint" optional Claim to the ID token that Entra ID returns, this will allow Entra ID to know who is logging out.

## Potential fix to ClaimPrincipalTransformationTests

I've added a "ResetUsersAndRolesDbContext()" to ClaimPrincipalTransformationTests in order to ensure it starts with a fresh InMemory db.

## Increased the access token lifetime for Keycloak

As a temporary fix for 401s caused by direct-to-API calls during slower UI test runs (e.g. admin_api.delete_test_user) when using Keycloak, I've increase the lifetime of the tokens it issues.  The real fix is to reauthenticate on getting 401s with these direct API calls during Robot test runs, but that'll take a bit longer to put in place properly.

# UI test results

## Using Entra ID

image

## Using Keycloak

image

### DIFF
--- a/docker/keycloak/keycloak-ees-realm.json
+++ b/docker/keycloak/keycloak-ees-realm.json
@@ -5,8 +5,8 @@
   "defaultSignatureAlgorithm" : "RS256",
   "revokeRefreshToken" : false,
   "refreshTokenMaxReuse" : 0,
-  "accessTokenLifespan" : 300,
-  "accessTokenLifespanForImplicitFlow" : 900,
+  "accessTokenLifespan" : 1800,
+  "accessTokenLifespanForImplicitFlow" : 1800,
   "ssoSessionIdleTimeout" : 1800,
   "ssoSessionMaxLifespan" : 36000,
   "ssoSessionIdleTimeoutRememberMe" : 0,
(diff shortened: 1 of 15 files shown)


sambiramairelogic:
image
Works for me :D

sambiramairelogic (approved):

PR #4602, opened 2024-02-26
EES-4517 Add GetDataSet endpoint

mmyoungman:
This PR adds a new endpoint that will be used by the new Data Catalogue Data Set pages.

No caching has been included as part of this work. We believe this will be done on the frontend. See EES-4931.


src/GovUk.Education.ExploreEducationStatistics.Content.ViewModels/DataSetDetailsViewModel.cs
```diff
@@ -0,0 +1,46 @@
...
+    public DateTime Published { get; set; }
+
+    public DataSetDetailsPublicationViewModel Publication { get; set; } = null!;
+}
+
+public record DataSetDetailsFileViewModel
+{
+    public Guid Id { get; set; }
+    public string Name { get; set; } = null!;
+    public string Size { get; set; } = null!;
```
benoutram, line 45: Couldn't all the properties on these view models use `init` acessors rather than `set`? I don't think you ever expect them to be set again once initialised even in unit tests.

src/GovUk.Education.ExploreEducationStatistics.Content.ViewModels/DataSetDetailsViewModel.cs
```diff
@@ -0,0 +1,46 @@
...
+using Newtonsoft.Json.Converters;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.ViewModels;
+
+public record DataSetDetailsViewModel
+{
+    public string Title { get; set; } = null!;
+    public string Summary { get; set; } = null!;
+    public DataSetDetailsFileViewModel File { get; set; } = null!;
+    public DataSetDetailsReleaseViewModel Release { get; set; } = null!;
```
benoutram, line 12: Just wondering if we should have blank lines separating fields to be consistent with the spacing in other models like `DataSetDetailsReleaseViewModel` where annotations force the blank lines.

I've been wondering this in some of my own changes recently. Hopefully the code standards coming in soon will enforce this one way or the other.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/DataSetServiceTests.cs
```diff
@@ -0,0 +1,205 @@
...
+    public class GetDataSetTests : DataSetServiceTests
+    {
+        private readonly DataFixture _fixture = new();
+
+        [Fact]
+        public async Task Success()
+        {
+            var publication = _fixture.DefaultPublication()
+                .WithReleaseParents(
+                    _fixture.DefaultReleaseParent(publishedVersions: 1, draftVersion: true,  year: 2000)
```
benoutram, line 25: I don't think `year: 2000` is important to any of these tests so I would consider removing the  year property.

Year will be initialised by default.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/DataSetServiceTests.cs
```diff
@@ -0,0 +1,205 @@
...
+using Xunit;
+using static GovUk.Education.ExploreEducationStatistics.Content.Model.Tests.Utils.ContentDbUtils;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Services.Tests;
+
+public class DataSetServiceTests
+{
+    public class GetDataSetTests : DataSetServiceTests
+    {
+        private readonly DataFixture _fixture = new();
```
benoutram, line 18: Changing `_dataFixture` to be declared outside of `GetDataSetTests` will mean its accessible to any other methods which get tested in this service which are added later on.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/DataSetServiceTests.cs
```diff
@@ -0,0 +1,205 @@
...
+            {
+                var dataSetService = BuildService(contentDbContext);
+
+                var result = await dataSetService.GetDataSet(
+                    Guid.NewGuid(), Guid.NewGuid());
+
+                result.AssertNotFound();
+
+            }
+        }
```
benoutram, line 99: This could be testing no release _or_ no files.

I would still save a release here with a file and for the method under test make the following call.

```csharp
GetDataSet(releaseId: Guid.NewGuid(),
    fileId: releaseFile.FileId)
```

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/DataSetServiceTests.cs
```diff
@@ -0,0 +1,205 @@
...
+
+            var releaseFile = _fixture.DefaultReleaseFile()
+                .WithRelease(publication.Releases[0])
+                .WithFile(_fixture.DefaultFile())
+                .Generate();
+
+            var contentDbContextId = Guid.NewGuid().ToString();
+            await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId))
+            {
+                await contentDbContext.ReleaseFiles.AddRangeAsync(releaseFile);
```
benoutram, line 39: [Styleguide 2.3.2](https://github.com/dfe-analytical-services/explore-education-statistics/blob/dev/styleguides/styleguide-backend.md#232-entity-framework-ef) - Use `contentDbContext.ReleaseFiles.Add(releaseFile);` rather than the async variant.

Applies to multiple occurrences in this file.

Could also consider making this `contentDbContext.Publications.Add(publication);` Feels more natural to be saving the graph top-down but I don't think it matters at all.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/DataSetServiceTests.cs
```diff
@@ -0,0 +1,205 @@
...
+                    _fixture.DefaultReleaseParent(publishedVersions: 0, draftVersion: true, year: 2000)
+                        .Generate(1))
+                .WithTopic(_fixture.DefaultTopic()
+                    .WithTheme(_fixture.DefaultTheme()))
+                .Generate();
+
+            var releaseFile = _fixture.DefaultReleaseFile()
+                .WithRelease(publication.Releases[0])
+                .WithFile(_fixture.DefaultFile())
+                .Generate();
```
benoutram, line 144: Fyi if you want to you can avoid two `.Generate()` lines here and elsewhere in this file where you use `.Generate()` by using the implicit operator already defined by `Generator`:

```csharp
    /// <summary>
    /// Automatically generate an instance of T from the generator
    /// using implicit type conversion.
    /// </summary>
    public static implicit operator T(Generator<T> generator)
    {
        return generator.Generate();
    }
```

As an example:

```csharp
Publication publication = _fixture.DefaultPublication()
      .WithReleaseParents(
          _fixture.DefaultReleaseParent(publishedVersions: 0, draftVersion: true, year: 2000)
              .Generate(1))
      .WithTopic(_fixture.DefaultTopic()
          .WithTheme(_fixture.DefaultTheme()));
```

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/DataSetServiceTests.cs
```diff
@@ -0,0 +1,205 @@
...
+            {
+                var dataSetService = BuildService(contentDbContext);
+
+                var result = await dataSetService.GetDataSet(
+                    publication.Releases[0].Id, Guid.NewGuid());
+
+                result.AssertNotFound();
+            }
+        }
+    }
```
benoutram, line 197: I think I would leave the previous test to test an unpublished version which it's already doing, and change this into one which tests a release which isn't the latest published version.

It would still test `publication.Releases[0].Id` but you would give it `publishedVersions: 2` and it could also have a draft version.

You would set up the release file to be on all 3 versions.

If you don't do this, you haven't got a test which checks you can't get a data set that exists on a previously published  version.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Services/DataSetService.cs
```diff
@@ -118,6 +123,57 @@ public DataSetService(ContentDbContext contentDbContext)
...
+                    Published = releaseFile.Release.Published!.Value,
+                    Publication = new DataSetDetailsPublicationViewModel()
+                    {
+                        Id = releaseFile.Release.PublicationId,
+                        Title = releaseFile.Release.Publication.Title,
+                        Slug = releaseFile.Release.Publication.Slug,
+                        ThemeTitle = releaseFile.Release.Publication.Topic.Theme.Title,
+                    },
+                },
+                File = new DataSetDetailsFileViewModel()
```
benoutram, line 161: Parentheses here in `new DataSetDetailsFileViewModel()` are redundant when using object initialiser syntax.

This comment applies elsewhere in this file too.

src/GovUk.Education.ExploreEducationStatistics.Content.ViewModels/DataSetDetailsViewModel.cs
mmyoungman: I put blank lines on the one view model because it had the `JsonConvertor` on one property - the rest I figured I could condense. I've added spaces between the properties of the others as well now.

I'd not be inclined to stipulate a way in coding standards, to be honest - I don't think it matters.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/DataSetServiceTests.cs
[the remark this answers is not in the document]
mmyoungman: I didn't want to have to rewrite the tests again as integration tests so I was intending to leave them. But if you feel strongly about it I'll change them.

src/GovUk.Education.ExploreEducationStatistics.Content.ViewModels/DataSetDetailsViewModel.cs
benoutram: Looks good now. 👍

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/DataSetServiceTests.cs
benoutram: Reviewed latest changes and I still think it's a bit confusing as its trying to be both a "FileIsNotAssociatedWithRelease" and a "ReleaseIsNotLatestPublishedVersion" test at the same time.

You could argue you don't need the first of these cases as you've already got a "NoFile" test.

For the latter case I still think it would be better if the test set up the same file for all versions.

All other changes seem fine btw.

A blank line has crept in on L102.

I didn't mention this earlier but there's a style guide pointer now about test method names now following `Scenario_Result` pattern with  the simple happy-path scenario being called `Success` for succinctness. I'm not sure if this means some of your tests here should be suffixed with `_ReturnsNotFound` or not, e.g. `NoFile` - `NoFile_ReturnsNotFound`.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/DataSetService.cs
```diff
@@ -118,6 +123,57 @@ public DataSetService(ContentDbContext contentDbContext)
...
     }
+
+    public async Task<Either<ActionResult, DataSetDetailsViewModel>> GetDataSet(
+        Guid releaseId,
+        Guid fileId)
+    {
+        if (!await _releaseRepository.IsLatestPublishedReleaseVersion(releaseId))
+        {
+            return new NotFoundResult();
+        }
```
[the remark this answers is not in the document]
mmyoungman: Looked into this and have discussed it with Ben. 

There is a `userService.CheckCanViewRelease` auth handler, but when I tried to use it that introduces trouble when mocking this - meaning it wouldn't actually have tested the method as thoroughly, as the `IsLatestPublishedReleaseVersion` wouldn't have actually got run against the test data provided.

If there were an easy way to create an `IUserService` via the integration tests, then using the auth handler would have been (almost) equivalent - and this may be possible in the future if the Content API integration tests are set up similar to how the Admin integration tests are.

My overall analysis of the benefits of using the auth handler is that there is really only one benefit: signalling to anyone reading this code that this check is important.

I've added a `WARN` comment above the `IsLatestPublishedReleaseVersion` check.

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/DataSetsControllerTests.cs
```diff
@@ -1644,14 +1640,201 @@ private List<ReleaseFile> GenerateDataSetsForRelease(Release release, int number
...
+                    contentDbContext ?? s.GetRequiredService<ContentDbContext>()));
+                services.AddTransient<IDataSetService>(
+                    s =>
+                    {
+                        var context = contentDbContext ?? s.GetRequiredService<ContentDbContext>();
+
+                        return new DataSetService(
+                            context,
+                            new ReleaseRepository(context));
+                    });
```
benoutram, line 1837: You could do this to simplify the configuration of the `IDataSetService` type:

```csharp
                services.AddTransient<IDataSetService>(
                    s => new DataSetService(
                        contentDbContext ?? s.GetRequiredService<ContentDbContext>(),
                        s.GetRequiredService<IReleaseRepository>()));
```

Alternatively you could leave that as it is, and remove the configuration for the `IReleaseRepository` type which isn't strictly needed with your current setup of `DataSetService`.

benoutram (approved):

PR #4606, opened 2024-02-28
Add 'Block auto-squash commits' GitHub workflow

ntsim:
This PR adds a GitHub workflow to prevent auto-squash commits (e.g. `fixup`) from being merged via pull requests.

To resolve this, these commits should be auto-squashed via `git rebase {branch} -i --autosquash` before the branch is merged.


jack-hive (approved):
If it's doing what I think it's doing, it's a nice idea!

PR #4610, opened 2024-03-01
EES- 4822/Use 301 for Redirects

sambiramairelogic:
This PR reconfigures the public site to redirect using HTTP Status Code 301 rather than 308. 

NextJS uses Http Status 308 for permanent redirects, as oppose to the usual 301. This is a deliberate design choice as 308s preserve the HTTP Verb when redirecting, however 301s turn everything to a GET. Unfortunately, NextJS does not provide a configuration option to use 301s instead. Luckily, we are using a custom server, so we are able to configure redirects in Express JS before reaching Next at all. 

By default, Next sets the [trailingSlash](https://nextjs.org/docs/app/api-reference/next-config-js/trailingSlash) config to false, meaning urls ending with slashes get redirected with the slash curtailed. This behaviour is re-created in `server.js` with the 301 status code. 

We also have two manual redirects, `meta-guidance` -> `data-guidance` and `download-latest-data` -> `data-catalogue`. These ideally need to be removed, however Google has trawled them within the past 6 months so I don't believe we can just yet. For now, these have also been re-created using 301s. 

New robot tests have been added to verify that these redirects still perform correctly, although sadly I don't believe robot can assert on which status code was used for the redirect. 

Exactly the same tests like-for-like have been created in Playwright as an experiment (spoiler alert: the playwright tests were SO much easier and took less than a quarter of the time...)

### DIFF
--- a/src/explore-education-statistics-frontend/server.js
+++ b/src/explore-education-statistics-frontend/server.js
@@ -69,13 +69,16 @@ async function startServer() {
   const server = express();
 
   function replaceLastOccurance(input, pattern, replacement) {
-    if (input === undefined || input === null || input.length === 0) {
+    if (
+      input === undefined ||
+      input === null ||
+      input.length === 0 ||
+      !input.endsWith(pattern)
+    ) {
       return input;
     }
 
-    return input.endsWith(pattern)
-      ? `${input.slice(0, -pattern.length)}${replacement}`
-      : input;
+    return `${input.slice(0, -pattern.length)}${replacement}`;
   }
 
   // Redirect URLs with trailing slash to equivalent without slash with 301
@@ -89,7 +92,7 @@ async function startServer() {
       '/data-catalogue',
     );
 
-    if (newUri !== req.url) {
+    if (newUri !== req.url && newUri !== '') {
       return res.redirect(301, newUri);
     }
     nextNotShadowed();
--- a/src/explore-education-statistics-frontend/server.js
+++ b/src/explore-education-statistics-frontend/server.js
@@ -68,6 +68,34 @@ async function startServer() {
 
   const server = express();
 
+  function replaceLastOccurance(input, pattern, replacement) {
+    if (input === undefined || input === null || input.length === 0) {
+      return input;
+    }
+
+    return input.endsWith(pattern)
+      ? `${input.slice(0, -pattern.length)}${replacement}`
+      : input;
+  }
+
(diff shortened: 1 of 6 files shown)


src/explore-education-statistics-frontend/server.js
```diff
@@ -68,6 +68,37 @@ async function startServer() {
...
+      input === undefined ||
+      input === null ||
+      input.length === 0 ||
+      !input.endsWith(pattern)
+    ) {
+      return input;
+    }
+
+    return `${input.slice(0, -pattern.length)}${replacement}`;
+  }
```
sambiramairelogic, line 82: I very much confess that this abstraction is quite noddy - but technically the "rule of three" applies! :D 

Very happy to revert to repeated lines if anyone feels strongly

src/explore-education-statistics-frontend/server.js
```diff
@@ -68,6 +68,37 @@ async function startServer() {
...
+    newUri = replaceLastOccurance(
+      newUri,
+      '/download-latest-data',
+      '/data-catalogue',
+    );
+
+    if (newUri !== req.url && newUri !== '') {
+      return res.redirect(301, newUri);
+    }
+    nextNotShadowed();
```
sambiramairelogic, line 98: If anyone can come up with a better name for this, be my guest. 

This method is ExpressJS' callback to just "go onto the next thing", which they call `next` by default. This is however shadowed if you import `next` (as in NextJS) above! I considered `continue` or `nextMiddleware` or `proceed` but for obvious reasons they're also problematic 🤷

mmyoungman (changes requested):

tests/robot-tests/tests/libs/public-utilities.py
```diff
@@ -24,6 +24,23 @@ def cookie_should_have_value(name, value):
     raise_assertion_error(f"Couldn't find cookie {name} with value {value}")
 
 
+def browser_should_be_at_address(expected):
```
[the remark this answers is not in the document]
sambiramairelogic: Aha, I hadn't realised. Will change :)

mmyoungman (approved):
The `server.js` changes look straightforward and I struggle to imagine what could be wrong with them, but I've hardly ever looked at Express so would suggest someone else taking a look before merging.

amyb-hiveit (approved):
LGTM 👍

sambiramairelogic:
Benchmark on Dev
With301Redirects

PR #4611, opened 2024-03-01
EES-4668 Backend rename of Release to ReleaseVersion

benoutram:
⚠️ This change relies on clearing private and public blob caches, see [EES-4957](https://dfedigital.atlassian.net/browse/EES-4957).

This PR is a backend renaming of Release to ReleaseVersion.

Currently when we refer to multiple releases it's not clear whether this could be...

Multiple versions of the same release:
2020 Academic Year version 0
2020 Academic Year version 1
2020 Academic Year version 2

Multiple releases:
2020 Academic Year version 0
2021 Academic Year version 0
2022 Academic Year version 0

Multiple releases with multiple versions:
2020 Academic Year version 0
2020 Academic Year version 1
2021 Academic Year version 0
2021 Academic Year version 1

The introduction of ReleaseParent and renaming of Release to ReleaseVersion makes it clear that a version belongs to a ReleaseParent and that all versions with the same parent belong to the same Release.

In a further change, ReleaseParent could also be renamed to Release, similar to the Methodology/MethodologyVersion hierarchy, and renaming of DataBlockParent to DataBlock that that we also want to do at a later data when DataBlock is split from ContentBlock.

Unfortunately there's no getting away from the fact that even though this is a rename, it's  quite a big change, hence why I limited it to the backend rename only for now.

I have tried to review all changes as I've been adding them so I recommend the PR review concentrates on the database migrations and looking for anything that I may have forgotten to rename or thinking about problems that this rename might cause.

As it's limited to the backend rename only, there are no view model changes, for example, where `ReleaseId` is returned to the frontend in a view model.

### UI test report

image


mmyoungman (approved):
I haven't found any issues - here's what I've checked:
- That every dropped foreign key is recreated
- That there are no table columns missed in the renaming
- That no DB columns that contains JSON (like DataBlock_Charts/Query) contain a reference to `releaseId`
- That there is no raw SQL query referencing `Releases` - although there are some in `useful-scripts`, but given the size of this PR, probably do that another time.

I've not checked all the other changes.

PR #4616, opened 2024-03-04
EES-4938 update table header reordering for WCAG 2.2

amyb-hiveit:
Updates the table header reordering to be compliant with WCAG 2.2 success criterion [2.5.7 Dragging Movements](https://www.w3.org/WAI/WCAG22/Understanding/dragging-movements.html). To be compliant we need to update the component so that reordering can be achieved via single pointer operation as well as by drag and drop and keyboard operation.

### Reordering items in a group:
Added a 'Move' button to each item

reorder-items-1

When clicked it shows arrow buttons to move the item up and down and a 'Done' button
reorder-items-2

### Reordering groups:

Changed the 'Move to columns / rows' button to just 'Move'

reorder-group-2

When clicked it shows arrow buttons to move the group left or right (or up and down on mobile),  the 'Move to columns / rows' button and a 'Done' button.

reorder-group


src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersAxis.tsx
```diff
@@ -103,17 +119,29 @@ export default function TableHeadersAxis({
               <div className={styles.groupsContainer}>
                 {values.map((group: Filter[], index: number) => {
                   const key = `group-${index}`;
+                  const groupName = `${name}[${index}]`;
+                  const groupLegend = getGroupLegend(group);
+                  const groupId = getTableHeaderGroupId(groupLegend);
```
sambiramairelogic, line 124: Purely a stylistic suggestion, but I wonder if declaring an interface with these four properties might be clearer than instantiating four variables? 

Then, since `id`, `index`, and the `moveUp`/`moveDown` methods are all passed down twice, we could even have each of them extend it in their props then just use the spread operator all the way down. 

Doesn't make a lick of difference, just thought it might be more concise :)

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersGroup.tsx
```diff
@@ -77,10 +86,15 @@ const TableHeadersGroup = ({
           <TableHeadersGroupControls
             defaultNumberOfItems={defaultNumberOfItems}
             groupName={name}
-            id={groupId}
+            id={id}
+            index={index}
+            isLastGroup={isLastGroup}
```
sambiramairelogic, line 91: If one were to be suuuuuuuuper pedantic, one might point out that we pull an `activeGroup` id out of the TableHeadersContext rather than an `isActiveGroup` flag, so we're introducing two different strategies for how we check for "is this the active group?" and "is this the last group?". 

Personally, I don't have the slightest problem with this as each one makes perfect sense in its context :D Just pointing it out...

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/TableHeadersForm.test.tsx
```diff
@@ -217,6 +217,8 @@ describe('TableHeadersForm', () => {
       columnAxis.getByRole('heading', { name: 'Locations' }),
     ).toBeInTheDocument();
 
+    userEvent.click(screen.getByRole('button', { name: 'Move Locations' }));
```
sambiramairelogic, line 220: Aha, was about to comment that since we're editing these tests there might be argument for converting them to the `setup ... / await user.click(element)` sort of style, but then realised of course we haven't merged the upgrade yet. Shame to create extra tech debt, but ho hum

sambiramairelogic:
Reordering
As far as I can tell the new single pointer operations work flawlessly, but unfortunately I think we've introduced a bug in the mouse drag & drop controls. I was able to get it into a state where dragging a group to the right didn't move it, or it did but then you can't then drag it back again. The ability to insert groups between two others also breaks quite easily, but all of these bugs go away if you click on an individual group's Reorder button then cancel again, so probably just a minor state not-updating-properly thingy? 

Vague steps to replicate are: 
* Follow table_tool_absence_by_characteristic.robot or otherwise load up a table with a few groups
* Move groups until a row contains at least 3
* (drag & drop) Move the left-most to the right-far-end, then attempt to drag it into the middle. For me this step fails.
* (drag & drop) Attempt to move the left-most to the right-most, this won't work and you're now in a bugged state.

If you have any trouble replicating let me know - maybe I was doing something silly!

sambiramairelogic (changes requested):
Damn fine work on the new single pointer buttons, they're super slick and easy to use, and considering their complexity the state logic is so concise. I'd have made a much larger mess of this :D 

Unfortunately I think one tiny bug has been introduced in the old drag & drop functionality, so requesting changes for now

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/TableHeadersForm.test.tsx
amyb-hiveit: yeah :( I'll rebase this and the other recent changes into the React branch once it gets into dev.

amyb-hiveit:
> Reordering [ Reordering ](https://private-user-images.githubusercontent.com/63285990/309939105-18998584-3195-463f-9f4f-f2f1f2a734a0.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDk2MzY2NDQsIm5iZiI6MTcwOTYzNjM0NCwicGF0aCI6Ii82MzI4NTk5MC8zMDk5MzkxMDUtMTg5OTg1ODQtMzE5NS00NjNmLTlmNGYtZjJmMWYyYTczNGEwLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAzMDUlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMzA1VDEwNTkwNFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTdlNDdmOWQ5NjkyMDRmNTc5M2RiNzE2NDZjN2VhM2QwYjY4MGY1MGExNDlhOGQwM2VjYTBkZTRjOWEzZDRiMzkmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.ku3YY9aal3-pqaX4uxsk7rHRaQ6yeK7eYd7isAMer8w) [ ](https://private-user-images.githubusercontent.com/63285990/309939105-18998584-3195-463f-9f4f-f2f1f2a734a0.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDk2MzY2NDQsIm5iZiI6MTcwOTYzNjM0NCwicGF0aCI6Ii82MzI4NTk5MC8zMDk5MzkxMDUtMTg5OTg1ODQtMzE5NS00NjNmLTlmNGYtZjJmMWYyYTczNGEwLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAzMDUlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMzA1VDEwNTkwNFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTdlNDdmOWQ5NjkyMDRmNTc5M2RiNzE2NDZjN2VhM2QwYjY4MGY1MGExNDlhOGQwM2VjYTBkZTRjOWEzZDRiMzkmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.ku3YY9aal3-pqaX4uxsk7rHRaQ6yeK7eYd7isAMer8w) As far as I can tell the new single pointer operations work flawlessly, but unfortunately I think we've introduced a bug in the mouse drag & drop controls. I was able to get it into a state where dragging a group to the right didn't move it, or it did but then you can't then drag it back again. The ability to insert groups between two others also breaks quite easily, but all of these bugs go away if you click on an individual group's Reorder button then cancel again, so probably just a minor state not-updating-properly thingy?
> 
> Vague steps to replicate are:
> 
> * Follow table_tool_absence_by_characteristic.robot or otherwise load up a table with a few groups
> * Move groups until a row contains at least 3
> * (drag & drop) Move the left-most to the right-far-end, then attempt to drag it into the middle. For me this step fails.
> * (drag & drop) Attempt to move the left-most to the right-most, this won't work and you're now in a bugged state.
> 
> If you have any trouble replicating let me know - maybe I was doing something silly!

Good spot! I've fixed this now - the draggable didn't like the `draggableId` being changed to use the `id` instead of the `name`.

sambiramairelogic (approved):
Awesome job!

PR #4617, opened 2024-03-04
EES-4404 add region role to banners

amyb-hiveit:
Adds the `region` role to the phase and user testing banners as it is best practice for all content, except skip links, to be contained within landmarks.

We had two almost identical `PageBanner` components so I've combined them into one called `PhaseBanner` as that's what it's called in the design system (https://design-system.service.gov.uk/components/phase-banner/)


src/explore-education-statistics-admin/src/components/Page.tsx
```diff
@@ -45,7 +45,9 @@ const Page = ({
           'dfe-width-container--wide': wide,
         })}
       >
-        {pageBanner ?? <PageBanner />}
+        {pageBanner ?? (
+          <PhaseBanner url="https://forms.office.com/Pages/ResponsePage.aspx?id=yXfS-grGoU2187O4s0qC-VQ56HAfKLpBrG0LxbfxbVdUQjVJQVdMOFlSMURGQ1kyMzRNWlpKN1NMVy4u" />
```
sambiramairelogic, line 49: Just a minor thought, but I wonder if it's worth abstracting this URL to somewhere? 

In previous projects I've seen things like this all built up a `Constants.ts` file, or pulled out of config. Only a thought/suggestion to help readability/maintainability, definitely not a blocking comment :-)

sambiramairelogic (dismissed):
Left one casual suggestion but LGTM 👍

src/explore-education-statistics-admin/src/components/Page.tsx
amyb-hiveit: Yeah that's worth considering, I'll not do it for this hotfix though.

sambiramairelogic (approved):

PR #4619, opened 2024-03-05
EES-4802: Add redirects for broken links Google has trawled

sambiramairelogic:
Google has picked up nearly a thousand broken links on our site, which negatively affect our SEO rating. These ideally should be redirected to parent pages for the next 6 months, so that Google picks up a new canonical address for each and clears the old url out of its registry. 

As ever, these redirects need to be 301s rather than 308s, so have been implemented in server.js. 

I'd still like to find a better way of doing this than pattern matching the strings themselves. For example an Apache server could simply upload a .htaccess file to achieve this, but I haven't found an equivalent for ExpressJS.

Where Google has crawled things like: `/api/methodologies/{methodologyId}/images/9348e5bc-c09d-43b4-de90-08d90ae3d820` I've implemented the redirect using URL escape characters, as that is what the browser will swap them out for anyway.


sambiramairelogic:
Benchmark on Dev
WithAllRedirects
Amazingly, bafflingly, somehow, this doesn't appear to have an impact on performance... I'm skeptical. If this gets merged, I'd like to do some more thorough testing against dev.

tests/playwright-tests/tests/public/redirects.spec.ts
```diff
@@ -1,6 +1,12 @@
+/* eslint-disable no-restricted-syntax */
+/* eslint-disable no-await-in-loop */
```
sambiramairelogic, line 2: I'd rather not have to break rules to write a parameterised test, but Playwright lacks Jest's async `it.each` method and so I'm sticking to doing it how the docs tell you to. Unless we definitely decide to adopt Playwright, I don't think it's worth spending more time on as this isn't even a test I'd want to run in the pipeline.

src/explore-education-statistics-frontend/server.js
```diff
@@ -81,20 +83,42 @@ async function startServer() {
...
     let newUri = req.url;
+    // Redirect URLs with trailing slash to equivalent without slash with 301
     newUri = replaceLastOccurance(newUri, '/', '');
+
+    // These are generated by spam sites. We could redirect them too to reduce Google's indexed 404 pages,
+    // but it seems current advice is actually to ignore these:
+    // https://support.google.com/webmasters/thread/253569816/google-crawler-adding-a-1000-to-the-end-of-a-ton-of-urls?hl=en
+    // newUri = replaceLastOccurance(newUri, '/1000', '');
+
+    // TODO: Remove these redirects after Google's index is updated (will happen within 6 months)
```
amyb-hiveit, line 105: Is there a jira ticket we can reference for this to do?

src/explore-education-statistics-frontend/redirects.js
```diff
@@ -0,0 +1,2262 @@
...
+
+  { from: '/data-table', to: '/' },
+  {
+    from: '/data-tablesnbsp;-nbsp;springnbsp;2022',
+    to: '/',
+  },
+  { from: '/data-tables/fast-track/[dataBlockId]', to: '/' },
+  { from: '/data-tables/fast-track/[dataBlockParentId]', to: '/' },
+  {
+    from: '/data-tables/fast-track/0164bb4a-6777-4d8f-a1a0-34ce69f75153',
```
amyb-hiveit, line 62: Have we investigated why these fast tracks are broken? The couple I've looked at are 500ing.

amyb-hiveit (approved):
Code-wise this seems fine. 

Have we done any investigation into why there are so many broken links on EES (am I right in thinking that the links are on EES not on other websites linking to EES)? The fast tracks and ones with things like `releaseId` in seem like they could be bugs.

src/explore-education-statistics-frontend/redirects.js
sambiramairelogic: Good spot - I hadn't realised that. The ones I tried were all 404s so I assumed just tables that no longer exist, or incorrect guids. I wonder if we could run them all through a tool to assess which are 404s and which are 500s...

sambiramairelogic: I wrote a Postman script to call every redirect link and assess the result. From this: 
* Many yield 404s as expected and should indeed be redirected
* Some yield 500s as you've noticed, which will be dealt with in a new ticket: https://dfedigital.atlassian.net/browse/EES-4978 
* Some result in infinite redirect loops, and will also be addressed in a new ticket: https://dfedigital.atlassian.net/browse/EES-4977

src/explore-education-statistics-frontend/server.js
sambiramairelogic: Good shout, have created (& referenced) https://dfedigital.atlassian.net/browse/EES-4979

PR #4640, opened 2024-03-09
EES-4716 Part 1 - Add `TimeIdentifier` and `GeographicLevel` schema filters and enum utils

ntsim:
This PR provides various bits of prep work for EES-4716 by providing `TimeIdentifier` and `GeographicLevel` schema filters. These are commonly used throughout the public API and it consequently makes sense to modify their schemas so that they output their codes (instead of their default representations).

## Related changes

- Added various `EnumUtil` methods for extra convenience when getting enum values or labels. These are cached and should be used in preference to calling `EnumUtil.GetEnums` and iterating over it.
- Added `TimeIdentifierUtils` to separate out the `AYNTP` release-only time identifier from the rest. All other time identifiers can participate in data queries.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Utils/EnumUtil.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Utils/EnumUtil.cs
@@ -1,6 +1,7 @@
 #nullable enable
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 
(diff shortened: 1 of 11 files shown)


src/GovUk.Education.ExploreEducationStatistics.Common/Utils/EnumUtil.cs
```diff
@@ -13,20 +14,47 @@ private static class EnumCache<TEnum> where TEnum : Enum
...
         {
             return enumValue;
         }
 
         throw new ArgumentException($"The value '{value}' is not a valid {typeof(TEnum).Name}");
     }
 
+    public static bool TryGetFromEnumValue<TEnum>(
+        string value,
+        [MaybeNullWhen(false)] out TEnum @enum)
```
jack-hive, line 45: Never seen this attribute before! You learn something every day :)

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/EnumUtil.cs
```diff
@@ -46,4 +89,22 @@ private static class EnumCache<TEnum> where TEnum : Enum
...
+        => EnumCache<TEnum>.ValuesSet.Value;
+
+    private static IEnumerable<string> GetEnumLabelsUncached<TEnum>() where TEnum : Enum
+        => GetEnums<TEnum>().Select(e => e.GetEnumLabel());
+
+    public static IReadOnlyList<string> GetEnumLabels<TEnum>() where TEnum : Enum
+        => EnumCache<TEnum>.LabelsList.Value;
+
+    public static IReadOnlySet<string> GetEnumLabelsSet<TEnum>() where TEnum : Enum
+        => EnumCache<TEnum>.LabelsSet.Value;
```
jack-hive, line 109: Might be best to order these so that `private` methods are together, and `public` methods are together. Just makes it slightly more readable IMO

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/GeographicLevelUtils.cs
```diff
@@ -8,8 +8,11 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Utils;
 
 public static class GeographicLevelUtils
 {
-    private static readonly Lazy<GeographicLevel[]> LevelsLazy =
-        new(EnumUtil.GetEnumsArray<GeographicLevel>);
+    private static readonly Lazy<IReadOnlyList<GeographicLevel>> LevelsLazy =
+        new(EnumUtil.GetEnums<GeographicLevel>());
+
+    private static readonly Lazy<IReadOnlyList<GeographicLevel>> OrderedLevelsLazy =
```
jack-hive, line 14: do we want to add some extra lazy's just for the list of values and list of labels? Utilising your new cached `GetEnumValues<TEnum>` and `GetEnumLabels<TEnum>` `EnumUtil` methods. Then these cached values can be used in the schema filter

jack-hive (changes requested):
Looks good! Only a minor comment or two

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Swagger/TimeIdentifierSchemaFilter.cs
```diff
@@ -0,0 +1,47 @@
...
+
+                The allowed values are:
+
+                - `AY` - academic year
+                - `AYQ1 - AYQ4` - academic year quarter 1 to 4
+                - `T1` - academic year's autumn term
+                - `T2` - academic year's spring term
+                - `T3` - academic year's summer term
+                - `T1T2` - academic year's autumn and spring term
+                - `CY` - calendar year
```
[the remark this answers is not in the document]
ntsim: Good spots, thanks!

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/GeographicLevelUtils.cs
ntsim: I've actually just gone in another direction and just scrapped the cached orderings completely. On reflection, it's not really necessary and it just creates a bunch more code we don't really need.

Just changed this to order within the schema filter instead.

PR #4641, opened 2024-03-09
EES-4716 Part 2 - Allow `SeparatedQueryModelBinder` to bind on string lists via attribute

ntsim:
:warning: Depends on #4640 

This changes `SeparatedQueryModelBinder` so that it can bind enumerable string types from a query parameter containing comma separated strings e.g. `?items=option-1,option-2`.

Previously, this was not possible as the model binder would ignore these model types on the basis that strings often contain commas without being intended to be used as comma-separated list.

We can now enable binding to enumerable string types by adding the `QuerySeparator` attribute to the property or parameter. Note that `FromQuery` is also required.

For example:

```cs
[HttpGet("api/test")]
public ActionResult MyEndpoint([FromQuery, QuerySeparator] IList<string> items)
{
}
```

We can make a request like `GET /api/test?items=item1,item2`, and the `items` list will now be bound with `item1` and `item2` as individual elements. Previously, this would only have bound as a single element like `item1,item2`.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/ModelBinding/SeparateQueryModelBinderTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/ModelBinding/SeparateQueryModelBinderTests.cs
@@ -1,0 +1,225 @@
+#nullable enable
+using System;
+using System.Collections.Generic;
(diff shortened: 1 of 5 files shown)


jack-hive (approved):
Nice!

PR #4642, opened 2024-03-10
EES-4716 Part 1b - Add security policies to public API

ntsim:
This PR adds the initial security policies required for the public API.

As part of this, we've updated `DataSetStatus` and `DataSetVersionStatus` to reflect recently agreed terminology for statuses in [EES-4898](https://dfedigital.atlassian.net/browse/EES-4898). This has required an update to `PublicDataDbContext` to rename the `Published` datetime column (on both `DataSet` and `DataSetVersion`) to `Withdrawn`.

The security policies have now been applied to the relevant `DataSetService` methods.

## Other changes

- Removed unneeded `ParquetFilename` from `DataSetVersion`. This isn't needed as we can compute Parquet file paths from the data set ID and version number.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetsControllerTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetsControllerTests.cs
@@ -84,7 +84,7 @@ public abstract class DataSetsControllerTests : IntegrationTestFixture
         [Theory]
         [InlineData(DataSetStatus.Draft)]
         [InlineData(DataSetStatus.Withdrawn)]
-        public async Task DataSetNotAvailable_Returns404(DataSetStatus dataSetStatus)
+        public async Task DataSetNotAvailable_Returns403(DataSetStatus dataSetStatus)
         {
             DataSet dataSet = DataFixture
                 .DefaultDataSet()
@@ -94,7 +94,7 @@ public abstract class DataSetsControllerTests : IntegrationTestFixture
 
             var response = await GetDataSet(dataSet.Id);
 
-            response.AssertNotFound();
+            response.AssertForbidden();
         }
 
         [Fact]
@@ -481,6 +481,22 @@ public abstract class DataSetsControllerTests : IntegrationTestFixture
             validationProblem.AssertHasInclusiveBetweenError("pageSize");
         }
 
+        [Theory]
+        [InlineData(DataSetStatus.Draft)]
+        [InlineData(DataSetStatus.Withdrawn)]
+        public async Task UnavailableDataSet_Returns503(DataSetStatus status)
+        {
+            DataSet dataSet = DataFixture
+                .DefaultDataSet()
+                .WithStatus(status);
+
+            await TestApp.AddTestData<PublicDataDbContext>(context => context.DataSets.Add(dataSet));
+
+            var response = await ListVersions(dataSetId: dataSet.Id);
+
+            response.AssertForbidden();
+        }
+
         [Fact]
         public async Task InvalidDataSetId_Returns404()
         {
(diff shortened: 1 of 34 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetService.cs
```diff
@@ -61,28 +70,40 @@ public DataSetService(PublicDataDbContext publicDataDbContext)
...
+        int pageSize)
     {
         var queryable = _publicDataDbContext.DataSetVersions
-            .Where(ds => ds.DataSetId == dataSetId)
+            .Where(ds => ds.DataSetId == dataSet.Id)
             .Where(ds => ds.Status == DataSetVersionStatus.Published
-                || ds.Status == DataSetVersionStatus.Unpublished
-                || ds.Status == DataSetVersionStatus.Deprecated);
+                         || ds.Status == DataSetVersionStatus.Withdrawn
+                         || ds.Status == DataSetVersionStatus.Deprecated);
```
jack-hive, line 97: I've been looking at this for a while wondering if we could possibly re-use the authorization logic we have in the authorization handlers here... Otherwise if we change the rule in one place we have to remember to change it here too...

I can't think of a way to do this currently, but It would be nice not to have this logic in 2 places. The authorization handlers currently work fine if checking against individual records, but isn't so great when looking for a list of records - we can't use it here unless we bring them all into memory first and then filter them :/

Have you got any ideas?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Security/AuthorizationHandlers/ViewDataSetVersionAuthorizationHandlerTests.cs
```diff
@@ -0,0 +1,54 @@
...
+        DataSetVersion dataSetVersion = _dataFixture
+            .DefaultDataSetVersion()
+            .WithStatus(status);
+
+        var handler = BuildHandler();
+        var context = CreateAnonymousAuthContext<ViewDataSetVersionRequirement, DataSetVersion>(dataSetVersion);
+
+        handler.HandleAsync(context);
+
+        Assert.False(context.HasSucceeded);
```
jack-hive, line 47: Or could do `context.HasFailed`. Either-or :)

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetsControllerTests.cs
```diff
@@ -481,6 +481,22 @@ public async Task PageSizeOutOfBounds_Returns400(int pageSize)
             validationProblem.AssertHasInclusiveBetweenError("pageSize");
         }
 
+        [Theory]
+        [InlineData(DataSetStatus.Draft)]
+        [InlineData(DataSetStatus.Withdrawn)]
+        public async Task UnavailableDataSet_Returns503(DataSetStatus status)
```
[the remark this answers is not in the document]
ntsim: D'oh :smile:

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Security/AuthorizationHandlers/ViewDataSetVersionAuthorizationHandlerTests.cs
ntsim: Not quite :smile: We have to explicitly call `context.Fail` for `context.HasFailed` to be set to true. 

We don't do this in the authorization handlers, so it'll just be set to false.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetService.cs
ntsim: As discussed on Slack, there isn't a great way to do this currently that isn't more involved due to how LINQ works. There's potentially ways to make this work better by wrangling with expressions, but in the interest of time, we'll continue with the current style.

To neaten things up some more, I've refactored these checks into more re-usable auth extension methods.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Security/AuthorizationHandlers/ViewDataSetVersionAuthorizationHandlerTests.cs
jack-hive: ahhh, fair enough haha

jack-hive (approved):

PR #4653, opened 2024-03-11
EES-4709 Adding wrapper for encoding and decoding unique IDs as 'Sqids'

jack-hive:
As part of the Public API work, we want to be able to provider users with unique, collision-free, and obfuscated IDs to be used when querying API Data Sets. These 'public IDs' point to metadata values such as filter options.

These changes introduce the [Sqids](https://github.com/sqids/sqids-dotnet?tab=readme-ov-file) NuGET package. It provides us with the above functionality and allows us to decode the 'Sqids' back into the original integer ID stored in our database. It also gives us the ability to make the IDs unique to our application by providing a shuffled alphabet.

This PR adds a wrapper around the Sqids library, providing our application with unique IDs (via a shuffled alphabet), and ensures that the generated IDs are at least 5 characters in length. It provides some utility methods to encode and decode. When decoding, it will ensure that the Sqid is the 'Canonical' encoding of the corresponding integer ID before successfully decoding it.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Utils/SqidEncoderTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Utils/SqidEncoderTests.cs
@@ -11,8 +11,6 @@ public abstract class SqidEncoderTests
         MinLength = 5,
     });
 
-    private readonly SqidsEncoder<int> _defaultSqidsEncoder = new();
-
     public class EncodeTests() : SqidEncoderTests
     {
         [Fact]
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Utils/SqidEncoderTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Utils/SqidEncoderTests.cs
@@ -1,0 +1,132 @@
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Utils;
+using Sqids;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests.Utils;
+
+public abstract class SqidEncoderTests
+{
+    private readonly SqidsEncoder<int> _sqidsEncoder = new(new()
+    {
+        Alphabet = "UjIo1rQXDHTmNsy6p4SlAtO0uqhfCiexFMVc5nG3v2Bbz8dKgWJRa79PZYwkEL",
+        MinLength = 5,
+    });
+
+    private readonly SqidsEncoder<int> _defaultSqidsEncoder = new();
+
+    public class EncodeTests() : SqidEncoderTests
+    {
+        [Fact]
+        public void Success()
+        {
+            var encodedNumber = SqidEncoder.Encode(1);
+
+            var expected = _sqidsEncoder.Encode(1);
+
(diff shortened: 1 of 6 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Utils/SqidEncoder.cs
```diff
@@ -0,0 +1,31 @@
+using Sqids;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Utils;
```
ntsim, line 3: Would put this in the `Public.Data.Utils` project

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Utils/SqidEncoderTests.cs
```diff
@@ -0,0 +1,130 @@
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Utils;
+using Sqids;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests.Utils;
```
ntsim, line 4: Would put this in the `Public.Data.Utils.Tests` project

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Utils/SqidEncoderTests.cs
jack-hive: I forgot that existed! Haha

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Utils/SqidEncoderTests.cs
```diff
@@ -0,0 +1,130 @@
...
+        [Fact]
+        public void Success()
+        {
+            var encodedNumber = SqidEncoder.Encode(1);
+
+            var expected = _sqidsEncoder.Encode(1);
+
+            Assert.Equal(expected, encodedNumber);
+        }
+    }
```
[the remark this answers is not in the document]
jack-hive: I agree. I almost didn't mind it at first because I thought if the alphabet/other options ever changed, in either place, I'd want the tests to fail. So it felt OK to me.

But I think hardcoding the hash in there would probably be better so I've changed it

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Utils/SqidEncoderTests.cs
```diff
@@ -0,0 +1,130 @@
...
+            var successful = SqidEncoder.TryDecodeSingleNumber(invalidSqid, out var decodedNumber);
+
+            Assert.False(successful);
+            Assert.Equal(0, decodedNumber);
+        }
+
+        [Theory]
+        [InlineData(1, 1)]
+        [InlineData(1, 1, 1)]
+        public void SqidRepresentsMultipleNumbers_Throws(params int[] numbers)
```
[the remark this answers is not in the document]
jack-hive: Copy/paste

Fixed :)

ntsim (approved):

PR #4656, opened 2024-03-12
EES-4987 Create Prod-Only Redirects.robot test

sambiramairelogic:
There are two specific redirects that are remaining for now, but want to be removed once Google agrees that there is no longer any traffic going to the old route. 

For now these are tested in redirects.robot but these tests require the use of seed data. This is currently being ran in prod where there isn't any seed data. This PR effectively removes those tests for Prod (but leaves the other redirects tests).

### DIFF
--- a/tests/robot-tests/tests/general_public/prod_only/redirects.robot
+++ b/tests/robot-tests/tests/general_public/prod_only/redirects.robot
@@ -5,7 +5,7 @@ Suite Setup         user opens the browser
 Suite Teardown      user closes the browser
 Test Setup          fail test fast if required
 
-Force Tags          GeneralPublic    Local    Dev    Test    Preprod
+Force Tags          GeneralPublic    Prod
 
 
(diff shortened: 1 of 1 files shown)


tests/robot-tests/tests/general_public/prod_only/redirects.robot
```diff
@@ -0,0 +1,42 @@
+*** Settings ***
+Resource            ../libs/public-common.robot
+
+Suite Setup         user opens the browser
+Suite Teardown      user closes the browser
+Test Setup          fail test fast if required
+
+Force Tags          GeneralPublic    Local    Dev    Test    Preprod
```
duncan-at-hiveit, line 8: Hey Sam, this combo of tags will unfortunately not run against Prod here.  For this test in this folder, I'd just include the "GeneralPublic" and "Prod" tags.

duncan-at-hiveit (approved):

PR #4658, opened 2024-03-12
Apply file_scoped_namespaces to the Common project

sambiramairelogic:
This is the sort of PR we can start generating lots of, to slowly apply the .editorconfig rules in reviewable chunks. 

This reformats ONLY the Common project, and applies only the file_scoped_namespaces rule. This was achieved by setting the severity of that rule to error then running 

`dotnet format style GovUk.Education.ExploreEducationStatistics.sln --severity error --include GovUk.Education.ExploreEducationStatistics.Common/`


sambiramairelogic:
Doesn't seem there's the appetite within the team to apply formatting rules en masse. We know the option is there if we change our minds in future

PR #4659, opened 2024-03-12
Remove root declaration from src/.editorconfig

sambiramairelogic:
Good spot @ntsim :)


src/.editorconfig
```diff
@@ -1,7 +1,6 @@
 ################################################################
 # Global preferences                                           #
 ################################################################
-root = true
```
ntsim, line 4: We'll need to put this in the actual root `.editorconfig` too

sambiramairelogic: Sorry, thought it was already was and didn't check. More haste less speed >.<

ntsim (approved):

PR #4674, opened 2024-03-18
Ees 4806 final cut

duncan-at-hiveit:
The output of this PR is a process for deploying the Public API Container App and having it up and running successfully with connection to the PSQL Flexible Server and its `public_data` database.

In order to facilitate this, this PR:
- adds `Publish` and `Docker Image Push` steps to the main EES Build pipeline.
- adds a pipeline completion trigger to the Public API Infrastructure / Deploy pipeline which will trigger the pipeline on completion of the main EES Build pipeline.
- specifies the correct Docker image to deploy to the Container App based upon the build number of the main EES Build pipeline run.
- adds configuration to the Public API container app to allow it to identify its Managed Identity when running in an Azure environment.
- adds code to `Startup.cs` in the Public API "API" project when running in Azure to use access tokens with a connection string to talk to PSQL Flexible Server using its Managed Identity.

# 2-step deploy process for Public API Infrastructure / Deploy pipeline

There is a chicken-and-egg scenario with the building and deployment of the Public API infrastructure and deployment.  The infrastructure has to be set up firstly in order to make PSQL available for adding users for the API Container App and the Data Processor Function App to use, then a manual process is needed to add those users to the server and appropriate permissions granted to the `public_data` database.

This is the purpose of the `psqlDbUsersAdded` flag in `main.bicep`.  The idea would be to firstly run the Public API Infrastructure / Deploy pipeline with this flag set to false in order to get the majority of the infrastructure set up in place (including PSQL).  Then the manual process would be carried out to add the appropriate db users and permissions.  *Then* the pipeline can be run with the flag set to true, which is its default.  This can be carried out either as a standalone run of the pipeline directly, or by allowing the main EES Build pipeline to run and the trigger this pipeline afterwards, which would then ensure that there is a Docker image to pull.

## Manual process for adding database users and permissions for Container App

After the first run of the pipeline to set up the PSQL infrastructure, we need a way to connect to the database.  At this point traffic is locked down to the VNet, so we need a way to connect to it.  In the EES Resource Groups, this *might* be possible via network peering that is already in place to support us in connecting to Azure SQL / Storage from our DfE laptops.  If this is not possible, we can spin up a temporary VM in the Resource Group and attach it to the VNet, and then SSH on.

After establishing contact with PSQL over the VNet, the following steps have to take place:

1. Add yourself as an Admin to PSQL - this can be done in the `Authentication` blade in Azure Portal when viewing PSQL.

image

2. Jump on the machine that has connection to PSQL over the VNet.
3. Install AZ CLI tools (`https://learn.microsoft.com/en-us/cli/azure/install-azure-cli-linux`), JQ (`sudo apt install jq`) and PSQL client (`sudo apt install postgresql-client`).
4. Log into Azure with `az login` and log in as yourself using the browser link and code provided.
5. Export environment variables to allow a connection to PSQL (replacing the PGHOST and PGUSER with the correct values):

```
export PGHOST=dfe-dev-dw-ees-publicapi-psql-flexibleserver.postgres.database.azure.com ; export PGUSER=duncan@hiveit.co.uk ; export PGPORT=5432 ; export PGDATABASE=postgres ; export PGPASSWORD="$(az account get-access-token --resource https://ossrdbms-aad.database.windows.net | jq -r .accessToken)"
```

6. Connect to the PSQL server by running `psql`.
7. Create a user representing the Managed Identity of the Container App's user-assigned identity (replacing with correct identity id):

`select * from pgaadauth_create_principal('dfe-dev-dw-ees-publicapi-id-api', false, false);`

8. Grant database access for this user (replacing with correct identity id):

```
\connect public_data;
GRANT ALL ON SCHEMA public TO "dfe-dev-dw-ees-publicapi-id-api";
```

We might want to fine-tune these permissions later!

9. We can now run the Public API Infrastructure / Deploy pipeline with `psqlDbUsersAdded` set to true!

# Other manual changes needed

## Adding new Resource Group-specific Variable Group but for non-KeyVault variables

This will be called "EES Bicep Infrastructure public - *resource group name*" and will contain the following variables specific to the target Resource Group:

image

We already have Resource Group-specific Variable Groups set up but they are linked to Key Vault and unfortunately the 2 types of variables (secrets from KV vs non-secrets) cannot coexist in a single Variable Group.

# Other miscellaneous changes in this PR

## Removing duplication of Admin build and test runs from the main EES Build PIpeline

It looked like we were building and running tests for Admin in the main stage of the main EES Build Pipeline, and then performing this again in the Admin-specific stage of the pipeline, so I've excluded Admin build steps from the main stage specifically.

## Removing "Restore" jobs from the main EES Build PIpeline

Dotnet Restores are implicitly part of Dotnet Builds, so there's no need to do them as separate jobs.

## Removal of Key Vault dependency on PSQL Bicep module

Given that we no longer need to store any secret connection strings in KV anymore, we don't need to reference KV at all in the Public API pipeline anymore.

## Extracting the creation of many sub-resources from Bicep modules to reduce redeploys of non-changing components

I found that a lot of extra time was being added to the deploy when one resource in a module needed updating but others did not - the deploy process also redeployed all of the resources that didn't need changing.  As an example, if PSQL redeployed previously, it would also redeploy the Private DNS zone and private links as well every time, despite them being unchanged.  As soon as these were pulled out of `postgresqlFlexibleServer.bicep` into `main.bicep`, this no longer occurred and the Private DNS zone etc only redeployed when actually necessary.  

## Upping default version of PSQL to 16

We are using features that required an upgrade from 14, so I have added this as part of this PR (otherwise the db migrations run by the Public API fail).

## Removed various explicit `dependsOn` statements

Seems that the general advice is to allow Bicep to work out the dependency tree implicitly, and fall back to explicit `dependsOn` only when necessary.

## Removed HTTPS support from Public API "API" project

As it's running behind Azure's Container App HTTPS support, the containers themselves are no longer required to run with HTTPS, and can be accessed purely on port 8080.

# Nice-to-haves that haven't been included in this PR

Int he existing main EES build pipeline, the Content and Data APIs are both published into a "public-api" subfolder currently, which muddies the water somewhat!  I've not tried to change this here as this would require a bunch of manual changes to the Classic Release pipeline and orchestration.  I think for a future piece of work, we will move these aside and allow the Public API components to take the "public-api" folder.  In the meantime in this PR, I've used the "public-api/api" subfolder for the Container App build, which would be my desired target for the published Container App artifacts anyway.


infrastructure/templates/public-api/components/containerApp.bicep
```diff
@@ -54,87 +51,39 @@ param minReplica int = 1
...
-    ]
+var containerAppIdentity = managedIdentityId != null ? {
+  type: 'UserAssigned'
+  userAssignedIdentities: {
+    '${managedIdentityId}': {}
   }
-  tags: tagValues
+} : {
+  type: 'SystemAssigned'
 }
```
benoutram, line 79: Rather than declaring this, why don't you reference the existing resource which has to exist at this point?

```
resource managedIdentity 'Microsoft.ManagedIdentity/userAssignedIdentities@2023-01-31' existing = {
  name: identityName
}
```

I think there's a reason why this has to be a managed identity rather than a system assigned identity because of the way a container app throws a provision error if it doesn't have the ACR pull role in advance of the container app being provisioned.

infrastructure/templates/public-api/main.bicep
```diff
@@ -116,68 +118,137 @@ module fileShareModule 'components/fileShares.bicep' = {
...
+  properties: {
+    sku: {
+      name: 'PerGB2018'
+    }
+  }
+  tags: tagValues
+}
+
+resource apiContainerAppEnvironment 'Microsoft.App/managedEnvironments@2023-05-01' = {
+  name: '${resourcePrefix}-cae-${apiContainerAppName}'
```
benoutram, line 188: I don't think it's really a problem here but the container app environment can be used for more than one container app where the apps are related. Apps would only use separate container app environments where they need to be isolated or shouldn't share the same resources. Should we name the container app environment something more generic than the app itself to reflect this?

infrastructure/templates/public-api/api-infrastructure-pipeline.yml
```diff
@@ -58,21 +66,29 @@ stages:
...
+          echo 'Using "$(dockerImagesTag)" as the tag to identify the Docker images to deploy with this pipeline run.'
+          echo
+          echo 'Structure of work folder of this pipeline:'
           tree $(Build.SourcesDirectory)
 
     - task: AzureCLI@2
       displayName: 'Install Bicep'
       inputs:
-        azureSubscription: ${{parameters.azureSubscription}}
+        azureSubscription: $(serviceConnection)
```
benoutram, line 83: Cool, is `serviceConnection` built in?

azure-pipelines.dfe.yml
```diff
@@ -107,22 +103,50 @@ jobs:
...
+        displayName: 'Build Public API - API Docker image'
+        condition: eq(variables.IsBranchDeployable, true)
+        inputs:
+          containerRegistry: 's101d-datahub-spn-ees-dfe-gov-uk-docker-managed-service-connection'
+          repository: 'ees-public-api/api'
+          command: 'build'
+          Dockerfile: 'docker/public-api/api/Dockerfile'
+          buildContext: '$(Build.ArtifactStagingDirectory)/public-api/api/GovUk.Education.ExploreEducationStatistics.Public.Data.Api'
+          tags: $(Build.BuildNumber)
+          arguments: '--build-arg BUILD_BUILDNUMBER=$(Build.BuildNumber)'
```
luke-h1, line 128: Just leaving a small comment (hope you don't mind - find this stuff interesting) - some good arguments I've seen passed to docker images like this are: who deployed it (can be sourced via git info I believe), age of the deploy (using a timestamp). Good thing about this is you can expose this info via an endpoint (i.e. `/api/v1/version`) so you can quickly debug stuff if things go haywire in prod

infrastructure/templates/public-api/components/containerApp.bicep
```diff
@@ -54,87 +51,39 @@ param minReplica int = 1
...
 
 resource containerApp 'Microsoft.App/containerApps@2023-05-01' = {
```
luke-h1, line 81: Was just curious - Have you folks considered putting a load balancer in front of the containers? Aware the usage that the public api is gonna get is a bit of an unknown but a load balancer would help with stampedes of traffic suddenly come thru to a given service. Also should provide lots of good things you can scale on out of the box i.e. if load balancer reqs > 10k then scale to 6 containers

duncan-at-hiveit: Well hello @luke-h1 :smile:  Hope you're doing well today!

I believe that Container Apps have built-in load balancing out of the box to a degree, but we'll need to see how that performs when all the code's good to go I guess.  The longer-term plans are still to have this sit behind Azure API Management for rate limiting, DDoS prevention etc, so this will likely add another dimension to this puzzle as well!

infrastructure/templates/public-api/api-infrastructure-pipeline.yml
duncan-at-hiveit: No, this comes from a Variable Group specific to the Resource Group being deployed to.  More to be seen of this in EES-4807 PR.

infrastructure/templates/public-api/main.bicep
duncan-at-hiveit: Hey Ben, yeah that sounds like a good idea.  I've already done this for the PostgreSQL Flexible Server to make it reusable elsewhere, so I'll follow the same convention here, ta

docker/public-api/api/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM mcr.microsoft.com/dotnet/aspnet:8.0
+WORKDIR /App
+
+EXPOSE 8080
+
+COPY [".", "."]
```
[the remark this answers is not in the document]
duncan-at-hiveit: I have removed, thanks

azure-pipelines.dfe.yml
```diff
@@ -107,22 +103,50 @@ jobs:
...
+        condition: eq(variables.IsBranchDeployable, true)
         inputs:
-          artifactName: 'public-api'
-          targetPath: '$(Build.ArtifactStagingDirectory)/public-api'
+          containerRegistry: 's101d-datahub-spn-ees-dfe-gov-uk-docker-managed-service-connection'
+          repository: 'ees-public-api/api'
+          command: 'push'
+          Dockerfile: 'docker/public-api/api/Dockerfile'
+          buildContext: '$(System.ArtifactStagingDirectory)/public-api/api'
+          tags: $(Build.BuildNumber)
```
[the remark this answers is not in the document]
duncan-at-hiveit: Amended this and the public front-end Docker image tasks to use the combined command now, thanks!

See EES-4807's https://github.com/dfe-analytical-services/explore-education-statistics/blob/EES-4807-build-and-deploy-function-app/azure-pipelines.dfe.yml

infrastructure/templates/public-api/main.bicep
```diff
@@ -116,68 +118,137 @@ module fileShareModule 'components/fileShares.bicep' = {
...
+
+@description('This allows the managed identity of the container app to access the registry, note scope is applied to the wider ResourceGroup not the ACR')
+resource apiContainerAppManagedIdentityRBAC 'Microsoft.Authorization/roleAssignments@2022-04-01' = {
+  name: guid(resourceGroup().id, apiContainerAppManagedIdentity.id, acrPullRole)
+  properties: {
+    roleDefinitionId: acrPullRole
+    principalId: apiContainerAppManagedIdentity.properties.principalId
+    principalType: 'ServicePrincipal'
+  }
+}
```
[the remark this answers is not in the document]
duncan-at-hiveit: I think I might have to trial running this on the Dev Resource Group, and make the necessary amends there should we run into issues.  I'm not sure either way what we'll have provisioned in terms of permissions for our existing service connections but we did a chunk of work with them a while back when rolling out the PNPM work for the first time.

infrastructure/templates/public-api/components/containerApp.bicep
duncan-at-hiveit: Yup that sounds like a good plan, thanks!  Updated as part of EES-4807.

infrastructure/templates/public-api/main.bicep
duncan-at-hiveit: Done.  Also done for the Log Analytics Workspace, as I believe that, as we've done with the Public API App Insights instance, it is shareable across several resources and might prove better insights by being combined.

azure-pipelines.dfe.yml
duncan-at-hiveit: Also, found the setting `addPipelineData` in the Docker build task which does a lot of this for free for you :)

infrastructure/templates/public-api/api-infrastructure-pipeline.yml
```diff
@@ -1,28 +1,30 @@
...
 
-name: Deploy Public API Bicep files
+resources:
+  pipelines:
+    - pipeline: EESBuildPipeline
+      source: Explore Education Statistics
+      trigger: true
+
+# This defines the "Build.BuildNumber" variable available in the pipeline.
+name: $(Date:yyyyMMdd).$(Rev:r)
```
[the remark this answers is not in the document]
duncan-at-hiveit: Yup you're right!  Removed now in EES-4807, thanks.

azure-pipelines.dfe.yml
luke-h1: Ah wicked, just thought: with it being a public api if you end up using semver versioning you could expose the version too

infrastructure/templates/public-api/components/containerApp.bicep
luke-h1: Cool, yeah having a waf / the API management stuff will help against the nefarious traffic I imagine. Another advantage of the load balancer approach I forgot to mention is that your DNS is attached to the load balancer, rather than the container app/app service. Makes things like server patching/upgrades a bit easier as you don't have to go around and edit any DNS records 😛

infrastructure/templates/public-api/main.bicep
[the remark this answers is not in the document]
benoutram: Ok yep I agree about trialling it out to find out.

The PNPM work is different because that's an app service running a container, not an app within a container app environment. That work also opted to use username/password authentication rather than a user managed identity where the identity would be granted the AcrPull role.

I can't see that the AcrPull role has any assignments for the existing ACR or that it's even a recommended way of doing it. We should look into sorting it at some point

benoutram (approved):

PR #4701, opened 2024-03-26
EES-4770 Add ability to order public release pages' release series

mmyoungman:
This PR allows the sorting of releases and legacy links under the "Releases in this series" section on public release pages. To achieve this, we now store both legacy links and a publications's releases in the `Publications` table's new `ReleaseSeries` column. Also as part of this change, releases are also displayed alongside legacy links on Admin's Legacy releases page for publications.

We still use the "legacy release" phrasing on public facing text, but internally the code now refers to them as "legacy links". Hopefully this difference will be unified in the future.

This PR builds on an original PR by Tom (https://github.com/dfe-analytical-services/explore-education-statistics/pull/4585), but has additional major changes due to two main reasons:
- After Tom had raised his PR, a change went in to create the `Release` and `ReleaseVersion` hierarchy, which created opportunities to simplify the code - specifically we didn't need to rely on `ReleaseVersionId`s in release series items, and so avoid having to handle amendments and other complications.
- I decided to take the opportunity to remove the `LegacyReleases` table completely.

There is a single change from the requirements - we don't display "Draft" or "Draft Amendment" labels against release series items. These have been substitued by an "Unpublished" label.

:rotating_light: There is a faffy deploy step of running the release series migration endpoint as this work is deployed across CI environments :rotating_light: 

:rotating_light: After this work has merged, to ensure existing legacy releases work correctly on your local environment, you'll need to use `ees-mssql-data-65.zip` :rotating_light: 

### Other notes

There are potentially other opportunities to simplify the code, but I didn't want to make them as this PR is already fairly big:
- I suspect that `PublicationViewModel` and `PublicationCacheViewModel` no longer need `Releases` - if `ReleaseSeries` doesn't already contain all the information required, it'd be easy to add what was needed.
- There is also potential for a Publication's ReleaseSeries to be used to determine a publication's latest release - allowing us to no longer rely on the ordering of the TimeIdentifier enum avoid issues like EES-5022.

The workhorse of the new endpoints is the new `UpdateReleaseSeries` endpoint. This sets the entire series in a single request, but is used by the frontend to edit, delete, and reorder particular release series items. This is why `UpdateReleaseSeries` checks that the release series contains all of a publication's releases - otherwise it would be possible for releases to be removed from the release series.


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
```diff
@@ -128,7 +128,7 @@ await using (var context = InMemoryApplicationDbContext(contextId))
             {
                 var actual = await context
                     .ReleaseVersions
-                    .SingleAsync(rv => rv.PublicationId == publication.Id);
```
benoutram, line 131: This is undoing a change made by the Release to ReleaseVersion renaming work, probably from resolving merge conflicts.

tomjonesdev (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/ReleaseVersionRepository.cs
```diff
@@ -172,6 +172,23 @@ public async Task<List<ReleaseVersion>> ListLatestReleaseVersions(CancellationTo
...
+            publishedOnly: publishedOnly,
+            cancellationToken: cancellationToken);
+
+        return await _contentDbContext.ReleaseVersions
+            .Where(rv =>
+                rv.ReleaseId == releaseId
+                && rv.Version == maxVersion)
+            .FirstOrDefaultAsync(cancellationToken);
+    }
+
```
benoutram, line 191: You could consider a version of this which doesn't involve two queries, because I think this could all be done with the following join in a single query (I didn't try this out, but should be something along these lines!).

```
            var maxVersionsQueryable = _contentDbContext.ReleaseVersions
                .Where(releaseVersion => releaseVersion.ReleaseId == releaseId)
                .Where(releaseVersion => !publishedOnly || releaseVersion.Published.HasValue)
                .GroupBy(releaseVersion => releaseVersion.ReleaseId)
                .Select(groupedVersions => groupedVersions.Max(releaseVersion => releaseVersion.Version));

            return await _contentDbContext.ReleaseVersions
                .Where(releaseVersion => releaseVersion.ReleaseId == releaseId)
                .Join(maxVersionsQueryable,
                    releaseVersion => releaseVersion.Version,
                    maxVersion => maxVersion,
                    (releaseVersion, _) => releaseVersion)
                .SingleAsync(cancellationToken: cancellationToken);
```

Rather than make this change in `ReleaseVersionRepository` you could also consider adding it to `ReleaseVersionPredicate` in a method something like this:

```
    public static IQueryable<ReleaseVersion> LatestReleaseVersion(this IQueryable<ReleaseVersion> releaseVersions,
        Guid releaseId,
        bool publishedOnly = false)
    {
        var maxVersionsQueryable = releaseVersions
            .Where(releaseVersion => releaseVersion.ReleaseId == releaseId)
            .Where(releaseVersion => !publishedOnly || releaseVersion.Published.HasValue)
            .GroupBy(releaseVersion => releaseVersion.ReleaseId)
            .Select(groupedVersions => groupedVersions.Max(releaseVersion => releaseVersion.Version));

        return releaseVersions
            .Where(releaseVersion => releaseVersion.ReleaseId == releaseId)
            .Join(maxVersionsQueryable,
                releaseVersion => releaseVersion.Version,
                maxVersion => maxVersion,
                (releaseVersion, _) => releaseVersion);
    }
```

The repository method wouldn't even be needed as anything using it can just make a call like ` _contentDbContext.ReleaseVersions.LatestReleaseVersion(releaseId, publishedOnly: true)` without needing to even depend on the repository.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/Interfaces/IReleaseVersionRepository.cs
```diff
@@ -33,6 +33,18 @@ public interface IReleaseVersionRepository
...
         CancellationToken cancellationToken = default);
 
+    /// <summary>
+    /// Retrieves the latest release version for a particular release.
+    /// </summary>
+    /// <param name="releaseId">The unique identifier of the release.</param>
+    /// <param name="publishedOnly">Return only published release version if true.</param>
+    /// <param name="cancellationToken">A <see cref="CancellationToken"/> to observe while waiting for the task to complete.</param>
+    /// <returns>The latest version that is associated with a particular release</returns>
+    Task<ReleaseVersion?> GetLatestReleaseVersionForParent(
```
benoutram, line 43: Now that `ReleaseParent` has been renamed to `Release`, `GetLatestReleaseVersionByRelease` or `GetLatestReleaseVersionForRelease` would be a more appropriate name.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/ReleaseVersionRepository.cs
```diff
@@ -172,6 +172,23 @@ public async Task<List<ReleaseVersion>> ListLatestReleaseVersions(CancellationTo
...
+            releaseId,
+            publishedOnly: publishedOnly,
+            cancellationToken: cancellationToken);
+
+        return await _contentDbContext.ReleaseVersions
+            .Where(rv =>
+                rv.ReleaseId == releaseId
+                && rv.Version == maxVersion)
+            .FirstOrDefaultAsync(cancellationToken);
+    }
```
benoutram, line 190: We have unit tests for all of the latest release version methods in this repository and I noticed this addition hasn't added any tests.

I think I would add a unit test in for consistency but also because these methods tend to be quite heavily relied upon.

In another comment I've suggested you could move this to `ReleaseVersionPredicates` but just noticed that class itself doesn't have it's own tests. I guess if they end up being tested indirectly via another method then it's ok to omit the tests.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/ReleaseVersionRepository.cs
mmyoungman: I've gone with a nested query in the hopes of making it more readable - happy to switch to the join if you think that's preferable.

I've also made a change to `ListLatestReleaseVersions` to use `ReleaseVersionPredicates.LatestReleaseVersions` so we don't need to inject `IReleaseVersionRepository` into `PublicationService`.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Predicates/ReleaseVersionPredicates.cs
```diff
@@ -47,4 +47,26 @@ public static class ReleaseVersionPredicates
...
+
+    /// <summary>
+    /// Filters a sequence of <see cref="IQueryable{T}"/> of type <see cref="ReleaseVersion"/> to only include the latest
+    /// published versions of each release.
+    /// </summary>
+    /// <param name="releaseVersionsQueryable">The source <see cref="IQueryable{T}"/> of type <see cref="ReleaseVersion"/> to filter.</param>
+    /// <param name="releaseId">Unique identifier of a release to filter by.</param>
+    /// <param name="publishedOnly">Flag to only include published release versions.</param>
+    /// <returns>An <see cref="IQueryable{T}"/> of type <see cref="ReleaseVersion"/> that contains elements from the input
+    /// sequence filtered to only include the latest versions of each release.</returns>
```
benoutram, line 59: The `summary` and `returns` descriptions aren't consistent with the method which can only return at most one release version.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Predicates/ReleaseVersionPredicates.cs
```diff
@@ -47,4 +47,26 @@ public static class ReleaseVersionPredicates
...
+        Guid releaseId,
+        bool publishedOnly = false)
+    {
+        return releaseVersionsQueryable
+            .Where(releaseVersion => releaseVersion.ReleaseId == releaseId)
+            .Where(releaseVersion => releaseVersion.Version == releaseVersionsQueryable
+                .Where(latestVersion => latestVersion.ReleaseId == releaseId)
+                .Where(latestVersion => !publishedOnly || latestVersion.Published.HasValue)
+                .Select(latestVersion => (int?)latestVersion.Version)
+                .Max());
```
benoutram, line 70: I've not seen this nesting before, I'm assuming this is working in both EF core in-memory  db for unit tests and for real with SQLServer?

Do you know what the difference is in the generated query with this nesting vs. if this was done as a join?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -462,17 +463,16 @@ await supersededPublications
                         }
                         else
                         {
-                            var latestVersion = await _releaseVersionRepository
-                                .GetLatestReleaseVersionForParent(
-                                    seriesItem.ReleaseId!.Value,
-                                    publishedOnly: true);
+                            var latestVersion = await _context.ReleaseVersions
+                                .LatestReleaseVersion(seriesItem.ReleaseId!.Value, publishedOnly: true)
+                                .SingleOrDefaultAsync();
```
benoutram, line 468: I think a comment above this like `// prefer getting the latest published version over a more recent draft version if one exists` might help what's going off here when read in context together with the block below `// if no published version, look for an unpublished version`.

I was a bit confused at first, wondering why we don't just get the latest release version.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/ReleaseVersionRepository.cs
benoutram: > I've also made a change to ListLatestReleaseVersions to use ReleaseVersionPredicates.LatestReleaseVersions so we don't need to inject IReleaseVersionRepository into PublicationService.

I think this is ok but is there much of an advantage to doing this?

`ReleaseVersionPredicates.LatestReleaseVersion` returns an unordered IQueryable so you've had to duplicate the logic for the reverse chronological ordering here.

I recently made an effort to tidy up as many of the places as possible where we were doing the ordering by year and then by time period to be contained within `ReleaseVersionRepository` methods. Except for methods in that repository which id's, they are using the method `OrderByReverseChronologicalOrder` to order results consistently. That logic is defined in a single place.

I think there's a couple of instances (MappingProfiles and MethodologyService) which were not easy to change. I'd have to revisit those to explain why.

Maybe you could reuse that logic here, or swap it back to how it was?

I feel like keeping the ordering logic in as fewer places as possible is a good idea, especially as we may swap it out to order by the release series at some point soon.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Predicates/ReleaseVersionPredicates.cs
mmyoungman: I believe we've used subqueries like this elsewhere. Tough to find documentation for it though: https://visualstudiomagazine.com/articles/2018/10/01/subqueries-in-linq.aspx

I double checked that the generated query was nested:
```
dbug: Microsoft.EntityFrameworkCore.Database.Command[20100]
      Executing DbCommand [Parameters=[@__releaseId_0='bf3167d4-6a81-4b5e-bab7-08dc554c3456'], CommandType='Text', CommandTimeout='30']
      SELECT TOP(2) [r].[Id], [r].[ApprovalStatus], [r].[Created], [r].[CreatedById], [r].[DataGuidance], [r].[NextReleaseDate], [r].[NotifiedOn], [r].[NotifySubscribers], [r].[PreReleaseAccessList], [r].[PreviousVersionId], [r].[PublicationId], [r].[PublishScheduled], [r].[Published], [r].[RelatedInformation], [r].[ReleaseId], [r].[ReleaseName], [r].[Slug], [r].[SoftDeleted], [r].[TimePeriodCoverage], [r].[Type], [r].[UpdatePublishedDate], [r].[Version]
      FROM [ReleaseVersions] AS [r]
      WHERE [r].[SoftDeleted] = CAST(0 AS bit) AND [r].[ReleaseId] = @__releaseId_0 AND [r].[Version] = (
          SELECT MAX([r0].[Version])
          FROM [ReleaseVersions] AS [r0]
          WHERE [r0].[SoftDeleted] = CAST(0 AS bit) AND [r0].[ReleaseId] = @__releaseId_0 AND [r0].[Published] IS NOT NULL)
```

Interestingly, in the article I linked above from 2018, it seemed while you could do subqueries in linq at that time, the article suggests at the end that it may have been generating multiple queries behind the scenes. Seems they've actually generated nested queries now.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Repository/ReleaseVersionRepository.cs
mmyoungman: I've switched back to the original

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Predicates/ReleaseVersionPredicates.cs
benoutram: Cool, I'm not sure what the difference in performance between the join and the nested query would be, but this seems fine.

Just thinking about this reminded me that as an alternative to adding this new method, calls to this should have also been possible by just making this instead I think? 🙂 

```
_contentDbContext.ReleaseVersions
.LatestReleaseVersions()
.FirstOrDefaultAsync(version => version.ReleaseId == releaseId)
```

Not saying you switch to this but if you have any time it would be interesting to see what this would have generated. As it's all in SQL rather than in-memory it should still only fetch at most one result back from the db.

PR #4703, opened 2024-03-26
EES-5029 Update 'sort' parameters in API requests to be consistent across project

ntsim:
This PR updates any 'sort' or 'order' parameters in API requests to be consistent across the entire project. These have been consolidated to simply:

- `sort` - the field to sort by
- `sortDirection` - the direction to sort in

Appropriate changes have been made across the frontend and backend.


src/GovUk.Education.ExploreEducationStatistics.Content.Requests/DataSetsListRequestSortBy.cs
```diff
@@ -1,6 +1,6 @@
 ﻿namespace GovUk.Education.ExploreEducationStatistics.Content.Requests;
 
-public enum DataSetsListRequestOrderBy
+public enum DataSetsListRequestSortBy
```
benoutram, line 3: Do you think we should also change the terminology in the XML documentation below, changing 'Order by' to 'Sort by'?

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
```diff
@@ -1537,7 +1537,7 @@ public async Task Search_SortByRelevance_Desc()
                 var pagedResult = (await service.ListPublications(
                     search: "term",
                     sort: null, // Sort should default to relevance
-                    order: null // Order should default to descending
+                    sortDirection: null // Order should default to descending
```
benoutram, line 1540: Code comments like this one could be changed to `Direction should default to descending`. This applies to multiple places so worth checking elsewhere.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/DataSetService.cs
```diff
@@ -176,28 +176,28 @@ internal static class FreeTextReleaseFileValueResultQueryableExtensions
 {
     internal static IOrderedQueryable<FreeTextValueResult<ReleaseFile>> OrderBy(
         this IQueryable<FreeTextValueResult<ReleaseFile>> query,
-        DataSetsListRequestOrderBy orderBy,
-        SortOrder sort)
+        DataSetsListRequestSortBy sortBy,
```
benoutram, line 179: Shouldn't the name here be `sort` rather than `sortBy`?

It doesn't really matter because it's the internal implementation of the service, but this method could also be renamed from `OrderBy` to `Sort`.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
ntsim: Simplified this to 'Defaults to descending'. Similar thing applied to rest of the comments too

benoutram (approved):

PR #4725, opened 2024-03-28
Ees 4807 small tweaks after merge

duncan-at-hiveit:
This PR:
- changes Container App Environment to be internal rather than having a Public IP address, in preparation for adding an Application Gateway in front, as requested by the Security Architecture group.
- adds in a conditional flag to omit the Container App from deployment until a user-assigned managed identity has been created and assigned the AcrPull role.
- swaps out the conditional inclusion of chunks of YAML in the pipeline for different environments (which turned out not to be working!) with "condition" property on Stages based on the current branch.
- removes PSQL VNet integration in favour of enabling public SQL access to whitelisted IP address ranges, as per existing SQL Server setups.  In the future this will be resolved for VNet traffic by EES-5052, by adding a Private Endpoint linking PSQL to the VNet.
- adds additional necessary subnets to the main ARM template's control.
- adds new appsettings to Container App, and SQL connection string secret creation for Admin, Publisher and Data Processor.
- amends the initial migration of the Container App to allow all PSQL db users to have access to future tables as well as initial ones.

## Optimisation flag for PSQL

I've added a flag for effectively skipping the PSQL stage of the deploy, as currently it seems that PSQL attempts to update every single time that we do a Public API Inf deploy, which leads to at least 5 minutes extra deploy time and at worst about 13 minutes extra.  We need to work out what is causing it to redeploy each time even when nothing has changed.  It may be down to the child "databases" and "firewallRules" resources, which could potentially be solved by moving them out into top-level resources in the postgres module rather than as child resources, but haven't investigated yet.

### DIFF
--- a/infrastructure/templates/public-api/main.bicep
+++ b/infrastructure/templates/public-api/main.bicep
@@ -181,6 +181,12 @@ resource postgreSqlPrivateDnsZone 'Microsoft.Network/privateDnsZones@2020-06-01'
 }
 */
 
+var formattedPostgreSqlFirewallRules = map(postgreSqlFirewallRules, rule => {
+  name: replace(rule.name, ' ', '_')
+  startIpAddress: rule.startIpAddress
+  endIpAddress: rule.endIpAddress
+})
+
 // Deploy PostgreSQL Database.
 module postgreSqlServerModule 'components/postgresqlDatabase.bicep' = {
   name: 'postgreSQLDatabaseDeploy'
@@ -202,7 +208,7 @@ module postgreSqlServerModule 'components/postgresqlDatabase.bicep' = {
     */
     // privateDnsZoneId: postgreSqlPrivateDnsZone.id
     // subnetId: vNetModule.outputs.postgreSqlSubnetRef
-    firewallRules: concat(postgreSqlFirewallRules, [
+    firewallRules: concat(formattedPostgreSqlFirewallRules, [
       {
         name: '${resourcePrefix}-ca-${apiContainerAppName}-subnet'
         startIpAddress: vNetModule.outputs.containerAppEnvironmentSubnetStartIpAddress
--- a/infrastructure/templates/public-api/main.bicep
+++ b/infrastructure/templates/public-api/main.bicep
@@ -218,28 +218,7 @@ module postgreSqlServerModule 'components/postgresqlDatabase.bicep' = if (update
     */
     // privateDnsZoneId: postgreSqlPrivateDnsZone.id
     // subnetId: vNetModule.outputs.postgreSqlSubnetRef
-    firewallRules: concat(formattedPostgreSqlFirewallRules, [
-      {
-        name: '${resourcePrefix}-ca-${apiContainerAppName}-subnet'
-        startIpAddress: vNetModule.outputs.containerAppEnvironmentSubnetStartIpAddress
-        endIpAddress: vNetModule.outputs.containerAppEnvironmentSubnetEndIpAddress
-      }
-      {
-        name: '${dataProcessorFunctionAppFullName}-subnet'
-        startIpAddress: vNetModule.outputs.dataProcessorSubnetStartIpAddress
-        endIpAddress: vNetModule.outputs.dataProcessorSubnetEndIpAddress
-      }
-      {
-        name: '${subscription}-as-ees-admin-subnet'
-        startIpAddress: vNetModule.outputs.adminAppServiceSubnetStartIpAddress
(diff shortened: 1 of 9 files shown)


benoutram (approved):

infrastructure/templates/public-api/main.bicep
```diff
@@ -185,32 +187,19 @@ module postgreSqlServerModule 'components/postgresqlDatabase.bicep' = {
...
-resource apiContainerAppManagedIdentityRBAC 'Microsoft.Authorization/roleAssignments@2022-04-01' = {
-  name: guid(resourceGroup().id, apiContainerAppManagedIdentity.id, acrPullRole)
-  properties: {
-    roleDefinitionId: acrPullRole
-    principalId: apiContainerAppManagedIdentity.properties.principalId
-    principalType: 'ServicePrincipal'
-  }
+resource apiContainerAppManagedIdentity 'Microsoft.ManagedIdentity/userAssignedIdentities@2023-01-31' existing = if (deployContainerApp) {
+  name: apiContainerAppManagedIdentityName
 }
```
benoutram, line 199: Could this be inside the containerApp module rather than in main?

infrastructure/templates/public-api/main.bicep
```diff
@@ -238,6 +227,14 @@ module apiContainerAppModule 'components/containerApp.bicep' = if (psqlDbUsersAd
         name: 'ContentApi__Url'
         value: publicUrls!.contentApi
       }
+      {
+        name: 'MiniProfiler__Enabled'
+        value: 'true'
+      }
+      {
+        name: 'ParquetFiles__BasePath'
+        value: 'data/public-api-parquet'
```
benoutram, line 236: I think we need a new ticket creating on configuring this Azure File Share at some point, in order to mount the data directory.

infrastructure/templates/public-api/main.bicep
```diff
@@ -252,9 +249,19 @@ module apiContainerAppModule 'components/containerApp.bicep' = if (psqlDbUsersAd
...
+        // tables in the "public_data" database successfully.
+        name: 'PublisherFunctionAppIdentityName'
+        value: publisherFunctionAppFullName
+      }
     ]
     tagValues: tagValues
   }
+  dependsOn: [
+    postgreSqlServerModule
+  ]
```
benoutram, line 264: Just wondering if this is still required? If deployContainerApp is true and updatePsqlFlexibleServer is false would this work?

duncan-at-hiveit: Yeah, so this is in the situation where `updatePsqlFlexibleServer` is false.  In that case, it's dependent on `postgreSqlServerModule` being complete before continuing - we have to be explicit about the dependency because we're not using any actual values from the `postgreSqlServerModule` output in this definition.

If `updatePsqlFlexibleServer` is true, `postgreSqlServerModule` will be undefined, and any undefined references in the `dependsOn` array get ignored, so this module can just deploy immediately.

infrastructure/templates/public-api/main.bicep
duncan-at-hiveit: I've created https://dfedigital.atlassian.net/browse/EES-5127 to cover this.

infrastructure/templates/public-api/main.bicep
```diff
@@ -56,8 +56,13 @@ param dateProvisioned string = utcNow('u')
...
 
-@description('Have database users been added to PSQL yet for Container App and Function App?')
-param psqlDbUsersAdded bool = true
+@description('Can we deploy the Container App yet?  This is dependent on the user-assigned Managed Identity for the API Container App being created with the AcrPull role, and the database users added to PSQL.')
+param deployContainerApp bool = true
+
+// Note that this has been added temporarily to avoid 10+ minute deploys where it appears that PSQL will redeploy even if no changes exist in
+// this deploy from the previous one.
+@description('Does the PostgreSQL Flexible Server require any updates? False by default to avoid unnecessarily lengthy deploys.')
+param updatePsqlFlexibleServer bool = false
```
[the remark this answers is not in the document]
duncan-at-hiveit: I've created https://dfedigital.atlassian.net/browse/EES-5128 to cover this, and added some TODOs to the codebase to reference it.

infrastructure/templates/public-api/main.bicep
duncan-at-hiveit: I've created https://dfedigital.atlassian.net/browse/EES-5128 to cover this, and added a TODO to the codebase to reference it.

benoutram (approved):

PR #4733, opened 2024-04-03
EES-4857: Add column and CSV logic to Location table for "LAESTAB" code

tomjonesdev:
Statistical data files uploaded via the Admin website contain column information for the "school_laestab" (local authority establishment) code, but this data isn't present in the files downloaded by users of the public website.

This change persists that column of data and allows those users to match the data they take away from the service without needing to find their own lookup from URN (unique reference number) to "LAESTAB" code.

image


src/GovUk.Education.ExploreEducationStatistics.Data.Model/LocationAttribute.cs
```diff
@@ -1,11 +1,11 @@
 #nullable enable
+using GovUk.Education.ExploreEducationStatistics.Common.Model.Data;
 using System;
 using System.Collections.Generic;
-using GovUk.Education.ExploreEducationStatistics.Common.Model.Data;
 
 namespace GovUk.Education.ExploreEducationStatistics.Data.Model;
 
-public abstract record LocationAttribute(string? Code, string? Name)
+public abstract record LocationAttribute(string? Code, string? Name, string? LaEstab = null)
```
duncan-at-hiveit, line 8: Does this need to be pushed down into the generic `LocationAttribute` superclass?  I think that it's only applicable to Schools, and I see that you have it in the `School` record type, so I would've assumed that it could just stay in that subclass.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/DataSetQueryLocation.cs
```diff
@@ -75,7 +75,7 @@ public static DataSetQueryLocation Parse(string location)
                 nameof(DataSetQueryLocationSchool.Urn) =>
                     new DataSetQueryLocationSchool { Urn = value },
 
-                nameof(DataSetQueryLocationSchool) =>
+                nameof(DataSetQueryLocationSchool.LaEstab) =>
```
duncan-at-hiveit, line 78: Haha, how did we get away with this one for so long?

ntsim: This is public API query endpoint work that is currently in progress and has already been fixed on my branch :wink:

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Models/FixedInformationDataFileReader.cs
```diff
@@ -113,6 +113,9 @@ private enum LocationColumn
         [EnumLabelValue("school_name")]
         SchoolName,
 
+        [EnumLabelValue("school_laestab")]
+        SchoolLaEstab,
```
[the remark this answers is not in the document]
tomjonesdev: Are you saying that introducing another column index means that the process which adds new records won't recognise that a particular location exists already, and just add it again?

Aren't duplicates prevented some other way, e.g. by an ID, a name, or some other identifiable code?

PR #4736, opened 2024-04-04
EES-5036 Accomodate BST and GMT changes in PRA frontend

sambiramairelogic:
There's a frontend bug caused by the Pre-Release Access screen not correctly accommodating GMT/BST changes. 

The `start` date of the PRA period comes from the backend as GMT. During winter this will be midnight on the first day of PRA, but during summer this will be 23:00 on the day BEFORE the start of PRA. The frontend needs to convert the GMT time to BST, but only during BST dates. 

Native Javascript dates don't accommodate timezones at all, and neither does the `date-fns` package that we use out of the box. Luckily, there's an extension to it which lets us do this -> https://www.npmjs.com/package/date-fns-tz

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/pre-release/PreReleasePageContainer.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/pre-release/PreReleasePageContainer.tsx
@@ -37,13 +37,13 @@ export const calculatePraPeriodAdvice = (
     'd MMMM yyyy',
   );
   const timeOfPraStart = formatInTimeZone(start, 'Europe/London', 'HH:mm');
-  const dayScheduledForPublish = formatInTimeZone(
+  const dateScheduledForPublish = formatInTimeZone(
     scheduledPublishDate,
     'Europe/London',
     'd MMMM yyyy',
   );
 
-  return `Pre-release access will be available from ${dateOfPraStart} at ${timeOfPraStart} until it is published on ${dayScheduledForPublish}.`;
+  return `Pre-release access will be available from ${dateOfPraStart} at ${timeOfPraStart} until it is published on ${dateScheduledForPublish}.`;
 };
 
 const PreReleasePageContainer = ({
@@ -123,8 +123,6 @@ const PreReleasePageContainer = ({
     }
 
     if (access === 'Before') {
-      // const utcStart = zonedTimeToUtc(start, start.getTimezoneOffset);
-
       return (
         <>
           <h1>Pre-release access is not yet available</h1>
--- a/src/explore-education-statistics-admin/src/pages/release/pre-release/PreReleasePageContainer.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/pre-release/PreReleasePageContainer.tsx
@@ -17,7 +17,7 @@ import LoadingSpinner from '@common/components/LoadingSpinner';
 import NotificationBanner from '@common/components/NotificationBanner';
 import { useErrorControl } from '@common/contexts/ErrorControlContext';
 import useAsyncRetry from '@common/hooks/useAsyncRetry';
-import { utcToZonedTime, format } from 'date-fns-tz';
+import { formatInTimeZone } from 'date-fns-tz';
 import React from 'react';
 import { generatePath } from 'react-router';
 import { Route, RouteComponentProps, Switch } from 'react-router-dom';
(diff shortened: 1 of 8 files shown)


src/explore-education-statistics-admin/src/pages/release/pre-release/PreReleasePageContainer.tsx
```diff
@@ -104,6 +104,18 @@ const PreReleasePageContainer = ({
     }
 
     if (access === 'Before') {
+      const zonedStartDate = utcToZonedTime(start, 'Europe/Berlin');
```
amyb-hiveit, line 107: Why Berlin not London here?

src/explore-education-statistics-admin/src/pages/release/pre-release/PreReleasePageContainer.tsx
```diff
@@ -104,6 +104,18 @@ const PreReleasePageContainer = ({
...
+      const dateOfPraStart = format(
+        utcToZonedTime(zonedStartDate, 'Europe/London'),
```
amyb-hiveit, line 109: Do we need to do `utcToZonedTime` again here?

src/explore-education-statistics-admin/src/pages/release/pre-release/__tests__/PreReleasePageContainer.test.tsx
```diff
@@ -76,9 +76,10 @@ describe('PreReleasePageContainer', () => {
       }),
     ).toBeInTheDocument();
 
+    // Need to match both 09:00 and 10:00, because mocking out the "London/Europe" timezone for BST/GMT testing isn't really feasible
```
amyb-hiveit, line 79: We mock date-fns here https://github.com/dfe-analytical-services/explore-education-statistics/blob/dev/src/explore-education-statistics-common/test/mockDate.ts, could it be extended or something similar done to mock the timezone stuff?

src/explore-education-statistics-admin/src/pages/release/pre-release/PreReleasePageContainer.tsx
sambiramairelogic: Because I forgot to remove that line whilst playing around with this func during development 🙈 Will remove!

src/explore-education-statistics-admin/src/pages/release/pre-release/PreReleasePageContainer.tsx
sambiramairelogic: We absolutely do not, no :) Will remove!

src/explore-education-statistics-admin/src/pages/release/pre-release/__tests__/PreReleasePageContainer.test.tsx
sambiramairelogic: That's a good shout, I'll have a go. What I'm tying to avoid here is this test suddenly failing in 6 months time when we enter GMT, so whatever mock we write would need to know when the timezones change. Not sure if that's the same date each year, so will go research

tests/robot-tests/tests/admin/bau/prerelease.robot
```diff
@@ -141,8 +141,8 @@ Validate prerelease has not started
     user checks nth breadcrumb contains    1    Home
     user checks nth breadcrumb contains    2    Pre-release access
 
-    ${tomorrow}=    get current datetime    %Y-%m-%dT00:00:00    1
-    ${day_after_tomorrow}=    get current datetime    %Y-%m-%dT%H:%M:%S    2
+    ${tomorrow}=    get current datetime    %Y-%m-%dT00:00:00    1    Europe/London
+    ${day_after_tomorrow}=    get current datetime    %Y-%m-%dT%H:%M:%S    2    Europe/London
 
     ${time_start}=    format uk to local datetime    ${tomorrow}    %-d %B %Y at %H:%M
     ${time_end}=    format uk to local datetime    ${day_after_tomorrow}    %-d %B %Y
```
mmyoungman, line 148: I am not sure we need this here. It looks like L147-8 already handle the translation?

sambiramairelogic: I wondered the same myself, but I --think-- it's safer like this. My reasoning is this: 

PRA Start comes from the backend as Utc, and the cron-job is adjusted twice a year to match BST/GMT. So on the Pre-Prod environment during summer this will be 23:00 on the day before, but during winter it will be 00:00 on the day of. 
Because we're calculating the date only in the line `${tomorrow}=    get current datetime    %Y-%m-%dT00:00:00`, this means that in summer we'll get two different dates if we use Utc or BST. 

Also, the test is currently the test is failing with the error "Page should have contained text Pre-release access will be available from 2 April 2024 at 23:00", but of course that's wrong. It's expecting 23:00, but it should be expecting 00:00 on 3 April. 

But I might have made some errors of logic somewhere along those lines!

mmyoungman (approved):

PR #4739, opened 2024-04-08
EES-5056 Adding relationship from the Content DB's Files table to the Public API DB's DataSetVersion table

jack-hive:
This ticket is to implement the database relationship so that CSV Datasets and Public API datasets can be linked - and then used across the Public API workstream.

The proposed solution is to add a new `PublicDataSetVersionId` column to the `Files` table in the Content DB. This links to the Id primary key in the `DataSetVersions` table in the Public API DB.

The `DataSetVersions` table in the Public API DB currently already has the inverse relationship in the form of a `CsvFileId` column which corresponds to the Id primary key column of the `Files` table in the Content DB.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Model/File.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Model/File.cs
@@ -22,6 +22,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Model
 
         public int? DataSetFileVersion { get; set; }
 
+        public Guid? PublicDataSetVersionId { get; set; }
+
         public Guid? ReplacedById { get; set; }
 
         public File? ReplacedBy { get; set; }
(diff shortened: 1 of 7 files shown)


benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Model/File.cs
```diff
@@ -22,6 +22,8 @@ public class File : ICreatedTimestamp<DateTime?>
 
         public int? DataSetFileVersion { get; set; }
 
+        public Guid? PublicDataSetVersionId { get; set; }
```
benoutram, line 25: I think it would be a good idea to add an index to this field by changing ContentDbContext with something like `entity.HasIndex(e => e.PublicDataSetVersionId)`;

jack-hive: Yes, it 100% would be! I intended on doing this and forgot.

Added it now :)

benoutram (approved):

PR #4740, opened 2024-04-08
EES-5015 Add tests for data set GET query endpoint

ntsim:
This PR adds test code for the data set GET query endpoint. This primarily covers `DataSetQueryParser` (which generates WHERE fragment SQL) and the endpoint itself in `DataSetController.QueryDataSetGet`.

## Related changes

- Updated various test assertions for FluentValidation errors to assert on more specific details for comparators.
- Added various test extensions for testing location, time period and sort strings.
- Reworked `ParquetPathResolver` to use default interface implementations. This makes it easier to create implementation for test projects where we need more control over the paths. In our case, we've implemented a `TestParquetPathResolver` that demonstrates this.

  Note that the previous utils and extension methods for working with Parquet paths have all been  removed in favour of just using `IParquetPathResolver`.


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/ViewModels/ValidationProblemViewModelTestExtensions.cs
```diff
@@ -0,0 +1,244 @@
...
+    {
+        var error = validationProblem.AssertHasError(
+            expectedPath: expectedPath,
+            expectedCode: ValidationMessages.TimePeriodFormat.Code
+        );
+
+        var errorDetail = error.GetDetail<FormatErrorDetail>();
+
+        Assert.Equal(value, errorDetail.Value);
+        Assert.Equal("{period}|{code}", errorDetail.ExpectedFormat);
```
jack-hive, line 206: I'd potentially move this check into the tests in `TimePeriodStringValidatorsTests`. And then wouldn't bother checking it here. Then you have tests covering the fact that this `ExpectedFormat` is generated as expected, and this extension method is just concerned with the value.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/ViewModels/ValidationProblemViewModelTestExtensions.cs
```diff
@@ -0,0 +1,244 @@
...
+    {
+        var error = validationProblem.AssertHasError(
+            expectedPath: expectedPath,
+            expectedCode: ValidationMessages.SortDirection.Code
+        );
+
+        var errorDetail = error.GetDetail<SortStringValidators.DirectionErrorDetail>();
+
+        Assert.Equal(direction, errorDetail.Value.Direction);
+        Assert.Equal(["Asc", "Desc"], errorDetail.AllowedDirections);
```
jack-hive, line 154: Same here, I'd move this check to `SortStringValidatorsTests` and only check the `direction` here

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/ViewModels/ValidationProblemViewModelTestExtensions.cs
```diff
@@ -0,0 +1,244 @@
...
+    {
+        var error = validationProblem.AssertHasError(
+            expectedPath: expectedPath,
+            expectedCode: ValidationMessages.SortFormat.Code
+        );
+
+        var errorDetail = error.GetDetail<FormatErrorDetail>();
+
+        Assert.Equal(value, errorDetail.Value);
+        Assert.Equal("{field}|{order}", errorDetail.ExpectedFormat);
```
jack-hive, line 117: Same here, I'd move this to `SortStringValidatorsTests`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/ViewModels/ValidationProblemViewModelTestExtensions.cs
```diff
@@ -0,0 +1,244 @@
...
+    {
+        var error = validationProblem.AssertHasError(
+            expectedPath: expectedPath,
+            expectedCode: ValidationMessages.LocationAllowedLevel.Code
+        );
+
+        var errorDetail = error.GetDetail<LocationStringValidators.AllowedLevelErrorDetail>();
+
+        Assert.Equal(level, errorDetail.Value.Level);
+        Assert.Equal(EnumUtil.GetEnumValues<GeographicLevel>().Order(), errorDetail.AllowedLevels);
```
jack-hive, line 61: Same here, I'd move this to `LocationStringValidatorsTests`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/ViewModels/ValidationProblemViewModelTestExtensions.cs
```diff
@@ -0,0 +1,244 @@
...
+    {
+        var error = validationProblem.AssertHasError(
+            expectedPath: expectedPath,
+            expectedCode: ValidationMessages.LocationFormat.Code
+        );
+
+        var errorDetail = error.GetDetail<FormatErrorDetail>();
+
+        Assert.Equal(value, errorDetail.Value);
+        Assert.Equal("{level}|{property}|{value}", errorDetail.ExpectedFormat);
```
jack-hive, line 43: Same here, I'd move this to `LocationStringValidatorsTests`

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetsControllerQueryTests.cs
```diff
@@ -0,0 +1,2093 @@
...
+                        path, new StringValues(
+                            [
+                                "",
+                                "invalid",
+                                "||",
+                                "LADD|code|12345",
+                                "NATT|code|12345",
+                                "NAT|invalid|12345",
+                                "LA|urn|12345",
+                                "SCH|code|12345",
```
jack-hive, line 1601: Might be nice to add some invalid properties for `PROV` and `RSC` too, just for completeness

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Validators/TimePeriodStringValidators.cs
```diff
@@ -14,8 +14,13 @@ public static partial class TimePeriodStringValidators
 
     public static IRuleBuilderOptionsConditions<T, string> TimePeriodString<T>(this IRuleBuilder<T, string> rule)
     {
-        return rule.Custom((value, context) =>
+        return rule.NotEmpty().Custom((value, context) =>
         {
+            if (value.IsNullOrWhitespace())
+            {
+                return;
+            }
```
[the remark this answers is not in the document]
ntsim: This is to prevent FluentValidation's somewhat annoying default cascading behaviour. Without this, if the user provides an empty string in the query parameter, it gets mapped to a null in the request model.

The null will fail the `NotEmpty` validation, but cascading behaviour will mean that the null still comes through to our custom validator. When the null hits the format validation, it then throws an exception, resulting in a 500 response.

It's possible to change FluentValidation's cascading behaviour, but this isn't very consistent when combined with `RuleForEach`. Unfortunately, this stops validating on the first item that has an error in the list, rather than report errors for every item in the list.

It's also annoying to have to remember to set the cascade behaviour manually. Consequently, I've opted for this approach as it avoids all the above issues.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetsControllerQueryTests.cs
```diff
@@ -0,0 +1,2093 @@
...
+                        Assert.InRange(int.Parse(kv.Value), 0, 10_000_000);
+                        break;
+                    case "sess_unauthorised":
+                        Assert.InRange(int.Parse(kv.Value), 0, 500_000);
+                        break;
+                    case "sess_unauthorised_percent":
+                        Assert.InRange(float.Parse(kv.Value), 0, 15);
+                        break;
+                }
+            }
```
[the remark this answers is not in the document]
ntsim: Yep fair points. I've updated this to check for explicit min / max values for each indicator

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Validators/TimePeriodStringValidators.cs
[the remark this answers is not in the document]
jack-hive: I see. That makes sense

jack-hive (approved):

PR #4741, opened 2024-04-10
EES-4704: Rewrite existing Slack Integrations from webhooks to Slack Apps

tomjonesdev:
Various webhooks were implemented by a colleague who has since left the team. Deactivating their Slack account causes the integrations to be removed, thus losing the various features integrated by custom scripts and third party apps. The independent "Incoming Webhooks" Slack app is also deprecated, and has been superseded by webhooks and APIs which can be implemented by a custom Slack workspace app. This method also allows additional "collaborators" to view and edit the app's config, and reinstall to the workspace when necessary, without the reliance on a single user account.

The changes replace existing "Incoming Webook" and custom app behaviour with a single custom app, currently named EESBot".

The options considered during development were:

1. Custom hand-written API, implementing the Slack API to funnel all messages through
**Pro**: full control of message content
**Con**: excessive amount of work compared with alternatives (i.e. an additional deployment and codebase to maintain), most of the existing implementations only needed a simple message posting (which can just be done with a webhook), potential difficulties integrating 3rd party app notifications

2. An app for each purpose/integration
**Pro**: would be clear from the app name/description what the purpose/source of the notifications were
**Con**: multiple apps to configure and maintain. New integrations would require a new app creating/configuring alongside the task, rather than just adding another webhook to the existing app

3. A single app to funnels all integrations through 👈
**Pro**: easier to configure and maintain a single, central app
**Con**: some third party integrations post a pre-defined message which can't be customised, so it's not always clear where the messages originated from

### DIFF
--- a/infrastructure/templates/datafactory/template.json
+++ b/infrastructure/templates/datafactory/template.json
@@ -47,12 +47,6 @@
     "actionGroupAlerts": {
       "type": "string"
     },
-    "slackAppToken": {
-      "type": "string",
-      "metadata": {
-        "description": "Slack bot user OAuth token for posting alerts"
-      }
-    },
     "slackAlertsChannel": {
       "type": "string",
       "metadata": {
@@ -61,7 +55,8 @@
     }
   },
   "variables":{
-    "templateBaseUrl": "[concat('https://raw.githubusercontent.com/dfe-analytical-services/explore-education-statistics/', parameters('branch'), '/infrastructure/templates/')]"
+    "templateBaseUrl": "[concat('https://raw.githubusercontent.com/dfe-analytical-services/explore-education-statistics/', parameters('branch'), '/infrastructure/templates/')]",
+    "ees-alerts-slackapptoken": "[resourceId('Microsoft.KeyVault/vaults/secrets', variables('keyVaultName'), 'ees-alerts-slackapptoken')]"
   },
   "resources":[
     {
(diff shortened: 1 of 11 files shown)


tests/robot-tests/tests/libs/slack.py
```diff
@@ -17,14 +15,10 @@
 class SlackService:
     def __init__(self):
         self.slack_bot_token = os.getenv("SLACK_BOT_TOKEN")
-        self.report_webhook_url = os.getenv("SLACK_TEST_REPORT_WEBHOOK_URL")
```
mmyoungman, line 20: Looks like this variable can be removed after this has been deployed? If you haven't already, I'd consider noting this down along with any other post-deploy tasks.

mmyoungman: I think removing `EES UI tests` from Slack apps is another post-deploy task.

infrastructure/templates/datafactory/template.json
```diff
@@ -47,10 +47,10 @@
     "actionGroupAlerts": {
       "type": "string"
     },
-    "slackWebhook": {
+    "slackAppToken": {
```
mmyoungman, line 50: I saw Ben was suggesting also fetching this variable from keyvault directly in this template rather than using template params, by adding something similar to this to the `variables` section:

`"ees-admin-govuknotify-api-key": "[resourceId('Microsoft.KeyVault/vaults/secrets', variables('keyVaultName'), 'ees-admin-govuknotify-api-key')]",`

mmyoungman (changes requested):

tests/robot-tests/tests/libs/slack.py
tomjonesdev: Both on my list 👍

mmyoungman (approved):

PR #4745, opened 2024-04-11
EES-4908 Add Admin endpoint for listing API datasets for a publication

benoutram:
This PR adds a new Admin endpoint for listing the available API datasets for a publication.

`GET /api/publications/{publicationId}/data-sets`

## Pagination
The endpoint offers query parameters `page` and `pageSize` to return a paginated response using offset pagination defaulting to a page size of 10.

## Ordering
The endpoint returns datasets in descending order of their latest live version's published date (if any), and then by ascending title order.

# Other changes
- Configures Admin to use FluentValidation.
- Adds new `LatestDraftVersionId` / `LatestDraftVersion` pointers to `DataSet`.
- Renames `DataSet.LatestVersion` to `LatestLiveVersion`.
- Corrects the `LatestLiveVersion` relationship between `DataSet` and `DataSetVersion` from one-to-one to one-to-zero-or-one, preventing the relationship from automatically being set which was the case even for non-live versions.
- Converts Admin `Startup` to use primary constructor.
- Converts Admin `PublicationController` to use primary constructor.
- Removes Admin `PublicationControllerTests` GetRoles unit tests which were not providing value.
- Rewrites remainder of existing Admin `PublicationControllerTests` from unit tests to integration tests.
- Adds dependency between Admin and Public.Data.Model to provide and configure `PublicDataDbContext`.
- Adds new `IntegrationTestFixture` and `TestApplicationFactory` to Admin to support integration testing with `PublicDataDbContext`. This uses  PostgreSQL [Testcontainer](https://dotnet.testcontainers.org/modules/postgres/), similar to how it's used in the Public API.
- Squashes `PublicDataDbContext` migrations into new `20240408074415_InitialMigration`, removing `EES4807_AddPermissionsToDataProcessorUser`.
- Fixes Public API unit test `DataSetsControllerTests.GetDataSetTests.DataSetIsAvailable_Returns200` which was missing setup data `FinishWith(dsv => dataSet.LatestLiveVersion = dsv)`.
- Adds missing `Changes` column in table `ChangeSetIndicators`. The code `Changes = table.Column<string>(type: "jsonb", nullable: true),` was missing from `InitialMigration`.

# Configuration of `PublicDataDbContext` in Admin

This PR has added a new Admin environment connection string variable `PublicDataDb` which depends on a new Key Vault secret `ees-admin-connectionstring-publicdatadb`.

This connection string contains an access token parameter which will be replaced by a real access token. This depends on the Admin having a user managed identity setup to access the public data PostgreSQL database. The remainder of this infrastructure work will be carried out by [EES-5049](https://dfedigital.atlassian.net/browse/EES-5049).

# New database seed data

After the database column rename and new initial migration there is a new seed data database `public-api-7.zip`.


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
```diff
@@ -0,0 +1,119 @@
...
+    private static DataSetViewModel MapDataSet(DataSet dataSet)
+    {
+        var latestDraftVersion = dataSet.Versions
+            .Where(version => version.Status is DataSetVersionStatus.Draft
+                or DataSetVersionStatus.Failed
+                or DataSetVersionStatus.Mapping
+                or DataSetVersionStatus.Processing)
+            .OrderBy(version => version.VersionMajor)
+            .ThenBy(version => version.VersionMinor)
+            .LastOrDefault();
```
ntsim, line 71: It's a bit inefficient to have to go through every single version to figure out what the latest draft is.

It's also possible we might end up with lots of data sets and lots of versions for those data sets, so this approach isn't super scaleable. In fact, it's basically the same type of problem we had / have with releases.

Would suggest that rather than do this, we add `LatestDraftVersion` / `LatestDraftVersionId` to `DataSetVersion` so that we have pointers to these in the database.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
```diff
@@ -0,0 +1,119 @@
...
+                Number = dataSetVersion.Version,
+                Published = dataSetVersion.Published!.Value,
+                Status = dataSetVersion.Status,
+                Type = dataSetVersion.VersionType,
+            }
+            : null;
+    }
+
+    private async Task<Either<ActionResult, Publication>> CheckPublicationExists(Guid publicationId,
+        CancellationToken cancellationToken = default)
```
ntsim, line 113: No need to provide a default here as we'll always be guaranteed to be passing a `CancellationToken` through from the public method

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/Public.Data/IDataSetService.cs
```diff
@@ -0,0 +1,18 @@
+using System;
```
ntsim, line 1: `#nullable enable`

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/Public.Data/DataSetListRequest.cs
```diff
@@ -0,0 +1,27 @@
+using FluentValidation;
```
ntsim, line 1: `#nullable enable`

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
```diff
@@ -1,10 +1,14 @@
 using System;
```
ntsim, line 1: `#nullable enable`?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Fixture/IntegrationTestFixture.cs
```diff
@@ -0,0 +1,28 @@
+using System.Threading.Tasks;
```
ntsim, line 1: `#nullable enable`

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/DataSetViewModels.cs
```diff
@@ -0,0 +1,87 @@
...
+    /// A summary of the data set’s contents.
+    /// </summary>
+    public required string Summary { get; init; }
+
+    /// <summary>
+    /// The status of the data set. Can be one of the following:
+    ///
+    /// - `Published` - the data set has been published and will receive updates
+    /// - `Deprecated` - the data set is being discontinued and will no receive updates
+    /// - `Withdrawn` - the data set has been withdrawn and can no longer be used
```
ntsim, line 32: This doesn't provide the full set of statuses. Would suggest that we remove them from here completely as they'll easily get out of sync (if we add any more statuses) and we don't need this level of detail as we don't really use the Swagger docs for the admin.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/DataSetViewModels.cs
```diff
@@ -0,0 +1,87 @@
+#nullable enable
+using System;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.ViewModels.Public.Data;
+
+/// <summary>
+/// Provides high-level information about a data set.
+/// </summary>
```
ntsim, line 9: We can keep all the docblocks in if you think they provide value, but we might just want to get rid of any that are self-explanatory in this view model and the ones below.

The public API view models need to be fully documented as these are used in Swagger doc generation. However, this is somewhat overkill for the admin, especially as we've never been in the habit of documenting our view models / requests in any great detail.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/TimePeriodRangeViewModel.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace GovUk.Education.ExploreEducationStatistics.Admin.ViewModels.Public.Data;
+
+/// <summary>
+/// Describes a time period range in human-readable format.
+/// </summary>
```
ntsim, line 5: Similar to `DataSetViewModels`, do we want to keep these kind of self-explanatory docblocks?

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/PublicationController.cs
```diff
@@ -153,17 +150,39 @@ public async Task<ActionResult<List<ReleaseSeriesTableEntryViewModel>>> GetRelea
...
                 .GetPublicationRolesForPublication(publicationId)
                 .HandleFailuresOrOk();
         }
+
+        /// <summary>
+        /// List a publication’s data sets
+        /// </summary>
+        /// <remarks>
+        /// Lists summary details of all the data sets related to a publication.
+        /// </remarks>
```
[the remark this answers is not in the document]
benoutram: I've removed this docblock now

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/PublicationControllerTests.cs
```diff
@@ -1,114 +1,745 @@
...
-                    })
-                )
+            await TestApp.AddTestData<ContentDbContext>(context =>
+            {
+                context.Topics.Add(topic);
+            });
+
+            var client = TestApp
+                .SetUser(BauUser())
+                .CreateClient();
```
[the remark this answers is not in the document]
benoutram: It was in there intentionally as the user could change inbetween tests, like it does in the `ListPublicationDataSetsTests`. 

I've set it now as the default client in `CreatePublication` now with an optional `HttpClient? client = null` parameter that allows an alternative client to be passed.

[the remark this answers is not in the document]
benoutram: Also added new test `UserHasNoAccessToCreatePublication_ReturnsForbidden`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
benoutram: I've added this to `DataSet` now, squashed the migration and regenerated `public-api-7.zip`.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/DataSetViewModels.cs
benoutram: I've left the high-level record type summaries but removed all the rest. I wouldn't mind getting rid of absolutely everything if you think it's not necessary.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/PublicationControllerTests.cs
```diff
@@ -1,114 +1,745 @@
...
+                    s.SetPublished(DateTimeOffset.UtcNow.AddDays(-2));
+                })
+                .ForIndex(6, s =>
+                {
+                    s.SetDataSet(dataSets[5]);
+                    s.SetStatusDraft();
+                })
+                .FinishWith(dsv =>
+                {
+                    if (dsv.Status is DataSetVersionStatus.Published or DataSetVersionStatus.Withdrawn or DataSetVersionStatus.Deprecated)
```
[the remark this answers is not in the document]
benoutram: Cool I've fixed this now. Off topic but just checking you are aware this new Admin endpoint isn't returning the entire version history currently?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/PublicationControllerTests.cs
```diff
@@ -1,114 +1,745 @@
...
+                    indicators: 1,
+                    locations: 1,
+                    timePeriods: 2)
+                .WithDataSet(dataSet)
+                .ForIndex(0, s => s.SetVersionNumber(major: 1, minor: 1))
+                .ForIndex(1, s => s.SetVersionNumber(major: 1, minor: 0))
+                .ForIndex(2, s => s.SetVersionNumber(major: 3, minor: 1))
+                .ForIndex(3, s => s.SetVersionNumber(major: 3, minor: 0))
+                .ForIndex(4, s => s.SetVersionNumber(major: 2, minor: 1))
+                .ForIndex(5, s => s.SetVersionNumber(major: 2, minor: 0))
```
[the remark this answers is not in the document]
benoutram: Given the suggestion to add in a `LatestDraftVersionId` / `LatestDraftVersion`, combined with the fact this endpoint doesn't return all version history currently, I don't think we need this test `PublicationHasSingleDataSetWithMultipleVersions_CorrectDraftVersion` anymore?

I've removed it.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/PublicationControllerTests.cs
```diff
@@ -1,114 +1,745 @@
...
+            Assert.Equal(draftDataSetVersion.Status, dataSetViewModel.DraftVersion.Status);
+            Assert.Equal(draftDataSetVersion.VersionType, dataSetViewModel.DraftVersion.Type);
+
+            Assert.NotNull(dataSetViewModel.LatestLiveVersion);
+            Assert.Equal(liveDataSetVersion.Id, dataSetViewModel.LatestLiveVersion.Id);
+            Assert.Equal(liveDataSetVersion.Version, dataSetViewModel.LatestLiveVersion.Number);
+            Assert.Equal(liveDataSetVersion.Status, dataSetViewModel.LatestLiveVersion.Status);
+            Assert.Equal(liveDataSetVersion.VersionType, dataSetViewModel.LatestLiveVersion.Type);
+            Assert.Equal(liveDataSetVersion.Published?.ToUnixTimeSeconds(),
+                dataSetViewModel.LatestLiveVersion?.Published.ToUnixTimeSeconds());
```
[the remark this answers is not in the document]
benoutram: I've changed this to use `Assert.Multiple` now.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/PublicationControllerTests.cs
```diff
@@ -1,114 +1,745 @@
...
+
+            await TestApp.AddTestData<ContentDbContext>(context => context.Publications.Add(publication));
+            await TestApp.AddTestData<PublicDataDbContext>(context => context.DataSets.AddRange(dataSets));
+
+            // Data set 0 has 1 live version published 3 days ago (expected position: 3)
+            // Data set 1 has 1 live version published 4 days ago and 1 ignored draft version (expected position: 4)
+            // Data set 2 has 1 live version published 1 day ago (expected position: 1)
+            // Data set 3 has 1 draft version and title Dataset B (expected position: 6)
+            // Data set 4 has 1 live version published 2 days ago (expected position: 2)
+            // Data set 5 has 1 draft version and title Dataset A (expected position: 5)
```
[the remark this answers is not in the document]
benoutram: I've added comments inline below now and added comments inline into the declaration of `expectedDataSetIds` as well.

jack-hive (commented):
Just wondering if we should change all the `...Public.Data...` namespaces to `...PublicData...`

What do people think?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/PublicationControllerTests.cs
```diff
@@ -1,114 +1,745 @@
...
+                .ToList();
+
+            await TestApp.AddTestData<PublicDataDbContext>(context =>
+            {
+                context.DataSetVersions.AddRange(dataSetVersions);
+                context.DataSets.UpdateRange(dataSets);
+            });
+
+            var expectedDataSetIds = dataSets
+                .OrderByDescending(ds => ds.LatestLiveVersion != null ? ds.LatestLiveVersion.Published : DateTimeOffset.MinValue)
```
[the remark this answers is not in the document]
benoutram: Yes it can be, I only did it like that so it would work regardless of the test data set up.

I've simplified it now from 

```csharp
var expectedDataSetIds = dataSets
    .OrderByDescending(ds =>
        ds.LatestLiveVersion != null ? ds.LatestLiveVersion.Published : DateTimeOffset.MinValue)
    .ThenBy(ds => ds.Title)
    .ThenBy(ds => ds.Id)
    .Select(ds => ds.Id)
    .Skip((page - 1) * pageSize)
    .Take(pageSize)
    .ToList();
```

to

```csharp
var expectedDataSetIds = dataSets
    .OrderByDescending(ds => ds.LatestLiveVersion!.Published)
    .Select(ds => ds.Id)
    .Skip((page - 1) * pageSize)
    .Take(pageSize)
    .ToList();
```

benoutram:
> Just wondering if we should change all the `...Public.Data...` namespaces to `...PublicData...`
> 
> What do people think?

Discussed this with @jack-hive and @ntsim away from Github and we are going to stick with the Admin using `Public.Data`, at least for now.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/PublicationControllerTests.cs
```diff
@@ -1,114 +1,745 @@
...
+                    ContactTelNo = "01234567890"
+                },
+                TopicId = topic.Id
+            };
+
+            var response = await CreatePublication(client, request);
+
+            var result = response.AssertOk<PublicationCreateViewModel>();
+
+            Assert.Multiple(
```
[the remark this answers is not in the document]
benoutram: Yes I agree, checking view models seem to be an ideal case for using it.🙂

ntsim (approved):

PR #4751, opened 2024-04-15
EES-5049 Configure Admin's connection to the Public Data PostgreSQL database in Azure environments

benoutram:
This PR configures the Admin's connection to the Public Data PostgreSQL database in Azure environments.

It builds on configuration for the Admin's connection to this database already added by EES-4908 in https://github.com/dfe-analytical-services/explore-education-statistics/pull/4745. This added configuration for the database connection in the local environment.

Connecting in Azure environments depends on the Admin having its own database principal in the Public Data database.
This is a manual step that will happen when the database is available in each environment.

It also relies on the Admin's database principal being granted permissions on tables in the database. Configuration to add this automatically was added by https://github.com/dfe-analytical-services/explore-education-statistics/pull/4750.

When the database is available and the Admin's database principal has been granted permissions in each environment we can enable the connection by changing the appropriate `<env>.parameters.json`.

```json
    "publicDataDbExists": {
      "value": false
    }
```

needs changing to 

```json
    "publicDataDbExists": {
      "value": true
    }
```

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
@@ -107,6 +107,9 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin
         // This method gets called by the runtime. Use this method to add services to the container.
         public virtual void ConfigureServices(IServiceCollection services)
         {
+            // TODO EES-5073 Remove this when the Public Data db exists in ALL Azure environments.
+            var publicDataDbExists = configuration.GetValue<bool>("PublicDataDbExists");
+
             services.AddHealthChecks();
 
             /*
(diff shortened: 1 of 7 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
```diff
@@ -222,19 +222,18 @@ public virtual void ConfigureServices(IServiceCollection services)
...
-                    // });
+                    services.AddDbContext<PublicDataDbContext>(options =>
+                    {
+                        var sqlServerTokenProvider = new DefaultAzureCredential();
+                        var accessToken = sqlServerTokenProvider.GetToken(
+                                new TokenRequestContext(scopes:
+                                    ["https://ossrdbms-aad.database.windows.net/.default"]))
+                            .Token;
+                        var connectionStringWithAccessToken =
+                            publicDataDbConnectionString.Replace("[access_token]", accessToken);
```
duncan-at-hiveit, line 233: We actually need to supply Admin with the Public Data DB connection string in Azure when this goes in, so presumably this change will start to fail Admin starting up when it finds that it doesn't have an appropriate piece of configuration.

Given that Admin is controlled via the ARM template, we'll need to make the change in the ARM template within this PR I think.

We'll need to provide the connection string in the form:

`Server=<subscription>-ees-psql-flexibleserver.postgres.database.azure.com;Database=public_data;Port=5432;User Id=<subscription>-as-ees-admin;Password=[access_token]`

benoutram: This was already done by the ARM template change in https://github.com/dfe-analytical-services/explore-education-statistics/pull/4745/files. 🙂 

The connection string has been added as a key vault secret which contains the connection in the form you posted above.

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
```diff
@@ -107,6 +107,9 @@ public class Startup(IConfiguration configuration, IHostEnvironment hostEnvironm
         // This method gets called by the runtime. Use this method to add services to the container.
         public virtual void ConfigureServices(IServiceCollection services)
         {
+            // TODO EES- Remove this when the Public Data db exists in ALL Azure environments.
+            var publicDataDbExists = configuration.GetValue<bool>("PublicDataDbExists");
```
duncan-at-hiveit, line 111: Does this just need adding to appsettings.Development.json?

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
```diff
@@ -222,19 +222,18 @@ public virtual void ConfigureServices(IServiceCollection services)
...
-                    //     var accessToken = sqlServerTokenProvider.GetToken(
-                    //             new TokenRequestContext(scopes:
-                    //                 ["https://ossrdbms-aad.database.windows.net/.default"]))
-                    //         .Token;
-                    //     var connectionStringWithAccessToken =
-                    //         publicDataDbConnectionString.Replace("[access_token]", accessToken);
-                    //     var dbDataSource = new NpgsqlDataSourceBuilder(connectionStringWithAccessToken).Build();
-                    //     options.UseNpgsql(dbDataSource);
-                    // });
+                    services.AddDbContext<PublicDataDbContext>(options =>
```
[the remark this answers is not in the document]
benoutram: Ok, hadn't thought about the duplication, but I think I will leave this duplicated for now due to the point you make about the common dependency on Azure.Identity. Thanks 🙂

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
benoutram: I debugged this when running locally and it defaults to false when not set, rather than throwing an exception.

I think that makes sense because looking at the source code for `GetValue<bool>` it is `return GetValue(configuration, key, default(T));`

benoutram: Added it in now so that we configure `DataSetService` in the local environment like this.

```csharp
services.AddTransient<IDataSetService, DataSetService>();
```

PR #4755, opened 2024-04-16
Resolve caniuse warnings

sambiramairelogic:
caniuse-warning

We've been getting warnings about an outdated version of `caniuse` each we start the frontend. This PR removes them


amyb-hiveit (approved):
Thanks! I've been meaning to fix this for the last 3 years but never got round to it 🤣

PR #4756, opened 2024-04-16
Resolve GovUk-Frontend warnings

sambiramairelogic:
Each time we start the frontend project we're currently getting warnings produced by the `GovUk Frontend` package because we're using a font value that will be deprecated in the next version. Following the guidance here - https://design-system.service.gov.uk/get-started/new-type-scale/ - this PR updates the font to the next lowest value. 

button-after-change
button-after-change-mobile
govuk-warning

### DIFF
--- a/src/explore-education-statistics-frontend/src/components/FilterClearButton.module.scss
+++ b/src/explore-education-statistics-frontend/src/components/FilterClearButton.module.scss
@@ -23,7 +23,7 @@
 }
 
 .filterType {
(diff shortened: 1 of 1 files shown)


amyb-hiveit (approved):
This makes the new data catalogue filters, where the type is shown, a bit big, but it's under a feature flag so I'm happy to leave it for now and I'll tweak it when I go back to data catalogue work.

PR #4757, opened 2024-04-16
EES-5067 add ref lines between data points on major axis

amyb-hiveit:
Adds the ability to position reference lines between data points on the major axis on vertical bar charts (they can't be added on other types of chart).

Recharts doesn't allow reference lines to be positioned like this: "If the specified x-axis is a category axis, the value of x must be one of the categorys (sic), otherwise no line will be drawn" (https://recharts.org/en-US/api/ReferenceLine), so I've had to be a bit 'creative':

- the reference line is added to the major axis configuration
- when the chart is rendered it:
  - moves the line to the minor axis, so the line is rendered horizontally between the two points
  - in `CustomReferenceLineLabel` it draws a new line perpendicular to the existing one, using the `viewbox`, chart height and margin to calculate its position
  -  hides the existing line by setting its `stroke` to `none`

Screenshot 2024-04-16 102419

Screenshot 2024-04-16 104600

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartAxisConfiguration.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartAxisConfiguration.tsx
@@ -24,6 +24,7 @@ import {
   TickConfig,
 } from '@common/modules/charts/types/chart';
 import { DataSetCategory } from '@common/modules/charts/types/dataSet';
+import { otherAxisPositionTypes } from '@common/modules/charts/types/referenceLinePosition';
 import createDataSetCategories, {
   toChartData,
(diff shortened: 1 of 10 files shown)


sambiramairelogic (approved):
Nice job! I wouldn't have even thought this was possible tbh...

I did notice a few improvements we could make to the validation in the configuration form, but nothing worth addressing in this PR. It's currently possible to add two identical reference lines, which render on top of each other, or to enter 0 as the label pixel width (or indeed something stupid like 100,000). Also, it's possible to get it into a state where it gives validation errors but doesn't say why: 
image

But also the problems are immediately obvious in the preview, and easily editable. Also we're never going to be able to create something totally foolproof with this sort of tool anyway, so the users are always going to be able to ask for something silly.

PR #4774, opened 2024-04-18
Ees 4909 add public api assets to publishing checklist

duncan-at-hiveit:
# UI test run 02/05/2024

image

## Overview

This PR adds a new check into the Release publishing checklist to ensure that there are no Public API data set imports still in progress when attempting to sign off the Release.

As we don't yet have the UI to show Public API data set imports in place, the link that appears in the checklist error will direct the user to the existing data import tab for now.  

## Testing from the front end

To simulate ongoing imports, I created a Release in Admin as per usual and imported a data file.

I then added some dummy data into my local PostgreSQL db to represent an initial DataSetVersion for the ongoing import:

```
INSERT INTO "DataSets"
	("Id", "Title", "Summary", "PublicationId", "Status", "Created")
	VALUES ('c55a3f12-6a4d-4327-8c57-bb0fc4471f46', 'Blah', 'Blah', 'cb07894b-ba45-45e0-8f94-c5de7e3a8191', 'OK', NOW());

-- This DataSetVersion refers to the data file's FileId that is being "uploaded".
INSERT INTO "DataSetVersions"
	("Id", "DataSetId", "Status", "CsvFileId", "Created", "VersionMajor", "VersionMinor", "Notes", "TotalResults", "MetaSummary")
	VALUES ('199c1ee8-3435-4c06-80a8-4d4701d998e9', 'c55a3f12-6a4d-4327-8c57-bb0fc4471f46', 'OK', 'cc4290fa-1e79-4a80-a55c-08dc5f91d76a', NOW(), '1', '1', 'Notes', 0, '{}');
```

and create an import row with a non-complete status (in this case, "Validating"):

```
INSERT INTO "DataSetVersionImports"
	("Id", "DataSetVersionId", "Status", "Created")
	VALUES ('2189bd2c-0fd6-4eda-9937-b507050d42a8', '199c1ee8-3435-4c06-80a8-4d4701d998e9', 'Validating', NOW());
```

and when visiting the Sign off page, the resulting checklist error appears: 

image

### Queued

`UPDATE "DataSetVersionImports" SET "Status" = 'Queued' WHERE "Id" = '2189bd2c-0fd6-4eda-9937-b507050d42a8';`

image

### Validating

`UPDATE "DataSetVersionImports" SET "Status" = 'Validating' WHERE "Id" = '2189bd2c-0fd6-4eda-9937-b507050d42a8';`

image

### Importing

`UPDATE "DataSetVersionImports" SET "Status" = 'Importing' WHERE "Id" = '2189bd2c-0fd6-4eda-9937-b507050d42a8';`

image

### Complete

`UPDATE "DataSetVersionImports" SET "Status" = 'Complete' WHERE "Id" = '2189bd2c-0fd6-4eda-9937-b507050d42a8';`

image

### Cancelled

`UPDATE "DataSetVersionImports" SET "Status" = 'Cancelled' WHERE "Id" = '2189bd2c-0fd6-4eda-9937-b507050d42a8';`

image

### Failed

`UPDATE "DataSetVersionImports" SET "Status" = 'Failed' WHERE "Id" = '2189bd2c-0fd6-4eda-9937-b507050d42a8';`

image


src/GovUk.Education.ExploreEducationStatistics.Admin/Validators/ValidationErrorMessages.cs
```diff
@@ -112,6 +112,7 @@ public enum ValidationErrorMessages
         RelatedDashboardsSectionContainsEmptyHtmlBlock,
         ReleaseMustContainKeyStatOrNonEmptyHeadlineBlock,
         SummarySectionContainsEmptyHtmlBlock,
+        PublicApiDataSetImportsMustBeCompleted,
```
benoutram, line 115: Should we add another checklist item for if the data set version is in the mapping stage still, and not reached 'draft' status yet?

Ultimately I think we only want to allow the approval if all the new data set versions added to this release are in 'draft'

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionImportService.cs
```diff
@@ -0,0 +1,33 @@
...
+            .Where(rf => rf.ReleaseVersionId == releaseVersionId && rf.File.Type == FileType.Data)
+            .Select(rf => rf.FileId)
+            .ToListAsync();
+
+        return await publicDataDbContext
+            .DataSetVersionImports
+            .AsNoTracking()
+            .AnyAsync(import => 
+                dataFileIds.Contains(import.DataSetVersion.CsvFileId) 
+                && !DataSetVersionImportStatusConstants.TerminalStates.ToList().Contains(import.Status));
```
benoutram, line 31: I think in future if we allow an import to be retried for the same dataset version this would not work as we'd have a history of imports for the same version and  only really need to consider the latest import.

I'm wondering if we really need to be concerned with the `DataSetVersionImport` though and be filtering by a list of the terminal states for the import, when `DataSetVersion` has a status where we can just check it's not still `DataSetVersionStatus.Processing` or `DataSetVersionStatus.Failed`.

benoutram (commented):
We've already discussed some changes to the checklist to work from values of  `DataSetVersionStatus` rather than values of `DataSetVersionImportStatus`, so I will leave reviewing the tests until this change has been made.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionImportService.cs
duncan-at-hiveit: Updated to use DataSetVersion.Status.

duncan-at-hiveit:
> We've already discussed some changes to the checklist to work from values of `DataSetVersionStatus` rather than values of `DataSetVersionImportStatus`, so I will leave reviewing the tests until this change has been made.

Updated to use DataSetVersion.Status.

src/GovUk.Education.ExploreEducationStatistics.Admin/Validators/ValidationErrorMessages.cs
benoutram: Cool 👍. I think it's fine to build in the checklist functionality for `Cancelled` statuses right now, but for the first iteration I don't think we will have any way of cancelling a data set version that is processing so for the time being it won't be possible to reach this status.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
```diff
@@ -0,0 +1,43 @@
...
+    : IDataSetVersionService
+{
+    public async Task<List<DataSetVersionStatusSummary>> GetStatusesForReleaseVersion(Guid releaseVersionId)
+    {
+        var releaseFileIds = await contentDbContext
+            .ReleaseFiles
+            .AsNoTracking()
+            .Where(rf => rf.ReleaseVersionId == releaseVersionId && rf.File.Type == FileType.Data)
+            .Select(rf => rf.Id)
+            .ToListAsync();
```
benoutram, line 26: I don't think it hurts to have `AsNoTracking` here but the result doesn't contain any entity types so no tracking would be done anyway.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
```diff
@@ -0,0 +1,43 @@
...
+            .AsNoTracking()
+            .Where(dataSetVersion => releaseFileIds.Contains(dataSetVersion.ReleaseFileId))
+            .Include(dataSetVersion => dataSetVersion.DataSet)
+            .ToListAsync())
+            .Select(dataSetVersion => new DataSetVersionStatusSummary(
+                Id: dataSetVersion.Id,
+                Title: dataSetVersion.DataSet.Title,
+                Status: dataSetVersion.Status)
+            )
+            .ToList();
```
benoutram, line 39: I would have thought you can get away without the `.Include` and the early `ToListAsync` here, moving it to the end and then no need for the additional parentheses?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataSetVersionServiceTests.cs
```diff
@@ -0,0 +1,110 @@
...
+            await GetEnums<DataSetVersionStatus>()
+                .ToAsyncEnumerable()
+                .ForEachAwaitAsync(AssertDataSetVersionStatusReturnedOk);
+        }
+        
+        /// <summary>
+        /// Test the scenario when data set imports are in progress, but for an unrelated ReleaseVersion.
+        /// </summary>
+        [Fact]
+        public async Task DataSetVersionForDifferentReleaseVersion()
```
benoutram, line 39: I think the name of this test class and the doc comments need updating here as they all still refer to import statuses.

I think ideally this test class wouldn't be needed if there were controller integration tests covering the release checklist. Is that what you were thinking too?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
duncan-at-hiveit: Unfortunately we can't, as the Select part is an expression tree and thus not usable in a Linq EF query.  This is why we have to do the actual query first and then the Select into the DataSetVersionStatusSummary afterwards.  If you've gotten examples of where we fetch directly into a DTO / ViewModel elsewhere though, feel free to point me at one and I'll see what we did there!  Thanks.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataSetVersionServiceTests.cs
duncan-at-hiveit: Ahh right you are!  I've updated to reflect the change in name and behaviour now, thanks.

Yes ideally we'd have a controller test but unfortunately I just got hit by time constraints.  Ideally we'd merge this test and the ReleaseChecklistServiceTests into an overarching Controller integration test, but I've blown the time estimate that we had for 4909 already and don't wanna push it any further at the present time, if that makes sense?  I could always raise a ticket to test this all at the Controller level, similar to the Publisher function integration test that is currently testing a Service but really would be better to test from the Function level.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
duncan-at-hiveit: Good spot - I've removed thanks - originally I think I had this returning entities but I can't remember!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
benoutram: I thought that's only the case if the expression tree cannot be translated to SQL by the provider?

In this case I can't see what can't be translated, so I would have thought this code should work?

```csharp
        return await publicDataDbContext
            .DataSetVersions
            .AsNoTracking()
            .Where(dataSetVersion => releaseFileIds.Contains(dataSetVersion.ReleaseFileId))
            .Select(dataSetVersion => new DataSetVersionStatusSummary(
                dataSetVersion.Id,
                dataSetVersion.DataSet.Title,
                dataSetVersion.Status)
            )
            .ToListAsync();
```

`RedirectsService#List`, 'MethodologyService#GetSummariesTree`, and `MethodologyService#GetPublishedPublicationsForMethodology` are some examples of where we are able to do this.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataSetVersionServiceTests.cs
benoutram: Cool, no I think that's all fine. The main thing was to sort out the naming and the comments here. Thanks.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
duncan-at-hiveit: Aha cool thanks.  It was trying to use named parameters that was confusing it.  Thanks for pointing out the examples.  I've switched it to not use explicit constructor param names now and it's happy.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataSetVersionServiceTests.cs
duncan-at-hiveit: Cool ta

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
benoutram: I've tried refreshing but I can't see that change 😆 .

benoutram (approved):

PR #4775, opened 2024-04-18
EES-5062 Move FilterSequence and IndicatorSequence from ReleaseSubject to ReleaseFile

mmyoungman:
This PR moves `ReleaseSubject.FilterSequence` and `IndicatorSequence` into `ReleaseFile`.

This is being done in preparation for introducing `DataSetFileMeta`. Without moving these values first, when listing all data sets on the Data catalogue page, we'd be forced to query both the content and statistics databases.

After this work is deployed, we can then remove `FilterSequence` and `IndicatorSequence` from `ReleaseSubject`.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseAmendmentServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseAmendmentServiceTests.cs
@@ -280,7 +280,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
             {
                 Id = Guid.NewGuid(),
                 Filename = "Filename 1",
-                SubjectId = Guid.NewGuid()
+                SubjectId = Guid.NewGuid(),
             };
 
             var dataFile2 = new File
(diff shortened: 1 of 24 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseAmendmentServiceTests.cs
```diff
@@ -298,7 +298,24 @@ public async Task CreateReleaseAmendment()
                     ReleaseVersion = originalReleaseVersion,
                     ReleaseVersionId = originalReleaseVersion.Id,
                     File = dataFile1,
-                    FileId = dataFile1.Id
+                    FileId = dataFile1.Id,
+                    FilterSequence = new List<FilterSequenceEntry>
```
tomjonesdev, line 302: Assuming `FilterSequence` (et al.) are concrete type properties rather than abstractions, you can shorthand the instantiations to simply use `new()` as the type is always inferred.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/ReleaseFileSequenceMigrationController.cs
```diff
@@ -0,0 +1,66 @@
...
+        var numReleaseSubjectsMigrated = 0;
+
+        foreach (var releaseSubject in releaseSubjects)
+        {
+            var releaseFile = await contentDbContext.ReleaseFiles
+                .Where(rf =>
+                    rf.File.Type == FileType.Data
+                    && rf.File.SubjectId == releaseSubject.SubjectId
+                    && rf.ReleaseVersionId == releaseSubject.ReleaseVersionId)
+                .SingleAsync(cancellationToken: cancellationToken);
```
tomjonesdev, line 47: Rather than have both a `Where` and a `SingleAsync`, you should be able to move the conditions into the latter.
```
 var releaseFile = await contentDbContext.ReleaseFiles.SingleAsync(...conditions, cancellationToken);
```

tomjonesdev (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseAmendmentServiceTests.cs
mmyoungman: I've used the new collection syntax, so `FilterSequence = [ ... ]`

PR #4785, opened 2024-04-23
EES 4809: Upgrade to NextJS v13

sambiramairelogic:
Upgrades NextJS in the `frontend` project to version 13.5.6.

Summary of changes: 

- The `Link` component must no longer contain an anchor tag as its child, so has been refactored accordingly.
- Babel is no longer needed, so has been replaced by SWC in the `frontend` project.
- NextJS will now report an `invalid next config` error if not provided a build number at launch. It looks like we're already doing this in production, but not locally. I've added one to the committed `.env` file and a fail-fast check in startup. To be extra safe and certain that this value doesn't make it through to production, we could create an uncommited `.env.local` and instruct people to add the build number there, but that seems like unnecessary extra manual setup steps. 
- The `jest.config.js` file in `frontend` has been updated to match changes in NextJS. 
- NextJS has been removed from the `common` project to avoid having to make the same changes to its `jest.config` file. 
- To achieve this, `next/head` has been replaced with `react-helmet`, and the dynamic wrapper for `MapBlock` has been moved into `frontend`. This means the "internal" version in `common` has now been renamed `MapBlockInternal` -> `MapBlock`. 
- This also requires that `react-leaflet` be mocked out in the `admin` and `common` projects.

The Jest tests output two warnings whilst running the `admin` tests, but these also occur on the `dev` branch so I'm not addressing them in this PR. They are:
* `Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value.`
* `Warning: validateDOMNesting(...): Whitespace text nodes cannot appear as a child of <tr>. Make sure you don't have any extra whitespace between tags on each line of your source code.`


src/explore-education-statistics-common/src/modules/charts/components/__tests__/MapBlockInternal.test.tsx
```diff
@@ -31,7 +32,7 @@ describe('MapBlockInternal', () => {
   };
 
   test('renders legends and polygons correctly', async () => {
-    render(<MapBlockInternal {...testBlockProps} />);
+    render(<MapBlock {...testBlockProps} />);
```
ntsim, line 35: We can rename this test file to `MapBlock.test.tsx` as well

src/explore-education-statistics-frontend/src/components/ButtonLink.tsx
```diff
@@ -19,13 +19,11 @@ const ButtonLink = ({
...
 }: Props) => {
-  const isAbsolute = typeof to === 'string' && to.startsWith('http');
-
-  const link = (
-    <a
-      // eslint-disable-next-line react/jsx-props-no-spreading
+  return (
+    <Link
+      href={to}
+      prefetch={prefetch}
```
ntsim, line 25: Explicit props should go after the spread to avoid being accidentally overridden in the future i.e.

```tsx
<Link
  {...props}
  href={to}
  prefetch={prefetch}
  classNames={...}
/>
```

src/explore-education-statistics-frontend/src/components/Link.tsx
```diff
@@ -40,24 +39,9 @@ const Link = ({
         className,
       )}
       data-testid={testId}
+      {...props}
```
ntsim, line 42: Similar to other comment - explicit props should go after the spread to avoid being accidentally overridden in the future i.e.

```tsx
<RouterLink
  {...props}
  href={to}
  prefetch={prefetch}
  // etc...
/>
```

src/explore-education-statistics-frontend/src/modules/find-statistics/components/DynamicDataBlockTabs.tsx
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Wrapper component so that we can lazily render {@see DataBlockTabs}.
+ * We need this as it otherwise causes issues with SSR due to Leaflet
+ * trying to initialise server-side (causing undefined `window` errors).
+ */
+const DynamicDataBlockTab = (props: DataBlockTabsProps) => {
+  return <DataBlockTabInternal {...props} />;
+};
+
+export default memo(DynamicDataBlockTab);
```
ntsim, line 21: Rather than call this `DynamicDataBlockTab` (note it should also be plural too i.e. 'tabs'), would suggest just keeping this as it's original name of `DataBlockTabs`.

There's no need to convey the implementation detail in the naming, and it's not really clear to a person just skimming the code what 'dynamic' is supposed to mean anyway.

This should have a tiny benefit of making fewer overall changes in this PR too.

src/explore-education-statistics-common/src/modules/find-statistics/components/DataBlockTabs.tsx
```diff
@@ -169,5 +169,7 @@ const DataBlockTabs = ({
 
 export default withLazyLoad(DataBlockTabs, {
   offset: 100,
-  placeholder: ({ dataBlock }) => <span data-testid={testId(dataBlock)} />,
+  placeholder: ({ dataBlock, id }) => (
+    <span id={id} data-testid={testId(dataBlock)} />
```
ntsim, line 173: I've added this `id` so that `publish_data.robot` can see this element (the release headlines) and scroll to it.

Now that we've wrapped the entire component with `dynamic` in the frontend, `react-lazyload` can't seem to detect if the lazy loaded component should render or not on initial page load. I suspect this might be due to either:

- Previously always rendering on page load for the release headlines
- Not rendering at the right time to calculate its position on the page relative to the scroll

Either way, this should fix this issue.

tests/robot-tests/tests/admin_and_public/bau/publish_data.robot
```diff
@@ -513,6 +513,7 @@ Add text block with link to a featured table to accordion section
     user clicks button    Insert    ${modal}
     user waits until modal is not visible    Insert featured table link
     user clicks button    Save & close    ${block}
+    sleep    0.5
```
ntsim, line 516: For some reason, a failure happens consistently here on your branch. Not entirely sure why this is, because you've not changed any of the admin code really :thinking:

I checked some of your dependency upgrades and couldn't see anything off really.

Regardless, adding this sleep seems to fix it even though it's a bit naff.

src/explore-education-statistics-frontend/package.json
```diff
@@ -24,8 +24,8 @@
     "leaflet": "^1.9.4",
     "lodash": "^4.17.21",
     "memoizee": "^0.4.14",
-    "next": "^12.3.4",
-    "next-transpile-modules": "^9.1.0",
+    "next": "^13.5.6",
+    "next-transpile-modules": "^10.0.1",
```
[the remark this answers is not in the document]
sambiramairelogic: Removed :)

src/explore-education-statistics-frontend/jest.config.js
```diff
@@ -1,3 +1,10 @@
+const nextJest = require('next/jest.js');
+
+const createJestConfig = nextJest({
+  // Provide the path to your Next.js app to load next.config.js and .env files in your test environment
```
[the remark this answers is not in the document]
sambiramairelogic: Aye fair shout, removed

src/explore-education-statistics-common/src/modules/charts/components/MapBlock.tsx
```diff
@@ -1,28 +1,65 @@
...
+  DataGroupingType,
+  MapConfig,
+} from '@common/modules/charts/types/chart';
+import { DataSetCategory } from '@common/modules/charts/types/dataSet';
+import { LegendConfiguration } from '@common/modules/charts/types/legend';
+import getDataSetCategoryConfigs, {
+  DataSetCategoryConfig,
+} from '@common/modules/charts/util/getDataSetCategoryConfigs';
+import { GeoJsonFeatureProperties } from '@common/services/tableBuilderService';
+
```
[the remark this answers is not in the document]
sambiramairelogic: This was an extant line that just looks like an addition cos of the rename, but yeah agreed and removed :)

src/explore-education-statistics-frontend/jest.config.js
```diff
@@ -1,3 +1,10 @@
...
+  dir: './',
+});
+
 /** @type {import('jest').Config} */
 const config = {
```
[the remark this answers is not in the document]
sambiramairelogic: I've removed the automatic ignoring of node_modules and .module/css etc, and the manual switch to SWC. I couldn't see anything else I think we can remove, but could easily be wrong

src/explore-education-statistics-common/src/components/ErrorPrefixPageTitle.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { Component } from 'react';
-import Head from 'next/head';
+import { Helmet } from 'react-helmet';
 
 class ErrorPrefixPageTitle extends Component {
```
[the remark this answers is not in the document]
sambiramairelogic: Absolutely agreed and removed. On frontend's behalf at least NextJS brings in better tools to do this anyway.

src/explore-education-statistics-frontend/src/components/ButtonLink.tsx
sambiramairelogic: Good spot, fixed

src/explore-education-statistics-frontend/src/components/Link.tsx
sambiramairelogic: Good spot, fixed

src/explore-education-statistics-frontend/src/modules/find-statistics/components/DynamicDataBlockTabs.tsx
sambiramairelogic: I'm not super keen on having two components named identically, but I do agree the word dynamic is misleading. I can't think of anything better that clearly denotes the intention of this component so have gone with your suggestion :)

src/explore-education-statistics-common/src/modules/find-statistics/components/DataBlockTabs.tsx
sambiramairelogic: Awesome, thank you :)

ntsim (approved):

PR #4793, opened 2024-04-24
EES-5085 Fix incorrect slugs for updated publication titles with dashes and spaces

ntsim:
This PR fixes incorrect slugs being shown when updating publication titles. Previously, titles like `My test - publication` would result in a slug like `my-test---publication`. With the fix, this is now `my-test-publication`.

As part of this, we've also updated the `publish_methodology_publication_update.robot` suite, which would previously fail on parallel test runs due to having a publication title that is a duplicate of another test suite (`publish_publication_update.robot`). It's likely that the publication title used in `publish_methodology_publication_update.robot` had previously been non-standard to avoid this slug issue.


mmyoungman (approved):

tomjonesdev (approved):

mmyoungman:
If you're keen, you could add the corresponding test to the backend - I think it won't have it

mmyoungman (approved):

PR #4794, opened 2024-04-24
EES-3192: Add error page redirection to Notifier on verification failure

tomjonesdev:
Public website users subscribing to publication notifications were presented with a generic, non-user-friendly error message if they attempted to verify their email address after the verification token had expired.

This change adds an error page matching existing error page format and style (and meeting the GDS standards) to inform users of the issue, and how to resolve it.

image

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/SubscriptionManager.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/SubscriptionManager.cs
@@ -1,16 +1,16 @@
-using System;
-using System.Collections.Generic;
-using System.Threading.Tasks;
 using GovUk.Education.ExploreEducationStatistics.Common.Utils;
 using GovUk.Education.ExploreEducationStatistics.Notifier.Configuration;
 using GovUk.Education.ExploreEducationStatistics.Notifier.Services.Interfaces;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Azure.Cosmos.Table;
-using Microsoft.Extensions.Logging;
-using Notify.Exceptions;
 using Microsoft.Azure.Functions.Worker;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
+using Notify.Exceptions;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
 using static GovUk.Education.ExploreEducationStatistics.Common.TableStorageTableNames;
 
 namespace GovUk.Education.ExploreEducationStatistics.Notifier.Functions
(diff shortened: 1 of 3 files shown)


src/explore-education-statistics-frontend/src/modules/subscriptions/VerificationErrorPage.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { NextPage } from 'next';
+import React from 'react';
+
+const VerificationErrorPage: NextPage = () => {
+  return (
+    <Page title="Verification failed">
+      <div className="govuk-panel__body">
+        <p>
+          Your subscription verification token has expired. You can try again by
+          re-subscribing from the publication's main screen.
```
sambiramairelogic, line 11: This may be scope creep, but it might be a nicer user journey if we were to make `publication's main screen` a link to said main screen. Could we send the publication id in the RedirectResult and add it as a prop here?

src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/SubscriptionManager.cs
```diff
@@ -254,7 +254,8 @@ public class SubscriptionManager
                 }
             }
 
-            return new BadRequestObjectResult("Unable to verify subscription");
+            return new RedirectResult(
+                $"{_appSettingOptions.PublicAppUrl}/subscriptions/verification-error");
```
sambiramairelogic, line 258: I'm not sure it's correct to use a redirect here - what's actually occurred? If the token has expired shouldn't we get a 403 or 401? Or if it's a bad payload maybe a 422? I'd prefer to return a 4xx status code and handle the problem in the frontend server.

Using a redirect will also have SEO implications - firstly we'd need to make sure it's a permeant not a temporary redirect, and secondly we'd then need to stop Google indexing the error verification page as the canonical. Perfectly doable, but feels like there are better options than using a redirect maybe?

src/explore-education-statistics-frontend/src/modules/subscriptions/VerificationErrorPage.tsx
tomjonesdev: I considered this, but the data isn't available at this stage because the token linking a user to a publication has been deleted. Mentioned to Cam and he's happy with current messaging, and can iterate down the line if it's deemed worth the effort investigating.

src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/SubscriptionManager.cs
tomjonesdev: It was a 400. The issue was that the Notifier is a separate endpoint unrelated to the Public site on which we wanted to display the error page, so it can't be handled by middleware. `RedirectResult` is used in other partS of this class for subscription update/unsubscribe.

PR #4800, opened 2024-04-25
EES-4995 show API data sets tag and allow filtering

amyb-hiveit:
- Adds an 'Available by API' tag to data sets with APIs on the data catalogue page. 
- Adds a filter to toggle between all data sets and only those with APIs
- Needs to be tested with the backend once that's ready

### DIFF
--- a/src/explore-education-statistics-frontend/src/modules/data-catalogue/components/Filters.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/data-catalogue/components/Filters.tsx
@@ -4,17 +4,30 @@ import {
   Theme,
 } from '@common/services/publicationService';
 import Button from '@common/components/Button';
-import { FormFieldset, FormGroup, FormSelect } from '@common/components/form';
-import { DataSetFileFilter } from '@frontend/services/dataSetFileService';
+import {
+  FormFieldset,
+  FormGroup,
+  FormRadioGroup,
+  FormSelect,
+} from '@common/components/form';
+import {
+  DataSetFileFilter,
+  DataSetType,
+} from '@frontend/services/dataSetFileService';
 import styles from '@frontend/modules/data-catalogue/components/Filters.module.scss';
 import React from 'react';
+import ButtonText from '@common/components/ButtonText';
+
+const formId = 'filters-form';
 
 interface Props {
+  dataSetType?: DataSetType;
   latestOnly?: string;
   publicationId?: string;
   publications?: PublicationTreeSummary[];
   releaseId?: string;
   releases?: ReleaseSummary[];
+  showClearFiltersButton?: boolean;
   themeId?: string;
   themes: Theme[];
   onChange: ({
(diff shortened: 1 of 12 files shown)


sambiramairelogic (approved):

src/explore-education-statistics-frontend/src/modules/data-catalogue/components/__tests__/Filters.test.tsx
```diff
@@ -97,12 +107,13 @@ describe('Filters', () => {
   });
 
   test('calls the onChange handler when the theme filter is changed', async () => {
+    const user = userEvent.setup();
```
ntsim, line 110: This is quite repetitive. Couldn't we stick this in our common `render` function?

For example:

```ts
// render.tsx

interface CustomRenderResult extends RenderResult {
  user: UserEvent;
}

/**
 * Custom test render function that is pre-configured with any contexts
 * that would otherwise create unnecessary boilerplate.
 */
export default function render(
  ui: ReactElement,
  options?: Omit<RenderOptions, 'queries'>,
): CustomRenderResult {
  return {
    ...(baseRender(ui, {
      wrapper: DefaultWrapper,
      ...options,
    }) as RenderResult),
    user: userEvent.setup(),
  };
}
```

We'd then be able to use it like:

```
const { user } = render(<Blah/>); 
```

This is something they suggest in the docs as well: https://testing-library.com/docs/user-event/intro/#writing-tests-with-userevent

src/explore-education-statistics-frontend/src/modules/data-catalogue/components/__tests__/Filters.test.tsx
```diff
@@ -52,6 +52,16 @@ describe('Filters', () => {
...
     expect(releases[1].selected).toBe(false);
+
+    const apiFilter = within(
+      screen.getByRole('group', {
+        name: 'Type of data',
+      }),
+    );
+    expect(apiFilter.getByLabelText('All data')).toBeInTheDocument();
+    expect(apiFilter.getByLabelText('All data')).toBeChecked();
+    expect(apiFilter.getByLabelText('API data sets only')).toBeInTheDocument();
```
ntsim, line 63: Don't need the `toBeInTheDocument` assertions? The `toBeChecked` is good enough by virtue of the element having to exist in the document for it to be checked

ntsim (changes requested):

src/explore-education-statistics-frontend/src/modules/data-catalogue/components/Filters.tsx
```diff
@@ -116,6 +128,33 @@ export default function Filters({
...
+        {showClearFiltersButton && (
+          <ButtonText onClick={onClearFilters}>Clear filters</ButtonText>
+        )}
+
+        <FormRadioGroup
+          formGroupClass="dfe-border-top govuk-!-padding-top-4 govuk-!-margin-top-2"
+          id="apiDataSetsOnly"
+          legend="Type of data"
+          legendSize="s"
+          name="apiDataSetsOnly"
```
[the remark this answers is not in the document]
amyb-hiveit: I've changed this to `dataSetType` with `all` and `api` as the options now.

src/explore-education-statistics-frontend/src/services/dataSetFileService.ts
```diff
@@ -86,6 +89,7 @@ export const dataSetFileFilters = [
 export type DataSetFileFilter = (typeof dataSetFileFilters)[number];
 
 export interface DataSetFileListRequest {
+  dataSetType?: 'all' | 'api';
```
ntsim, line 92: We could extract out a `DataSetType` for this and use it across the relevant components and props. Seems like we're currently using strings instead (which isn't type safe).

src/explore-education-statistics-frontend/src/modules/data-catalogue/components/Filters.tsx
```diff
@@ -116,6 +130,33 @@ export default function Filters({
             }}
           />
         </FormGroup>
+
+        {showClearFiltersButton && (
+          <ButtonText onClick={onClearFilters}>Clear filters</ButtonText>
+        )}
+
+        <FormRadioGroup
```
ntsim, line 138: As mentioned in another comment, we could use a `DataSetType` type here to constrain the option values and provide better type safety i.e.

```tsx
<FormRadioGroup<DataSetType> />
```

ntsim (approved):

PR #4802, opened 2024-04-25
EES-4993 Show which Datasets have associated API Datasets (Data Catalogue)

jack-hive:
On the Data Catalogue, results need to show whether a release has an associated API Dataset via a grey label and Users also need to be able to filter their search by API Datasets only. 

These changes introduce the new filter in the form an optional query parameter `ApiDataSetsOnly`, and include a new flag in the view-model to represent if a data set has an associated API data set or not `HasApiDataSet`.


src/GovUk.Education.ExploreEducationStatistics.Content.Services/DataSetFileService.cs
```diff
@@ -59,11 +60,14 @@ public class DataSetFileService : IDataSetFileService
...
 
         var query = _contentDbContext.ReleaseFiles
+            .AsNoTracking()
+            .Include(rf => rf.File)
             .OfFileType(FileType.Data)
             .HavingNoDataReplacementInProgress()
             .HavingThemeId(themeId)
             .HavingPublicationIdOrNoSupersededPublication(publicationId)
             .HavingReleaseVersionId(releaseVersionId)
+            .HavingApiDataSet(apiDataSetsOnly ?? false)
```
sambiramairelogic, line 70: Doesn't really matter, but I think you could use `apiDataSetsOnly.Value` here, like `latestOnly.Value` below?

src/GovUk.Education.ExploreEducationStatistics.Content.Services/DataSetFileService.cs
```diff
@@ -59,11 +60,14 @@ public class DataSetFileService : IDataSetFileService
             _contentDbContext.ReleaseVersions.LatestReleaseVersions(publicationId, publishedOnly: true);
 
         var query = _contentDbContext.ReleaseFiles
+            .AsNoTracking()
+            .Include(rf => rf.File)
```
benoutram, line 64: Is this change needed? I think that `BuildResultViewModel()` is not evaluated client side, so it shouldn't need every ReleaseFile to have the File loaded in order to evaluate `result.Value.File.PublicDataSetVersionId`.

jack-hive: Nope, I actually tested this and made a note to myself to remove the `Include`, but then forgot to do it!! So good spot!

I intentionally kept the `AsNoTracking` though. In theory it should speed it up? I did some performance testing and it made practically no difference with 10,000 records. But I was tempted to keep it in anyway. Do you have any thoughts?

benoutram: Oops, that comment was only meant to be on the `.Include(rf => rf.File)` 😆.

I hadn't really thought about the `AsNoTracking` here. I was just reading about tracking and custom projections in the docs [here](https://learn.microsoft.com/en-us/ef/core/querying/tracking#tracking-and-custom-projections).

I think it's pretty clear that if the result contains no entity types, like if you are projecting to a `Guid` id with no instances of the actual entity type, then no tracking will take place anyway. In those cases it's not going to make a difference.

In our case here I would have thought it doesn't make a difference either because the result shouldn't be materialised client side first in order to build the view model. I would have thought there should still be no trackable entities.

In reality though I'm not sure how it's working, especially with recent changes we are introducing now to include ordered metadata. The docs say "if EF Core materializes an entity instance for client evaluation, it will be tracked". Based on that I think it's a good to leave it in.

benoutram (approved):

PR #4806, opened 2024-04-26
EES-5099, EES-5094, EES-5102 Update public data schema with location and data set version changes

ntsim:
This PR updates the `PublicDataDbContext` for the public API with the following changes:

- [EES-5099](https://dfedigital.atlassian.net/browse/EES-5099) - Moved `PublicId` from  `LocationOptionMetaLink` to `LocationOptionMeta`

  Moving the `PublicId` to `LocationOptionMeta` was made possible by changing the data set query syntax so that users have to specify the location's level and the public ID to identify it correctly.

- [EES-5094](https://dfedigital.atlassian.net/browse/EES-5094) - Changed `MetaSummary` and `GeographicLevelMeta` to be nullable for `DataSetVersion` 

  Making `MetaSummary` and `GeographicLevelMeta` optional on `DataSetVersion` is an ergonomic change to make it easier to write tests. However, it also more accurately reflects the reality of a `DataSetVersion`, they can be in different states and don't necessarily have these relationships in place yet.

- [EES-5102](https://dfedigital.atlassian.net/browse/EES-5102) - Add `VersionPatch` to `DataSetVersion` for data replacement versioning in the future


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetVersion.cs
```diff
@@ -101,7 +109,9 @@ public void Configure(EntityTypeBuilder<DataSetVersion> builder)
                     );
             });
 
-            builder.Property(dsv => dsv.Status).HasConversion<string>();
+            builder.HasIndex(dsv => new { dsv.DataSetId, dsv.VersionMajor, dsv.VersionMinor, dsv.VersionPatch })
+                .HasDatabaseName("IX_DataSetVersions_DataSetId_VersionNumber")
+                .IsUnique();
```
benoutram, line 114: In the changes in `InitialCreate` it looks like this change is removing the index on `DataSetId` in favour of only having the multicolumn index on all the columns combined.

I was expecting this change only to create an additional index, rather than remove the exising `DataSetId` one.

Reading the docs it looks like this is ok as they say the search effectiveness is the same regardless of which index column(s) the query conditions use, but just thought I'd mention it.

ntsim: Yeh I noticed this, but EF was originally setting up the `DataSetId` index for us. I don't know if EF decided it didn't need to set this up as it was redundant, or because I was manually setting this index that affected `DataSetId`.

Either way, as you've said though, I don't think this will make any substantial difference so think it's fine to leave it is as-is.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/LocationOptionMetaGeneratorExtensions.cs
```diff
@@ -9,6 +9,11 @@ public static class LocationOptionMetaGeneratorExtensions
         int id) where TOptionMeta : LocationOptionMeta
         => generator.ForInstance(s => s.SetId(id));
 
+    public static Generator<TOptionMeta> WithPublicId<TOptionMeta>(
+        this Generator<TOptionMeta> generator,
+        string publicId) where TOptionMeta : LocationOptionMeta
+        => generator.ForInstance(s => s.SetPublicId(publicId));
+
```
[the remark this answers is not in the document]
ntsim: Not quite. `LocationOptionMeta` is abstract so there aren't actually defaults defined in these generator extensions.

The defaults get set on the generator extensions for each concrete type e.g. `LocationLocalAuthorityOptionMetaGeneratorExtensions`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetVersion.cs
benoutram: Cool, hopefully it realised the new one encompasses the old one and that it's not needed. 🙂

benoutram (approved):

PR #4813, opened 2024-05-01
DRAFT: EES-4802 Re-apply redirect map

sambiramairelogic:
Marking this as DRAFT because I really think there must be a different way of handling this. 

Google has picked up a slew of incorrect routes, which return a 404 because they don't match a path that NextJS can serve. Michael's brief requests us to add a redirect map (in [ticket](https://dfedigital.atlassian.net/browse/EES-4802)) with a status code of 301, for long enough that Google removes them from its indexes. 

We could do this in the custom server or in the NextJS middleware, but I'm using the custom server for full control over the status code, and the ability to redirect the trailingSlash with a 301 rather than a 307. 

What I don't quite see is why this is necessary, and what stops Google from just picking up more incorrect routes, given that there are literally infinite incorrect routes out there. It feels to me like the 404 is actually correct, and we should just take the SEO hit on this one, but this PR implements the brief as requested.

### DIFF
--- a/src/explore-education-statistics-frontend/redirects.js
+++ b/src/explore-education-statistics-frontend/redirects.js
@@ -758,7 +758,6 @@ const seoRedirects = [
     from: '/statistics/pupil-absence-in-schools-in-england',
     to: '/find-statistics',
   },
-  { from: '/subscriptions', to: '/' },
 ];
 
 module.exports = seoRedirects;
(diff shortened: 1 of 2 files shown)


amyb-hiveit (approved):
I'm approving this as code-wise it seems fine and some brief internet research shows that it seems to be a fairly standard thing for SEO people to recommend. I can also find people who say you don't need to, but it doesn't seem like it would cause harm.

So, even though it feels a bit wrong, I think we should probably bow to Michael's superior SEO knowledge. 

Will we need a ticket to remove these in X months once google has reindexed them?

PR #4815, opened 2024-05-01
EES-4891 add notification banner to new data catalogue page

amyb-hiveit:
Screenshot 2024-05-01 113123

### DIFF
--- a/src/explore-education-statistics-frontend/src/modules/data-catalogue/DataCataloguePageNew.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/data-catalogue/DataCataloguePageNew.tsx
@@ -9,6 +9,7 @@ import VisuallyHidden from '@common/components/VisuallyHidden';
 import ButtonText from '@common/components/ButtonText';
 import Tag from '@common/components/Tag';
 import Button from '@common/components/Button';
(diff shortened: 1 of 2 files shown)


src/explore-education-statistics-frontend/src/modules/data-catalogue/DataCataloguePageNew.tsx
```diff
@@ -261,6 +262,19 @@ export default function DataCataloguePageNew() {
           : undefined
       }
     >
+      <NotificationBanner title="This page has changed">
```
sambiramairelogic, line 265: How closely do we need to fit the design system? This is ludicrously niggly but I couldn't help but notice it asks us to "Position a notification banner immediately before the page h1.", and it looks like the <Page> component renders children just below its h1. I definitely don't think it matters though, just noticed it and thought I'd flag it

amyb-hiveit: Hmm good spot! The component has a h2 in so it would seem odd to me to have it above the h1. We're slightly misusing it here anyway and it's only temporary so I think it's probably ok.

sambiramairelogic (approved):

PR #4820, opened 2024-05-02
EES-5121 Refactor data set query string validators into POST request model validators

ntsim:
This PR refactors the facet string validators used by the data set GET query endpoint to re-use validators for the equivalent facet request models i.e.

- `LocationStringValidators` → `DataSetQueryLocation`
- `TimePeriodStringValidators` → `DataSetQueryTimePeriod`
- `SortStringValidators` → `DataSetQuerySort`

This has mostly involved separating out as much logic relevant to the facet strings and the facet request models. 

After the facet strings have been parsed and their own validations have been ran, we then parse the string into its facet request model and run its validations as well. Any errors from the facet request model are then re-mapped to validation errors for the facet string.

As part of this, we've also done a bunch of tweaks and cleanup including:

- Changes to the various error details for the string validators. These have been simplified and now just provide a `value` containing the portion of the string that was invalid.
- Updates to the relevant validation error messages to:
  - Cover various not empty errors
  - Make the error codes more consistent
- Updates test assertions for checking validation errors
- Re-using as much test data as possible between test suites for the facet strings and facet request models.
- Refactored most of the tests to leverage FluentValidation's test assertion extensions. We've also implemented some additional extensions to augment this, particularly when checking custom state.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Requests/DataSetQueryTimePeriodValidatorTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Requests/DataSetQueryTimePeriodValidatorTests.cs
@@ -1,0 +1,210 @@
+using FluentValidation.TestHelper;
+using GovUk.Education.ExploreEducationStatistics.Common.Database;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Common.Model;
+using GovUk.Education.ExploreEducationStatistics.Common.Tests.Validators;
+using GovUk.Education.ExploreEducationStatistics.Common.Utils;
+using GovUk.Education.ExploreEducationStatistics.Common.Validators.ErrorDetails;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Requests;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Validators.ErrorDetails;
+using ValidationMessages = GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Validators.ValidationMessages;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests.Requests;
+
+public class DataSetQueryTimePeriodValidatorTests
+{
+    private readonly DataSetQueryTimePeriod.Validator _validator = new();
+
+    public static TheoryData<string, string> ValidTimePeriods => new()
+    {
+        { "2020", "AY" },
+        { "2020/2021", "AY" },
+        { "2020/2021", "AYQ1" },
+        { "2020/2021", "AYQ3" },
+        { "2020", "FY" },
+        { "2021", "CY" },
+        { "2021", "CYQ1" },
+        { "2021", "CYQ2" },
+        { "2021", "RY" },
+        { "2021", "M1" },
+        { "2021", "M12" },
+        { "2021", "W20" },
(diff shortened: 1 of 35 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Requests/DataSetQueryTimePeriodValidatorTests.cs
```diff
@@ -0,0 +1,204 @@
...
+
+        _validator.TestValidate(timePeriod).ShouldNotHaveAnyValidationErrors();
+    }
+
+    public static TheoryData<string, string> InvalidTimePeriodYears => new()
+    {
+        { "2022-", "CY" },
+        { "-2022", "CY" },
+        { "2022 ", "AY" },
+        { " 2022", "AY" },
```
jack-hive, line 65: Maybe could test for negative numbers and pure alphabetic ones?

jack-hive (changes requested):

duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/DataSetQueryLocation.cs
```diff
@@ -36,6 +49,8 @@ public static DataSetQueryLocation Parse(string location)
         var property = parts[1].ToUpperFirst();
         var value = parts[2];
 
+        var parsedLevel = EnumUtil.GetFromEnumValue<GeographicLevel>(level);
+
         if (property == nameof(DataSetQueryLocationId.Id))
```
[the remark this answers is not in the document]
ntsim: Unfortunately this isn't possible as `KeyProperty` is an interface method so we need an actual instance of the object before we can use it.

In this situation, we don't have an instance yet as we're in the middle of trying to parse one out.

The language doesn't allow static interface methods, so the current approach is probably the best one we have!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Validators/LocationStringValidators.cs
```diff
@@ -30,159 +33,125 @@ public static partial class LocationStringValidators
...
+        var geographicLevel = EnumUtil.GetFromEnumValue<GeographicLevel>(level);
 
         string[] properties = geographicLevel switch
         {
             GeographicLevel.LocalAuthority =>
             [
-                "id",
-                nameof(LocationLocalAuthorityOptionMeta.Code),
-                nameof(LocationLocalAuthorityOptionMeta.OldCode)
+                nameof(DataSetQueryLocationId.Id),
```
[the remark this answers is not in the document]
ntsim: Similar to previous comment - can't do this as it's not a static method

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Requests/DataSetQueryTimePeriodValidatorTests.cs
ntsim: Already testing for negative a `-2022`. I've added in a few extra cases that contain alpha characters as well now

jack-hive: Sorry, I missed the `-2022`

jack-hive (approved):

PR #4823, opened 2024-05-02
EES-4910 Create an endpoint for Creating an API Dataset 

jack-hive:
This work is to create an endpoint in the Admin app, which can be used to create public API data sets from pre-existing data files. The endpoint acts as a proxy to the public Processor Function App which will be doing the processing/importing of the data set into an API data set.

The endpoint lives at: `POST api/public-data/data-sets`

And accepts a request body of:

```json
{
    "releaseFileId": "018c696c-ac3d-74f1-88b6-b6e78d4fc18a"
}
```

where `releaseFileId` is the primary key of the `ReleaseFiles` table in the Content DB, corresponding to the `File` that we want to create an associated API data set for.

The admin endpoint will authenticate that the user is a BAU user before allowing them to proceed with the creation of the API data set. Otherwise, the endpoint will respond with a `403`.

The endpoint also validates that there isn't already a `DataSetVersion` which has been created for the requested `releaseFileId`. If there is, then the endpoint will respond with a `400`.

If the Processor Function App responds with a `200`, this means that the new `DataSetVersion`, corresponding to the requested `releaseFileId`, has been created. We then return the ID for this in the response from the admin endpoint as a `200`:

```json
{
    "dataSetVersionId": "018c696c-ac3d-74f1-88b6-b6e78d4fc18b"
}
```

## Related Changes
The HTTP Processor Function App endpoint, which the above admin endpoint is a proxy to, can be seen at the following PR https://github.com/dfe-analytical-services/explore-education-statistics/pull/4825.

It lives at `POST api/orchestrators/processor`.


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/ProcessorClient.cs
```diff
@@ -0,0 +1,55 @@
...
+            {
+                default:
+                    var message = await response.Content.ReadAsStringAsync(cancellationToken);
+                    logger.LogError(
+                        StringExtensions.TrimIndent(
+                        $"""
+                         Failed to create data set version with status code: {response.StatusCode}. Message:
+                         {message}
+                         """));
+                    response.EnsureSuccessStatusCode();
```
ntsim, line 43: Appreciate that we discussed just throwing a server error, but does feel like this would lead to an unnecessarily bad user experience on the frontend.

I've mentioned this in various comments on #4825, but we could pipe validation errors from the processor endpoint through to the frontend so we can give better error feedback to the end user.

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/Public.Data/DataSetVersionCreateRequest.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    public class Validator : AbstractValidator<DataSetVersionCreateRequest>
+    {
+        public Validator(IDataSetVersionRepository dataSetVersionRepository)
+        {
+            RuleFor(request => request.ReleaseFileId).MustAsync(async (releaseFileId, cancellationToken) => 
+                    await dataSetVersionRepository.GetByReleaseFileId(
+                        releaseFileId: releaseFileId,
+                        cancellationToken: cancellationToken)
+                    is null)
```
benoutram, line 22: We might not need a repository method to do this as the code in that method is simple and it's also fetching back the data set version, whereas this could make use of `AnyAsync` to check for existence if using the database context directly and that would be more efficient.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Public.Data/ProcessorClientTests.cs
```diff
@@ -0,0 +1,78 @@
...
+        client.BaseAddress = BaseUri;
+        _processorClient = new ProcessorClient(Mock.Of<ILogger<ProcessorClient>>(), client);
+    }
+
+    public class ProcessTests : ProcessorClientTests
+    {
+        private static readonly Uri Uri = new(BaseUri, "api/orchestrators/processor");
+
+        [Fact]
+        public async Task HttpClientAccepted_ReturnsPublications()
```
benoutram, line 34: I think `_ReturnsPublications` in the name needs changing?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetsControllerTests.cs
```diff
@@ -952,4 +963,111 @@ private async Task<HttpResponseMessage> GetDataSet(Guid dataSetId, HttpClient? c
...
+                .WithStatusPublished();
+
+            await TestApp.AddTestData<PublicDataDbContext>(context => context.DataSets.Add(dataSet));
+
+            DataSetVersion dataSetVersion = DataFixture
+                .DefaultDataSetVersion(
+                    filters: 1,
+                    indicators: 1,
+                    locations: 1,
+                    timePeriods: 2)
```
benoutram, line 1027: There's been a recent change to make the data set version meta optional now. I think we could take advantage of that here and remove specifying these options as the meta is insignificant to this test.

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/Public.Data/DataSetVersionCreateRequest.cs
jack-hive: Moved to the `DataSetVersionService` as a method called `HasExistingVersion`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/ProcessorClient.cs
jack-hive: I've changed this to now check for 400s and 404s, and it will bubble the `ValidationProblemDetails` from the 400s back to the caller.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Public.Data/ProcessorClientTests.cs
```diff
@@ -51,6 +53,44 @@ public async Task HttpClientAccepted_ReturnsPublications()
...
+            left.AssertValidationProblem(Errors.Error1);
+        }
+
+        [Fact]
+        public async Task HttpClientNotFound_ReturnsNotFound()
+        {
+            _mockHttp.Expect(HttpMethod.Post, Uri.AbsoluteUri)
+                .Respond(HttpStatusCode.NotFound);
+
+            var response = await _processorClient.CreateInitialDataSetVersion(releaseFileId: It.IsAny<Guid>());
```
benoutram, line 86: I've not seen `It.IsAny` being used outside of mock Setup or Verify before. Is this correct?

benoutram (approved):

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Validators/ValidationMessages.cs
```diff
@@ -0,0 +1,11 @@
+using GovUk.Education.ExploreEducationStatistics.Common.Model;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Validators;
+
+public static class ValidationMessages
+{
+    public static readonly LocalizableMessage HasExistingApiDataSetVersion = new(
+        Code: "HasExistingApiDataSetVersion",
```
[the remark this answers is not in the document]
jack-hive: I'd agree with that :)

src/GovUk.Education.ExploreEducationStatistics.Admin/Validators/ValidationMessages.cs
```diff
@@ -0,0 +1,11 @@
...
+        Message: "The API data set version already exists."
```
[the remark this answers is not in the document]
jack-hive: I'd agree with that :)

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/CreateInitialDataSetVersionFunction.cs
```diff
@@ -35,17 +35,19 @@ public class CreateInitialDataSetVersionFunction(
                 instanceId,
                 cancellationToken: cancellationToken
             ))
-            .OnSuccess(async dataSetVersionId =>
+            .OnSuccess(async dataSetIdDataSetVersionIdPair =>
```
ntsim, line 38: Not a biggy, but for simplicity you could just call this `tuple`

PR #4826, opened 2024-05-03
EES-5095 Add ability to create ErrorViewModels from ValidationFailure directly

ntsim:
This PR adds the ability to construct `ErrorViewModel` directly from FluentValidation's `ValidationFailure` outside of `FluentValidationActionFilter` i.e. when calling `IValidator.Validate` manually.

We've also added an extension for `IValidator<T>` that returns a `Task<Either<ActionResult, T>`, which allows the validation call to integrate into Either returning chains.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/FluentValidationExtensions.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/FluentValidationExtensions.cs
@@ -1,13 +1,34 @@
 #nullable enable
 using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 using FluentValidation;
 using FluentValidation.Results;
 using GovUk.Education.ExploreEducationStatistics.Common.Model;
+using GovUk.Education.ExploreEducationStatistics.Common.Validators;
+using GovUk.Education.ExploreEducationStatistics.Common.ViewModels;
+using Microsoft.AspNetCore.Mvc;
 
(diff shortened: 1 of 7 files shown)


src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/FluentValidationExtensions.cs
```diff
@@ -1,13 +1,34 @@
...
+    {
+        var result = await validator.ValidateAsync(instance, cancellationToken);
+
+        if (result.IsValid)
+        {
+            return instance;
+        }
+
+        return ValidationUtils.ValidationResult(result.Errors.Select(ErrorViewModel.Create));
+    }
```
benoutram, line 30: Cool, this makes things much nicer. 🙂

src/GovUk.Education.ExploreEducationStatistics.Common/ViewModels/ErrorViewModel.cs
```diff
@@ -32,4 +34,17 @@ public record ErrorViewModel
...
+    {
+        var detail = failure.GetErrorDetail();
+
+        return new ErrorViewModel
+        {
+            Path = failure.PropertyName,
+            Code = failure.ErrorCode.Replace("Validator", ""),
+            Message = failure.ErrorMessage,
+            Detail = detail.Count > 0 ? detail : null
+        };
```
benoutram, line 48: Cool, I think this fixes the problem I was having with getting the additional details from the failure, and the reason why I had to put a custom message into my rule to identity the name of the field! Thanks.

benoutram (approved):

PR #4834, opened 2024-05-06
EES-5130 Add `InvalidRequestInputResultFilter` to handle invalid input

ntsim:
This PR adds an `InvalidRequestInputFilter` to the public API. This filter can convert errors that were caused by invalid request input into a standard validation error response.

The out-of-the-box errors are not adequate as they:

- Leak internals by reporting types that could not be deserialized from the JSON
- Don't have full paths to the part of the request causing the error (typically only reported partial paths)
- Don't have error codes

Importantly, fixing these issues provides a better user experience for the public API's data set query POST endpoint. As users can create highly nested queries, it's crucial that they get detailed validation errors that pinpoint precisely where any mistakes have been made.

## Other changes

- Fixed incorrect error paths in `LocationOptionMetaJsonConverter` when `JsonExceptions` thrown during conversion. We now re-throw these exceptions whilst capturing the original `JsonException` so that we can construct the full JSON path in the subsequent invalid input error.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
@@ -8,17 +8,19 @@ using GovUk.Education.ExploreEducationStatistics.Common.Cancellation;
 using GovUk.Education.ExploreEducationStatistics.Common.Config;
 using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Rules;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Options;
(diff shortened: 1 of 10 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
```diff
@@ -62,20 +64,21 @@ public void ConfigureServices(IServiceCollection services)
...
+                options.EnableEndpointRouting = false;
+
+                options.Filters.Add<OperationCancelledExceptionFilter>();
+                options.Filters.Add<ProblemDetailsResultFilter>();
+                options.AddInvalidRequestInputResultFilter();
                 options.AddCommaSeparatedQueryModelBinderProvider();
                 options.AddTrimStringBinderProvider();
+
+                // Adds correct camelCased paths for model errors
+                options.ModelMetadataDetailsProviders.Add(new SystemTextJsonValidationMetadataProvider());
```
duncan-at-hiveit, line 81: Nicely found

duncan-at-hiveit (approved):

PR #4838, opened 2024-05-08
Ees 5052 add private endpoint to connect psql to vnet

duncan-at-hiveit:
This PR:
- Adds a private endpoint to allow PSQL Flexible Server to be access securely directly from the VNet.
- Corrects the feeding of connection strings to the Data Processor function on deploy.
- Corrects the lookup of PSQL connection strings in Azure environment in the codebase by adding additional helpers to ConnectionUtils.

The resulting deployment shows that the `CountDataSets` Function can successfully look up the connection string and use it to query PSQL over the VNet.

image

### DIFF
--- a/infrastructure/templates/public-api/api-infrastructure-pipeline.yml
+++ b/infrastructure/templates/public-api/api-infrastructure-pipeline.yml
@@ -17,7 +17,7 @@ parameters:
 variables:
   - group: Public API Infrastructure - common
   - name: isDev
-    value: $[eq(variables['Build.SourceBranch'], 'refs/heads/EES-5052-add-private-endpoint-to-connect-psql-to-vnet')]
+    value: $[eq(variables['Build.SourceBranch'], 'refs/heads/dev')]
   - name: isTest
     value: $[eq(variables['Build.SourceBranch'], 'refs/heads/test')]
   - name: isMaster
--- a/infrastructure/templates/public-api/api-infrastructure-pipeline.yml
+++ b/infrastructure/templates/public-api/api-infrastructure-pipeline.yml
@@ -17,7 +17,7 @@ parameters:
 variables:
   - group: Public API Infrastructure - common
   - name: isDev
-    value: $[eq(variables['Build.SourceBranch'], 'refs/heads/dev')]
+    value: $[eq(variables['Build.SourceBranch'], 'refs/heads/EES-5052-add-private-endpoint-to-connect-psql-to-vnet')]
   - name: isTest
     value: $[eq(variables['Build.SourceBranch'], 'refs/heads/test')]
   - name: isMaster
--- a/infrastructure/templates/public-api/deploy-stage-template.yml
+++ b/infrastructure/templates/public-api/deploy-stage-template.yml
@@ -81,13 +81,21 @@ stages:
               scriptLocation: inlineScript
               inlineScript: |
                 set -e
+                
                 az functionapp config appsettings set \
                   --name '$(dataProcessorFunctionAppName)' \
                   --resource-group '$(resourceGroupName)' \
                   --slot 'staging' \
                   --settings \
-                    "ConnectionStrings__PublicDataDb=@Microsoft.KeyVault(VaultName=$(keyVaultName); SecretName=$(dataProcessorPsqlConnectionStringSecretKey))" \
                     "CoreStorage=@Microsoft.KeyVault(VaultName=$(keyVaultName); SecretName=$(coreStorageConnectionStringSecretKey))"
+                
+                az webapp config connection-string set \
+                  --name '$(dataProcessorFunctionAppName)' \
+                  --resource-group '$(resourceGroupName)' \
+                  --slot 'staging' \
+                  --connection-string-type=PostgreSQL \
+                  --settings \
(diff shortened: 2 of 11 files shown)


infrastructure/templates/public-api/api-infrastructure-pipeline.yml
```diff
@@ -17,7 +17,7 @@ parameters:
 variables:
   - group: Public API Infrastructure - common
   - name: isDev
-    value: $[eq(variables['Build.SourceBranch'], 'refs/heads/dev')]
+    value: $[eq(variables['Build.SourceBranch'], 'refs/heads/EES-5052-add-private-endpoint-to-connect-psql-to-vnet')]
```
benoutram, line 20: Just a reminder to revert this before merging.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
```diff
@@ -98,12 +99,12 @@ public void ConfigureServices(IServiceCollection services)
         // cause the data source builder to throw a host exception.
         if (!hostEnvironment.IsIntegrationTest())
         {
-            var connectionString = configuration.GetConnectionString("PublicDataDb")!;
+            var connectionString = ConnectionUtils.GetPostgreSqlConnectionString("PublicDataDb")!;
```
benoutram, line 102: Just wondering if you've tested a combination of this where you make the pipeline change to set the connection string using `az webapp config connection-string` but don't make this change to the Admin/Public API or Processor functions, to see if this change is required?

Based on the table shown in section [Connection string prefixes](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-8.0#connection-string-prefixes) , if there's an environment variable `CUSTOMCONNSTR_{KEY}` there should also be a converted configuration key created by removing the environment variable prefix and adding a configuration key section `ConnectionStrings:{KEY}`. If that's the case then I think this might have worked before, and our `ConnectionUtils` method of accessing the environment variables with the prefix could be redundant?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
```diff
@@ -98,12 +99,12 @@ public void ConfigureServices(IServiceCollection services)
...
 
             if (hostEnvironment.IsDevelopment())
             {
                 var dataSourceBuilder = new NpgsqlDataSourceBuilder(
-                    configuration.GetConnectionString("PublicDataDb"));
+                    ConnectionUtils.GetPostgreSqlConnectionString("PublicDataDb"));
```
benoutram, line 107: I think this can be `var dataSourceBuilder = new NpgsqlDataSourceBuilder(connectionString);` because `connectionString` already exists a few lines above.

infrastructure/templates/public-api/deploy-stage-template.yml
```diff
@@ -81,13 +81,21 @@ stages:
...
-                    "ConnectionStrings__PublicDataDb=@Microsoft.KeyVault(VaultName=$(keyVaultName); SecretName=$(dataProcessorPsqlConnectionStringSecretKey))" \
                     "CoreStorage=@Microsoft.KeyVault(VaultName=$(keyVaultName); SecretName=$(coreStorageConnectionStringSecretKey))"
+                
+                az webapp config connection-string set \
+                  --name '$(dataProcessorFunctionAppName)' \
+                  --resource-group '$(resourceGroupName)' \
+                  --slot 'staging' \
+                  --connection-string-type=PostgreSQL \
+                  --settings \
+                    "PublicDataDb=@Microsoft.KeyVault(VaultName=$(keyVaultName); SecretName=$(dataProcessorPsqlConnectionStringSecretKey))"
```
benoutram, line 98: Minor one, but `--connection-string-type 'PostgreSQL'` or `--connection-string-type PostgreSQL` looks like it should work and then be consistent with the other parameters.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
duncan-at-hiveit: I think what this is saying is that you would provide the connection string as a standard appsetting rather than explicitly as a connection string setting, but prefix it manually with one of the magic prefixes.  So it would be like:

```
                az functionapp config appsettings set \
                  --name '$(dataProcessorFunctionAppName)' \
                  --resource-group '$(resourceGroupName)' \
                  --slot 'staging' \
                  --settings \
                    "CoreStorage=@Microsoft.KeyVault(VaultName=$(keyVaultName); SecretName=$(coreStorageConnectionStringSecretKey))" \
                    "PublicDataDb=POSTGRESQLCONNSTR_@Microsoft.KeyVault(VaultName=$(keyVaultName); SecretName=$(dataProcessorPsqlConnectionStringSecretKey))"
                  
```

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
duncan-at-hiveit: Updated ta!

benoutram (approved):

PR #4849, opened 2024-05-09
EES-4722 Add data set query POST endpoint

ntsim:
This PR adds the data set query POST endpoint (`/api/v1/data-sets/{dataSetId}/query`). This endpoint provides the ability to compose much more complex queries than the GET endpoint by allowing `and`, `or` and `not` conditions.

Most of the underlying functionality has already been established for the GET endpoint, so most of this PR is just extending this prior work. 

Where possible, we try and re-use as much as possible throughout the source and tests. This should be particularly apparent in the validators, where we have very similar (but different) models for the facet representations of each endpoint (e.g. strings in GET vs JSON objects in POST).

## Other changes

- Added manual opening and closing of DuckDB connections during a query to prevent Dapper opening and closing connections on every single query. The connection is now held open until the request completes.
- Refactored out various `Create` methods for query criteria facets. This is mostly used for tests where we want to be able to provide a string comparator (via theory data) to construct the criteria facets.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetsControllerPostQueryTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetsControllerPostQueryTests.cs
@@ -1,0 +1,2878 @@
+using System.Net.Http.Json;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Common.Model;
+using GovUk.Education.ExploreEducationStatistics.Common.Model.Data;
+using GovUk.Education.ExploreEducationStatistics.Common.Tests.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Common.Utils;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests.Fixture;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests.ViewModels;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Requests;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Validators;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.ViewModels;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Database;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests.Fixtures;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Services.Interfaces;
+using Microsoft.AspNetCore.Mvc.Testing;
+using Microsoft.AspNetCore.WebUtilities;
+using Microsoft.Extensions.Primitives;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests.Controllers;
+
+public abstract class DataSetsControllerPostQueryTests(TestApplicationFactory testApp) : IntegrationTestFixture(testApp)
+{
+    private const string BaseUrl = "api/v1/data-sets";
+
+    private readonly TestParquetPathResolver _parquetPathResolver = new()
+    {
+        Directory = "AbsenceSchool"
+    };
+
+    public class AccessTests(TestApplicationFactory testApp) : DataSetsControllerPostQueryTests(testApp)
+    {
+        [Theory]
+        [InlineData(DataSetVersionStatus.Processing)]
+        [InlineData(DataSetVersionStatus.Failed)]
+        [InlineData(DataSetVersionStatus.Draft)]
+        [InlineData(DataSetVersionStatus.Mapping)]
+        [InlineData(DataSetVersionStatus.Withdrawn)]
+        [InlineData(DataSetVersionStatus.Cancelled)]
+        public async Task VersionNotAvailable_Returns403(DataSetVersionStatus versionStatus)
+        {
+            var dataSetVersion = await SetupDefaultDataSetVersion(versionStatus);
+
+            var response = await QueryDataSet(
+                dataSetId: dataSetVersion.DataSetId,
+                request: new DataSetQueryRequest
+                {
+                    Indicators = ["sess_authorised"]
+                }
+            );
+
+            response.AssertForbidden();
+        }
+
+        [Theory]
+        [InlineData(DataSetVersionStatus.Published)]
+        [InlineData(DataSetVersionStatus.Deprecated)]
+        public async Task VersionAvailable_Returns200(DataSetVersionStatus versionStatus)
+        {
+            var dataSetVersion = await SetupDefaultDataSetVersion(versionStatus);
+
+            var response = await QueryDataSet(
+                dataSetId: dataSetVersion.DataSetId,
+                request: new DataSetQueryRequest
+                {
+                    Indicators = ["sess_authorised"]
+                }
+            );
+
+            var viewModel = response.AssertOk<DataSetQueryPaginatedResultsViewModel>(useSystemJson: true);
+
+            Assert.Equal(1, viewModel.Paging.Page);
+            Assert.Equal(1, viewModel.Paging.TotalPages);
+            Assert.Equal(216, viewModel.Paging.TotalResults);
+
+            Assert.Empty(viewModel.Warnings);
+
+            Assert.Equal(216, viewModel.Results.Count);
+        }
+
+        [Fact]
+        public async Task DataSetDoesNotExist_Returns404()
+        {
+            var response = await QueryDataSet(
+                dataSetId: Guid.NewGuid(),
+                request: new DataSetQueryRequest
+                {
+                    Indicators = ["sess_authorised"]
+                }
+            );
+
+            response.AssertNotFound();
+        }
+
+        [Fact]
+        public async Task VersionDoesNotExist_Returns404()
+        {
+            var dataSetVersion = await SetupDefaultDataSetVersion();
+
+            var response = await QueryDataSet(
+                dataSetId: dataSetVersion.DataSetId,
+                dataSetVersion: "2.0",
+                request: new DataSetQueryRequest
+                {
+                    Indicators = ["sess_authorised"]
+                }
+            );
+
+            response.AssertNotFound();
+        }
+    }
+
+    public class IndicatorValidationTests(TestApplicationFactory testApp) : DataSetsControllerPostQueryTests(testApp)
+    {
+        [Fact]
+        public async Task Empty_Returns400()
+        {
+            var dataSetVersion = await SetupDefaultDataSetVersion();
+
+            var response = await QueryDataSet(
+                dataSetId: dataSetVersion.DataSetId,
+                request: new DataSetQueryRequest
+                {
+                    Indicators = []
+                }
+            );
+
+            var validationProblem = response.AssertValidationProblem();
+
+            Assert.Single(validationProblem.Errors);
+
+            validationProblem.AssertHasNotEmptyError("indicators");
+        }
+
+        [Fact]
+        public async Task Blank_Returns400()
+        {
+            var dataSetVersion = await SetupDefaultDataSetVersion();
+
+            var response = await QueryDataSet(
+                dataSetId: dataSetVersion.DataSetId,
+                request: new DataSetQueryRequest
+                {
+                    Indicators = ["", " ", "  "]
+                }
+            );
+
+            var validationProblem = response.AssertValidationProblem();
+
+            Assert.Equal(3, validationProblem.Errors.Count);
+
+            validationProblem.AssertHasNotEmptyError("indicators[0]");
+            validationProblem.AssertHasNotEmptyError("indicators[1]");
+            validationProblem.AssertHasNotEmptyError("indicators[2]");
+        }
+
+        [Fact]
+        public async Task TooLong_Returns400()
+        {
+            var dataSetVersion = await SetupDefaultDataSetVersion();
+
+            var response = await QueryDataSet(
+                dataSetId: dataSetVersion.DataSetId,
+                request: new DataSetQueryRequest
+                {
+                    Indicators = [new string('a', 41), new string('a', 42)]
+                }
+            );
+
+            var validationProblem = response.AssertValidationProblem();
+
+            Assert.Equal(2, validationProblem.Errors.Count);
+
+            validationProblem.AssertHasMaximumLengthError("indicators[0]", maxLength: 40);
+            validationProblem.AssertHasMaximumLengthError("indicators[1]", maxLength: 40);
+        }
+
+        [Fact]
+        public async Task MissingParam_Returns400()
+        {
+            var dataSetVersion = await SetupDefaultDataSetVersion();
+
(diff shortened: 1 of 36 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Requests/DataSetGetQueryLocationsValidatorTests.cs
```diff
@@ -9,147 +9,154 @@ public abstract class DataSetGetQueryLocationsValidatorTests
...
         }
 
         [Theory]
-        [MemberData(nameof(InvalidLocationStringsSingle))]
+        [MemberData(nameof(InvalidLocationFormatsSingle))]
         public void Failure_InvalidString(string location)
         {
             var query = new DataSetGetQueryLocations { Eq = location };
 
             _validator.TestValidate(query)
```
jack-hive, line 57: we could check for the specific error details here couldn't we? For instance, here I think it would be `ValidationMessages.LocationFormat.Code` etc.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Requests/DataSetQueryCriteriaLocationsValidatorTests.cs
```diff
@@ -0,0 +1,227 @@
...
+        new DataSetQueryLocationId { Level = "LA", Id = "12345" },
+        new DataSetQueryLocationLocalAuthorityCode { Code = "E08000019" },
+        new DataSetQueryLocationLocalAuthorityCode { Code = "E09000021 / E09000027" },
+        new DataSetQueryLocationLocalAuthorityOldCode { OldCode = "373" },
+        new DataSetQueryLocationLocalAuthorityOldCode { OldCode = "314 / 318" },
+        new DataSetQueryLocationId { Level = "SCH", Id = "12345" },
+        new DataSetQueryLocationSchoolUrn { Urn = "107029" },
+        new DataSetQueryLocationSchoolLaEstab { LaEstab = "3732060" },
+        new DataSetQueryLocationId { Level = "PROV", Id = "12345" },
+        new DataSetQueryLocationProviderUkprn { Ukprn = "10066874" },
```
jack-hive, line 26: Nitpick but might be nice to organise these so that the same types are next to each other. Makes it easier to digest which scenarios are covered

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Requests/DataSetQueryCriteriaLocationsValidatorTests.cs
```diff
@@ -0,0 +1,227 @@
...
+        [Theory]
+        [MemberData(nameof(InvalidLocationsMultiple))]
+        public void Failure(DataSetQueryLocation[] locations)
+        {
+            var query = new DataSetQueryCriteriaLocations { In = locations };
+
+            var result = _validator.TestValidate(query);
+
+            Assert.All(result.Errors, error =>
+                Assert.StartsWith("In", error.PropertyName));
```
jack-hive, line 168: maybe could do with a separate test(s) which covers all the individual error code/messages that are possible. And then, with this test only concerning itself with the error property name, we will have covered all ground.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Requests/DataSetQueryCriteriaTimePeriodsValidatorTests.cs
```diff
@@ -0,0 +1,338 @@
...
+    };
+
+    public static readonly TheoryData<DataSetQueryTimePeriod[]> InvalidTimePeriodsMultiple = new()
+    {
+        Array.Empty<DataSetQueryTimePeriod>(),
+        new DataSetQueryTimePeriod[]
+        {
+            new() { Period = "", Code = "AY" },
+            new() { Period = "Invalid", Code = "AY" },
+            new() { Period = "2022/2020", Code = "AY" },
```
jack-hive, line 68: Wonder if here you meant to put `2020/2022` to cover the case where the 2nd year isn't exactly **1** year after the 1st year. Because we already cover the 1st year being after the 2nd year in the line below

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Requests/DataSetQueryCriteriaTimePeriodsValidatorTests.cs
```diff
@@ -0,0 +1,338 @@
...
+
+        [Theory]
+        [MemberData(nameof(InvalidTimePeriodsMultiple))]
+        public void Failure(DataSetQueryTimePeriod[] timePeriods)
+        {
+            var query = new DataSetQueryCriteriaTimePeriods { In = timePeriods };
+
+            var result = _validator.TestValidate(query);
+
+            Assert.All(result.Errors, error =>
```
jack-hive, line 150: Might be worth counting the number of errors and checking it's equal to the number of items in each list in `InvalidTimePeriodsMultiple`

Same for other tests in this class and the others where we are testing for multiple errors

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/DataSetQueryCriteria.cs
```diff
@@ -1,3 +1,19 @@
...
 
-public abstract record DataSetQueryCriteria;
+[JsonConverter(typeof(DataSetQueryCriteriaJsonConverter))]
+public abstract record DataSetQueryCriteria
+{
+    protected static void InheritanceValidator<TCriteria>(
+        PolymorphicValidator<TCriteria, DataSetQueryCriteria> validator)
+        where TCriteria : DataSetQueryCriteria
+    {
+        validator.Add(_ => new DataSetQueryCriteriaAnd.Validator());
```
jack-hive, line 14: Can this just be `validator.Add(new DataSetQueryCriteriaAnd.Validator());`?

Same with the other lines

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/DataSetQueryRequest.cs
```diff
@@ -44,4 +45,38 @@ public record DataSetQueryRequest
...
+                    v.Add(new DataSetQueryCriteriaNot.Validator());
+                    v.Add(new DataSetQueryCriteriaFacets.Validator());
+                });
+
+            When(q => q.Sorts is not null, () =>
+            {
+                RuleFor(q => q.Sorts)
+                    .NotEmpty();
+                RuleForEach(q => q.Sorts)
+                    .SetValidator(new DataSetQuerySort.Validator());
```
jack-hive, line 73: Do we need to add a `.NotNull()` here for each item in the `Sorts` list?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Requests/DataSetQueryCriteriaNotValidatorTests.cs
```diff
@@ -0,0 +1,283 @@
...
+                ],
+            },
+        };
+
+        _validator.TestValidate(query)
+            .ShouldNotHaveAnyValidationErrors();
+    }
+
+    [Fact]
+    public void Condition_MultipleSingleFacets_Failure()
```
jack-hive, line 155: Maybe rename to `Condition_MultipleFacets_Failure` to be consistent with elsewhere

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Services/DataSetQueryParserTests.cs
```diff
@@ -1457,123 +1449,1125 @@ public async Task AllFacetsWithMixedComparatorsAndMultipleOptions()
...
+            var parsed = await service.ParseCriteria(criteria, _dataSetVersion, queryState);
+
+            const string expectedSql = """
+                                       data."field_a" = ?
+                                       """;
+
+            Assert.Equal(expectedSql, parsed.Sql);
+
+            Assert.Single(parsed.SqlParameters);
+            Assert.Equal(filterOptions[0].Id, parsed.SqlParameters[0].Argument);
```
jack-hive, line 1532: could do 

```csharp
var sqlParameter = Assert.Single(parsed.SqlParameters);
Assert.Equal(filterOptions[0].Id, sqlParameter.Argument);
```

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Requests/DataSetGetQueryLocationsValidatorTests.cs
ntsim: Nope, don't think so as we're testing a mixture of cases in the theory data so it's not simple to test every type of error that can occur.

Additionally, we already test the exact errors in great detail in `LocationStringValidatorTests` so this would be needlessly redundant.

Think it'd be better to keep these tests fairly high level just to prove integration and leave it at that.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Requests/DataSetQueryCriteriaLocationsValidatorTests.cs
ntsim: As per my other comment - don't think this is necessary. This would lead to needless redundancy as we're already testing location strings in great depth in `LocationSTringValidatorTests` and in the integration tests.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Requests/DataSetQueryCriteriaTimePeriodsValidatorTests.cs
ntsim: Nope I deliberately wanted to test it being two years after, however, you're right, we should also test your case as well! I've added this in now.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/DataSetQueryCriteria.cs
ntsim: Nope, we need to defer this or we run into a recursive loop which crashes the server.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/DataSetQueryRequest.cs
ntsim: Good point, I've added this in

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Requests/DataSetQueryRequestValidatorTests.cs
```diff
@@ -0,0 +1,427 @@
...
+                .ShouldHaveValidationErrorFor("Indicators[0]")
+                .WithErrorCode(FluentValidationKeys.MaximumLengthValidator);
+
+            result
+                .ShouldHaveValidationErrorFor("Indicators[1]")
+                .WithErrorCode(FluentValidationKeys.NotEmptyValidator);
+        }
+    }
+
+    public class SortsTests : DataSetQueryRequestValidatorTests
```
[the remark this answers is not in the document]
ntsim: Added a test case for this. I've generally tightened up the query validator tests to check null cases too.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/Converters/DataSetQueryCriteriaJsonConverter.cs
```diff
@@ -0,0 +1,83 @@
...
+        }
+        catch (JsonException exception)
+        {
+            throw new JsonException(message: null, exception);
+        }
+    }
+
+    public override void Write(Utf8JsonWriter writer, DataSetQueryCriteria value, JsonSerializerOptions options)
+    {
+        try
```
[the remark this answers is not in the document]
ntsim: Nope we don't need this. Firstly, we're in a nullable context and the value to be serialized should be non-null so we shouldn't need to guard against this.

Secondly, the `JsonSerializer.Serialize` calls return null when they receives a null, so this is already providing null protection anyway.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/Converters/DataSetQueryLocationJsonConverter.cs
```diff
@@ -0,0 +1,120 @@
...
+            .ToHashSet();
+
+        if (!doc.RootElement.TryGetProperty(nameof(DataSetQueryLocation.Level).ToLowerFirst(), out var levelProperty))
+        {
+            throw new JsonException();
+        }
+
+        try
+        {
+            var levelString = levelProperty.GetString() ?? string.Empty;
```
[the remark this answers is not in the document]
ntsim: No we don't want to throw. If possible, we want to avoid throwing so that we can actually deserialize to a request model to allow the validation layer to hit.

Not being able to deserialize usually leads to a worse user experience as it prevents us from providing more specific errors to the user.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetsControllerPostQueryTests.cs
```diff
@@ -0,0 +1,2872 @@
...
+        }
+
+        [Theory]
+        [InlineData("In")]
+        [InlineData("NotIn")]
+        public async Task InvalidMix_Returns400(string comparator)
+        {
+            var dataSetVersion = await SetupDefaultDataSetVersion();
+
+            DataSetQueryTimePeriod[] invalidTimePeriods = [
```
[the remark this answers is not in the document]
ntsim: Yeh just going to leave it tbh. This doesn't actually need to be exhaustive as we cover all this type of stuff in the validator tests.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetsControllerPostQueryTests.cs
```diff
@@ -0,0 +1,2872 @@
...
+                }
+            );
+
+            var viewModel = response.AssertOk<DataSetQueryPaginatedResultsViewModel>(useSystemJson: true);
+
+            Assert.Equal(page, viewModel.Paging.Page);
+            Assert.Equal(pageSize, viewModel.Paging.PageSize);
+            Assert.Equal(totalPages, viewModel.Paging.TotalPages);
+            Assert.Equal(216, viewModel.Paging.TotalResults);
+        }
```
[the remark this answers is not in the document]
ntsim: I've added this as `pageResults`, and for the equivalent in the GET tests

jack-hive (approved):

PR #4856, opened 2024-05-13
 Fixing ui test failure of admin and public tests  in playwright

saicharan2789:


tests/playwright-tests/general-public/pages/FindStatisticsPage.ts
```diff
@@ -12,5 +12,6 @@ export default class FindStatisticsPage {
 
   async navigateToPublicReleasePage(publicationName: string) {
     await this.releaseLink(publicationName).click();
+    await this.page.waitForTimeout(2000);
```
sambiramairelogic, line 15: It might be nice to find another way of handling this; setting manual timeouts is often considered bad practice in UI tests unless absolutely unavoidable because they can be flaky (i.e. this might usually pass, but one day it takes longer than 2 seconds for some reason and fails). They also slow things down unnecessary. The docs discourage them for exactly these reasons too - https://playwright.dev/docs/api/class-page#page-wait-for-timeout 

We should be able to wait for a condition that tells us the page has transitioned, maybe the presence of the new header or something. I believe `getByRole` implicitly waits, so something like this might work:
image

Looks like there are a few different ways of doing this in the docs, I think `page.waitForSelector` might work too :)

saicharan2789: yeah we can do that way as well, but it will break overall consistency  if write await this.page.getByRole('heading', {name:publicationName });

tests/playwright-tests/general-public/pages/FindStatisticsPage.ts
```diff
@@ -11,7 +11,7 @@ export default class FindStatisticsPage {
   }
 
   async navigateToPublicReleasePage(publicationName: string) {
-    await this.releaseLink(publicationName).click();
-    await this.page.waitForTimeout(2000);
+    await this.releaseLink(publicationName).click({ force: true });
+    await this.page.waitForNavigation();
```
sambiramairelogic, line 15: Nice, looks like waiting for all the navigation events to occur is a good way to go! Looks like this method is deprecated though, the docs say to use `waitForURL` instead -> https://playwright.dev/docs/api/class-page#page-wait-for-navigation

tests/playwright-tests/tests/admin-and-public/publishPublication.spec.ts
```diff
@@ -69,6 +69,7 @@ test('Verify that user is able to create a release via admin', async ({
 test('Validate that user is able to see the created release in public and title is same as publication name', async () => {
   await homePage.navigateToExploreFindStatisticsPage();
   await findStatisticsPage.navigateToPublicReleasePage(publicationName);
-  const title = await releasePage.pageTitle.textContent();
+  
+  const title = await releasePage.pageTitle(publicationName).innerText();
   await expect(title).toContain(publicationName);
```
sambiramairelogic, line 74: Can't help but notice the docs discourage the use of `innerText()` -> https://playwright.dev/docs/api/class-locator#locator-inner-text

Does it work if we replace both these lines with something like `await expect(releasePage.pageTitle(publicationName)).toHaveText(publicationName);` ? If it doesn't, we can pass in the `useInnerText` config option like it says in the docs - 
image

sambiramairelogic (approved):

PR #4859, opened 2024-05-15
EES-5096 Add Processor function to move CSV files from Azure Blob Storage to Azure File Share

benoutram:
This PR adds a new activity function `CopyCsvFiles` which allows copying the data and metadata csv files from private blob storage to the mounted azure file share. These csv files will eventually be used by other functions to import data.

The `CopyCsvFiles` function is chained into the `ProcessInitialDataSetVersion` function orchestration.

Whilst copying the files we also take the opportunity to compress them prior to saving them on the file share using the gzip file format.

## Other changes

- 58b02fd3b572b71c6f241e2356532bd3a5f787c6 Rename `ParquetPathResolver` to `DataSetVersionPathResolver` to reflect that it can be used for resolving other files related to a data set version (e.g. csv files).
- 0c92803263d5fa67b31badcb5c0a409eff5183ec Move `TestDataSetVersionPathResolver` to the Public.Data.Services.Tests project so that it can be reused. 
-  004dbe80d2c59c3d469443e031df9a036b74d89f Move `IDataSetService` into `Public.Data.Processor.Services.Interfaces` namesepace.
- 0743a2ffbcf04b70510ff532f9e829b3a457c2e9 Inline `CompleteProcessing` and `HandleProcessingFailure` activity functions into the class `ProcessInitialDataSetVersionFunction`.


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/CopyCsvFilesFunctionTests.cs
```diff
@@ -0,0 +1,127 @@
...
+                    .Generate(1))
+                .FinishWith(dsv => dsv.DataSet.LatestDraftVersion = dsv);
+
+            await AddTestData<PublicDataDbContext>(context =>
+            {
+                context.DataSetVersions.Add(dataSetVersion);
+                context.DataSets.Update(dataSet);
+            });
+
+            var blobStorageService = GetRequiredService<IPrivateBlobStorageService>();
```
duncan-at-hiveit, line 80: Is this a dummy implementation that writes to the filesystem, or using a test container for Azurite, or do we need storage running when running these tests?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/CopyCsvFilesFunctionTests.cs
```diff
@@ -0,0 +1,127 @@
...
+                .FirstAsync(dsv => dsv.Id == dataSetVersion.Id);
+
+            var savedImport = savedDataSetVersion.Imports.Single();
+            Assert.Equal(DataSetVersionImportStage.CopyingCsvFiles, savedImport.Stage);
+
+            var dataSetVersionPathResolver = GetRequiredService<IDataSetVersionPathResolver>();
+
+            var expectedCsvDataPath = dataSetVersionPathResolver.CsvDataPath(dataSetVersion);
+            Assert.True(File.Exists(expectedCsvDataPath));
+            Assert.Equal(DataFileContent, await DecompressFileToString(expectedCsvDataPath));
```
duncan-at-hiveit, line 114: Awesome test :)

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/CreateInitialDataSetVersionFunctionTests.cs
```diff
@@ -46,6 +49,7 @@ public async Task Success()
 
             await AddTestData<ContentDbContext>(context =>
             {
+                context.ReleaseVersions.Add(releaseVersion);
```
duncan-at-hiveit, line 52: Does the saving of the ReleaseFiles not cascade save the ReleaseVersions that they're attached to?  Just wondering why you need to explicitly add an `Add` here, not that it's a biggie or owt :)

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ProcessInitialDataSetVersionFunctionTests.cs
```diff
@@ -97,4 +118,108 @@ public async Task ActivityFunctionThrowsException_CallsHandleFailureActivity()
...
+            await AddTestData<PublicDataDbContext>(
+                context =>
+                {
+                    context.DataSetVersions.Add(dataSetVersion);
+                    context.DataSets.Update(dataSet);
+                }
+            );
+
+            var function = GetRequiredService<ProcessInitialDataSetVersionFunction>();
+            await function.HandleProcessingFailure(
```
duncan-at-hiveit, line 156: I was half-expecting to see that we'd be stashing an error of some kind on the import at this point?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/CopyCsvFilesFunctionTests.cs
duncan-at-hiveit: Aha, I've just found:

```
<PackageReference Include="Testcontainers.Azurite" Version="3.8.0" />
```

:slightly_smiling_face:

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ProcessInitialDataSetVersionFunctionTests.cs
```diff
@@ -1,6 +1,11 @@
+using GovUk.Education.ExploreEducationStatistics.Common.Tests.Extensions;
```
duncan-at-hiveit, line 1: I like your organisation of the various function tests here.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/CopyCsvFilesFunction.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        var blobStream = await privateBlobStorageService.StreamBlob(
+            BlobContainers.PrivateReleaseFiles,
+            csvFile.Path(),
+            cancellationToken: cancellationToken);
+
+        await using var fileStream = new FileStream(destinationPath, FileMode.Create, FileAccess.Write);
+        await CompressionUtils.CompressToStream(blobStream, fileStream, ContentEncodings.Gzip, cancellationToken);
+        fileStream.Close();
```
duncan-at-hiveit, line 73: Do you need to explicitly close if using `await using`?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/ProcessInitialDataSetVersionFunction.cs
```diff
@@ -43,7 +54,59 @@ public class ProcessInitialDataSetVersionFunction
...
+        await publicDataDbContext.SaveChangesAsync(cancellationToken);
+
+        // Any additional logic to tidy up after importing will be added here
+
+        dataSetVersion.Status = DataSetVersionStatus.Draft;
+        dataSetVersionImport.Completed = DateTimeOffset.UtcNow;
+        await publicDataDbContext.SaveChangesAsync(cancellationToken);
+    }
+
+    [Function(nameof(HandleProcessingFailure))]
```
duncan-at-hiveit, line 84: Is the plan that we'll supply some details as to the error encountered here?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Services.Tests/DataSetVersionPathResolverTests.cs
```diff
@@ -14,12 +14,12 @@
 
 namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Services.Tests;
 
-public abstract class ParquetPathResolverTests
+public abstract class DataSetVersionPathResolverTests
```
duncan-at-hiveit, line 17: I see you're making the outer classes abstract and I've seen Nick making them static I think.  Be good to agree upon a best convention.

duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/CopyCsvFilesFunctionTests.cs
```diff
@@ -0,0 +1,127 @@
...
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Database;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests.Fixtures;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Functions;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Services.Interfaces;
+using Microsoft.EntityFrameworkCore;
+using File = System.IO.File;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests.Functions;
+
+public class CopyCsvFilesFunctionTests(ProcessorFunctionsIntegrationTestFixture fixture)
```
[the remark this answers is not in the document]
benoutram: I think they have been made abstract which makes sense to me. I'll do the same here, thanks.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Services.Tests/DataSetVersionPathResolverTests.cs
benoutram: Spoke to @ntsim about this and he has been doing a mixture of both, creating abstract base classes where test classes need to inherit dependencies like the Test application, and in other cases like util tests which don't need to inherit, make the outer class static.

When searching the codebase for classes with the regex `public static class.+Tests` they all seem to be util type tests except for `DataSetFilesControllerCachingTests`, but it's not so clear cut when searching for `public abstract class.+Tests` where there's a mixture of util and non-util type tests.

I've made a change to `DataSetFilesControllerCachingTests`, and two existing processor test classes `CreateInitialDataSetVersionFunctionTests`, and `ProcessInitialDataSetVersionFunctionTests` to make the base classes abstract.

I've also added a note for discussion on the backend test style guide proposals board.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/CopyCsvFilesFunction.cs
benoutram: Good point, we shouldn't need this. Removed `.Close()`, thanks.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/CreateInitialDataSetVersionFunctionTests.cs
benoutram: Yes it should do.

I split these out in an attempt to fix some flaky tests, but it made no difference so I've undone the explicit `Add` of the release version now, thanks.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/ProcessInitialDataSetVersionFunction.cs
benoutram: I think we could do this and stash some error details in the `DataSetVersionImport`, but we should also be able to get any exception details from the durable orchestrations build in HTTP API's to retrieve the status of the instance. That's the reason why I've stashed the instance id on the `DataSetVersionImport`.

I've spoken with @ntsim about this previously and I think we said there could be a proxy in the Admin to that built-in API to retrieve the status from the Processor function app, abstracting away some of the complexity of it.

For now though, we haven't done anything like that so we would need to query the built-in API directly with the instance id to get the error.

Btw, EES-5096 hasn't implemented the `HandleProcessingFailure` function, it's only moved the location of it, which is why your seeing this change here.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ProcessInitialDataSetVersionFunctionTests.cs
benoutram: Replied in other related comment, thanks.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/ProcessInitialDataSetVersionFunction.cs
duncan-at-hiveit: Yup no probs, I figured, thanks!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Services.Tests/DataSetVersionPathResolverTests.cs
duncan-at-hiveit: Brill thanks for investigating

PR #4866, opened 2024-05-16
EES-5129/Add manual confirmation pages for publication subscriptions

sambiramairelogic:
Whilst awaiting DfE assistance in adding the e-mail headers detailed in https://dfedigital.atlassian.net/browse/EES-5129, we need to introduce manual confirmation steps to users subscribing to/unsubscribing from publications. 

To facilitate this I've added two new pages to the Frontend to confirm subscribing/unsubscribing to publications. The links sent out in emails now point to these FE pages instead, and the FE itself calls the Notifier functions. This means we no longer need the verification-error page, this has become merely a component rendered when an error is encountered. 

This has sparked quite a big refactor of both the FE and BE components of this functionality, and as since it was actually quite lacking in tests I've added quite a few to be sure. Admittedly, they could be much more robust - I'm mostly only testing the happy paths, and we could be tad smarter about disposing our test data on dispose, but it's a trade-off with time spent.


src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/SubscriptionManagerTests.cs
```diff
@@ -0,0 +1,554 @@
...
+{
+    private static readonly GovUkNotifyOptions.EmailTemplateOptions EmailTemplateOptions = new()
+    {
+        ReleaseAmendmentPublishedId = "release-amendment-published-id",
+        ReleaseAmendmentPublishedSupersededSubscribersId = "release-amendment-published-superseded-subscribers-id",
+        ReleasePublishedId = "release-published-id",
+        ReleasePublishedSupersededSubscribersId = "release-published-superseded-subscribers-id",
+        SubscriptionConfirmationId = "subscription-confirmation-id",
+        SubscriptionVerificationId = "subscription-verification-id"
+    };
```
tomjonesdev, line 35: Could the function and mocked services which are built in each test be instantiated at a single point in the constructor instead? Would save a lot of repetition in the "Arrange" portion of each test.

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/SubscriptionManagerTests.cs
```diff
@@ -0,0 +1,554 @@
...
+            Title = "Test Publication Title 1"
+        };
+
+        // Act
+        var result =
+            await notifierFunction.RequestPendingSubscriptionFunc(request, new TestFunctionContext(),
+                new CancellationToken());
+
+        // Assert
+        var okResult = Assert.IsType<OkObjectResult>(result);
```
tomjonesdev, line 83: Not super important, but I'd suggest that the initial type assertion here is sufficient - the null and status code checks which follow don't confirm anything new/different (i.e. an `OkObjectResult` is equivalent to a 200, and non-null is implied by the type check).

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/SubscriptionManagerTests.cs
```diff
@@ -0,0 +1,554 @@
...
+        Assert.NotNull(okResult);
+        Assert.Equal(200, okResult.StatusCode);
+
+        emailService.Verify(mock =>
+            mock.SendEmail(notificationClient, "test1@test.com", "subscription-verification-id",
+                It.Is<Dictionary<string, dynamic>>(d =>
+                    AssertEmailTemplateValues(d, "Test Publication Title 1",
+                        "https://localhost:3000/subscriptions/test-publication-slug-1/confirm-subscription/activation-code-1",
```
tomjonesdev, line 91: Is this base URL/port guaranteed to be remain constant across dev machines/build pipelines/environments, or would a config variable be more robust here?

src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/SubscriptionManager.cs
```diff
@@ -167,44 +154,44 @@ public class SubscriptionManager
...
 
-                if (sub != null)
-                {
-                    await _storageTableService.RemoveSubscriber(table, sub);
-                    return new RedirectResult(
-                        $"{_appSettingOptions.PublicAppUrl}/subscriptions?slug={sub.Slug}&unsubscribed=true",
-                        true);
-                }
+            var table = await _storageTableService.GetTable(NotifierSubscriptionsTableName);
+            var sub = _storageTableService.RetrieveSubscriber(table, new SubscriptionEntity(id, email)).Result;
```
mmyoungman, line 170: I know this isn't you, but I'd switch this `.Result` out for an `await`.

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/SubscriptionManagerTests.cs
sambiramairelogic: It is actually is, because it's being fed in when the class is built (`Options.Create(new AppSettingOptions { PublicAppUrl = "https://localhost:3000" })` . 
I did think about making it a config variable, but wasn't sure if we already had those set in all environments and was being lazy :D I do agree it would be neater

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/SubscriptionManagerTests.cs
sambiramairelogic: Yeah you're 100% right, will remove

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/SubscriptionManagerTests.cs
sambiramairelogic: They absolutely could, and I'd much prefer them that way myself to be honest. I left them like this because it's what we're doing elsewhere, and when I checked the styleguide for this sort of thing I found this:
image

src/explore-education-statistics-frontend/src/modules/subscriptions/ConfirmSubscriptionPage.tsx
```diff
@@ -0,0 +1,102 @@
...
+  Subscription,
+} from '@frontend/services/notificationService';
+import withAxiosHandler from '@frontend/middleware/ssr/withAxiosHandler';
+import SubscriptionStatusMessage from '@frontend/modules/subscriptions/components/SubscriptionStatusMessage';
+import VerificationErrorMessage from '@frontend/modules/subscriptions/components/VerificationErrorMessage';
+import Head from 'next/head';
+
+interface Props {
+  publicationSlug: string;
+  data: PublicationTitle;
```
mmyoungman, line 19: I'd be inclined to rename this from `data` to `publication`

Same with ConfirmUnsubscriptionPage

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/ReleaseNotifierTests.cs
mmyoungman, line 1: I notice this PR is being merged to `dev`. I think that's probably right, but just pointing it out in case we want to hotfix this to `master`.

src/explore-education-statistics-frontend/src/modules/subscriptions/NewSubscriptionPage.tsx
```diff
@@ -0,0 +1,131 @@
...
+  slug: string;
+  data: PublicationTitle;
+}
+
+const SubscriptionPage: NextPage<Props> = ({ data, slug }) => {
+  const { isMounted } = useMounted();
+
+  if (!isMounted) {
+    return null;
+  }
```
mmyoungman, line 33: Could be me, but I don't understand why this is necessary

src/explore-education-statistics-frontend/src/modules/subscriptions/ConfirmUnsubscriptionPage.tsx
```diff
@@ -0,0 +1,93 @@
...
+  publicationSlug,
+  token,
+}) => {
+  const [unsubscribedSubscription, setUnsubscribedSubscription] = useState<
+    Subscription | undefined
+  >(undefined);
+
+  const onConfirmClicked = async () => {
+    // TODO: Currently we get a 400 if the token isn't valid, but aren't
+    // properly handling this like we are for verification errors.
```
tomjonesdev, line 33: Do these 400s return an error similar to [EES-3192](https://dfedigital.atlassian.net/browse/EES-3192)? Does there need to be a ticket to implement similar error handling?

src/explore-education-statistics-frontend/src/modules/subscriptions/components/VerificationErrorMessage.tsx
```diff
@@ -0,0 +1,28 @@
...
+  slug: string;
+}
+
+const VerificationErrorMessage = ({ slug }: VerificationErrorMessageProps) => {
+  return (
+    <div className="govuk-panel__body">
+      <p>
+        Your subscription verification token has expired. You can try again by
+        re-subscribing from the{' '}
+        <Link to={`/find-statistics/${slug}`}>publication's main screen.</Link>
```
tomjonesdev, line 14: Cam will be happy with this. A link to the applicable publication was a nice-to-have when the verification error page was added before, but the data wasn't available because the token linking it with a publication no longer existed.

src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/SubscriptionManager.cs
```diff
@@ -10,155 +11,141 @@
...
-            _logger.LogInformation("{FunctionName} triggered", context.FunctionDefinition.Name);
+            logger.LogInformation("{FunctionName} triggered", context.FunctionDefinition.Name);
 
-            var subscriptionsTable = await _storageTableService.GetTable(NotifierSubscriptionsTableName);
-            var pendingSubscriptionsTable = await _storageTableService.GetTable(NotifierPendingSubscriptionsTableName);
+            var validationResult = await requestValidator.Validate(req, cancellationToken);
+            if (validationResult.IsLeft) // If validation failed
+            {
+                logger.LogError("{FunctionName} failed because the {req} did not contain a valid JSON object.", context.FunctionDefinition.Name, req);
+                return validationResult.Left;
```
tomjonesdev, line 54: What sort of information is included in this result? Does it need to be "cleansed" or made more client-friendly before being returned (i.e. returned as a new `BadRequestResult` (or similar) with a custom message like the other methods)?

tomjonesdev (approved):

src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/SubscriptionManager.cs
sambiramairelogic: Ooh good spot, yeah will change

src/explore-education-statistics-frontend/src/modules/subscriptions/ConfirmSubscriptionPage.tsx
sambiramairelogic: Completely agreed - it also irks me that we've got the slug in two places, but I appreciate there's actually a technical benefit in that case. I only left it as-was at first because I was trying not to change too much existing code, but at this point I mean... why not :D

src/explore-education-statistics-frontend/src/modules/subscriptions/ConfirmSubscriptionPage.tsx
```diff
@@ -0,0 +1,102 @@
...
+}) => {
+  const [confirmedSubscription, setConfirmedSubscription] = useState<
+    Subscription | 'verification-error' | undefined
+  >(undefined);
+
+  const onConfirmClicked = async () => {
+    const response = await notificationService.confirmPendingSubscription(
+      data.id,
+      token,
+    );
```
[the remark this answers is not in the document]
sambiramairelogic: Good shout - I did actually wonder about wrapping this up in a hook so we can return an `{ result: T | undefined, isLoading: boolean, Errors: Error[] }` sort of object, and tie the new `<LoadingSpinner>`'s behaviour to the `isLoading` value. It's a pattern I quite like, but I can't find any examples of us doing the same and I didn't wanna rock the boat too much. Will take a look at wrapping it in a `Queries.ts` instead :)

src/explore-education-statistics-frontend/src/modules/subscriptions/NewSubscriptionPage.tsx
sambiramairelogic: Neither do I, but it was already there and I'm scared to touch it 🤣 I believe I've seen it in other files with a `<FormProvider>`, and I think it's something to do with the forms loading properly. @amyb-hiveit might know more?

src/explore-education-statistics-frontend/src/modules/subscriptions/ConfirmSubscriptionPage.tsx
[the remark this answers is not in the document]
sambiramairelogic: Ah okay, so looks like the `*queries.ts` files are to wrap them in the Tanstack react-query stuff, which allows us better caching-control through the use of query keys. There's probably no benefit to that in this case, so might not be necessary, but I'll have a think overnight and address first thing tomorrow :)

sambiramairelogic:
This work was duplicated and re-targeted at the master branch

PR #4867, opened 2024-05-16
Ees 5127 add parquet file share mounts remove existing fileshares

duncan-at-hiveit:
# Overview

This PR:
- mounts the Public API Parquet file share into the Data Processor Function App and the API Container App (and Container App Environment)
- updates the `functionApp.bicep` and Data Processor's `host.json` to be in line with Durable Functions zero-downtime deploy strategy
- Introduces a new user-assigned identity for the Data Processor and its staging slot to use for a single consistent database user within PostgreSQL that both slots can use, and that can survive the Data Processor Function App being deleted and recreated.
- Adds some basic debug "Health Check" Functions into Data Processor to aid in ensuring that the file mount and database connection is OK to use.


infrastructure/templates/public-api/main.bicep
```diff
@@ -196,18 +219,55 @@ resource apiContainerAppManagedIdentity 'Microsoft.ManagedIdentity/userAssignedI
...
     managedEnvironmentId: containerAppEnvironmentModule.outputs.containerAppEnvironmentId
+    volumeMounts: [
+      {
+        volumeName: parquetFileShareMountName
+        mountPath: parquetFileShareMountPath
+      }
+    ]
+    volumes: [
+      {
+        name: parquetFileShareMountName
```
jack-hive, line 264: so is the mount name the same as the volume name?

infrastructure/templates/public-api/components/storageAccount.bicep
```diff
@@ -53,7 +48,7 @@ resource storageAccount 'Microsoft.Storage/storageAccounts@2023-01-01' = {
         value: ipRule
         action: 'Allow'
       }]
-      virtualNetworkRules: [for ipRule in storageSubnetRules: {
+      virtualNetworkRules: [for ipRule in allowedSubnetIds: {
```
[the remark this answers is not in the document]
duncan-at-hiveit: Good spot - done, thanks!

jack-hive (approved):

infrastructure/templates/public-api/components/functionApp.bicep
```diff
@@ -68,67 +95,63 @@ resource appServicePlan 'Microsoft.Web/serverfarms@2023-01-01' = {
   }
 }
 
-var dedicatedStorageAccountName = replace('${subscription}eessa${functionAppName}', '-', '')
+// Configuring a single shared storage account for task management, and 2 individual storage accounts to be split
```
benoutram, line 101: Going by the docs, I think the shared storage account is used for normal operations - key management, timer trigger management etc.

I think it would be good to adjust this comment to say that, and that the individual storage accounts are for use by the Durable Functions extension, for reliable execution.

infrastructure/templates/public-api/components/functionApp.bicep
```diff
@@ -68,67 +95,63 @@ resource appServicePlan 'Microsoft.Web/serverfarms@2023-01-01' = {
...
+// between the production slot and staging slot. See
+// https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-zero-downtime-deployment#status-check-with-slot
```
benoutram, line 103: This is really cool. Just looking at the docs, it mentions implementing a status check function for use by the CI/CD pipeline, so that the slots are swapped only when there's no pending or running orchestrations. Don't we also need to do that here?

infrastructure/templates/public-api/components/functionApp.bicep
```diff
@@ -68,67 +95,63 @@ resource appServicePlan 'Microsoft.Web/serverfarms@2023-01-01' = {
...
+var durableManagementStorageAccountName = replace('${subscription}eessa${functionAppName}mg', '-', '')
+var slot1StorageAccountName = replace('${subscription}eessa${functionAppName}s1', '-', '')
+var slot2StorageAccountName = replace('${subscription}eessa${functionAppName}s2', '-', '')
```
benoutram, line 106: The `DurableManagementStorage` connection string are set using the outputs of the `slot1StorageAccountModule` and `slot2StorageAccountModule` which use the names of `slot1StorageAccountName` and `slot1StorageAccountName`.

I think the `AzureWebJobsStorage` connection string is for the shared storage account, and that its name should be something like `sharedStorageAccountName` here?

infrastructure/templates/public-api/components/functionApp.bicep
```diff
@@ -68,67 +95,63 @@ resource appServicePlan 'Microsoft.Web/serverfarms@2023-01-01' = {
...
+  name: '${durableManagementStorageAccountName}StorageAccountDeploy'
+  params: {
+    location: location
+    storageAccountName: durableManagementStorageAccountName
+    allowedSubnetIds: [subnetId]
+    skuStorageResource: 'Standard_LRS'
+    keyVaultName: keyVaultName
+    tagValues: tagValues
   }
 }
```
benoutram, line 125: Related to the comment above, shouldn't this be named something like `sharedStorageAccountName` as its used to configure the `AzureWebJobsStorage` connection string?

infrastructure/templates/public-api/components/functionApp.bicep
```diff
@@ -39,6 +39,16 @@ param applicationInsightsKey string
 @description('Specifies the subnet id')
 param subnetId string
 
+@description('Specifis whether this Function App is accessible from the public internet')
```
benoutram, line 42: Typo here in 'specifies'

infrastructure/templates/public-api/components/functionApp.bicep
```diff
@@ -143,14 +169,47 @@ resource functionApp 'Microsoft.Web/sites@2023-01-01' = {
...
+       shareName: next.fileShareName
+       mountPath: next.mountPath
+       accountName: next.storageAccountName
+       accessKey: next.storageAccountKey
+     }
+   }))
+}
+
+resource slot1StorageAccountFileShare 'Microsoft.Storage/storageAccounts/fileServices/shares@2023-01-01' = {
+  name: '${durableManagementStorageAccountName}/default/${fullFunctionAppName}1'
```
benoutram, line 195: I don't understand why the name here is `durableManagementStorageAccountName` rather than `slot1StorageAccountName`?

infrastructure/templates/public-api/components/functionApp.bicep
```diff
@@ -143,14 +169,47 @@ resource functionApp 'Microsoft.Web/sites@2023-01-01' = {
...
+  dependsOn: [
+    durableManagementStorageAccountModule
+  ]
+}
+
+resource slot2StorageAccountFileShare 'Microsoft.Storage/storageAccounts/fileServices/shares@2023-01-01' = {
+  name: '${durableManagementStorageAccountName}/default/${fullFunctionAppName}2'
```
benoutram, line 202: Same as above, I don't understand why the name here is `durableManagementStorageAccountName` rather than `slot2StorageAccountName`?

infrastructure/templates/public-api/components/functionApp.bicep
duncan-at-hiveit: I've updated various comments to better reflect this now, thanks!

infrastructure/templates/public-api/components/functionApp.bicep
duncan-at-hiveit: I think that there's a potential decision to be made here on how to time the slot swapping.  One way is to wait until there's no activity.  The other way is to allow the Durable Functions that were originally running in the production slot to continue running while it is being swapped into the staging slot.  That way, long-running operations don't need to block deploys and, by virtue of each slot having its own dedicated storage account, it's able to continue its execution even after being swapped into staging.

I think we need to spend some time looking into these strategies, but I'm inclined not to try to add it to this particular PR.  Happy for me to spin off an additional ticket?

infrastructure/templates/public-api/components/functionApp.bicep
duncan-at-hiveit: :+1: done ta

infrastructure/templates/public-api/components/functionApp.bicep
duncan-at-hiveit: Corrected ta

infrastructure/templates/public-api/components/functionApp.bicep
duncan-at-hiveit: This was by design, opting to put the file shares that are used for the individual slots to hold their code into the shared storage account.

However, I'm happy to change this to try to put it into the dedicated storage accounts for slot1 and slot2 though and see how it goes.

infrastructure/templates/public-api/components/functionApp.bicep
duncan-at-hiveit: As above

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/ProcessorHostBuilder.cs
```diff
@@ -73,7 +73,14 @@ public static IHostBuilder ConfigureProcessorHostBuilder(this IHostBuilder hostB
...
                         services.AddDbContext<PublicDataDbContext>(options =>
                         {
-                            var sqlServerTokenProvider = new DefaultAzureCredential();
+                            var sqlServerTokenProvider = new DefaultAzureCredential(
+                                new DefaultAzureCredentialOptions
+                                {
+                                    // Unlike Container Apps and App Services, DefaultAzureCredential does not pick up 
+                                    // the "AZURE_CLIENT_ID" environment variable automatically when operating within
+                                    // a Function App.  We therefore provide it manually.
+                                    ManagedIdentityClientId = Environment.GetEnvironmentVariable("AZURE_CLIENT_ID")
```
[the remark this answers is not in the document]
duncan-at-hiveit: I'll give it a shot

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/HealthCheckFunctions.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                logger.LogError("Unable to find the file share mount");
+            }
+        }
+        catch (Exception e)
+        {
+            logger.LogError(e, "Error encountered when attempting to find the file share mount");
+            throw;
+        }
+    }
```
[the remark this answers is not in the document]
duncan-at-hiveit: It was a bit of a throwaway comment in the PR description, but I'd changed this function to be a little safer and just check for the existence of the file mount instead.

Updated both to be HTTP triggered Functions.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/host.json
```diff
@@ -2,8 +2,10 @@
   "version": "2.0",
   "extensions": {
     "durableTask": {
+      "hubName": "DataProcessorTaskHub",
```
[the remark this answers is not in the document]
duncan-at-hiveit: I don't think we'll need this because each slot uses a distinct Storage Account connection string for the "DurableManagementStorage" appsetting.

infrastructure/templates/public-api/components/functionApp.bicep
duncan-at-hiveit: Seems to work fine and makes a lot more sense - I've made the change thanks.

PR #4868, opened 2024-05-16
EES-5140 Fix DataSetFileMeta / Add TimePeriodRangeMeta

mmyoungman:
This PR fixes an issue where `DataSetFileMeta` only stored one `TimeIdentifier` for a whole data set. This assumption I made is incorrect - a data set can have multiple `TimeIdentifier`'s e.g. months.

`File.DataSetFileMeta` now has a `TimePeriodRangeMeta` which stores the data set's starting and ending `TimeIdentifier`/`Year`.

We are finished with the DataSetFileMeta migration endpoint, so I repurposed it in this PR to set the new `TimePeriodRange` field and remove `TimeIdentifier` and `Years`.

The `TimeIdentifier` and `Year` fields in `DataSetFileMeta` can be removed after this migration. I've added a note to do so to EES-4918.

### DIFF
--- a/tests/robot-tests/tests/snapshots/find_statistics_snapshot.json
+++ b/tests/robot-tests/tests/snapshots/find_statistics_snapshot.json
@@ -79,7 +79,7 @@
   {
     "publication_summary": "Children accommodated in secure children's homes in England and Wales including approved places, availability, occupancy, sex, age, ethnicity, length of stay",
     "publication_title": "Children accommodated in secure children's homes",
-    "published": "18 May 2023",
+    "published": "16 May 2024",
     "release_type": "National statistics",
     "theme": "Children's social care"
   },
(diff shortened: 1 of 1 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Model/DataSetFileMeta.cs
```diff
@@ -11,31 +12,52 @@ public class DataSetFileMeta
...
+
+    public TimePeriodRangeBoundMeta End { get; set; } = null!;
+}
+
+public class TimePeriodRangeBoundMeta
+{
+    [JsonConverter(typeof(TimeIdentifierJsonConverter))]
+    public TimeIdentifier TimeIdentifier { get; set; }
+
+    public int Year { get; set; }
```
ntsim, line 40: Would change this to `string Period` as we've modelled it in the public API. 

We've purposefully done this because singular dates and date ranges are eventually coming down the pipeline ([EES-3109](https://dfedigital.atlassian.net/browse/EES-3109)). 

These new time periods are potentially going in the `time_period` column, so the current assumptions that we make all over the place this is numeric year are likely to be incorrect in the future. Modelling our meta summary as suggested helps prevent this being an issue later down the line.

PR #4871, opened 2024-05-17
EES-5140 Fix DataSetFileMeta / Add TimePeriodRangeMeta

mmyoungman:
Copy of this PR, since I accidentally blitzed my changes :fearful: 

This PR fixes an issue where DataSetFileMeta only stored one TimeIdentifier for a whole data set. This assumption I made is incorrect - a data set can have multiple TimeIdentifier's e.g. months.

File.DataSetFileMeta now has a TimePeriodRangeMeta which stores the data set's starting and ending TimeIdentifier/Year.

We are finished with the DataSetFileMeta migration endpoint, so I repurposed it in this PR to set the new TimePeriodRange field and remove TimeIdentifier and Years.

The TimeIdentifier and Year fields in DataSetFileMeta can be removed after this migration. I've added a note to do so to EES-4918.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Converters/GeographicLevelsListJsonConverter.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Converters/GeographicLevelsListJsonConverter.cs
@@ -1,0 +1,51 @@
+#nullable enable
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Common.Model.Data;
+using Newtonsoft.Json;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Converters;
+
+public class GeographicLevelsListJsonConverter : JsonConverter<IList<GeographicLevel>>
+{
+    public override void WriteJson(JsonWriter writer, IList<GeographicLevel>? value, JsonSerializer serializer)
+    {
+        writer.WriteStartArray();
+        if (value != null)
+        {
+            foreach (var gl in value)
+            {
+                writer.WriteValue(gl.GetEnumValue());
+            }
+        }
+        writer.WriteEndArray();
+    }
+
+    public override List<GeographicLevel>? ReadJson(
+        JsonReader reader,
+        Type objectType,
+        IList<GeographicLevel>? existingValue,
+        bool hasExistingValue,
+        JsonSerializer serializer)
+    {
+        var geographicLevels = (GeographicLevel[])Enum.GetValues(typeof(GeographicLevel));
+        var result = new List<GeographicLevel>();
+
+        while (reader.Read() && reader.TokenType == JsonToken.String)
+        {
+            var value = (string)reader.Value;
+            var geographicLevel = geographicLevels.First(gl => gl.GetEnumValue() == value
+                                                               || gl.GetEnumLabel() == value); // EES-4670 Can be removed after migration done in EES-5140
+            result.Add(geographicLevel);
+        }
+
+        if (reader.TokenType != JsonToken.EndArray)
+        {
+            return null;
+        }
+
+        return result;
+    }
+}
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/DataSetFilesControllerTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/DataSetFilesControllerTests.cs
@@ -6,6 +6,7 @@ using System.Net.Http;
 using System.Threading.Tasks;
 using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Model;
+using GovUk.Education.ExploreEducationStatistics.Common.Model.Data;
 using GovUk.Education.ExploreEducationStatistics.Common.Tests;
 using GovUk.Education.ExploreEducationStatistics.Common.Tests.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Tests.Fixtures;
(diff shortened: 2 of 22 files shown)


mmyoungman:
Nick left a comment on the previous PR: https://github.com/dfe-analytical-services/explore-education-statistics/pull/4868#discussion_r1603658291

```
Would change this to string Period to match how we've modelled it in the public API.

We've purposefully done this because singular dates and date ranges are eventually coming down the pipeline ([EES-3109](https://dfedigital.atlassian.net/browse/EES-3109)).

These new time periods are potentially going in the time_period column, so the current assumptions that we make all over that this is a single numeric year are likely to be incorrect in the future. Modelling our meta summary as suggested helps prevent this being an issue later down the line.
```

ntsim:
In the future, yes it probably makes sense to do a migration to remove `Year` as we'd likely want to consolidate these into a single string `Period` column (or similar).

Will speak with you on Slack about your other changes!

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/DataSetFilesControllerTests.cs
```diff
@@ -1825,7 +1839,18 @@ public async Task FetchDataSetDetails_Success()
...
                 .WithFile(_fixture.DefaultFile()
-                    .WithDataSetFileMeta(_fixture.DefaultDataSetFileMeta())
+                    .WithDataSetFileMeta(_fixture.DefaultDataSetFileMeta()
+                        .WithTimePeriodRange(new TimePeriodRangeMeta
+                        {
+                            Start = new TimePeriodRangeBoundMeta {
+                                TimeIdentifier = TimeIdentifier.CalendarYear,
+                                Period = "2000",
+                            },
+                            End = new TimePeriodRangeBoundMeta {
```
ntsim, line 1849: Formatting for these open braces should be on new lines

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/DataSetFilesControllerTests.cs
```diff
@@ -1600,8 +1600,19 @@ public async Task DataSetFileMetaCorrectlyReturned_Success()
...
+                                {
+                                    TimeIdentifier = TimeIdentifier.AcademicYear,
+                                    Period = "2000",
+                                },
+                                End = new TimePeriodRangeBoundMeta
+                                {
+                                    TimeIdentifier = TimeIdentifier.AcademicYear,
+                                    Period = "2002",
+                                },
+                            })
```
ntsim, line 1615: Would be good to have some generator extensions for this to make this a bit nicer e.g. something like this:

```cs
_fixture.DefaultTimePeriodRangeMeta()
  .WithStart("2000", TimeIdentifier.AcademicYear)
  .WithEnd("2002", TimeIdentifier.AcademicYear)
```

Similar thing applies in the other test case further down

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Fixtures/DataSetFileMetaGeneratorExtensions.cs
```diff
@@ -15,8 +15,11 @@ public static Generator<DataSetFileMeta> WithDefaults(this Generator<DataSetFile
     public static InstanceSetters<DataSetFileMeta> SetDefaults(this InstanceSetters<DataSetFileMeta> setters)
         => setters
             .SetGeographicLevels(new List<string> { "National " })
```
ntsim, line 17: We can change this and the other lists below to collection expressions i.e.

```cs
.SetGeographicLevels(["National"])
```

src/GovUk.Education.ExploreEducationStatistics.Content.Model/DataSetFileMeta.cs
```diff
@@ -11,31 +12,52 @@ public class DataSetFileMeta
     public List<string> GeographicLevels { get; set; } = new();
```
ntsim, line 12: This would probably be better as `List<GeographicLevel>` to provide type safety here.

In the public API model, we also store the level codes (the enum values) in the DB JSON e.g. `NAT`, `LA`, etc, rather than the labels. The codes are not supposed to be changed often, so they are a better choice for this.

This is a more flexible approach that allows us to change the labels we return in the view models, without having to change them in the DB.

Given that we're already running a new migration, would suggest we also update the geographic levels too to store the codes instead.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/DataSetFileMeta.cs
```diff
@@ -11,31 +12,52 @@ public class DataSetFileMeta
...
     [JsonConverter(typeof(TimeIdentifierJsonConverter))]
-    public TimeIdentifier TimeIdentifier { get; set; }
+    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+    public TimeIdentifier? TimeIdentifier { get; set; } // EES-4918 to remove
+
+    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+    public List<int>? Years { get; set; } = new(); // EES-4918 to remove
 
-    public List<int> Years { get; set; } = new();
+    public TimePeriodRangeMeta TimePeriodRange { get; set; } = null!;
```
ntsim, line 21: Mentioned in our chat - we can add `required` across the model properties here and below and remove any unneeded default values.

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/DataSetFilesControllerTests.cs
```diff
@@ -1641,12 +1652,15 @@ public async Task DataSetFileMetaCorrectlyReturned_Success()
                 dataSetFileMetaViewModel.GeographicLevels
                     .AssertDeepEqualTo(originalMeta!.GeographicLevels);
 
-                dataSetFileMetaViewModel.TimePeriod.AssertDeepEqualTo(
-                    new DataSetFileTimePeriodViewModel
+                dataSetFileMetaViewModel.TimePeriodRange.AssertDeepEqualTo(
```
ntsim, line 1655: Don't need `AssertDeepEqualTo` as we're using records, can use standard `Assert.Equal`

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/DataSetFilesControllerTests.cs
```diff
@@ -1641,12 +1652,15 @@ public async Task DataSetFileMetaCorrectlyReturned_Success()
                 dataSetFileMetaViewModel.GeographicLevels
                     .AssertDeepEqualTo(originalMeta!.GeographicLevels);
```
ntsim, line 1653: As below - don't need deep equals, can just use `Assert.Equal`

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/DataSetFilesControllerTests.cs
```diff
@@ -1641,12 +1652,15 @@ public async Task DataSetFileMetaCorrectlyReturned_Success()
...
-                        To = TimePeriodLabelFormatter.Format(originalMeta.Years.Last(), originalMeta.TimeIdentifier),
+                        From = TimePeriodLabelFormatter.Format(
+                            originalMeta.TimePeriodRange.Start.Period,
+                            originalMeta.TimePeriodRange.Start.TimeIdentifier),
+                        To = TimePeriodLabelFormatter.Format(
+                            originalMeta.TimePeriodRange.End.Period,
+                            originalMeta.TimePeriodRange.End.TimeIdentifier),
                     });
 
                 dataSetFileMetaViewModel.Filters
```
ntsim, line 1666: Similar to above - can just use `Assert.Equal`.

Same for indicators below as well!

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/DataSetFilesControllerTests.cs
```diff
@@ -1874,13 +1899,15 @@ public async Task FetchDataSetDetails_Success()
             viewModel.Meta.GeographicLevels
                 .AssertDeepEqualTo(dataSetFileMeta!.GeographicLevels);
```
ntsim, line 1900: Can just use `Assert.Equal` rather than deep equals.

Same thing for the other meta facets below as well.

src/GovUk.Education.ExploreEducationStatistics.Content.ViewModels/DataSetFileMetaViewModel.cs
```diff
@@ -3,14 +3,14 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.ViewModels;
 public record DataSetFileMetaViewModel
 {
     public List<string> GeographicLevels { get; init; } = new();
```
ntsim, line 5: Can add `required` on all these view model properties and remove any unneeded default value - here and below

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Model/DataSetFileMeta.cs
mmyoungman: I've made the change. I ended up doing it with Newtonsoft.Json - my rational being that's what we are currently using for all other ContentDbContext JSON stuff so this keeps it consistent.

src/GovUk.Education.ExploreEducationStatistics.Common/Converters/GeographicLevelsListJsonConverter.cs
```diff
@@ -0,0 +1,50 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Common.Model.Data;
+using Newtonsoft.Json;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Converters;
+
+public class GeographicLevelsListJsonConverter : JsonConverter<List<GeographicLevel>>
```
ntsim, line 10: Might want to change this to `IList` to make it more generic

ntsim (approved):

PR #4887, opened 2024-05-23
EES-5161 Add noindex,nofollow tags to all pages in non-production environments

sambiramairelogic:
As the [brief ](https://dfedigital.atlassian.net/browse/EES-5161) says, all sites on the `dev` and `pre-production` environments should bear the `noindex, nofollow` tags for all robots.

### DIFF
--- a/src/explore-education-statistics-frontend/src/components/PageMeta.tsx
+++ b/src/explore-education-statistics-frontend/src/components/PageMeta.tsx
@@ -30,6 +30,12 @@ const PageMeta = ({
         content="jWf4Mg_pzTOgXDWccGcv9stMsdyptYwHeVpODHdesoY"
       />
 
+      {process.env.APP_ENV !== 'Production' && (
+        <>
+          <meta name="robots" content="noindex,nofollow" />
+          <meta name="googlebot" content="noindex,nofollow" />
+        </>
+      )}
       {/* <!-- Open Graph / Facebook --> */}
       <meta property="og:type" content="website" />
       <meta property="og:title" content={title} />
--- a/src/explore-education-statistics-frontend/src/components/PageMeta.tsx
+++ b/src/explore-education-statistics-frontend/src/components/PageMeta.tsx
@@ -31,10 +31,7 @@ const PageMeta = ({
       />
 
       {process.env.APP_ENV !== 'Production' && (
-        <>
-          <meta name="robots" content="noindex,nofollow" />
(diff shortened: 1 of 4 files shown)


src/explore-education-statistics-frontend/src/components/PageMeta.tsx
```diff
@@ -30,6 +30,12 @@ const PageMeta = ({
         content="jWf4Mg_pzTOgXDWccGcv9stMsdyptYwHeVpODHdesoY"
       />
 
+      {process.env.APP_ENV !== 'Production' && (
+        <>
+          <meta name="robots" content="noindex,nofollow" />
+          <meta name="googlebot" content="noindex,nofollow" />
```
ntsim, line 36: Pretty sure we don't need this as the `robots` tag should be sufficient to block all crawlers, including Google's according to [their docs](https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag): 

> In this example, the robots meta tag instructs search engines not to show the page in search results. The value of the name attribute (robots) specifies that the rule applies to all crawlers.

src/explore-education-statistics-frontend/src/components/PageMeta.tsx
```diff
@@ -30,6 +30,12 @@ const PageMeta = ({
         content="jWf4Mg_pzTOgXDWccGcv9stMsdyptYwHeVpODHdesoY"
       />
 
+      {process.env.APP_ENV !== 'Production' && (
+        <>
+          <meta name="robots" content="noindex,nofollow" />
```
ntsim, line 35: Would suggest we also put this in admin too under it's `public/index.html` file.

This would be applied to every environment, including prod.

src/explore-education-statistics-frontend/src/components/PageMeta.tsx
sambiramairelogic: I just erred on the side of caution and didn't see a downside to supplying both, but fine can remove

src/explore-education-statistics-frontend/src/components/PageMeta.tsx
sambiramairelogic: Certainly happy to, I just didn't think we cared about the admin site from an SEO perspective. Everything except the login page should be giving a 401 anyway I believe so I'm sure its rating is atrocious already :D

ntsim (approved):

PR #4891, opened 2024-05-24
EES-5003 - Create sitemap.xml and robots.txt files

sambiramairelogic:
Uses [next-sitemap](https://www.npmjs.com/package/next-sitemap) to generate sitemap.xml and robots.txt files. 

To generate the dynamic pages, we need new endpoints to fetch all available slugs. These need to be filtered appropriately, I've taken my best guess but would appreciate input. 

I've also taken my best guess when it comes to the priority of pages. I think the only one that really matters is that the root has priority 1 (it's bounded 0:1, with 0.5 the default). 

I've chosen not to generate a Change Frequency parameter for each sitemap entry, but we should look at adding this if we have sufficient information to do so.


src/explore-education-statistics-frontend/src/services/sitemapService.ts
```diff
@@ -0,0 +1,141 @@
...
+    //  find-statistics/[publication]/prerelease-access-list
+    //  find-statistics/[publication]/[release]/data-guidance
+    //  find-statistics/[publication]/[release]/prerelease-access-list
+    //  data-catalogue/[publicationSlug]/[releaseSlug]
+    //  data-tables/[publicationSlug]/[releaseSlug]
+    //  find-statistics/[publication]/[release]
+    ...buildPublicationRoutes(publications),
+
+    // matches dynamic routes following the patterns:
+    //  data-catalogue/data-set/[dataSetFieldId]
```
ntsim, line 36: Similar to lower comment, do we need these comments? They'll get out of sync really easily and they don't really add anything that can't be read in the code itself

src/explore-education-statistics-frontend/src/services/sitemapService.ts
```diff
@@ -0,0 +1,141 @@
...
+        lastmod: publication.lastModified
+          ? `${publication.lastModified}`
+          : undefined,
+      },
+      {
+        loc: `${process.env.PUBLIC_URL}data-tables/${publication.slug}`,
+        lastmod: publication.lastModified
+          ? `${publication.lastModified}`
+          : undefined,
+      },
```
ntsim, line 79: Do these paths need to be indexed? This probably isn't going to help our SEO much and don't feel like pages that should be indexed and show up in search results IMO.

Might want to check this?

src/explore-education-statistics-frontend/src/services/sitemapService.ts
```diff
@@ -0,0 +1,141 @@
...
+          priority: 0.2,
+        },
+        {
+          loc: `${process.env.PUBLIC_URL}data-catalogue/${publication.slug}/${release.slug}`,
+          lastmod: release.lastModified ? `${release.lastModified}` : undefined,
+        },
+        {
+          loc: `${process.env.PUBLIC_URL}data-tables/${publication.slug}/${release.slug}`,
+          lastmod: release.lastModified ? `${release.lastModified}` : undefined,
+        },
```
ntsim, line 127: Similar to othe rcomment - do we need these paths?

src/explore-education-statistics-frontend/src/services/sitemapService.ts
```diff
@@ -0,0 +1,141 @@
...
+    //  find-statistics/[publication]/[release]/data-guidance
+    //  find-statistics/[publication]/[release]/prerelease-access-list
+    //  data-catalogue/[publicationSlug]/[releaseSlug]
+    //  data-tables/[publicationSlug]/[releaseSlug]
+    //  find-statistics/[publication]/[release]
+    ...buildPublicationRoutes(publications),
+
+    // matches dynamic routes following the patterns:
+    //  data-catalogue/data-set/[dataSetFieldId]
+    ...buildDataSetRoutes(dataSets),
```
ntsim, line 37: Would comment this out for the time being and leave a TODO.

These paths are still in development and shouldn't be indexed right now.

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/DataSetFileServiceTests.cs
```diff
@@ -0,0 +1,63 @@
+using System;
```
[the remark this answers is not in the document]
sambiramairelogic: It would be redundant as this is already set for the project as a whole

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/DataSetFilesControllerTests.cs
```diff
@@ -1779,6 +1779,78 @@ public async Task ReleaseFileSummariesContainHtml_HtmlTagsAreStripped()
...
+        {
+            var client = (app ?? BuildApp()).CreateClient();
+
+            return await client.GetAsync("/api/data-set-files/sitemap-items");
+        }
+        
+        [Fact]
+        public async Task ListSitemapItems()
+        {
+            Publication publication = _fixture.DefaultPublication()
```
tomjonesdev, line 1797: Not super important, but I think the preference is to use `var` in variable assignments rather than the type.

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/PublicationControllerTests.cs
```diff
@@ -100,6 +104,45 @@ public async Task GetPublicationTree()
...
+            var controller = BuildPublicationController(publicationService: publicationService.Object);
+
+            var response = await controller.ListSitemapItems();
+            var sitemapItems = response.AssertOkResult();
+
+            var item = Assert.Single(sitemapItems);
+            Assert.Equal("test-publication", item.Slug);
+
+            Assert.NotNull(item);
+            Assert.NotNull(item.Releases);
```
tomjonesdev, line 139: `NotNull` checks don't really add anything here as non-null is inferred by other assertions.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
```diff
@@ -2027,6 +2027,97 @@ await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId)
...
+            private readonly Guid publicationId = Guid.NewGuid();
+
+            private readonly Guid firstReleaseId = Guid.NewGuid();
+            private readonly string firstReleaseSlug = "first-release-slug";
+
+            private readonly Guid secondReleaseId = Guid.NewGuid();
+            private readonly string secondReleaseSlug = "second-release-slug";
+
+            private readonly string firstReleaseVersionPublishedDate = "2019-02-03T07:34:12";
+            private readonly string firstReleaseVersionUpdateDate = "2019-02-04T08:29:54";
```
tomjonesdev, line 2042: If these fields are only used by a single test, then I'd suggest they be declared as part of the test setup, rather than at class-level.

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/PublicationControllerTests.cs
sambiramairelogic: Technically, you're right, but because we're in a `#nullable` context here `item.Releases` generates a possibly null warning unless it's explicitly checked first. Frustrating, I know!

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/DataSetFilesControllerTests.cs
sambiramairelogic: Aye good spot, I wasn't sure which way to go here because the rest of the file uses the explicit types. 

I chose consistency because I know that's something we're really keen on - I'm very happy to switch it over to var, but would you like me to switch over the rest of the extant tests in this file too?

src/GovUk.Education.ExploreEducationStatistics.Content.ViewModels/DataSetSitemapItemViewModel.cs
```diff
@@ -0,0 +1,8 @@
+namespace GovUk.Education.ExploreEducationStatistics.Content.ViewModels;
+
+public record DataSetSitemapItemViewModel
+{
+    public required string Id { get; init; }
+
+    public DateTime? LastModified { get; init; }
```
tomjonesdev, line 7: Should these dates be _non_-nullable? Although a modification date would typically be nullable for most entities, `lastmod` on a sitemap item would always be present because presumably the initial value is when it was first created.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
sambiramairelogic: Yeah I had the same thought myself, but was trying to be consistent with how we do this elsewhere. I'll move them in

tests/playwright-tests/tests/public/seoMetadataFiles.spec.ts
```diff
@@ -0,0 +1,46 @@
+import { test, expect } from '@playwright/test';
+import environment from '@util/env';
+
+const { PUBLIC_URL } = environment;
+
+test.describe('SEO Metadata Files', () => {
+  test('An xml sitemap file can be found at the expected route', async ({
```
tomjonesdev, line 7: Since a lot of the logic relates to setting the URL and last modified date, I'd suggest checking that these values are displaying properly - not all of them of course, maybe just one from each of the main sections since they're handled independently?

tomjonesdev (approved):

PR #4899, opened 2024-05-29
EES-4995 only show type filter if there are api data sets

amyb-hiveit:
Checks to see if there are any API data sets before showing the 'Type of data' filter on the new Data Catalogue.

### DIFF
--- a/src/explore-education-statistics-frontend/src/modules/data-catalogue/DataCataloguePage.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/data-catalogue/DataCataloguePage.tsx
@@ -16,6 +16,7 @@ import { QueryClient, dehydrate } from '@tanstack/react-query';
 import publicationQueries from '@frontend/queries/publicationQueries';
 
 interface Props {
+  showTypeFilter?: boolean;
   releases?: ReleaseSummary[];
   selectedPublication?: PublicationTreeSummary;
   selectedRelease?: ReleaseSummary;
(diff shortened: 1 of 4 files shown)


src/explore-education-statistics-frontend/src/modules/data-catalogue/DataCataloguePage.tsx
```diff
@@ -63,6 +67,13 @@ export const getServerSideProps: GetServerSideProps<Props> = async context => {
         publicationFilter: 'DataCatalogue',
       }),
     );
+
+    if (!showTypeFilter) {
+      const apiDataSets = await queryClient.fetchQuery(
+        dataSetFileQueries.list({ dataSetType: 'api' }),
+      );
+      showTypeFilter = !!apiDataSets.results.length;
```
tomjonesdev, line 75: Not seen the double `!!` before, does it behave similarly to `!apiDataSets.results`?

tomjonesdev (approved):

src/explore-education-statistics-frontend/src/modules/data-catalogue/DataCataloguePage.tsx
amyb-hiveit: It's a shorthand way of doing something like `showTypeFilter = apiDataSets.results.length ? true : false;`. It's essentially using the not operator (`!`) twice to make an inverted boolean. We need to check for the length of array as just checking results exists would return true for an empty results array.

PR #4911, opened 2024-05-30
EES-5119 - adding Managed Identity authentication to the Data Processor

duncan-at-hiveit:
This PR:
* connects a manually-created App Registration to the Data Processor.
* whitelists the Admin App Service and managed identity to be able to access the Function App.
* adds access token requesting into Admin to allow it to authenticate with the Data Processor.
* enables temporary public network access to the Data Processor staging slot in order to be able to perform deploys.

As another prerequisite to running the pipeline, this App Registration must be set up firstly and its Id supplied as a Library variable for the Public API and a Release pipeline variable for the ARM deploy.

## Creating the App Registration

image

image

image

## Updating the pipeline / Library variables

image

### Public API Library

image

### ARM Release pipeline

image

image

(it also needs to be added to the "template override variables" in the usual fashion for the "Validate" and "Deploy" stages.)

### DIFF
--- a/infrastructure/templates/public-api/components/functionApp.bicep
+++ b/infrastructure/templates/public-api/components/functionApp.bicep
@@ -49,6 +49,13 @@ param userAssignedManagedIdentityParams {
   principalId: string
 }?
 
+@description('An existing App Registration registered with Entra ID that will be used to control access to this Function App')
+param entraIdAuthentication {
+  appRegistrationClientId: string
+  allowedClientIds: string[]
+  allowedPrincipalIds: string[]
+}?
+
 @description('Specifies the SKU for the Function App hosting plan')
 param sku object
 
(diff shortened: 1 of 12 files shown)


infrastructure/templates/public-api/components/functionApp.bicep
```diff
@@ -183,58 +190,94 @@ resource slot2FileShare 'Microsoft.Storage/storageAccounts/fileServices/shares@2
...
+ siteConfig: {
+   alwaysOn: alwaysOn ?? null
+   preWarmedInstanceCount: preWarmedInstanceCount ?? null
+   netFrameworkVersion: '8.0'
+   linuxFxVersion: appServicePlanOS == 'Linux' ? 'DOTNET-ISOLATED|8.0' : null
+   keyVaultReferenceIdentity: keyVaultReferenceIdentity
+ }
+ keyVaultReferenceIdentity: keyVaultReferenceIdentity
+ publicNetworkAccess: publicNetworkAccessEnabled ? 'Enabled' : 'Disabled'
+}
```
benoutram, line 212: Formatting might be off by 1 char here? I see more indentation under `siteConfig` than in the parent `commonSiteProperties`.

infrastructure/templates/public-api/components/functionApp.bicep
```diff
@@ -49,6 +49,13 @@ param userAssignedManagedIdentityParams {
   principalId: string
 }?
 
+@description('An existing App Registration registered with Entra ID that will be used to control access to this Function App')
+param entraIdAuthentication {
+  appRegistrationClientId: string
+  allowedClientIds: string[]
+  allowedPrincipalIds: string[]
+}?
```
benoutram, line 57: I was wondering why we can't create the app registration in this template. Is it because it's not supported in Bicep or because our deployment principal has insufficient role permissions?

infrastructure/templates/public-api/components/functionApp.bicep
```diff
@@ -183,58 +190,94 @@ resource slot2FileShare 'Microsoft.Storage/storageAccounts/fileServices/shares@2
...
 
-    clientAffinityEnabled: true
-    reserved: reserved
-    siteConfig: {
-      keyVaultReferenceIdentity: keyVaultReferenceIdentity
+var authSettingsV2Properties = entraIdAuthentication == null ? {} : {
+  globalValidation: {
+    requireAuthentication: true
+    unauthenticatedClientAction: 'Return401'
+  }
```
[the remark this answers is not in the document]
duncan-at-hiveit: Hi yes.  That'll be because I was trying all sorts of different solutions during Friday, and I ended up leaving the auth disabled right at the end.  The health checks has been working successfully prior to this though.  Once I can get a successful deploy out again, I'll show that it's working OK with the auth (although I'm doubting myself now)!

I think that it works by virtue of the pings coming from the host machine itself against the containers running within it, and it is therefore able to bypass the auth layer.

infrastructure/templates/public-api/components/functionApp.bicep
duncan-at-hiveit: Got it, thanks

infrastructure/templates/public-api/components/functionApp.bicep
duncan-at-hiveit: Haha.  Sooo...

ARM and Bicep operate on resources in Subscriptions and Resource Groups, but App Registrations are part of Entra ID, not ARM resources.

The only way to achieve this in ARM / Bicep is to use the "DeploymentScript" resource type, and then use it to run Azure CLI scripts and capture the outputs.  This needs a user-assigned identity creating who is privileged enough to do the actions that you're asking them to do.

Looking into it, there was no way that we'd get permission to have our SPNs be able to create App Registrations against DfE's Entra ID (especially for Prod and Pre-Prod, it's Service Tickets only), so I went down this approach.

benoutram: Ok cool, I understand the separation now, thanks.

benoutram (approved):

PR #4914, opened 2024-05-30
EES-5175 and EES-5177 data catalogue updates

amyb-hiveit:
Updates to the new data catalogue page:
- add the themes guidance modal to the themes filter
- banner text update
- change 'clear filters' to 'reset filters'
- add methodology related link

### DIFF
--- a/src/explore-education-statistics-common/src/components/form/FormSelect.tsx
+++ b/src/explore-education-statistics-common/src/components/form/FormSelect.tsx
@@ -27,8 +27,9 @@ export interface FormSelectProps
   error?: string;
   id: string;
   inline?: boolean;
+  inlineHint?: boolean;
   inputRef?: Ref<HTMLSelectElement>;
-  hint?: string;
+  hint?: string | ReactNode;
   name: string;
   onBlur?: FocusEventHandler;
   onChange?: SelectChangeEventHandler;
@@ -48,6 +49,7 @@ const FormSelect = ({
   error,
   id,
   inline = false,
+  inlineHint = false,
   inputRef,
   hint,
   hideLabel,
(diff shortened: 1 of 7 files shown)


src/explore-education-statistics-common/src/components/form/FormSelect.tsx
```diff
@@ -27,8 +27,9 @@ export interface FormSelectProps
   error?: string;
   id: string;
   inline?: boolean;
+  inlineHint?: boolean;
```
ntsim, line 30: Is this prop necessary? Would have assume that this can be inferred from `inline` and `hint` being set?

src/explore-education-statistics-common/src/components/form/FormSelect.tsx
```diff
@@ -65,24 +67,30 @@ const FormSelect = ({
...
-          className={classNames({ 'govuk-!-margin-right-2': inline })}
-          id={id}
-          label={label}
-          hideLabel={hideLabel}
-        />
+        <div
+          className={classNames({
+            'dfe-flex dfe-justify-content--space-between': inlineHint,
+          })}
+        >
```
ntsim, line 74: Could we make this into a conditional container so we don't get the extra div in the markup when this shouldn't be there?

We could also write a snapshot test to cover this

ntsim (changes requested):

src/explore-education-statistics-common/src/components/form/FormSelect.tsx
amyb-hiveit: It is, they do different things:
- `inline` makes the label and select be inline, in this case the hint is rendered below them (it's usually between the label and select)
`inlineHint` makes the hint be inline with the label, it can only be used when then label and select aren't inline.

I've added comments to Props to explain this.

src/explore-education-statistics-common/src/components/form/FormSelect.tsx
amyb-hiveit: Done

src/explore-education-statistics-common/src/components/form/FormSelect.tsx
```diff
@@ -26,9 +26,19 @@ export interface FormSelectProps
...
   id: string;
+  /**
+   * Renders the label and select inline.
+   * If a hint is provided it appears below the label
+   * and select instead of between them.
+   */
   inline?: boolean;
+  /**
+   * Renders the hint inline with the label.
+   * Has no effect if `inline` is true.
```
ntsim, line 37: Is this actually true? It looks like if `inline` and `inlineHint` are true, then the div applying the `inlineHint` styling still gets applied?

Shouldn't this be a bit more precise so that we don't add the container?

We could also add a test for this.

amyb-hiveit: Yep good point, that's sorted now.

ntsim (approved):

PR #4915, opened 2024-05-30
Ees 5147 add data processor function app health checks

duncan-at-hiveit:
This PR:
- enables Health Checks in the Data Processor Function App.
- adds "Unhealthy" Metric Alerts for production and staging slots.
- simplifies and reduces Data Processor redeploy attempts.

### Healthy

The endpoint will return 200 when everything is healthy.  An example response body:

```
{
    "psqlConnection": {
        "healthy": true,
        "unhealthyReason": null
    },
    "fileShareMount": {
        "healthy": true,
        "unhealthyReason": null
    },
    "healthy": true
}
```

### Unhealthy

The endpoint will return 500 when anything is unhealthy.  An example response body:

```
{
    "psqlConnection": {
        "healthy": false,
        "unhealthyReason": "An exception has been raised that is likely due to a transient failure."
    },
    "fileShareMount": {
        "healthy": true,
        "unhealthyReason": null
    },
    "healthy": false
}
```

### Production slot

image

### Staging slot

image

### Resources

image

### Slack alerts

image


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/HealthCheckFunctions.cs
benoutram, line 1: I noticed that http requests to the health check function are failing and the alert rule is regularly in a fired state for long periods of time.  The Processor app is unavailable with a gateway timeout.

It seems like it might be related to this exception which is being thrown regularly:

```
System.InvalidOperationException: An error was generated for warning 'Microsoft.EntityFrameworkCore.Infrastructure.ManyServiceProvidersCreatedWarning': More than twenty 'IServiceProvider' instances have been created for internal use by Entity Framework. This is commonly caused by injection of a new singleton service instance into every DbContext instance. For example, calling 'UseLoggerFactory' passing in a new instance each time--see https://go.microsoft.com/fwlink/?linkid=869049 for more details. This may lead to performance issues, consider reviewing calls on 'DbContextOptionsBuilder' that may require new service providers to be built. This exception can be suppressed or logged by passing event ID 'CoreEventId.ManyServiceProvidersCreatedWarning' to the 'ConfigureWarnings' method in 'DbContext.OnConfiguring' or 'AddDbContext'.
   at Microsoft.EntityFrameworkCore.Diagnostics.EventDefinition.Log[TLoggerCategory](IDiagnosticsLogger`1 logger, Exception exception)
   at Microsoft.EntityFrameworkCore.Diagnostics.CoreLoggerExtensions.ManyServiceProvidersCreatedWarning(IDiagnosticsLogger`1 diagnostics, ICollection`1 serviceProviders)
   at Microsoft.EntityFrameworkCore.Internal.ServiceProviderCache.<GetOrAdd>g__BuildServiceProvider|4_1(IDbContextOptions _, ValueTuple`2 arguments)
   at Microsoft.EntityFrameworkCore.Internal.ServiceProviderCache.GetOrAdd(IDbContextOptions options, Boolean providerRequired)
   at Microsoft.EntityFrameworkCore.DbContext..ctor(DbContextOptions options)
```

I'm not sure if it's related to this set of changes, or the EES-5119 changes that this builds on, or it it was happening before either set of changes, but just mentioning this here because it's obviously something we need to look at next.

I noticed that immediately after a restart of the function app in Azure, the health check is fine for the first few requests.

I thought that adding this code to the `IHostBuilder` config might give some insights, but for me the error doesn't occur locally and this change makes no difference:

```
.UseDefaultServiceProvider((context, options) => options.ValidateScopes = true)
```

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/HealthCheckFunctions.cs
```diff
@@ -1,63 +1,82 @@
...
-        
-        return Task.CompletedTask;
     }
+
+    public record HealthCheckResponse(HealthCheckSummary PsqlConnection, HealthCheckSummary FileShareMount)
+    {
+        public bool Healthy => PsqlConnection.Healthy && FileShareMount.Healthy;
+    };
+
+    public record HealthCheckSummary(bool Healthy, string? UnhealthyReason = null);
```
benoutram, line 81: I was thinking this alternative implementation:

```
    public record HealthCheckSummary(string? UnhealthyReason = null)
    {
        public bool Healthy => UnhealthyReason == null;
    }
```

would avoid needing the boolean param where there's a possibility that a summary can be constructed as healthy together with an unhealthy reason which wouldn't make sense

 `new HealthCheckSummary(Healthy: true, UnhealthyReason: "feeling a bit unhealthy");`

This is fairly academic though so I'm happy if you don't want to make any changes.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/HealthCheckFunctionTests.cs
```diff
@@ -0,0 +1,54 @@
...
+
+            var httpContext = new DefaultHttpContext();
+            var result = await function.HealthCheck(httpContext.Request);
+
+            var expectedHealthCheckResult = new HealthCheckFunctions.HealthCheckResponse(
+                PsqlConnection: new HealthCheckFunctions.HealthCheckSummary(Healthy: true),
+                FileShareMount: new HealthCheckFunctions.HealthCheckSummary(Healthy: true));
+            
+            result.AssertOkObjectResult(expectedHealthCheckResult);
+            Assert.Equal((int) HttpStatusCode.OK, httpContext.Response.StatusCode);
```
benoutram, line 33: Is an additional status code check required here? I would have though that the `AssertOkObjectResult` check would be doing that.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/HealthCheckFunctions.cs
```diff
@@ -1,63 +1,82 @@
...
+    private async Task<HealthCheckSummary> CheckPsqlConnectionHealthy()
+    {
+        logger.LogInformation("Attempting to test PSQL health");
+
         try
         {
-            var message = $"Found {await publicDataDbContext.DataSets.CountAsync()} datasets.";
-            logger.LogInformation(message);
-            return message;
+            await publicDataDbContext.DataSets.AnyAsync();
```
benoutram, line 46: I doubt there's any difference in effort on the database but if we wanted to make this check super generic and not depend on any particular table we could change this to `await publicDataDbContext.Database.ExecuteSqlRawAsync("SELECT 1");`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/HealthCheckFunctions.cs
```diff
@@ -1,63 +1,82 @@
...
 #pragma warning disable IDE0060
-        [HttpTrigger(AuthorizationLevel.Function, "get")] HttpRequestData request)
+        [HttpTrigger(AuthorizationLevel.Anonymous, "get")] HttpRequest request)
 #pragma warning restore IDE0060
     {
+        var psqlConnectionHealthCheck = await CheckPsqlConnectionHealthy();
+        var fileShareMountHealthCheck = CheckFileShareMountHealth();
+        var healthCheckResponse = new HealthCheckResponse(
+            PsqlConnection: psqlConnectionHealthCheck,
+            FileShareMount: fileShareMountHealthCheck);
```
benoutram, line 28: We've also got connections on the content database and on blob storage, so I'm wondering if you considered also making those part of the health check?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/HealthCheckFunctions.cs
duncan-at-hiveit: Yeah the problem we are facing is that we need to keep supplying new access tokens to the connection string throughout the lifetime of the Function App, and thus we were constructing a new connection string and data source each time.

After some digging, I've now found `NpgsqlDataSourceBuilder.UsePeriodicPasswordProvider` which I think will fit the bill very well, allowing us to create the data source outside of the AddDbContext block.

infrastructure/templates/public-api/components/functionApp.bicep
```diff
@@ -280,6 +287,68 @@ resource stagingSlotAuthSettings 'Microsoft.Web/sites/slots/config@2022-03-01' =
...
+var commonUnhealthyMetricAlertRuleProperties = {
+  enabled: true
+  severity: 1
+  evaluationFrequency: 'PT5M'
+  windowSize: 'PT5M'
+  criteria: {
+    'odata.type': 'Microsoft.Azure.Monitor.SingleResourceMultipleMetricCriteria'
+    allOf: [
+      {
+        name: 'Metric1'
```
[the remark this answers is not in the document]
duncan-at-hiveit: Yeah, seems fairly standard naming convention, and also the convention that we currently have in the ARM template too, so I'm happy to keep as is if that's OK.

infrastructure/templates/public-api/components/functionApp.bicep
```diff
@@ -280,6 +287,68 @@ resource stagingSlotAuthSettings 'Microsoft.Web/sites/slots/config@2022-03-01' =
...
 
+resource alertsActionGroup 'Microsoft.Insights/actionGroups@2023-01-01' existing = {
+  name: '${subscription}-ag-ees-alertedusers'
+}
+
+var commonUnhealthyMetricAlertRuleProperties = {
+  enabled: true
+  severity: 1
+  evaluationFrequency: 'PT5M'
+  windowSize: 'PT5M'
```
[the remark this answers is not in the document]
duncan-at-hiveit: Good point!  Updated to evaluate every minute, and removed the window size config altogether.

infrastructure/templates/public-api/main.bicep
```diff
@@ -381,6 +381,10 @@ module dataProcessorFunctionAppModule 'components/functionApp.bicep' = {
       family: 'EP'
     }
     preWarmedInstanceCount: 1
+    healthCheck: {
+      path: '/api/HealthCheck'
+      unhealthyMetricName: '${subscription}PublicDataProcessorUnhealthy'
+    }
```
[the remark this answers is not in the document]
duncan-at-hiveit: We do!  It is https://dfedigital.atlassian.net/browse/EES-5148

infrastructure/templates/public-api/components/functionApp.bicep
```diff
@@ -280,6 +287,68 @@ resource stagingSlotAuthSettings 'Microsoft.Web/sites/slots/config@2022-03-01' =
...
+    criteria: {
+      allOf: [union(
+        commonUnhealthyMetricAlertRuleProperties.criteria.allOf[0],
+        {
+          metricNamespace: 'Microsoft.Web/sites/slots'
+        }
+      )]
+    }
+  })
+}
```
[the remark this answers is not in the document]
duncan-at-hiveit: Yes I think we'll want all of that, but as part of another ticket.  We'll also want to configure backups for PSQL and the Public API storage account, and all that additional hardening work.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/HealthCheckFunctionTests.cs
duncan-at-hiveit: Yup sorry that was a hangover from previous work - removed now ta

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/HealthCheckFunctions.cs
duncan-at-hiveit: I suppose this gives us the extra advantage that we can also test that the managed identity has correct privileges to access the "public" schema of the DB this way, so I'd probably be keen to keep it this way.

benoutram: That's fine by me, good point.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/HealthCheckFunctions.cs
duncan-at-hiveit: Added in thanks!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/HealthCheckFunctions.cs
duncan-at-hiveit: That did the trick!

infrastructure/templates/public-api/components/functionApp.bicep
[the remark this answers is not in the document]
duncan-at-hiveit: Didn't like the window size being removed, so left at 5M as you suggested

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/HealthCheckFunctions.cs
benoutram: 👍 Just wondering if you are planning to add more tests to `HealthCheckFunctionTests` as currently the only failure variant is for the missing file share? No worries if not though.

duncan-at-hiveit: I couldn't think how to provide failures in the other cases without a decent element of overengineering to make a DbContext fail or shut down storage prior to calling it :)  I guess I *could* just shut down storage, but the DbContexts I reckon would be fiddly!

PR #4937, opened 2024-06-06
EES-5182 redirect data catalogue slug urls

amyb-hiveit:
Release pages have links to the data catalogue on using publication and release slugs, e.g. `/data-catalogue/<publication-slug>/<release-slug>`. This PR adds a redirect on the new version of the data catalogue page so that these links will work correctly. The link itself will be updated as part of the tidy up ticket, [EES-4781](https://dfedigital.atlassian.net/browse/EES-4781), but we'll want to keep the redirect so that bookmarked links will still work.

Also:
- fixes a bug where removing a release filter wasn't updating the data sets list to show all releases
- fixes a UI test that broke when I changed 'clear filters' to 'reset filters'

### DIFF
--- a/tests/robot-tests/tests/admin_and_public/bau/data_catalogue.robot
+++ b/tests/robot-tests/tests/admin_and_public/bau/data_catalogue.robot
@@ -252,6 +252,7 @@ Remove publication filter
 Remove release filter
     user chooses select option    id:filters-form-theme    ${PUPILS_AND_SCHOOLS_THEME_TITLE}
     user chooses select option    id:filters-form-publication    ${PUPIL_ABSENCE_PUBLICATION_TITLE}
+    user chooses select option    id:filters-form-release    ${PUPIL_ABSENCE_RELEASE_NAME}
     user clicks button    ${PUPIL_ABSENCE_RELEASE_NAME}
     user checks page contains button    ${PUPILS_AND_SCHOOLS_THEME_TITLE}
     user checks page contains button    ${PUPIL_ABSENCE_PUBLICATION_TITLE}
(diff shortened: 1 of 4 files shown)


tests/robot-tests/tests/admin_and_public/bau/data_catalogue.robot
```diff
@@ -298,3 +299,21 @@ Validate zip contains correct files
     sleep    8    # wait for file to download
     ${list}=    create list    data/dates.csv    data-guidance/data-guidance.txt
     zip should contain directories and files    ui-tests-data-catalogue-%{RUN_IDENTIFIER}_2021-22-q1.zip    ${list}
+
+Validate data catalogue page redirect from slug based urls
+    environment variable should be set    PUBLIC_URL
+    user navigates to public frontend    %{PUBLIC_URL}/data-catalogue/${PUPIL_ABSENCE_PUBLICATION_SLUG}/2016-17?newDesign=true
+    user waits until h1 is visible    Data catalogue
```
sambiramairelogic, line 306: Happy either way, but I might just be tempted to add an explicit url check using something like 

`user checks url equals    %{PUBLIC_URL}/data-catalogue?publicationId=${PUPIL_ABSENCE_PUBLICATION_ID}&releaseId=2016-17&newDesign=true`

That said, this could fail if the url parameters occur in a different order so maybe we don't actually gain anything from it. Approving either way but wanted to throw the idea out

sambiramairelogic (approved):
Left one optional comment but otherwise looks good :)

tests/robot-tests/tests/admin_and_public/bau/data_catalogue.robot
amyb-hiveit: That's a good idea but I'm not sure it's possible as we'd need to know the GUIDs for the publication and release, which, I'm assuming, are generated when the seed data is created so will be different in different environments.

PR #4938, opened 2024-06-06
EES-5201 Refactor functions in preparation for creating next data set versions

benoutram:
Some of the functions developed as part of EES-4903 for processing the initial data set version will be reusable by functions developed for other activity sequences responsible for creating subsequent data set versions.

The function `CopyCsvFilesTests` was already written with this in mind, but other functions like `ImportMetadata` and probably `HandleProcessingFailure` will end up being used by multiple orchestrators for processing the initial data set version and creating/processing subsequent versions.

This task is to refactor these functions to make them available to multiple different orchestrator functions.

Doing this early will make it easier to work on EES-5152 (writing function tests) and EES-4953 (creating the next data set version) in parallel.

### Other changes

- Rename `DataSetService.CreateDataSetVersion` to `DataSetService.CreateDataSet` after new `DataSetVersionService` was introduced by #4902.
- Rename `CreateInitialDataSetVersionFunction/ CreateInitialDataSetVersion` to `CreateDataSetFunction / CreateDataSet` along with similar renames to request and response types.
- Pass `instanceId` as input to `HandleProcessingFailure`. While not strictly necessary since `instanceId` will be set automatically, this is done to be consistent with the other activity functions.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/ProcessorFunctionsIntegrationTest.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/ProcessorFunctionsIntegrationTest.cs
@@ -153,7 +153,7 @@ public class ProcessorFunctionsIntegrationTestFixture : FunctionsIntegrationTest
     {
         return
         [
-            typeof(CreateInitialDataSetVersionFunction),
+            typeof(CreateDataSetFunction),
             typeof(ProcessInitialDataSetVersionFunction),
             typeof(DeleteDataSetVersionFunction),
             typeof(CopyCsvFilesFunction),
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/ProcessorFunctionsIntegrationTest.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/ProcessorFunctionsIntegrationTest.cs
@@ -1,7 +1,9 @@
 using GovUk.Education.ExploreEducationStatistics.Common.Tests;
 using GovUk.Education.ExploreEducationStatistics.Common.Tests.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model;
 using GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Database;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests.Fixtures;
 using GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Functions;
 using GovUk.Education.ExploreEducationStatistics.Public.Data.Services.Interfaces;
 using Microsoft.EntityFrameworkCore;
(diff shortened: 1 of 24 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/ProcessorTestData.cs
```diff
@@ -0,0 +1,19 @@
...
+
+public record ProcessorTestData(string Filename)
+{
+    private static string DataFilesDirectoryPath => Path.Combine(
+        Assembly.GetExecutingAssembly().GetDirectoryPath(),
+        "Resources",
+        "DataFiles"
+    );
+
+    public static ProcessorTestData AbsenceByCharacteristic2016 => new(nameof(AbsenceByCharacteristic2016));
```
benoutram, line 14: Just remembered that these files aren't Absence By Characteristic 2016 at all, they are a copy of the test files taken from the Public API, so I will need to rename this to something more suitable. I might also revert the structure back to using an additional folder to group the files by as it would better support other test data sets being added to this project that might have several files like the set of parquet files.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/ProcessorFunctionsIntegrationTest.cs
```diff
@@ -32,6 +35,67 @@ public void Dispose()
...
+        params string[] expectedFiles)
+    {
+        var dataSetVersionPathResolver = GetRequiredService<IDataSetVersionPathResolver>();
+        var actualFiles = Directory.GetFiles(dataSetVersionPathResolver.DirectoryPath(dataSetVersion))
+            .Select(Path.GetFileName)
+            .ToArray();
+
+        // Assert that the directory contains the expected files and no others
+        Assert.Equal(expectedFiles.Length, actualFiles.Length);
+        Assert.True(ComparerUtils.SequencesAreEqualIgnoringOrder(expectedFiles, actualFiles));
```
ntsim, line 97: In a recent C# version (not sure), the `IEnumerable<T>.Order` method was introduced which means we don't really need this util here. This can be simplified to just:

```cs
Assert.Equal(expectedFiles.Order(), actualFiles.Order());
```

This is a little better for debugging as well because `Assert.True` won't output the enumerables on failure (unlike `Assert.Equal`).

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/ProcessorTestData.cs
ntsim: Yep! I was about to leave you a comment saying exactly this :smile:

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/ProcessorFunctionsIntegrationTest.cs
benoutram: Cool! Yeah the debug is much better and the length check that I had prior to it is pretty pointless as well. Thanks for the suggestion.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/CreateInitialDataSetVersionFunction.cs
```diff
@@ -30,7 +30,7 @@ public class CreateInitialDataSetVersionFunction(
         var instanceId = Guid.NewGuid();
 
         return await requestValidator.Validate(request, cancellationToken)
-            .OnSuccess(() => dataSetService.CreateDataSetVersion(
+            .OnSuccess(() => dataSetService.CreateDataSet(
```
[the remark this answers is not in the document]
benoutram: Ok yep I think we did say we would change it. I've done this now in the second fixup commit. I will adjust the commit message accordingly later.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/ProcessorTestData.cs
benoutram: 🙂 I've reverted the move of those test data files now and made some changes to `ProcessorTestData`. Changes are in the first fixup commit.

ntsim (approved):

PR #4939, opened 2024-06-06
EES-5138 Add linking and release information to data set version

ntsim:
This PR adds additional linking and release information to data set versions returned by the public API. This allow us to populate the 'API data set version history' section of the data set page in the public frontend.

The API response now looks like:

```json
{
  "file": {
    "id": "8cb89baf-9acd-4310-9f9c-0af306a4ec24"
  },
  "release": {
    "title": "Academic year 2020/2021"
    "slug": "2021-22"
  }
}
```

This functionality relies on a new content API endpoint to list release files (`/api/release-files`) that the public API uses when getting or listing data set versions. This just accepts a list of IDs of release files to filter on currently, but we may want to extend it further for other use cases if required.

## Other changes

- Updated seed data so that any data sets without an actual release file in the content DB are placed under the performance testing publication. A new seed data dump will be uploaded as part of this.
- Removed unneeded `ReleaseName` from `ReleaseSummaryViewModel`


src/GovUk.Education.ExploreEducationStatistics.Content.Services/ReleaseFileService.cs
```diff
@@ -15,62 +15,100 @@
...
+
+            var allowedReleaseVersions = new HashSet<ReleaseVersion>();
+
+            foreach (var releaseVersion in releaseVersions)
+            {
+                if (await userService.CheckCanViewReleaseVersion(releaseVersion).IsRight())
+                {
+                    allowedReleaseVersions.Add(releaseVersion);
+                }
+            }
```
benoutram, line 71: Did you consider returning a forbidden error if any or all of the release versions cannot be viewed?

src/GovUk.Education.ExploreEducationStatistics.Content.ViewModels/DataGuidanceViewModel.cs
```diff
@@ -1,13 +1,15 @@
...
 
 public record DataGuidanceViewModel : ReleaseSummaryViewModel
 {
-    public string DataGuidance { get;  }
+    public required string DataGuidance { get; init; }
 
-    public List<DataGuidanceDataSetViewModel> DataSets { get; }
+    public required List<DataGuidanceDataSetViewModel> DataSets { get; init; }
 
+    [SetsRequiredMembers]
```
benoutram, line 12: Might be good to communicate use of this new attribute to the rest of the team whenever there's a suitable time! 🙂

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/ReleaseFileControllerTests.cs
```diff
@@ -13,40 +15,213 @@
...
 using static GovUk.Education.ExploreEducationStatistics.Common.Services.CollectionUtils;
 using static Moq.MockBehavior;
 
-namespace GovUk.Education.ExploreEducationStatistics.Content.Api.Tests.Controllers
+namespace GovUk.Education.ExploreEducationStatistics.Content.Api.Tests.Controllers;
+
+public abstract class ReleaseFileControllerTests : IntegrationTestFixture
 {
-    public class ReleaseFileControllerTests(TestApplicationFactory testApp) : IntegrationTestFixture(testApp)
+    private ReleaseFileControllerTests(TestApplicationFactory testApp) : base(testApp)
```
benoutram, line 34: Do you need this private constructor on the abstract class? I would have thought you could declare the class with a primary constructor like

```csharp
public abstract class ReleaseFileControllerTests(TestApplicationFactory testApp) : IntegrationTestFixture(testApp)
```

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetService.cs
```diff
@@ -122,23 +126,65 @@ internal class DataSetService(
...
-                .Take(pageSize)
-                .ToListAsync(cancellationToken: cancellationToken)
-            )
-            .Select(MapDataSetVersion)
+        var dataSetVersions = await queryable
+            .OrderByDescending(dsv => dsv.VersionMajor)
+            .ThenByDescending(dsv => dsv.VersionMinor)
+            .Skip((page - 1) * pageSize)
+            .Take(pageSize)
+            .ToListAsync(cancellationToken: cancellationToken);
```
benoutram, line 134: I think you can use `.Paginate(page: page, pageSize: pageSize)` here if you want.

src/GovUk.Education.ExploreEducationStatistics.Content.ViewModels/ReleaseSummaryViewModel.cs
```diff
@@ -1,36 +1,42 @@
...
-    [JsonConverter(typeof(StringEnumConverter))]
-    public ReleaseType Type { get; }
+    public required bool LatestRelease { get; init; }
 
-    public bool LatestRelease { get; }
+    public PublicationSummaryViewModel? Publication { get; init; }
 
-    public PublicationSummaryViewModel? Publication { get; }
+    [JsonConstructor]
+    [Newtonsoft.Json.JsonConstructor]
```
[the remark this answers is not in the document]
ntsim: Yeh good point. It's because I originally had this as a private constructor, but changed it once I realised I could use the `SetsRequiredMembers` attribute!

src/GovUk.Education.ExploreEducationStatistics.Admin/Models/DataFileInfo.cs
```diff
@@ -7,12 +7,9 @@
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Models
 {
-    public class DataFileInfo : FileInfo
+    public record DataFileInfo : FileInfo
```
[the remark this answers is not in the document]
ntsim: Yeh, it's basically a view model. Would prefer to leave it for another PR when someone's more bothered to change it though!

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ReleaseFileService.cs
ntsim: Nope I didn't as I've opted for a filtering approach. It'd be a little confusing for the endpoint to return a forbidden response as it's not particularly informative and won't help you pinpoint where the problem is.

src/GovUk.Education.ExploreEducationStatistics.Content.ViewModels/DataGuidanceViewModel.cs
ntsim: Yep! Although one thing to note - it doesn't provide a compiler error if the required properties aren't filled out by the constructor. Rider will highlight this as a potential inspection warning, but I don't think the compiler will.

Something to be aware of when changing properties on anything that uses these types of constructors.

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/ReleaseFileControllerTests.cs
ntsim: I originally went with this because I was getting this warning:

> Parameter 'testApp' is captured into the state of the enclosing type and its value is also passed to the base constructor. The value might be captured by the base class as well

However, this was fixable by changing the base `IntegrationTestFixture` constructor to be internal.

This is something that we'll probably run into the more we use primary constructors due to their state capturing nature!

src/GovUk.Education.ExploreEducationStatistics.Content.ViewModels/DataGuidanceViewModel.cs
benoutram: Cool, just tried this out and I saw the warning in Rider you mentioned. I will bear this in mind, thanks.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/ReleaseFileService.cs
benoutram: Cool, I think it's fine as it is then. Thanks for going through the reasoning.

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/ReleaseFileControllerTests.cs
benoutram: From reading up on this it sounds like the captured `testApp` variable can be stored multiple times through different levels of inheritance (especially with our multiple levels of subclasses for partitioning our tests).

If a derived class references `testApp` it will be captured as a separate copy leading to a copy in our test class and a separate version of it in the base class. 

It sounds like the recommended way to go is to avoid using the captured variable and use the protected `TestApp` field from the base class like you've already done.

If we ditch the protected `TestApp` property and made it private wouldn't we still have multiple copies of it?
I see the potential for misuse of `testApp` but maybe it's good enough just to be aware of the warning and not use it? 

I see two places where we should correct it in `Admin.Tests.Services.DataSetVersionServiceTests`.

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/ReleaseFileControllerTests.cs
ntsim: Not sure, would have to test it, but my guess would be that because `testApp` is a reference type this should be fine?

As far as I understand it, this wouldn't result in multiple actual copies of it in memory. It's also less likely to be an issue as the `WebApplicationFactory` APIs are immutable anyway.

It's a shame that there isn't functionality to change the privacy of primary constructor parameters, or a way to pipe them through to the base class. Given the potential issues, it actually feels like the private constructor approach is the safer approach.

Anyway, I'll probably do a little investigation at some point and gather some more evidence / thoughts.

PR #4941, opened 2024-06-07
EES-5196 Add data set status and summary to admin details page

ntsim:
This PR adds the data set's status and summary information to the admin's details page:

image


amyb-hiveit (approved):
Could add a test for this to ReleaseApiDataSetDetailsPage.test.tsx, but otherwise all good.

PR #4947, opened 2024-06-11
EES-5194 Add Processor bulk delete endpoint for multiple API data set versions

jack-hive:
We need to add an endpoint to the Processor to delete multiple data set versions in bulk. 

We will be using this endpoint within the deletion plan for an entire release version.

The deletion is blocked if ANY data set versions linked to the target release version are not in a state which can be deleted.

Similar to the existing Processor endpoint which is used to delete SINGLE data set versions, this endpoint is wrapped in a transaction such that all changes will be rolled back if any single deletion step fails. It will attempt to delete the following:
- The `DataSets` linked to the release version
- The `DataSetVersions`, and any related metadata, linked to the release version
- The `DataSetVersionImports` for all the `DataSetVersions` being deleted
- The parquet files (used by DuckDB) associated with ALL of the imports for each `DataSet` being deleted 
- The `Files` linked to the release version will be unlinked from the deleted `DataSets` and `DataSetVersions`

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionService.cs
@@ -100,10 +100,14 @@ internal class DataSetVersionService(
         IReadOnlyList<ReleaseFile> releaseFiles,
         CancellationToken cancellationToken)
     {
+        var releaseFileIds = releaseFiles
+            .Select(rf => rf.Id)
+            .ToList();
+
         return await publicDataDbContext.DataSetVersions
             .AsNoTracking()
             .Include(dsv => dsv.DataSet)
-            .Where(dsv => releaseFiles.Any(rf => rf.Id == dsv.ReleaseFileId))
+            .Where(dsv => releaseFileIds.Contains(dsv.ReleaseFileId))
             .ToListAsync(cancellationToken);
     }
 
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionService.cs
@@ -59,33 +59,23 @@ internal class DataSetVersionService(
 
     public async Task<Either<ActionResult, Unit>> BulkDeleteVersions(Guid releaseVersionId, CancellationToken cancellationToken = default)
     {
-        var strategy = contentDbContext.Database.CreateExecutionStrategy();
+        return await contentDbContext.RequireTransaction(() =>
+            GetReleaseFiles(releaseVersionId, cancellationToken)
+                .OnSuccessCombineWith(async releaseFiles => await GetDataSetVersions(releaseFiles, cancellationToken))
+                .OnSuccess(releaseFilesAndDataSetVersions =>
+                {
+                    var (releaseFiles, dataSetVersions) = releaseFilesAndDataSetVersions;
 
-        return await strategy.ExecuteAsync(
-            async () =>
-            {
-                using var transactionScope = new TransactionScope(
-                    TransactionScopeOption.Required,
-                    new TransactionOptions { IsolationLevel = IsolationLevel.ReadCommitted },
-                    TransactionScopeAsyncFlowOption.Enabled);
-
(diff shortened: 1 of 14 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionService.cs
```diff
@@ -76,6 +143,17 @@ await UpdateFilePublicApiDataSetId(releaseFile, cancellationToken))
...
     }
 
+    private async Task DeleteDataSetVersions(IReadOnlyList<DataSetVersion> dataSetVersions, CancellationToken cancellationToken)
+    {
+        var dataSets = dataSetVersions.Select(dsv => dsv.DataSet);
+
+        publicDataDbContext.DataSetVersions.RemoveRange(dataSetVersions);
+        await publicDataDbContext.SaveChangesAsync(cancellationToken);
+
+        publicDataDbContext.DataSets.RemoveRange(dataSets);
```
duncan-at-hiveit, line 153: I don't think you're safe to delete the DataSets at this point - only if the DataSetVersion getting deleted from it is its only version.

As an example, imagine a DataSet that has a published DataSetVersion on a different Release altogether (e.g. a 2021 Release but you are deleting a 2022 Release).  In this case, you shouldn't delete the DataSet because it still has valid DataSetVersions hanging off it relating to the 2021 Release.

jack-hive: I think we spoke about this and kind of agreed it was safe because:
- If you are deleting a release version, it must be an amendment
- If it's an amendment, until it's published, only the **new** `Files` it's linked to will be linked to a `DataSetVersion` which itself is linked to the new release amendment via the `ReleaseFileId` field
- So, when we find `DataSetVersions` to delete, via the `ReleaseFileId` field, we will only pick up ones that are linked to **new** `Files`
- If they're new `Files`, then the `DataSetVersion` its linked to **has** to be the 1st version of it's series
- Therefore the `DataSet` linked to that `DataSetVersion` is guaranteed to not have any other `DataSetVersions` hanging off it

So we only end up deleting 1st versions of `DataSetVersions`, and hence it is safe to delete the `DataSet`


However.........

This got me thinking, and now I think we missed a trick with the unlinking of the `Files`. We said that was also safe, but it's only safe if I unlink **just** the new `Files`. But I think the way I've written the code, I unlink them all at the minute. Even the old `Files`

duncan-at-hiveit: Hey there, let me respond to some of these:

> * If you are deleting a release version, it must be an amendment

Yup, correct!

> * If it's an amendment, until it's published, only the **new** `Files` it's linked to will be linked to a `DataSetVersion` which itself is linked to the new release amendment via the `ReleaseFileId` field

Yup also correct :)

> * So, when we find `DataSetVersions` to delete, via the `ReleaseFileId` field, we will only pick up ones that are linked to **new** `Files`

Correctamundo once again!

> * If they're new `Files`, then the `DataSetVersion` its linked to **has** to be the 1st version of it's series

Now this isn't correct, so I'll give you an example:

* The Academic Year 2021 release, we will call Release1.  It has a single ReleaseVersion on it which we'll call Release1Version1.  Release1Version1 has a single ReleaseFile, and this is used as a basis for a new DataSet and DataSetVersion.  We'll call this DataSet1 and DataSet1Version1.  This gets published.
* We then a year later start work on the Academic Year 2022 release, which we'll call Release2. It has one ReleaseVersion on it called Release2Version1.
* We do some work on this and it gets published.
* We then create an amendment of it, which we'll call Release2Version2.  We add a ReleaseFile to it and use it as a basis for the DataSet1Version2!  So now we have a ReleaseFile linked to Release2Version2.
* We decide that we want to delete this amendment.
* At this point, I believe your code above would find that DataSet1Version2 is linked to Release2Version2 and assume that it's the only version of that DataSet, which it is not.  DataSet1Version1 is still out there and live, and so it is not safe to delete this DataSet just yet.

> * Therefore the `DataSet` linked to that `DataSetVersion` is guaranteed to not have any other `DataSetVersions` hanging off it
> So we only end up deleting 1st versions of `DataSetVersions`, and hence it is safe to delete the `DataSet`

As in the above example, I think this is incorrect.

duncan-at-hiveit (approved):

PR #4952, opened 2024-06-12
Ees 4944 read next data set version metadata

duncan-at-hiveit:
# Update 19/06/2024

Based on feedback from Nick, we've decided to go with an in-memory-only extraction of metadata from the next DataSetVersion rather than storing them in the database.  As a result of this, the following changes have changed since originally raising this PR:

* We have a new function called "CreateMappings" - this in turn reads all the appropriate meta for generating the diffs and the meta summary.
* In order to gather the required metadata, each BlahMetaRepository in use now has a Read* and a Create* method which under the hood use a bunch of shared code for extracting the metadata. 
* CompleteProcessingFunction is no longer broken out as a separate shared function - instead, it's moved back into ProcessInitialDataSetVersionFunction, and a new CompleteNextDataSetVersionMappingsProcessing function is broken out.

# Overview

This PR covers EES-4944 (the importing of the next data set version's metadata) and EES-4953 (the Admin API endpoint from which to initiate this action).

# Behaviour

## Importing

The import process begins in the same way as the initial DataSet workflow:
1. The Subject CSVs are copied to the file share.
2. The Metadata is imported.

Then it skips the import of the data itself and jumps straight to completion whereupon the DataSetVersionImport is marked as completed and the DataSetVersion set to its next state, which in this case is "Mapping".

The import process is initiated from the Admin endpoint `POST /api/public-data/data-set-versions` which takes a DataSetId and the ReleaseFileId for the next candidate version.

## Validation

This PR adds validation to the choice of ReleaseFile when it receives the import request from Admin.  We'd expect most / all of this to be handled by the dropdown that shows the available Subjects that can be chosen for the next DataSetVersion, but this extra validation is done down in the importer just in case:

There is some shared validation from the initial version import:
1. The chosen ReleaseFile must not already be in use by a DataSet.
2. The chosen ReleaseFile must come from a ReleaseVersion that is currently in Draft.
3. The chosen ReleaseFile must be of type "Data".

And there is some new validation specific to the "next" version import:

1. The chosen ReleaseFile must belong to a Release of the same Publication as the DataSet.
2. The chosen ReleaseFile must not belong to the same Release as any of the DataSet's currrent Versions (i.e. you can't add a version "1.1" from a file on the 2021 Release if version "1.0" also lives on the 2021 Release - they must be from another Release entirely and never from any of its amendments).
3. The chosen DataSet must already have a current live Version.

# Refactoring

## Breaking "CompleteProcessing" into its own file
 
As I've found that the CompleteProcessing stage seemed to be relevant in its current form for both the initial DataSetVersion and "next" Versions for now, I have separated it into its own class for reuse between the "initial" and "next" workflows.  The only change in behaviour for now is the fact that initial versions end up in a "Draft" state whereas next versions end up in 
a "Mapping" state.  It may prove down the line that it's best separated out should the 2 workflows differ significantly for any reason but this is pretty trivial to do.

Ben and I talked about the potential for the Import Metadata Function to break into 2 parts which may also affect the CompleteProcessing Function, but we agreed not to try to tackle that now as it would cause issues on his testing branch. 

# Miscellaneous changes

## Adding DbContext.RequireTransaction extension method

Given that more occurrences of the following code block were starting to appear around the codebase (including in this PR), I broke it out into a DbContext extension method for ease of maintenance and readability of the code.

```
        var strategy = context.Database.CreateExecutionStrategy();

        return await strategy.ExecuteAsync(
            async () =>
            {
                using var transactionScope = new TransactionScope(
                    TransactionScopeOption.Required,
                    new TransactionOptions {IsolationLevel = IsolationLevel.ReadCommitted},
                    TransactionScopeAsyncFlowOption.Enabled);
                
                // Do something with dbs here

                transactionScope.Complete();
            }); 
``` 

now becomes:

```
await context.RequireTransaction(async () => {
  // Do something with dbs here
});
```

I added a whole suite of tests around this that share a transaction amongst different DbContexts and tests the behaviour if shared transactions and rollback, the support for DbContexts that use `EnableRetryOnFailure()`, and nested transactions. 

## Adding EnableRetryOnFailure to all PostgreSQL DbContexts

As this option was available for setting up as we register DbContexts, I have added it to all of our PostgreSQL DbContexts.


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionService.cs
```diff
@@ -23,31 +23,50 @@ internal class DataSetVersionService(
     IDataSetVersionPathResolver dataSetVersionPathResolver
 ) : IDataSetVersionService
 {
+    public async Task<Either<ActionResult, (Guid dataSetId, Guid dataSetVersionId)>> CreateInitialVersion(
```
ntsim, line 26: Does this need to return the `dataSetId` as a tuple? This method already receives a `dataSetId` as a parameter so we already know what this is in the consumer.

Would simplify this by changing the return type to `Task<Either<ActionResult, Guid>>`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionService.cs
```diff
@@ -23,31 +23,50 @@ internal class DataSetVersionService(
...
+        return await GetDataSet(dataSetId, cancellationToken)
+            .OnSuccess(ValidateInitialDataSet)
+            .OnSuccess(dataSet => CreateDataSetVersion(
+                releaseFileId: releaseFileId,
+                instanceId: instanceId,
+                dataSet: dataSet, 
+                cancellationToken: cancellationToken));
+    }
+
+    public async Task<Either<ActionResult, (Guid dataSetId, Guid dataSetVersionId)>> CreateNextVersion(
```
ntsim, line 41: Same as above - no need to return the `dataSetId` in a tuple when we already know what the `dataSetId` is in the consumer.

Would simplify this by changing the return type to `Task<Either<ActionResult, Guid>>`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionService.cs
```diff
@@ -23,31 +23,50 @@ internal class DataSetVersionService(
...
-            });
+        return await publicDataDbContext.RequireTransaction(() => 
+            GetDataSetVersion(dataSetVersionId, cancellationToken)
+                .OnSuccessDo(CheckCanDeleteDataSetVersion)
+                .OnSuccessDo(async dataSetVersion => await GetReleaseFile(dataSetVersion, cancellationToken)
+                    .OnSuccessVoid(async releaseFile =>
+                        await UpdateFilePublicApiDataSetId(releaseFile, cancellationToken))
+                    .OnFailureDo(_ =>
+                        throw new KeyNotFoundException(
+                            $"The expected 'ReleaseFile', with ID '{dataSetVersion.ReleaseFileId}', was not found.")))
```
ntsim, line 67: This has been changed on dev - need to rebase.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionService.cs
```diff
@@ -123,4 +142,264 @@ private void DeleteParquetFiles(DataSetVersion dataSetVersion)
...
+    {
+        var dataSet = await publicDataDbContext
+            .DataSets
+            .Include(dataSet => dataSet.LatestLiveVersion)
+            .Include(dataSet => dataSet.Versions)
+            .FirstOrDefaultAsync(dataSet => dataSet.Id == dataSetId, cancellationToken);
+
+        return dataSet is null
+            ? ValidationUtils.ValidationResult(CreateDataSetIdError(
+                message: ValidationMessages.FileNotFound,
```
ntsim, line 158: Should be `ValidationMessages.DataSetNotFound`?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionService.cs
```diff
@@ -123,4 +142,264 @@ private void DeleteParquetFiles(DataSetVersion dataSetVersion)
...
+        {
+            errors.Add(CreateReleaseFileIdError(
+                message: ValidationMessages.NoMetadataFile,
+                releaseFileId: releaseFile.Id));
+        }
+
+        return errors.Count == 0 ? Unit.Instance : ValidationUtils.ValidationResult(errors);
+    }
+    
+    private async Task<Either<ActionResult, Unit>> ValidateReleaseFileAndDataSet(
```
ntsim, line 279: Bit confusing to have another validation method as it looks like we could just combine this all into `ValidateReleaseFile`?

Would suggest we merge these two methods together.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionService.cs
```diff
@@ -123,4 +142,264 @@ private void DeleteParquetFiles(DataSetVersion dataSetVersion)
...
+        List<ErrorViewModel> errors = [];
+
+        if (releaseFile.ReleaseVersion.PublicationId != dataSet.PublicationId)
+        {
+            errors.Add(CreateReleaseFileIdError(
+                message: ValidationMessages.NextReleaseFileMustBeForSamePublicationAsDataSet,
+                releaseFileId: releaseFile.Id));
+        }
+
+        var historicReleaseFileIds = dataSet
```
ntsim, line 293: We don't typically use 'historical' to describe things, and given my earlier comment around the validation message, perhaps `previousReleaseFileIds` would be more consistent.

Same could be applied to `historicalReleaseIds` (i.e. `previousReleaseIds`) below

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionService.cs
```diff
@@ -123,4 +142,264 @@ private void DeleteParquetFiles(DataSetVersion dataSetVersion)
...
+    }
+
+    private async Task CreateDataSetVersionImport(
+        DataSetVersion dataSetVersion,
+        Guid instanceId,
+        CancellationToken cancellationToken)
+    {
+        var dataSetVersionImport = new DataSetVersionImport
+        {
+            DataSetVersionId = dataSetVersion.Id, InstanceId = instanceId, Stage = DataSetVersionImportStage.Pending
```
ntsim, line 350: Formatting in this file (as per other comments)

ntsim (changes requested):

ntsim:
This branch really needs to be rebased onto dev as it's missing some significant new changes. Think there's probably a fair bit of merge conflict pain heading your way!

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetsControllerTests.cs
```diff
@@ -450,15 +450,9 @@ public async Task PageSizeInAllowedRange_Success(int pageSize)
...
-                },
-                {
-                    "pageSize", pageSize?.ToString()
-                },
-                {
-                    "publicationId", publicationId.ToString()
-                },
+                {"page", page?.ToString()},
+                {"pageSize", pageSize?.ToString()},
+                {"publicationId", publicationId.ToString()},
```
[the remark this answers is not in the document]
duncan-at-hiveit: Seems to no longer be an issue after rebase, thanks!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/Interfaces/IDataSetMetaService.cs
```diff
@@ -1,8 +1,19 @@
...
+
     Task CreateDataSetVersionMeta(
         Guid dataSetVersionId,
         CancellationToken cancellationToken = default);
 }
+
+public record DataSetVersionMappingMeta(
+    List<(FilterMeta, List<FilterOptionMeta>)> Filters,
+    List<(LocationMeta, List<LocationOptionMetaRow>)> Locations,
+    DataSetVersionMetaSummary MetaSummary);
```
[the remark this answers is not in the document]
duncan-at-hiveit: Okeydokes - done!

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Public.Data/ProcessorClientTests.cs
```diff
@@ -45,9 +45,7 @@ public async Task HttpClientSuccess()
         {
             var responseBody = new CreateDataSetResponseViewModel
             {
-                DataSetId = Guid.NewGuid(),
-                DataSetVersionId = Guid.NewGuid(),
-                InstanceId = Guid.NewGuid()
```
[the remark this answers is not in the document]
duncan-at-hiveit: Done ta

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Public.Data/ProcessorClientTests.cs
```diff
@@ -129,96 +122,4 @@ private enum Errors
...
-            });
-
-            _mockHttp.VerifyNoOutstandingExpectation();
-        }
-
-        private enum Errors
-        {
-            Error1,
-        }
-    }
```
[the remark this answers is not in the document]
duncan-at-hiveit: After rebasing we have it all back ta

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/ProcessorClient.cs
```diff
@@ -32,52 +32,77 @@ internal class ProcessorClient(
...
-        await AddBearerToken(cancellationToken);
-
-        var request = new DataSetCreateRequest
-        {
-            ReleaseFileId = releaseFileId,
-        };
-
-        var response = await httpClient
-            .PostAsJsonAsync("api/CreateDataSet", request, cancellationToken: cancellationToken);
+        var request = new DataSetCreateRequest {ReleaseFileId = releaseFileId};
```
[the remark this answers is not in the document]
duncan-at-hiveit: Done ta

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/DbContextTransactionExtensionsTests.cs
```diff
@@ -0,0 +1,395 @@
...
+            });
+
+        private async Task WriteToAllDbContexts(int id = 1)
+        {
+            await dbContext1.Entities.AddAsync(new TestEntity {Id = id});
+            await dbContext1.SaveChangesAsync();
+            await dbContext2.Entities.AddAsync(new TestEntity {Id = id});
+            await dbContext2.SaveChangesAsync();
+            await dbContext3WithoutRetry.Entities.AddAsync(new TestEntity {Id = id});
+            await dbContext3WithoutRetry.SaveChangesAsync();
```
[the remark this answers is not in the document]
duncan-at-hiveit: Done ta

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Requests/Validators/ValidationMessages.cs
```diff
@@ -6,32 +6,61 @@ namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Reque
...
+
+    public static readonly LocalizableMessage NextReleaseFileMustBeForSamePublicationAsDataSet = new(
+        Code: nameof(NextReleaseFileMustBeForSamePublicationAsDataSet),
+        Message: "The file must belong to the same publication as the data set."
+    );
+
+    /// <summary>
+    /// This rule ensures that we're not creating the next data set version based on a release file that is an
+    /// amendment of the release that the currently live data set version   
+    /// </summary>
```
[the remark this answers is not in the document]
duncan-at-hiveit: This was the only rule where I felt like it wasn't immediately obvious why the restriction was put in place, hence I decided to give it some extra doco.  Happy to remove if you feel like the message is clear enough after rewording.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionService.cs
duncan-at-hiveit: Seen that during rebase, thanks.  Updated

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/CreateNextDataSetVersionFunctionTests.cs
```diff
@@ -0,0 +1,487 @@
...
+            
+            // Assert the response view model values match the created data set version and import.
+            Assert.Equal(updatedDataSet.Id, responseViewModel.DataSetId);
+            Assert.Equal(nextDataSetVersion.Id, responseViewModel.DataSetVersionId);
+            Assert.Equal(dataSetVersionImport.InstanceId, responseViewModel.InstanceId);
+            
+            // Assert the processing orchestrator was scheduled with the correct arguments
+            Assert.NotNull(processNextDataSetVersionContext);
+            Assert.NotNull(startOrchestrationOptions);
+            Assert.Equal(new ProcessDataSetVersionContext {DataSetVersionId = nextDataSetVersion.Id},
```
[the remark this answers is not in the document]
duncan-at-hiveit: Done ta

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/CreateNextDataSetVersionFunctionTests.cs
```diff
@@ -0,0 +1,487 @@
...
+            validationProblem.AssertHasError(
+                expectedPath: nameof(NextDataSetVersionCreateRequest.ReleaseFileId).ToLowerFirst(),
+                expectedCode: ValidationMessages.ReleaseFileMustBeFromDifferentReleaseToHistoricalVersions.Code
+            );
+        }
+
+        private async Task<(DataSet, DataSetVersion)> AddDataSetAndLatestLiveVersion()
+        {
+            Publication publication = DataFixture
+                .DefaultPublication();
```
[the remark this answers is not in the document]
duncan-at-hiveit: Or with no publication id of any sort as it goes, thanks!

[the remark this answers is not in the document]
duncan-at-hiveit: Or with no publication id of any sort as it goes, thanks!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ProcessNextDataSetVersionFunctionTests.cs
```diff
@@ -0,0 +1,265 @@
...
+using GovUk.Education.ExploreEducationStatistics.Common.Tests.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Database;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests.Fixtures;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Functions;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Model;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Services.Interfaces;
+using Microsoft.DurableTask;
+using Microsoft.DurableTask.Entities;
```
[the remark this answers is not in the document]
duncan-at-hiveit: Cleaned up ta

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/ProcessorTestData.cs
```diff
@@ -17,5 +25,22 @@ public record ProcessorTestData(string Name)
...
+            {
+                Code = TimeIdentifier.AcademicYear,
+                Period = "2020/2021"
+            },
+            End = new TimePeriodRangeBound
+            {
+                Code = TimeIdentifier.AcademicYear,
+                Period = "2022/2023"
+            }
+        });
```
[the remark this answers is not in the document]
duncan-at-hiveit: Haha yeah was happy to see that we'd both found the need for it!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/ProcessorClient.cs
```diff
@@ -32,50 +32,73 @@ internal class ProcessorClient(
...
+        return await SendRequest<Unit>(
+            () => httpClient.DeleteAsync(
+                url,
+                cancellationToken: cancellationToken),
+            cancellationToken);
     }
 
-    public async Task<Either<ActionResult, Unit>> DeleteDataSetVersion(
-        Guid dataSetVersionId,
+    private async Task<Either<ActionResult, TResponse>> SendRequest<TResponse>(
```
jack-hive, line 94: I was speaking to Ben about this on one of my previous PR's. We were torn which way to go but in the end we decided not to break the handling of the HTTP response out into a common method - because it felt like each public method should be responsible for handling it's HTTP response individually, as the way it handles it **may** differ from endpoint to endpoint. For example, the `DeleteDataSetVersion` endpoint previously handled `NotFound` status codes; whereas now that case would fall into a `500`.

I'm still torn which way I'd go on it personally. It does save a lot of code, that is mostly just reused. But at the same time it feels like each endpoint should be handled individually as there isn't technically anything to say they won't be different. They're just the same now by coincidence (aside from `DeleteDataSetVersion` like I mentioned)

src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/DbContextTransactionExtensions.cs
```diff
@@ -0,0 +1,113 @@
...
+    /// own transaction open.
+    ///
+    /// Note that if you have a unit of work that requires interaction with multiple
+    /// DbContexts and any use the EnableRetryOnFailure() strategy, you should use any
+    /// one of the DbContexts that use this strategy to call "RequireTransaction" on,
+    /// and thereafter any other DbContexts that also use the EnableRetryOnFailure()
+    /// strategy will be functional. Not doing this will result in
+    /// InvalidOperationExceptions when those DbContexts are used. 
+    /// </summary>
+    public static async Task<TResult> RequireTransaction<TDbContext, TResult>(
```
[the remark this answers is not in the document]
jack-hive: Just my 2 cents --> I think `RequireTransaction` is more clear as to what it is doing. I think `RunTransaction` could be interpreted as always opening up a new transaction when called. But we're actually, by default, only opening a new transaction if there isn't already an ambient transaction in scope. `RunTransaction` would also suggest that the transaction is complete after the unit of work within scope is complete - but if there is a root (ambient) transaction in scope then the transaction isn't actually complete until the outer scope calls `transactionScope.Complete`.

If we choose to add an an optional parameter to change the `TransactionScopeOption`, then we may have to consider another name. Because `TransactionScopeOption.Suppress` causes the unit of work to take part under **NO** transaction, even if there is an ambient transaction present. In this case, I'd suggest both `RequireTransaction` and `RunTransaction` are both misleading. I can't think of a good generic name that would cover all scenarios though. So I'd suggest maybe we don't create an optional parameter, and instead create additional extension methods when the scenario arises whereby we need to use the other `TransactionScopeOptions`? So keep `RequireTransaction` for `TransactionScopeOption.Required`, create a new one (only when we need to) for `TransactionScopeOption.RequiresNew` called `RequiresNewTransaction`, and create a new one (only when we need to) for `TransactionScopeOption.Suppress` called `SuppressTransaction`.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/ActivityNames.cs
```diff
@@ -4,9 +4,17 @@ internal static class ActivityNames
...
 
     public const string ImportData = nameof(ProcessInitialDataSetVersionFunction.ImportData);
     public const string WriteDataFiles = nameof(ProcessInitialDataSetVersionFunction.WriteDataFiles);
-    public const string CompleteProcessing = nameof(ProcessInitialDataSetVersionFunction.CompleteProcessing);
+
+    public const string CompleteInitialDataSetVersionProcessing = 
+        nameof(ProcessInitialDataSetVersionFunction.CompleteInitialDataSetVersionProcessing);
+
+    public const string CompleteNextDataSetVersionMappingProcessing = 
+        nameof(ProcessNextDataSetVersionFunction.CompleteNextDataSetVersionMappingProcessing);
```
benoutram, line 19: There was spacing before to divide the activities which are generic with the activities which relate to processing the initial version. The spacing seems ad-hoc now so I think it would be good to try and group them again.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/ProcessInitialDataSetVersionFunction.cs
```diff
@@ -69,9 +67,9 @@ public class ProcessInitialDataSetVersionFunction(
         await UpdateImportStage(dataSetVersionImport, DataSetVersionImportStage.WritingDataFiles, cancellationToken);
         await parquetService.WriteDataFiles(dataSetVersionImport.DataSetVersionId, cancellationToken);
     }
-
-    [Function(ActivityNames.CompleteProcessing)]
-    public async Task CompleteProcessing(
+
```
benoutram, line 70: Minor one, there's some whitespace crept in here into `ProcessInitialDataSetVersionFunction`.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/DbContextTestExtensions.cs
```diff
@@ -0,0 +1,32 @@
+using System;
```
benoutram, line 1: `#nullable enable` is missing here

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/DbContextTestExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+public static class DbContextTestExtensions
+{
+    public static async Task ClearTestData<TDbContext>(this TDbContext context) where TDbContext : DbContext
+    {
+        if (context.Database.IsNpgsql())
+        {
+            var tables = context.Model.GetEntityTypes()
+                .Select(type => type.GetTableName())
+                .Distinct()
+                .ToList();
```
benoutram, line 17: `GetTableName` can return null her.  In some places where this code has been moved from we've handled this with `.Cast<string>()` so you could add that in here?

ntsim: Would suggest using `OfType<string>()` instead. When using `Cast`, if the type is incorrect it can throw an exception, whereas `OfType` just filters unexpected types out

benoutram: Great, thanks @ntsim.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/DbContextTransactionExtensionsTests.cs
```diff
@@ -0,0 +1,395 @@
...
+        public async Task InitializeAsync() =>
+            await Task.WhenAll(PostgreSqlContainers.SelectAsync(async container =>
+            {
+                await container.StartAsync();
+                await container.ExecScriptAsync("ALTER SYSTEM SET max_prepared_transactions = 100");
+                await container.StopAsync();
+                await container.StartAsync();
+                await container.ExecScriptAsync("""CREATE TABLE IF NOT EXISTS "Entities" ("Id" int PRIMARY KEY)""");
+                return Task.CompletedTask;
+            }));
```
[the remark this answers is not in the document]
ntsim: Ah right, fair enough!

src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/DbContextTransactionExtensions.cs
```diff
@@ -0,0 +1,113 @@
...
+        this TDbContext context,
+        Func<Task<TResult>> transactionalUnit)
```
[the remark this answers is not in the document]
ntsim: Happy to leave it for now as it'd be easy to add in later anyway

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/CreateNextDataSetVersionFunctionTests.cs
```diff
@@ -0,0 +1,499 @@
...
+            // are untouched.
+            Assert.Equal(DataSetStatus.Published, updatedDataSet.Status);
+            Assert.Equal(dataSet.Title, updatedDataSet.Title);
+            Assert.Equal(dataSet.Summary, updatedDataSet.Summary);
+            Assert.Equal(dataSet.PublicationId, updatedDataSet.PublicationId);
+
+            // Assert that the new DataSetVersion is set as the latest draft version, and that the
+            // prior DataSetVersion is still set as the latest live version. 
+            Assert.NotNull(updatedDataSet.LatestDraftVersion);
+            Assert.NotEqual(liveDataSetVersion.Id, updatedDataSet.LatestDraftVersion!.Id);
```
jack-hive, line 86: would this be better as:
`Assert.Equal(responseViewModel.DataSetVersionId, updatedDataSet.LatestDraftVersion!.Id);`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/CreateNextDataSetVersionFunctionTests.cs
```diff
@@ -0,0 +1,499 @@
...
+                .WithFiles([
+                    DataFixture
+                        .DefaultFile(FileType.Data)
+                        .WithSubjectId(subjectId),
+                    DataFixture
+                        .DefaultFile(FileType.Metadata)
+                        .WithSubjectId(subjectId)
+                ])
+                .GenerateList()
+                .ToTuple2();
```
jack-hive, line 167: Could maybe use your `AddDataAndMetadataFiles` method here to shorten?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/FilterMetaRepository.cs
```diff
@@ -128,4 +128,56 @@ await optionTable
             }
         }
     }
+
+    private async Task<List<FilterMeta>> ReadFilterMeta(
```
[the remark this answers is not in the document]
ntsim: Not sure if I agree with the reasoning. It's also inconsistent with `LocationMetaRepository` which has a `GetLocationOptionMetas` which is getting the options in a very similar similar way.

Maybe we could just make this more explicit as `GetFilterMetasFromCsv`?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ProcessNextDataSetVersionFunctionTests.cs
```diff
@@ -0,0 +1,278 @@
...
+                .WithDataSetVersionId(initialDataSetVersion.Id)
+                .GenerateList(4);
+
+            await AddTestData<PublicDataDbContext>(context =>
+            {
+                context.LocationMetas.AddRange(initialLocationMeta);
+                context.FilterMetas.AddRange(initialFilterMeta);
+                context.IndicatorMetas.AddRange(initialIndicatorMeta);
+                context.TimePeriodMetas.AddRange(initialTimePeriodMeta);
+            });
```
jack-hive, line 171: Don't think you need any of this now since you've changed it to do the work in-memory

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/LocationMetaRepository.cs
```diff
@@ -153,16 +146,58 @@ FROM read_csv('{dataSetVersionPathResolver.CsvDataPath(dataSetVersion):raw}', AL
         }
     }
 
+    private async Task<List<LocationOptionMetaRow>> GetLocationOptionMetas(
+        IDuckDbConnection duckDbConnection,
+        DataSetVersion dataSetVersion,
+        CancellationToken cancellationToken,
```
ntsim, line 152: Lint here - the cancellation token should be the last parameter

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/DbContextTransactionExtensionsTests.cs
```diff
@@ -0,0 +1,395 @@
...
+
+    private void AssertNoEntitiesInAnyDbContexts()
+    {
+        var dbContext1 = TestApp.Services.GetRequiredService<TestDbContext1>();
+        var dbContext2 = TestApp.Services.GetRequiredService<TestDbContext2>();
+        var dbContext3 = TestApp.Services.GetRequiredService<TestDbContext3WithoutRetry>();
+        Assert.Empty(dbContext1.Entities);
+        Assert.Empty(dbContext2.Entities);
+        Assert.Empty(dbContext3.Entities);
+    }
```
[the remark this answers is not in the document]
ntsim: Hmm, but in this context, we don't need to use it again within the method, so it's safe to dispose?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ProcessNextDataSetVersionFunctionTests.cs
```diff
@@ -0,0 +1,278 @@
...
+                dataSet.Id == initialDataSetVersion.DataSet.Id);
+
+            var (nextDataSetVersion, instanceId) = await CreateDataSetVersionAndImport(
+                dataSet: dataSet,
+                importStage: Stage.PreviousStage(),
+                versionMajor: 1,
+                versionMinor: 1);
+
+            var dataSetVersionPathResolver = GetRequiredService<IDataSetVersionPathResolver>();
+            Directory.CreateDirectory(dataSetVersionPathResolver.DirectoryPath(nextDataSetVersion));
```
jack-hive, line 256: don't think you need these 2 lines?

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/LocationMetaRepository.cs
```diff
@@ -153,16 +146,58 @@ FROM read_csv('{dataSetVersionPathResolver.CsvDataPath(dataSetVersion):raw}', AL
         }
     }
 
+    private async Task<List<LocationOptionMetaRow>> GetLocationOptionMetas(
```
ntsim, line 149: As per our conversation about the naming of a similar method in `FilterMetaRepository`, if we go with a more explicit naming of `GetFilterMetasFromCsv`, we should adjust the naming in this repository as well.

The same could be done for `TimePeriodMetaRepository`, `GeographicLevelMetaRepository` as well.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/DbContextTransactionExtensionsTests.cs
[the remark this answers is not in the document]
duncan-at-hiveit: Yeah it's fine to.  It adds a decent number of extra lines to the test suite as a whole but it's fine.  I've added usings in everywhere.  It does give a warning when trying to use a looked-up context within an `Assert.ThrowsAsync(() => RequireTransaction(...))` block as you'll see, whereas it's fine when it's just within a single `RequireTransaction(...)` block.

src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/DbContextTransactionExtensions.cs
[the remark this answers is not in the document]
duncan-at-hiveit: Yeah that was along my line of thinking too as it goes @jack-hive.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/FilterMetaRepository.cs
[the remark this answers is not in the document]
duncan-at-hiveit: No worries - changed to GetFilterMetas ta

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/FilterMetaRepository.cs
```diff
@@ -128,4 +128,56 @@ await optionTable
...
+                    PublicId = row.ColName,
+                    DataSetVersionId = dataSetVersion.Id,
+                    Label = row.Label,
+                    Hint = row.FilterHint ?? string.Empty
+                }
+            )
+            .ToList();
+    }
+
+    private async Task<List<FilterOptionMeta>> ReadFilterOptionMeta(
```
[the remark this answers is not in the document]
duncan-at-hiveit: Already done :)

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
```diff
@@ -41,13 +43,53 @@ public async Task<List<DataSetVersionStatusSummary>> GetStatusesForReleaseVersio
...
+    }
+
+    public async Task<bool> FileHasVersion(
+        Guid releaseFileId,
+        CancellationToken cancellationToken = default)
+    {
+        return await publicDataDbContext.DataSetVersions
+            .AsNoTracking()
+            .AnyAsync(dsv => dsv.ReleaseFileId == releaseFileId, cancellationToken);
+    }
```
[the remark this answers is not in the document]
duncan-at-hiveit: Aha right you are, good spot!  Sorted

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/ProcessInitialDataSetVersionFunction.cs
duncan-at-hiveit: Heh - I'm not seeing what you're seeing I'm afraid.  I have a method `CompleteInitialDataSetVersionProcessing` which has no additional whitespace under it

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/DbContextTestExtensions.cs
duncan-at-hiveit: Nice one.  Added in and also to DbContextTransactionExtensions as well.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/DbContextTestExtensions.cs
duncan-at-hiveit: Cool added that, thanks @benoutram

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/LocationMetaRepository.cs
duncan-at-hiveit: Cool, amended thanks

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetVersion.cs
```diff
@@ -11,6 +11,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Model;
 
 public class DataSetVersion : ICreatedUpdatedTimestamps<DateTimeOffset, DateTimeOffset?>
 {
+    public const string FirstVersionString = "1.0";
```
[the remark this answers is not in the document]
duncan-at-hiveit: Haha.  You know what, it was because I was originally using it in tests, but have since rolled that back out.  I'm just going to return it to the way it was - no const.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionService.cs
```diff
@@ -122,4 +142,266 @@ private void DeleteParquetFiles(DataSetVersion dataSetVersion)
...
+        CancellationToken cancellationToken)
+    {
+        var dataSetVersion = new DataSetVersion
+        {
+            ReleaseFileId = releaseFile.Id,
+            DataSetId = dataSet.Id,
+            Status = DataSetVersionStatus.Processing,
+            Notes = "",
+            VersionMajor = dataSet.LatestLiveVersion?.VersionMajor ?? 1,
+            VersionMinor = dataSet.LatestLiveVersion?.VersionMinor + 1 ?? 0
```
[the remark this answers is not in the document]
duncan-at-hiveit: Yeah we initially create it as a minor version update, but it's only when the mapping is complete that we actually decide on what the version should be.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/DbContextTransactionExtensionsTests.cs
```diff
@@ -0,0 +1,364 @@
...
+    private class SimulateFailureException : Exception;
+
+    // ReSharper disable once ClassNeverInstantiated.Global
+    public class DbContextTransactionExtensionsTestFixture :
+        IAsyncLifetime
+    {
+        public readonly PostgreSqlContainer[] PostgreSqlContainers =
+            Enumerable.Range(0, 3).Select(_ => new PostgreSqlBuilder()
+                    .WithImage("postgres:16.1-alpine")
+                    .WithCommand("-c", "max_prepared_transactions=100")
```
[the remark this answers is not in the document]
duncan-at-hiveit: Haha that's what I said!

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/DbContextTransactionExtensionsTests.cs
```diff
@@ -0,0 +1,364 @@
...
+        public DbSet<TestEntity> Entities { get; init; } = null!;
+    }
+
+    private class SimulateFailureException : Exception;
+
+    // ReSharper disable once ClassNeverInstantiated.Global
+    public class DbContextTransactionExtensionsTestFixture :
+        IAsyncLifetime
+    {
+        public readonly PostgreSqlContainer[] PostgreSqlContainers =
```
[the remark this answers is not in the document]
duncan-at-hiveit: I suspect it would, but I think I would err on leaving that til a different PR if that's OK.  Given we don't have a Test Container for SQL Server just yet, we might have to rely on manual testing of this initially.

src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/DbContextTransactionExtensions.cs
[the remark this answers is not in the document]
ntsim: Cool, seems like a fair analysis. To be honest, I hadn't really appreciated what all the different scopes did!

It might get a little trickier if we start using the different types of scopes. There's a fair bit of nuance and complexity to this stuff so we'd want to avoid exposing too many footguns for people to trip up on.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/CreateNextDataSetVersionFunctionTests.cs
duncan-at-hiveit: I wanted to test it against the actual entity that was saved, as testing against `responseViewModel.DataSetVersionId` would be a false positive if `responseViewModel.DataSetVersionId` was set incorrectly.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ProcessNextDataSetVersionFunctionTests.cs
duncan-at-hiveit: We're only reading the *next* data set version's meta in-memory, but the original data set version's meta is still read from the db as it's already in there and ready to go!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/LocationMetaRepository.cs
duncan-at-hiveit: I went for the simple "Get" in the end, so leaving this be :)

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ProcessNextDataSetVersionFunctionTests.cs
duncan-at-hiveit: 100% correct!  Removed thanks

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/CreateNextDataSetVersionFunctionTests.cs
duncan-at-hiveit: Oh yeah!  Thanks, changed that

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/ActivityNames.cs
duncan-at-hiveit: Updated to leave the generic ones at the top, then initial data set ones next, and then next data set version ones last if that's OK?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/ProcessorHostBuilder.cs
```diff
@@ -83,8 +83,9 @@ public static IHostBuilder ConfigureProcessorHostBuilder(this IHostBuilder hostB
                     .AddScoped<ITimePeriodMetaRepository, TimePeriodMetaRepository>()
                     .AddScoped<IParquetService, ParquetService>()
                     .AddScoped<IPrivateBlobStorageService, PrivateBlobStorageService>()
-                    .AddScoped<IValidator<DataSetCreateRequest>,
-                        DataSetCreateRequest.Validator>()
+                    .AddScoped<IValidator<DataSetCreateRequest>, DataSetCreateRequest.Validator>()
+                    .AddScoped<IValidator<NextDataSetVersionCreateRequest>,
+                        NextDataSetVersionCreateRequest.Validator>()
```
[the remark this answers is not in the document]
jack-hive: Hmmmm... I'd expect `.AddFluentValidation` to do that for you. If you take a look under 'FluentValidationServiceCollectionExtensions' it seems like it should work with that 1 line alone :/ 

I'm not sure why it's failing

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/DbContextTransactionExtensionsTests.cs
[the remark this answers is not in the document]
jack-hive: Yeah that's cool with me! We technically test this as part of some of the endpoints anyway

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ProcessNextDataSetVersionFunctionTests.cs
jack-hive: Oh sorry

Do we do anything with it though? The test seems to pass if you remove all this

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/ProcessorHostBuilder.cs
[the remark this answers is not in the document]
duncan-at-hiveit: Not sure tbh.  We also do this in the Content API Startup.cs:

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ProcessNextDataSetVersionFunctionTests.cs
duncan-at-hiveit: Oh yeah of course, because this PR doesn't contain the "create mappings" work!  Cool, removed here, but it's reinstated in the 4945 PR!  Ta

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/Public.Data/ProcessorClientTests.cs
```diff
@@ -163,22 +163,6 @@ public async Task HttpClientBadRequest_ReturnsBadRequest()
...
-            _mockHttp.Expect(HttpMethod.Delete, $"{Uri.AbsoluteUri}/{dataSetVersionId}")
-                .Respond(HttpStatusCode.NotFound);
-
-            var response = await _processorClient.DeleteDataSetVersion(dataSetVersionId);
-
-            _mockHttp.VerifyNoOutstandingExpectation();
-
-            var left = response.AssertLeft();
-            left.AssertNotFoundResult();
-        }
```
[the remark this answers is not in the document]
duncan-at-hiveit: I've reinstated this endpoint's special-case handling of 404s by allowing each endpoint to supply an optional "additionalResponseValidator" so that each endpoint can get first stab at deciding what to do with the response.  404s are now returning NotFoundResults for this endpoint alone, although I'd personally be tempted to consider doing this with all endpoints :)

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/ProcessInitialDataSetVersionFunction.cs
benoutram: It's not a big deal but it's strange that you are not seeing this. The introduction of that line containing only whitespace is shown on Github when I look through the 'Files changed' list. It's line 70 directly above the line containing the annotation `[Function(ActivityNames.CompleteInitialDataSetVersionProcessing)]`.

If I check out this branch locally and reformat this file it shows as a change which can be seen here:

image

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/ProcessorClient.cs
```diff
@@ -77,22 +80,26 @@ internal class ProcessorClient(
...
             () => httpClient.DeleteAsync(
                 url,
                 cancellationToken: cancellationToken),
+            additionalResponseValidator,
             cancellationToken);
     }
 
     private async Task<Either<ActionResult, TResponse>> SendRequest<TResponse>(
         Func<Task<HttpResponseMessage>> requestFunction,
+        Func<HttpResponseMessage, ActionResult?>? additionalResponseValidator = null,
```
jack-hive, line 102: I'd potentially rename this to `additionalFailedResponseValidator`, since it is only used in non-success cases. But other than that, I think this looks good :)

jack-hive (approved):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/ProcessInitialDataSetVersionFunction.cs
duncan-at-hiveit: Aha ok got it ta

duncan-at-hiveit: fixed!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/ProcessorClient.cs
duncan-at-hiveit: I've renamed to `customResponseHandler` to keep it open-ended and moved it so that it can potentially handle any and all responses, failures or otherwise.  Doesn't have to be solely for validation anymore but could be used to do something custom with success cases as well.

PR #4957, opened 2024-06-13
EES-5173 Change Release Version DELETION plan to include deleting draft API data sets

jack-hive:
When a user decides to delete a specific (amendment) release version, we will first check if it is associated with any API data sets. If it is, then we will also delete these as part of the plan.

As part of this, we need to change the Release Version deletion plan to include details about the linked API data set versions that will be deleted - and whether or not each of them are in a state which can be deleted. We do this via `Valid` flags

The deletion plan for the release version now looks like:

```json
"scheduledMethodologies": [],
"apiDataSetVersions": [
      {
          "dataSetId": "afbeb08c-b6c8-4e9d-d24e-08dc1cbc809a",
          "dataSetName": "My First API Data Set",
          "dataSetVersionId": "bfbeb08c-b6c8-4e9d-d24e-08dc1cbc809b",
          "version": "v1.0",
          "status": "Draft",
          "valid": true
      },
]
"valid": true
```

We have added the `apiDataSetVersions` and `valid` fields.

As can be seen from the example JSON above, the release version is linked to 1 API data set version. The API data set versions itself is in a `valid` state to be deleted, with it being in `Draft` - Therefore, the overall `valid` flag, on the root of the JSON, is also set to `true`. Deleting the release version will subsequently delete all the linked API data set versions.


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseService.cs
```diff
@@ -650,6 +677,25 @@ public class DeleteDataFilePlan
...
+        public IReadOnlyList<DeleteApiDataSetVersionPlan> ApiDataSetVersions { get; init; } = [];
+
+        public bool Valid => ApiDataSetVersions.All(dsv => dsv.Valid);
+    }
+
+    public class DeleteApiDataSetVersionPlan
+    {
+        public Guid DataSetId { get; init; }
+
+        public string DataSetName { get; init; } = null!;
```
ntsim, line 691: `DataSetTitle` is more accurate as it's formed from the `DataSet.Title` property

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseService.cs
```diff
@@ -650,6 +677,25 @@ public class DeleteDataFilePlan
...
+
+        public Guid DataSetVersionId { get; init; }
```
ntsim, line 693: Could this be simply `Id` given that the view model is concerning a data set version?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionExtensions.cs
```diff
@@ -0,0 +1,11 @@
...
+
+public static class DataSetVersionExtensions
+{
+    public static bool CanBeDeleted(this DataSetVersion dataSetVersion) => dataSetVersion.Status.IsDeletableState();
+
+    public static bool IsDeletableState(this DataSetVersionStatus status) => status is DataSetVersionStatus.Failed
+        or DataSetVersionStatus.Mapping
+        or DataSetVersionStatus.Draft
+        or DataSetVersionStatus.Cancelled;
+}
```
ntsim, line 11: Not sure why we've gone for extensions here when we can just have these as methods on the model itself?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionExtensions.cs
```diff
@@ -0,0 +1,11 @@
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Extensions;
+
+public static class DataSetVersionExtensions
+{
+    public static bool CanBeDeleted(this DataSetVersion dataSetVersion) => dataSetVersion.Status.IsDeletableState();
+
+    public static bool IsDeletableState(this DataSetVersionStatus status) => status is DataSetVersionStatus.Failed
```
ntsim, line 7: Why two methods? Just a single one is sufficient as they do the same thing.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/ReleasesController.cs
```diff
@@ -60,11 +61,11 @@ public class ReleasesController : ControllerBase
         }
 
         [HttpDelete("release/{releaseVersionId:guid}")]
-        public async Task<ActionResult<ReleaseViewModel>> DeleteRelease(Guid releaseVersionId)
+        public async Task<ActionResult> DeleteRelease(Guid releaseVersionId)
```
duncan-at-hiveit, line 64: I would name this "DeleteReleaseVersion" given that it's not the full Release that you'll be deleting here.  I know that there are other areas where we simply say "Release", but I think we're trying to generally move over to renaming things correctly to ReleaseVersion where we can (this is historically because we didn't used to have separate Release and ReleaseVersion entities).

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Interfaces/IReleaseService.cs
```diff
@@ -15,7 +16,9 @@ public interface IReleaseService
     {
         Task<Either<ActionResult, ReleaseViewModel>> CreateRelease(ReleaseCreateRequest releaseCreate);
 
-        Task<Either<ActionResult, DeleteReleasePlan>> GetDeleteReleasePlan(Guid releaseVersionId);
+        Task<Either<ActionResult, DeleteReleasePlan>> GetDeleteReleasePlan(
```
duncan-at-hiveit, line 19: Might be a good time to rename this to GetDeleteReleaseVersionPlan

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/EitherTestExtensions.cs
```diff
@@ -79,6 +80,11 @@ public static void AssertStatusCodeResult<T>(this Either<ActionResult, T> result
             badRequest.AssertValidationProblem(expectedValidationErrors);
             return either.Left;
         }
+        public static ValidationProblemViewModel AssertBadRequestWithValidationProblem<TRight>(this Either<ActionResult, TRight> either)
```
duncan-at-hiveit, line 83: Cheeky newline above?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionExtensions.cs
jack-hive: I've moved them to the model

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionExtensions.cs
jack-hive: One is for the model, one is for the status. In the viewmodels for the deletion plan I didn't have access to the full model when calculating the `Valid` flag, because it was auto-calculated. So I needed to use the status

jack-hive: After discussing, I've turned this into just the one method and moved it to the model itself. We no longer auto-calculate the `valid` flag, so can remove the extension methods

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/ReleasesControllerTests.cs
```diff
@@ -296,24 +299,89 @@ public async Task GetDeleteDataFilePlan()
...
+
+            var result = await controller.DeleteReleaseVersion(_releaseVersionId);
+            VerifyAllMocks(releaseService);
+
+            var validationProblem = result.AssertBadRequestWithValidationProblem();
+
+            validationProblem.AssertHasError(
+                expectedPath: "error path",
+                expectedCode: "error code");
+        }
```
ntsim, line 383: These tests are the the old school non-integration controller tests which are essentially useless. Our controllers are usually super thin and the main logic resides in the service that's called. As we mock out the service, these tests basically prove nothing and have next to zero value.

We've previously agreed in dev team meetings that we wouldn't continue writing these styles of tests.

Of course, we can keep them now you've written them, but in the future would advise against spending any time working on these types of tests. Instead, either focus on the service tests (where the value actually is) or convert to integration tests. The choice should depend on what's most appropriate for that PR.

ntsim (approved):

PR #4960, opened 2024-06-14
Make it easier and smoother to run playwright tests locally

sambiramairelogic:
This is purely an optional quality of life PR, but it's been bugging me that we have to install playwright each time in order to run the tests, and that we can't run the playwright tests from the root directory with a single command. 

I've also been meaning to install @frontend to the playwright-tests project so it can more safely import `redirects.js`. 

This PR adds the playwright-tests project to the pnpm workspace so that playwright is installed during `pnpm i`, and adds a command to run tests from root.

### DIFF
--- a/README.md
+++ b/README.md
@@ -552,6 +552,8 @@ These scripts can generally be run from most `package.json` files across the pro
 
 - `pnpm format` - Format codebase using Prettier.
 
+- `pnpm test:playwright` - Run the Playwright tests (requires all services running)
+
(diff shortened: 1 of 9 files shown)


saicharan2789:
Hi @sambiramairelogic - I had spined up the local environment and ran the below commands : 1 `pnpm i ` and 2 
`pnpm test:playwright`. I am able to run the playwright tests, However all the tests are failing now 
image

PR #4966, opened 2024-06-14
Improve Methodology Sitemap Item query

sambiramairelogic:
The Dev environment is encountering Null Argument Exceptions when trying to fetch the dynamic sitemap. 

For some bizarre reason this works when running locally, but our query should be properly loading the Methodology of the MethodologyVersion through a `.ThenInclude`. 

This PR also adds cancellation tokens to all ListSitemapItems routes.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Services/Interfaces/IDataSetFileService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Services/Interfaces/IDataSetFileService.cs
@@ -29,5 +29,5 @@ public interface IDataSetFileService
     Task<Either<ActionResult, DataSetFileViewModel>> GetDataSetFile(
         Guid dataSetId);
 
-    Task<Either<ActionResult, List<DataSetSitemapItemViewModel>>> ListSitemapItems();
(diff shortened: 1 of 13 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Services/Interfaces/IDataSetFileService.cs
```diff
@@ -29,5 +29,5 @@ public interface IDataSetFileService
     Task<Either<ActionResult, DataSetFileViewModel>> GetDataSetFile(
         Guid dataSetId);
 
-    Task<Either<ActionResult, List<DataSetSitemapItemViewModel>>> ListSitemapItems();
+    Task<Either<ActionResult, List<DataSetSitemapItemViewModel>>> ListSitemapItems(CancellationToken cancellationToken);
```
ntsim, line 32: To simplify the tests, we've typically been adding `default` for the cancellation tokens i.e.

```cs
ListSitemapItems(CancellationToken canellationToken = default);
```

This way, we don't need to call the service methods with this argument as it's not relevant most of the time.

This can be done with the other interfaces and the implementations as well, and then the tests can be simplified too

ntsim (approved):

PR #4970, opened 2024-06-14
EES-4906 Filter API data set candidates that are still importing

ntsim:
This PR fixes an issue where the API data set candidates list would include files that were still importing. This has now been resolved to filter these files out.

## Other changes

- Updated generator extensions for `File`. This importantly introduces different defaults depending on what `FileType` is passed to the `DefaultFile` method. Previously, the assumption was that the default should be a data file which is irrelevant for things like ancillary files.
- Updated generator extensions for `DataImport`. This removes `Set/WithRowCounts` in favour of the typical generator extensions for different properties.
- Introduces default `DataFixture` instances for `Generators`. This fixes various issues like potential `-1` values for fixture indices, but also allows fixtures to be propagated to sub-contexts during generation (e.g. setting a child relationship).


src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Fixtures/FileGeneratorExtensions.cs
```diff
@@ -122,9 +98,53 @@ public static InstanceSetters<File> SetDefaults(this InstanceSetters<File> sette
...
 
+    public static InstanceSetters<File> SetDefaults(this InstanceSetters<File> setters, FileType? fileType)
+        => fileType switch
+        {
+            FileType.Data => setters.SetDataFileDefaults(),
+            FileType.Metadata => setters.SetMetaFileDefaults(),
+            FileType.Ancillary => setters.SetAncillaryFileDefaults(),
+            null => setters.SetDefaults(),
+            // TODO: Implement other file types
+            _ => throw new ArgumentOutOfRangeException(nameof(fileType), fileType, null)
```
benoutram, line 112: Would it be too difficult to implement the other file types as part of this change even if we don't change any tests to use it? Maybe there's some types that we aren't that bothered about generating so happy if you want to just leave this.

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Fixtures/DataImportGeneratorExtensions.cs
```diff
@@ -12,55 +12,140 @@ public static Generator<DataImport> DefaultDataImport(this DataFixture fixture)
...
     public static Generator<DataImport> WithStatus(
         this Generator<DataImport> generator,
         DataImportStatus status)
-        => generator.ForInstance(s => s.Set(d => d.Status, status));
+        => generator.ForInstance(s => s.SetStatus(status));
+
+    public static InstanceSetters<DataImport> WithStagePercentageComplete(
+        this InstanceSetters<DataImport> setters,
+        int stagePercentageComplete)
+        => setters.Set(d => d.StagePercentageComplete, stagePercentageComplete);
```
benoutram, line 48: The body of this looks like it should be named `SetStagePercentageComplete`, but then we are missing `WithStagePercentageComplete`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetCandidateService.cs
```diff
@@ -44,19 +44,27 @@ internal class DataSetCandidateService(
...
+            .ToListAsync(cancellationToken: cancellationToken))
+            .Select(fileId => releaseFilesByFileId[fileId])
             .Select(rf => new DataSetCandidateViewModel
             {
                 ReleaseFileId = rf.Id,
                 Title = rf.Name!,
             })
             .OrderBy(rf => rf.Title)
-            .ToListAsync(cancellationToken: cancellationToken);
+            .ToList();
```
benoutram, line 68: I initially found this a bit confusing to read because my first thought was 'why is the source of the candidates result originating from `DataImports`?'.

I understand it now but just wondering if you think it would be more readable if the file id's with data imports list was assigned a variable, something like this?

```csharp
var fileIdsWithCompletedImports = await contentDbContext.DataImports
            .Where(di => di.Status == DataImportStatus.COMPLETE)
            .Where(di => releaseFilesByFileId.Keys.Contains(di.FileId))
            .Select(di => di.FileId)
            .ToListAsync(cancellationToken: cancellationToken);

return fileIdsWithCompletedImports 
            .Select(fileId => releaseFilesByFileId[fileId])
            .Select(rf => new DataSetCandidateViewModel
            {
                ReleaseFileId = rf.Id,
                Title = rf.Name!,
            })
            .OrderBy(rf => rf.Title)
            .ToList();
```

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Fixtures/FileGeneratorExtensions.cs
```diff
@@ -8,35 +8,11 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Model.Tests.Fixture
...
-            {
-                GeographicLevels = [GeographicLevel.Country],
-                TimePeriodRange = new TimePeriodRangeMeta
-                {
-                    Start = new TimePeriodRangeBoundMeta { TimeIdentifier = TimeIdentifier.CalendarYear, Period = "2000", },
-                    End = new TimePeriodRangeBoundMeta { TimeIdentifier = TimeIdentifier.CalendarYear, Period = "2001", }
-                },
-                Filters = [new() { Id = Guid.NewGuid(), Label = "Filter 1", ColumnName = "filter_1", }],
-                Indicators = [new() { Id = Guid.NewGuid(), Label = "Indicator 1", ColumnName = "indicator_1", }],
-            })
```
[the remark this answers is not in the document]
ntsim: This is still included in the defaults for the `Data` file type. This is constructed via invoking `context.Fixture.DefaultDataSetFileMeta` (L130), which sets this up in its defaults.

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Fixtures/FileGeneratorExtensions.cs
ntsim: It's not too difficult, but I didn't want to expand the scope of the PR any further. Thought it'd be best to leave it to the next person that actually needs one of the missing file types to implement the defaults as they'll have better context on the requirements.

Will leave this for now if that's okay!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetCandidateService.cs
ntsim: Fair point, but also I realised that there's a more efficient way to do this using a `Join` rather than two queries. The code now looks like this:

```cs
return await contentDbContext.ReleaseFiles
    .Where(rf => rf.ReleaseVersionId == releaseVersionId)
    .Where(rf => rf.File.Type == FileType.Data)
    .Where(rf => rf.File.PublicApiDataSetId == null)
    .Where(rf => rf.File.ReplacedById == null)
    .Where(rf => rf.File.ReplacingId == null)
    .Join(
        contentDbContext.DataImports,
        rf => rf.FileId,
        di => di.FileId,
        (rf, di) => new { ReleaseFile = rf, DataImport = di }
    )
    .Where(tuple => tuple.DataImport.Status == DataImportStatus.COMPLETE)
    .Select(tuple => tuple.ReleaseFile)
    .Select(rf => new DataSetCandidateViewModel
    {
        ReleaseFileId = rf.Id,
        Title = rf.Name!,
    })
    .OrderBy(rf => rf.Title)
    .ToListAsync(cancellationToken: cancellationToken);
```

How's this look to you?

benoutram: Looks great to me, and I saw you added `AsNoTracking()` too 🙂 .

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Fixtures/FileGeneratorExtensions.cs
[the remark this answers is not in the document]
benoutram: Cool, I see that now, thanks. 🙂

benoutram (approved):

PR #4979, opened 2024-06-18
EES-4993 Move `PublicApiDataSet*` columns from `Files` to `ReleaseFiles`

ntsim:
This PR moves the `PublicApiDataSetId` and `PublicApiDataSetVersion` columns from the content DB's `Files` table to `ReleaseFiles`.

This fixes an issue where creating a draft API data set for a `ReleaseFile` in an amendment can immediately show in the Data Catalogue as an API data set. This was due to the `PublicApiDataSet*` columns previously being on the `Files` table, meaning any changes would be shared between all `ReleaseFiles` sharing the same `File`.

:warning: **Note** - To implement this, we also require an endpoint migration to unset the `PublicApiDataSet*` columns for any amendment `ReleaseFiles` that have already been created with this issue.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
@@ -409,6 +409,27 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Model.Database
                         v => v.HasValue
                             ? DateTime.SpecifyKind(v.Value, DateTimeKind.Utc)
                             : null);
+                entity.Property(f => f.PublicApiDataSetVersion)
+                    .HasMaxLength(20)
+                    .HasConversion(
+                        v => v.ToString(),
+                        v => SemVersion.Parse(v, SemVersionStyles.Strict, 20)
+                    );
+
+                entity.HasIndex(rf => new
+                {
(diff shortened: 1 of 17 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
```diff
@@ -409,6 +409,27 @@ private static void ConfigureReleaseFile(ModelBuilder modelBuilder)
...
+                        v => v.ToString(),
+                        v => SemVersion.Parse(v, SemVersionStyles.Strict, 20)
+                    );
+
+                entity.HasIndex(rf => new
+                {
+                    rf.ReleaseVersionId,
+                    rf.FileId,
+                })
+                .IsUnique();
```
jack-hive, line 424: I'm guessing this is unrelated to this PR but we just wanted to add this in because it should have always been there?

jack-hive (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
ntsim: Yep, this is just to add a unique constraint that should have been there in the first place!

PR #4985, opened 2024-06-19
EES-5243 Add CORS policy to allow access to public API from public frontend

ntsim:
This PR adds a CORS policy for the public API so that clients from the public frontend can access it.

## Other changes

- Disabled Renovate for Bicep resources as this was suggesting incorrect versions that do not have official documentation and consequently have broken intellisense. In the future, we should manually upgrade these resource as required.

### DIFF
--- a/infrastructure/templates/public-api/parameters/main-preprod.bicepparam
+++ b/infrastructure/templates/public-api/parameters/main-preprod.bicepparam
@@ -3,6 +3,11 @@ using '../main.bicep'
 // Environment Params
 param environmentName = 'Pre-Production'
 
+param publicUrls = {
+  contentApi: 'https://s101p02-as-ees-content.azurewebsites.net'
+  publicApp: 'https://pre-production.explore-education-statistics.service.gov.uk'
+}
+
 // PostgreSQL Database Params
 param postgreSqlSkuName = 'Standard_B1ms'
 param postgreSqlStorageSizeGB = 32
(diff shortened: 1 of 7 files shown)


infrastructure/templates/public-api/parameters/main-preprod.bicepparam
```diff
@@ -3,6 +3,11 @@ using '../main.bicep'
 // Environment Params
 param environmentName = 'Pre-Production'
 
+param publicUrls = {
+  contentApi: 'https://s101p02-as-ees-content.azurewebsites.net'
```
benoutram, line 7: Is it not possible to use `https://content.pre-production.explore-education-statistics.service.gov.uk` here?

benoutram (approved):

infrastructure/templates/public-api/parameters/main-preprod.bicepparam
ntsim: For some reason, this doesn't have a working SSL certificate. I'm not sure about the context surrounding this, but someone else might know.

For now, I've just used this URL as it's also the one being used in the ARM template `pre-prod.json` parameters as well.

benoutram: Cool, I wasn't aware of the certificate warning. Thanks. 🙂

PR #4989, opened 2024-06-20
EES-4781 remove data catalogue feature flag

amyb-hiveit:
Removes the feature flag on the new Data Catalogue page, plus:
- updates unit and UI tests
- improves the check for api data sets in DataCataloguePage
- moves `publicationSortOptions`, `publicationFilters`, `DataSetFileSortOption` and `DataSetFileFilters` to their own files to prevent mysterious errors in tests
- removes old data catalogue files


src/explore-education-statistics-frontend/src/modules/data-catalogue/DataCataloguePage.tsx
```diff
@@ -66,64 +639,51 @@ export const getServerSideProps: GetServerSideProps<Props> = async context => {
...
+    const selectedPublication = themes
+      .flatMap(option => option.topics)
+      .flatMap(option => option.publications)
+      .find(option => option.slug === publicationSlug);
 
-    // Redirect old slug based links to new format.
-    if (newDesign) {
-      const redirectUrlQuery = selectedRelease
-        ? `publicationId=${selectedPublication.id}&releaseId=${selectedRelease.id}`
+    if (selectedPublication) {
```
sambiramairelogic, line 649: I wonder if it's worth being a bit stricter about this - If you put in any old garbage in place of the slugs this will evaluate `selectedPublication` as undefined, fall through and display the page as if you'd just visited `./data-catalogue` without the slugs. This will be a pain point for SEO at the very least because we're serving duplicated content for two URLs, but also Google could end up indexing it too leading to more redirects. I'd recommend we serve a 404 in this case (or redirect to `./data-catalogue`, but then that raises the question of why we don't do that elsewhere for other routes). 

Finally, I tried this out on the live site and realised we have the same problem, so apologies for flagging something you haven't introduced but it would be nice to sort it :)

sambiramairelogic:
Changing the data catalogue path to use search parameters rather than slugs will have implications for the sitemap, which is still being populated with the old routes, e.g. 
`          
{
    loc: `${process.env.PUBLIC_URL}data-catalogue/${publication.slug}/${release.slug}`,
    lastmod: release.lastModified,
},
`
So far I haven't included any search parameters in the sitemap paths, but in this case it's probably fair to argue that "they alter the content of the page significantly" and therefore should be included. Naturally this then raises another problem which is the ordering of those parameters because each unique combination will be a valid url query but counted as a different route by Google! I _think_ what we do in this case is provide a `canonical` header to tell it which order is the definitively correct one, but I'll do a bit of reading / run this by Michael.

src/explore-education-statistics-frontend/src/modules/data-catalogue/DataCataloguePage.tsx
amyb-hiveit: Good point! I've set it to redirect to `/data-catalogue` instead of 404ing as it feels more user friendly, and hopefully prevent us having to set up specific redirects for urls that 404 later on.

amyb-hiveit:
> Changing the data catalogue path to use search parameters rather than slugs will have implications for the sitemap, which is still being populated with the old routes, e.g.
> 
> { loc: `${process.env.PUBLIC_URL}data-catalogue/${publication.slug}/${release.slug}`, lastmod: release.lastModified, }
> 
> So far I haven't included any search parameters in the sitemap paths, but in this case it's probably fair to argue that "they alter the content of the page significantly" and therefore should be included. Naturally this then raises another problem which is the ordering of those parameters because each unique combination will be a valid url query but counted as a different route by Google! I _think_ what we do in this case is provide a `canonical` header to tell it which order is the definitively correct one, but I'll do a bit of reading / run this by Michael.

As discussed in Slack, I've:
- removed the sitemap entries for the existing data catalogue with publication slugs etc
- uncommented the sitemap entries for the new data set files entries ([EES-5202](https://dfedigital.atlassian.net/browse/EES-5202))
- added a canonical link to `/data-catalogue` as Google seems to suggest this is what you should do for pages where the content can be filtered and sorted, see https://support.google.com/webmasters/answer/10347851

sambiramairelogic (approved):

PR #4992, opened 2024-06-21
EES-5235 Prevent collisions of Public API PublicId's in filter and location options

benoutram:
This PR prevents the collision of Public API Public Id's in `FilterOptionMetaLink` and `LocationOptionMeta` by generating the `PublicId` [Sqids](https://sqids.org/dotnet) based on Id sequences.

Each `PublicId` is  a short unique string known as a Sqid generated from a number. The generation is deterministic, i.e. when repeated with the same number, the same `PublicId` string will be generated.

The possibility for collisions comes from where we generate `PublicId` from a number that's based on the `FilterOptionMetaLink` or `LocationOptionMeta` table row count.

If any rows are deleted (e.g. when a  data set version gets deleted) the row count reduces. When a new data set version is created and new options are inserted, the `PublicId` generation which is based on the count which is now overlapping again leads to a `PublicId` string being generated that already exists.

Here's an example where can see collisions that have occurred in `PublicId`'s of filter options:

Collisions-4

### Use database sequences as basis for `PublicId` generation

This PR switches out the use of table row counts to use database sequences.

This relies on the use of native PostreSQL functions `nextval` and `setval` to manipulate the sequences.

First `nextval` is used to advance the the sequence and return the value which is then used as the basic for our `PublicId` generation in new rows being inserted as part of a batch.

At the end of the batch `setval` is used to update the current value of the sequence. The new value depends on the number of rows that were inserted.

For `LocationOptionMeta` we can use the existing sequence `LocationOptionMetas_Id_seq` which is an implicit sequence  belonging to the `LocationOptionMeta.Id` identity column. For `FilterOptionMetaLink` we need to configure a new sequence `FilterOptionMetaLink_seq`.

During integration testing, truncating all tables to clear all data will automatically reset the value of `LocationOptionMetas_Id_seq`. However we also need to ensure `FilterOptionMetaLink_seq` is reset.

### Set Id values for `LocationOptionMeta` 

This PR also fixes a bug with our usage of LinqToDB's `BulkCopyAsync` to insert batches of `LocationOptionMeta`'s.

The `Id` key column is configured with the default `NpgsqlPropertyBuilderExtensions.UseIdentityByDefaultColumn` which is the equivalent of  `GENERATED BY DEFAULT AS IDENTITY`. This should allow our application to override the value of Id when inserting new rows which we need to do for `LocationOptionMeta` rows where we expect `Id` and `PublicId` to be in sync.

When using `BulkCopyAsync` I found the database was always generating new key values from its implicit sequence and ignoring the `Id` value provided. This is fixed by supplying configuration to `BulkCopyAsync`:

`new BulkCopyOptions { KeepIdentity = true }`

### Add new `MetaInsertBatchSize` app setting to Public Data Processor

This PR adds a new `MetaInsertBatchSize` configuration option with a default setting provided in `appsettings.json` which is 1000 rows.

This is made a configurable option so that it can be varied during integration testing to ensure there are no issues with the partitioning of collections that occurs during the batching and the  setting and getting of sequence values during those batch inserts. Any issues may not have been seen with a high batch size which results in a single batch.

### New Public API seed data

There's a new version `public-api-15.zip` of the Public API seed data to reflect changes to the data made by this PR.


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Options/ProcessorOptions.cs
```diff
@@ -0,0 +1,12 @@
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Options;
+
+public class ProcessorOptions
+{
+    public static readonly string Section = "Processor";
```
ntsim, line 5: Not too sure about this name. We're in the `Public.Data.Processor` project, so everything is implicitly a processor option.

For the `Data.Processor`, we've called a similar thing `AppSettings`. Maybe it'd be better to be consistent with this? Or is `AppSettings` a bit weird and we should call it something else?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Options/ProcessorOptions.cs
```diff
@@ -0,0 +1,12 @@
...
+
+    /// <summary>
+    /// Batch size to use when inserting location option meta rows, location option meta link rows,
+    /// and filter option meta link rows into the public data db.
+    /// </summary>
+    public required int InsertBatchSize { get; set; }
```
ntsim, line 11: Might want to name this more explicitly to mention meta i.e. `MetaInsertBatchSize`, otherwise, it sounds a bit too generic and could apply in many other contexts.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/FilterMetaRepository.cs
```diff
@@ -79,17 +82,19 @@ await optionTable
                 .InsertWhenNotMatched()
                 .MergeAsync(cancellationToken);
 
-            var startIndex = await publicDataDbContext.FilterOptionMetaLinks.CountAsync(token: cancellationToken) + 1;
+            var startIndex = (await publicDataDbContext.Database
+                    .SqlQuery<int>($"SELECT nextval('public.\"FilterOptionMetaLink_Id_seq\"')")
+                    .ToListAsync(cancellationToken: cancellationToken))
+                .First();
```
ntsim, line 88: Can this be written as the following:

```cs
var startIndex = await publicDataDbContext.Database
    .SqlQuery<int>($"SELECT nextval('public.\"FilterOptionMetaLink_Id_seq\"')")
    .FirstAsync(cancellationToken: cancellationToken);
```

Would also be tempted to abstract this into a method on `PublicDataDbContext`:

```cs
public Task<int> NextSequenceValue(string sequenceName, CancellationToken cancellationToken = default)
    => Database
        .SqlQuery<int>($"SELECT nextval('public.\"{sequenceName}\"')")
        .FirstAsync(cancellationToken);
```

This can then be re-used for importing location options and the seeder

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/FilterMetaRepository.cs
```diff
@@ -126,6 +132,10 @@ await optionTable
                     $"Inserted incorrect number of filter option meta links for {meta.PublicId}. " +
                     $"Inserted: {insertedLinks}, expected: {options.Count}");
             }
+
+            await publicDataDbContext.Database.ExecuteSqlAsync(
+                $"SELECT setval('public.\"FilterOptionMetaLink_Id_seq\"', {startIndex + insertedLinks - 1})",
+                cancellationToken);
```
ntsim, line 138: Similar to earlier comment, might be good to abstract this into a method on `PublicDataDbContext`:

```cs
public Task SetSequenceValue(string sequenceName, long value, CancellationToken cancellationToken = default)
    => Database.ExecuteSqlAsync(
        $"SELECT setval('public.\"{sequenceName}\"', {value})",
        cancellationToken);
```

This can be re-used for importing location options and the seeder

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/LocationMetaRepository.cs
```diff
@@ -113,16 +114,24 @@ FROM read_csv('{dataSetVersionPathResolver.CsvDataPath(dataSetVersion):raw}', AL
                         .Where(o => !existingRowKeys.Contains(o.GetRowKey()))
                         .ToList();
 
-                    var startIndex =
-                        await publicDataDbContext.LocationOptionMetas.CountAsync(token: cancellationToken) + 1;
+                    var startIndex = (await publicDataDbContext.Database
+                            .SqlQuery<int>($"SELECT nextval('public.\"LocationOptionMetas_Id_seq\"')")
```
ntsim, line 118: Could abstract a constant on `PublicDataDbContext` for `LocationOptionMetas_Id_seq` to re-use when interacting with the sequence in this method?

Similar thing can be done in the seeder

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Fixture/TestApplicationFactory.cs
```diff
@@ -26,18 +27,20 @@ public override async ValueTask DisposeAsync()
...
 
+    [SuppressMessage("Security", "EF1002:Risk of vulnerability to SQL injection.")]
     public async Task ClearTestData<TDbContext>() where TDbContext : DbContext
     {
         await using var context = GetDbContext<TDbContext>();
 
         var tables = context.Model.GetEntityTypes()
             .Select(type => type.GetTableName())
             .Distinct()
+            .Cast<string>()
```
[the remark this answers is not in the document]
benoutram: Cool, made this change, thanks.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Database/PublicDataDbContext.cs
```diff
@@ -22,6 +22,8 @@ private void Configure(bool updateTimestamps = true)
     protected override void OnModelCreating(ModelBuilder modelBuilder)
     {
         modelBuilder.ApplyConfigurationsFromAssembly(typeof(PublicDataDbContext).Assembly);
+
+        modelBuilder.HasSequence<int>("FilterOptionMetaLink_Id_seq");
```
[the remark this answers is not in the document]
benoutram: Good point, made this change, thanks.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/FilterMetaRepository.cs
benoutram: The reason for using `First` after `ToListAsync` and not `FirstAsync` is because that was giving me the error 

```column t.Value does not exist```

I've just realised it's documented that with `SqlQuery`/`SqlQueryRaw` the output column in the `SELECT` must be named `Value` and that it's case sensitive. In Postgres this means we need to wrap `Value` in quotes.

I've abstracted out `NextSequenceValue` and `SetSequenceValue` as suggested which makes things quite a bit neater, thanks. Unfortunately that string interpolation doesn't work well with the function calls, just like we've seen elsewhere executing DDL commands.

This is the neatest syntax that I've managed to come up with to get this working so far:

```csharp
    [SuppressMessage("Security", "EF1002:Risk of vulnerability to SQL injection.")]
    public Task<int> NextSequenceValue(string sequenceName, CancellationToken cancellationToken = default)
    {
        return Database.SqlQueryRaw<int>($"""
                                          SELECT nextval('public."{sequenceName}"') AS "Value"
                                          """)
            .FirstAsync(cancellationToken);
    }

    [SuppressMessage("Security", "EF1002:Risk of vulnerability to SQL injection.")]
    public Task<int> SetSequenceValue(string sequenceName, long value, CancellationToken cancellationToken = default)
    {
        return Database.SqlQueryRaw<int>($"""
                                          SELECT setval('public."{sequenceName}"', {value}) AS "Value"
                                          """)
            .FirstAsync(cancellationToken);
    }
```

The method `SetSequenceValue` doesn't really need to return a value but I've put it in anyway because it does return a value from the database and both methods then look more consistent.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Options/AppSettingOptions.cs
```diff
@@ -0,0 +1,12 @@
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Options;
+
+public class AppSettingOptions
```
ntsim, line 3: Minor, but shouldn't `AppSettings` be plural?

ntsim (approved):

PR #4993, opened 2024-06-21
EES-5003 - Minor fixes

sambiramairelogic:
Quick intermediary PR which fixes a few niggles ahead of the next deploy to master -

- The `PUBLIC_URL_WITH_WWW` environment variable was incorrectly named, and has been corrected (this will require renaming the similarly incorrectly named pipeline variable).
- To combat the issue of the robots/sitemap files being constructed at build time, which only occurs once and does not have access to environment variables configured for each environment, we introduced `.env.production` so that the build would have access to the correct `PUBLIC_URL` both when running locally and on the pipeline. This however has consequences for the redirect tests, because the redirects are now going to the production url when testing locally. There's a few ways we could resolve this and keep `.env.production`, but this is all just getting too messy and hacky at this stage. I think the simplest thing is to revert `.env.production` and introduce a `PROD_PUBLIC_URL` value to the normal `.env` which only the sitemap config file uses. 
- The SEO metadata playwright tests have been flaky when running locally because the sitemap file produces xml which is confusing playwright's html locators. It finds an xml `<loc>` within an html `span`, so `getByText('foo')` finds two elements, as if both the loc and the span have the same innertext. I spent 15 minutes trying to resolve this but stopped after that since we've decided not to support Playwright tests. For now I've skipped the suite.

### DIFF
--- a/src/explore-education-statistics-frontend/.env
+++ b/src/explore-education-statistics-frontend/.env
@@ -5,5 +5,5 @@ PUBLIC_API_DOCS_URL=TODO-GUIDANCE-URL
 NOTIFICATION_API_BASE_URL=http://localhost:7073/api
 GA_TRACKING_ID=
 PUBLIC_URL=http://localhost:3000/
-PROD_PUBLIC_URL=https://explore-education-statistics.service.gov.uk/
+PROD_PUBLIC_URL=https://explore-education-statistics.service.gov.uk
 APP_ENV=Local
--- a/src/explore-education-statistics-frontend/.env
+++ b/src/explore-education-statistics-frontend/.env
@@ -5,4 +5,5 @@ PUBLIC_API_DOCS_URL=TODO-GUIDANCE-URL
 NOTIFICATION_API_BASE_URL=http://localhost:7073/api
 GA_TRACKING_ID=
 PUBLIC_URL=http://localhost:3000/
+PROD_PUBLIC_URL=https://explore-education-statistics.service.gov.uk/
 APP_ENV=Local
--- a/src/explore-education-statistics-frontend/next-sitemap.config.js
+++ b/src/explore-education-statistics-frontend/next-sitemap.config.js
@@ -1,6 +1,6 @@
 /** @type {import('next-sitemap').IConfig} */
 module.exports = {
-  siteUrl: process.env.PUBLIC_URL,
+  siteUrl: process.env.PROD_PUBLIC_URL,
   sitemapSize: 5000,
   exclude: ['/server-sitemap.xml'],
   generateRobotsTxt: true,
@@ -16,7 +16,7 @@ module.exports = {
         ],
       },
     ],
-    additionalSitemaps: [`${process.env.PUBLIC_URL}server-sitemap.xml`],
+    additionalSitemaps: [`${process.env.PROD_PUBLIC_URL}server-sitemap.xml`],
   },
   transform: async (config, path) => {
     if (path === '/') {
--- a/src/explore-education-statistics-frontend/next-sitemap.config.js
+++ b/src/explore-education-statistics-frontend/next-sitemap.config.js
@@ -16,7 +16,7 @@ module.exports = {
         ],
       },
     ],
-    additionalSitemaps: [`${process.env.PROD_PUBLIC_URL}server-sitemap.xml`],
+    additionalSitemaps: [`${process.env.PROD_PUBLIC_URL}/server-sitemap.xml`],
   },
   transform: async (config, path) => {
     if (path === '/') {
--- a/tests/playwright-tests/tests/public/seoMetadataFiles.spec.ts
+++ b/tests/playwright-tests/tests/public/seoMetadataFiles.spec.ts
@@ -3,7 +3,10 @@ import environment from '@util/env';
 
 const { PUBLIC_URL, PROD_PUBLIC_URL } = environment;
 
-test.describe('SEO Metadata Files', () => {
+// Playwright finds two elements with getByText('...sitemal.xml') because the html span they're displayed
+// in contains the xml <loc> tag, which playwright then treats as another html element.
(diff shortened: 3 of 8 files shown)


sambiramairelogic:
FYI this is the playwright issue, seems to be using `#webkit-xml-viewer-source-xml` to parse the xml 
image

tests/robot-tests/.env.example
```diff
@@ -1,4 +1,5 @@
 PUBLIC_URL=http://localhost:3000  # URL of the public frontend
+PUBLIC_URL_WITH_WWW=http://www.localhost:3000  # URL of the public frontend with www
```
ntsim, line 2: We added a `user navigates to public frontend with www` keyword which means we don't need to add a new environment variable.

Can use this keyword for the UI test changes in this PR.

src/explore-education-statistics-frontend/.env
```diff
@@ -5,4 +5,5 @@ PUBLIC_API_DOCS_URL=TODO-GUIDANCE-URL
 NOTIFICATION_API_BASE_URL=http://localhost:7073/api
 GA_TRACKING_ID=
 PUBLIC_URL=http://localhost:3000/
+PROD_PUBLIC_URL=https://explore-education-statistics.service.gov.uk/
```
ntsim, line 8: Could get rid of the trailing slash given? We don't need to change this in all the environments or infrastructure so this should be pretty easy to do.

Will need adjustments in the `next-sitemap.config.js`

src/explore-education-statistics-frontend/next-sitemap.config.js
```diff
@@ -16,7 +16,7 @@ module.exports = {
         ],
       },
     ],
-    additionalSitemaps: [`${process.env.PUBLIC_URL}server-sitemap.xml`],
+    additionalSitemaps: [`${process.env.PROD_PUBLIC_URL}server-sitemap.xml`],
```
ntsim, line 19: Don't we need to make additional changes in the `server-sitemap.xml` as well? All the locations are using `PUBLIC_URL` instead of `PROD_PUBLIC_URL`

ntsim (changes requested):

tests/playwright-tests/tests/public/seoMetadataFiles.spec.ts
```diff
@@ -3,7 +3,10 @@ import environment from '@util/env';
 
 const { PUBLIC_URL, PROD_PUBLIC_URL } = environment;
 
-test.describe('SEO Metadata Files', () => {
+// Playwright finds two elements with getByText('...sitemal.xml') because the html span they're displayed
+// in contains the xml <loc> tag, which playwright then treats as another html element.
+// TODO: Either resolve this, or re-create these tests in robot if the team prefers
+test.describe.skip('SEO Metadata Files', () => {
```
ntsim, line 9: Wouldn't it be better to skip just the offending test cases?

tests/playwright-tests/tests/public/seoMetadataFiles.spec.ts
```diff
@@ -3,7 +3,10 @@ import environment from '@util/env';
 
 const { PUBLIC_URL, PROD_PUBLIC_URL } = environment;
 
-test.describe('SEO Metadata Files', () => {
+// Playwright finds two elements with getByText('...sitemal.xml') because the html span they're displayed
+// in contains the xml <loc> tag, which playwright then treats as another html element.
```
ntsim, line 7: Could use an XPath selector if there's an issue finding the correct element?

Also, it's weird that it's picking up a `span` because the sitemap doesn't actually contain this element!

tests/robot-tests/.env.example
sambiramairelogic: D'oh, yes of course we did. Will do

src/explore-education-statistics-frontend/next-sitemap.config.js
sambiramairelogic: Hah, yeah happily. I went with the slash to be consistent with `PUBLIC_URL` but yes would much prefer without

tests/playwright-tests/tests/public/seoMetadataFiles.spec.ts
sambiramairelogic: Well, maybe. Ideally! 

Only one of them fails for me, but looking at the assertions in other tests I can't actually explain why the others don't fail too. They use `getByText...toHaveCount(1)` too. 
I'm just a bit concerned that this work has been causing an awful lot of pipeline failures, and these tests aren't really adding a supreme amount of value right now. Also, because the purpose of this PR is to iron out silly issues before a merge to master I'm trying to be as safe as possible. 

If you disagree then I'll happily just skip the one failing test - but wanted to share my reasoning

tests/playwright-tests/tests/public/seoMetadataFiles.spec.ts
sambiramairelogic: Yep was considering that, alternatively could actually make an API GET request to the `/sitemap.xml` route with content type set to xml, then parse the response and assert on it all properly that way. Just cautious not to spend too long on Playwright without approval to do so 

And, yeah, really no idea why it's getting the span!

src/explore-education-statistics-frontend/.env
sambiramairelogic: Hah, yeah happily. I went with the slash to be consistent with PUBLIC_URL but yes would much prefer without

tests/playwright-tests/tests/public/seoMetadataFiles.spec.ts
ntsim: Okay fair enough, happy to leave it then!

ntsim (approved):

PR #4995, opened 2024-06-21
EES-5170: Update geospatial boundary data import process to handle GeoJSON files

tomjonesdev:
The current process for implementing new geospatial boundary data requires a developer to complete several manual steps, using various tools, to convert shapefile data into the GeoJSON format used by EES each time new data is published.

This work adds a section to the administrative portal which allows BAU users to manage the data themselves, removing the dependency on a developer and streamlining the overall process. This section allows users to view the current levels, amend their labels, and upload new data files as they become available.


src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/BoundaryLevelRepository.cs
```diff
@@ -18,9 +19,32 @@ public BoundaryLevelRepository(StatisticsDbContext context)
...
 
+        public async Task<IEnumerable<BoundaryLevel>> Get()
+        {
+            return await _context.BoundaryLevel.ToListAsync();
+        }
+
         public Task<BoundaryLevel?> Get(long id)
         {
-            return _context.BoundaryLevel.SingleOrDefaultAsync(level => level.Id == id);
+            return id == 0
```
mmyoungman, line 29: Why is it important that the id not be 0?

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/BoundaryDataUtils.cs
```diff
@@ -0,0 +1,40 @@
...
+            .ToString()
+            .Equals("OBJECTID", StringComparison.InvariantCultureIgnoreCase);
+
+        var nameIndex = propertiesContainsObjectId ? 2 : 1;
+        var codeIndex = propertiesContainsObjectId ? 1 : 0;
+
+        var nameKey = properties.ElementAt(nameIndex).Key;
+        var codeKey = properties.ElementAt(codeIndex).Key;
+
+        if (!nameKey.EndsWith("NM", StringComparison.OrdinalIgnoreCase))
```
mmyoungman, line 23: I don't know if it's worth adding a comment for someone to understand why this is necessary. Same for the `CD` check.

src/explore-education-statistics-admin/src/pages/bau/BoundaryDataPage.tsx
```diff
@@ -0,0 +1,198 @@
...
+          </dl>
+
+          <h2 className="govuk-heading-lg">Level codes</h2>
+          <dl
+            className={classNames(
+              styles.glossarySection,
+              styles.boundaryLevels,
+            )}
+          >
+            <dt>EDA</dt>
```
mmyoungman, line 67: Can we use `locationsLevelsMap` here rather than hardcoding these? Could add the code (`EDA` or whatever) to `locationsLevelsMap` and then just loop over them all. This way if someone adds another location type in the future, they're less likely to miss adding it here.

src/explore-education-statistics-admin/src/pages/bau/BoundaryDataPage.tsx
```diff
@@ -0,0 +1,198 @@
...
+      <LoadingSpinner
+        loading={isLoadingBoundaryLevels}
+        text="Loading boundary levels"
+      >
+        <Details summary="Glossary" id="glossary">
+          <h2 className="govuk-heading-lg">Boundary types</h2>
+          <dl
+            className={classNames(styles.glossarySection, styles.boundaryTypes)}
+          >
+            <dt>BFC</dt>
```
mmyoungman, line 35: Is there any chance the frontend is going to want access to these boundary type descriptions in other places? If so I'd consider creating something like `locationsLevelsMap` for boundary types, and then using that to display them here.

src/explore-education-statistics-admin/src/pages/bau/BoundaryDataUploadPage.tsx
```diff
@@ -0,0 +1,157 @@
...
+              >
+                <FormGroup>
+                  <FormFieldSelect<FormValues>
+                    id="boundaryLevel"
+                    name="boundaryLevel"
+                    label="Select boundary level"
+                    className="govuk-!-width-one-third"
+                    options={[
+                      { value: '', label: '' },
+                      {
```
mmyoungman, line 72: I'd also be inclined to use `locationLevelsMap` here.

mmyoungman (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Statistics/BoundaryDataController.cs
```diff
@@ -0,0 +1,95 @@
...
+    [HttpGet("boundary-data/levels")]
+    public async Task<ActionResult<List<BoundaryLevelViewModel>>> GetBoundaryLevels()
+    {
+        var levels = await _boundaryLevelService.Get();
+
+        return Ok(levels);
+    }
+
+    [HttpGet("boundary-data/levels/{id}")]
+    public async Task<ActionResult<List<BoundaryLevelViewModel>>> GetBoundaryLevel(long id)
```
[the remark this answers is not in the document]
tomjonesdev: The boundary level ID on the database is a `long` for some reason, this avoided the need to convert an `int` to a `string` then parse it to a `long`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Statistics/BoundaryDataController.cs
```diff
@@ -0,0 +1,95 @@
...
+
+    [HttpGet("boundary-data/levels")]
+    public async Task<ActionResult<List<BoundaryLevelViewModel>>> GetBoundaryLevels()
+    {
+        var levels = await _boundaryLevelService.Get();
+
+        return Ok(levels);
+    }
+
+    [HttpGet("boundary-data/levels/{id}")]
```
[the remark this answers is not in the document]
tomjonesdev: Will that constraint not cause issues with consumers trying to hit the endpoint? It's not common for IDs to be `long`

src/GovUk.Education.ExploreEducationStatistics.Data.Model.Tests/Repository/BoundaryDataRepositoryTests.cs
```diff
@@ -0,0 +1,58 @@
...
+public sealed class BoundaryDataRepositoryTests : IDisposable
+{
+    private readonly DbContextOptions<StatisticsDbContext> _contextOptions;
+    private readonly SqliteConnection _connection;
+    private readonly StatisticsDbContext _context;
+    private readonly BoundaryDataRepository _sut;
+
+    public BoundaryDataRepositoryTests()
+    {
+        _connection = new SqliteConnection("Filename=:memory:");
```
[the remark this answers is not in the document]
tomjonesdev: Yeah I was just trying it out, it's easy enough to switch over. I didn't get round to finishing tests or adding integration tests before going off, so this aspect was very much a WIP.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/BoundaryLevelRepository.cs
tomjonesdev: `0` implies a value wasn't provided, the default for a numeric value type.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Statistics/BoundaryDataController.cs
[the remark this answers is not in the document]
mmyoungman: I'm not 100%, but I assumed that was a hint to the model binder to turn the id included in the route into a long - although it's not really necessary as the model binder can infer the type from the controller method's arguments, but we generally add the type to the route string as well, for whatever reason.

src/explore-education-statistics-admin/src/pages/bau/BoundaryDataPage.tsx
tomjonesdev: Possibly, wasn't aware of that so I'll have a look how it's used elsewhere to make sure adding another property doesn't break stuff.

YAGNI principle put me off adding reusability complexity to some of this work, that's why most of the view stuff is in one file.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/BoundaryLevelRepository.cs
```diff
@@ -29,5 +53,23 @@ public IEnumerable<BoundaryLevel> FindByGeographicLevels(IEnumerable<GeographicL
...
+        public async Task<BoundaryLevel> Create(
+            GeographicLevel level,
+            string label,
+            DateTime published)
+        {
+            var newBoundaryLevel = await _context.BoundaryLevel.AddAsync(new()
+            {
+                Level = level,
+                Label = label,
+                Created = DateTime.UtcNow,
```
benoutram, line 66: We haven't been setting any created values explicitly recently because we have interfaces in `Timestamps.cs` which can be used which when present mean the entities get `Created` or `Created` and `Updated` values set automatically.

It should be just a case of changing `BoundaryLevel` to `public class BoundaryLevel : ICreatedTimestamp<DateTime>`.

Since they can now be updated too when labels change you could consider using `ICreatedUpdatedTimestamp<DateTime,DateTime?>` . This will force you to add a `DateTime? Updated` property and then you will need to recreate your database migration to add the new column.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/BoundaryLevelRepository.cs
```diff
@@ -18,9 +19,32 @@ public BoundaryLevelRepository(StatisticsDbContext context)
...
+            {
+                throw new ArgumentNullException(nameof(id));
+            }
+
+            var level = await _context.BoundaryLevel.FindAsync(id)
+                ?? throw new KeyNotFoundException();
+
+            level.Label = label;
+            await _context.SaveChangesAsync();
         }
```
benoutram, line 48: See comment left on `Create` about adding an `Updated` datetime to audit when these get updated.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/BoundaryLevelRepository.cs
benoutram: I agree with @mmyoungman, this seems a bit strange to me. I'd expect trying to get an id that's not found in the database to ultimately return a 404 rather than an exception.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Statistics/BoundaryDataController.cs
```diff
@@ -0,0 +1,95 @@
...
+    private readonly IBoundaryDataService _boundaryDataService;
+    private readonly IBoundaryLevelService _boundaryLevelService;
+
+    public BoundaryDataController(
+        IBoundaryDataService boundaryDataService,
+        IBoundaryLevelService boundaryLevelService)
+    {
+        _boundaryDataService = boundaryDataService;
+        _boundaryLevelService = boundaryLevelService;
+    }
```
benoutram, line 31: You could neaten this up with a primary constructor here so this all becomes

```csharp
public class BoundaryDataController(
    IBoundaryDataService boundaryDataService,
    IBoundaryLevelService boundaryLevelService)
    : ControllerBase
```

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Statistics/BoundaryDataController.cs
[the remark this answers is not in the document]
benoutram: It will mean that if the value is not a valid _whatever-the-type-is_, then the route will not be matched and the caller will get a 404, which I think is a good thing.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/BoundaryLevelService.cs
```diff
@@ -0,0 +1,66 @@
...
+
+public class BoundaryLevelService : IBoundaryLevelService
+{
+    private readonly IBoundaryLevelRepository _boundaryLevelRepository;
+
+    public BoundaryLevelService(
+        IBoundaryLevelRepository boundaryLevelRepository)
+    {
+        _boundaryLevelRepository = boundaryLevelRepository;
+    }
```
benoutram, line 21: You could neaten this up by using a primary constructor so this is replaced by

```csharp
public class BoundaryLevelService(IBoundaryLevelRepository boundaryLevelRepository) : IBoundaryLevelService
```

I would also be tempted to not use a repository and depend on `StatisticsDbContext` directly for these non-complex CRUD operations.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Statistics/BoundaryLevelUpdateRequest.cs
```diff
@@ -0,0 +1,8 @@
+namespace GovUk.Education.ExploreEducationStatistics.Admin.ViewModels.Statistics;
+
+public class BoundaryLevelUpdateRequest
+{
+    public int Id { get; set; }
```
benoutram, line 5: The type here should match the type of `BoundaryLevel.Id` so `long` rather than `int`.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Statistics/BoundaryLevelUpdateRequest.cs
```diff
@@ -0,0 +1,8 @@
...
+
+    public string Label { get; set; }
```
benoutram, line 7: You could consider adding some Fluent validation to this request type.
I've not seen yet if you enforce a minimum label in the FE but if you do it can be enforced here too like so...

(not tested! 🙂 )

```csharp
    public class Validator : AbstractValidator<BoundaryLevelUpdateRequest>
    {
        public Validator()
        {
            RuleFor(request => request.Id)
                .GreaterThan(0);

            RuleFor(request => request.Label)
                .NotEmpty()
                .MaximumLength(5);
        }
    }
```

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/BoundaryLevelService.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        if (id == 0)
+        {
+            throw new ArgumentNullException(nameof(id));
+        }
+
+        if (string.IsNullOrWhiteSpace(label))
+        {
+            throw new ArgumentNullException(nameof(label));
+        }
```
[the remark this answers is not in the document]
benoutram: Also see the comment I've left on `BoundaryLevelUpdateRequest.cs` with an example of doing the validation there using FluentValidation which would replace all this.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Statistics/BoundaryLevelViewModel.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.ViewModels.Statistics;
+
+public class BoundaryLevelViewModel
+{
+    public long Id { get; set; }
+
+    public string Level { get; set; }
```
benoutram, line 9: Rather than using `string` here I would change the type to be `GeographicLevel`. That way it's more obvious what the values of `Level` are going to be and it constrains it to those values rather than an arbitrary string.

I think you will need to annotate it like so

```csharp
[JsonConverter(typeof(StringEnumConverter), typeof(CamelCaseNamingStrategy))]
public required GeographicLevel Level { get; init; }
```

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
```diff
@@ -80,6 +83,7 @@
 using DataGuidanceService = GovUk.Education.ExploreEducationStatistics.Admin.Services.DataGuidanceService;
 using DataSetService = GovUk.Education.ExploreEducationStatistics.Admin.Services.Public.Data.DataSetService;
 using GlossaryService = GovUk.Education.ExploreEducationStatistics.Admin.Services.GlossaryService;
+using HeaderNames = Microsoft.Net.Http.Headers.HeaderNames;
```
benoutram, line 86: Not sure what this is, could it simply be `using Microsoft.Net.Http.Headers;`?

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
```diff
@@ -1,8 +1,5 @@
 #nullable enable
-using System;
-using System.Collections.Generic;
-using System.Threading;
-using System.Threading.Tasks;
+using Azure.Core;
```
benoutram, line 2: I don't think this is being used.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Statistics/BoundaryLevelViewModel.cs
```diff
@@ -0,0 +1,14 @@
...
+
+    public string Label { get; set; }
+
+    public DateTime Published { get; set; }
```
benoutram, line 13: Similar comments to those I left on `BoundaryLevelUpdateRequest`, i.e. you could use `record` instead of `class` here, you could use the `required` keyword and all the `set` properties could be replaced with `init`.

It would look a bit like this

```csharp
public record BoundaryLevelViewModel
{
    public required long Id { get; init; }

    public required GeographicLevel Level { get; init; }

    public required string Label { get; init; }

    public required DateTime Published { get; init; }
}
```

src/explore-education-statistics-admin/src/pages/bau/BoundaryDataPage.tsx
mmyoungman: I think I'm keen on it because we have had to add new location types multiple times and I think every time we've missed updating one place or another, so I think anything to reduce that burden is worthwhile.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/GovUk.Education.ExploreEducationStatistics.Data.Model.csproj
```diff
@@ -11,6 +11,7 @@
   </ItemGroup>
 
   <ItemGroup>
+	<PackageReference Include="GeoJSON.Net" Version="1.2.19" />
```
benoutram, line 14: Looks like an alignment issue here or in this file as it's not formatted in-line with the other elements, but also see the comment I left in Admin.csproj about using `GeoJSON.Text` instead.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/GeoJsonRepository.cs
```diff
@@ -34,31 +34,31 @@ public class GeoJsonRepository : IGeoJsonRepository
...
         }
 
-        private IEnumerable<GeoJson> FindAndCache(long boundaryLevelId, IEnumerable<string> codes)
+        private IEnumerable<BoundaryData> FindAndCache(long boundaryLevelId, IEnumerable<string> codes)
         {
-            var geoJsonList = _context.GeoJson.Where(geoJson =>
-                geoJson.BoundaryLevelId == boundaryLevelId &&
+            var geoJsonList = _context.BoundaryData.Where(geoJson =>
+                geoJson.BoundaryLevel.Id == boundaryLevelId &&
                 codes.Contains(geoJson.Code)).ToList();
```
benoutram, line 41: I think this query on the `BoundaryData` table would benefit from having an index on the `Code` column which doesn't exist in the configuration you've applied to it so far.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Database/StatisticsDbContext.cs
```diff
@@ -102,6 +102,7 @@ private void Configure(int? timeout = null, bool updateTimestamps = true)
             }
         }
 
+        public DbSet<BoundaryData> BoundaryData { get; set; } = null!;
```
benoutram, line 105: I think you can delete the `GeoJson` type now mentioned in L113 below and the reference to the `geojson` view in method `ConfigureGeoJson`.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240621123907_EES5170_AddBoundaryDataTable.cs
```diff
@@ -0,0 +1,48 @@
...
+
+#nullable disable
+
+namespace GovUk.Education.ExploreEducationStatistics.Data.Model.Migrations
+{
+    /// <inheritdoc />
+    public partial class EES5170_AddBoundaryDataTable : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
```
benoutram, line 11: I think you can add custom code a the end of this method to drop the `dbo.geojson` view.

You can also delete the `Current_View_geojson.sql` file in the `Migrations` folder.

I think we discussed some other stuff which can eventually be cleaned up at a later date like the `geometry2json` function and the geometry tables so it would be good to create a separate 'tidy-up' ticket for these.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/BoundaryLevelServiceTests.cs
```diff
@@ -0,0 +1,95 @@
...
+    //    // Arrange
+    //    _boundaryLevelRepository
+    //        .Setup(blr => blr.Update(
+    //            It.IsAny<long>(),
+    //            It.IsAny<string>()))
+    //        .Returns(Task.CompletedTask);
+
+    //    // Act & Assert
+    //    await Assert.IsAssignableFrom<Task>(async () => await _sut.UpdateLabel(1, "Boundary Level 1 (UPDATED)")).;
+    //}
```
benoutram, line 94: Rather than concentrating on finishing these tests in `BoundaryLevelServiceTests` and `BoundaryDataServiceTests` I would look into rewriting `BoundaryDataControllerTests` as integration tests which then exercise all the functionality in the services and repositories.

See `Admin.Tests.Controller.Api.PublicationControllerTests` as an example.

You can then delete all the service and repository unit tests being added here unless there's anything shared by features which wouldn't be covered by that controller test, or some edge cases which still make sense to be unit tested.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Statistics/BoundaryDataControllerTests.cs
```diff
@@ -0,0 +1,106 @@
...
+        var response = await _sut.UploadBoundaryFile(
+            GeographicLevel.School,
+            "Boundary Data 1",
+            geoJsonFile,
+            DateTime.UtcNow);
+
+        // Assert
+        Assert.IsAssignableFrom<BadRequestObjectResult>(response);
+        Assert.Equal("Invalid file format. This function only accepts GeoJSON files", (response as BadRequestObjectResult).Value);
+    }
```
benoutram, line 105: See comment left in `BoundaryLevelServiceTests.cs` about switching to integration tests.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/BoundaryDataServiceTests.cs
benoutram, line 1: See my comment left in `BoundaryLevelServiceTests.cs` about replacing this unit test class with controller integration tests.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/BoundaryDataService.cs
```diff
@@ -0,0 +1,51 @@
...
+        foreach (var feature in featureCollection.Features)
+        {
+            var (name, code) = BoundaryDataUtils.GetFeatureDetails(feature.Properties);
+
+            boundaryData.Add(new()
+            {
+                BoundaryLevel = boundaryLevel,
+                Name = name,
+                Code = code,
+                GeoJson = JsonConvert.SerializeObject(feature)
```
benoutram, line 45: Same comment that @mmyoungman has already left you in `BoundaryData.cs`. I don't think we should be doing this serialisation here, but make the field of type `Feature`  and move the configuration to `StatisticsDbContext`.

src/GovUk.Education.ExploreEducationStatistics.Data.ViewModels/LocationViewModelBuilder.cs
```diff
@@ -10,7 +11,7 @@ public static class LocationViewModelBuilder
     public static Dictionary<GeographicLevel, List<LocationAttributeViewModel>> BuildLocationAttributeViewModels(
         IList<Location> locations,
         Dictionary<GeographicLevel, List<string>>? hierarchies,
-        Dictionary<GeographicLevel, Dictionary<string, GeoJson>>? geoJson = null)
+        Dictionary<GeographicLevel, Dictionary<string, BoundaryData>>? geoJson = null)
```
benoutram, line 14: Because we are renaming the type from `BoundaryData` to `GeoJson` I think you should consider renaming all the `geoJson` variables that occur in many places, not just here.

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/SubjectResultMetaServiceTests.cs
```diff
@@ -670,236 +670,236 @@ await using (var statisticsDbContext = InMemoryStatisticsDbContext(statisticsDbC
...
+
+        //        var regionOption1SubOption3 = regionOption1.Options[2];
+        //        Assert.Equal(observations[2].Location.Id, regionOption1SubOption3.Id);
+        //        Assert.Equal(_eastMidlands.Name, regionOption1SubOption3.Label);
+        //        Assert.Equal(_eastMidlands.Code, regionOption1SubOption3.Value);
+        //        Assert.NotNull(regionOption1SubOption3.GeoJson);
+        //        Assert.Null(regionOption1SubOption3.Level);
+        //        Assert.Null(regionOption1SubOption3.Options);
+        //    }
+        //}
```
benoutram, line 902: Just a reminder of the TODO still outstanding to get this fixed up.

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/LocationViewModelBuilderTests.cs
```diff
@@ -131,130 +131,130 @@ public void BuildLocationAttributeViewModels_Flat()
...
+    //    // GeoJson is missing as it was not provided for this code
+    //    Assert.Null(la2.GeoJson);
+
+    //    var la3 = localAuthorities[2];
+    //    Assert.Equal(locations[4].Id, la3.Id);
+    //    Assert.Equal(_sunderland.Name, la3.Label);
+    //    Assert.Equal(_sunderland.Code, la3.Value);
+    //    Assert.NotNull(la3.GeoJson);
+    //    Assert.Equal(_sunderland.Code, la3.GeoJson["code"].ToString());
+    //}
```
benoutram, line 257: Just a reminder of the TODO still outstanding to get this fixed up.

benoutram (changes requested):
Finishing the review I originally submitted on 26th.

src/GovUk.Education.ExploreEducationStatistics.Admin/GovUk.Education.ExploreEducationStatistics.Admin.csproj
```diff
@@ -26,6 +26,7 @@
 
   <ItemGroup>
     <PackageReference Include="AspectInjector" Version="2.8.2" PrivateAssets="all" />
+    <PackageReference Include="GeoJSON.Net" Version="1.2.19" />
```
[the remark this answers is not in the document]
tomjonesdev: I tried that one first but couldn't get it to deserialise the data. Check our earlier DMs from around the time I sent the example console app, IIRC I mentioned the issue I saw with the newer version.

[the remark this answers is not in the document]
tomjonesdev: image

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/BoundaryLevelService.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        await _boundaryLevelRepository.Update(id, label);
+    }
+
+    private static BoundaryLevelViewModel Map(BoundaryLevel boundaryLevel)
```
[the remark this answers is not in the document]
tomjonesdev: This would mean that the view model class would need to be declared within (or accessible to) the data layer, which I think should be avoided.

I'll leave it where it is for now because it's not used anywhere else, but can easily be moved to a utility class in the future if needed.

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/BoundaryDataUtils.cs
tomjonesdev: I'll refactor as per Ben's suggestion above, so comments shouldn't be necessary.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240621123907_EES5170_AddBoundaryDataTable.cs
tomjonesdev: Created EES-5292 for the removals.

src/explore-education-statistics-admin/src/pages/bau/BoundaryDataPage.tsx
tomjonesdev: It looks like `locationLevelsMap` is a dictionary, rather than a list I can iterate over.

I can add codes to the existing object, but I'd still need to manually query each one by a key, rather than just adding a simple loop, so I'm not convinced on the benefit of refactoring this from hard-coded. The same applies to the boundary types (assuming I mimic the dictionary structure from location levels).

src/explore-education-statistics-admin/src/pages/bau/BoundaryDataUploadPage.tsx
tomjonesdev: See comment on `BoundaryDataPage`.

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/BoundaryDataUtils.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            throw new ArgumentException($"{nameKey} does not match expected format (expects key ending \"NM\")");
```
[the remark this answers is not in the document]
tomjonesdev: I've added `Either` to the calling methods in the controller/service layers, but would rather avoid adding MVC features to `Common`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/BoundaryLevelService.cs
tomjonesdev: I've implemented the primary constructor, but this class existed before and made use of a separate repo layer, so I'd rather maintain that pattern than bring in database-related dependencies and muddying the current separation.

src/explore-education-statistics-admin/src/pages/bau/BoundaryDataPage.tsx
sambiramairelogic: Dictionary is just a custom type we've written to wrap up TS's `<Record>` type anyway. It's just an object - so I think we could just iterate over it like so: 
image

(We could also return that as actual JSX rather than the interpolated string, but I was just being quick to prove the point)

sambiramairelogic: Tom has a very good point with YAGNI, but my personal tiebreaker votes goes to:
image

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
```diff
@@ -70,17 +68,24 @@
...
+using System.Threading;
+using System.Threading.Tasks;
 using Thinktecture;
 using static GovUk.Education.ExploreEducationStatistics.Common.Utils.StartupUtils;
 using ContentGlossaryService = GovUk.Education.ExploreEducationStatistics.Content.Services.GlossaryService;
 using ContentMethodologyService = GovUk.Education.ExploreEducationStatistics.Content.Services.MethodologyService;
 using ContentPublicationService = GovUk.Education.ExploreEducationStatistics.Content.Services.PublicationService;
 using ContentReleaseService = GovUk.Education.ExploreEducationStatistics.Content.Services.ReleaseService;
 using DataGuidanceService = GovUk.Education.ExploreEducationStatistics.Admin.Services.DataGuidanceService;
+//using DataImportService = GovUk.Education.ExploreEducationStatistics.Admin.Services.DataImportService;
```
mmyoungman, line 88: Rogue comment

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
```diff
@@ -499,6 +499,9 @@ public virtual void ConfigureServices(IServiceCollection services)
             });
             services.AddTransient<IEmailService, EmailService>();
 
+            //services.AddSingleton<IDatabaseHelper, DatabaseHelper>();
```
mmyoungman, line 502: Rogue comments. More below too.

src/GovUk.Education.ExploreEducationStatistics.Data.ViewModels/LocationViewModelBuilder.cs
```diff
@@ -10,61 +11,69 @@ public static class LocationViewModelBuilder
...
-        var geoJson = code.IsNullOrEmpty()
+        var feature = code.IsNullOrEmpty()
             ? null
-            : geoJsonByCode?.GetValueOrDefault(code)?.Deserialized;
+            : boundaryDataByCode?.GetValueOrDefault(code)?.GeoJson;
+
+        if (feature != null)
+        {
+            // A recent change (from either of the two latest commits) is causing these to be added multiple times, throwing exceptions due to duplicates.
+            // TryAdd prevents the exception, but doesn't resolve the duplicate calls
```
mmyoungman, line 68: Something to address here?

src/explore-education-statistics-admin/src/pages/bau/BoundaryDataPage.tsx
```diff
@@ -0,0 +1,140 @@
...
+            )}
+          >
+            {Object.keys(locationLevelsMap).map(key => {
+              return (
+                <>
+                  <dt>{locationLevelsMap[key].code}</dt>
+                  <dd>{locationLevelsMap[key].label}</dd>
+                </>
+              );
+            })}
```
mmyoungman, line 60: Still a non-believer? :)

mmyoungman (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Data.Model/GovUk.Education.ExploreEducationStatistics.Data.Model.csproj
benoutram: Just checked and I think the issue is with a tab character that's crept in here.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Database/StatisticsDbContext.cs
```diff
@@ -152,6 +154,32 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+                .IsRequired();
+
+            modelBuilder.Entity<BoundaryData>()
+                .HasOne(boundaryData => boundaryData.BoundaryLevel)
+                .WithMany()
+                .IsRequired();
+
+            modelBuilder.Entity<BoundaryData>()
+                .Property(boundaryData => boundaryData.GeoJson)
+                .IsRequired();
```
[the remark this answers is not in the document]
benoutram: Configuration looks better now that some other issues have been resolved but I'm guessing you decided not to go with this version to remain consistent with the other 'Configure' methods?

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Statistics/BoundaryDataController.cs
```diff
@@ -0,0 +1,107 @@
...
+        [FromForm] GeographicLevel level,
+        [FromForm] string label,
+        IFormFile file,
+        [FromForm] DateTime published)
+    {
+        return await boundaryLevelService
+            .ListBoundaryLevels()
+            .OnSuccessDo(async existingLevels =>
+            {
+                if (file.FileName.Split('.')[1] != "geojson")
```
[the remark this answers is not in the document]
tomjonesdev: I've moved file name and label validation to a FluentValidation class.

Regarding the view model creation, I moved this function from where it was before based on another comment. I wasn't happy moving it to a controller because the presentation layer now has a reference to a database entity. I can see that other places have view model creation in service layers (which I'm quite sure is where I put it before, imo where it should be), but essentially there doesn't appear to be a consistent approach atm so I'm not moving it again :D.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Database/StatisticsDbContext.cs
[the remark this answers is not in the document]
tomjonesdev: Correct 👍

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/BoundaryLevelService.cs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    public async Task<Either<ActionResult, BoundaryLevel>> UpdateBoundaryLevel(
+        long id,
+        string label)
+    {
+        return await statisticsPersistenceHelper
+            .CheckEntityExists<BoundaryLevel, long>(id)
+            .OnSuccessDo(async _ => await boundaryLevelRepository.UpdateBoundaryLevel(id, label));
+    }
```
benoutram, line 54: I know you've already mentioned the existence of the `BoundaryLevelRepository` as the reason for wanting to keep separation between the service and database logic here which proably makes sense for the `CreateBoundaryLevel` as that's a bit more involved but I think the rest of the methods would benefit from using the `StatisticsDbContext` as a dependency here rather than `IPersistenceHelper<StatisticsDbContext>`.

Recently we are attempting to avoid the use of `IPersistenceHelper` because it can be replaced with extension methods that we've added for the database context like `FirstOrNotFoundAsync` and `SingleOrNotFoundAsync`.

If you depend on the database context directly, at least the `ListBoundaryLevels`, `GetBoundaryLevel` and `UpdateBoundaryLevel` code that's split across the service and repository can all be simplified as follows:

```csharp
    public async Task<Either<ActionResult, List<BoundaryLevel>>> ListBoundaryLevels(
    {
        return await context.BoundaryLevel.ToListAsync(cancellationToken);
    }

    public async Task<Either<ActionResult, BoundaryLevel>> GetBoundaryLevel(
        long id, CancellationToken cancellationToken = default)
    {
        return await context.BoundaryLevel.FirstOrNotFoundAsync(level => level.Id == id, cancellationToken);
    }

    public async Task<Either<ActionResult, BoundaryLevel>> UpdateBoundaryLevel(
        long id,
        string label,
    {
        return await context.BoundaryLevel
            .FirstOrNotFoundAsync(level => level.Id == id, cancellationToken)
            .OnSuccess(async level =>
            {
                level.Label = label;
                await context.SaveChangesAsync(cancellationToken);
                return level;
            });
    }
```

Note I've also added `CancellationToken cancellationToken` to these methods as we've recently tried to start propagating `CancellationToken` down from requests.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/BoundaryData.cs
```diff
@@ -0,0 +1,24 @@
...
+    public string? Code { get; set; }
+
+    [Required]
+    public string? Name { get; set; }
+
+    [Required]
+    public Feature? GeoJson { get; set; }
+
+    [Required]
+    public BoundaryLevel? BoundaryLevel { get; set; }
```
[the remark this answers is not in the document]
tomjonesdev: They are, but `Feature` can't just be `new()` as the constructor requires a dictionary of properties.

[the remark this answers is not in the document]
benoutram: I think you want something like this instead

```csharp
public class BoundaryData
{
    public int Id { get; set; }

    public required string Code { get; set; }

    public required string Name { get; set; }

    public Feature GeoJson { get; set; } = null!;

    public required long BoundaryLevelId { get; set; }

    public BoundaryLevel? BoundaryLevel { get; set; } = null!;
}
```

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Statistics/BoundaryLevelControllerTests.cs
```diff
@@ -0,0 +1,178 @@
...
+
+        var client = TestApp
+            .SetUser(AuthenticatedUser(RoleClaim(RoleNames.BauUser)))
+            .CreateClient();
+
+        var response = await client.GetAsync("api/boundary-level");
+
+        var result = response.AssertOk<List<BoundaryLevelViewModel>>();
+
+        Assert.Multiple(
```
mmyoungman, line 55: We should make this the standard way of asserting - it is inefficient with other unit tests where you only see one failure at a time.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Statistics/BoundaryLevelControllerTests.cs
```diff
@@ -0,0 +1,178 @@
...
+                    Level = GeographicLevel.Region,
+                    Published = DateTime.Now,
+                });
+        });
+
+        var client = TestApp
+            .SetUser(AuthenticatedUser(RoleClaim(RoleNames.BauUser)))
+            .CreateClient();
+
+        var path = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)!, @"REG - Regions England BUC 202212.geojson");
```
mmyoungman, line 154: I'd move the geojson file into the `Resources` directory. This is the pattern we use in other places - see `DataArchiveValidationServiceTests#CreateFormFileFromResource`

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/GovUk.Education.ExploreEducationStatistics.Admin.Tests.csproj
```diff
@@ -54,4 +54,10 @@
     </Content>
   </ItemGroup>
 
+  <ItemGroup>
+    <None Update="REG - Regions England BUC 202212.geojson">
```
mmyoungman, line 58: This can be removed once the file is moved to the `Resources` directory.

mmyoungman (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/BoundaryDataUtils.cs
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    public static string? GetName(IDictionary<string, object> properties)
+    {
+        var key = GetKeyBySuffix(properties, "NM");
+
+        return properties.TryGetValue(key, out var name)
+            ? name.ToString() ?? ""
+            : throw new ArgumentException("Required key not found (expects key ending 'NM')");
+    }
```
benoutram, line 25: Return type should be `string` rather than `string?` now after recent changes.

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/BoundaryLevelCreateRequest.cs
```diff
@@ -0,0 +1,50 @@
...
+                .NotEmpty()
+                .MustAsync(async (label, cancellation) =>
+                {
+                    var existingLevels = await boundaryLevelService.ListBoundaryLevels();
+
+                    return !existingLevels.IsLeft && !existingLevels.Right
+                        .Select(el => el.Label)
+                        .ToList()
+                        .Contains(label);
+                })
```
benoutram, line 38: I don't think we've got any examples of using services within validators to know how best to handle a service failure like this off the top of my head, but you could at least simplify this as follows:

```csharp
                .MustAsync(async (label, cancellationToken) =>
                {
                    var existingLevels = await boundaryLevelService.ListBoundaryLevels(cancellationToken);
                    return existingLevels.IsRight && !existingLevels.Right.Any(bl => bl.Label == label);
                })
```

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Statistics/BoundaryLevelControllerTests.cs
```diff
@@ -0,0 +1,178 @@
...
+                    Level = GeographicLevel.Country,
+                    Published = DateTime.Now,
+                },
+                new()
+                {
+                    Id = 2,
+                    Created = DateTime.Now,
+                    Label = "Boundary Level 2",
+                    Level = GeographicLevel.Country,
+                    Published = DateTime.Now,
```
benoutram, line 43: You could make these different levels and compare levels in the assertion.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Statistics/BoundaryLevelControllerTests.cs
```diff
@@ -0,0 +1,178 @@
...
+        await using var context = TestApp.GetDbContext<StatisticsDbContext>();
+
+        var savedBoundaryLevel = await context.BoundaryLevel.SingleAsync(bl => bl.Id == 1);
+        var savedBoundaryData = await context.BoundaryData.Include(bd => bd.BoundaryLevel).FirstAsync();
+
+        Assert.Multiple(
+            () => Assert.NotNull(savedBoundaryLevel),
+            () => Assert.NotNull(savedBoundaryData),
+            () => Assert.Equal(GeographicLevel.Region, savedBoundaryLevel.Level)
+        );
```
benoutram, line 176: I'm not sure the `Include(bd => bd.BoundaryLevel)` is doing anything here? I think .Where(bd => db.BoundaryLevel.Id == 1)` would be more appropriate.

`FirstAsync` is going to blow up if there aren't any so I think that makes the check `Assert.NotNull(savedBoundaryData)`  redundant too. Maybe you could assert on the count `BoundaryData` for that level?

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/BoundaryDataUtilsTests.cs
```diff
@@ -0,0 +1,143 @@
+using GovUk.Education.ExploreEducationStatistics.Common.Utils;
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils;
+public class BoundaryDataUtilsTests
+{
+    [Fact]
+    public void GetCode_NoMatchingSuffix_ThrowsArgumentException()
```
benoutram, line 10: Not really a big deal but the new tests which have been written aren't following the backend styleguide point 3.1.1 which we came up with fairly recently which says

> Create a nested class on a per-public-method being tested basis {MethodUnderTest}Tests

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Repository/GeoJsonRepository.cs
```diff
@@ -1,68 +1,68 @@
...
 using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Data.Model.Database;
 using GovUk.Education.ExploreEducationStatistics.Data.Model.Repository.Interfaces;
+using System;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace GovUk.Education.ExploreEducationStatistics.Data.Model.Repository
 {
     public class GeoJsonRepository : IGeoJsonRepository
```
benoutram, line 11: I think this should be renamed `BoundaryDataRepository`.

When this code is deployed and the migration is run, the table is initially empty. I think we might need to test that existing charts are cached and continue to render ok? I'm assuming you are planning to run the script to populate the `BoundaryData` table as soon as you see the table get created?

mmyoungman: Might be worth adding details about how this will be deployed to the PR description. Seems like you're going to need a faffy deploy ticket too.

To be super safe, it might also be worth ensuring this doesn't get merged into dev with any other major changes so if we do need to roll back we can do it easily.

tomjonesdev: Yep that's right @benoutram - presumably the existing charts will be thoroughly checked during QA, but they seemed to function locally.

I've raised EES-5330 for the faffy deployment.

mmyoungman (approved):

benoutram (approved):

PR #4996, opened 2024-06-24
EES-4716 Fix various issues with data set query endpoint

ntsim:
This PR:

- Fixes query keys that are set to empty strings in the request being bound as null in the request model. For example, `?locations.eq=` would be bound as null instead of an empty string, resulting in unexpected behaviour in the API.
- Fixes invalid values not resulting in validation errors. These are typically where a value has been provided, but is the wrong type or format e.g. `?page=ten` (instead of `?page=10`). We now return `InvalidValue` errors in these cases.
- Fixes unknown fields in the request not resulting in validation errors. We now output an `UnknownField` error whenever these are encountered in the body or query parameters.

## Relevant changes

- Renamed the `RequiredField` error to `RequiredValue`.
- Changed `InvalidRequestInputFilter` to be an `IAsyncActionFilter` instead of `IResultFilter`.
- Added new `IsSimple` and `IsComplex` extensions for `Type`. Any primitives or types that can be represented as a single value are considered 'simple'.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Filters/InvalidRequestInputFilter.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Filters/InvalidRequestInputFilter.cs
@@ -1,9 +1,9 @@
 using System.Diagnostics.CodeAnalysis;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Utils;
 using GovUk.Education.ExploreEducationStatistics.Common.Validators;
 using GovUk.Education.ExploreEducationStatistics.Common.ViewModels;
 using Microsoft.AspNetCore.Mvc;
-using Microsoft.AspNetCore.Mvc.Abstractions;
 using Microsoft.AspNetCore.Mvc.Filters;
 using Microsoft.AspNetCore.Mvc.Infrastructure;
 using Microsoft.AspNetCore.Mvc.ModelBinding;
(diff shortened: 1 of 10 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Filters/InvalidRequestInputFilter.cs
```diff
@@ -105,33 +105,51 @@ public Task OnActionExecutionAsync(ActionExecutingContext context, ActionExecuti
...
+            else
+            {
+                completed = true;
+            }
         }
 
-        return JsonPathUtils.Concat(paths);
+        // Not ideal, but there isn't any better way of figuring out if we have this type
+        // of JsonException as they don't provide any error codes or other identifier.
+        var message = currentException is not null
```
duncan-at-hiveit, line 156: Gnarly!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Filters/InvalidRequestInputFilter.cs
```diff
@@ -91,6 +93,21 @@ public Task OnActionExecutionAsync(ActionExecutingContext context, ActionExecuti
         return Task.CompletedTask;
     }
 
+    private static List<ErrorViewModel> GetUnknownQueryParameterErrors(
```
duncan-at-hiveit, line 96: Nice work!

duncan-at-hiveit (approved):

PR #5007, opened 2024-06-26
Ees 5113 investigate partial json updates

duncan-at-hiveit:
# Overview

This PR:
- remodels the JSON structure of the DataSetVersionMapping JSON columns to allow for easier updates using JSONB paths and PostgreSQL commands.

This effectively replaces Lists with Dictionaries in the situations where we would originally be using one of the fields within an array element as its key to determine which piece of JSON to update.  Now, instead of looping through an array to find an element with an appropriate key, we instead use this same key as the *Key* for a Dictionary entry and the rest of the element's values as the *Value* for that same Dictionary entry.

An example of this change can be seen on the investigation ticket https://dfedigital.atlassian.net/browse/EES-5113, where we use the GeographicLevel as a Key to a Dictionary that contains Location Mappings for different GeographicLevels, and then uses the Key of the source LocationOption as the Key to a Dictionary that contains mappings from source LocationOptions to their respective candidate LocationOptions.

# Additional changes

## Simpler DbContext configuration

Because we're now no longer looking to leverage Entity Framework to do the JSON fragment updates (because this requires the full JSON graph to be deserialised before we can update tiny parts of it, which would not be performant), we no longer need to tell EF the ownership structure of the JSON elements.  Thus they can simply be registered as JSONB column fields.

## Convenience constructors

Because now we've removed the "Key" field from `MappableElement`, this leaves only the "Label" field remaining.  I've therefore changed this to a single-arg constructor for brevity.

## Extracted repeated code around generating keys for mappable elements

There were various places where we repeat the process of generating keys for LocationOptions, FilterOptions and Filters.  Rather than repeat in multiple places, I've extracted them into reusable compents.


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetVersionMapping.cs
```diff
@@ -38,52 +38,27 @@ public void Configure(EntityTypeBuilder<DataSetVersionMapping> builder)
...
+                    v => JsonSerializer.Serialize(v, new JsonSerializerOptions(JsonSerializerDefaults.General)),
+                    v => JsonSerializer.Deserialize<LocationMappingPlan>(v,
+                        new JsonSerializerOptions(JsonSerializerDefaults.General))!);
+
+            builder.Property(p => p.FilterMappingPlan)
+                .HasColumnType("jsonb")
+                .HasConversion(
+                    v => JsonSerializer.Serialize(v, new JsonSerializerOptions(JsonSerializerDefaults.General)),
+                    v => JsonSerializer.Deserialize<FilterMappingPlan>(v,
+                        new JsonSerializerOptions(JsonSerializerDefaults.General))!);
```
jack-hive, line 61: Just wondering if we need to bother with the `new JsonSerializerOptions(JsonSerializerDefaults.General)` bits. Would it work if we removed the optional options parameter in each of the `JsonSerializer.Serialize(...)` and `JsonSerializer.Deserialize(...)` lines?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionMappingService.cs
```diff
@@ -186,44 +208,38 @@ private static void AutoMapFilters(FilterMappingPlan filtersPlan)
...
+                        Candidates = candidatesForLevel
+                            .ToDictionary(
+                                keySelector: LocationOptionKeyGenerator,
+                                elementSelector: CreateLocationOptionFromMetaRow)
+                    };
+                });
+
+        var sourceLevels = sourceMappingsByLevel
+            .Select(level => level.Key)
+            .Distinct();
```
jack-hive, line 242: Think you can just do `var sourceLevels = sourceMappingsByLevel.Keys` here :)

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ProcessNextDataSetVersionFunctionTests.cs
```diff
@@ -225,72 +228,80 @@ public async Task Success_Mappings()
...
+                                        Code = option.ToRow().Code,
+                                        OldCode = option.ToRow().OldCode,
+                                        Urn = option.ToRow().Urn,
+                                        LaEstab = option.ToRow().LaEstab,
+                                        Ukprn = option.ToRow().Ukprn
+                                    }
+                                })
+                    });
 
             // There should be 5 levels of mappings when combining all the source and target levels.
```
jack-hive, line 281: think this will be 6 now won't it?

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ProcessNextDataSetVersionFunctionTests.cs
duncan-at-hiveit: It's still 5 because of the shared levels between the 2 data set versions.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionMappingService.cs
duncan-at-hiveit: Ahh yeah, right you are again.  Changed ta

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetVersionMapping.cs
duncan-at-hiveit: Done, and it's working fine

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ProcessNextDataSetVersionFunctionTests.cs
jack-hive: Ah yes! sorry, I didn't realise `School` was in the new data set

jack-hive (approved):

PR #5013, opened 2024-06-28
EES-4823: Pre-empt content urls that will redirect

sambiramairelogic:
Google has crawled over 200 links on our site which hit redirects, and is marking us down for this (in Michael's [spreadsheet](https://docs.google.com/spreadsheets/d/1Hqha2UHV-0L0dqOFFIo-UJRsLUgzFK3Lhz9Uq57k_6U/edit?gid=1190542568#gid=1190542568))

A sizable chunk of these are renamed publications/methodologies as per `<contentApi>/redirects`, and a few others also hit redirect rules introduced to `server.js` (i.e., ending in a trailing slash or having a hostname starting with www). 

This PR applies all our redirect rules to hrefs before they are rendered by `ContentHtml` to pre-empt these redirects. This unfortunately means making an API call to the content API each time we render a link, which is hardly ideal. I've left a TODO comment in the files below to prompt a discussion on this - there's a few ways we could go.


src/explore-education-statistics-common/src/services/redirectService.ts
```diff
@@ -10,11 +12,14 @@ interface Redirect {
...
 
-const contentApiUrl = process.env.CONTENT_API_BASE_URL;
+export const redirectPathStarts = {
+  methodologies: '/methodology',
+  publications: '/find-statistics',
+};
 
 const redirectService = {
   async list(): Promise<Redirects> {
-    return (await fetch(`${contentApiUrl}/redirects`)).json();
```
sambiramairelogic, line 17: Eurgh, so apparently this causes build errors in the frontend. We could re-implement the frontend version of this service and have two, but that's messy. Problem as things stand I need to make this call available from `ContentHtml` which is in common. 

@ntsim  I'm gonna hold off solving this pending your thoughts on my big TODO: comment in `AnchorHtml`, as I suspect there might be a few changes 
image

src/explore-education-statistics-common/src/components/AnchorHtml.tsx
```diff
@@ -0,0 +1,40 @@
+import { OmitStrict } from '@common/types';
+import applyRedirectRules from '@common/utils/url/applyRedirectRules';
+import React, { AnchorHTMLAttributes, useEffect, useState } from 'react';
+
+interface AnchorHtmlProps
+  extends OmitStrict<AnchorHTMLAttributes<HTMLAnchorElement>, 'href'> {
+  href: string;
+}
+
+export default function AnchorHtml({
```
ntsim, line 10: Would just call this `Link`, or `ContentLink` if you want to avoid confusion with the other `Link` components.

src/explore-education-statistics-common/src/components/ContentHtml.tsx
```diff
@@ -87,11 +88,11 @@ export default function ContentHtml({
         return !node.attribs.href.includes(
           'explore-education-statistics.service.gov.uk',
         ) && typeof node.attribs['data-featured-table'] === 'undefined' ? (
-          <a href={url} target="_blank" rel="noopener noreferrer">
+          <AnchorHtml href={url} target="_blank" rel="noopener noreferrer">
```
ntsim, line 91: Do we need redirects for external links? We don't have redirects for external URLs, so this doesn't seem necessary

src/explore-education-statistics-common/src/utils/url/applyRedirectRules.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const prodPublicUrl = 'https://explore-education-statistics.service.gov.uk/';
+
+export default async function applyRedirectRules(url: string): Promise<string> {
+  let newUrl: string = url;
+
+  newUrl = await noTrailingSlashUnlessOnlyHost(newUrl);
+  newUrl = await pathMustNotEndInSlash1000(newUrl);
+  newUrl = await publicSiteHostMustNotStartWithWww(newUrl);
+  newUrl = await mustNotTriggerAContentApiRedirect(newUrl);
```
ntsim, line 14: IMO, this approach is potentially over-complicating things by trying to be functional. 

I'm not really sure it's worth the effort as we don't typically write code like this, and it's less performant as it has to parse into a `URL` object multiple times and is using regexes over the entire URL in multiple places.

Would suggest that we just construct the `URL` object once and potentially just simplifying this back to some `if` statements (with comments) which modify the `url` i.e.

```ts
const newUrl = new URL(url);

// Do a check
if (blah) {
  newUrl.pathname = ...;
}

// Do another check
if (blah) {
  newUrl.hostname = ...;
}

// Using functions is still fine for more
// contained pieces of logic.
await callAFunction(newUrl);

return newUrl.toString();
```

src/explore-education-statistics-common/src/utils/url/applyRedirectRules.ts
```diff
@@ -0,0 +1,93 @@
...
+  RedirectType,
+  redirectPathStarts,
+} from '@common/services/redirectService';
+
+const prodPublicUrl = 'https://explore-education-statistics.service.gov.uk/';
+
+export default async function applyRedirectRules(url: string): Promise<string> {
+  let newUrl: string = url;
+
+  newUrl = await noTrailingSlashUnlessOnlyHost(newUrl);
```
ntsim, line 11: If we don't modify as suggested in comment below, then we should remove any unnecessary `awaits`. This is only necessary on the function that calls `redirectService.list`.

src/explore-education-statistics-common/src/utils/url/applyRedirectRules.ts
```diff
@@ -0,0 +1,93 @@
+import redirectService, {
+  RedirectType,
+  redirectPathStarts,
+} from '@common/services/redirectService';
+
+const prodPublicUrl = 'https://explore-education-statistics.service.gov.uk/';
```
ntsim, line 6: Is it the right approach to use this as the base? If we're on non-prod environments, won't this potentially result in incorrect links in those environments?

I mentioned this in your other PR, but we should probably set the host's URL using `setHostUrl` / `getHostUrl` functions and use this instead?

src/explore-education-statistics-common/src/utils/url/applyRedirectRules.ts
```diff
@@ -0,0 +1,93 @@
...
+            .join('/');
+        }
+      }
+      return acc;
+    }, '');
+
+    if (redirectPath) {
+      return new URL(`${redirectPath}${parsedUrl.search}`, prodPublicUrl).href;
+    }
+  }
```
ntsim, line 90: Don't need the `reduce` and we're splitting out the path segments too many times.

This can be simplified to:

```ts
const redirectPathStart = Object.entries(redirectPathStarts).find(
  ([_, path]) => parsedPath.startsWith(path),
);
const pathSegments = parsedPath.split('/');

if (redirectPathStart && pathSegments.length > 1) {
  const [key] = redirectPathStart;
  const redirectType = key as RedirectType;

  const rewriteRule = redirectsFromContentApi[redirectType].find(
    ({ fromSlug }) => pathSegments[2].toLowerCase() === fromSlug,
  );

  if (rewriteRule) {
    const path = pathSegments
      .map(segment =>
        segment.toLowerCase() === rewriteRule?.fromSlug
          ? rewriteRule?.toSlug
          : segment.toLowerCase(),
      )
      .join('/');

    return new URL(`${path}${parsedUrl.search}`, prodPublicUrl).href;
  }
}
```

ntsim (changes requested):

benoutram:
Closing this PR as Nick has suggested an alternative approach in the discussion comments. Although Sam has had a good go at implementing it, it has been left as a WIP in the most recent commits on this branch and this PR is now stale. I have documented the history of this in EES-6337 for whoever picks this up again so they can look at this PR and see Nick's alternative approach.

PR #5023, opened 2024-07-02
EES-5112 - endpoint for autosaving location mappings

duncan-at-hiveit:
This PR:
* Adds an endpoint to allow updates to be made by the user of the Location-to-Location mappings in the DataSetVersionMapping table.

The new endpoint will be used by the autosave functionality that is planned for the Location (and Filter) mapping pages when adding a new data set version.  We plan to use a debounced autosave endpoint and therefore are expecting to batch up multiple autosave requests.  The `BatchLocationMappingUpdatesRequest` therefore is expecting a list of update requests at a time and will action all of them.

If any of the update requests should fail for whatever reason, the whole batch of autosave updates will also fail (as in the transaction will roll them all back together).


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionMappingService.cs
```diff
@@ -0,0 +1,136 @@
...
+    private async Task<Either<ErrorViewModel, LocationMappingUpdateResponse>> UpdateLocationOptionMapping(
+        Guid nextDataSetVersionId,
+        LocationMappingUpdateRequest updateRequest,
+        int index,
+        CancellationToken cancellationToken = default)
+    {
+        var updateJsonRequest = new JsonbPathRequest<Guid>
+        {
+            TableName = nameof(PublicDataDbContext.DataSetVersionMappings),
+            IdColumnName = "TargetDataSetVersionId",
```
jack-hive, line 86: could this be `IdColumnName = nameof(DataSetVersionMapping.TargetDataSetVersionId)`

src/GovUk.Education.ExploreEducationStatistics.Common/Services/PostgreSqlRepository.cs
```diff
@@ -0,0 +1,99 @@
...
+        where TResponse : class
+    {
+        var jsonPathParam = new NpgsqlParameter("jsonPath", request.PathSegments);
+        var rowIdParam = new NpgsqlParameter("rowId", request.RowId);
+
+        var response = await context
+            .Set<JsonFragment>()
+            .FromSqlRaw(
+                sql: $"""
+                      SELECT "{request.JsonbColumnName}"#>@jsonPath "JsonString" FROM "{request.TableName}"
```
jack-hive, line 31: Would it be worth doing `"{nameof(JsonFragment.JsonString)}"` in place of `"JsonString"`?

src/GovUk.Education.ExploreEducationStatistics.Common/Services/PostgreSqlRepository.cs
```diff
@@ -0,0 +1,99 @@
...
+                      WHERE "{request.IdColumnName}" = @rowId
+                      """,
+                parameters: [jsonPathParam, rowIdParam])
+            .SingleOrDefaultAsync(cancellationToken);
+
+        return response is not null
+               && response.JsonString is not null
```
jack-hive, line 38: could simplify this to 

```csharp
        return response?.JsonString is not null
            ? JsonSerializer.Deserialize<TResponse>(response.JsonString)
            : null;
```

src/GovUk.Education.ExploreEducationStatistics.Common/Services/PostgreSqlRepository.cs
```diff
@@ -0,0 +1,99 @@
...
+            cancellationToken);
+
+        var jsonUpdateResult = updateValueFn.Invoke(json);
+
+        if (jsonUpdateResult.IsLeft)
+        {
+            return jsonUpdateResult;
+        }
+
+        return await SetJsonbAtPath(context, request, jsonUpdateResult.Right, cancellationToken);
```
jack-hive, line 97: I think you could simplify this to:

```csharp
        return updateValueFn.Invoke(json)
            .OnSuccess(async jsonUpdateResult => await SetJsonbAtPath(context, request, jsonUpdateResult.Right, cancellationToken));
```

src/GovUk.Education.ExploreEducationStatistics.Common/Services/PostgreSqlRepository.cs
```diff
@@ -0,0 +1,99 @@
...
+using Microsoft.EntityFrameworkCore;
+using Npgsql;
+using NpgsqlTypes;
+using JsonSerializer = System.Text.Json.JsonSerializer;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Services;
+
+public class PostgreSqlRepository : IPostgreSqlRepository
+{
+    public async Task<TResponse?> GetJsonbFromPath<TDbContext, TRowId, TResponse>(
```
jack-hive, line 17: this could potentially be `private` for now. But I can definitely see a case where you would use it. Happy for you to leave it as either. Just wanted to point it out :)

src/GovUk.Education.ExploreEducationStatistics.Common/Services/PostgreSqlRepository.cs
```diff
@@ -0,0 +1,99 @@
...
+            .SingleOrDefaultAsync(cancellationToken);
+
+        return response is not null
+               && response.JsonString is not null
+            ? JsonSerializer.Deserialize<TResponse>(response.JsonString)
+            : null;
+    }
+
+#pragma warning disable EF1002
+    public async Task<TValue?> SetJsonbAtPath<TDbContext, TValue, TRowId>(
```
jack-hive, line 44: this could potentially be `private` for now. But I can definitely see a case where you would use it. Happy for you to leave it as either. Just wanted to point it out :)

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/BatchMappingUpdateViewModels.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            RuleForEach(request => request.Updates)
+                .SetValidator(new LocationMappingUpdateRequest.Validator());
+        }
+    }
+}
+
+public record LocationMappingUpdateRequest
+{
+    public GeographicLevel? Level { get; init; }
```
jack-hive, line 26: should this be non-nullable?

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/BatchMappingUpdateViewModels.cs
```diff
@@ -0,0 +1,80 @@
...
+
+    public string? SourceKey { get; init; } = null;
```
jack-hive, line 28: should this be non-nullable?

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/BatchMappingUpdateViewModels.cs
```diff
@@ -0,0 +1,80 @@
...
+
+    public string? CandidateKey { get; init; }
+
+    public MappingType? Type { get; init; }
```
jack-hive, line 32: should this be non-nullable?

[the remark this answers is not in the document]
jack-hive: Would `None` never be valid here? Wonder if we should add that to the validator below. I think it should only be in `AutoNone`, `AutoMapped`, `ManualNone` or `ManualMapped` when it is coming through to this endpoint right?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionMappingControllerTests.cs
jack-hive, line 1: if we do end up moving the trigger into the `DataSetVersionsController` then this would need moving too

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionMappingControllerTests.cs
```diff
@@ -0,0 +1,530 @@
...
+            client ??= BuildApp().CreateClient();
+
+            var uri = new Uri($"{BaseUrl}/{nextDataSetVersionId}/mapping/locations", UriKind.Relative);
+
+            return await client.PatchAsync(uri,
+                new JsonNetContent(new BatchLocationMappingUpdatesRequest { Updates = updates }));
+        }
+    }
+
+    // permission tests (and do we have these for other controller integration tests too)?
```
jack-hive, line 524: is this comment meant to be here?

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Fixtures/TestApplicationFactory.cs
```diff
@@ -52,6 +52,11 @@ protected override IHostBuilder CreateHostBuilder()
                     builder
                         .AddFilter<ConsoleLoggerProvider>("Default", LogLevel.Warning)
                         .AddFilter<ConsoleLoggerProvider>("Microsoft", LogLevel.Warning);
+
+                    // Uncomment to add SQL logging to the debug console.
+                    // .AddFilter<ConsoleLoggerProvider>((category, level) =>
+                    //     category == DbLoggerCategory.Database.Command.Name
+                    //     && level == LogLevel.Information);
```
jack-hive, line 59: are we leaving this here on purpose?

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Public.Data/DataSetVersionMappingController.cs
```diff
@@ -0,0 +1,33 @@
...
+        [FromRoute] Guid nextDataSetVersionId,
+        [FromBody] BatchLocationMappingUpdatesRequest request,
+        CancellationToken cancellationToken)
+    {
+        return await mappingService
+            .ApplyBatchMappingUpdates(
+                nextDataSetVersionId: nextDataSetVersionId,
+                request: request,
+                cancellationToken: cancellationToken)
+            .HandleFailuresOrOk();
```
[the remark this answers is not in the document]
duncan-at-hiveit: As per previous comment, I have updated the base path for the controller

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Public.Data/DataSetVersionMappingController.cs
```diff
@@ -0,0 +1,34 @@
...
+    }
```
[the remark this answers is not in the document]
duncan-at-hiveit: I have gone ahead and updated the base path of the Controller itself to include the mapping path.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/BatchMappingUpdateViewModels.cs
```diff
@@ -0,0 +1,80 @@
...
+                .NotNull()
+                .NotEmpty()
+                .When(request => request.Type is not null && MappedTypes.Contains(request.Type.Value));
+
+            RuleFor(request => request.CandidateKey)
+                .Null()
+                .When(request => request.Type is not null && !MappedTypes.Contains(request.Type.Value));
+        }
+    }
+}
```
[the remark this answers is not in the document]
duncan-at-hiveit: Absolutely!  I have moved the requests into there but kept the responses in the ViewModels namespace (folder).

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/BatchMappingUpdateViewModels.cs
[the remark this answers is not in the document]
duncan-at-hiveit: > when this is `AutoMapped` or `AutoNone`, is it just the user confirming that the automapping was correct?

I flipped and flopped over this one.  Initially it would only be `Manual*` that we'd expect as the user worked through the form, but I envisaged the possibility of a user "undoing" a manual decision they'd made, and returning it back to "Auto*".

Tbh since the UI designs don't currently allow for this, I'll just accept manual mappings for now, and it's not hard to change things down the line if we need to.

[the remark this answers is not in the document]
duncan-at-hiveit: > Would `None` never be valid here? Wonder if we should add that to the validator below. I think it should only be in `AutoNone`, `AutoMapped`, `ManualNone` or `ManualMapped` when it is coming through to this endpoint right?

As per my previous reply, I'm changing this to only accept "Manual*" for now.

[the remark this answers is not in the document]
duncan-at-hiveit: Done, and added some more useful error messages as well to help out the front end a bit more.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionMappingService.cs
duncan-at-hiveit: Good suggestion - done

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionMappingService.cs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    private async Task<Either<ActionResult, DataSetVersionMapping>> CheckMappingExists(
+        Guid nextDataSetVersionId,
+        CancellationToken cancellationToken)
+    {
+        return await publicDataDbContext
+            .DataSetVersionMappings
+            .AsNoTracking()
+            .FirstOrNotFoundAsync(
```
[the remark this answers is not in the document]
duncan-at-hiveit: I've changed

src/GovUk.Education.ExploreEducationStatistics.Common/Services/PostgreSqlRepository.cs
duncan-at-hiveit: Yes nice one, changed!

src/GovUk.Education.ExploreEducationStatistics.Common/Services/PostgreSqlRepository.cs
duncan-at-hiveit: Yup absolutely, good suggestion ta

src/GovUk.Education.ExploreEducationStatistics.Common/Services/PostgreSqlRepository.cs
duncan-at-hiveit: I can if I cheat a bit and have `updateValueFn` return a `Task<Either>` rather than a straight `Either` :grin: 

And I'm actually happy to make this change despite the fact that we're not currently needing Task support in our current use case, because this makes it more flexible for the future.

So I've updated the method signature and added this simplfication too, thanks!

src/GovUk.Education.ExploreEducationStatistics.Common/Services/PostgreSqlRepository.cs
duncan-at-hiveit: Yup going to leave it public for when I implement your suggested validation for checking that the candidate keys are valid candidates!

src/GovUk.Education.ExploreEducationStatistics.Common/Services/PostgreSqlRepository.cs
duncan-at-hiveit: Yup thanks - gonna leave it public for the very reason that it could be useful for the future, ta!

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/BatchMappingUpdateViewModels.cs
duncan-at-hiveit: I was forced to do this so that it is not auto-allocated the default value for this enum if the user doesn't include it in their request body.  By default it'll choose the first GeographicLevel value in the enum otherwise.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/BatchMappingUpdateViewModels.cs
duncan-at-hiveit: I was forced to do this so that it is not auto-allocated the default value for this enum if the user doesn't include it in their request body.  By default it'll choose the first MappingType value in the enum otherwise.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/BatchMappingUpdateViewModels.cs
duncan-at-hiveit: Because I was forced to change the enums to be nullable so that they don't get auto-set, I did the same with these.  But actually there's no technical reason why they need to be, so I've set them back to being non-nullable to avoid future confusion!

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionMappingControllerTests.cs
duncan-at-hiveit: As per previous comment and change, I'm planning to keep this separate if that's OK

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionMappingControllerTests.cs
duncan-at-hiveit: Nope!  I removed it in a subsequent PR, but I'll remove it here too :)

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Fixtures/TestApplicationFactory.cs
duncan-at-hiveit: I'd like to, as currently I'm finding it very useful to just be able to include this again when I need it.  I did try to control with configuration, but it was getting messy!

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Public.Data/DataSetVersionMappingController.cs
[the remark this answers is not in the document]
jack-hive: Yeah that sounds like a good solution to me! :)

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/BatchMappingUpdateViewModels.cs
```diff
@@ -0,0 +1,80 @@
...
+
+    public string SourceKey { get; init; } = string.Empty;
+
+    public LocationOptionMapping Mapping { get; init; } = null!;
+}
+
+public record BatchLocationMappingUpdatesResponseViewModel
+{
+    public List<LocationMappingUpdateResponse> Updates { get; init; } = [];
+}
```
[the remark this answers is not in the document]
jack-hive: I think (correct me if I'm wrong!) your reasoning is:
- You'd put `LocationMappingUpdateResponse` above `BatchLocationMappingUpdatesResponseViewModel` because it feels like it should be defined first in order to then use it in `BatchLocationMappingUpdatesResponseViewModel`

Whereas, my reasoning would be:
- I'd put `BatchLocationMappingUpdatesResponseViewModel` above `LocationMappingUpdateResponse` because `BatchLocationMappingUpdatesResponseViewModel` is the main view model that is returned from the endpoint - so if you read it like a book from top to bottom it reads nicer (in my opinion)

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/BatchMappingUpdateViewModels.cs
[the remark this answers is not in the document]
jack-hive: Ah, to be honest you make a good point about the user changing their minds and changing it back! I'm a bit torn now.

But I must agree, at first I was expecting only `Manual*` to be accepted. So I'll approve as-is and we can always change it later if needs be :)

src/GovUk.Education.ExploreEducationStatistics.Common/Services/PostgreSqlRepository.cs
jack-hive: I thought we had an `.OnSuccess` extension for use with non-task `Eithers`? So I thought you might have been ok to leave the `updateValueFn` type as if was

src/GovUk.Education.ExploreEducationStatistics.Common/Services/PostgreSqlRepository.cs
jack-hive: sounds good!

src/GovUk.Education.ExploreEducationStatistics.Common/Services/PostgreSqlRepository.cs
jack-hive: sounds good!

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/BatchMappingUpdateViewModels.cs
jack-hive: ah, I see. That's a good point!

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/BatchMappingUpdateViewModels.cs
jack-hive: makes sense!

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/Public.Data/BatchMappingUpdateRequests.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            RuleFor(request => request.CandidateKey)
+                .NotEmpty()
+                .When(request => request.Type is not null && request.Type.Value == MappingType.ManualMapped)
+                .WithErrorCode(ValidationMessages.CandidateKeyMustBeSpecifiedWithMappedMappingType.Code)
+                .WithMessage(ValidationMessages.CandidateKeyMustBeSpecifiedWithMappedMappingType.Message);
+            
+            RuleFor(request => request.CandidateKey)
+                .Null()
+                .When(request => request.Type is not null && request.Type.Value != MappingType.ManualMapped)
```
jack-hive, line 67: wonder if this should be `.When(request => request.Type is not null && request.Type.Value == MappingType.ManualNone)` to match the error code more accurately

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/Public.Data/BatchMappingUpdateRequests.cs
jack-hive: Just seen that you've fixed this in (EES-5115)[https://github.com/dfe-analytical-services/explore-education-statistics/pull/5028] so I'm happy for you to leave this as-is here as it will get fixed when 5115 gets merged :)

jack-hive (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/BatchMappingUpdateViewModels.cs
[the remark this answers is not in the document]
duncan-at-hiveit: Yup totally - I tend to try to do the same with my methods as well :)

PR #5028, opened 2024-07-03
EES-5115 - Filter Option mapping endpoint

duncan-at-hiveit:
This PR:
- adds an endpoint for updating the mappings for Filter Options via a debounced autosave mechansim.

Much the same as the PR for autosaving Location mappings!

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionMappingControllerTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionMappingControllerTests.cs
@@ -4,8 +4,8 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
 using System.Security.Claims;
-using System.Text.Json;
 using System.Threading.Tasks;
+using GovUk.Education.ExploreEducationStatistics.Admin.Requests.Public.Data;
 using GovUk.Education.ExploreEducationStatistics.Admin.Tests.Fixture;
 using GovUk.Education.ExploreEducationStatistics.Admin.Validators;
 using GovUk.Education.ExploreEducationStatistics.Admin.ViewModels.Public.Data;
@@ -706,7 +706,7 @@ public abstract class DataSetVersionMappingControllerTests(
         {
             var client = BuildApp().CreateClient();
 
-            var response = await ApplyBatchMappingUpdates(
+            var response = await ApplyBatchLocationMappingUpdates(
                 Guid.NewGuid(),
                 [
                     new LocationMappingUpdateRequest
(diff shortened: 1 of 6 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/BatchMappingUpdateViewModels.cs
jack-hive, line 1: I think you can remove the `.NotNull()` on the `FilterOptionMappingUpdateRequest.Validator` as the `.NotEmpty()` covers it

**I also think the changes in 5114 were intended to be here**

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionMappingControllerTests.cs
```diff
@@ -876,6 +873,469 @@ public async Task DataSetVersionMappingDoesNotExist_Returns404()
...
+            var client = BuildApp().CreateClient();
+        
+            List<FilterOptionMappingUpdateRequest> updates =
+            [
+                new()
+                {
+                    FilterKey = "Filter 1 key",
+                    SourceKey = "Filter 1 option 1 key",
+                    Type = MappingType.ManualMapped,
+                    CandidateKey = "target-filter-option-1-key"
```
jack-hive, line 983: I've just realised, we don't have anything in the code which validates that the candidate key selected is a legitimate candidate key. Should we validate it against the list of candidate keys in the mapping plan?

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/BatchMappingUpdateViewModels.cs
jack-hive: I think you can get rid of the nullable types `string?` and `MappingType?` on the `Type` and `SourceKey`, respectively, in `MappingUpdateRequest`

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/Public.Data/BatchMappingUpdateViewModels.cs
duncan-at-hiveit: I've removed the nullable string types but left the enum types as nullable, otherwise they're auto-bound to the first value within their enum if left empty.

jack-hive: makes sense!

jack-hive:
I think this might need rebasing back onto 5111 again, as it looks like a commit from EES-5111 has snuck in?

jack-hive (approved):

PR #5031, opened 2024-07-04
EES-5011 Replace obsolete queue client library `Microsoft.Azure.Storage.Queue`

benoutram:
This PR replaces usage of the obsolete queue client library `Microsoft.Azure.Storage.Queue` with `Azure.Storage.Queues`.

It replaces `IStorageQueueService` / `StorageQueueService` with a new `IQueueServiceClient` / `QueueServiceClient` which is a lightweight wrapper around the Azure `QueueServiceClient` providing the ability to send messages to any queue using the same connection and also serialize objects as JSON.

### New function clients

Unlike the previous usage of `StorageQueueService` where there could be multiple instances of `IStorageQueueService` configured with different connections leading to the potential for connection mismatch during dependency injection, there are now three new function clients `DataProcessorClient`, `PublisherClient` and `NotifierClient`. This simplifies the dependency injection configuration and removes the possibility of another service being wired up with the wrong candidate / connection.

These new clients give ownership to the Function apps for the types of actions which can be queued. They also have responsibility for queuing messages onto the correct queues removing potential for any mismatch between message types and the intended queues.

### Change to use `System.Text.Json` for serialization

This PR depends on #5025 due to changing from `Newtonsoft.Json` to `System.Text.Json` for serializing message objects onto queues. This is done to be consistent with a recent change to Azure Functions which uses `System.Text.Json` by default for deserializing queue messages.

This means the workaround for `System.Text.Json` deserializing messages which was added to `PublisherHostBuilderExtensions.cs` is also removed.

### Other changes

- Renames `PoisonQueueHandler.cs` to `PoisonQueueHandlerFunction.cs`.


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublishingService.cs
```diff
@@ -1,26 +1,25 @@
...
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Validators.ValidationErrorMessages;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Validators.ValidationUtils;
-using static GovUk.Education.ExploreEducationStatistics.Publisher.Model.PublisherQueues;
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
 {
     public class PublishingService(
-        IPersistenceHelper<ContentDbContext> persistenceHelper,
```
mmyoungman, line 22: Out of interest, do we still need the PersistenceHelper at all now or can it be removed?

To be honest, I'm not sure the history of it and why it was added in the first place. I assume to limit the number of db connections?

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
```diff
@@ -397,15 +396,7 @@ public virtual void ConfigureServices(IServiceCollection services)
...
-
-            services.AddTransient<IPublishingService, PublishingService>(provider =>
-                new PublishingService(
-                    provider.GetService<IPersistenceHelper<ContentDbContext>>(),
-                    new StorageQueueService(
-                        publisherStorageConnectionString,
-                        new StorageInstanceCreationUtil()),
-                    provider.GetService<IUserService>(),
-                    provider.GetRequiredService<ILogger<PublishingService>>()));
+            services.AddTransient<IPublishingService, PublishingService>();
```
mmyoungman, line 399: I thought we might want to use `AddSingleton` here? And with the others that use `QueueServiceClient`. 

I could be completely off here though...

src/GovUk.Education.ExploreEducationStatistics.Publisher/Functions/PublishTaxonomyFunction.cs
```diff
@@ -13,7 +13,9 @@ public class PublishTaxonomyFunction(
 {
     [Function("PublishTaxonomy")]
     public async Task PublishTaxonomy(
+#pragma warning disable IDE0060
         [QueueTrigger(PublishTaxonomyQueue)] PublishTaxonomyMessage message,
+#pragma warning restore IDE0060
```
mmyoungman, line 18: I see the `message` is unused. Is there a reason we cannot remove it?

If it can and you're avoiding it, don't feel the need to remove it on my behalf.

mmyoungman (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublishingService.cs
benoutram: I think it was only ever intended to be a helper to provide `CheckEntityExists` but I think that can be replaced in every case now with the methods such as `FirstOrNotFoundAsync` now. See https://github.com/dfe-analytical-services/explore-education-statistics/blob/290cb4b7b2548641bde87b76031b79042899b813/src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/QueryableExtensions.cs#L13 now.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Functions/PublishTaxonomyFunction.cs
benoutram: You need to have some parameter to add the `QueueTrigger` attribute to.

You also can't use `PublishTaxonomyMessage _` otherwise the following startup error occurs:

`The 'PublishTaxonomy' function is in error: The binding name _ is invalid. Please assign a valid name to the binding.`

src/GovUk.Education.ExploreEducationStatistics.Admin/Startup.cs
benoutram: Added a new commit where I've changed the scope of our new 'clients' to Singleton. They wrap `QueueServiceClient` which in turn wraps `Azure.Storage.Queues.QueueServiceClient`.

The reason why I didn't set the scope to Singleton to begin with was that I wasn't sure if `QueueServiceClient` is thread-safe, and not using Singleton was just the safest option.

Looks like `Azure.Storage.Queues.QueueServiceClient` is thread-safe and we should be ok to make this change. Thanks.

mmyoungman (approved):

PR #5034, opened 2024-07-08
EES-5112 - adding validation to ensure Candidate Keys are valid

duncan-at-hiveit:
This PR:
- adds in validation for the selected candidate keys for update requests, for both Locations and Filter Options.

## Location candidate validation

This validation is pretty simple.  We only need to take the "Level" supplied in the update request, and see if there is a candidate Location within that same Level with the given CandidateKey.

## Filter option validation

This validation is a little more involved.  In order to validate, we:

1. Take the "FilterKey" from the update request, which identifies the owning filter of the filter option being updated.
2. Find the filter that that owning filter is mapped to.
3. If the owning filter has not yet been mapped, return an error.
4. If the owning filter has been mapped, look for the candidate filter option underneath the filter that the owning filter has been mapped to.

## New "KeyExistsAtJsonbPath" method

This is an optimised method that in this PR we are using to check for the existence of a key within a given chunk of JSON.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionMappingControllerTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionMappingControllerTests.cs
@@ -66,8 +66,7 @@ public abstract class DataSetVersionMappingControllerTests(
                     Levels = new Dictionary<GeographicLevel, LocationLevelMappings>
                     {
                         {
-                            GeographicLevel.LocalAuthority,
-                            new LocationLevelMappings
+                            GeographicLevel.LocalAuthority, new LocationLevelMappings
                             {
                                 Mappings = new Dictionary<string, LocationOptionMapping>
                                 {
@@ -253,8 +252,7 @@ public abstract class DataSetVersionMappingControllerTests(
                     Levels = new Dictionary<GeographicLevel, LocationLevelMappings>
                     {
                         {
-                            GeographicLevel.LocalAuthority,
-                            new LocationLevelMappings
+                            GeographicLevel.LocalAuthority, new LocationLevelMappings
                             {
                                 Mappings = new Dictionary<string, LocationOptionMapping>
                                 {
@@ -546,8 +544,7 @@ public abstract class DataSetVersionMappingControllerTests(
                     Levels = new Dictionary<GeographicLevel, LocationLevelMappings>
                     {
                         {
-                            GeographicLevel.LocalAuthority,
-                            new LocationLevelMappings
+                            GeographicLevel.LocalAuthority, new LocationLevelMappings
                             {
                                 Mappings = new Dictionary<string, LocationOptionMapping>
                                 {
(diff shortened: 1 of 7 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionMappingControllerTests.cs
```diff
@@ -1193,17 +1455,36 @@ public async Task UpdateMappingDoesNotExist_Returns400_AndRollsBackTransaction()
...
+                            {
+                                Options =
+                                {
+                                    {
+                                        "Target filter 1 option 1 key", new MappableFilterOption("Filter 1 option 1 label")
+                                    },
+                                    {
+                                        "Target filter 1 option 2 key", new MappableFilterOption("Filter 1 option 2 label")
+                                    },
+                                    { "Target filter 1 option 3 key", new MappableFilterOption("Filter 1 option 3 label") }
```
jack-hive, line 1473: might be nice to format this the same as the ones above :)

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionMappingControllerTests.cs
```diff
@@ -578,15 +661,176 @@ public async Task UpdateMappingDoesNotExist_Returns400_AndRollsBackTransaction()
...
+                    Type = MappingType.ManualMapped,
+                    CandidateKey = "target-la-location-2-key"
+                },
+                // This candidate does not exist.
+                new()
+                {
+                    Level = GeographicLevel.Country,
+                    SourceKey = "source-la-location-3-key",
+                    Type = MappingType.ManualMapped,
+                    CandidateKey = "target-la-location-1-key"
```
jack-hive, line 812: I'm not super fussed, but at first I didn't realise that the reason this candidate is invalid is because it is not a candidate for the `Country` level, even though it is a candidate for the `LocalAuthority` level.

This is a different failure scenario to the candidate key `target-la-location-2-key` above, which doesn't exist at all.

I just wondered if we should break these out into 2 separate tests to make it clearer - and because they could potentially go down different code paths? But I don't feel strongly either way

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Common/Services/PostgreSqlRepository.cs
```diff
@@ -14,6 +14,38 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Services;
...
+        var rowIdParam = new NpgsqlParameter("rowId", request.RowId);
+
+        // We use the COALESCE function below to ensure that we're testing the existence of
+        // the given key against a non-null JSON fragment. Without this, an exception is thrown
+        // whereas we are wanting to return false. 
+        var response = await context
+            .Set<JsonBool>()
+            .FromSqlRaw(
+                sql: $"""
+                      SELECT COALESCE("{request.JsonbColumnName}" #> @jsonPath, '[]'::jsonb)
```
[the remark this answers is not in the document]
duncan-at-hiveit: Yup, this means that if `{request.JsonbColumnName}" #> @jsonPath` resolves to NULL, then this will fall back to a non-null JSON object for the key to be compared against.

Without this fallback, it goes kaboom!

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionMappingControllerTests.cs
duncan-at-hiveit: I've given this file a blanket reformat, which cleans this one up too!

src/GovUk.Education.ExploreEducationStatistics.Common/Services/PostgreSqlRepository.cs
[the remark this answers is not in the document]
jack-hive: but doesn't `COALESCE` return the first non-null value in the parameter list? And if none are non-null, then it returns null itself? So I thought it would have the same behaviour?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionMappingControllerTests.cs
jack-hive: Nice! Yep, I prefer this compromise actually :)

jack-hive (approved):

PR #5035, opened 2024-07-08
EES-5213 change form validation mode and fix publication form issues

amyb-hiveit:
Following feedback in the DAC report about validation on the publication form in the table tool this PR makes the following changes to all forms:
- changes validation to be triggered when the form is submitted instead of `onBlur` (when a form field loses focus). This complies better with the GDS guidance as well as the DAC feedback.
- once a form has been submitted revalidation happens `onChange` so users get instant feedback that they've resolved the problem.
- set `noValidate` on forms by default so that the build in html form validation isn't triggered as this can cause accessibility issues.

There are some exceptions to this where we have very complex validation for the chart builder that would require a lot of reworking to get to work with only validating on submit, here I've set the form to validate `onBlur` (`ChartDataGroupingForm` and `ChartLegendConfiguration`) or set it on some individual fields (`ChartReferenceLinesConfiguration`)

I've updated all the tests which were checking for error messages appearing `onBlur`, while I was there I've changed these to use the new `user` from the common `render` instead of `userEvent`, so it's a bit noisey, sorry.

Also:
- fixed a validation problem on the publication form where you could end up with the error 'themeId cannot be null'
- fixed the `redirects_www` robot test now the data catalogue page has switched over

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/users/__tests__/UserInvitePage.test.tsx
+++ b/src/explore-education-statistics-admin/src/pages/users/__tests__/UserInvitePage.test.tsx
@@ -13,7 +13,6 @@ import { screen, waitFor, within } from '@testing-library/react';
 import React from 'react';
 import { MemoryRouter, Route } from 'react-router';
 import { administrationUserInviteRoute } from '@admin/routes/administrationRoutes';
-import userEvent from '@testing-library/user-event';
 
 jest.mock('@admin/services/publicationService');
 jest.mock('@admin/services/userService');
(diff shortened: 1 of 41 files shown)


src/explore-education-statistics-admin/src/pages/users/__tests__/UserInvitePage.test.tsx
```diff
@@ -447,9 +476,5 @@ describe('UserInvitePage', () => {
         </TestConfigContextProvider>
       </MemoryRouter>,
     );
-
-    expect(
-      await screen.findByText('Manage access to this service'),
-    ).toBeInTheDocument();
```
mmyoungman, line 453: You get stung by missing that this was here in one of the render methods doing this?

I agree with moving it out - even if it's more lines I wouldn't expect the render method to be doing something like this.

mmyoungman (approved):

PR #5041, opened 2024-07-10
EES-5337: Increase dev/prod statistics DB size to 600GB/1.5TB respectively

tomjonesdev:
Preventative measure to accommodate growing statistics database to ensure we don't encounter issues inserting/updating data.


duncan-at-hiveit (approved):
Sanity checked the new numbers, and I came out with the same figures!

PR #5043, opened 2024-07-10
EES-4928 API Subscriptions: Add Endpoints to subscribe users

jack-hive:
## Overview

This PR:
- Adds the ability for a user to subscribe to notifications for an API Data Set
- We have added two Http triggers in this PR:
  - One for requesting a subscription (of which the status will be set to `SubscriptionPending`)
  - One for verifying the subscription (which will be triggered via the user following a link we send to their email)

## Other miscellaneous changes

Whilst making these changes, we have made some slight refactors with the current functionality used to subscribe a user to a publication.

Namely:

### The `INotificationClientProvider.cs` is no longer required

This class was being used to retrieve an instance of a `NotificationClient` (which belongs to an external NuGet package `GovukNotify`). This was then passed into every method call to `IEmailService.SendEmail` when sending emails. We therefore had to inject `INotificationClientProvider` into every class we needed to send emails, and we had a pattern of constructing a new `NotificationClient` everywhere we needed to send an email. This also caused unnecessary bloating in our test code as we had to mock not only the `IEmailService`, but also the `INotificationClientProvider`, and also a mock of the `NotificationClient` itself - which we were creating via the `NotifierTestUtils.cs`.

The `INotificationClientProvider` has now been removed, and instead we inject a **scoped** instance of an `INotificationClient` at startup. We don't currently have a need for the configuration to ever change, and the only service which requires the use of the `INotificationClient` is the `IEmailService`. `IEmailService` is the only interface we currently need to use as a proxy to the external `GovukNotify` project.

This simplified the code, and greatly simplified the tests. We no longer have to mock 3 things, only the 1; `IEmailService`. 

The reason we still mock `IEmailService` is because we do not want to be running our tests against the actual external dependency of the `GovukNotify` project.

**EDIT: on reflection, we could have potentially span this a different way and mocked the `INotificationClient` instead of the `IEmailService`. Currently, it doesn't have too many consequences on the tests which way we go. However if the `EmailService` ever got more complex (rather than doing what it does now and just directly calls `NotificationClient.SendEmail` with no extra business logic) then it would be worth mocking the `INotificationClient` instead. This was an oversight on my part. I will wait until someone has reviewed and given their opinion before I make this change.**

### Integration test setup

The project was not previously fully set-up to run integration tests. We have:
- Moved the program configuration into `NotifierHostBuilder` such that we can access it via the integration test project startup in `NotifierFunctionsIntegrationTestFixture`.
- Created an `appsettings.IntegrationTest.json` which is used to override configuration settings in `NotifierFunctionsIntegrationTestFixture.ConfigureTestHostBuilder` to be specific to the integration test environment.
- Added `InitializeAsync` and `DisposeAsync` methods to `NotifierFunctionsIntegrationTest`. At the end of each individual test run we want to clear any data associated with the test. We have further added a reusable `ClearAzureDataTableTestData` method to `FunctionsIntegrationTest`, which is used to clear all rows of data for all tables within Azure Table Storage - this uses the new `Azure.Data.Tables` package, as opposed to the deprecated `Microsoft.Azure.Cosmos.Table` package.
- Due to not foreseeing a situation whereby we would like to **unit/integration** test against the real `NotificationClient`, we have created a permanently mocked `IEmailService _emailService`  in `NotifierFunctionsIntegrationTestFixture`, which is injected at startup to replace the `EmailService` implementation. This mock is then re-used across all tests within the same fixture scope. We automatically verify all of the mocked invocations, and reset the mock, after each test run inside the `NotifierFunctionsIntegrationTest.DisposeAsync` method.
- The tests written in `ApiSubscriptionManagerTests.cs` resemble much closer to integration tests than those in `PublicationSubscriptionManagerTests.cs`. In the future I envisage we will migrate those tests over to the new style, and won't be mocking all of the dependencies, now that the infrastructure is in place


src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/NotifierFunctionsIntegrationTest.cs
```diff
@@ -1,17 +1,98 @@
...
+
+    public async Task AddTestSubscription(string tableName, SubscriptionEntity subscription)
+    {
+        var storageAccount = CloudStorageAccount.Parse(TableStorageConnectionString());
+        var tableClient = storageAccount.CreateCloudTableClient();
+        var table = tableClient.GetTableReference(tableName);
+        await table.CreateIfNotExistsAsync();
+
+        await table.ExecuteAsync(TableOperation.InsertOrReplace(subscription));
+    }
```
benoutram, line 61: Just wondering if using the `DataTableStorageService` to add this would work here, or is the interface on `SubscriptionEntity` different and preventing this?

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/AssertExtensions.cs
```diff
@@ -107,5 +107,29 @@ public static void AssertUtcNow(this DateTimeOffset? dateTimeOffset, int withinM
...
+        /// <param name="expectedDateTimeOffset">the expected timestamp.</param>
+        public static void AssertEqual(
+            this DateTimeOffset? dateTimeOffset, 
+            DateTimeOffset expectedDateTimeOffset, 
+            int withinMillis = TimeWithinMillis)
+        {
+            Assert.NotNull(dateTimeOffset);
+            dateTimeOffset.Value.AssertEqual(expectedDateTimeOffset: expectedDateTimeOffset, withinMillis: withinMillis);
+        }
     }
```
benoutram, line 134: Cool, was just needing something like this myself so it will be nice to switchover. 🙂 

I imagine there's quite a lot of places we could switch to use this, on another day!

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/NotifierFunctionsIntegrationTest.cs
```diff
@@ -1,17 +1,98 @@
...
+        var appSettingsOptions = GetRequiredService<IOptions<AppSettingsOptions>>();
+
+        var dataTableStorageService = new DataTableStorageService(appSettingsOptions.Value.TableStorageConnectionString);
+
+        return await dataTableStorageService.GetEntityIfExistsAsync<ApiSubscription>(
+            tableName: appSettingsOptions.Value.ApiSubscriptionsTableName,
+            partitionKey: dataSetId.ToString(),
+            rowKey: email,
+            select: select);
+    }
```
benoutram: Could these methods be moved to the test classes that they are used by? E.g. `AddTestSubscription` is only used by `PublicationSubscriptionManagerTests`, but currently you have to navigate to this class to find out what the implementation is.

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/PublicationSubscriptionManagerTests.cs
```diff
@@ -38,37 +38,29 @@ public class SubscriptionManagerFunctionTests(NotifierFunctionsIntegrationTestFi
     public async Task SendsSubscriptionVerificationEmail()
     {
         // Arrange (mocks)
-        var storageTableService = new StorageTableService(Options.Create(new AppSettingsOptions
+        var publicationSubscriptionRepository = new PublicationSubscriptionRepository(Options.Create(new AppSettingsOptions
```
benoutram, line 41: How come you can't use `GetRequiredService<IPublicationSubscriptionRepository>()` here? I would have thought it would be set up with the integration test app settings where the connection string has been overriden?

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/ApiSubscriptionManagerTests.cs
```diff
@@ -0,0 +1,375 @@
...
+using Xunit;
+
+namespace GovUk.Education.ExploreEducationStatistics.Notifier.Tests.Functions;
+
+public abstract class ApiSubscriptionManagerTests(NotifierFunctionsIntegrationTestFixture fixture)
+    : NotifierFunctionsIntegrationTest(fixture)
+{
+    private readonly Guid dataSetId = Guid.NewGuid();
+    private readonly string dataSetTitle = "data set title";
+    private readonly string email = "test@test.com";
```
benoutram, line 27: I think the names of these should be `_dataSetId`, `_dataSetTitle`, `_email`, which would help some with some warnings throughout this test class where these names are currently colliding with the names of method parameters.

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/ApiSubscriptionManagerTests.cs
```diff
@@ -0,0 +1,375 @@
...
+            Assert.Equal(email, response.Email);
+            Assert.Equal(ApiSubscriptionStatus.SubscriptionPending, response.Status);
+
+            // Assert that the verification link contains a valid token
+            var extractedEmail = ExtractEmailFromSubscriptionLinkToken(verificationLink);
+            Assert.Equal(extractedEmail, email);
+
+            var subscription = await GetApiSubscriptionIfExists(
+                dataSetId: dataSetId,
+                email: email);
```
benoutram, line 68: Cool, it's great to check actual table data afterwards! 🙂

src/GovUk.Education.ExploreEducationStatistics.Notifier/Repositories/PublicationSubscriptionRepository.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            return new Subscription
+            {
+                Subscriber = activeSubscriber,
+                Status = SubscriptionStatus.Subscribed
+            };
+        }
+
+        return new Subscription { Status = SubscriptionStatus.NotSubscribed };
+    }
```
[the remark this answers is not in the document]
jack-hive: I wouldn't want to risk breaking anything and having to re-test it all 😂

I was thinking we could come back to refactor the publication subscription stuff in 1 swoop later on?

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/NotifierFunctionsIntegrationTest.cs
jack-hive: I just left this bit as-is from the previous code - this method adds the **publication** subscriptions, and all of that code is still using the deprecated Cosmos library, including this method. So I thought best to leave all of the publication related stuff as untouched as possible for now, until we come to swap everything out to use the new stuff (if/when we decide that the new stuff is the way to go!)

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/AssertExtensions.cs
jack-hive: yeah, quite possibly!

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/NotifierFunctionsIntegrationTest.cs
jack-hive: Possibly, however I can see these having the potential to be used elsewhere. For example, when I write the tests for the function that will send notifications - which may be in a different class to the other triggers. I think we do a similar thing in the Processor test project which I think is kinda nice. Somewhat undiscoverable, but the reusability makes up for it in my opinion.

What do you think?

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/PublicationSubscriptionManagerTests.cs
jack-hive: Ah, these are the old publication subscription tests - I tried to leave them untouched for now. In these tests, we mock every service, and also mock the settings. So I left this as-is for now, but later it might be good to come back and change these to be in-line with the integration test style we've introduced in this PR for the API subscriptions, at the same time as refactoring all of that code to use the new `DataTableStorageService`?

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/FunctionsIntegrationTest.cs
```diff
@@ -50,6 +52,57 @@ public abstract class FunctionsIntegrationTest<TFunctionsIntegrationTestFixture>
...
+                var batch = items.Take(100);
+                items = items.Skip(100);
+
+                actions = [.. batch.Select(e => new TableTransactionAction(tableTransactionActionType, e))];
+                var response = await tableClient.SubmitTransactionAsync(actions).ConfigureAwait(false);
+                responses.Add(response);
+            }
+        }
+        return responses;
+    }
```
benoutram, line 104: Just discovered that there's also a method `.WithInMemoryPersistence` which we could try adding to the following code where we configure the Azurite container to enable the `--inMemoryPersistence option`.

https://github.com/dfe-analytical-services/explore-education-statistics/blob/65600c392af863af19bcfcee601ec60ff621acbb/src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/NotifierFunctionsIntegrationTest.cs#L137-L139

That might also help boost performance?

```
    private readonly AzuriteContainer _azuriteContainer = new AzuriteBuilder()
        .WithImage("mcr.microsoft.com/azure-storage/azurite:3.27.0")
        .WithInMemoryPersistence()
        .Build();
```

src/GovUk.Education.ExploreEducationStatistics.Common/Services/Interfaces/IDataTableStorageService.cs
```diff
@@ -0,0 +1,37 @@
...
+        string tableName,
+        ITableEntity entity,
+        TableUpdateMode updateMode = TableUpdateMode.Replace,
+        CancellationToken cancellationToken = default);
+
+    Task DeleteEntityAsync(
+        string tableName,
+        string partitionKey,
+        string rowKey,
+        CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
jack-hive: Done! :)

src/GovUk.Education.ExploreEducationStatistics.Notifier.Model/ApiSubscription.cs
```diff
@@ -0,0 +1,16 @@
...
+public class ApiSubscription : ITableEntity
+{
+    public string PartitionKey { get; set; }
+    public string RowKey { get; set; }
+    public string DataSetTitle { get; set; }
+    public ApiSubscriptionStatus Status { get; set; }
+    public DateTimeOffset? ExpiryDateTime { get; set; }
+    public DateTimeOffset? Timestamp { get; set; }
+    public ETag ETag { get; set; }
+}
```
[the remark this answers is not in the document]
jack-hive: Very good shout indeed! I've added these :)

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/appsettings.IntegrationTest.json
```diff
@@ -0,0 +1,22 @@
...
+    "ApiKey": "notifier-client-api-key",
+    "EmailTemplates": {
+      "ReleaseAmendmentPublishedId": "release-amendment-published-id",
+      "ReleaseAmendmentPublishedSupersededSubscribersId": "release-amendment-published-superseded-subscribers-id",
+      "ReleasePublishedId": "release-published-id",
+      "ReleasePublishedSupersededSubscribersId": "release-published-superseded-subscribers-id",
+      "SubscriptionConfirmationId": "subscription-confirmation-id",
+      "SubscriptionVerificationId": "subscription-verification-id",
+      "ApiSubscriptionConfirmationId": "api-subscription-confirmation-id",
+      "ApiSubscriptionVerificationId": "api-subscription-verification-id"
```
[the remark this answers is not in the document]
jack-hive: well spotted!

src/GovUk.Education.ExploreEducationStatistics.Notifier/Configuration/AppSettingsOptions.cs
```diff
@@ -11,4 +11,6 @@ public class AppSettingsOptions
     public string TableStorageConnectionString { get; init; } = null!;
 
     public string TokenSecretKey { get; init; } = null!;
+
+    public string ApiSubscriptionsTableName { get; init; } = null!;
```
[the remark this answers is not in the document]
jack-hive: I was planning on adding these to the infrastructure in this ticket https://dfedigital.atlassian.net/jira/software/c/projects/EES/boards/105?label=publicAPI&selectedIssue=EES-4927

Do you think we're OK to leave all of the infrastructure changes until then? I don't think these endpoints will be being used until that point so it shouldn't blow up I don't think?

src/GovUk.Education.ExploreEducationStatistics.Notifier/Configuration/GovUkNotifyOptions.cs
```diff
@@ -21,5 +21,9 @@ public class EmailTemplateOptions
         public string SubscriptionConfirmationId { get; init; } = null!;
 
         public string SubscriptionVerificationId { get; init; } = null!;
+
+        public string ApiSubscriptionConfirmationId { get; init; } = null!;
+
+        public string ApiSubscriptionVerificationId { get; init; } = null!;
```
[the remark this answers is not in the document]
jack-hive: I was planning on adding these to the infrastructure in this ticket https://dfedigital.atlassian.net/jira/software/c/projects/EES/boards/105?label=publicAPI&selectedIssue=EES-4927

Do you think we're OK to leave all of the infrastructure changes until then? I don't think these endpoints will be being used until that point so it shouldn't blow up I don't think?

src/GovUk.Education.ExploreEducationStatistics.Notifier/ViewModels/ApiSubscriptionViewModel.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace GovUk.Education.ExploreEducationStatistics.Notifier.ViewModels;
+
+public record ApiSubscriptionViewModel
+{
+    public Guid DataSetId { get; init; }
+
+    public string DataSetTitle { get; init; } = null!;
+
+    public string Email { get; init; } = null!;
```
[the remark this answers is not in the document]
jack-hive: good shout!

src/GovUk.Education.ExploreEducationStatistics.Notifier/Requests/NewPendingApiSubscriptionRequest.cs
```diff
@@ -0,0 +1,25 @@
+using FluentValidation;
+using System;
+
+namespace GovUk.Education.ExploreEducationStatistics.Notifier.Requests;
+
+public record NewPendingApiSubscriptionRequest
```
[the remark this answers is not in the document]
jack-hive: changed :)

src/GovUk.Education.ExploreEducationStatistics.Notifier/Repositories/Interfaces/IApiSubscriptionTableStorageService.cs
```diff
@@ -0,0 +1,7 @@
+using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces;
+
+namespace GovUk.Education.ExploreEducationStatistics.Notifier.Repositories.Interfaces;
+
+public interface IApiSubscriptionTableStorageService : IDataTableStorageService
+{
+}
```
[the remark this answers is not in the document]
jack-hive: nice!

src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/ApiSubscriptionManager.cs
```diff
@@ -0,0 +1,68 @@
...
+using System;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+using FluentValidation;
+using GovUk.Education.ExploreEducationStatistics.Common.Model;
+using Microsoft.Extensions.Logging;
+using Microsoft.AspNetCore.Http;
+
+namespace GovUk.Education.ExploreEducationStatistics.Notifier.Functions;
+
+public class ApiSubscriptionManager(
```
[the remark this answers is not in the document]
jack-hive: Yeah I think that would be better, I'll change it!

src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/PublicationSubscriptionManager.cs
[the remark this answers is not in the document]
jack-hive: Ah that's git recognizing this as an add/delete rather than a rename!

I believe that was one change yes. Another was to remove the need for the `INotificationClientProvider` and constructing a `NotificationClient` for every email.

I think that might be it, but it might be worth diffing the old `SubscriptionManager.cs` class with this in an online compare tool to be sure?

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/ApiSubscriptionManagerTests.cs
jack-hive: good shout

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/FunctionsIntegrationTest.cs
jack-hive: Nice find!

I've added this little beauty in. I've had to upgrade our Azurite version to 3.31.0 in doing so (well, it needed to be on 3.28.0 or greater, but I took this opportunity to upgrade to the latest!). I've also fixed the other places in the code we use this TestContainer (some of them were actually inconsistent with our data-storage docker container image!)

src/GovUk.Education.ExploreEducationStatistics.Notifier/Configuration/AppSettingsOptions.cs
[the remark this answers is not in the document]
benoutram: I didn't realise you had that ticket already thanks.

src/GovUk.Education.ExploreEducationStatistics.Notifier/Repositories/PublicationSubscriptionRepository.cs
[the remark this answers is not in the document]
benoutram: That makes sense to me! 🙂

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/PublicationSubscriptionManagerTests.cs
benoutram: Thanks, that makes sense now. Yes I imagine we'd swap over these tests to use the integration test approach when we do the Publication subscription refactoring.

src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/PublicationSubscriptionManager.cs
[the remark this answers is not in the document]
benoutram: I think the changes to the tests match up with our expectations of what's changed  here so it's all cool. 🙂

src/GovUk.Education.ExploreEducationStatistics.Common/Services/DataTableStorageService.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// <summary>
+    /// Gets a table by name, will create the table if it does not exist
+    /// </summary>
+    /// <param name="tableName">The name of the table to get.</param>
+    /// <returns>The table</returns>
+    private async Task<TableClient> GetTableClient(string tableName)
+    {
+        var tableClient = _client.GetTableClient(tableName);
+
+        await tableClient.CreateIfNotExistsAsync();
```
benoutram, line 87: I think this means that the change described in the infrastructure ticket EES-4927 to create the table is not required in advance.

That ticket also describes creating a new queue, but we do a similar thing for queues

We do a similar `CreateIfNotExists` for getting a handle on storage queues, see

https://github.com/dfe-analytical-services/explore-education-statistics/blob/afa5dd143813b0400360e34830e8bc0022bcd21b/src/GovUk.Education.ExploreEducationStatistics.Common/Services/QueueServiceClient.cs#L42-L47

I think that means you don't need this ticket at all?

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/FunctionsIntegrationTest.cs
benoutram: Cool! Glad to know this is working👍

src/GovUk.Education.ExploreEducationStatistics.Notifier/Configuration/GovUkNotifyOptions.cs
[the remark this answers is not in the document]
benoutram: I didn't realise you had that ticket already thanks. I've left another comment related to that Jira ticket.

I think I would do this work early to keep the infrastructure template change with the code changes and have a faffy deploy ticket to role out the template id's to KeyVault.

It's fine if you want to make the infrastructure template change separately though.

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/NotifierFunctionsIntegrationTest.cs
benoutram: I think it's fine to leave these in then, for potential reusability like you say.

src/GovUk.Education.ExploreEducationStatistics.Common/Services/DataTableStorageService.cs
jack-hive: You're absolutely right! I think I'll repurpose that ticket to be a faffy-deploy ticket for the keyvault values we need

benoutram (approved):

benoutram (approved):

PR #5044, opened 2024-07-10
Create new `public_data` app user for local environment and new `public_data_admin` group role for granting privileges

benoutram:
This PR adds a new Postgres initialisation script which creates a new `public_data` application user to be used as the database user for our services running locally.

It also grants privileges to a new `public_data_admin` role. Membership of this role can be granted to individual user roles in Azure who require write privileges on public schema objects.

In the local environment the `public_data_admin` role is created by the initialisation script, although it's unlikely to be used since we can easily login as the `postgres` superuser or `public_data` application users.

In Azure environments this role needs to be created manually before the database migration is run which grants privileges.

### Testing the `public_data_admin` role

Create a new user  `test_user_1` without `public_data_admin` role membership:

```bash
psql -U postgres -d public_data <<-EOSQL
    CREATE ROLE test_user_1 WITH LOGIN PASSWORD 'password';
    GRANT USAGE ON SCHEMA public TO test_user_1;
EOSQL
```

2. Attempt to SELECT from a table in the public schema logged in as `test_user_1`:

```bash
psql -U test_user_1 -d public_data -c "SELECT COUNT(*) from public.\"DataSets\""
```

🔴 Prints

```
ERROR:  permission denied for table DataSets
```

Create a new user  `test_user_2` without `public_data_admin` role membership:

```bash
psql -U postgres -d public_data <<-EOSQL
    CREATE ROLE test_user_2 WITH LOGIN PASSWORD 'password';
    GRANT USAGE ON SCHEMA public TO test_user_2;
    GRANT public_data_admin TO test_user_2;
EOSQL
```

2. Attempt to SELECT from a table in the public schema logged in as `test_user_2`:

```bash
psql -U test_user_2 -d public_data -c "SELECT COUNT(*) from public.\"DataSets\""
```

🟢  Prints

```
 count
-------
     9
(1 row)
```


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240501113006_InitialMigration.cs
```diff
@@ -579,40 +580,67 @@ protected override void Up(MigrationBuilder migrationBuilder)
                 principalTable: "DataSetVersions",
                 principalColumn: "Id");
 
+            var isLocalEnvironment = 
+                Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT") == Environments.Development;
+
             // Grant privileges on objects created by this resource's database user to the Admin App Service user.
-            var adminAppServiceIdentityName = Environment.GetEnvironmentVariable("AdminAppServiceIdentityName");
-            if (adminAppServiceIdentityName != null)
+            var adminAppServiceRoleName = isLocalEnvironment
```
duncan-at-hiveit, line 587: I know it's just grown this way, but this pattern repeats a few times in this file and is quite verbose:

```
            var adminAppServiceRoleName = isLocalEnvironment
                ? "app_admin" 
                : Environment.GetEnvironmentVariable("AdminAppServiceIdentityName");
            if (adminAppServiceRoleName != null)
            {
                migrationBuilder.Sql(
                    $"""GRANT SELECT, INSERT, UPDATE, DELETE ON ALL TABLES IN SCHEMA public TO {adminAppServiceRoleName}""");
                migrationBuilder.Sql(
                    $"""ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT, INSERT, UPDATE, DELETE ON TABLES TO {adminAppServiceRoleName}""");
            }
```

I wonder if we can extract into a method that just takes the environment variable name and the PSQL role name?  Then we can very quickly reuse if and when we need more roles adding?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240501113006_InitialMigration.cs
```diff
@@ -579,40 +580,67 @@ protected override void Up(MigrationBuilder migrationBuilder)
...
+                ? "app_admin" 
+                : Environment.GetEnvironmentVariable("AdminAppServiceIdentityName");
+            if (adminAppServiceRoleName != null)
             {
                 migrationBuilder.Sql(
-                    $"GRANT SELECT, INSERT, UPDATE, DELETE ON ALL TABLES IN SCHEMA public TO \"{adminAppServiceIdentityName}\"");
+                    $"""GRANT SELECT, INSERT, UPDATE, DELETE ON ALL TABLES IN SCHEMA public TO {adminAppServiceRoleName}""");
```
duncan-at-hiveit, line 593: I wondered if you were going to go with an approach of defining roles up front with their associated permissions to do `SELECT, INSERT, UPDATE, DELETE`, and then assign the Azure users those roles.  Had you looked at that approach at all?

E.g. every environment always has an "app_admin" role with NOLOGIN and `SELECT, INSERT, UPDATE, DELETE` permissions, and then we assign the Azure Admin managed identity user that role, and locally we assign a db user that role.

Or alternatively we have a `public_api_read_write` role that is NOLOGIN with `SELECT, INSERT, UPDATE, DELETE` permissions, and then we just assign all our application users that role.

There's possibly not much advantage in either alternative approach to be fair though, if they even work.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240501113006_InitialMigration.cs
```diff
@@ -579,40 +580,67 @@ protected override void Up(MigrationBuilder migrationBuilder)
...
             {
                 migrationBuilder.Sql(
-                    $"GRANT SELECT, INSERT, UPDATE, DELETE ON ALL TABLES IN SCHEMA public TO \"{publisherFunctionAppIdentityName}\"");
+                    $"""GRANT SELECT, INSERT, UPDATE, DELETE ON ALL TABLES IN SCHEMA public TO {publisherFunctionAppRoleName}""");
                 migrationBuilder.Sql(
-                    $"ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT, INSERT, UPDATE, DELETE ON TABLES TO \"{publisherFunctionAppIdentityName}\"");
+                    $"""ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT, INSERT, UPDATE, DELETE ON TABLES TO {publisherFunctionAppRoleName}""");
             }
+
+            // Grant privileges on objects created by this resource's database user to the 'public_data_admin' role.
```
duncan-at-hiveit, line 627: For simplicity, would it be worth just having this role in local as well as Azure?  That way, if it needs tweaking in any way, it can be maintained and tested locally before changing in Azure.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240628135434_EES5235_GrantSequencePrivileges.cs
```diff
@@ -12,14 +13,19 @@ public partial class EES5235_GrantSequencePrivileges : Migration
     /// <inheritdoc />
     protected override void Up(MigrationBuilder migrationBuilder)
     {
+        var isLocalEnvironment = 
+            Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT") == Environments.Development;
```
duncan-at-hiveit, line 17: Well found - I struggled to work out how to determine if you're in development or not when away from the host builder code!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240710081112_EES5329_CreatePublicDataAdminRole.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Migrations;
+
+/// <inheritdoc />
+[ExcludeFromCodeCoverage]
+public partial class EES5329_CreatePublicDataAdminRole : Migration
+{
+    /// <inheritdoc />
+    protected override void Up(MigrationBuilder migrationBuilder)
+    {
+        // Grant privileges on objects created by this resource's database user to the 'public_data_admin' role.
```
duncan-at-hiveit, line 15: Is this a repeat of the code in InitialMigration?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240501113006_InitialMigration.cs
```diff
@@ -579,40 +580,67 @@ protected override void Up(MigrationBuilder migrationBuilder)
...
 
+            var isLocalEnvironment = 
+                Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT") == Environments.Development;
+
             // Grant privileges on objects created by this resource's database user to the Admin App Service user.
-            var adminAppServiceIdentityName = Environment.GetEnvironmentVariable("AdminAppServiceIdentityName");
-            if (adminAppServiceIdentityName != null)
+            var adminAppServiceRoleName = isLocalEnvironment
+                ? "app_admin" 
+                : Environment.GetEnvironmentVariable("AdminAppServiceIdentityName");
```
duncan-at-hiveit, line 589: I wonder, should we just fail initial migration if we're running in Azure and we haven't been supplied the appropriate environment variables for the various users?  Failing to do so will allow the migration to complete successfully but without all our required users set up, at which point we'd have to do manual cleanup.

data/public-api-db/00-init.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
duncan-at-hiveit, line 1: This is very cool - nicely found that you can do this.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/PublicDataDbContextModelSnapshot.cs
```diff
@@ -25,6 +25,26 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
             modelBuilder.HasSequence<int>("FilterOptionMetaLink_seq");
 
+            modelBuilder.Entity("GovUk.Education.ExploreEducationStatistics.Common.Model.JsonBool", b =>
```
duncan-at-hiveit, line 28: Oh, I wonder how these didn't get generated for my PR?  :)

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240710081112_EES5329_CreatePublicDataAdminRole.cs
benoutram: Yes it is, this one is the migration required to make this change in the Dev environment as it stands now, because the `InitialMigration` has already run there.

This changes was copied back to `IntiialMigration` so we don't forget it if we squash the migrations again and produce a new `IntiialMigration`.

If you clear the Postgres volume locally and allow all the migrations to run from scratch it does mean these grants get applied twice but I don't think that matters.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240628135434_EES5235_GrantSequencePrivileges.cs
benoutram: I've added this to `MigrationBuilderExtensions` now so that within a migration you can do `migrationBuilder.IsLocalEnvironment()`.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240501113006_InitialMigration.cs
benoutram: I did do this and put these methods in `MigrationBuilderExtensions` but then with my most recent changes I don't think it's warranted anymore as there's now no need for so many occurrences of this, and there will be even less when we next squash all the migrations.

Just in case we need it again, this is what I was working on for the role names:

```csharp
private record AppService(string LocalPsqlRoleName, string AppIdentityEnvironmentVariableName)
{
    public static AppService AdminAppService => new("app_admin", "AdminAppServiceIdentityName");
    public static AppService DataProcessorFunctionApp => new("app_public_data_processor", "DataProcessorFunctionAppIdentityName");
    public static AppService PublisherFunctionApp => new("app_publisher", "PublisherFunctionAppIdentityName");

    private static bool IsLocalEnvironment => Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT") == Environments.Development;

    public string PsqlRoleName => IsLocalEnvironment
        ? LocalPsqlRoleName
        : Environment.GetEnvironmentVariable(AppIdentityEnvironmentVariableName) ?? throw new Exception($"Environment variable {AppIdentityEnvironmentVariableName} is not set");
}
```

A role name could then be accessed like

```csharp
AppService.AdminAppService.PsqlRoleName
```

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240501113006_InitialMigration.cs
benoutram: Previous comment referred to before I replaced `public_data_admin` with `public_data_read_write`, but the same is still true.

If you want to test it locally,  create a user like so

```bash
 psql -U postgres -d public_data <<-EOSQL
    CREATE ROLE my_user WITH LOGIN PASSWORD 'password';
    GRANT USAGE ON SCHEMA public TO my_user;
    GRANT public_data_read_write TO my_user;
EOSQL
```

I did this after deleting the public data docker volume and the `public_data.sql` init script in the data directory, started the database and then allowed the Public API app to migrate the database.

Logging in with this new user we see it has write privileges, for example it can insert data:

```bash
psql -U my_user -d public_data <<-EOSQL
    INSERT INTO "DataSets" ("Id", "Title", "Summary", "PublicationId", "Status", "Created")
    VALUES (gen_random_uuid(),'Title','Summary','cbbd299f-8297-44bc-92ac-558bcf51f8ad','Published',now());
EOSQL
```

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240710081112_EES5329_CreatePublicDataAdminRole.cs
benoutram: This has been replaced now by another migration `EES5329_GrantPublicDataReadWriteRolePrivileges` , but yes it's repeating some code also added to `InitialMigration`. 

This migration is required to make the change in the Dev environment as it stands now, because `InitialMigration` has already run there, assuming that we're not immediately about to tear down that database and recreate it.

I copied the code back into `InitialMigration` so we don't forget it if we squash the migrations again and produce a new `InitialMigration`. I thought we would be quite likely to diff any new `InitialMigration` with the existing one and then spot this addition and copy it over.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240501113006_InitialMigration.cs
benoutram: See previous comment about this not being required any more, plus a possible solution for it it's required again.

I've also removed these variables from the bicep template.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240501113006_InitialMigration.cs
benoutram: Discussed this on Slack and explained that I'd initially tried something like this but couldn't get it to work unless the Public Data API's app user which creates and owns all the objects also did the grants.

I think I've got this working now but done this as a separate commit so we can revert back if necessary.

I went with this approach:

> Or alternatively we have a public_api_read_write role that is NOLOGIN with SELECT, INSERT, UPDATE, DELETE permissions, and then we just assign all our application users that role.

We would assign this role to all our application user roles and also our own Azure admin user roles that we want to have write privileges.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240628135434_EES5235_GrantSequencePrivileges.cs
benoutram: I've added this to `MigrationBuilderExtensions` now so that within a migration you can do something like `migrationBuilder.IsEnvironment(Environments.Development)` or `migrationBuilder.IsEnvironment(Environments.Production)`.

The reason why we check `migrationBuilder.IsEnvironment(Environments.Production)` rather than `!migrationBuilder.IsEnvironment(Environments.Development)`  is that we don't want to run `ALTER DEFAULT PRIVILEGES` in integration tests because that role doesn't exist. In the local environment it's been run by the Docker entrypoint init script.

I wasn't sure if we should be doing something in the integration tests before running the database migration to run the init script, or even going as far as to make sure each integration test migrates the database with the same user setup that it would do in a production and then runs the test as the same user that it would do in production, which might be a different user. That seems like it might be worthwhile but not an easy change to make. Using a pre-existing image that already has a migrated database seems like it would be better for performance anyway. I have left this for now.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240501113006_InitialMigration.cs
duncan-at-hiveit: Nice one!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240501113006_InitialMigration.cs
duncan-at-hiveit: Cool!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240628135434_EES5235_GrantSequencePrivileges.cs
duncan-at-hiveit: Makes sense, thanks!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240501113006_InitialMigration.cs
duncan-at-hiveit: Cool

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240710081112_EES5329_CreatePublicDataAdminRole.cs
duncan-at-hiveit: Makes sense, thanks!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240501113006_InitialMigration.cs
duncan-at-hiveit: :+1:

PR #5047, opened 2024-07-11
EES-4954 - continue import of next data set version, using the mappings

duncan-at-hiveit:
This PR:
- uses the auto mappings and manual mappings established previously to allow reuse of PublicIds from the previous data set version to use again for location options and filter options in the next data set version that have been mapped to.
- moves PublicId back onto LocationOptionMetaLink from LocationOptionMeta (which was its previous home a while back) in order to achieve the reuse of PublicIds without needing to create duplicate LocationOptionMeta rows with different Public Ids.

## Reusing PublicIds for backwards-compatibility with existing data set version queries

This PR allows existing queries built for, say, a `V1.0` data set version to be compatible with `V1.1`, by adopting PublicIds from elements in `V1.0` for their counterparts in `V1.1`.  Thus a user's query that was built using PublicId "abcd" to reference "Bath" in `V1.0` will be compatible with `V1.1`'s "Bath" also, by virtue of `V1.1`'s "Bath" adopting "abcd" as *its* PublicId too (so long as the mappings say that V1.0's "Bath" is mapped to V1.1's "Bath"!).

### Moving PublicId back onto LocationOptionMetaLink

In order to support this PublicId reuse strategy for Location Options, it was necessary to move PublicId back onto the LocationOptionMetaLink table, where it used to exist a while back.

The following scenario highlights why this is necessary, based on the previous model:

> So currently, users of the Public API will be using Public Ids to identify things like LocationOptions and FilterOptions in their data set queries.  So a given Data Set might contain information relating to a LocationOption called "City of Bath", which has a Public Id of "abcd".
>
> People putting together queries against that Data Set will use the Public Id of "abcd" if they're filtering results based on using the "City of Bath" LocationOption.
>
> If another Data Set comes along which also contains information related to a LocationOption of "City of Bath", we can reuse the existing LocationOption in the database (assuming all its Location codes match the one currently in the database), and thus we don't end up with an explosion of LocationOptions in the db.  As a result, people building queries against this other Data Set will also use the Public Id of "abcd" to reference it.
>
> All this works well so far!
>
> If another Data Set comes along and in its V1.0 version it contains a "Bath" LocationOption with a Public Id  of "efgh", people building queries against that Data Set will be using the Public Id of "efgh" to reference it.
If a user then creates a V1.1 data set version for that same Data Set that no longer contains a "Bath" LocationOption, but now contains a "City of Bath" LocationOption, the user will map "Bath" to "City of Bath" during the mapping phase.
When we come to complete the import of this next version, we will use the mappings the user carried out to identify that "Bath" has now become "City of Bath", and here is where it gets tricky with the current model.
>
> In order to allow queries that the user wrote against the V1.0 version to continue working correctly, we need to allow them to continue using the Public Id of "efgh" to reference the now "City of Bath" LocationOption.
> 
> This means we have to be able to either:
> 1. Support a single LocationOption of "City of Bath" with 2 or more Public Ids (in our case so far, "abcd" and "efgh".  In order to support this, I would suggest we move PublicId to the link tables that link data set versions to their meta, rather than on the meta tables themselves.  This would mean in this scenario moving PublicId from LocationOptionMeta to LocationOptionMetaLink.
> 2. Support multiple LocationOptions of "City of Bath", each with a different PublicId.  So in the scenario above, we'd have 2 "City of Bath" LocationOptionMeta entries in the db with the same Label, the same Location Codes, but different PublicIds ("abcd" and "efgh" respectively).

This PR opted to implement option 1 as the preferred option, thus reducing duplication of LocationOptionMeta entries in the database and preventing ambiguity when the importer next comes along to reuse which "City of Bath" to reuse for the next data set that uses it for a location option.

As it went, FilterOptionMetaLinks already had the PublicId on it so it wasn't necessary to make any changes to the filter model.

## Adding PublicIds to the mappings for optimised mapping application

During the applying of mappings when importing the next data set version's metadata, for each new piece of meta that's going into the tables we need to see if they should be adopting a predecessor's PublicId, or generating a new one because they haven't been mapped.

To do this, we generate the "mapping key" (or "candidate key") for that new piece of meta, and see if any mappings that were made in the previous step are referencing that mapping key via their "CandidateKey" field.  If they are, we need to find the original piece of meta's PublicId and adopt it.  This could have been achieved by reading the previous data set version's meta from the database and attempting to match pieces of old meta to their counterparts in the new version, and then finding the original one's Public Id from there, but I felt that it would be a lot more efficient just to store the PublicId in the mapping JSON itself so that it's immediately available in the code when it's needed, without any additional db lookups or object comparisons being necessary.

## Using GetRowKeyPretty() over GetRowKey() for generating mapping keys

Whilst showing the mapping JSON to Amy, I became aware that the mapping keys we were currently using for Location Options were quite mysterious-looking and prone to breaking in a tricky-to-debug way if for any reason someone were to choose to reorder the various codes being output in `GetRowKey()`.

I therefore created `GetRowKeyPretty()` to include the names of the various Code fields alongside their respective values, so that we can more easily see at a glance what all the various codes mean.  To improve space-saving and efficiency of looking up the key, only the non-null Code fields are included in the generated key.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/LocationOptionMetaRowTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/LocationOptionMetaRowTests.cs
@@ -11,25 +11,25 @@ public abstract class LocationOptionMetaRowTests
         {
             HashSet<string> excludedProperties =
             [
-                nameof(LocationOptionMetaRow.Id),
+                $"{nameof(LocationOptionMetaRow.Id)} value",
             ];
 
             var expectedProperties = typeof(LocationOptionMetaRow)
                 .GetProperties()
-                .Select(p => p.Name)
+                .Select(p => $"{p.Name} value")
                 .Except(excludedProperties)
                 .ToHashSet();
 
             var optionRow = new LocationOptionMetaRow
             {
                 Id = 1,
-                Type = nameof(LocationOptionMetaRow.Type),
-                Label = nameof(LocationOptionMetaRow.Label),
-                Code = nameof(LocationOptionMetaRow.Code),
-                OldCode = nameof(LocationOptionMetaRow.OldCode),
-                Urn = nameof(LocationOptionMetaRow.Urn),
-                LaEstab = nameof(LocationOptionMetaRow.LaEstab),
-                Ukprn = nameof(LocationOptionMetaRow.Ukprn),
+                Type = nameof(LocationOptionMetaRow.Type) + " value",
+                Label = nameof(LocationOptionMetaRow.Label) + " value",
+                Code = nameof(LocationOptionMetaRow.Code) + " value",
+                OldCode = nameof(LocationOptionMetaRow.OldCode) + " value",
+                Urn = nameof(LocationOptionMetaRow.Urn) + " value",
+                LaEstab = nameof(LocationOptionMetaRow.LaEstab) + " value",
+                Ukprn = nameof(LocationOptionMetaRow.Ukprn) + " value"
             };
 
             var rowKey = optionRow.GetRowKey();
@@ -37,6 +37,75 @@ public abstract class LocationOptionMetaRowTests
 
             Assert.Equal(expectedProperties, properties);
         }
+    }
+    
+    public class GetRowKeyPrettyTests
+    {
+        [Fact]
+        public void ContainsKeyProperties()
+        {
+            HashSet<string> excludedProperties =
+            [
+                $"{nameof(LocationOptionMetaRow.Id)}:{nameof(LocationOptionMetaRow.Id)} value",
+            ];
+
+            var expectedProperties = typeof(LocationOptionMetaRow)
+                .GetProperties()
+                .Select(p => $"{p.Name}:{p.Name} value")
+                .Except(excludedProperties)
+                .ToHashSet();
+
+            var optionRow = new LocationOptionMetaRow
+            {
+                Id = 1,
+                Type = nameof(LocationOptionMetaRow.Type) + " value",
+                Label = nameof(LocationOptionMetaRow.Label) + " value",
+                Code = nameof(LocationOptionMetaRow.Code) + " value",
+                OldCode = nameof(LocationOptionMetaRow.OldCode) + " value",
+                Urn = nameof(LocationOptionMetaRow.Urn) + " value",
+                LaEstab = nameof(LocationOptionMetaRow.LaEstab) + " value",
+                Ukprn = nameof(LocationOptionMetaRow.Ukprn) + " value"
+            };
+
+            var rowKey = optionRow.GetRowKeyPretty();
+            var properties = rowKey.Split(',').ToHashSet();
+
+            Assert.Equal(expectedProperties, properties);
+        }
+    }
+    
+    [Fact]
+    public void OmitsNullProperties()
+    {
+        HashSet<string> excludedProperties =
+        [
+            $"{nameof(LocationOptionMetaRow.Id)}:{nameof(LocationOptionMetaRow.Id)} value",
+            $"{nameof(LocationOptionMetaRow.OldCode)}:{nameof(LocationOptionMetaRow.OldCode)} value",
+            $"{nameof(LocationOptionMetaRow.Ukprn)}:{nameof(LocationOptionMetaRow.Ukprn)} value",
+        ];
+
+        var expectedProperties = typeof(LocationOptionMetaRow)
+            .GetProperties()
+            .Select(p => $"{p.Name}:{p.Name} value")
+            .Except(excludedProperties)
+            .ToHashSet();
+
+        var optionRow = new LocationOptionMetaRow
+        {
+            Id = 1,
+            Type = nameof(LocationOptionMetaRow.Type) + " value",
+            Label = nameof(LocationOptionMetaRow.Label) + " value",
+            Code = nameof(LocationOptionMetaRow.Code) + " value",
+            OldCode = null,
+            Urn = nameof(LocationOptionMetaRow.Urn) + " value",
+            LaEstab = nameof(LocationOptionMetaRow.LaEstab) + " value",
+            Ukprn = null
+        };
+
+        var rowKey = optionRow.GetRowKeyPretty();
+        var properties = rowKey.Split(',').ToHashSet();
+
+        Assert.Equal(expectedProperties, properties);
     }
 
     public class PropertyTests
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/LocationOptionMetaRowTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/LocationOptionMetaRowTests.cs
@@ -12,7 +12,6 @@ public abstract class LocationOptionMetaRowTests
             HashSet<string> excludedProperties =
             [
                 nameof(LocationOptionMetaRow.Id),
-                nameof(LocationOptionMetaRow.PublicId),
             ];
 
             var expectedProperties = typeof(LocationOptionMetaRow)
@@ -24,7 +23,6 @@ public abstract class LocationOptionMetaRowTests
             var optionRow = new LocationOptionMetaRow
             {
                 Id = 1,
-                PublicId = nameof(LocationOptionMetaRow.PublicId),
                 Type = nameof(LocationOptionMetaRow.Type),
                 Label = nameof(LocationOptionMetaRow.Label),
                 Code = nameof(LocationOptionMetaRow.Code),
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetVersionMapping.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetVersionMapping.cs
@@ -232,8 +232,11 @@ public record FilterMappingPlan
 
 public static class MappingKeyFunctions
 {
-    public static Func<LocationOptionMetaRow, string> LocationOptionKeyGenerator =>
-        option => $"{option.Label} :: {option.GetRowKey()}";
+    public static Func<LocationOptionMetaRow, string> LocationOptionMetaRowKeyGenerator =>
+        option => $"{option.Label} :: {option.GetRowKeyPretty()}";
+    
+    public static Func<LocationOptionMeta, string> LocationOptionMetaKeyGenerator =>
+        option => LocationOptionMetaRowKeyGenerator(option.ToRow());
 
     public static Func<FilterMeta, string> FilterKeyGenerator =>
         filter => filter.PublicId;
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetVersionMapping.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetVersionMapping.cs
@@ -31,6 +31,26 @@ public class DataSetVersionMapping : ICreatedUpdatedTimestamps<DateTimeOffset, D
     public DateTimeOffset Created { get; set; }
 
     public DateTimeOffset? Updated { get; set; }
+    
+    public LocationLevelMappings GetLocationLevelMappings(GeographicLevel level)
+    {
+        return LocationMappingPlan.Levels[level];
+    }
+    
+    public LocationOptionMapping GetLocationOptionMapping(GeographicLevel level, string locationOptionKey)
+    {
+        return GetLocationLevelMappings(level).Mappings[locationOptionKey];
+    }
+
+    public FilterMapping GetFilterMapping(string filterKey)
+    {
+        return FilterMappingPlan.Mappings[filterKey];
+    }
+    
+    public FilterOptionMapping GetFilterOptionMapping(string filterKey, string filterOptionKey)
+    {
+        return GetFilterMapping(filterKey).OptionMappings[filterOptionKey];
+    }
 
     internal class Config : IEntityTypeConfiguration<DataSetVersionMapping>
     {
@@ -103,10 +123,13 @@ public enum MappingType
 /// <summary>
 /// This base class represents an element from the DataSetVersions that can be mapped.
 /// </summary>
-public abstract record MappableElement(string Label);
+public abstract record MappableElement
+{
+    public string Label { get; init; }
+};
 
-public abstract record MappableElementWithOptions<TMappableOption>(string Label)
-    : MappableElement(Label)
+public abstract record MappableElementWithOptions<TMappableOption>
+    : MappableElement
     where TMappableOption : MappableElement
 {
     public Dictionary<string, TMappableOption> Options { get; init; } = [];
@@ -149,7 +172,7 @@ public abstract record ParentMapping<TMappableElement, TOption, TOptionMapping>
 /// This represents a location option that is potentially mappable to another location option
 /// from the same geographic level. 
 /// </summary>
-public record MappableLocationOption(string Label) : MappableElement(Label)
+public record MappableLocationOption : MappableElement
 {
     public string? Code { get; init; }
 
@@ -191,20 +214,19 @@ public class LocationMappingPlan
 /// <summary>
 /// This represents a filter option that is potentially mappable to another filter option. 
 /// </summary>
-public record MappableFilterOption(string Label) : MappableElement(Label);
+public record MappableFilterOption : MappableElement;
 
 /// <summary>
 /// This represents a filter that is potentially mappable to another filter.
 /// </summary>
-public record MappableFilter(string Label) : MappableElement(Label);
+public record MappableFilter : MappableElement;
 
 /// <summary>
 /// This represents a candidate filter and all of its candidate filter options from
 /// the target data set version that could be mapped to from filters and filter options
 /// from the source version.
 /// </summary>
-public record FilterMappingCandidate(string Label)
-    : MappableElementWithOptions<MappableFilterOption>(Label);
+public record FilterMappingCandidate : MappableElementWithOptions<MappableFilterOption>;
 
 /// <summary>
 /// This represents a potential mapping of a filter option from the source data set version
@@ -234,7 +256,7 @@ public static class MappingKeyFunctions
 {
     public static Func<LocationOptionMetaRow, string> LocationOptionMetaRowKeyGenerator =>
         option => $"{option.Label} :: {option.GetRowKeyPretty()}";
-    
+
     public static Func<LocationOptionMeta, string> LocationOptionMetaKeyGenerator =>
         option => LocationOptionMetaRowKeyGenerator(option.ToRow());
 
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetVersionMapping.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetVersionMapping.cs
@@ -109,7 +109,7 @@ public abstract record MappableElementWithOptions<TMappableOption>(string Label)
     : MappableElement(Label)
     where TMappableOption : MappableElement
 {
-    public Dictionary<string, TMappableOption> Options { get; set; } = [];
+    public Dictionary<string, TMappableOption> Options { get; init; } = [];
 }
 
 /// <summary>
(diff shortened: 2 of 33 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/LocationOptionMetaRowTests.cs
```diff
@@ -11,27 +11,25 @@ public void ContainsKeyProperties()
...
                 Id = 1,
-                PublicId = nameof(LocationOptionMetaRow.PublicId),
-                Type = nameof(LocationOptionMetaRow.Type),
-                Label = nameof(LocationOptionMetaRow.Label),
-                Code = nameof(LocationOptionMetaRow.Code),
-                OldCode = nameof(LocationOptionMetaRow.OldCode),
-                Urn = nameof(LocationOptionMetaRow.Urn),
-                LaEstab = nameof(LocationOptionMetaRow.LaEstab),
-                Ukprn = nameof(LocationOptionMetaRow.Ukprn),
+                Type = nameof(LocationOptionMetaRow.Type) + " value",
```
jack-hive, line 26: Are we adding this `" value"` on the end to check for spaces etc?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ImportMetadataFunctionTests.cs
```diff
@@ -156,6 +157,150 @@ public async Task DataSetVersionMeta_CorrectLocations(ProcessorTestData testData
...
+                                Mappings = level
+                                    .Options
+                                    .ToDictionary(
+                                        keySelector: MappingKeyFunctions.LocationOptionMetaKeyGenerator,
+                                        elementSelector: option =>
+                                            option == mappedOption1 ? option1Mapping
+                                            : option == mappedOption2 ? option2Mapping
+                                            : new LocationOptionMapping
+                                            {
+                                                Type = random.Next(1) == 0
```
jack-hive, line 230: I think `random.Next(1)` will always produce `0`

jack-hive: wonder if it would be better to make this deterministic, and not use `random`. Declare our own integer, and take the modulus so that it switches between `AutoNone` and `ManualNone` for consecutive options

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ImportMetadataFunctionTests.cs
```diff
@@ -241,6 +386,142 @@ public async Task DataSetVersionMeta_CorrectFilters(ProcessorTestData testData,
...
+            await using var publicDataDbContext = GetDbContext<PublicDataDbContext>();
+
+            var actualFilters = await publicDataDbContext.FilterMetas
+                .Include(fm => fm.Options.OrderBy(o => o.Label))
+                .ThenInclude(fom => fom.MetaLinks)
+                .Where(fm => fm.DataSetVersionId == targetDataSetVersion.Id)
+                .OrderBy(fm => fm.Label)
+                .ToListAsync();
+
+            var globalOptionIndex = 0;
```
jack-hive, line 471: Just wondering, I think I'd actually expect this test to fail... Shouldn't this start counting from the number of options that are already in the database from data set version 1.0 ? 

Or am I not thinking through this correctly?

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240710122937_EES4954_MovePublicIdToLocationMetaOptionLinkTable.cs
```diff
@@ -0,0 +1,69 @@
...
+        private const string MigrationId = "20240710122937";
+
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<string>(
+                name: "PublicId",
+                table: "LocationOptionMetaLinks",
+                type: "text",
+                nullable: true);
```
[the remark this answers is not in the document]
duncan-at-hiveit: It's nullable at first, and then we migrate the values from LocationOptionMeta into it.  Finally, we set it to non-nullable.  You can find all of these steps in this one migration.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/LocationOptionMetaRowTests.cs
duncan-at-hiveit: I'm just wanting to make sure it's not just using nameof(field) as the method-under-test's output, so adding something that definitely proves that it's using the values of those fields and not just the names of the fields.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ImportMetadataFunctionTests.cs
duncan-at-hiveit: Good idea - implemented this thanks

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ImportMetadataFunctionTests.cs
duncan-at-hiveit: It goes off the starting value of the sequence that is used to control this.  We don't actually set up any prior standing data, so this sequence starts at 0 at this point.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/FilterMetaRepository.cs
```diff
@@ -199,4 +218,64 @@ await optionTable
...
+                    .Values
+                    .Where(mapping => mapping.Type is MappingType.AutoMapped or MappingType.ManualMapped)
+                    .ToDictionary(
+                        keySelector: mapping => mapping.CandidateKey!,
+                        elementSelector: mapping => mapping.PublicId));
+
+        return new PublicIdMappings { Filters = mappingsByFilter };
+    }
+
+    private static string CreatePublicIdForFilterOptionMetaLink(
```
ntsim, line 254: We're not really creating anything so this might not be the best naming. Could also this to `GetFilterOptionMetaLinkPublicId`.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/FilterMetaRepository.cs
```diff
@@ -199,4 +218,64 @@ await optionTable
...
+        public Dictionary<string, Dictionary<string, string>> Filters { get; init; } = [];
+
+        public string? GetPublicIdForFilterOptionCandidate(string filterKey, string filterOptionCandidateKey)
+        {
+            if (!Filters.TryGetValue(filterKey, out var filterOptionMappings))
+            {
+                return null;
+            }
+
+            return filterOptionMappings.GetValueOrDefault(filterOptionCandidateKey);
```
ntsim, line 278: Could simplify this to:
```cs
return Filters
    .GetValueOrDefault(filterKey)
    ?.GetValueOrDefault(filterOptionCandidateKey);
```

ntsim, line 278: Could simplify this to:
```cs
return Filters.GetValueOrDefault(filterKey)
    ?.GetValueOrDefault(filterOptionCandidateKey);
```

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/LocationMetaRepository.cs
```diff
@@ -226,35 +239,74 @@ private static List<GeographicLevel> ListLocationLevels(IReadOnlySet<string> all
...
+
+    private async Task<PublicIdMappings> CreatePublicIdMappings(
+        DataSetVersion dataSetVersion,
+        CancellationToken cancellationToken)
+    {
+        var mappings = await EntityFrameworkQueryableExtensions
+            .SingleOrDefaultAsync(publicDataDbContext
+                    .DataSetVersionMappings,
+                mapping => mapping.TargetDataSetVersionId == dataSetVersion.Id,
+                cancellationToken);
```
ntsim, line 274: Similar to earlier comment - looks like the extension hasn't been used correctly as there is an ambiguous invocation. Can be fixed as:

```cs
var mappings = await publicDataDbContext
    .DataSetVersionMappings
    .Where(mapping => mapping.TargetDataSetVersionId == dataSetVersion.Id)
    .SingleOrDefaultAsync(cancellationToken: cancellationToken);
```

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/LocationMetaRepository.cs
```diff
@@ -226,35 +239,74 @@ private static List<GeographicLevel> ListLocationLevels(IReadOnlySet<string> all
...
+
+        if (mappings is null)
+        {
+            return new PublicIdMappings();
+        }
+
+        var mappingsByLevel = mappings
+            .LocationMappingPlan
```
ntsim, line 282: Similar to earlier comment - could select the `LocationMappingPlan` in the query to avoid fetching the rest of the row.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/LocationMetaRepository.cs
```diff
@@ -226,35 +239,74 @@ private static List<GeographicLevel> ListLocationLevels(IReadOnlySet<string> all
...
+        public Dictionary<GeographicLevel, Dictionary<string, string>> Levels { get; init; } = [];
+
+        public string? GetPublicIdForCandidate(GeographicLevel level, string candidateKey)
+        {
+            if (!Levels.TryGetValue(level, out var mappingLevel))
+            {
+                return null;
+            }
+
+            return mappingLevel.GetValueOrDefault(candidateKey);
```
ntsim, line 309: Similar to earlier comment - could be simplified to:

```cs
return Levels.GetValueOrDefault(level)
    ?.GetValueOrDefault(candidateKey);
```

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/FilterMetaRepository.cs
```diff
@@ -199,4 +218,64 @@ await optionTable
...
+    private static string CreatePublicIdForFilterOptionMetaLink(
+        PublicIdMappings publicIdMappings,
+        FilterMeta filter,
+        FilterOptionMeta option,
+        Func<string> defaultPublicIdFn)
+    {
+        return publicIdMappings
+                   .GetPublicIdForFilterOptionCandidate(
+                       filterKey: MappingKeyFunctions.FilterKeyGenerator(filter),
+                       MappingKeyFunctions.FilterOptionKeyGenerator(option))
```
ntsim, line 263: Would use a named parameter here for the candidate key as well, as per style guide

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetVersionMapping.cs
```diff
@@ -223,7 +225,22 @@ public record FilterMapping : ParentMapping<MappableFilter, MappableFilterOption
 /// </summary>
 public record FilterMappingPlan
 {
-    public Dictionary<string, FilterMapping> Mappings { get; set; } = [];
+    public Dictionary<string, FilterMapping> Mappings { get; init; } = [];
+
+    public Dictionary<string, FilterMappingCandidate> Candidates { get; init; } = [];
+}
+
+public static class MappingKeyFunctions
```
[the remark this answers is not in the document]
duncan-at-hiveit: Yup no probs - done

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/LocationOptionMetaLink.cs
```diff
@@ -1,12 +1,25 @@
...
 
     public required int OptionId { get; set; }
 
     public LocationOptionMeta Option { get; set; } = null!;
+
+    internal class Config : IEntityTypeConfiguration<LocationOptionMetaLink>
+    {
+        public void Configure(EntityTypeBuilder<LocationOptionMetaLink> builder)
+        {
+            builder.HasIndex(o => o.PublicId);
```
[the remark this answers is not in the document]
duncan-at-hiveit: Cool yup makes sense - done!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240710122937_EES4954_MovePublicIdToLocationMetaOptionLinkTable.cs
```diff
@@ -0,0 +1,69 @@
...
+            
+            migrationBuilder.CreateIndex(
+                name: "IX_LocationOptionMetaLinks_PublicId",
+                table: "LocationOptionMetaLinks",
+                column: "PublicId");
+
+            migrationBuilder.SqlFromFile("Migrations", 
+                $"{MigrationId}_{nameof(EES4954_MovePublicIdToLocationMetaOptionLinkTable)}.sql");
```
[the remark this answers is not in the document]
duncan-at-hiveit: Nope, and I saw that Ben and you were issuing direct SQL in other migrations after this, so I've changed it to do the same ta

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ImportMetadataFunctionTests.cs
```diff
@@ -156,6 +157,150 @@ public async Task DataSetVersionMeta_CorrectLocations(ProcessorTestData testData
...
+                                        keySelector: MappingKeyFunctions.LocationOptionMetaKeyGenerator,
+                                        elementSelector: option =>
+                                            option == mappedOption1 ? option1Mapping
+                                            : option == mappedOption2 ? option2Mapping
+                                            : new LocationOptionMapping
+                                            {
+                                                Type = i++ % 2 == 0 
+                                                    ? MappingType.AutoNone
+                                                    : MappingType.ManualNone
+                                            })
```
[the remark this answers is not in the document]
duncan-at-hiveit: Beautiful - changed ta

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ImportMetadataFunctionTests.cs
```diff
@@ -156,6 +157,150 @@ public async Task DataSetVersionMeta_CorrectLocations(ProcessorTestData testData
...
+            });
+
+            await ImportMetadata(testData, targetDataSetVersion, instanceId, metaInsertBatchSize);
+
+            await using var publicDataDbContext = GetDbContext<PublicDataDbContext>();
+
+            var actualLocations = await publicDataDbContext.LocationMetas
+                .Include(lm => lm.Options)
+                .Include(lm => lm.OptionLinks)
+                .ThenInclude(locationOptionMetaLink => locationOptionMetaLink.Option)
```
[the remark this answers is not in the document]
duncan-at-hiveit: Cool changed ta

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ImportMetadataFunctionTests.cs
```diff
@@ -241,6 +386,142 @@ public async Task DataSetVersionMeta_CorrectFilters(ProcessorTestData testData,
                 });
         }
 
+        [Theory]
+        [MemberData(nameof(TestDataFilesWithMetaInsertBatchSize))]
+        public async Task DataSetVersionMeta_CorrectFilters_WithMappings(ProcessorTestData testData,
```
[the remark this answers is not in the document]
duncan-at-hiveit: I think `DataSetVersionMeta_CorrectFilters_WithMappings()` is OK as we're testing not only the FilterOptions but also the Filters themselves in this test.  The name originates from the test method above which is called `DataSetVersionMeta_CorrectFilters()`.

Formatted `testData` as suggested, ta.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Tests/Functions/ImportMetadataFunctionTests.cs
```diff
@@ -156,6 +157,150 @@ public async Task DataSetVersionMeta_CorrectLocations(ProcessorTestData testData
...
+                {
+                    var actualLocation = actualLocations[index];
+                    actualLocation.AssertDeepEqualTo(expectedLocation,
+                        notEqualProperties: AssertExtensions.Except<LocationMeta>(
+                            l => l.Id,
+                            l => l.DataSetVersionId,
+                            l => l.Options,
+                            l => l.OptionLinks,
+                            l => l.Created
+                        ));
```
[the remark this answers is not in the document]
duncan-at-hiveit: This was inherited from the test method above which is called `DataSetVersionMeta_CorrectLocationOptions()`. I was just staying consistent with that, but adding in the additional side-effects of this being the next data set version being imported with the mappings, so I wanted to keep as close to the original "initial" import test method as possible to ensure no other unexpected side effects crept in as part of this work.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/FilterMetaRepository.cs
```diff
@@ -199,4 +218,64 @@ await optionTable
...
+
+    private async Task<PublicIdMappings> CreatePublicIdMappings(
+        DataSetVersion dataSetVersion,
+        CancellationToken cancellationToken)
+    {
+        var mappings = await EntityFrameworkQueryableExtensions
+            .SingleOrDefaultAsync(publicDataDbContext
+                    .DataSetVersionMappings,
+                mapping => mapping.TargetDataSetVersionId == dataSetVersion.Id,
+                cancellationToken);
```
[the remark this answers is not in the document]
duncan-at-hiveit: Ahh fantastic, yes good spot - looks like it works fine with your suggestion thanks

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/FilterMetaRepository.cs
```diff
@@ -199,4 +218,64 @@ await optionTable
...
+
+        if (mappings is null)
+        {
+            return new PublicIdMappings();
+        }
+
+        var mappingsByFilter = mappings
+            .FilterMappingPlan
+            .Mappings
```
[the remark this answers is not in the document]
duncan-at-hiveit: Good shout - done here and also in LocationMetaRepository

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/FilterMetaRepository.cs
duncan-at-hiveit: Sounds good to me - renamed here and also the equivalent in LocationMetaRepository.  Also renamed `CreatePublicIdMappings` to `GetPublicIdMappings` in both.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/FilterMetaRepository.cs
duncan-at-hiveit: Nice simplification - done ta

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/LocationMetaRepository.cs
duncan-at-hiveit: Done as part of the other change - ta

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/LocationMetaRepository.cs
duncan-at-hiveit: Done as part of the other change - ta

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/LocationMetaRepository.cs
duncan-at-hiveit: Nice - done ta

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/FilterMetaRepository.cs
duncan-at-hiveit: Cool np - updated ta

ntsim (approved):

PR #5049, opened 2024-07-11
EES-5281-backend Adding `IsLinkedToApiDataSet` flag to the `DataFileInfo` View Model

jack-hive:
This is the small backend change to add the `IsLinkedToApiDataSet` flag to the `DataFileInfo` View Model. This will be used by the 'Data Uploads' page to detect whether or not it should hide the 'Delete files' and 'Replace data' buttons for a subject. If the subject is linked to an API data set, we will hide these buttons. Otherwise, we will show them.

This was a quick solution, trying to avoid making any major changes/refactors.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
@@ -282,6 +282,10 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
                                 replacingDataFile: replacingFile,
                                 order: releaseDataFileOrder);
 
+                            var dataReleaseFile = await _releaseFileRepository.Find(
+                                releaseVersionId: releaseVersionId,
+                                fileId: dataFile.Id);
+
                             var metaFile = await _releaseDataFileRepository.Create(
                                 releaseVersionId: releaseVersionId,
                                 subjectId: subjectId,
@@ -300,7 +304,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
 
                             var permissions = await _userService.GetDataFilePermissions(dataFile);
 
-                            return BuildDataFileViewModel(dataFile: dataFile,
+                            return BuildDataFileViewModel(
+                                dataReleaseFile: dataReleaseFile!,
                                 metaFile: metaFile,
                                 validSubjectName,
                                 dataImport.TotalRows,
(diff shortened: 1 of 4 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -282,6 +282,10 @@ await _releaseVersionRepository
                                 replacingDataFile: replacingFile,
                                 order: releaseDataFileOrder);
 
+                            var dataReleaseFile = await _releaseFileRepository.Find(
+                                releaseVersionId: releaseVersionId,
+                                fileId: dataFile.Id);
```
benoutram, line 287: This returns an optional `ReleaseFile`, and I don't think the `ReleaseFile.File` is guaranteed to be included in the result of it, but is depended on by the `DataFileViewModel` method.

Just thinking it might be better to use the `ContentDbContext` variable directly here and remove these couple of potential issues with `Find`?

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
jack-hive: I'm undecided on this one. I was also thinking it was weird to use `_releaseFileRepository.Find` as it returns a nullable `ReleaseFile`. However, I thought it was probably OK because in this context both the `ReleaseFile` and `File` **should** exist. So I was happy to let it blow up if it was `null`, because it shouldn't ever happen. So In the end I decided to still use it, as it's nice to re-use the existing method.

But I also get why you'd want to use `ContextDbContext` directly to be more explicit, and I could use `.Single` rather than `.SingleOrDefault`.

benoutram: Yes I think it might be better to switchover or at least refactor it slightly so that you can specify what's included?

The `BuildDataFileViewModel` has a lot of references to `dataReleaseFile.File`. If that's working, I think it's only going to be because the `ReleaseFileRepository.Find` method has the line `.Include(releaseFile => releaseFile.File.CreatedBy)` which I'm guessing means the whole of `releaseFile.File` is being hydrated.

jack-hive: Yeah it's currently working because of that line, you're right.

I've decided to make it more explicit and write the query inline! :)

PR #5051, opened 2024-07-15
EES-5239 Add `PreviewToken` to Public API data model

benoutram:
This PR adds and configures the `PreviewToken` entity type to the Public API data model, following the writeup of the model in [EES-5239](https://dfedigital.atlassian.net/browse/EES-5239).

There is a zero-to-many relationship between `DataSetVersions` and `PreviewTokens`. `PreviewTokens` is made up of:

- **Id** (uuid) - The primary key of the table and also the security token, assuming that EES-4913 decides to secure the Public API with unique tokens.
- **Label** (character, max 100) - Label of the PreviewToken.
- **DataSetVersionId** (uuid) - Foreign key of the DataSetVersion which owns the PreviewToken.
- **CreatedByUserId** (uuid) - Id of the user in the Admin’s Identity model who created the PreviewToken. This is going to be used when building the view model to get or list PreviewTokens to look up the user and include their email address.
- **Created** (timestamp with timezone) - The created date of the PreviewToken.
- **Expiry** (timestamp with timezone) - The expiry date of the PreviewToken. This will always be set on creation and be updated if a user chooses to ‘expire’ a PreviewToken ahead of its scheduled expiry date.
- **Updated** (timestamp with timezone, optional) - This field exists for consistency with other entity types in the Public Data model. It’s value would only ever be expected to be null,  or the same value as Expiry if a PreviewToken token is updated to ‘expire’ it early. This might change if for example in future we allow a label to be updated or the validity to be extended.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/PublicDataDbContextModelSnapshot.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/PublicDataDbContextModelSnapshot.cs
@@ -25,6 +25,26 @@ namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Migration
 
             modelBuilder.HasSequence<int>("FilterOptionMetaLink_seq");
 
+            modelBuilder.Entity("GovUk.Education.ExploreEducationStatistics.Common.Model.JsonBool", b =>
+                {
+                    b.Property<bool>("BoolValue")
+                        .HasColumnType("boolean");
+
+                    b.ToTable((string)null);
+
+                    b.ToView(null, (string)null);
+                });
+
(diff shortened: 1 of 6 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/PublicDataDbContextModelSnapshot.cs
```diff
@@ -25,6 +25,26 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
             modelBuilder.HasSequence<int>("FilterOptionMetaLink_seq");
 
+            modelBuilder.Entity("GovUk.Education.ExploreEducationStatistics.Common.Model.JsonBool", b =>
```
duncan-at-hiveit, line 28: These again!

duncan-at-hiveit (approved):

PR #5056, opened 2024-07-16
EES-5335 Adding Timer Trigger to remove expired API subscriptions

jack-hive:
This PR adds a Timer Trigger to the Notifier project, which runs every hour, to remove all expired API subscriptions. The API subscriptions have an expiry time of 1 hour, in which you must verify your subscription. If you do not verify the subscription in time, this trigger will remove the pending subscription and you will need to request another.

Technically, this could let subscriptions sit around for up to 1 hour 59 minutes 59 seconds. But we have this issue for the publication subscriptions too.


src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/ApiSubscriptionManager.cs
```diff
@@ -65,4 +65,13 @@ public class ApiSubscriptionManager(
             return new StatusCodeResult(StatusCodes.Status500InternalServerError);
         }
     }
+
+    [Function("RemoveExpiredApiSubscriptions")]
+    public async Task RemoveExpiredApiSubscriptions(
+        [TimerTrigger("0 0 * * * *")] TimerInfo timerInfo,
```
benoutram, line 71: Just a comment around this bit of the PR description:

> Technically, this could let subscriptions sit around for up to 1 hour 59 minutes 59 seconds. But we have this issue for the publication subscriptions too.

Having this trigger set to every hour might lead someone to think there's a relationship between this interval and the duration that pending subscriptions are valid for, which is also one hour, which I think is actually the case but it took a fair bit of investigation to work this out!

I had thought that the pending subscriptions which are expired could be left indefinitely, because once they are expired they will not work, and that we only clean them up so that we don't retain unnecessary personal data in the storage table. If that was true, this function could be scheduled as frequently as once every few minutes or as little as once a month, as I don't think it would matter. More frequent is probably better!

After investigating though, it looks like when we get a pending subscription or verify a token we don't check the pending subscription expiry date and they only really expire as a result of being deleted?

I think I would be tempted to decouple this, possibly by changing https://github.com/dfe-analytical-services/explore-education-statistics/pull/5043 before that goes in, so that pending subscriptions have an actual expiry, and then they can be purged at any time after that, otherwise they haven't really 'expired'.

Just wondering what you think?

jack-hive: I think we actually **need** to clean up the expired subscriptions, because the table storage will only allow us to insert one entity with a unique combination of it's `PartitionKey` and `RowKey`. So if we then let the user try to re-request a pending subscription it will fail to create it.

But yes, currently you're right I think. It only really treats it as expired once deleted. I agree I should decouple this, and I'll make the changes in #5043. However, I think we would then introduce the following problem:

1. Pending subscription created with expiry of **now + 1 hour** - let's say '**now**' is 13:00pm, so expiry is 14:00pm
2. The timer trigger runs at 13:59pm --> this **does not** pick up this subscription to delete yet
3. at 14:00pm someone requests to verify the subscription. But we throw an error and say 'you can't, this subscription has expired'
4. So, the user tries to re-request another pending subscription. But they can't because we throw an error again to say 'you can't do this, a pending subscription already exists for this data set' (as we haven't yet deleted it!)
5. The user has to wait until 14:59pm for when the next timer trigger runs. At which point, the subscription is deleted and they can now successfully request another

So the user has ended up in this weird predicament where they can't do anything for anything between 0 --> 59 minutes depending when the first timer trigger has ran and when the subscription was first created

jack-hive: I've changed it to run every 1 minute, and have decoupled the code to check it has genuinely expired before verifying (so it is no longer just relying on it being deleted)

benoutram: Spoke with @jack-hive about this previously and we talked about there still being the same problem now, for up to a duration of one minute.

In future we could think about making a change to both Publication and API subscriptions so that if they request a pending subscription and one already exists which is expired, we send out a new token, but reuse the same pending subscription entity in table storage updated with the new expiry date on it.

That would mean we don't hit an error trying to insert a new entity with the same combination of `PartitionKey` and `RowKey`.

benoutram (approved):

PR #5058, opened 2024-07-16
EES-5324 api data set locations mapping page

amyb-hiveit:
Adds the API data set locations mapping page. For now this is read-only, updating the mappings will be done in subsequent PRs.

Note: I'm going to need to update the data structures used here to include the source and candidate keys, but I'll do that in the PR to update the mappings, so don't worry about it here.

Screenshot 2024-07-16 121050


src/explore-education-statistics-admin/src/pages/release/data/ReleaseApiDataSetLocationsMappingPage.tsx
```diff
@@ -0,0 +1,214 @@
...
+                errors={unmappedLocationErrors}
+                headingTag="h3"
+                title="Unmapped locations"
+                updateDocumentTitle={false}
+              />
+            )}
+            <p>
+              Existing locations in the publication data set should be mapped to
+              an equivalent location in the new data set. We try to
+              automatically match up locations that have the same name, but you
```
duncan-at-hiveit, line 117: `name and codes` strictly speaking, if you feel like that needs feeding back in?

src/explore-education-statistics-admin/src/pages/release/data/components/ApiDataSetNewLocationsTable.tsx
```diff
@@ -0,0 +1,44 @@
...
+      </thead>
+      <tbody>
+        {locations.map(({ candidate }) => {
+          return (
+            <tr key={`candidate-${candidate.code}`}>
+              <td>Not applicable to current data set</td>
+              <td>
+                {candidate.label}
+                <br />
+                {candidate.code}
```
duncan-at-hiveit, line 36: Note that there may be several codes to choose from depending on the level of the Location (e.g. `code` and `oldCode` for Local Authorities, `ukprn` for Providers etc, so there will possibly be login already in the front end code to find the appropriate codes to show based on the Location level.

src/explore-education-statistics-common/src/components/ErrorSummary.tsx
```diff
@@ -43,7 +54,11 @@ export const BaseErrorSummary = forwardRef<
       data-testid={testId}
     >
       <div role="alert">
-        <h2 className="govuk-error-summary__title">{title}</h2>
+        {createElement(
```
duncan-at-hiveit, line 57: Why the need for createElement here, just out of interest?

src/explore-education-statistics-common/src/components/PageNav.tsx
```diff
@@ -0,0 +1,105 @@
+import styles from '@common/components/PageNav.module.scss';
```
duncan-at-hiveit, line 1: Can the existing nav when viewing release contents on the public site be changed to use these common components at some point, or are they different beasts altogether?

duncan-at-hiveit (approved):

src/explore-education-statistics-admin/src/pages/release/data/components/ApiDataSetNewLocationsTable.tsx
amyb-hiveit: I can't find anywhere in the frontend where we currently handle this. Do you have a list of what it should be for each level?

src/explore-education-statistics-common/src/components/ErrorSummary.tsx
amyb-hiveit: Because you can now specify what heading tag to use, instead of always h2, so it needs to create the element using that.

src/explore-education-statistics-common/src/components/PageNav.tsx
amyb-hiveit: The one on the new data catalogue details page can probably use this and maybe other places, I didn't want to go far down a refactoring rabbit hole though so have left if for now.

PR #5059, opened 2024-07-16
EES-5238 - extracting Function logic into service layer following PR …

duncan-at-hiveit:
This PR:
- implements a suggestion in the initial PR review for EES-5238 to extract function logic into the service layer.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/CompleteNextDataSetVersionImportFunction.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/CompleteNextDataSetVersionImportFunction.cs
@@ -1,30 +1,25 @@
 using FluentValidation;
 using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Model;
-using GovUk.Education.ExploreEducationStatistics.Common.Validators;
-using GovUk.Education.ExploreEducationStatistics.Common.Validators.ErrorDetails;
-using GovUk.Education.ExploreEducationStatistics.Common.ViewModels;
-using GovUk.Education.ExploreEducationStatistics.Public.Data.Model;
 using GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Database;
 using GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Model;
 using GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Requests;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.ViewModels;
 using Microsoft.AspNetCore.Mvc;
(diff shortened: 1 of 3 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/CompleteNextDataSetVersionImportFunction.cs
```diff
@@ -35,10 +30,7 @@ public class CompleteNextDataSetVersionImportFunction(
     {
         return await requestValidator
             .Validate(request, cancellationToken)
-            .OnSuccess(_ => GetNextDataSetVersionInMappingStatus(request, cancellationToken))
-            .OnSuccessDo(_ => GetCompletedDataSetVersionMapping(request, cancellationToken))
-            .OnSuccessCombineWith(nextDataSetVersion =>
-                GetImportInManualMappingStage(request, nextDataSetVersion))
+            .OnSuccess(_ => mappingService.GetManualMappingVersionAndImport(request, cancellationToken))
```
jack-hive, line 33: I think I'd have moved the logic in the `.OnSuccess` below this into this `GetManualMappingVersionAndImport` method too. So that the service returns the view model. And I'd have named the method `CompleteNextDataSetVersionImport` or something instead.

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/CompleteNextDataSetVersionImportFunction.cs
jack-hive: So all the function ends up doing is  ==> receiving the request ==> delegating the business logic to the service layer, which returns a viewmodel ==> then the function handles the response

duncan-at-hiveit: A couple of examples are CreateDataSetFunction and CreateNextDataSetVersionMappingsFunction

jack-hive: I think `CreateDataSetFunction` and `CreateNextDataSetVersionMappingsFunction` could still be changed to return the view model in the service layer though couldn't they (not in this PR)?

duncan-at-hiveit: And what to do with the code which kicks off the main orchestration following the completion of the http trigger?  Would you suggest that goes in the service layer also?  Feels to me that you'd wanna retain that in the function code personally.

jack-hive: I think I'd still do that part in the function code! I'd agree it feels like it belongs in that layer

jack-hive (approved):

PR #5060, opened 2024-07-16
EES-5275 Add backend validation to request table queries

mmyoungman:
This PR adds validation to table queries provided in request objects to the backend. This is being done in response to EES-5256 where data blocks were created with no locations, which shouldn't be possible.

The FullTable™ requests are when we expect the request query to be enough to get details for a full table.
The LocationsOrTimePeriods requests are for two particular steps of the table tool - when the user has only selected at most a subject, locations, and a time period.

These request records are used to validate all steps of the table tool after selecting a location. They are also included in requests for creating and updating datablocks, and creating permalinks. They are validated in those cases too.

I did consider splitting endpoints that include a `LocationsOrTimePeriodsQueryRequest` object into two separate endpoints, but ultimately decided it wasn't worth it. If we decided to do this in the future it wouldn't be hard - it's more grunt work than anything else.

### Other changes
- `DataBlockCreateViewModel` and `DataBlockUpdateViewModel` are only used as requests, so have both been moved and renamed appropriately.


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Statistics/TableBuilderControllerIntegrationTests.cs
```diff
@@ -44,15 +45,18 @@ public class TableBuilderControllerIntegrationTests(TestApplicationFactory testA
...
             SubjectMeta = new SubjectResultMetaViewModel
             {
-                TimePeriodRange = new List<TimePeriodMetaViewModel>
-                {
-                    new(2020, AcademicYear),
-                    new(2021, AcademicYear),
-                }
+                TimePeriodRange =
+                [
+                    new TimePeriodMetaViewModel(2020, AcademicYear),
```
tomjonesdev, line 57: Does this new C# "feature" mean you have to define the types explicitly again? Normally the `TimePeriodMetaViewModel` type would be inferred for each item in the collection.

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/DataBlockRequests.cs
```diff
@@ -0,0 +1,59 @@
...
+using FluentValidation;
+using GovUk.Education.ExploreEducationStatistics.Common.Model.Chart;
+using GovUk.Education.ExploreEducationStatistics.Common.Model.Data;
+using GovUk.Education.ExploreEducationStatistics.Common.Requests;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Requests;
+
+public record DataBlockCreateRequest
+{
+    [Required] public string Heading { get; init; } = string.Empty;
```
tomjonesdev, line 13: I've been using the `required` _keyword_ for `...Request` classes, as my understanding was that the `[Required]` attribute was more of a code-first EF Core approach (based on other classes across the project).

src/GovUk.Education.ExploreEducationStatistics.Common/Requests/QueryRequests.cs
```diff
@@ -0,0 +1,90 @@
...
+    public class Validator : AbstractValidator<FullTableQueryRequest>
+    {
+        public Validator()
+        {
+            RuleFor(context => context.LocationIds)
+                .NotEmpty();
+            RuleFor(context => context.TimePeriod)
+                .NotNull();
+
+            // NOTE: No rule for filters - a data set might have no filters!
```
tomjonesdev, line 30: I don't think it's super necessary to explain the absence of something here.

src/GovUk.Education.ExploreEducationStatistics.Common/Requests/QueryRequests.cs
```diff
@@ -0,0 +1,90 @@
...
+
+public record FullTableQueryRequest
+{
+    public Guid SubjectId { get; set; }
+
+    public List<Guid> LocationIds { get; set; } = new();
+
+    public TimePeriodQuery? TimePeriod { get; set; }
+
+    public IEnumerable<Guid> Filters { get; set; } = new List<Guid>();
```
tomjonesdev, line 17: Should this be nullable if it's optional, or is it important for it to always be a list, even if it's an empty one?

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Controllers/TableBuilderControllerTests.cs
```diff
@@ -57,25 +56,25 @@ public class TableBuilderControllerTests(TestApplicationFactory testApp)
...
                     TimePeriod = new TimePeriodQuery
                     {
                         StartYear = 2021,
                         StartCode = CalendarYear,
                         EndYear = 2022,
                         EndCode = CalendarYear
-                    }
+                    },
+                    Filters = new List<Guid>(),
+                    Indicators = new List<Guid> // use collection expression -> test failures
```
tomjonesdev, line 68: Not sure on the meaning/purpose of this comment - leftover note-to-self?

tomjonesdev (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Statistics/TableBuilderControllerIntegrationTests.cs
mmyoungman: It doesn't but Rider was recommending showing the type name again - I assume the rationale here is that since we don't see the type in `new List<TimePeriodMetaViewModel>` it prefers to explicitly say the type when creating elements.

I'm not really fussed either way - I think my preference would be to have the type stated somewhere, but it's not too hard to hover over `TimePeriodRange` to see the type.

src/GovUk.Education.ExploreEducationStatistics.Common/Requests/QueryRequests.cs
mmyoungman: My thinking is that it's possible someone might see this code and ask themselves the question and wonder if it's been missed, and this comment prevents someone wasting any time exploring it as it answers the question immediately.

src/GovUk.Education.ExploreEducationStatistics.Common/Requests/QueryRequests.cs
mmyoungman: This is the way it was handled before my changes. I think you've got a point here, but it might be a bit of an involved change - I would probably have to change the frontend as well, and it could be nullable in several instances - for what I consider very little benefit.

src/GovUk.Education.ExploreEducationStatistics.Data.Api.Tests/Controllers/TableBuilderControllerTests.cs
mmyoungman: It's just a note that if you switch this from `new List<Guid>` to a collection expression, the unit test fails. I'll change the comment to make it clearer.

PR #5062, opened 2024-07-16
EES-5170: Add read permission to new BoundaryData table.

tomjonesdev:


src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240716123651_EES5170_AddBoundaryDataTablePermission.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace GovUk.Education.ExploreEducationStatistics.Data.Model.Migrations
+{
+    /// <inheritdoc />
+    public partial class EES5170_AddBoundaryDataTablePermission : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql("" +
+                "IF EXISTS (SELECT name " +
```
duncan-at-hiveit, line 14: Happy with the IF EXISTS, but I don't think you actually need it.  We always have it locally and in the Azure environment.

duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240716123651_EES5170_AddBoundaryDataTablePermission.cs
tomjonesdev: I'll leave it for now, no harm in an extra `SqlExcepion` failsafe!

PR #5064, opened 2024-07-16
EES-5270 Fix snapshots

mmyoungman:
This PR fixes snapshots after changes to the Find statistics and Data catalogue pages.

It also updates the snapshots.

### DIFF
--- a/tests/robot-tests/scripts/create_snapshots.py
+++ b/tests/robot-tests/scripts/create_snapshots.py
@@ -12,6 +12,7 @@ from selenium.common import NoSuchElementException
 from selenium.webdriver.chrome.options import Options
 from selenium.webdriver.common.by import By
 from selenium.webdriver.support.ui import WebDriverWait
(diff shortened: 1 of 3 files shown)


tests/robot-tests/scripts/create_snapshots.py
```diff
@@ -128,11 +129,11 @@ def _create_find_statistics_snapshot(self) -> str:
                 except NoSuchElementException:
                     publication_summary = ""
 
-                release_type = publication.find_element(By.XPATH, "./dl//dd[@data-testid='release-type']").text
+                release_type = publication.find_element(By.XPATH, "./dl//dd[@data-testid='Release type-value']").text
```
tomjonesdev, line 132: Feel free to ignore if it's not important, but the other `testid` formats shown here are snake-cased, i.e. `Release-type-value` instead of `Release type-value` (don't know if this matters).

N-moh (approved):

PR #5066, opened 2024-07-16
EES-4906 - added restrictions on when to show "Create new version" bu…

duncan-at-hiveit:
This PR:
- fixes a problem whereby the front end was letting you select data set candidates for the next version of an existing data set that would be invalid from the Data Processor's point of view.

## A quick note on terminology in this PR

The front end code still refers to a ReleaseVersion as "Release", and thus the ReleaseVersion's Id is `release.id`.

I actually need to provide ReleaseVersion.ReleaseId so we're able to provide that as a comparison when looking at rendering the "Create new version" buttons for a data set.  Having `release.id` and `release.releaseId` in the front end seemed confusing, so I opted to go for `release.id` and `release.releaseSeriesId` instead in an attempt to be clearer.  Let me know if I succeeded or not!

## The problem scenario

Specifically, the case that it was failing on was attempting to create a next version by using a data set candidate from a release amendment, where a previous version of that amendment contained a previous version of that data set.

For example, we have a "2023 Academic Year" `v0` Release, which contains a `V1.0` version of a data set.

"2023 Academic Year" `v0` gets published.

We then create an amendment of "2023 Academic Year", `v1`, AND upload a new Subject and then attempt to use it as the next `V1.1` version of the same data set that had its `V1.0` version on the `v0` release version.

The Data Processor considers this invalid, because after the amendment "2023 Academic Year" `v1` is published, this would result in the data set's `V1.0` version belonging to a release version that is no longer publicly visible (because the amendment has superseded it).

This PR now will hide the "Create new version" button in both the `Live Data Sets` list view and also on the `API Data Set` page itself when you're looking at it in the context of a release version that already has a version of that same data set in its amendment history.

## The correct user journey

We have a "2023 Academic Year" `v0` Release, which contains a `V1.0` version of a data set.

"2023 Academic Year" `v0` gets published.

The user would then create a "2024 Academic Year" `v0` Release (i.e. a new Release for the next year's statistics), which contains a `V1.1` version of the data set originally created on the 2023 Release.

"2023 Academic Year" `v0` gets published.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetsControllerTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetsControllerTests.cs
@@ -43,14 +43,41 @@ public abstract class DataSetsControllerTests(
         [Fact]
         public async Task PublicationHasSingleDataSet_Success_CorrectViewModel()
         {
-            Publication publication = DataFixture.DefaultPublication();
+            Publication publication = DataFixture
+                .DefaultPublication()
+                .WithReleases(
+                    DataFixture
+                        .DefaultRelease(publishedVersions: 1, draftVersion: true)
+                        .Generate(3)
+                );
+
+            var liveReleaseVersion = publication.ReleaseVersions.First(rv => rv.Published is not null);
+
+            var draftReleaseVersion = publication.ReleaseVersions.First(rv =>
+                rv.Published is null
+                && rv.ReleaseId != liveReleaseVersion.ReleaseId);
+
+            ReleaseFile liveReleaseFile = DataFixture
+                .DefaultReleaseFile()
+                .WithFile(DataFixture.DefaultFile(FileType.Data))
+                .WithReleaseVersion(liveReleaseVersion);
+
+            ReleaseFile draftReleaseFile = DataFixture
+                .DefaultReleaseFile()
+                .WithFile(DataFixture.DefaultFile(FileType.Data))
+                .WithReleaseVersion(draftReleaseVersion);
+
+            await TestApp.AddTestData<ContentDbContext>(context =>
+            {
+                context.Publications.Add(publication);
+                context.ReleaseFiles.AddRange(liveReleaseFile, draftReleaseFile);
+            });
 
             DataSet dataSet = DataFixture
                 .DefaultDataSet()
                 .WithStatusPublished()
                 .WithPublicationId(publication.Id);
 
-            await TestApp.AddTestData<ContentDbContext>(context => context.Publications.Add(publication));
             await TestApp.AddTestData<PublicDataDbContext>(context => context.DataSets.Add(dataSet));
 
             DataSetVersion draftDataSetVersion = DataFixture
@@ -58,12 +85,14 @@ public abstract class DataSetsControllerTests(
                 .WithVersionNumber(1, 1)
                 .WithStatusDraft()
                 .WithDataSet(dataSet)
+                .WithReleaseFileId(draftReleaseFile.Id)
                 .FinishWith(dsv => dataSet.LatestDraftVersion = dsv);
 
             DataSetVersion liveDataSetVersion = DataFixture
                 .DefaultDataSetVersion(filters: 1, indicators: 1, locations: 1, timePeriods: 2)
                 .WithStatusPublished()
                 .WithDataSet(dataSet)
+                .WithReleaseFileId(liveReleaseFile.Id)
                 .FinishWith(dsv => dataSet.LatestLiveVersion = dsv);
 
             await TestApp.AddTestData<PublicDataDbContext>(context =>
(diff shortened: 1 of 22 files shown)


src/explore-education-statistics-admin/src/pages/release/data/__tests__/ReleaseApiDataSetDetailsPage.test.tsx
```diff
@@ -371,6 +372,30 @@ describe('ReleaseApiDataSetDetailsPage', () => {
     ).not.toBeInTheDocument();
   });
 
+  test('does not render the create new version button when release series includes data set version', async () => {
```
amyb-hiveit, line 375: Is this the right wording? Isn't it checking for the data set previous release ids, not the data set version? (Same question for the live table version of this test below).

duncan-at-hiveit: Hi there, I've reworded slightly to:

`does not render the create new version button when release series includes a previous version of this data set`

and

`does not render 'Create new version' buttons when release series contains previous version of this data set`

Is this clearer?  Thanks!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
```diff
@@ -157,9 +163,36 @@ private async Task<DataSetViewModel> MapDataSet(DataSet dataSet, CancellationTok
...
+    }
+
+    private static async Task<List<Guid>> GetReleaseIdsForReleaseFiles(
+        ContentDbContext contentDbContext,
+        List<Guid> releaseFileIds,
+        CancellationToken cancellationToken)
+    {
+        return await contentDbContext
+            .ReleaseFiles
+            .Include(releaseFile => releaseFile.ReleaseVersion)
```
benoutram, line 190: Minor one but you don't need an `Include` here.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetsControllerTests.cs
```diff
@@ -43,27 +43,56 @@ public class ListDataSetsTests(
...
-            Publication publication = DataFixture.DefaultPublication();
+            Publication publication = DataFixture
+                .DefaultPublication()
+                .WithReleases(
+                    DataFixture
+                        .DefaultRelease(publishedVersions: 1, draftVersion: true)
+                        .Generate(3)
+                );
+
+            var liveReleaseVersion = publication.ReleaseVersions.First(rv => rv.Published is not null);
```
benoutram, line 54: Does this test rely on the there being a total of three releases?

I was struggling to understand this but I think this code is picking any 1 of the 3 published release versions, and then getting a draft version from one of the two _other_ releases.

I'm wondering if it would help by creating variables for the releases prior to declaring the publication?

```csharp
var (release1, release2) = _fixture.DefaultRelease(publishedVersions: 1, draftVersion: true).Generate(2).ToTuple2();
var liveReleaseVersion = release1.Versions.Single(rv => rv.Published is not null);
var draftReleaseVersion = release2.Versions.Single(rv => rv.Published is null);
```

or something like this?

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ReleaseViewModels.cs
```diff
@@ -13,6 +13,8 @@ public record ReleaseViewModel
     {
         public Guid Id { get; set; }
 
+        public Guid ReleaseSeriesId { get; set; }
```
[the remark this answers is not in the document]
duncan-at-hiveit: I have renamed to releaseId in this PR, and renamed all existing releaseIds to releaseVersionIds where appropriate in a follow-up PR (in admin codebase only).

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
duncan-at-hiveit: Gotcha - removed ta.  Must've been a hangover from previous attempts!

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetsControllerTests.cs
duncan-at-hiveit: I just wanted to prove that we'd end up with 3 `previousReleaseId` entries in the view model, rather than just the 2 that are attached to the live version and the draft version, should anyone be tempted to switch it up in the future and accidentally break this scenario.  I've added a comment as to why we're generating 3 here.

src/explore-education-statistics-admin/src/pages/release/data/components/LiveApiDataSetsTable.tsx
```diff
@@ -74,44 +76,45 @@ export default function LiveApiDataSetsTable({
...
+                        await apiDataSetVersionService.createVersion({
+                          dataSetId: dataSet.id,
+                          releaseFileId,
+                        });
+                        history.push(
+                          generatePath<ReleaseDataSetRouteParams>(
+                            releaseApiDataSetDetailsRoute.path,
+                            {
+                              publicationId,
+                              releaseId: releaseVersionId,
```
benoutram, line 110: Guessing this one and another on L79 is correct and will be tidied up by the big rename!

benoutram (approved):

src/explore-education-statistics-admin/src/pages/release/data/components/LiveApiDataSetsTable.tsx
duncan-at-hiveit: Yup!

PR #5069, opened 2024-07-17
EES-4915 Add endpoint to revoke a `PreviewToken`

benoutram:
This PR adds a new secured Admin endpoint to revoke a `PreviewToken`.

## Revoke a `PreviewToken`
Revoking a token is a `POST` request which accepts the `PreviewToken` id appended as a path parameter.

Example request:

```http
POST https://localhost:5021/api/public-data/preview-tokens/00000000-0000-0000-0000-000000000000/revoke
```

### Possible responses

If the request is valid the preview token is revoked by updating the `Expiry` to the current date, and a 200 OK response is returned with body containing the updated preview token resource. 

Example response:

```http
HTTP/1.1 200 OK

{
  "id": "00000000-0000-0000-0000-000000000000",
  "label": "Label",
  "status": "Expired",
  "createdByEmail": "ees-test.bau1@education.gov.uk",
  "created": "2024-07-19T12:42:30.762579+00:00",
  "expiry": "2024-07-19T12:42:39.7281552+00:00",
  "updated": "2024-07-19T12:42:39.7282109+00:00"
}
```

If the preview token is not found a 404 Not Found response is returned.

If the preview token is expired a 400 Bad Request is returned with body containing the error code and message:

```json
  "errors": [
    {
      "message": "The preview token is expired.",
      "path": "previewTokenId",
      "code": "PreviewTokenExpired",
      "detail": {
        "value": "00000000-0000-0000-0000-000000000000"
      }
    }
  ]
```

If the request is not authorised because of an invalid token or the user is not a BAU user a 401 Unauthorized response is returned.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Public.Data/PreviewTokenController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Public.Data/PreviewTokenController.cs
@@ -47,4 +47,14 @@ public class PreviewTokenController(IPreviewTokenService previewTokenService) : 
             .ListPreviewTokens(dataSetVersionId, cancellationToken)
             .HandleFailuresOrOk();
     }
+
+    [HttpDelete("{previewTokenId:guid}")]
+    public async Task<ActionResult> RevokePreviewToken(
+        Guid previewTokenId,
+        CancellationToken cancellationToken)
+    {
+        return await previewTokenService
+            .RevokePreviewToken(previewTokenId, cancellationToken)
+            .HandleFailuresOrNoContent(convertNotFoundToNoContent: false);
+    }
 }
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Public.Data/PreviewTokenController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Public.Data/PreviewTokenController.cs
@@ -49,12 +49,12 @@ public class PreviewTokenController(IPreviewTokenService previewTokenService) : 
     }
 
     [HttpPost("{previewTokenId:guid}/revoke")]
-    public async Task<ActionResult> RevokePreviewToken(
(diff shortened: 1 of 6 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Public.Data/PreviewTokenController.cs
```diff
@@ -36,4 +36,14 @@ public class PreviewTokenController(IPreviewTokenService previewTokenService) :
             .GetPreviewToken(previewTokenId, cancellationToken)
             .HandleFailuresOrOk();
     }
+
+    [HttpDelete("{previewTokenId:guid}")]
```
duncan-at-hiveit, line 40: My only comment here is that I would've personally expected a POST / PUT / PATCH rather than a DELETE, as I'd imagine that I'd be surprised to call a DELETE API endpoint but still see the entity I'd deleted in the `List` endpoint.

benoutram: Good point, I originally considered `PATCH` as the most appropriate as the `Expiry` field is updated, but since the caller won't be providing the value of `Expiry` and there will be no body, I ruled it out and went with `DELETE`.

I've updated this now from `DELETE` to `POST`, changing the request from 

```
DELETE https://localhost:5021/api/public-data/preview-tokens/00000000-0000-0000-0000-000000000000
```

to

```
POST https://localhost:5021/api/public-data/preview-tokens/00000000-0000-0000-0000-000000000000/revoke
```

Do you think the response should contain the updated resource or be empty (gone with empty for now)?

duncan-at-hiveit: I think I might've expected the updates resource in the response, but tbh empty is fine for me as well.

benoutram: Response added in! 🙂

PR #5076, opened 2024-07-18
EES-4929 Adding endpoint to unsubscribe users from API data sets

jack-hive:
This PR adds an endpoint to the Notifier project to allow users to unsubscribe from API data sets.

The status codes that the endpoint returns are:
- `204` the subscription was successfully deleted in the database (hence, the user is now unsubscribed)
- `400` either the token supplied is invalid, **or** the subscription has not yet been verified (you can't unsubscribe to something you were never subscribed to!)
- `404` the subscription does not exist (applies to both pending and verified subscriptions)

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/ApiSubscriptionFunctions.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/ApiSubscriptionFunctions.cs
@@ -44,7 +44,8 @@ public class ApiSubscriptionFunctions(
 
     [Function("VerifyApiSubscription")]
     public async Task<IActionResult> VerifyApiSubscription(
-        [HttpTrigger(AuthorizationLevel.Anonymous, "patch", Route = "public-api/{dataSetId:guid}/verify-subscription/{token}")]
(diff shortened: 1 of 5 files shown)


src/GovUk.Education.ExploreEducationStatistics.Notifier/Functions/ApiSubscriptionFunctions.cs
```diff
@@ -45,7 +45,7 @@ public class ApiSubscriptionFunctions(
 
     [Function("VerifyApiSubscription")]
     public async Task<IActionResult> VerifyApiSubscription(
-        [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = "public-api/{dataSetId:guid}/verify-subscription/{token}")]
+        [HttpTrigger(AuthorizationLevel.Anonymous, "patch", Route = "public-api/{dataSetId:guid}/verify-subscription/{token}")]
```
benoutram, line 48: Controversial one but my gut feeling is that this should be a `POST `since a `PUT` or a `PATCH` both expect a document in the body representing the resource or partial resource to be updated.

benoutram (approved):

PR #5078, opened 2024-07-18
EES-5220: Retain selected sorting method when applying search term, and default to 'Newest'

tomjonesdev:
Reduces confusion for end users who are currently seeing less relevant results and reduces frustration for internal users who are looking for recent publications and are finding the search doesn’t really help them.


src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPage.tsx
```diff
@@ -130,7 +130,7 @@ const FindStatisticsPage: NextPage = () => {
         : {
             ...omit(router.query, 'page'),
             [filterType]: nextValue,
-            sortBy: filterType === 'search' ? 'relevance' : sortBy,
+            sortBy: filterType === 'search' ? sortBy : 'newest',
```
amyb-hiveit, line 133: This is only keeping the user's previous sort order when you search, if you change another filter it resets to 'newest'.

amyb-hiveit (changes requested):
There's a ui test, `statistics_page.robot`, that checks that relevance is selected when you search, so that will need to be updated.

amyb-hiveit (approved):

PR #5079, opened 2024-07-19
EES-5345: Add Publisher's Guide page.

tomjonesdev:
This will give a clear set of links and guidance so that all users on the service can be aware of the surrounding support offering on the service.

### DIFF
--- a/src/explore-education-statistics-admin/src/components/PageHeader.tsx
+++ b/src/explore-education-statistics-admin/src/components/PageHeader.tsx
@@ -117,6 +117,13 @@ interface LoggedInLinksProps {
 
 const LoggedInLinks = ({ user }: LoggedInLinksProps) => (
   <>
+    {user.permissions.canAccessAnalystPages && (
+      <li className="govuk-header__navigation-item">
+        <a className="govuk-header__link" href="/publishers-guide">
+          Publisher's guide
+        </a>
+      </li>
+    )}
     {user.permissions.isBauUser && (
       <li className="govuk-header__navigation-item">
         <a className="govuk-header__link" href="/administration">
--- a/src/explore-education-statistics-admin/src/pages/publishers-guide/PublishersGuide.tsx
+++ b/src/explore-education-statistics-admin/src/pages/publishers-guide/PublishersGuide.tsx
@@ -1,0 +1,94 @@
+import Page from '@admin/components/Page';
+import React from 'react';
+
+const PublishersGuide = () => {
+  return (
+    <Page
+      title="Publisher's guide"
+      caption="Publishing on Explore education statistics"
+      wide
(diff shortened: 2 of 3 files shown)


src/explore-education-statistics-admin/src/pages/publishers-guide/PublishersGuide.tsx
```diff
@@ -0,0 +1,98 @@
...
+      <h2>Guidance</h2>
+      <p>
+        Detailed guidance on publishing statistics, covering everything from how
+        to implement Reproducible Analytical Pipelines, through to Open Data
+        Standards and details of using the Explore education statistics service
+        are available on the{' '}
+        <a
+          href="https://github.com/dfe-analytical-services/analysts-guide.git"
+          target="_blank"
+          rel="noreferrer"
```
ntsim, line 27: Use `noopener noreferrer` for external links. Same for other links below.

src/explore-education-statistics-admin/src/pages/publishers-guide/PublishersGuide.tsx
```diff
@@ -0,0 +1,98 @@
...
+        >
+          Statistics production and publishing teams area (opens in a new tab)
+        </a>
+        .
+      </p>
+
+      <ol>
+        <li>
+          <p>
```
ntsim, line 59: This is a list of so don't think we need this extra `p`?

Same applies for other list items below.

src/explore-education-statistics-admin/src/components/PageHeader.tsx
```diff
@@ -117,6 +117,13 @@ interface LoggedInLinksProps {
 
 const LoggedInLinks = ({ user }: LoggedInLinksProps) => (
   <>
+    {user.permissions.canAccessSystem && (
```
ntsim, line 120: Is this is the right permission? 

We have pre-release only users who should only have limited access to see specific releases. These types of users typically live outside of the statistics publishing team.

I'd imagine we probably shouldn't be letting these users see this type of information.

The `canAccessAnalystPages` permission is probably more appropriate.

src/explore-education-statistics-admin/src/routes/routes.ts
```diff
@@ -75,6 +76,13 @@ export const homeRoute: ProtectedRouteProps = {
   exact: true,
 };
 
+export const publishersGuideRoute: ProtectedRouteProps = {
+  path: '/publishers-guide',
+  component: PublishersGuide,
+  protectionAction: permissions => permissions.canAccessSystem,
```
ntsim, line 82: Similar to other comment, this probably should be `canAccessAnalystPages`

ntsim (changes requested):

ntsim (approved):

PR #5084, opened 2024-07-22
EES-4529 Upgrade minor dependencies that don't break stuff

mmyoungman:
This PR is to update minor dependencies that don't break stuff.

All the updates were taken from the renovate PR: https://github.com/dfe-analytical-services/explore-education-statistics/pull/5017

Updates that were included in the renovate PR that broke stuff so haven't been included in this PR:
- DuckDB.NET.Data.Full 0.10.3
- Swashbuckle.AspNetCore 6.6.2
- Microsoft.Azure.Functions.Worker.Extensions.DurableTask 1.1.4

I tried updating all these dependencies independent of any of the other updates, and they all caused unit tests to fail.

I also snuck in an update to pabot for the UI tests. This is to remove the deprecation warnings every time you run the UI tests.

### Xunit update

Updating xunit to 2.9.0 broke `Requests` unit tests in `GovUK.Education.EES.Public.Data.Api.Tests`. These tests would pass individually, but fail when run collectively with any integration test. When failing, property names were being camel cased while the tests expected this not to happen.

The reason this happened was due to FluentValidation global configuration found in `FluentValidationServiceCollectionExtensions#AddFluentValidation`. Here we set configuration that use camel case property names. Once this has run, any test that runs afterwards, the result of `TestValidate` calls will have a result with camel-cased property names.

To fix this, I created a test collection that ensures this camel case global config is set before any of the affected tests run, and then changed all the affected tests to expect camel case property names.

### Edit - Update note
The other three dependencies that weren't included in this PR have been merged to dev separately, so now this should complete all the backend depdency updates.


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/xunit.runner.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$schema": "https://xunit.net/schema/current/xunit.runner.schema.json",
+  "parallelAlgorithm": "conservative",
+  "parallelizeTestCollections": false,
```
tomjonesdev, line 4: Did this impact test suite duration much?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Filters/InvalidRequestInputFilterTests.cs
```diff
@@ -13,6 +13,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests.Filte
 
 public class InvalidRequestInputFilterTests(TestApplicationFactory testApp) : IntegrationTestFixture(testApp)
 {
+    /*
```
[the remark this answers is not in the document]
mmyoungman: I've uncommented them all. I was commenting them out while trying to debug the issue. Then I forgot to uncomment them :facepalm:

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/xunit.runner.json
mmyoungman: This was another attempt of mine to resolve the issue by tweaking xunit settings. It turned out these changes didn't solve the issue. I'll remove this `xunit.runner.json` file. Another :facepalm:

tomjonesdev (approved):

PR #5087, opened 2024-07-23
🐛 EES-5329 Add missing `ASPNETCORE_ENVIRONMENT` environment variable and force migration to rerun

benoutram:
This PR is a bugfix for https://github.com/dfe-analytical-services/explore-education-statistics/pull/5044.

While testing this ticket it appears the code in the migration `EES5329_GrantPublicDataReadWriteRolePrivileges` has not run in Azure, most probably due to the check `migrationBuilder.IsEnvironment(Environments.Production)` returning false.

We fix this by:

- Adding missing environment variable `ASPNETCORE_ENVIRONMENT` to the Public API Dockerfile.
- Changing `MigrationBuilderExtensions.IsEnvironment` to use `Production` as the default environment if `ASPNETCORE_ENVIRONMENT` is not set, consistent with the behaviour of `IHostEnvironment`.
-  Regenerating the migration `EES5329_GrantPublicDataReadWriteRolePrivileges` to force it to be rerun.


mmyoungman (approved):

src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/MigrationBuilderExtensions.cs
```diff
@@ -41,5 +45,10 @@ public static class MigrationBuilderExtensions
                 migrationBuilder.Sql(line);
             }
         }
+
+        private static string GetEnvironment() =>
+            Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT") ??
+            Environment.GetEnvironmentVariable("DOTNET_ENVIRONMENT") ??
+            Environments.Production;
```
luke-h1, line 52: Interesting to know - does Azure set this to production by default for the other APIs? 🤔 I assume so since the boxes are pre-configured to run dotnet but just curious

benoutram: Hey @luke-h1,

Yes it does and that's kind of the reason why I did it like this.

Here's a few references to it behaving this way in the docs:

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/host/web-host?view=aspnetcore-8.0#environment

image

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-8.0#azure-app-service

image

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-8.0#environments

image

PR #5089, opened 2024-07-23
EES-4950 Re-model data set version change sets

ntsim:
This PR re-models changelogs for data set versions (i.e. v1.1+) towards a more traditional normalised schema (instead of using JSONB). This provides the usual benefits of normalisation (referential integrity, de-duplication) and produces less overhead in terms of the code that needs to be written.

The downside is that larger changelogs can have many changes, potentially resulting in poorer performance than de-normalised JSONB (due to joining across potentially hundreds / thousands of rows).

After a little discussion with @duncan-at-hiveit, we agreed on balance, the benefits of the normalised approach outweighed the downsides. Additionally, any potential performance issues are addressable via caching (e.g. blob cache) of the entire changelog.


src/GovUk.Education.ExploreEducationStatistics.Common/Comparers/ListValueComparer.cs
```diff
@@ -0,0 +1,14 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.EntityFrameworkCore.ChangeTracking;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Comparers;
+
+public class ListValueComparer<TValue>(IEqualityComparer<TValue>? equalityComparer = null)
+    : ValueComparer<List<TValue>>(
+        (c1, c2) => c1!.SequenceEqual(c2!, equalityComparer),
```
duncan-at-hiveit, line 11: Are we always assured we're going to be given non-null Lists when using this Comparer, or should we build in some null-checking here?  I expect that where we've used this pattern previously in the codebase, we've always known that we're guaranteed to have non-null Lists, but now this is generic and reusable maybe we should build null safety in.

src/GovUk.Education.ExploreEducationStatistics.Common/Converters/EnumToEnumValueListConverter.cs
```diff
@@ -0,0 +1,30 @@
+#nullable enable
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Converters;
+
+public class EnumToEnumValueListConverter<TEnum>(ConverterMappingHints? mappingHints = null)
```
duncan-at-hiveit, line 9: Good to have

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Fixtures/DataSetVersionGeneratorExtensions.cs
```diff
@@ -408,81 +418,110 @@ public static InstanceSetters<DataSetVersion> SetStatusDraft(this InstanceSetter
...
         );
+    
+    public static InstanceSetters<DataSetVersion> SetGeographicLevelMetaChange(
+        this InstanceSetters<DataSetVersion> instanceSetter,
+        GeographicLevelMetaChange geographicLevelChange)
+        => instanceSetter.Set(
+            dsv => dsv.GeographicLevelMetaChange,
+            (_, dsv) =>
+            {
+                geographicLevelChange.DataSetVersion = dsv;
```
duncan-at-hiveit, line 460: Looks like elsewhere, you've opted to just call this parameter `change`, so you might wanna shorten this one too to match?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Change.cs
```diff
@@ -1,14 +1,170 @@
...
+                .HasOne(c => c.CurrentState)
+                .WithMany()
+                .IsRequired(false);
+
+            builder
+                .HasOne(c => c.PreviousState)
+                .WithMany()
+                .IsRequired(false);
+
+            builder.Navigation(c => c.PreviousState).AutoInclude();
```
duncan-at-hiveit, line 34: Good idea - we've used this once before for DataBlocks but I do wonder sometimes if there are other places where we could leverage this.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Change.cs
```diff
@@ -1,14 +1,170 @@
...
+    internal abstract class BaseConfig<TChange, TChangeEntity> : IEntityTypeConfiguration<TChange>
+        where TChange : Change<TChangeEntity>
+        where TChangeEntity : class
+    {
+        public virtual void Configure(EntityTypeBuilder<TChange> builder)
+        {
+            builder
+                .HasOne(c => c.CurrentState)
+                .WithMany()
+                .IsRequired(false);
```
duncan-at-hiveit, line 27: I guess the IsRequired being false for this and the Previous State allow things to get added and removed between data set versions?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Change.cs
```diff
@@ -1,14 +1,170 @@
-using GovUk.Education.ExploreEducationStatistics.Common.Utils;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
 
 namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Model;
 
-public class Change<TState>
+public abstract class Change<TEntity>
```
duncan-at-hiveit, line 6: Had you considered adding an additional `TId` generic param to this base class?  Then you could have:

```
    public TId? CurrentStateId { get; set; }

    public TId? PreviousStateId { get; set; }
```

on this base class and not have to add it to each subclass?

ntsim: I originally did model it like this, but it's not a correct generalisation for all the change types. For example, location and filter option changes are using owned types and don't use a single ID, instead they reference meta / option IDs like the link tables do. Consequently, this isn't really possible.

duncan-at-hiveit: Gotcha that makes sense, thanks!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Change.cs
duncan-at-hiveit: Cool np

duncan-at-hiveit (approved):

PR #5098, opened 2024-07-25
EES-4367 generate api data set preview tokens

amyb-hiveit:
Adds the ability to generate, view and revoke API data set preview tokens.

I've added `apiBaseUrl` and `apiDocsUrl` to the admin config, but they aren't defined however they get defined in the back end yet, and may end up being called other things.


src/explore-education-statistics-admin/src/pages/release/data/components/ApiDataSetPreviewTokenCreateForm.tsx
```diff
@@ -0,0 +1,87 @@
...
+    <FormProvider validationSchema={validationSchema}>
+      {({ formState }) => {
+        return (
+          <Form
+            id="apiDataSetTokenCreateForm"
+            onSubmit={values => onSubmit(values.label)}
+          >
+            <FormFieldTextInput<FormValues>
+              name="label"
+              label="Name this token instance"
```
ntsim, line 48: If we continue to reference this field with 'name' (instead of 'label'), would suggest we simplify this to 'Token name'.

src/explore-education-statistics-admin/src/pages/release/data/components/ApiDataSetPreviewTokenCreateForm.tsx
```diff
@@ -0,0 +1,87 @@
...
+              hint="Add a name so you can easily reference this token"
+            />
+            <FormFieldset
+              error={formState.errors?.terms?.message}
+              id="apiDataSetTokenCreateForm-terms"
```
ntsim, line 53: `FormFieldset` will already prefix the parent form's `id`, meaning that we currently end up with a rendered `id` of `apiDataSetTokenCreateForm-apiDataSetTokenCreateForm-terms`

We only need to define the part of the id that will be appended to the parent form id i.e. `id="terms"`

src/explore-education-statistics-admin/src/pages/release/data/components/ApiDataSetPreviewTokenCreateForm.tsx
```diff
@@ -0,0 +1,87 @@
...
+              legend="Terms of usage"
+              legendSize="s"
+            >
+              <FormFieldCheckbox
+                name="terms"
```
ntsim, line 58: Given the above comment, this field name will result in an ID clash with the fieldset. We could change this to something else like `agreeTerms` or `acceptTerms` to avoid this.

src/explore-education-statistics-admin/src/pages/release/data/components/ApiDataSetPreviewTokenCreateForm.tsx
```diff
@@ -0,0 +1,87 @@
...
+export default function ApiDataSetPreviewTokenCreateForm({
+  onCancel,
+  onSubmit,
+}: Props) {
+  const validationSchema = useMemo<ObjectSchema<FormValues>>(() => {
+    return Yup.object({
+      label: Yup.string().required('Enter a label'),
+      terms: Yup.boolean()
+        .required('The terms of usage must be accepted')
+        .oneOf([true], 'The terms of usage must be accepted'),
```
ntsim, line 35: Minor, but should this be 'The terms of usage must be agreed' to be consistent with what the checkbox actually says?

src/explore-education-statistics-admin/src/pages/release/data/components/__tests__/ApiDataSetPreviewTokenCreateForm.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+    render(
+      <ApiDataSetPreviewTokenCreateForm onCancel={noop} onSubmit={noop} />,
+    );
+
+    expect(
+      screen.getByLabelText('Name this token instance'),
+    ).toBeInTheDocument();
+    expect(
+      screen.getByLabelText(
+        'I agree to only share this token with individuals that have been granted production access to the unpublished data',
```
ntsim, line 17: Minor, but could avoid having to copy this entire thing in verbatim by using a regex e.g.

```ts
screen.getByLabelText(/I agree/);
```
This would be a little more robust to wording changes so we don't need to update the tests when it changes.

Similar could be done later in this test suite as well.

src/explore-education-statistics-admin/src/pages/release/data/ReleaseApiDataSetPreviewTokenPage.tsx
```diff
@@ -0,0 +1,163 @@
...
+    history.push(
+      generatePath<ReleaseDataSetPreviewTokenRouteParams>(
+        releaseApiDataSetPreviewRoute.path,
+        {
+          publicationId,
+          releaseId,
+          dataSetId,
+        },
+      ),
+    );
```
ntsim, line 74: Think there's a bug here. I'm finding that if I create a token, then revoke it, I'll end up back on the initial preview landing page, but the 'Generate API token' modal will already be open again.

Imagine this is because the modal state hasn't updated correctly and as we're currently redirecting to the same page component.

This should be resolvable by splitting out separate pages for the initial landing page and the page for viewing the preview token (as per earlier comment).

src/explore-education-statistics-admin/src/routes/releaseRoutes.ts
```diff
@@ -123,6 +129,13 @@ export const releaseApiDataSetLocationsMappingRoute: ReleaseRouteProps = {
   protectionAction: permissions => permissions.isBauUser,
 };
 
+export const releaseApiDataSetPreviewRoute: ReleaseRouteProps = {
+  path: '/publication/:publicationId/release/:releaseId/api-data-sets/:dataSetId/preview/:previewTokenId?',
+  title: 'API data set locations mapping',
+  component: ReleaseApiDataSetPreviewTokenPage,
+  protectionAction: permissions => permissions.isBauUser,
+};
```
ntsim, line 137: As per earlier comment - would split out routes for:

- initial preview landing page at `/preview` 
- page for viewing a preview token at `/preview/:previewTokenId`

src/explore-education-statistics-common/src/components/CopyLinkButton.tsx
```diff
@@ -1,51 +1,80 @@
...
 import useToggle from '@common/hooks/useToggle';
 import classNames from 'classnames';
-import React from 'react';
+import { OmitStrict } from '@common/types';
+import React, { ReactNode } from 'react';
 
-interface Props {
-  className?: string;
-  url: string;
+interface Props extends OmitStrict<CopyLinkModalProps, 'buttonClassName'> {
```
ntsim, line 10: This isn't very clean and it'll create an unnecessary circular dependency with `CopyLinkModal`.

Would suggest moving the props from `CopyLinkModal` to here instead and then extending `CopyLinkButtonProps` in `CopyLinkModal` i.e.

```ts
// CopyLinkButton.tsx

export interface CopyLinkButtonProps {
  buttonClassName?: string;
  className?: string;
  confirmMessage?: string;
  inlineButton?: boolean;
  url: string;
}

// CopyLinkModal.tsx

interface CopyLinkModalProps extends CopyLinkButtonProps {
  // ...
}
```

src/explore-education-statistics-common/src/components/CopyLinkButton.tsx
```diff
@@ -1,51 +1,80 @@
...
+  buttonText?: string | ReactNode;
+  label?: string;
 }
 
-const CopyLinkButton = ({ className, url }: Props) => {
+export default function CopyLinkButton({
```
ntsim, line 15: Is this the correct naming for this? 

We're using this component for more than just links so a more appropriate name would be something more generic like `CopyTextButton`

src/explore-education-statistics-common/src/components/CopyLinkButton.tsx
```diff
@@ -1,51 +1,80 @@
...
+  buttonText = 'Copy',
+  className,
+  confirmMessage = 'Link copied to the clipboard.',
```
ntsim, line 18: As above comment - a more generic default would be 'Text copied to the clipboard' if we change this component to `CopyTextButton`.

src/explore-education-statistics-common/src/components/CopyLinkButton.tsx
```diff
@@ -1,51 +1,80 @@
...
-      showClose
-      title="Copy link to the clipboard"
-      triggerButton={
-        <Button className={className} variant="secondary">
-          <LinkIcon />
-          <VisuallyHidden>Copy link to the clipboard</VisuallyHidden>
-        </Button>
-      }
-    >
+    <div className={className}>
```
ntsim, line 26: Do we need the extra div? Can't this be combined with the child container div?

src/explore-education-statistics-common/src/components/CopyLinkButton.tsx
```diff
@@ -1,51 +1,80 @@
...
 }
 
-const CopyLinkButton = ({ className, url }: Props) => {
+export default function CopyLinkButton({
+  buttonText = 'Copy',
+  className,
+  confirmMessage = 'Link copied to the clipboard.',
+  inlineButton = true,
+  label,
+  url,
```
ntsim, line 21: As per other comments - would make this component more generic and change this prop to `text`

ntsim (changes requested):

src/explore-education-statistics-common/src/components/CopyLinkButton.tsx
amyb-hiveit: It provides a container for the whole component so you can do things like add margin around it, the child container div doesn't include the message when it's inline.

src/explore-education-statistics-admin/src/pages/release/data/components/ApiDataSetPreviewTokenCreateForm.tsx
```diff
@@ -0,0 +1,87 @@
...
+      {({ formState }) => {
+        return (
+          <Form
+            id="apiDataSetTokenCreateForm"
+            onSubmit={values => onSubmit(values.label)}
+          >
+            <FormFieldTextInput<FormValues>
+              name="label"
+              label="Name this token instance"
+              hint="Add a name so you can easily reference this token"
```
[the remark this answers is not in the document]
amyb-hiveit: I've gone for 'enter a token name'.

ntsim (approved):

PR #5103, opened 2024-07-26
EES-5205 Update RebuildIndexes stored procedure

mmyoungman:
This PR updates the `RebuildIndexes` stored procedure to finish whatever index it is currently reorganising after 10 hours. So it will not stop exactly at 10 hours, but will not start reorganising any new index after 10 hours.

This was a suggestion made by Paul - you can read his investigation that is attached to the EES-5205 ticket.

I've also increased the Data Factory timeout so there is less chance that a long-running sole reindexing will take longer than the Data Factory timeout.

I've also reduced the timeout for Data Factory tasks related to posting success/failure messages. As they just send Slack messages, I see no reason for them to have a timeout of 12 hours.


infrastructure/templates/datafactory/components/db-maintenance-template.json
```diff
@@ -130,7 +130,7 @@
               },
               "body": {
                 "channel": "[parameters('slackAlertsChannel')]",
-                "text": "Data Factory Success!",
+                "text": "Reindexing complete or Data Factory timeout at 16hrs - stored procedure may still be running!",
```
benoutram, line 133: If you make this change we are going to question whether it's still running or not every time it's reported, or at least we would need to look at the run time to see if it's 16 hours or not, to determine if it was successful or a timeout.

If it completes after 16 hours, the data factory task should timeout at 16 hours and it should trigger the `Report failure` activity, so I think the text on L355 should be changed to report something like `Data Factory Failure! Stored procedure may still be running`, rather than this change here.

If it completes successfully within 10 hours it would report `Data Factory Success!`.

If it completes somewhere between 10 hours and 16 hours because it began stopping, it would either report `Data Factory Success` or `Data Factory Failure! Stored procedure may still be running` depending on whether you make the change suggested in my other comment about the stored procedure, which will affect whether it's considered a success or failure.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_RebuildIndexes.sql
```diff
@@ -110,7 +112,9 @@ BEGIN
 
                     SET @StatsRowIndex += 1;
                 END
-
-            RAISERROR ('Completed updating statistics on tables.', 0, 1, NULL) WITH NOWAIT;
+            IF(@StopTime > GETDATE())
+                RAISERROR ('Completed updating statistics on tables.', 0, 1, NULL) WITH NOWAIT;
+            ELSE
+                RAISERROR ('Optimising indexes and updating statistics on tables took longer than %d minutes. Stopping.', 0, 1, @StopInMinutes) WITH NOWAIT;
```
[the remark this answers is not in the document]
mmyoungman: Ah nice - and so that means the "Reindexing success" message doesn't need all the extra qualification.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_RebuildIndexes.sql
```diff
@@ -110,7 +112,9 @@ BEGIN
 
                     SET @StatsRowIndex += 1;
                 END
-
-            RAISERROR ('Completed updating statistics on tables.', 0, 1, NULL) WITH NOWAIT;
+            IF(@StopTime > GETUTCDATE())
+                RAISERROR ('Completed updating statistics on tables.', 0, 1, NULL) WITH NOWAIT;
+            ELSE
+                RAISERROR ('Reindexing did not complete in %d minutes. Remaining reindexing skipped.', 16, 1, @StopInMinutes) WITH NOWAIT;
```
benoutram, line 118: I think there's a case here where all the rebuilding/reorganising of indexes and updating of statistics has completed, but after `StopTime`.

I.e. in the second loop, `StatsRowIndex` is now greater than `StatsCount` so there's no rows left to process, and nothing will be skipped if it were to stop.  The last iteration to be processed was started before `StopTime` but finished after `StopTime`.

If the total run time on completion was over 16 hours, the data factory task will have timed out earlier and we would already have seen a notification like `Data Factory Failure! Stored procedure may still be running!`, which is fine as we may want to manually intervene early.

But if the total run time at this point is between 10 and 16 hours it will fail with a notification `Data factory Failure! Reindexing did not complete in 600 minutes. Remaining reindexing skipped.` giving the impression that there's been a failure and reindexing has been skipped, when reindexing did actually complete in the 16 hours allowed and nothing has been skipped.

If you are bothered about this I think you could make a change to only raise the 'failure' severity level error if processing in either blocks was skipped.

benoutram (approved):
Approved changes. Left a comment about when the error gets raised which you might want to consider changing.

PR #5104, opened 2024-07-28
EES-4921 Add public API endpoint for data set version changelog

ntsim:
This PR adds a new endpoint to the public API for getting a data set version's changelog.

The endpoint for this is: `GET /api/v1/data-sets/{dataSetId}/versions/{dataSetVersion}/changes`

An example response looks like:

```json5
{
  "majorChanges": {
    "filterOptions": [
      // Deletion
      {
        "previousState": {
          "id": "deleted-option",
          "label": "Deleted option"
        }
      }
    ]
  },
  "minorChanges": {
    "filterOptions": [
      // Updated label
      {
        "previousState": {
          "id": "updated-option",
          "label": "Original label"
        },
        "currentState": {
          "id": "updated-option",
          "label": "Updated label"
        }
      },
      // Addition
      {
        "currentState": {
          "id": "new-option",
          "label": "New option"
        }
      }
    ]
  }
}
```


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetVersionChangeService.cs
```diff
@@ -0,0 +1,258 @@
...
+                    CurrentState = c.CurrentState is not null
+                        ? FilterViewModel.Create(c.CurrentState)
+                        : null,
+                    PreviousState = c.PreviousState is not null
+                        ? FilterViewModel.Create(c.PreviousState)
+                        : null,
+                })
+                .GroupBy(c => c.IsMajor() ? ChangeType.Major : ChangeType.Minor)
+                .ToDictionary(g => g.Key, g => g.ToList())
+            : null;
```
benoutram, line 137: I was wondering if there is a generic method like `GetChanges` that could be abstracted out here. Did you already try anything like this?

Was thinking about something like this, but this would need some further work to make it compatible:

```csharp
    private static Dictionary<ChangeType, List<ChangeViewModel<TViewModel>>>? GetChanges<TViewModel, TEntity>(
        List<Change<TEntity>> changes,
        Func<Change<TEntity>, ChangeViewModel<TViewModel>> selector)  =>
        changes.Count != 0
            ? changes
                .Select(selector)
                .GroupBy(c => c.IsMajor() ? ChangeType.Major : ChangeType.Minor)
                .ToDictionary(g => g.Key, g => g.ToList())
            : null;
```

ntsim: Gave this a bit of a shot, but the generics are pretty gnarly as `Change` and `ChangeViewModel` themselves have their own type parameters.

Didn't really think it was worth pursuing this further as a bit of copy and paste is fine in this case.

benoutram: Ok, makes sense. Thanks for giving it a try! 🙂

benoutram (approved):

PR #5106, opened 2024-07-29
EES-5274 Added cron jobs to run multiple times on friday, saturday an…

saicharan2789:


azure-pipelines-ui-tests.dfe.yml
```diff
@@ -1,11 +1,30 @@
...
+    - dev
+  always: true
+
+- cron: 0 6,14,18,23 * * 6
+  branches:
+    include:
+    - dev
+  always: true
+
+- cron: 0 6,14,18,23 * * 0
```
mmyoungman, line 22: These could be merged into one cron job: `0 6,14,18,23 * * 0,5,6`

mmyoungman, line 22: I'm inclined to think that running them 4 times on saturday and sunday is excessive, especially as the tests rerun failed test suites themselves.

Maybe change this so they run just a couple of times, say once on saturday night and once on sunday night? I think that would be enough to reveal flaky test issues.

So I'd be inclined to go for one new cron string of `0 23 * * 0,6`

mmyoungman (changes requested):

azure-pipelines-ui-tests.dfe.yml
mmyoungman: Discussed this with Sai. We're adding these extra runs as a temporary measure just ascertain how flaky the tests are - so given that we're just doing this for one weekend, I'm happy for this to be approved.

mmyoungman (approved):

PR #5109, opened 2024-07-30
EES-5365 Update to `DuckDB.NET.Data.Full` 1.0.2

benoutram:
This PR updates projects related to the Public API to use `DuckDB.NET.Data.Full` 1.0.2.

It fixes the incorrect placement of `CreateRow` in methods `LocationsDuckDbRepository.CreateLocationsTable` and `SeedDataCommand.Seeder.CreateParquetLocationMetaTable` which was causing an error when upgrading between 0.10.2 and 1.0.0.

> System.IndexOutOfRangeException : The table location_options has 9 columns but you are trying to append value for column 10

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/LocationsDuckDbRepository.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/LocationsDuckDbRepository.cs
@@ -45,11 +45,11 @@ public class LocationsDuckDbRepository(PublicDataDbContext publicDataDbContext) 
         {
             using var appender = duckDbConnection.CreateAppender(table: LocationOptionsTable.TableName);
 
-            var insertRow = appender.CreateRow();
-
             foreach (var link in location.OptionLinks.OrderBy(l => l.Option.Label))
             {
                 var option = link.Option;
+
+                var insertRow = appender.CreateRow();
 
                 insertRow.AppendValue(id++);
                 insertRow.AppendValue(option.Label);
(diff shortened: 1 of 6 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/LocationsDuckDbRepository.cs
```diff
@@ -45,12 +45,12 @@ await publicDataDbContext
         {
             using var appender = duckDbConnection.CreateAppender(table: LocationOptionsTable.TableName);
 
-            var insertRow = appender.CreateRow();
```
ntsim, line 48: Pretty surprised this was working before, wonder how that was?

ntsim (approved):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/LocationsDuckDbRepository.cs
benoutram: I know, I was surprised too 🙂 .

I think the explanation is in the implementation of `DuckDB.NET.Data.DuckDBAppenderRow` in 0.10.2 which back then looked like this:

https://github.com/Giorgi/DuckDB.NET/blob/71a2908767a58a48f39a7486683fc6350fe7a93e/DuckDB.NET.Data/DuckDBAppender.cs#L19-L31

and from the docs linked in that code comment we see this:

image

so it wasn't doing anything before.

In 0.10.3 `DuckDbAppender` was rewritten to not use native methods and the `CheckColumnAccess` method was added to the new `AppendValue` method by https://github.com/Giorgi/DuckDB.NET/commit/9c1b5c8d5ae7ffef3605b6e8fb0b0d69065caf06.

PR #5113, opened 2024-07-31
EES-5362 Add admin endpoint for data set version changelog

ntsim:
## :warning: Depends on #5104

This PR adds the admin endpoint for fetching a data set version's changelog.

The endpoint is accessed via: `GET /api/public-data/data-set-versions/{dataSetVersionId}/changes`

We have implemented this by proxying requests to the public API's changelog endpoint via the admin. We have implemented it like this to avoid sharing view models and other code between the public API and the admin (keeping separation of concerns). Additionally, we can optimise by streaming the public API's response directly to the client rather than re-serialising into a second JSON response.

**Note** - This endpoint will not be fully functional until authentication is added to the public API to allow the admin API to fetch changelogs for draft data set versions. This should be implemented in [EES-5374](https://dfedigital.atlassian.net/browse/EES-5374). 

In the meantime, the endpoint should return a 500 response for any draft data set versions as the public API will respond to the admin client with a 403.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Settings/PublicDataApiOptions.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Settings/PublicDataApiOptions.cs
@@ -1,0 +1,9 @@
+#nullable enable
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Settings;
+
+public class PublicDataApiOptions
+{
+    public const string Section = "PublicDataApi";
+
+    public string Url { get; init; } = string.Empty;
+}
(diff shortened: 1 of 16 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Settings/PublicDataApiOptions.cs
```diff
@@ -0,0 +1,9 @@
+#nullable enable
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Settings;
+
+public class PublicDataApiOptions
+{
+    public static readonly string Section = "PublicDataApi";
```
benoutram, line 6: Minor one, this could be `public const string` rather than `public static readonly string` , but I see it wouldn't be consistent with `PublicDataProcessorOptions` then. Might be best to update both if you decide to change this. 🙂

benoutram (approved):

PR #5114, opened 2024-08-01
EES-4952 - Changelog: Save user inputted Changelog notes

jack-hive:
This PR adds a generic PATCH endpoint for BAU users to update a `DataSetVersion`.

Currently, the only properties that can be updated through this endpoint are the changelog `Notes`. In the future, we may add more properties to the request model. The idea is that the request parameters are all optional (aside from the `DataSetVersionId`) - and only the non-null fields will be updated.

The endpoint will return the following HTTP status codes:
- `200` 
  - with a body of `DataSetDraftVersionViewModel` when successfully updated
- `400`
  - when either:
    - the target `DataSetVersion` is in a status whereby it cannot be updated. Namely: `Processing`/`Published`/`Deprecated`/`Withdrawn`
    - the `DataSetVersion` is the **first** version, `1.0`, and we are trying to update the `Notes`. Only versions `>=1.1` can have changelog notes
- `403`
  - if the request was made by a non-BAU user
- `404`
  - if the target `DataSetVersion` does not exist

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionsControllerTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionsControllerTests.cs
@@ -11,12 +11,16 @@ using GovUk.Education.ExploreEducationStatistics.Admin.Requests.Public.Data;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.Public.Data;
 using GovUk.Education.ExploreEducationStatistics.Admin.Tests.Fixture;
 using GovUk.Education.ExploreEducationStatistics.Admin.ViewModels.Public.Data;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Model;
 using GovUk.Education.ExploreEducationStatistics.Common.Tests.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils;
 using GovUk.Education.ExploreEducationStatistics.Common.Utils;
 using GovUk.Education.ExploreEducationStatistics.Common.Validators;
 using GovUk.Education.ExploreEducationStatistics.Common.ViewModels;
+using GovUk.Education.ExploreEducationStatistics.Content.Model;
+using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
+using GovUk.Education.ExploreEducationStatistics.Content.Model.Tests.Fixtures;
 using GovUk.Education.ExploreEducationStatistics.Public.Data.Model;
 using GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Database;
 using GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests.Fixtures;
@@ -25,7 +29,14 @@ using LinqToDB;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Testing;
 using Moq;
+using System;
+using System.Linq;
+using System.Net.Http;
+using System.Security.Claims;
+using System.Threading;
+using System.Threading.Tasks;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Tests.Security.Utils.ClaimsPrincipalUtils;
+using ValidationMessages = GovUk.Education.ExploreEducationStatistics.Admin.Validators.ValidationMessages;
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Controllers.Api.Public.Data;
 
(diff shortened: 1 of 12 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
```diff
@@ -100,6 +123,115 @@ private static DataSetVersionSummaryViewModel MapDraftSummaryVersion(DataSetVers
...
+            GeographicLevels = dataSetVersion.MetaSummary?.GeographicLevels
+                .Select(l => l.GetEnumLabel())
+                .ToList() ?? null,
+            TimePeriods = dataSetVersion.MetaSummary?.TimePeriodRange is not null
+                ? TimePeriodRangeViewModel.Create(dataSetVersion.MetaSummary.TimePeriodRange)
+                : null,
+            Filters = dataSetVersion.MetaSummary?.Filters ?? null,
+            Indicators = dataSetVersion.MetaSummary?.Indicators ?? null
+        };
+    }
```
benoutram, line 206: Just wondering if we should call this `MapDraftVersion` like we do in `DataSetService`, although that could be because that also has a `MapLiveVersion`.

There's kind of a loose connection implied her,e that because the data set version `CanBeUpdated` property is false, that it must be a draft version (or one of the other draft statuses), and so we are mapping to `DataSetDraftVersionViewModel` rather than one of the other view models like `DataSetLiveVersionViewModel`.

I'm not sure whether there's anything we can do to improve on this? Perhaps a check at the beginning of this method which would throw an exception if the status is not one of the draft statuses?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetVersion.cs
```diff
@@ -82,7 +82,7 @@ public class DataSetVersion : ICreatedUpdatedTimestamps<DateTimeOffset, DateTime
     public DataSetVersionType VersionType
         => VersionMinor == 0 ? DataSetVersionType.Major : DataSetVersionType.Minor;
 
-    public bool CanBeDeleted => Status is DataSetVersionStatus.Failed
+    public bool CanBeUpdated => Status is DataSetVersionStatus.Failed
         or DataSetVersionStatus.Mapping
         or DataSetVersionStatus.Draft
         or DataSetVersionStatus.Cancelled;
```
benoutram, line 88: Should we allow updating changelong notes on a data set version that's failed or cancelled?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionsControllerTests.cs
```diff
@@ -302,6 +308,210 @@ public async Task ProcessorFailureStatusCode_Returns500()
...
     }
 
+    public class UpdateVersionTests(
+        TestApplicationFactory testApp) : DataSetVersionsControllerTests(testApp)
+    {
+        [Theory]
+        [InlineData(DataSetVersionStatus.Failed)]
+        [InlineData(DataSetVersionStatus.Mapping)]
+        [InlineData(DataSetVersionStatus.Draft)]
+        [InlineData(DataSetVersionStatus.Cancelled)]
```
benoutram, line 318: Something like what I did in `CreatePreviewTokenControllerTests` might work well here?

```csharp
[MemberData(nameof(NonDraftDataSetVersionStatuses))]
```

Perhaps the definition  of the draft and non draft data set statuses could be put somewhere else, except that for Preview tokens I think draft really means only `DataSetVersionStatus.Draft`.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetVersion.cs
```diff
@@ -82,7 +82,7 @@ public class DataSetVersion : ICreatedUpdatedTimestamps<DateTimeOffset, DateTime
     public DataSetVersionType VersionType
         => VersionMinor == 0 ? DataSetVersionType.Major : DataSetVersionType.Minor;
 
-    public bool CanBeDeleted => Status is DataSetVersionStatus.Failed
+    public bool CanBeUpdated => Status is DataSetVersionStatus.Failed
```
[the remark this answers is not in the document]
jack-hive: Good shout

However, on reflection of recent discussions, I'm going to add a new lambda for `CanBeUpdated` and leave `CanBeDeleted` as-is. I think we've agreed to only **update** when `Draft` or `Mapping`. So I need to use different criteria :)

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionsControllerTests.cs
jack-hive: Good idea

I've changed to using `MemberData` but have named it `UpdateableDataSetVersionStatuses` as I don't think `DraftDataSetVersionStatuses` would make sense anymore due to us only using `Draft` and `Mapping`. However, I'm also not completely happy with the name I have chose either, so I'm open to suggestions!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
```diff
@@ -100,6 +123,115 @@ private static DataSetVersionSummaryViewModel MapDraftSummaryVersion(DataSetVers
...
+
+    private async Task<Either<ActionResult, DataSetVersion>> UpdateVersion(
+        DataSetVersion dataSetVersion,
+        DataSetVersionUpdateRequest updateRequest,
+        CancellationToken cancellationToken)
+    {
+        if (updateRequest.Notes is not null)
+        {
+            dataSetVersion.Notes = updateRequest.Notes;
+        }
```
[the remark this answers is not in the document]
jack-hive: Torn on that one! I can't decide!

Since I can't decide, I'm tempted to leave it as-is if you think it is more readable :)

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
jack-hive: I agree `MapDraftVersion` would be better, I've renamed it :) . We no longer cover **all** of the draft statuses when updating now, but `MapDraftVersion` is still correctly named even if it will only be called with a subset of all the possible draft statuses.

In terms of adding a check, I'm not sure if this is necessary as we will be duplicating the check on line 141 in `CheckCanUpdateVersion`. I guess when looking at this method in isolation, it can be confusing though.

**EDIT:** Thinking out loud though - If someone comes along in the future and tries to re-use this method for mapping draft versions, then it should probably throw an error if it isn't a draft status. But in this particular scenario, we don't want to cover **all** draft statuses, only the ones that allow a version to be updated. So....... maybe the exception should be thrown if it isn't `Failed` or `Mapping`? And the method should more accurately be called `MapUpdateableVersion`. But I don't like that either!


What do you think?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetVersion.cs
jack-hive: We've now discussed this and agreed to only allow it on `Draft` and `Mapping` :)

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
```diff
@@ -100,6 +123,115 @@ private static DataSetVersionSummaryViewModel MapDraftSummaryVersion(DataSetVers
             Type = dataSetVersion.VersionType,
         };
     }
+
+    private async Task<Either<ActionResult, DataSetVersion>> GetDataSetVersion(
+        Guid dataSetVersionId,
+        CancellationToken cancellationToken)
+    {
+        return await publicDataDbContext.DataSetVersions
+            .AsNoTracking()
```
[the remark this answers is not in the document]
jack-hive: Ah, I did this because `GetDataSetVersion` could, in theory, be used by other things - and in those scenarios you may only need to read the data so don't need the tracking. And then here (which, granted, is the only use case for now) I've explicitly started tracking inside the update method.

Do you think it would be better to track it from the beginning instead?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/DataSetVersion.cs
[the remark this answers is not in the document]
benoutram: That sounds like the best way to go! 🙂

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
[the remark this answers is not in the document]
benoutram: That's fine by me! 🙂

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionsControllerTests.cs
benoutram: Great!

I haven't really got any other suggestions for the name. I think you could drop the 'e' to make it `Updatable` and you could also drop `DataSet` to make it `UpdatableVersionStatuses`?

Or how about `EligibleUpdateDataSetVersionStatuses` or some variation of that? 🙂

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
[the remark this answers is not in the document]
benoutram: Good to know, thanks. I think it's fine to leave it like this. 🙂

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
benoutram: I can't think of a good solution here that avoids duplicating the check already made on L141 and makes it reusable by any future work that needs to map draft versions.

It would also be a bit weird to call that check again here since it's named `CanBeUpdated` whereas you'd expect to see a check like `IsDraft`.

I think that given we don't have anything else using this right now and the `CanBeUpdated` check is already being made outside of this, that we leave it for now with just the rename to `MapDraftVersion`. 🙂

jack-hive: I've gone ahead with your suggestion and just done the rename for now :)

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionsControllerTests.cs
jack-hive: That was my poor spelling! 😆

I prefer your last suggestion! I've gone with `EligibleUpdateDataSetVersionStatuses` :)

benoutram (approved):

PR #5115, opened 2024-08-01
EES-5260 Added 'Run_identifier' variable to include test run instance…

saicharan2789:
This PR focusses on creating unique test run identifier value at each test run


tests/robot-tests/run_tests.py
```diff
@@ -208,7 +209,7 @@ def run():
                     os.makedirs(f"{results_foldername}/downloads")
 
                 # Create a unique run identifier so that this test run's data will be unique.
-                run_identifier = create_run_identifier()
+                run_identifier = f"{run_identifier_initial_value}-{test_run_index}"
```
mmyoungman, line 212: If the goal of this change is to prevent the intermittent errors we're getting around unique run ids, I don't think this will fix it. But I still like the change as it will be useful to link a run with all its reruns.

mmyoungman (approved):

PR #5117, opened 2024-08-01
EES-5314 Fix ui test failure

saicharan2789:
This PR focusses on fixing the intermittent failures of two test suites - '`publish_methodology_publication_update`' and '`create_data_block_with_Chart`'


saicharan2789:
image

tests/robot-tests/tests/admin_and_public_2/bau/publish_methodology_publication_update.robot
```diff
@@ -131,11 +131,13 @@ Update methodology details
     user waits until page contains element    xpath://h2[contains(text(),'Edit methodology summary')]
     user clicks radio    Set an alternative title
     user enters text into element    id:updateMethodologyForm-title    ${PUBLICATION_NAME}-methodology update
+    user waits until button is clickable    Update methodology
```
mmyoungman, line 134: You might already be aware, but this does have a little bit of overlap with `user clicks button`, which in turn calls `user clicks element` which also checks the element is enabled. But I'm inclined to says that's fine - `user waits until button is clickable` is doing other checks besides checking the element is enabled that may prevent a test failure.

tests/robot-tests/tests/libs/utilities.py
```diff
@@ -163,31 +163,40 @@ def parent_does_not_contain_matching_elements() -> bool:
...
+        except Exception as err:
+            if attempt < retries - 1:
+                logger.info(f"Retrying due to error... ({attempt + 1}/{retries})")
+                time.sleep(delay)
+                continue
+            else:
+                logger.warning(
+                    f"Error in get_child_element() with parent '{parent_locator}' and child locator '{child_locator}': {err}"
+                )
+                raise AssertionError(err)
```
mmyoungman, line 199: Could use `log_and_raise` here as well.

tests/robot-tests/tests/libs/utilities.py
```diff
@@ -163,31 +163,40 @@ def parent_does_not_contain_matching_elements() -> bool:
...
-            logger.warn(
-                f"Found {len(children)} child elements matching child locator {child_locator} "
-                f"under parent locator {parent_locator} in utilities.py#get_child_element() - "
-                f"was expecting only one. Consider making the parent selector more specific. "
-                f"Returning the first element found."
-            )
+def get_child_element(parent_locator: object, child_locator: str, retries: int = 5, delay: float = 1.0):
+    def log_and_raise(message: str):
+        logger.warning(message)
+        raise AssertionError(message)
```
mmyoungman, line 169: I'd be tempted to pull this out of `get_child_element` so it can be used elsewhere.

mmyoungman, line 169: Rather than `raise AssertionError(message)`, I think this should be `raise_assertion_error(message)`.

mmyoungman (changes requested):

tests/robot-tests/tests/libs/utilities.py
saicharan2789: I've pulled out `log_and_raise` by making use of existing method  ` raise_assertion_error`

tests/robot-tests/tests/libs/utilities.py
saicharan2789: I have completed removed above lines and made use of existing raise_assertion_error method defined from line 84-86

tests/robot-tests/tests/admin_and_public_2/bau/publish_methodology_publication_update.robot
saicharan2789: Yeah I am aware of that , rather than interested to put some sleep, I just thought to do this way for the sake of time killing . The actual problem is DOM is not getting updated as fast as the robot framework would require due to which its failing. Ideally I would be more interested in using some method like waitforFunction which exist in playwright - helps in preventing racing conditions.

mmyoungman (approved):

PR #5118, opened 2024-08-02
EES-5371 Added retries mechanism until child element is found inside …

saicharan2789:


tests/robot-tests/tests/admin/bau/create_data_block_with_chart.robot
```diff
@@ -206,6 +206,7 @@ Validate data block is in list
     user checks table column heading contains    1    4    Created date    testid:dataBlocks
     user checks table column heading contains    1    5    Actions    testid:dataBlocks
 
+    user reloads page
```
amyb-hiveit, line 209: I'm not sure about reloading the page here and in the other places, it could disguise problems that occur when you're navigating around the app that don't occur when you have a fresh page load.

saicharan2789: For some reason after saving data block, robot tests are unable to read the value inside the table. Due which ui tests are failing abnormally. The fix for this kind of problems is to refresh the page

tests/robot-tests/tests/libs/utilities.py
```diff
@@ -87,8 +87,20 @@ def raise_assertion_error(err_msg):
 
 
 def user_waits_until_parent_contains_element(
-    parent_locator: object, child_locator: str, timeout: int = None, error: str = None, count: int = None
+    parent_locator: object, child_locator: str, timeout: int = None, error: str = None, count: int = None, retries: int = 5, delay: float = 1.0
 ):
+    def _try_with_retries_child_element_found(func, *args, **kwargs):
```
tomjonesdev, line 92: Assuming that this sort of nesting is the equivalent of a `private` function, i.e. only accessible within the parent function scope, I think this would be better positioned _outside_ of the parent function as it appears to be quite generic, and could be used in other places where retries are needed.

Also I think the function's name would benefit from being a bit clearer, e.g. `retry_or_fail_with_delay`, or something similarly generic.

tests/robot-tests/tests/admin_and_public_2/bau/publish_methodology_publication_update.robot
```diff
@@ -133,6 +133,7 @@ Update methodology details
     user enters text into element    id:updateMethodologyForm-title    ${PUBLICATION_NAME}-methodology update
     user waits until button is clickable    Update methodology
     user clicks button    Update methodology
+    user reloads page
```
amyb-hiveit, line 136: @saicharan2789 can we use `user waits until page finishes loading` here as well?

saicharan2789: done

amyb-hiveit (approved):

PR #5119, opened 2024-08-02
EES-5373: Remove FullTableQuery.BoundaryLevel

tomjonesdev:
On the back end Admin and Data API services, `FullTableQueryRequest`/`FullTableQuery`
are being used by the frond end to execute a table query and optionally specify any boundary level data which should be returned so that for content page map charts, the boundary level data can be fetched and maps can be rendered.

The front end should be responsible for including the `BoundaryLevel` in the request if boundary level data should be returned, i.e. because boundary level data is needed for rendering a map chart.

This PR moves the property from the query into a separate parameter.


src/explore-education-statistics-admin/src/pages/release/datablocks/__tests__/ReleaseDataBlockEditPage.test.tsx
```diff
@@ -119,6 +119,7 @@ describe('ReleaseDataBlockEditPage', () => {
             dataSets: [],
           },
         },
+        boundaryLevel: undefined,
```
benoutram, line 122: Is it necessary to include this is `boundaryLevel` is optional?

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Statistics/TableBuilderController.cs
```diff
@@ -62,31 +63,34 @@ public class TableBuilderController : ControllerBase
...
         public async Task<ActionResult<TableBuilderResultViewModel>> QueryForDataBlock(
             Guid releaseVersionId,
             Guid dataBlockParentId,
+            long? boundaryLevelId,
             CancellationToken cancellationToken = default)
         {
             return await _dataBlockService
                 .GetDataBlockVersionForRelease(releaseVersionId: releaseVersionId, dataBlockParentId: dataBlockParentId)
-                .OnSuccess(dataBlockVersion => GetReleaseDataBlockResults(dataBlockVersion, cancellationToken))
+                .OnSuccess(dataBlockVersion => GetReleaseDataBlockResults(dataBlockVersion, boundaryLevelId, cancellationToken))
```
benoutram, line 79: Worth checking formatting on this file for long lines.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Controllers/TableBuilderController.cs
```diff
@@ -94,22 +94,23 @@ public class TableBuilderController : ControllerBase
...
+                .Query(releaseVersionId, request.AsFullTableQuery(), request.BoundaryLevel, cancellationToken)
                 .HandleFailuresOr(Ok);
         }
 
         // Note that releaseVersionId is not necessary for this method to function any more in the Data API, but remains in place
         // both in order to support legacy URLs in bookmarks and in content, but also to remain consistent with the equivalent
         // endpoint in the Admin API, which does require the Release Id in order to differentiate between different 
         // DataBlockVersions rather than simply picking the latest published one.
-        [HttpGet("tablebuilder/release/{releaseVersionId:guid}/data-block/{dataBlockParentId:guid}")]
+        [HttpGet("tablebuilder/release/{releaseVersionId:guid}/data-block/{dataBlockParentId:guid}/{boundaryLevelId:long?}")]
```
benoutram, line 105: See similar comment in Admin about changing this to use a query parameter `[FromQuery] long? boundaryLevelId`

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Controllers/TableBuilderController.cs
```diff
@@ -94,22 +94,23 @@ public class TableBuilderController : ControllerBase
             }
 
             return await _tableBuilderService
-                .Query(releaseVersionId, request.AsFullTableQuery(), cancellationToken)
+                .Query(releaseVersionId, request.AsFullTableQuery(), request.BoundaryLevel, cancellationToken)
```
benoutram, line 97: Does any request hit this endpoint with a `BoundaryLevel` value set. The intention here is that we get rid of `BoundaryLevel` from `FullTableQueryRequest` and hopefully nothing uses this and the value can be `null`, otherwise we should turn it into a query parameter like the other endpoints.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/ManageContentPageService.cs
```diff
@@ -1,17 +1,11 @@
 #nullable enable
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Threading.Tasks;
```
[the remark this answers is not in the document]
tomjonesdev: What's Rider's rationale for `System` being first when the other imports are alphabetical? [This ticket](https://youtrack.jetbrains.com/issue/RSRP-471356/Support-sorting-all-using-directives-alphabetically-including-System) references the option being added to Rider.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Statistics/TableBuilderController.cs
tomjonesdev: I run `pnpm format` before raising a PR but that doesn't seem to run on .cs files. Afaik VS doesn't have a line length setting because there's a _visual_ word wrapper which avoids changing the actual code/structure.
image

src/explore-education-statistics-admin/src/pages/release/datablocks/__tests__/ReleaseDataBlockEditPage.test.tsx
tomjonesdev: Yep, I don't understand why, maybe some TS quirk. I tried with null/undefined but it still insisted on it being declared.
image

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/reducers/__tests__/chartBuilderReducer.test.ts
```diff
@@ -604,6 +604,7 @@ describe('chartBuilderReducer', () => {
             },
           },
         },
+        boundaryLevel: undefined,
```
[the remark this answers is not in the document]
tomjonesdev: Yep, see comment on `ReleaseDataBlockEditPage.test.tsx`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/ManageContentPageService.cs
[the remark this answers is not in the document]
benoutram: I think this will have be the default option in Rider, and up until recently all team members used Rider, so a large percentage of the code will have `System` imports ordered first.

Taken from StyleCop analyzer, the rationale for ordering `System` imports first is:

>  Placing all System using directives at the top of the using directives can make the code cleaner and easier to read, and can help make it easier to identify the namespaces that are being used by the code.

I've put this as an item to discuss at the team meeting as it turns out we can try and enforce this an EditorConfig rule one way or the other which we are missing right now depending on what we decide.

See https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/dotnet-formatting-options

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Statistics/TableBuilderController.cs
benoutram: Ok I remember that we've chatted about this before. Might be worth chatting to @jack-hive to see if he's found a solution? I've added a note to the team meeting board to discuss the differences between IDE's when formatting long lines.

benoutram (approved):

PR #5120, opened 2024-08-02
EES-5368 RebuildIndexes now writes run data to the DB

mmyoungman:
This PR updates the RebuildIndexes stored procedure to write run data to the statistics database.

It writes data to these tables:
- `__RebuildIndexes` for the start/end time of the entire SP run
- `__RebuildIndexesAlterIndexes` for indexes which were either reorganised, rebuilt, or left alone
- `__RebuildIndexesUpdateStatistics` for the `UPDATE STATISTICS` queries run against tables

Unfortunately due to turning the initial `IF LEN(ISNULL(@Tables, '')) = 0` into a guard clause, this has mean the git diff between the old and new `Current_Routine_RebuildIndexes.sql`. After that made the git diff basically useless for comparison, I formatted the whole file with Rider.

I figured that the file had changed enough anyway that this isn't a big deal, but my sincere apologies anyway for making this PR harder to review.


src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
```diff
@@ -0,0 +1,283 @@
...
+      IF @StatsRowIndex >= @StatsCount
+        RAISERROR ('Completed optimising indexes for tables.', 0, 1, NULL) WITH NOWAIT;
+
+      SET @StatsRowIndex += 1;
+    END
+
+  /*-----------------------------------------------------------------------*/
+
+  SET @TablesRowIndex = 1;
+  SELECT @TablesCount = COUNT(Id) FROM #TableList;
```
benoutram, line 240: I think you might have missed a point that we only want to do `UPDATE STATISTICS` for tables that have had indexes touched by the first stage, but this refactoring to base this from `#TableList` instead of `#Stats` has changed this to do `UPDATE STATISTICS` for each table regardless.

I think this is unnecessary so I would revert this change unless you found something that says we should always be doing `UPDATE STATISTICS`?

Quite often on the run of this stored procedure, no indexes are touched because all the percentages fall below `FragmentationThresholdReorganize`.

If we wanted to do this every time regardless because there's no dependency on `#Stats` we could have looked two separate stored procedures would have probably made more sense.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
```diff
@@ -0,0 +1,283 @@
...
+      FROM #TableList
+      WHERE Id = @TablesRowIndex;
+
+      SET @Command = 'UPDATE STATISTICS ' + @CurrentTable;
+
+      -- Save start time
+      UPDATE __RebuildIndexesUpdateStatistics
+      SET StartTime = GETUTCDATE()
+      WHERE RunId = @RunId
+        AND TableName = @CurrentTable;
```
benoutram, line 254: Because there's no other data like `FragPercent` kept on this table, maybe you don't need to insert rows into this table early and update them with `StartTime` here, but put the insert here instead?

Saying that though, for the previous stage you have a method of updating each row when the timeout is hit which you don't have here. So maybe `__RebuildIndexesUpdateStatistics` should get a `HitTimeout` column which gets updated here on timeout, in which case keep the original method of inserting the rows early! 🙂

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
```diff
@@ -0,0 +1,283 @@
...
+      SET EndTime = GETUTCDATE()
+      WHERE RunId = @RunId
+        AND TableName = @CurrentTable;
+
+      -- If we completed the last update statistics query, it doesn't matter if we've exceeded @StopTime: we've not skipped anything
+      IF @TablesRowIndex + 1 <= @TablesCount AND @StopTime < GETUTCDATE()
+        SET @SkipRemainingTasks = 1;
+
+      IF @TablesRowIndex + 1 > @TablesCount
+        RAISERROR ('Completed updating statistics on tables.', 0, 1, NULL) WITH NOWAIT;
```
benoutram, line 271: Could move this outside the `WHILE` without the check on count to become a log message to signal that this stage has completed even if something has been skipped.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
```diff
@@ -0,0 +1,283 @@
...
+      SET @TablesRowIndex += 1;
+    END
+
+  IF @SkipRemainingTasks = 1
+    RAISERROR ('Reindexing did not complete in %d minutes. Remaining indexes skipped.', 16, 1, @StopInMinutes) WITH NOWAIT;
+
+  UPDATE __RebuildIndexes
+  SET EndTime    = GETDATE(),
+      HitTimeout = @SkipRemainingTasks
+  WHERE Id = @RunId;
```
benoutram, line 282: Have you checked to see if processing continues to this if the `RAISERROR` with severity level 16 is hit in the previous block? Just wondering if there's a bug here.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
```diff
@@ -0,0 +1,283 @@
...
+        RAISERROR ('Completed updating statistics on tables.', 0, 1, NULL) WITH NOWAIT;
+
+      SET @TablesRowIndex += 1;
+    END
+
+  IF @SkipRemainingTasks = 1
+    RAISERROR ('Reindexing did not complete in %d minutes. Remaining indexes skipped.', 16, 1, @StopInMinutes) WITH NOWAIT;
+
+  UPDATE __RebuildIndexes
+  SET EndTime    = GETDATE(),
```
benoutram, line 280: This should be `GETUTCDATE` to be consistent with all the other dates used throughout this stored procedure.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_EES5368_UpdateRebuildIndexesSPToWriteRunDataToDB.cs
```diff
@@ -0,0 +1,30 @@
...
+
+#nullable disable
+
+namespace GovUk.Education.ExploreEducationStatistics.Data.Model.Migrations
+{
+    /// <inheritdoc />
+    [ExcludeFromCodeCoverage]
+    public partial class EES5368_UpdateRebuildIndexesSPToWriteRunDataToDB : Migration
+    {
+        private const string PreviousRebuildIndexesMigrationId = "20240726153723";
```
benoutram, line 13: I would reference `EES5205_UpdateRebuildIndexesStoredProc.MigrationId` here rather than `20240726153723` so it can be clicked through to. That's the reason why we mark `MigrationId` as `internal` rather than `private` so that it can be referenced from the following migration.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
```diff
@@ -0,0 +1,283 @@
...
+  );
+
+  IF OBJECT_ID(N'__RebuildIndexesUpdateStatistics', N'U') IS NULL
+  CREATE TABLE __RebuildIndexesUpdateStatistics
+  (
+    [RunId]     UNIQUEIDENTIFIER NOT NULL,
+    [TableName] NVARCHAR(MAX)    NOT NULL,
+    [StartTime] DATETIME2,
+    [EndTime]   DATETIME2,
+  );
```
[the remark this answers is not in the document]
mmyoungman: I've added the `Log_` but kept the starting double underscore - so they're now `__Log_RebuildIndexes` etc.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
```diff
@@ -0,0 +1,283 @@
...
+             NULL,
+             FragPercent,
+             NULL,
+             NULL,
+             NULL
+      FROM #Stats
+      WHERE Id = @StatsRowIndex;
+
+      SET @StatsRowIndex += 1;
+    END
```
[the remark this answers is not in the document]
mmyoungman: Good point - removed the loop

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
```diff
@@ -0,0 +1,283 @@
...
+  (
+    [RunId]     UNIQUEIDENTIFIER NOT NULL,
+    [TableName] NVARCHAR(MAX)    NOT NULL,
+    [StartTime] DATETIME2,
+    [EndTime]   DATETIME2,
+  );
+
+  -- Create new row for this stored procedure run
+  INSERT INTO __RebuildIndexes (Id, StartTime, EndTime, HitTimeout, ReportSent)
+  VALUES (@RunId, GETUTCDATE(), NULL, NULL, NULL);
```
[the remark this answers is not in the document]
mmyoungman: Left a comment on the SCOPE_IDENTITY line that hopefully makes it clear what is happening.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
```diff
@@ -0,0 +1,283 @@
...
+  FROM #TableList TL;
+
+  /*-----------------------------------------------------------------------*/
+
+  -- Rebuild or reorganize each index depending on the fragmentation percentage
+  SET @StatsRowIndex = 1;
+
+  WHILE @StatsRowIndex <= @StatsCount AND @SkipRemainingTasks = 0
+    BEGIN
+      DECLARE @FragPercentBefore INT;
```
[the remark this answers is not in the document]
mmyoungman: I've moved @FragPercentBefore to be decalred alongside @Command at the top of the stored procedure.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
```diff
@@ -0,0 +1,283 @@
...
+    [StartTime]  DATETIME2        NOT NULL,
+    [EndTime]    DATETIME2,
+    [HitTimeout] BIT,
+    [ReportSent] DATETIME2,
+  );
+
+  IF OBJECT_ID(N'__RebuildIndexesAlterIndexes', N'U') IS NULL
+  CREATE TABLE __RebuildIndexesAlterIndexes
+  (
+    [Id]                 INT              NOT NULL,
```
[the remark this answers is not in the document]
mmyoungman: I've put the `PRIMARY KEY` bit at the bottom of the `CREATE TABLE` call, but otherwise it's the same.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
```diff
@@ -0,0 +1,283 @@
...
+          RAISERROR ('Skipping command - fragmentation percent did not hit reorganisation threshold ''%s''.', 0, 1, @Command) WITH NOWAIT;
+
+          UPDATE __RebuildIndexesAlterIndexes
+          SET Info = 'Skipped: fragmentation percentage did not hit reorganisation threshold'
+          WHERE RunId = @RunId
+            AND Id = @StatsRowIndex;
+
+          SET @StatsRowIndex += 1;
+          CONTINUE;
+        END
```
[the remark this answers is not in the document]
mmyoungman: I'm inclined to leave it - mostly to save me the time rejigging it. 

My preference is towards guard clauses because it's already confusing what's happening with all the indenting - although I do agree it's easy to get caught out not incrementing `@StatsRowIndex` - in fact I think I had that bug when developing this.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
```diff
@@ -0,0 +1,283 @@
...
+                        ON ips.object_id = OBJECT_ID(tl.ObjectName, 'U')
+                   JOIN sys.indexes idx ON idx.object_id = ips.object_id AND idx.index_id = ips.index_id
+                   JOIN sys.objects obj ON obj.object_id = ips.object_id
+                   JOIN sys.schemas sch ON sch.schema_id = obj.schema_id
+            WHERE ips.alloc_unit_type_desc = 'IN_ROW_DATA'
+              AND tl.ObjectName = @ObjectName
+              AND sch.Name = @SchemaName
+              AND idx.name = @IndexName) AS Fragmented
+      WHERE __RebuildIndexesAlterIndexes.RunId = @RunId
+        AND __RebuildIndexesAlterIndexes.Id = @StatsRowIndex;
```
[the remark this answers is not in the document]
mmyoungman: > I don't understand why we are doing this? What's the difference between InitialFragPercent which was written when the row was inserted based on the value in #Stats and StartFragPercent which is being worked out and set again here?

My thought was that rebuilding/reorganising a different index might affect the frag percent of a different index. 

I tried searching whether this is the case but didn't find anything and didn't find another. Google's AI said it's not - not sure how trustworthy that is. But regardless I've removed `InitialFragPercent` entirely and removed the setting of `StartFragPercent` out of the while loop.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
mmyoungman: Not sure if I tested it. I added a return after the RAISERROR to ensure the stored procedure stops at that point. 

Rider is greying out the RETURN, so it's probably unnecessary, but I didn't find a definitive answer online - this post is claiming execution does continue: https://dba.stackexchange.com/questions/48912/how-to-break-sql-script-execution

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
mmyoungman: I'm sure you're aware, but to make it clear, as there is no frag percent threshold with update statistics, there is no ambiguity as to why those rows are skipped if they have no start time - it's because of the timeout.

The action I'm most tempted to take would be to add a HitTimeout column, just so it makes it absolutely clear what has happened to anyone looking at the `__Log_RebuildIndexesUpdateStatistics` table with no other context about how it works.

I'm inclined to keep creating the rows early and not add a HitTimeout column - it will mean we execute one extra query, but I don't think that's a big deal. I think it makes sense for readability to follow the same pattern as the rebuilding rather than have two different approaches. And we avoid having to add an extra INSERT into `IF @TablesRowIndex + 1 <= @TablesCount AND @StopTime < GETUTCDATE()`.

All that said, I'm inclined to think this is all pretty minor and think any benefit of changing this isn't worth the time to do so.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
mmyoungman: > I think you might have missed a point that we only want to do UPDATE STATISTICS for tables that have had indexes touched by the first stage

You're right I didn't realise this.

Rather than revert this entirely, I've added a check that skips the update stats query if no indexes were rebuilt/reorganized for the current table.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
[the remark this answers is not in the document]
mmyoungman: I hadn't altered the formatting settings at all.

It seems the reason it was indenting 2 spaces, even though Rider was set to indent 4 spaces by default, was due Rider using the settings in `ees/.editorconfig`, so I added a section in `.editorconfig` for sql files.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_RebuildIndexes.sql
```diff
@@ -1,120 +1,302 @@
...
+              AND TableName = @CurrentTable;
+
+            -- Skip if no indexes related to table were rebuilt/reorganized
+            SELECT @CurrentTableIndexesProcessed = COUNT(*)
+            FROM __Log_RebuildIndexesAlterIndexes
+            WHERE RunId = @RunId
+              AND ObjectName = @CurrentTable
+              AND EndTime IS NOT NULL
+
+            IF @CurrentTableIndexesProcessed = 0
```
benoutram, line 259: This could all  be replaced with

`IF EXISTS (SELECT 1 FROM __Log_RebuildIndexesAlterIndexes WHERE ...)`

which would be more efficient than using `COUNT` since it usually ends up scanning the entire table.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_RebuildIndexes.sql
```diff
@@ -1,120 +1,302 @@
...
+                    CONTINUE;
                 END
-            IF(@StopTime > GETUTCDATE())
-                RAISERROR ('Completed updating statistics on tables.', 0, 1, NULL) WITH NOWAIT;
-            ELSE
-                RAISERROR ('Reindexing did not complete in %d minutes. Remaining reindexing skipped.', 16, 1, @StopInMinutes) WITH NOWAIT;
+
+            -- Run
+            RAISERROR ('Updating statistics on table ''%s''.', 0, 1, @CurrentTable) WITH NOWAIT;
+            EXEC (@Command);
```
benoutram, line 269: I would move the assignment of `@Command` directly before this so that it's set at the point that it's needed. If you go back a few revisions it was like this previously.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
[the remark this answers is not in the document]
benoutram: Cool. I'm more familiar with it being in-line with the column definition and seeing the `CONSTRAINT` key word if it's defined elsewhere, but I think this works.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_RebuildIndexes.sql
```diff
@@ -1,120 +1,302 @@
...
 
-            SET @StatsRowIndex = 1;
-            SET @LastTable = '';
-            SELECT @StatsCount = COUNT(Id) FROM #Stats;
-            WHILE(@StatsRowIndex <= @StatsCount AND @StopTime > GETUTCDATE())
+            -- Save end info for reporting
+            UPDATE __Log_RebuildIndexesAlterIndexes
+            SET __Log_RebuildIndexesAlterIndexes.EndTime        = GETUTCDATE(),
+                __Log_RebuildIndexesAlterIndexes.EndFragPercent = Fragmented.FragPercent
```
benoutram, line 197: I think I still have performance concerns about running this for a second time albeit on an object by object basis, but I guess we could see how it goes. Have you tried running this query for just one index on one of the big tables to see how long this takes?

`EndTime` is going to be recorded as the end time plus whatever time this takes, rather than the time `EXEC (@Command)` finished executing so could be a bit misleading if this takes a long time.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
benoutram: I think it's fine to leave this.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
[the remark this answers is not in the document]
benoutram: Cool, yeah I only left this comment because it requires more careful managing of the counter.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_RebuildIndexes.sql
```diff
@@ -1,120 +1,302 @@
...
-                        END
-
-                    RAISERROR ('Executing command ''%s''.', 0, 1, @Command) WITH NOWAIT;
-                    EXEC (@Command);
+                Fragmented.FragPercent,
+                CASE
+                    WHEN Fragmented.FragPercent >= @FragmentationThresholdRebuild THEN 'Rebuild'
+                    WHEN Fragmented.FragPercent >= @FragmentationThresholdReorganize THEN 'Reorganize'
+                    ELSE 'None'
+                    END As ActionRequired
```
benoutram, line 62: Minor one but `As` is inconsistent here.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_RebuildIndexes.sql
```diff
@@ -1,120 +1,302 @@
...
+                END
+
+            SET @TablesRowIndex += 1;
+        END
+
+    RAISERROR ('Completed updating statistics on tables.', 0, 1, NULL) WITH NOWAIT;
+
+    UPDATE __Log_RebuildIndexes
+    SET EndTime    = GETUTCDATE(),
+        HitTimeout = @SkipRemainingTasks
```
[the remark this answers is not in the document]
mmyoungman: What is the difference? Unless I'm missing something, this is what the formatter set, not me.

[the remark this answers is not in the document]
mmyoungman: Unless I'm missing something, I believe the formatter did this, not me. It always pads the equals signs so they're aligned.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_RebuildIndexes.sql
```diff
@@ -1,120 +1,302 @@
...
+        HitTimeout BIT,
+        ReportSent DATETIME2,
+        PRIMARY KEY (Id),
+    );
+
+    IF OBJECT_ID(N'__Log_RebuildIndexesAlterIndexes', N'U') IS NULL
+    CREATE TABLE __Log_RebuildIndexesAlterIndexes
+    (
+        Id               INT IDENTITY (1,1),
+        StatsRowIndex    INT           NOT NULL,
```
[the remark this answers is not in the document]
mmyoungman: I forgot to remove it after refactoring it so we no longer used it.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_RebuildIndexes.sql
mmyoungman: I tested this against the prod database - every index for the Observation and ObservationFilterItems tables took 8 seconds to fetch the fragmentation percent, except for IX_Observation_LocationId, which took 5 mins 30 secs.

Discussed this with Ben and we're going to leave this for now. We can work out how long the initial fetching of all frag percents takes by comparing the overall start time with the start time of the first index. Once we have that, we can determine whether we want to do other changes here, such as ensuring fetching the EndFragPercent doesn't increase the Start/End times we record of the reindexing or even the more extreme case of dropping EndFragPercent entirely.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20240802123234_Routine_RebuildIndexes.sql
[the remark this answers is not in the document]
mmyoungman: My bad... all sorted now

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_RebuildIndexes.sql
```diff
@@ -1,120 +1,298 @@
...
+           ObjectName,
+           FragPercent,
+           ActionRequired
+    FROM #Stats;
+
+    -- Create rows for tables
+    INSERT INTO __Log_RebuildIndexesUpdateStatistics (RunId, TableName)
+    SELECT @RunId,
+           ObjectName
+    FROM #TableList TL;
```
benoutram, line 151: In the latest set of changes `tl.ObjectName` was changed to `ObjectName` following a previous comment,

but rather than

```sql
INSERT INTO __Log_RebuildIndexesUpdateStatistics (RunId, TableName)
    SELECT @RunId,
           ObjectName
    FROM #TableList TL;
```

this can be

```sql
INSERT INTO __Log_RebuildIndexesUpdateStatistics (RunId, TableName)
    SELECT @RunId,
           ObjectName
    FROM #TableList;
```

benoutram (approved):

PR #5122, opened 2024-08-05
EES-5379 Fixed ui test failure

saicharan2789:


tests/robot-tests/tests/admin_and_public/bau/legacy_releases_reordering.robot
```diff
@@ -51,8 +52,11 @@ Validate that publication and legacy releases exist in the page
     user clicks link    Legacy releases
     user waits until h2 is visible    Legacy releases
 
+    ${PUBLIC_URL_WITHOUT_AUTH}=    remove auth from url    %{PUBLIC_URL}
+    set suite variable    ${PUBLIC_URL_WITHOUT_AUTH}
+
     user checks table cell contains    1    1    Academic year 2020/21
-    user checks table cell contains    1    2    %{PUBLIC_URL}/find-statistics/${PUBLICATION_NAME}/2020-21
+    user checks table cell contains    1    2    ${PUBLIC_URL_WITHOUT_AUTH}/find-statistics/${PUBLICATION_NAME}/2020-21
```
amyb-hiveit, line 59: @saicharan2789 can you explain why this is needed please? It's just checking a link exists, which is done using the `PUBLIC_URL` in lots of UI tests, so I don't understand how auth relates to it.

saicharan2789: Hi @amyb-hiveit,

What's happening is that while executing this step, it generates the PUBLIC_URL with the user ID and password appended. This causes it to fail locally when running against the dev environment (although it doesn't fail in the pipeline). That's why I created a new variable, PUBLIC_URL_WITHOUT_AUTH, which stores the URL without the authentication details.

amyb-hiveit: Ok, that makes sense thanks. Though I'm wondering why this isn't a problem with running other tests that reference `PUBLIC_URL` locally against dev?

saicharan2789: Yeah, it does fail in other tests as well. For instance, if you take a look at the `re-directs.robot` test suite, we experienced similar failures when comparing the URLs. For some reason, the robot tests are retrieving the `PUBLIC_URL` with the user ID and password appended, while the expected step doesn't have `PUBLIC_URL` with auth details. This causes it to fail consistently when running locally against the dev environment.

amyb-hiveit (approved):

PR #5124, opened 2024-08-05
EES-4920 Version History: Admin Endpoint to list published versions of an API dataset

jack-hive:
This PR adds a GET endpoint for BAU users to retrieve a **paginated** list of previously published Data Set Versions for a specific Data Set.

BAU Users should be able to see the version history for all previous versions of an API data set (but not any that are in draft status) within the “Api Datasets” tab under any current versions. 
For each version listed they will see its version number, related release, status and a link to view the changelog and a link to view details.

I have chosen to return slightly more data than necessary in the view model, in pursuit of trying to re-use the view-models we already have. I thought the tradeoff of re-using some view-models, on both the backend and client, outweighed the downside of returning more data than we need, for this particular view. In the future, we may need to display more data on this page anyway.

The endpoint will return the following HTTP status codes:
- `200` 
  - with a paginated list of previously published data set versions, if the request was successful
- `403`
  - if the request was made by a non-BAU user

When a successful request is made, the following body is returned:

```json
{
  "paging": {
    "page": 1,
    "pageSize": 10,
    "totalResults": 1,
    "totalPages": 1
  },
  "results": [
    {
      "id": "afbeb08c-b6c8-4e9d-d24e-08dc1cbc809a",
      "version": "1.0",
      "status": "Published",
      "type": "Major",
      "release": {
        "id": "bfbeb08c-b6c8-4e9d-d24e-08dc1cbc809b",
        "title": "My Release Version"
      },
      "published": "2024-08-05T16:34:27.432373+00:00"
    }
  ]
}
```

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
@@ -52,11 +52,11 @@ internal class DataSetService(
                     .Paginate(page: page, pageSize: pageSize)
                     .ToListAsync(cancellationToken);
 
-                var releasesByDataSetVersionByDataSet =
-                    await GetPreviousReleasesByVersionByDataSet(dataSets, cancellationToken);
+                var releaseVersionsByDataSetVersionIdByDataSetId =
+                    await GetPreviousReleaseVersionsByVersionIdByDataSetId(dataSets, cancellationToken);
 
                 var results = dataSets
-                    .Select(ds => MapDataSetSummary(ds, releasesByDataSetVersionByDataSet[ds.Id]))
+                    .Select(ds => MapDataSetSummary(ds, releaseVersionsByDataSetVersionIdByDataSetId[ds.Id]))
                     .ToList();
 
                 return new PaginatedListViewModel<DataSetSummaryViewModel>(
@@ -148,7 +148,7 @@ internal class DataSetService(
 
     private async Task<DataSetViewModel> MapDataSet(DataSet dataSet, CancellationToken cancellationToken)
     {
-        var previousReleaseFilesByVersion = await GetDataSetPreviousReleaseFilesByVersion(dataSet, cancellationToken);
+        var releaseFilesByDataSetVersionId = await GetReleaseFilesByDataSetVersionId(dataSet, cancellationToken);
 
         var draftVersion = dataSet.LatestDraftVersion is null
             ? null
@@ -157,17 +157,17 @@ internal class DataSetService(
                 mappingStatus: await GetMappingStatus(
                     nextDataSetVersionId: dataSet.LatestDraftVersion.Id,
                     cancellationToken),
-                previousReleaseFilesByVersion[dataSet.LatestDraftVersion.Id]
+                releaseFilesByDataSetVersionId[dataSet.LatestDraftVersion.Id]
             );
 
         var latestLiveVersion = dataSet.LatestLiveVersion is null
             ? null
             : MapLiveVersion(
                 dataSet.LatestLiveVersion,
-                previousReleaseFilesByVersion[dataSet.LatestLiveVersion.Id]
+                releaseFilesByDataSetVersionId[dataSet.LatestLiveVersion.Id]
             );
 
-        var previousReleaseIds = previousReleaseFilesByVersion
+        var allReleaseIds = releaseFilesByDataSetVersionId
             .Select(r => r.Value.ReleaseVersion.ReleaseId)
             .ToList();
 
@@ -180,11 +180,11 @@ internal class DataSetService(
             SupersedingDataSetId = dataSet.SupersedingDataSetId,
             DraftVersion = draftVersion,
             LatestLiveVersion = latestLiveVersion,
-            PreviousReleaseIds = previousReleaseIds
+            PreviousReleaseIds = allReleaseIds
         };
     }
 
-    private async Task<Dictionary<Guid, Dictionary<Guid, ReleaseVersion>>> GetPreviousReleasesByVersionByDataSet(
+    private async Task<Dictionary<Guid, Dictionary<Guid, ReleaseVersion>>> GetPreviousReleaseVersionsByVersionIdByDataSetId(
         IReadOnlyList<DataSet> dataSets, 
         CancellationToken cancellationToken)
     {
@@ -192,7 +192,7 @@ internal class DataSetService(
             .SelectMany(ds => ds.Versions)
             .ToDictionary(dsv => dsv.ReleaseFileId, dsv => dsv.DataSet);
 
-        var releasesByReleaseFileId = await contentDbContext
+        var releaseVersionsByReleaseFileId = await contentDbContext
             .ReleaseFiles
             .Include(rf => rf.ReleaseVersion)
             .Where(releaseFile => dataSetsByPreviousReleaseFileId.Keys.Contains(releaseFile.Id))
@@ -201,7 +201,7 @@ internal class DataSetService(
                 rf => rf.ReleaseVersion,
                 cancellationToken);
 
-        return releasesByReleaseFileId
+        return releaseVersionsByReleaseFileId
             .Select(d => new
             {
                 ReleaseFileId = d.Key,
@@ -218,11 +218,11 @@ internal class DataSetService(
             );
     }
 
-    private async Task<IReadOnlyDictionary<Guid, ReleaseFile>> GetDataSetPreviousReleaseFilesByVersion(
+    private async Task<IReadOnlyDictionary<Guid, ReleaseFile>> GetReleaseFilesByDataSetVersionId(
         DataSet dataSet, 
         CancellationToken cancellationToken)
     {
-        var dataSetVersionsByPreviousReleaseFileId = dataSet
+        var dataSetVersionsByReleaseFileId = dataSet
             .Versions
             .ToDictionary(dsv => dsv.ReleaseFileId);
 
@@ -230,9 +230,9 @@ internal class DataSetService(
             .ReleaseFiles
             .Include(rf => rf.ReleaseVersion)
             .Include(rf => rf.File)
-            .Where(releaseFile => dataSetVersionsByPreviousReleaseFileId.Keys.Contains(releaseFile.Id))
+            .Where(releaseFile => dataSetVersionsByReleaseFileId.Keys.Contains(releaseFile.Id))
             .ToDictionaryAsync(
-                rf => dataSetVersionsByPreviousReleaseFileId[rf.Id].Id,
+                rf => dataSetVersionsByReleaseFileId[rf.Id].Id,
                 cancellationToken);
     }
 
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
@@ -52,11 +52,11 @@ internal class DataSetService(
                     .Paginate(page: page, pageSize: pageSize)
                     .ToListAsync(cancellationToken);
 
-                var releaseVersionsByDataSetVersionIdByDataSetId =
-                    await GetPreviousReleaseVersionsByVersionIdByDataSetId(dataSets, cancellationToken);
+                var dataSetReleaseVersions =
+                    await GetDataSetReleaseVersions(dataSets, cancellationToken);
 
                 var results = dataSets
-                    .Select(ds => MapDataSetSummary(ds, releaseVersionsByDataSetVersionIdByDataSetId[ds.Id]))
+                    .Select(ds => MapDataSetSummary(ds, dataSetReleaseVersions[ds.Id]))
                     .ToList();
 
                 return new PaginatedListViewModel<DataSetSummaryViewModel>(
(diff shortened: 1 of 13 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
```diff
@@ -170,66 +184,56 @@ private async Task<DataSetViewModel> MapDataSet(DataSet dataSet, CancellationTok
...
-            .Select(releaseFile => releaseFile.ReleaseVersion.ReleaseId)
-            .ToListAsync(cancellationToken);
+            .Include(rf => rf.ReleaseVersion)
+            .Where(releaseFile => dataSetsByPreviousReleaseFileId.Keys.Contains(releaseFile.Id))
+            .ToDictionaryAsync(
+                rf => rf.Id,
+                rf => rf.ReleaseVersion,
+                cancellationToken);
+
+        return releasesByReleaseFileId
```
duncan-at-hiveit, line 204: Would rename this to `releaseVersionsByReleaseFileId`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
```diff
@@ -170,66 +184,56 @@ private async Task<DataSetViewModel> MapDataSet(DataSet dataSet, CancellationTok
...
-
-    private static async Task<List<Guid>> GetReleaseIdsForReleaseFiles(
-        ContentDbContext contentDbContext,
-        List<Guid> releaseFileIds,
+    private async Task<Dictionary<Guid, Dictionary<Guid, ReleaseVersion>>> GetPreviousReleasesByVersionByDataSet(
+        IReadOnlyList<DataSet> dataSets, 
         CancellationToken cancellationToken)
     {
-        return await contentDbContext
+        var dataSetsByPreviousReleaseFileId = dataSets
```
duncan-at-hiveit, line 191: Feels like we're missing a trick - we seem to do a lot of this same operation (well, maybe not a lot but enough that I notice it).  Wondering if this is an indicator that there's a complexity in our model that we could simplify somehow, or otherwise create some reusable code to skip around this whole ReleaseFile link.

duncan-at-hiveit: (nothing to do with this PR though - just highlighting it to me again)

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
```diff
@@ -170,66 +184,56 @@ private async Task<DataSetViewModel> MapDataSet(DataSet dataSet, CancellationTok
...
+                g => g.ToDictionary(
+                        a => a.DataSet.Versions.Single(v => v.ReleaseFileId == a.ReleaseFileId).Id,
+                        a => a.ReleaseVersion
+                    )
+            );
     }
 
-    private async Task<IReadOnlyDictionary<DataSetVersion, ReleaseFile>> GetReleaseFilesByDataSetVersion(
-        DataSet dataSet,
+    private async Task<IReadOnlyDictionary<Guid, ReleaseFile>> GetDataSetPreviousReleaseFilesByVersion(
```
duncan-at-hiveit, line 221: I would call this `GetPreviousReleaseFilesByDataSetVersionId` personally

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
```diff
@@ -170,66 +184,56 @@ private async Task<DataSetViewModel> MapDataSet(DataSet dataSet, CancellationTok
...
-            predicate = predicate.Or(rf => rf.Id == dataSet.LatestDraftVersion.ReleaseFileId);
-        }
-
-        if (dataSet.LatestLiveVersion is not null)
-        {
-            dataSetVersions.Add(dataSet.LatestLiveVersion);
-
-            predicate = predicate.Or(rf => rf.Id == dataSet.LatestLiveVersion.ReleaseFileId);
-        }
+        var dataSetVersionsByPreviousReleaseFileId = dataSet
```
duncan-at-hiveit, line 225: I'm not sure what makes these `previous`.  Can it not also contain the latest version's ReleaseFile as well?  In which case, maybe it needs a slight rename to `GetReleaseFilesByDataSetVersionId`

duncan-at-hiveit (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
```diff
@@ -34,6 +36,40 @@ public class DataSetVersionService(
...
+        int pageSize,
+        CancellationToken cancellationToken = default)
+    {
+        return await userService.CheckIsBauUser()
+            .OnSuccess(async () =>
+            {
+                var dataSetVersionsQueryable = publicDataDbContext.DataSetVersions
+                    .AsNoTracking()
+                    .Where(ds => ds.DataSetId == dataSetId)
+                    .WherePublicStatus();
```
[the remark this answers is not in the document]
jack-hive: I think `.WherePublicStatus()` is checking for:
- `Published`
- `Withdrawn`
- `Deprecated`

So I think they're in sync?

However, I wonder if it's worth creating an equivalent extension method, which isn't for an `IQueryable`, of `.WherePublicStatus` that can be used in the `ViewDataSetVersionAuthorizationHandler.cs` class. And the two extension methods would reuse the same logic - so can always keep them in sync

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
```diff
@@ -111,17 +163,60 @@ public class DataSetVersionService(
                 dataSetVersion: dataSetVersion,
                 updateRequest: updateRequest,
                 cancellationToken: cancellationToken))
-            .OnSuccess(MapDraftVersion);
+            .OnSuccess(async dataSetVersion => await MapDraftVersion(dataSetVersion, cancellationToken));
     }
 
-    private static DataSetVersionSummaryViewModel MapDraftSummaryVersion(DataSetVersion dataSetVersion)
+    private async Task<IReadOnlyDictionary<Guid, ReleaseVersion>> GetReleasesByDataSetVersion(
```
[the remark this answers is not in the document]
jack-hive: good shout! done

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
```diff
@@ -111,17 +163,60 @@ public class DataSetVersionService(
...
+        IReadOnlyList<DataSetVersion> dataSetVersions,
+        CancellationToken cancellationToken)
     {
+        var dataSetVersionsByReleaseFileId = dataSetVersions
+            .ToDictionary(dsv => dsv.ReleaseFileId);
+
+        var releasesByReleaseFileId = await contentDbContext
```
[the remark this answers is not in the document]
jack-hive: good shout! done

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
```diff
@@ -111,17 +163,60 @@ public class DataSetVersionService(
...
+                cancellationToken);
+
+        return releasesByReleaseFileId
+            .ToDictionary(
+                d => dataSetVersionsByReleaseFileId[d.Key].Id,
+                d => d.Value
+            );
+    }
+
+    private static DataSetLiveVersionSummaryViewModel MapLiveSummaryVersion(
```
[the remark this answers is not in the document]
jack-hive: good shout! done

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
```diff
@@ -111,17 +163,60 @@ public class DataSetVersionService(
...
+            Id = dataSetVersion.Id,
+            Version = dataSetVersion.Version,
+            Status = dataSetVersion.Status,
+            Type = dataSetVersion.VersionType,
+            ReleaseVersion = MapReleaseVersion(releaseVersion),
+            Published = dataSetVersion.Published!.Value
+        };
+    }
+
+    private async Task<DataSetVersionSummaryViewModel> MapDraftSummaryVersion(
```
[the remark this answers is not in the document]
jack-hive: good shout! done

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
jack-hive: good shout! done

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
jack-hive: good shout! done

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
jack-hive: Now that you mention it, this doesn't make sense to me either!

I've renamed things to make it a bit clearer :)

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
jack-hive: Are we talking about having to navigate through the related `DataSetVersions`, from the `DataSet`, in order to get the `ReleaseFileId`?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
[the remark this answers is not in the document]
duncan-at-hiveit: Aha sorry - I see that there are 2 `WherePublicStatus` implementations.  I was looking at the one for DataSets, not DataSetVersions!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
duncan-at-hiveit: Cool!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
duncan-at-hiveit: Yeah - as I said, nothing to be done in this PR and maybe nothing that needs to be done at all - it's just in the back of my mind that this area of our model feels complex.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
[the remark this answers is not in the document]
duncan-at-hiveit: Yeah good idea on having another extension method.  I guess we'd put it in `DataSetVersionAuthExtensions` so the 2 pieces of code are right next to each other - also could extract the statuses being checked into a readonly array so they can both check against exactly the same list?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
```diff
@@ -52,11 +52,11 @@ internal class DataSetService(
                     .Paginate(page: page, pageSize: pageSize)
                     .ToListAsync(cancellationToken);
 
-                var releasesByDataSetVersionByDataSet =
-                    await GetPreviousReleasesByVersionByDataSet(dataSets, cancellationToken);
+                var releaseVersionsByDataSetVersionIdByDataSetId =
```
duncan-at-hiveit, line 55: I see the reasoning behind the rename, but this is quite a hefty variable name and I can't quite get my head around what it represents!  Happy to approve the PR but I think that possibly it might be clearer if the Dictionary of Dictionaries that we return from `GetPreviousReleaseVersionsByVersionIdByDataSetId` was actually some class / record with a meaningful name rather than a nested Dictionary.

One to mull over!

duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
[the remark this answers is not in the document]
jack-hive: we've now broken this out into an extension method as discussed :)

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetService.cs
jack-hive: It is a bit lengthy, I must admit!

I'll create a private record that I can use for it. Please let me know if you think it's any better :)

PR #5126, opened 2024-08-06
EES-5253 Squash Public API migrations

benoutram:
This PR squashes all the Public API database model migrations and creates a new `InitialMigration`.

### New version of Public API seed data

There is a new version `public-api-17.zip` of the seed data which has been generated with the new migration.


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240723085620_EES5329_GrantPublicDataReadWriteRolePrivileges.cs
```diff
@@ -1,54 +0,0 @@
...
-                  BEGIN
-                      IF EXISTS (SELECT 1 FROM pg_roles WHERE rolname = '{PublicDataDbContext.PublicDataReadWriteRole}') THEN
-                          GRANT SELECT, INSERT, UPDATE, DELETE, TRUNCATE, REFERENCES ON ALL TABLES IN SCHEMA public TO {PublicDataDbContext.PublicDataReadWriteRole};
-                          GRANT SELECT, UPDATE ON ALL SEQUENCES IN SCHEMA public TO {PublicDataDbContext.PublicDataReadWriteRole};
-
-                          ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT, INSERT, UPDATE, DELETE, TRUNCATE, REFERENCES ON TABLES TO {PublicDataDbContext.PublicDataReadWriteRole};
-                          ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT, UPDATE ON SEQUENCES TO {PublicDataDbContext.PublicDataReadWriteRole};
-                      END IF;
-                  END $$;
-                  """);
```
jack-hive, line 45: I think this has an extra couple of lines compared to the equivalent SQL we've brought over to the initial migration. Not 100% sure if they're needed though?

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240723085620_EES5329_GrantPublicDataReadWriteRolePrivileges.cs
benoutram: The two additional lines, i.e.

```sql
GRANT SELECT, INSERT, UPDATE, DELETE, TRUNCATE, REFERENCES ON ALL TABLES IN SCHEMA public TO {PublicDataDbContext.PublicDataReadWriteRole};
GRANT SELECT, UPDATE ON ALL SEQUENCES IN SCHEMA public TO {PublicDataDbContext.PublicDataReadWriteRole};
```

were necessary because this was not originally run as part of the initial migration in Azure environments and we needed them here to make sure permissions were granted to all *existing*  objects in the public schema, not just those created subsequently after this had run (which  is covered by the later two lines).

In the equivalent code block in `InitialMigration` we only need the two lines to alter default privileges, since it's stuck right at the very beginning of the migration it should cover all objects created after it.

jack-hive: Awesome, happy with that then! :)

jack-hive (approved):

PR #5127, opened 2024-08-06
Ees 5374 add easyauth to public api endpoints with app roles

duncan-at-hiveit:
:warning: Adding "Do Not Merge" for now, as when this goes out it'll probably cause some disruption on Dev!

# Overview

This PR:
- separates Azure Authentication configuration into a separate reusable module.
- enables Azure Authentication in the Public API Container App, with unauthenticated calls allowed.
- whitelists the Admin App and its Managed Identity to authenticate with the Public API.
- updates Public API security checks around the Changelog access to allow callers with the "UnpublishedData.Read" App Role access to unpublished data.

# The setup

1. We are enabling Azure Authentication on the Public API Container App, and allowing unauthenticated access as well.
2. We are configuring the authentication to only allow the Admin Application and Managed Identity to authenticate.  No other resources will be able to pass through its authentication. 
3. We will create a new "UnpublishedData.Read" App Role to the Public API Container App's App Registration.
4. We will assign the Admin App's managed identity to the "UnpublishedData.Read" App Role.
5. When the Admin API calls the Public API, it will request an access token to authenticate itself.  This will include the "UnpublishedData.Read" App Role in a "Role" Claim.  It will then add this access token as a Bearer token to its call to the Public API.
6. The security check in the Changelog endpoint in the Public API is amended to now check that any authenticated calls to it (which at this current point in time can only ever be the Admin App) has been assigned the "UnpublishedData.Read" App Role. If it has, it can see any Changelog, regardless of its published status.
7. For any unauthenticated calls (e.g. from public users), everything should continue to work as per usual.

# Manual infrastructure prerequisites

In order for the above to work, the Public API Container App must be represented in Entra ID with an Enterprise Application security principal.  We create this manually, as ARM and Bicep are unable to interact with Entra ID in this manner (plus we won't have permissions to create the Pre-Prod and Prod Enterprise Applications ourselves).

## Create and configure App Registration for Public API

Go into DfE Platform Identity, Enterprise Applications, "New".

Give it a name and create it, then go to "Expose an API" and create an "UnpublishedData.Read" App Role.

## Assign the "UnpublishedData.Read" App Role to the Admin App's managed identity

There is currently no way of achieving this via the Azure Portal UI.

## Add the new App Registration's Application (Client) ID to our Public API DevOps variable group.

image

### Add the new variables (and dummy values) for each environment

image

### Add override template parameters for each environment's "Validate" and "Deploy" jobs

image


ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetCandidatesControllerTests.cs
```diff
@@ -77,8 +77,8 @@ public async Task NoPermissionsToViewReleaseVersion_Returns403(SecurityClaimType
 
             var authenticatedUser =
                 claimType.HasValue
-                ? AuthenticatedUser(SecurityClaim(claimType.Value))
-                : AuthenticatedUser();
+                    ? DataFixture.AuthenticatedUser().WithClaim(claimType.Value.ToString())
```
ntsim, line 80: Smart use of generators for this :smile:

ntsim (approved):

benoutram:
I had to revert the `apiAppRegistrationClientId` override template parameters that were added to the Pre-prod and Prod environment's Validate and Deploy tasks as they were added a bit too early and caused the infrastructure deploy to fail without the corresponding template.json change being present. We just need to remember to revert this revert as a faffy deploy step when the time comes to deploy this to Pre-prod. 🙂

PR #5128, opened 2024-08-07
EES-5382 - attempt to prevent ARM template deploy from clearing down other policies

duncan-at-hiveit:
:warning: Added the "Do Not Merge" label to this until I'm back, as this could result in some downtime if it doesn't go 100% to plan!

This PR:
- is trying a change to how we add / update access policies in the ARM template's Key Vault deploy, to prevent it from removing any policies that aren't explicitly listed.

This is in an attempt to allow the Bicep templates to deploy any policies that it needs to separately from the main ARM template, by moving the access policies to be a child resource of the main Key Vault deploy.

The alternative is to add all the Public API Key Vault policies into the ARM template, but this isn't ideal, so fingers crossed!

### DIFF
--- a/infrastructure/templates/template.json
+++ b/infrastructure/templates/template.json
@@ -3628,175 +3628,186 @@
         "enabledForDiskEncryption": true,
         "enabledForTemplateDeployment": true,
         "enableSoftDelete": true,
-        "accessPolicies": [
-          {
-            "tenantId": "[subscription().tenantId]",
-            "objectId": "[reference(concat('Microsoft.DataFactory/factories/', variables('dataFactoryName')), '2018-06-01', 'Full').identity.principalId]",
-            "permissions": {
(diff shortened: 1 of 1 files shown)


infrastructure/templates/template.json
```diff
@@ -3628,175 +3628,186 @@
...
+                    "GetIssuers",
+                    "ListIssuers",
+                    "SetIssuers",
+                    "DeleteIssuers"
+                  ]
+                }
+              }
+            ]
+          }
+        },
```
benoutram, line 3810: Appreciate it's not your change but we could turn the two objectId's into template parameters and remove all three of the code comments. Those object id's shouldn't really be hardcoded like that.

If we did this we could also choose parameter names similar to the Entra Id group names which they represent.

benoutram (approved):

PR #5133, opened 2024-08-09
EES-4913 Allow preview access to Public API data sets

benoutram:
This PR adds preview access to Public API data set version data using preview tokens.

A preview token can be set for a request using a HTTP request header `Preview-Token`.

It makes a change to the two authorisation handlers associated with the guards `IUserService.CheckCanQueryDataSetVersion` and `IUserService.CheckCanViewDataSetVersion`. They now ensure a valid preview token is present in the request before succeeding where they would have otherwise failed due to the version being in an inaccessible status.

A preview token is valid for the `DataSetVersion` in the context of the request if there exists a `PreviewToken` entity with matching `Id`, which is not expired, and the token has been issued for the same `DataSetVersion` which must also be in `Draft` status.

### Get a data set version

```http
GET http://localhost:5050/api/v1.0/data-sets/3d369101-f7e9-2f72-a8fa-64e1717cd5eb/versions/1.0
Preview-Token: 3f369101-4f36-8774-a183-04910906b704
```

### Get a data set version's changes

```http
GET http://localhost:5050/api/v1.0/data-sets/3d369101-f7e9-2f72-a8fa-64e1717cd5eb/versions/1.0/changes
Preview-Token: 3f369101-4f36-8774-a183-04910906b704
```

###  Get a data set’s metadata

```http
GET http://localhost:5050/api/v1.0/data-sets/3d369101-f7e9-2f72-a8fa-64e1717cd5eb/meta?dataSetVersion=1.0
Preview-Token: 3f369101-4f36-8774-a183-04910906b704
```

### Query a data Set

Example for (POST) method but preview tokens can also be used to query a data set (GET).

```http
POST http://localhost:5050/api/v1.0/data-sets/3d369101-f7e9-2f72-a8fa-64e1717cd5eb/query?dataSetVersion=1.0
Preview-Token: 3f369101-4f36-8774-a183-04910906b704

{
  "criteria": {
   "filters": {
      "in": ["IfDbB"]
    },
    "geographicLevels": {
      "eq": "LA"
    },
    "locations": {
      "in": [
        {
          "level": "LA",
          "id": "QgWPw"
        }
      ]
    },
    "timePeriods": {
      "gte": {
        "period": "2022/2023",
        "code": "AY"
      }
    }
  },
  "indicators" : ["enrolments"]
}
```


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetVersionsControllerTests.cs
```diff
@@ -574,17 +581,129 @@ public async Task DataSetDoesNotExist_Returns404()
...
+                dataSetVersion: dataSetVersion1.PublicVersion,
+                previewTokenId: dataSetVersion2.PreviewTokens[0].Id);
+
+            response.AssertForbidden();
+        }
+
+        [Theory]
+        [InlineData(DataSetVersionStatus.Processing)]
+        [InlineData(DataSetVersionStatus.Failed)]
+        [InlineData(DataSetVersionStatus.Mapping)]
```
ntsim, line 673: We've also got a `Cancelled` status that we should also cover.

To make it easier to add more statuses in the future, we might want to refactor out some re-usable theory data that can be used across all our different theories.

We have a bunch of different test suites that could re-use this theory data (including the authorization handler tests), so we could add a new `TheoryData` namespace and add the theory data there e.g.

```cs
public static class DataSetVersionStatusTheoryData
{
    public static readonly TheoryData<DataSetVersionStatus> UnavailableStatuses = new()
    {
        DataSetVersionStatus.Processing,
        DataSetVersionStatus.Failed,
        DataSetVersionStatus.Mapping,
        DataSetVersionStatus.Withdrawn,
        DataSetVersionStatus.Cancelled,
    };
}
```

It might be worth considering implementing it using `ClassData` as doing this with  `MemberData` is a bit clunkier. I'd be fine with either approach.

It should be noted that we have a bunch of other test suites outside the scope of this ticket that are also doing similar theory tests. Might be a good idea to tidy these up as well whilst we're here.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Security/RequestHeaderConstants.cs
```diff
@@ -0,0 +1,9 @@
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Security;
+
+public static class RequestHeaderConstants
```
ntsim, line 3: Could rename this to `RequestHeaderNames`? This way we can short the fairly verbose property name to simply `PreviewToken`.

Would be accessed like `RequestHeaderNames.PreviewToken`.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Security/RequestHeaderConstants.cs
```diff
@@ -0,0 +1,9 @@
...
+{
+    /// <summary>
+    /// HTTP request header name for setting a valid Preview Token to access unpublished data.
```
ntsim, line 6: Minor, but we could shorten this down to: `Sets preview tokens for accessing unpublished data.`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Security/RequestHeaderConstants.cs
```diff
@@ -0,0 +1,9 @@
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Security;
```
ntsim, line 1: Might be better to put this in a more generic namespace as we could use this class for other request headers in the future.

`Requests`, `Utils` or a new namespace like `Constants` would probably be acceptable places

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Security/RequestHeaderConstants.cs
benoutram: Good idea, made this change, thanks.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Security/RequestHeaderConstants.cs
benoutram: Good idea, it's now `RequestHeaderNames` in `Public.Data.Api.Constants`, thanks.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetVersionsControllerTests.cs
```diff
@@ -574,17 +581,129 @@ public async Task DataSetDoesNotExist_Returns404()
...
+
+            await TestApp.AddTestData<PublicDataDbContext>(context => context.DataSets.Add(dataSet));
+
+            var (dataSetVersion1, dataSetVersion2) = DataFixture
+                .DefaultDataSetVersion(filters: 1, indicators: 1, locations: 1, timePeriods: 2)
+                .WithStatus(DataSetVersionStatus.Draft)
+                .WithDataSetId(dataSet.Id)
+                .WithPreviewTokens(() => [DataFixture.DefaultPreviewToken()])
+                .Generate(2)
+                .ToTuple2();
```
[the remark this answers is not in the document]
benoutram: Good idea, this certainly saves a lot of lines! I ended up doing this across all the codebase as it was just easier to do that and saves us updating them piecemeal in future.

src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/HttpContextExtensions.cs
```diff
@@ -0,0 +1,29 @@
+#nullable enable
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.AspNetCore.Http;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+
+public static class HttpContextExtensions
+{
+    public static bool TryGetRequestHeader(
```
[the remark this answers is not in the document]
benoutram: I added this extension primarily to do a not-null check on `httpContext`  because it made things look a bit cleaner than adding a check to make sure `httpContextAccessor.HttpContext` is not null prior to calling `httpContextAccessor.HttpContext.Request.TryGetHeader` or assuming it's not null and using `httpContextAccessor.HttpContext!.Request.TryGetHeader`.

I used a not-null check before I added this extension but that code got repeated in both auth handlers.

src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/HttpRequestExtensions.cs
```diff
@@ -62,4 +67,34 @@ public static bool Accepts(this HttpRequest request, bool exact = false, params
...
+        }
+
+        return JsonConvert.DeserializeObject<TJsonType>(requestBody) ??
+               throw new ArgumentException($"Could not deserialize request body to type {typeof(TJsonType)}");
+    }
+
+    public static bool TryGetHeader(
+        this HttpRequest httpRequest,
+        string headerName,
+        [MaybeNullWhen(false)] out string headerValue)
```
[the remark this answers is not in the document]
benoutram: Good point, thanks for the article. I've made this change now, although struggling to get rid of the null-forgiving operator in the usages of this method now.

I added the check `previewTokenValue.Count == 1` to ensure `previewTokenValue` is a `string` rather than `null` but don't think it's super necessary.

```csharp
        return httpContextAccessor.HttpContext.TryGetRequestHeader(
                   RequestHeaderNames.PreviewToken,
                   out var previewTokenValue)
               && previewTokenValue.Count == 1
               && await previewTokenService.ValidatePreviewToken(previewTokenValue!, dataSetVersion.Id);
```

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Security/AuthorizationHandlers/QueryDataSetVersionAuthorizationHandler.cs
```diff
@@ -32,8 +33,9 @@ protected override async Task HandleRequirementAsync(
     private async Task<bool> RequestHasValidPreviewToken(DataSetVersion dataSetVersion)
     {
         return httpContextAccessor.HttpContext.TryGetRequestHeader(
-                   RequestHeaderConstants.PreviewTokenRequestHeaderName,
-                   out var previewToken)
-               && await previewTokenService.ValidatePreviewToken(previewToken, dataSetVersion.Id);
+                   RequestHeaderNames.PreviewToken,
+                   out var previewTokenValue)
+               && previewTokenValue.Count == 1
+               && await previewTokenService.ValidatePreviewToken(previewTokenValue!, dataSetVersion.Id);
```
ntsim, line 39: Don't need the count check? Could do:

```cs
return httpContextAccessor.HttpContext.TryGetRequestHeader(
           RequestHeaderNames.PreviewToken,
           out var previewToken)
       && await previewTokenService.ValidatePreviewToken(previewToken.ToString(), dataSetVersion.Id);
```

The `ToString` will concatenate any preview token header values together and if there are multiple values, this would create an invalid `Guid` so would fail `ValidatePreviewToken`.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Security/AuthorizationHandlers/ViewDataSetVersionAuthorizationHandler.cs
```diff
@@ -40,8 +41,9 @@ protected override async Task HandleRequirementAsync(
     private async Task<bool> RequestHasValidPreviewToken(DataSetVersion dataSetVersion)
     {
         return httpContextAccessor.HttpContext.TryGetRequestHeader(
-                   RequestHeaderConstants.PreviewTokenRequestHeaderName,
-                   out var previewToken)
-               && await previewTokenService.ValidatePreviewToken(previewToken, dataSetVersion.Id);
+                   RequestHeaderNames.PreviewToken,
+                   out var previewTokenValue)
+               && previewTokenValue.Count == 1
+               && await previewTokenService.ValidatePreviewToken(previewTokenValue!, dataSetVersion.Id);
```
ntsim, line 47: Same as other comment - don't need count check

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Security/AuthorizationHandlers/QueryDataSetVersionAuthorizationHandler.cs
benoutram: Good idea, I've made this change in the latest fixup commit. Thanks

ntsim (approved):

PR #5135, opened 2024-08-11
EES-4997 Add API data set changelog to public frontend

ntsim:
This PR adds the the API data set changelog to the public frontend:

image

## Backend view model changes

Changes were made to `filterOptions` and `locationOptions` to group changes by their respective filter or location group i.e.

```jsonc
{
  "filterOptions": [
    {
      "filter": {
        "id": "filter-1",
        "label": "Filter 1"
      },
      "options": [
        { "currentState": { ... } }
      ]
    }
  ],
  "locationOptions": [
    {
      "level": {
        "code": "REG",
        "label": "Regional"
      },
      "options": [
        { "currentState": { ... } }
      ]
    }
  ]
}
```

This allows us to provide additional labelling for each group, which is necessary to display each group in the rendered changelog.

Each group will be displayed like the following:

- Deleted Filter 1 filter options
- Updated Filter 1 filter options
- Added Regional location options

## Relevant changes

- Refactored out `getLocationCodeEntries` util which is reused for creating location code labels in the admin's API data set location mapping page.
- Added `intersperse` util that can be used for adding extra items between array items e.g. separators between React elements.
- Removed `DataSetFilePageNav` in favour of common's `PageNav` component.


src/explore-education-statistics-common/src/modules/data-catalogue/components/ChangeList.tsx
```diff
@@ -0,0 +1,94 @@
...
+}: ChangeListProps<TState>) {
+  if (!changes.length) {
+    return null;
+  }
+
+  const { additions, deletions, updates } = groupByChangeType(changes);
+
+  const metaTypeLabel = customMetaTypeLabel ?? metaTypeLabels[metaType];
+
+  return (
```
amyb-hiveit, line 37: This might be a easier to visually parse if was grouped by the meta type, it currently feels a bit unordered if you end up with something like this:
Screenshot 2024-08-12 131549

Something like this might work better:

Screenshot 2024-08-12 132036

src/explore-education-statistics-common/src/modules/data-catalogue/components/ChangeItem.tsx
```diff
@@ -0,0 +1,84 @@
...
+  TimePeriodOption,
+} from '@common/services/types/apiDataSetMeta';
+import React from 'react';
+
+export interface ChangeItemProps<TState> {
+  change: Change<TState>;
+  metaType: keyof ChangeSet;
+}
+
+export default function ChangeItem<TState>({
```
amyb-hiveit, line 25: It'd be good to have at least some basic tests for this and the other untested new components.

src/explore-education-statistics-common/src/modules/data-catalogue/utils/groupByChangeType.ts
```diff
@@ -0,0 +1,33 @@
+import { Change } from '@common/services/types/apiDataSetChanges';
+
+export interface ChangesByType<TChange extends Change<unknown>> {
+  additions: TChange[];
+  deletions: TChange[];
+  updates: TChange[];
+}
+
+export default function groupByChangeType<TChange extends Change<unknown>>(
```
amyb-hiveit, line 9: Could you add some tests for this?

amyb-hiveit (commented):
Just a few queries, otherwise looks good.

src/explore-education-statistics-common/src/utils/array/intersperse.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Intersperse a separator item between each item in an array.
+ */
+export default function intersperse<TItem, TSeparator>(
```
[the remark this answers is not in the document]
ntsim: Not sure I'd agree there. Dictionary.com defines it as:

> to scatter here and there or place at intervals among other things:
_to intersperse flowers among shrubs._
>
> to diversify with something placed or scattered at intervals:
_to intersperse a dull speech with interesting anecdotes._

Merriam-Webster's is:

> 1: to insert at intervals among other things
_interspersing drawings throughout the text_
>
> 2: to place something at intervals in or among
_intersperse a book with pictures_

The definition is potentially up for debate here, but regardless, I mostly picked this name because this was the name of the util proposed for Lodash for this exact purpose: https://github.com/lodash/lodash/issues/2339

I'm not too keen on the proposed alternative so I'd prefer to stick with `intersperse` as it's nice, succinct and may be in Lodash one day.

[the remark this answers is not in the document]
amyb-hiveit: Fair enough, I read it more as 'to scatter here and there' rather than at regular intervals so it confused me a bit, but being consistent with a proposed lodash util makes sense.

src/explore-education-statistics-common/src/modules/data-catalogue/components/ChangeItem.tsx
ntsim: I've just moved this component into `ChangeList.tsx` and added tests for this.

However, I've not added any further tests around the change label components as these are pretty thoroughly tested by the `ChangeSection` suite. There isn't really any value in testing the label components individually as they're not meant to be used in isolation, and are only in  separate files for tidiness.

amyb-hiveit (approved):

PR #5137, opened 2024-08-12
EES-4879 Adding a blanket 30s timeout across all Public API endpoints

jack-hive:
This PR adds a 30 second `504` timeout to all Public API endpoints.

We also have a change to document these `504` timeout responses in the swagger documentation across all endpoints.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Options/RequestTimeoutOptions.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Options/RequestTimeoutOptions.cs
@@ -2,7 +2,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Options;
 
 public class RequestTimeoutOptions
 {
-    public const string Section = "RequestTimeouts";
+    public const string Section = "RequestTimeout";
 
-    public int? RequestTimeoutMilliseconds { get; init; }
+    public int? TimeoutMilliseconds { get; init; }
 }
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Options/RequestTimeoutOptions.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Options/RequestTimeoutOptions.cs
@@ -1,0 +1,8 @@
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Options;
+
+public class RequestTimeoutOptions
+{
+    public const string Section = "RequestTimeouts";
+
+    public int? RequestTimeoutMilliseconds { get; init; }
+}
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
@@ -1,9 +1,6 @@
-using System.Diagnostics.CodeAnalysis;
-using System.Text.Json.Serialization;
 using AngleSharp.Io;
 using Dapper;
 using FluentValidation;
-using GovUk.Education.ExploreEducationStatistics.Common.Cancellation;
 using GovUk.Education.ExploreEducationStatistics.Common.Config;
 using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.ModelBinding;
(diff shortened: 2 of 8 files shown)


duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
```diff
@@ -63,14 +63,23 @@ public void ConfigureServices(IServiceCollection services)
 
         // MVC
 
+        services.AddRequestTimeouts(options =>
+        {
+            options.DefaultPolicy =
+                new RequestTimeoutPolicy
+                {
+                    Timeout = TimeSpan.FromMilliseconds(30000),
```
duncan-at-hiveit, line 71: I don't knwo if it's worth making this configurable (with a 30s fallback), and set to something like 500ms in our integration tests?  That way, we can register a test controller that delays for 1s and then we can test the response.

jack-hive: Good idea, I'll make that change!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/appsettings.Development.json
```diff
@@ -19,5 +19,6 @@
   },
   "DataFiles": {
     "BasePath": "data/public-api-data"
-  }
+  },
```
duncan-at-hiveit, line 22: I guess with the fallback, you don't really need this value here.

duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/appsettings.Development.json
jack-hive: No that's true! I thought I'd leave it there for discoverability. But I'll remove it :)

duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
```diff
@@ -62,14 +63,23 @@ public void ConfigureServices(IServiceCollection services)
 
         // MVC
 
+        services.AddRequestTimeouts(options =>
+        {
+            options.DefaultPolicy =
+                new RequestTimeoutPolicy
+                {
+                    Timeout = TimeSpan.FromMilliseconds(configuration.GetValue<int?>("RequestTimeoutMilliseconds") ?? 30000),
```
ntsim, line 71: Might be good for us to extract out an `AppSettingsOptions` class to house this configuration. We already do this in other parts of the service and would give some type-safety over this.

This could then be re-used in the tests as well.

jack-hive: I was thinking of doing this, but then thought since it's just a one-off setting and I couldn't envisage us nesting any other options beneath the 'timeout' section as such, I thought I'd just simplify it by removing the need for the extra class. But I agree, it might be slightly better to do this. I've made a change :)

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Options/RequestTimeoutOptions.cs
```diff
@@ -0,0 +1,8 @@
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Options;
+
+public class RequestTimeoutOptions
+{
+    public const string Section = "RequestTimeouts";
+
+    public int? RequestTimeoutMilliseconds { get; init; }
```
ntsim, line 7: Given we've opted to put this in a `RequestTimeout` section, we could simplify this key to just `TimeoutMilliseconds`.

I was also wondering if you could just put the default value (30000) in here as well? That way we don't need to put the default in `Startup`. Not entirely sure if this works as we'd expect or not though (would suggest checking).

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Options/RequestTimeoutOptions.cs
```diff
@@ -0,0 +1,8 @@
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Options;
+
+public class RequestTimeoutOptions
+{
+    public const string Section = "RequestTimeouts";
```
ntsim, line 5: This should be `RequestTimeout` to align with the class

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Options/RequestTimeoutOptions.cs
jack-hive: Renamed :)

I tested the default, but it doesn't work unfortunately :(

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Options/RequestTimeoutOptions.cs
jack-hive: done :)

ntsim (commented):
There are currently some build errors that need to be fixed

jack-hive:
> There are currently some build errors that need to be fixed

I have no idea how this happened. I'm pretty sure I used F2 to rename the variable **everywhere!**. But my Visual Studio has been playing up a lot lately with detecting renames. I might have to get into the habit of running tests again even after something as simple as a rename :P

ntsim (approved):

PR #5139, opened 2024-08-13
EES-5411 add option to show all axis labels on vertical bar charts

amyb-hiveit:
Adds a 'Show all' option to the tick display type on the major axis configuration. When selected this will override Recharts' default behaviour of automatically showing / hiding axis labels depending on space and instead set a width on the labels that will cause them to wrap.

Also, I've added some styling to make the chart in the chart builder preview be the same size as when they're embedded in releases (EES-5410).

Some caveats:
- only on vertical bar charts (I looked at also adding it to line charts but there were issues with the last label being cut off)
- only on desktop, on mobile it will revert to the previous behaviour to prevent the labels from overlapping
- analysts might have to adjust the axis size to make sure the wrapped labels don't get cut off at the bottom.

Before - automatic tick display:
Screenshot 2024-08-13 131319

After - show all ticks:
Screenshot 2024-08-13 131210


src/explore-education-statistics-common/src/modules/charts/components/ChartRenderer.tsx
```diff
@@ -52,11 +52,41 @@ function ChartRenderer({
...
     switch (props.type) {
       case 'line':
-        return <LineChartBlock {...props} />;
+        return (
+          <>
+            {props.axes === undefined ||
+            props.axes.major === undefined ||
+            props.axes.minor === undefined ? (
+              <div>Unable to render chart, chart incorrectly configured</div>
+            ) : (
```
ntsim, line 61: Thinking about it, this type of error handling / validation isn't great as it's not particularly robust and doesn't actually allow us to track these types of issues properly in Application Insights.

Would suggest that we just ditch this stuff completely and just let any errors bubble up so that they can be logged at the nearest error boundary.

Same applies for other cases below.

src/explore-education-statistics-common/src/modules/charts/components/__tests__/ChartRenderer.test.tsx
```diff
@@ -75,4 +79,42 @@ describe('ChartRenderer', () => {
...
+      legend: undefined as unknown as LegendConfiguration,
+    };
+
+    render(<ChartRenderer id="test-id" {...invalidTestChartRenderer} />);
+    expect(
+      screen.getByText('Unable to render chart, chart incorrectly configured'),
+    ).toBeInTheDocument();
+  });
+
+  test('shows an error message with bad map data ', () => {
```
ntsim, line 100: Similar to earlier comment - this error handling isn't very robust and prevents us from tracking these issues in Application Insights.

Would suggest that in these situations, we just allow the `ChartRenderer` to throw an error with an appropriate message. This would then be caught and logged by the nearest error boundary.

src/explore-education-statistics-common/src/modules/charts/components/__tests__/ChartRenderer.test.tsx
```diff
@@ -75,4 +79,42 @@ describe('ChartRenderer', () => {
...
   });
+
+  test('shows an error message with bad chart data ', () => {
+    const invalidTestChartRenderer: ChartRendererProps = {
+      type: 'verticalbar',
+      meta: testFullTableMeta.subjectMeta,
+      data: testFullTableMeta.results,
+      alt: '',
+      height: 100,
+      axes: undefined as unknown as LineChartProps['axes'],
```
ntsim, line 90: Minor, but for any of these type assertions, we could just simplify them to `undefined as never`. 

Same applies elsewhere in this PR.

ntsim (changes requested):

src/explore-education-statistics-common/src/modules/charts/util/domainTicks.ts
```diff
@@ -177,15 +177,13 @@ export function getMinorAxisDomainTicks(
 }
 
 export function getMajorAxisDomainTicks(
-  chartData: ChartData[],
+  majorAxisCategories: string[],
```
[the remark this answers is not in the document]
amyb-hiveit: It was because I need to know the length of `majorAxisCategories` for the tick width calculation, so this was just to avoid doing `chartData.map(({ name }) => name)` twice.

src/explore-education-statistics-common/src/modules/charts/components/ChartRenderer.tsx
amyb-hiveit: Yep, makes sense. I've removed these and the tests for it.

src/explore-education-statistics-common/src/modules/charts/components/__tests__/ChartRenderer.test.tsx
amyb-hiveit: These tests are removed now.

ntsim (approved):

PR #5145, opened 2024-08-14
EES-4859 api data set subscriptions

amyb-hiveit:
Adds the front end for API data set subscriptions. It's basically the same as for publications but with better error handling.

I've made some small changes to the publications version to replace `SubscriptionStatusMessage` with a more generic `Panel` element.


src/explore-education-statistics-frontend/src/modules/api-subscriptions/NewSubscriptionPage.tsx
```diff
@@ -0,0 +1,139 @@
...
+  dataSetFile: DataSetFile;
+}
+
+const NewSubscriptionPage: NextPage<Props> = ({ dataSetFile }) => {
+  const { id, title } = dataSetFile;
+
+  const handleFormSubmit = async ({ email }: FormValues) => {
+    await apiNotificationService.requestPendingSubscription({
+      email,
+      dataSetId: id,
```
benoutram, line 47: Is this the correct `id` here? I'm wondering if it should be `dataSetFile.api.id`?

It might be worth renaming `id` to whatever type of id it is.

Same comment applies to the de-referencing of `id` from `dataSetFile` in the other pages.

src/explore-education-statistics-frontend/src/services/apiNotificationService.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const apiNotificationService = {
+  requestPendingSubscription(query: SubscribeQuery): Promise<ApiSubscription> {
+    return notificationApi.post(
+      '/public-api/request-pending-subscription',
+      query,
+    );
+  },
+  async confirmPendingSubscription(
+    id: string,
```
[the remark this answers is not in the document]
amyb-hiveit: Yep good point, I've changed it to consistently use `dataSetId` here and in the components.

src/explore-education-statistics-frontend/src/services/apiNotificationService.ts
```diff
@@ -0,0 +1,34 @@
...
+    token: string,
+  ): Promise<ApiSubscription | undefined> {
```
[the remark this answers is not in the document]
amyb-hiveit: I think that was a copy n paste error from the publication notification service, I've sorted it out now.

src/explore-education-statistics-frontend/src/modules/api-subscriptions/NewSubscriptionPage.tsx
amyb-hiveit: I think it is meant to be `dataSetId.id` as that's the id we need to fetch the data set details on the confirmation page, if I try with `dataSetFile.api.id` the request works but the link to the confirmation page in the email has the api id in which can't then be used to fetch the details.

benoutram: I've just checked with the backend code and you are correct, thanks. There are that many different types of id's relating to files and data sets that it's getting very confusing! 😆 

On the backend the Notifier knows nothing about the Public API data model, so although the parameter `dataSetId` sounds like it would be a Public API data set id, like you say, it needs to be the same type of id that's used to link back to the data catalogue page. On the backend this is a `File.DataSetFileId`! 😆

benoutram (approved):

PR #5150, opened 2024-08-15
EES-5409 Added UI tests to create a release with two small API Datasets

saicharan2789:
This PR introduces comprehensive UI tests to verify that the system behaves as expected when managing public API datasets while publishing releases. It focusses on meeting below acceptance criteria - 

- A release cannot be published if it has a failed or processing api dataset associated with it.
- A release cannot be published if it has a draft api data set that requires action (ie it has a draft api dataset version where the mappings have not been completed).
- User is able to create a new API Dataset
- A user may see the following statuses for an API Data set:

> 1. “Processing” - the api data set is being processed/imported
> 2. “Failed” - the api data set has failed to import
> 3. “Ready” - the api data set was successfully imported


saicharan2789:
image

mmyoungman (changes requested):

tests/robot-tests/tests/public_api/public_api_restricted.robot
```diff
@@ -0,0 +1,300 @@
...
+    user waits until element contains    css:#release-notes li:nth-of-type(1) time    ${date}
+    user waits until element contains    css:#release-notes li:nth-of-type(1) p    Test release note two
+
+Validate checklist error while API data set is still processing or being failed
+    user edits release status
+    user checks checklist errors contains
+    ...    1 issue that must be resolved before this release can be published.
+    user checks checklist errors contains either link
+    ...    All public API data set processing must be completed
+    ...    All failed public API data sets must be retried or removed
```
mmyoungman, line 177: I'm happy to leave this - I don't think there is any case here where we could miss a failure? - but I'd expect us to be checking for one then the other. Maybe the data set is processed so quickly and that's why you've done it this way?

mmyoungman (approved):

tests/robot-tests/tests/public_api/public_api_restricted.robot
```diff
@@ -0,0 +1,300 @@
...
+    user waits until element contains    css:dl[data-testid="draft-version-summary"] > div:nth-of-type(2) > dt + dd     Action required    %{WAIT_LONG}
+
+    ${mapping_status}=    get text    css:dl[data-testid="draft-version-summary"] > div:nth-of-type(2) > dt + dd
+    should be equal as strings    ${mapping_status}    Action required
+
+Add headline text block to Content page
+    user navigates to content page    ${PUBLICATION_NAME}
+    user adds headlines text block
+    user adds content to headlines text block    Headline text block text
+
```
mmyoungman, line 293: Do we need to create this headline text block here?

tests/robot-tests/tests/public_api/public_api_restricted.robot
saicharan2789: When processing large data files, two specific errors can occur depending on the speed of processing. Sometimes the files are processed quickly, while other times the process is slower. Also, there's an active [bug ticket](https://dfedigital.atlassian.net/browse/EES-5420) because large data files are consistently getting failing while processing. Given this situation, I decided to add some checks that can handle either outcome.

tests/robot-tests/tests/public_api/public_api_restricted.robot
saicharan2789: Its not mandatory required to create this headline text block, but I have added this so that I can only focus on specific errors which I am looking for. if I chose to leave this, then I would be tempted to add headline specific text block checklist errors.

tests/robot-tests/tests/public_api/public_api_restricted.robot
mmyoungman: Cool. If we will want to change this later when the devs have changed / fixed stuff, then you could add a comment here - but I'll leave it to you to decide what's best here.

saicharan2789: Added comments accordingly

PR #5152, opened 2024-08-15
EES-5419 import new data set version

amyb-hiveit:
Adds the ability to import a new API data set version once the mappings have been completed. The notification banner goes through the following stages: 

Before import:
Screenshot 2024-08-15 163939

While importing:
Screenshot 2024-08-15 163101

On successful import - changes to green and has `role="alert":
Screenshot 2024-08-15 162408

Coming back to the page after import - back to blue and no alert as it's not the result of an action you've just taken:  
Screenshot 2024-08-15 162327

After discussion with Cam I've removed the button to re-import the version and updated the wording.


src/explore-education-statistics-admin/src/pages/release/data/ReleaseApiDataSetDetailsPage.tsx
```diff
@@ -157,16 +183,96 @@ export default function ReleaseApiDataSetDetailsPage() {
...
+    ) {
+      return (
+        <ErrorSummary
+          errors={[
+            {
+              id: 'draft-version-summary',
+              message: 'Data set version import failed',
+            },
+          ]}
+        />
```
ntsim, line 271: Do we need to use error summary for this? We're not on a form, so it's a bit weird and inconsistent given we've already been using the notification banner for the other states.

Could just add an error variant of `NotificationBanner`?

src/explore-education-statistics-admin/src/pages/release/data/__tests__/ReleaseApiDataSetDetailsPage.test.tsx
```diff
@@ -396,10 +399,209 @@ describe('ReleaseApiDataSetDetailsPage', () => {
     ).not.toBeInTheDocument();
   });
 
+  describe('importing', () => {
```
ntsim, line 402: If we go with the 'finalise' terminology, would make any relevant text changes to this test suite here and below.

src/explore-education-statistics-admin/src/pages/release/data/ReleaseApiDataSetDetailsPage.tsx
```diff
@@ -157,16 +183,96 @@ export default function ReleaseApiDataSetDetailsPage() {
...
+      );
     }
-  }
+
+    if (dataSet?.draftVersion?.status === 'Draft') {
+      return (
+        <NotificationBanner
+          fullWidthContent
+          heading="This new API data set version has now been mapped and imported ready for final publication"
+          role={importingStatus === 'imported' ? 'alert' : undefined}
```
ntsim, line 238: Not entirely sure, but changing the `alert` role like this might not trigger the banner to be read out. Think the alert only works if the browser is aware of the element (with `role="alert"`) before its text is updated. From the [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role): 

> As with all other [live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions), alerts will only be announced when the content of the element with role="alert" is updated. Make sure that the element with the role is present in the page's markup first - this will "prime" the browser and screen reader to keep watching the element for changes. After this, any changes to the content will be announced. Do not try to dynamically add/generate an element with role="alert" that is already populated with the alert message you want announced - this generally does not lead to an announcement, as it is not a content change.

I'm not sure if the notification banner element ends up being unmounted from the DOM at any point, so the browser may not detect the element's contents change changing. 

Probably worth testing with a screen reader.

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/release/data/ReleaseApiDataSetDetailsPage.tsx
```diff
@@ -157,16 +183,96 @@ export default function ReleaseApiDataSetDetailsPage() {
...
+    dataSet?.draftVersion?.mappingStatus &&
+    dataSet.draftVersion.mappingStatus.filtersComplete &&
+    dataSet.draftVersion.mappingStatus.locationsComplete;
+
+  const showDraftVersionTasks =
+    importingStatus !== 'importing' &&
+    (dataSet?.draftVersion?.status === 'Draft' ||
+      dataSet?.draftVersion?.status === 'Mapping');
+
+  const getImportBanner = () => {
```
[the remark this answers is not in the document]
amyb-hiveit: I've moved this to its own component now.

src/explore-education-statistics-admin/src/pages/release/data/ReleaseApiDataSetDetailsPage.tsx
```diff
@@ -157,16 +183,96 @@ export default function ReleaseApiDataSetDetailsPage() {
...
+          <p>This process can take a few minutes</p>
+        </NotificationBanner>
+      );
+    }
+
+    if (dataSet?.draftVersion?.status === 'Mapping') {
+      return (
+        <NotificationBanner
+          fullWidthContent
+          heading="This new API data set version has been mapped and is ready to be imported"
```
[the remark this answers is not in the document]
amyb-hiveit: Changed to 'finalise' consistently throughout now.

src/explore-education-statistics-admin/src/pages/release/data/ReleaseApiDataSetDetailsPage.tsx
amyb-hiveit: Changed to use `NotificationBanner` now.

src/explore-education-statistics-admin/src/pages/release/data/ReleaseApiDataSetDetailsPage.tsx
amyb-hiveit: It does work here because the role isn't changed dynamically - it only has the role `alert` when the import process has just finished (`finalisingStatus === 'finalised'`), when you visit the page again there isn't a `finalisingStatus` so the role isn't applied.

ntsim: Okay, will trust that you're right about this!

ntsim (approved):

PR #5154, opened 2024-08-16
EES-5400: Create new endpoint to get GeoJSON for implementing multiple boundary files

tomjonesdev:
This PR adds a new endpoint which returns location data for a data block, based on the supplied boundary level. Once implemented, this will allow the frontend to have data blocks which reference numerous boundary levels (i.e. if the data spans a number of years, and the boundaries have changed over time), and then call the endpoint to retrieve the appropriate location data.

Due to the relatively large volume of data returned, this data is also cached in blob storage for efficiency/performance.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/BauCacheController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/BauCacheController.cs
@@ -1,10 +1,4 @@
 #nullable enable
-using System;
-using System.Collections.Generic;
-using System.ComponentModel.DataAnnotations;
-using System.Linq;
-using System.Text.RegularExpressions;
-using System.Threading.Tasks;
 using GovUk.Education.ExploreEducationStatistics.Common;
 using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Services;
@@ -14,6 +8,12 @@ using GovUk.Education.ExploreEducationStatistics.Common.Validators;
 using GovUk.Education.ExploreEducationStatistics.Content.Services.Interfaces.Cache;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Models.GlobalRoles;
 using static GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces.IBlobStorageService;
 
@@ -29,6 +29,11 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api.Bau
         private readonly IGlossaryCacheService _glossaryCacheService;
         private readonly IMethodologyCacheService _methodologyCacheService;
         private readonly IPublicationCacheService _publicationCacheService;
+
+        private const string GuidPattern = "[\\da-zA-Z]{8}-([\\da-zA-Z]{4}-){3}[\\da-zA-Z]{12}";
+        private const string ReleasePeriodPattern = "[0-9]{4}(-[^/]+)?\\";
+        private const string BoundaryLevelIdPattern = "\\d{1,3}";
+        private const string WildcardDirectoryNamePattern = "[^/]+";
 
         public BauCacheController(
             IPrivateBlobStorageService privateBlobStorageService,
@@ -55,7 +60,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api.Bau
                     BlobContainers.PrivateContent,
                     options: new DeleteBlobsOptions
                     {
-                        IncludeRegex = new Regex($"^releases/[^/]+/({pathString})/")
+                        IncludeRegex = new Regex($"^releases/{WildcardDirectoryNamePattern}/({pathString})/")
                     }
                 );
             }
@@ -112,7 +117,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api.Bau
                     BlobContainers.PublicContent,
                     options: new DeleteBlobsOptions
                     {
-                        IncludeRegex = new Regex($"^publications/[^/]+/releases/[^/]+/({pathString})/")
+                        IncludeRegex = new Regex($"^publications/{WildcardDirectoryNamePattern}/releases/{WildcardDirectoryNamePattern}/({pathString})/")
                     }
                 );
             }
@@ -163,7 +168,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api.Bau
                 BlobContainers.PublicContent,
                 options: new DeleteBlobsOptions
                 {
-                    IncludeRegex = new Regex($"^publications/[^/]+/{publicationJsonFilenameRegex}$")
+                    IncludeRegex = new Regex($"^publications/{WildcardDirectoryNamePattern}/{publicationJsonFilenameRegex}$")
                 });
 
             return NoContent();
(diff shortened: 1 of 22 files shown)


src/GovUk.Education.ExploreEducationStatistics.Data.Api/Controllers/TableBuilderController.cs
```diff
@@ -121,6 +121,25 @@ public async Task<ActionResult<TableBuilderResultViewModel>> QueryForTableBuilde
             return actionResult;
         }
 
+        [HttpGet("tablebuilder/release/data-block/{dataBlockParentId:guid}")]
+        public async Task<ActionResult<Dictionary<string, List<LocationAttributeViewModel>>>> QueryForTableBuilderWithGeoJsonResult(
```
duncan-at-hiveit, line 127: Does this being separate from the main query that brings back table data mean that we're having to issue multiple table queries per data block?  One for the table data, another for each boundary level as it's selected?

Reason I ask is that this could potentially become expensive on large tables, so we'll have to consider performance if I've got the right end of the stick!

duncan-at-hiveit: As an alternative, if this is issued as another query after the 1st query for table data (where we'd then know which Locations will appear in the results), could we instead post an array of Location Ids that we want boundary data for here rather than needing to issue a 2nd table query?

duncan-at-hiveit, line 127: Out of interest, do we not need an equivalent method in Admin's TableBuilderController?  Also, I know that we cache DataBlock results in Admin (see Admin's TableBuilderController):

```
        [BlobCache(typeof(DataBlockTableResultCacheKey))]
        private async Task<Either<ActionResult, TableBuilderResultViewModel>> GetReleaseDataBlockResults(
            DataBlockVersion dataBlockVersion,
            long? boundaryLevelId,
        {
            return await _userService
                .CheckCanViewReleaseVersion(dataBlockVersion.ReleaseVersion)
                .OnSuccess(_ => _tableBuilderService.Query(releaseVersionId: dataBlockVersion.ReleaseVersionId,
                    dataBlockVersion.Query,
                    boundaryLevelId,
                    cancellationToken));
        }
```

I wonder if we need to look at caching in Admin too?

tomjonesdev: I queried this with @amyb-hiveit, who implied that we could possibly use the same endpoint for both.

tomjonesdev: Alternative sounds good to me, but the ticket specified endpoint arguments of "boundary level id and data block id" - I don't recall if it was considered as an option and thrown out, or missed as a potential solution - I'll see if anyone can remember, or if they agree it should be amended to how you suggest.

mmyoungman: I think you will still need another endpoint creating in the backend, because (as I think you're already aware) while the frontend sometimes uses `dataApi` to make Admin API calls, the backend explicitly has separate endpoints.

So `QueryForTableBuilderResults` in the Data API's `TableBuilderController` has a twin `QueryForDataBlock` in Admin's `TableBuilderController`. The only difference between the two, as far as I'm aware, is the Admin's prefixes `data/` to the endpoint path and has a permissions check.

The frontend utilises both of these endpoints via a sole method, `tableBuilderService.ts`'s `getDataBlockTableData` - basically `dataApi` (and confusingly) when called from admin, actually calls admin. This is to allow the component `DataBlockTabs` (and possibly others?) to be used in `src/ees-admin` and `src/ees-frontend`.

amyb-hiveit: I think when this was discussed the preference was for a GET request so it could be cached. We wouldn't necessarily be able to fit all the required ids in a GET url so it was decided that using the data block id and letting the back end figure out what geojson it needed to return would be best.

src/GovUk.Education.ExploreEducationStatistics.Common/Services/FileStoragePathUtils.cs
```diff
@@ -67,6 +67,15 @@ public static string PublicContentDataBlockPath(
             return $"{PublicContentDataBlockParentPath(publicationSlug, releaseSlug)}/{dataBlockId}.json";
         }
 
+        public static string PublicContentDataBlockLocationsPath(
+            string publicationSlug,
+            string releaseSlug,
+            Guid dataBlockId,
+            long boundaryLevelId)
+        {
+            return $"{PublicContentDataBlockParentPath(publicationSlug, releaseSlug)}/{dataBlockId}/{boundaryLevelId}-locations.json";
```
mmyoungman, line 76: I think it might be worth asking the rest of their devs on where this should live, because I can imagine people have opinions about it, and is quick and easy to change.

It would be good to have a scheme where anyone looking at the files in blob storage without context can immediately recognise what these cache files are for. So my suggestion would be to change it to `$"{PublicContentDataBlockParentPath(publicationSlug, releaseSlug)}/{dataBlockId}-boundary-levels/{dataBlockId}-{boundaryLevelId}.json";`

This would also be easier to select by regex if we wanted to create/update an endpoint to remove them as suggested in another comment.

src/GovUk.Education.ExploreEducationStatistics.Common/Services/FileStoragePathUtils.cs
mmyoungman, line 1: It might be worth adding a new `BauCacheController` endpoint or updating an existing endpoint so that we can remove cached geojson without removing all data blocks. Although happy for you to not do this if you don't think it's worthwhile.

Right now, if we had to remove all cached geojson, we would have to do it by removing all data-block directories via `BauCacheController#ClearPublicCacheReleases`. But I think that's probably fine. I cannot imagine any reason why that would cause us a problem beyond the unnecessary generating of all data block caches again.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/Controllers/TableBuilderController.cs
```diff
@@ -121,6 +123,25 @@ public async Task<ActionResult<TableBuilderResultViewModel>> QueryForTableBuilde
             return actionResult;
         }
 
+        [HttpGet("tablebuilder/release/data-block/{dataBlockParentId:guid}")]
```
mmyoungman, line 126: To make this convenient for the frontend code, this endpoint's path needs to align with its twin, minus the `data/`: `[HttpGet("data/tablebuilder/release/{releaseVersionId:guid}/data-block/{dataBlockParentId:guid}/geojson")]`

We need the releaseId for the Admin endpoint, so we'll also include it in the Data API endpoint path even though it isn't used. There is precedent for this - see the comment for the Data API's `TableBuilderController#QueryForTableBuilderResult`.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/TableBuilderService.cs
```diff
@@ -110,6 +116,33 @@ public async Task<Either<ActionResult, TableBuilderResultViewModel>> Query(
                 });
         }
 
+        public async Task<Either<ActionResult, Dictionary<string, List<LocationAttributeViewModel>>>> Query(
+            Guid releaseVersionId,
+            FullTableQuery query,
+            long boundaryLevelId,
+            CancellationToken cancellationToken = default)
```
mmyoungman, line 123: I'd be inclined to rename this method to distinguish it from the other `Query` method - maybe `QueryForBoundaryLevel` - to help people who spot the methods are named the same but not catch that the return types are different.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/TableBuilderService.cs
```diff
@@ -68,14 +73,15 @@ public TableBuilderService(
...
         public async Task<Either<ActionResult, TableBuilderResultViewModel>> Query(
             FullTableQuery query,
             CancellationToken cancellationToken = default)
         {
             return await FindLatestPublishedReleaseVersionId(query.SubjectId)
-                .OnSuccess(releaseVersionId => Query(releaseVersionId, query, null, cancellationToken));
+                .OnSuccess(releaseVersionId => Query(releaseVersionId, query, boundaryLevelId: null, cancellationToken));
         }
 
         public async Task<Either<ActionResult, TableBuilderResultViewModel>> Query(
```
mmyoungman, line 87: I'd be inclined to add a comment to the `boundaryLevelId` argument in this method - something like
``` long? boundaryLevelId, // remove in EES-XXXX```
in case that sticks around for longer than expected.

**Review by mmyoungman — CHANGES_REQUESTED** [review]

File: src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Bau/BauCacheController.cs [hunk 6]
```diff
@@ -30,6 +30,10 @@ public class BauCacheController : ControllerBase
         private readonly IMethodologyCacheService _methodologyCacheService;
         private readonly IPublicationCacheService _publicationCacheService;
 
+        private const string GuidPattern = "[\\da-zA-Z]{8}-([\\da-zA-Z]{4}-){3}[\\da-zA-Z]{12}";
+        private const string ReleasePeriodPattern = "\\d{4}-\\d{2}";
```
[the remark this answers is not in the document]
tomjonesdev: It looks like one of the other methods (`ClearPublicCacheReleaseJson`) has made use of a pattern to accommodate for those variations, so I'll switch it that to see if it works as intended.

mmyoungman (approved):

PR #5156, opened 2024-08-16
EES-5421 Error if ListSubjects finds different number of subjects in DBs

mmyoungman:
This PR was created in response to the bug found in EES-5404. We believe - but aren't sure, as we struggled to recreate this issue - that this is due data not copying across to the replica database quick enough, so if someone used the table tool shortly after a release was published, under the right conditions, the Stats DB would return no subjects, and this result would be cached, meaning the service would continue to treat a release that did have subjects as if it had no subjects, in the table tool.

This PR implements a preventative measure of cross-checking the number of subjects in the Stats DB with the number of subjects in the Content DB.


src/GovUk.Education.ExploreEducationStatistics.Data.Services/ReleaseService.cs
```diff
@@ -73,9 +74,22 @@ private async Task<List<SubjectViewModel>> GetSubjects(Guid releaseVersionId, Li
 
             var releaseFiles = await QueryReleaseDataFiles(releaseVersionId)
                 .Where(rf => rf.File.SubjectId.HasValue
+                             && rf.ReleaseVersionId == releaseVersionId
                              && subjectsToInclude.Contains(rf.File.SubjectId.Value))
                 .ToListAsync();
 
+            // This check was added to protect against EES-5404
+            if (releaseSubjects.Count != releaseFiles.Count)
```
benoutram, line 82: This might be ok, but I was thinking that we'd do this check earlier than this, after assigning `releaseSubjects`.

The `subjectsToInclude` parameter is a list of subject id's found in the content database. See how it gets worked out in the method `GetPublishedSubjectIds`.

So based on this list of subject id's that we expect to exist from data in the content database, we get the subjects from statistics database and assign `releaseSubjects`.

If any subject in that list is not found we'd return an error.

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/ReleaseServiceTests.cs
```diff
@@ -243,6 +244,139 @@ await statisticsDbContext.IndicatorGroup.AddRangeAsync(
...
+            };
+
+            var subject1IndicatorGroup1 = new IndicatorGroup
+            {
+                Subject = releaseSubject1.Subject,
+                Label = "subject 1 indicator group 1",
+                Indicators = new List<Indicator>
+                {
+                    new() { Label = "subject 1 indicator group 1 indicator 1" },
+                }
```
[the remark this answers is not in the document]
mmyoungman: This code got removed

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ReleaseService.cs
mmyoungman: I've made the change. 

I will just highlight that this means that if someone changes the way `subjectsToInclude` is fetched for some reason, then  this check might be meaningless. But I don't see how or why that change would happen, and if it did it would be in the far future, so I think it's fine. I was a little tempted to change the variable name to make it explicit, but something like`contentDbSubjectsToInclude` seems a bit extreme.

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/ReleaseServiceTests.cs
```diff
@@ -243,6 +244,139 @@ await statisticsDbContext.IndicatorGroup.AddRangeAsync(
...
+            {
+                File = releaseFile1.File,
+                Status = DataImportStatus.COMPLETE
+            };
+
+            var import2 = new DataImport
+            {
+                File = releaseFile2.File,
+                Status = DataImportStatus.COMPLETE
+            };
```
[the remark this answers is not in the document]
mmyoungman: Because I copied the test data from another test. I've changed it to use the test data generator now, although not sure it was worth the time switching, in my opinion.

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ReleaseService.cs
```diff
@@ -71,6 +72,18 @@ private async Task<List<SubjectViewModel>> GetSubjects(Guid releaseVersionId, Li
                 .Where(rs => rs.ReleaseVersionId == releaseVersionId && subjectsToInclude.Contains(rs.SubjectId))
                 .ToListAsync();
 
+            if (!ComparerUtils.SequencesAreEqualIgnoringOrder(
+                    releaseSubjects.Select(rs => rs.SubjectId).ToList(),
+                    subjectsToInclude
+                    ))
```
benoutram, line 78: Not one for now but I'm wondering if we should eventually get rid of that util method in favour of something like this where it's clearer there's still ordering going on, and it's shorter!

`subjectsToInclude.Order().SequenceEqual(releaseSubjects.Select(rs => rs.SubjectId).Order())`

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/ReleaseServiceTests.cs
```diff
@@ -243,6 +249,107 @@ await statisticsDbContext.IndicatorGroup.AddRangeAsync(
...
+
+            var import1 = _fixture.DefaultDataImport()
+                .WithFile(releaseFile1.File)
+                .WithStatus(DataImportStatus.COMPLETE)
+                .Generate();
+
+            var import2 = _fixture.DefaultDataImport()
+                .WithFile(releaseFile2.File)
+                .WithStatus(DataImportStatus.COMPLETE)
+                .Generate();
```
benoutram, line 323: You could avoid all of these unnecessary `.Generate()`  by declaring the type rather than using `var`, which will use the implicit operator, e.g.

```csharp
            ReleaseSubject releaseSubject1 = _fixture.DefaultReleaseSubject()
                .WithReleaseVersion(statisticsReleaseVersion)
                .WithSubject(_fixture.DefaultSubject());
```

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/ReleaseServiceTests.cs
```diff
@@ -243,6 +249,107 @@ await statisticsDbContext.IndicatorGroup.AddRangeAsync(
...
+                    .WithType(FileType.Data)
+                    .WithSubjectId(releaseSubject1.SubjectId))
+                .Generate();
+
+            var releaseFile2 = _fixture.DefaultReleaseFile()
+                .WithReleaseVersion(contentReleaseVersion)
+                .WithFile(_fixture.DefaultFile()
+                    .WithType(FileType.Data)
+                    .WithSubjectId(Guid.NewGuid()))
+                .Generate();
```
benoutram, line 313: The two `.WithType(FileType.Data)` are unnecessary here since you can now do `_fixture.DefaultFile(FileType.Data)` which sets up the files with all the defaults we'd expect for data files.

If you use `.WithType(FileType.Data)` then you don't get that.

src/GovUk.Education.ExploreEducationStatistics.Data.Services.Tests/ReleaseServiceTests.cs
```diff
@@ -243,6 +249,107 @@ await statisticsDbContext.IndicatorGroup.AddRangeAsync(
...
+            var import2 = _fixture.DefaultDataImport()
+                .WithFile(releaseFile2.File)
+                .WithStatus(DataImportStatus.COMPLETE)
+                .Generate();
+
+            var contentDbContextId = Guid.NewGuid().ToString();
+            await using (var contentDbContext = InMemoryContentDbContext(contentDbContextId))
+            {
+                await contentDbContext.AddRangeAsync(releaseFile1, releaseFile2);
+                await contentDbContext.AddRangeAsync(import1, import2);
```
benoutram, line 329: In the style guide we've said

> Prefer to use `context.{EntityType}.Add` over `context.Add`

src/GovUk.Education.ExploreEducationStatistics.Data.Services/ReleaseService.cs
mmyoungman: I'm not really fussed about it - I think either way it's clear what it does at a glance and there isn't any ambiguity which means you need to look inside the method. If I were starting from scratch, I'd prefer the new way you're proposing.

So you're right I do think this is too minor to be spending any time on :)

benoutram (approved):

PR #5157, opened 2024-08-16
EES-5405 Stash all release related fields of a data set version in a new Release entity

benoutram:
This PR simplifies Public API endpoints for getting a data set version and listing data set versions by removing the need to make an external call to the Content API to get release file and release information related to data set versions. 

The fields currently gathered from the external Content API client call are the `File.DataSetFileId`, `ReleaseVerson.Slug`, and `ReleaseVersion.Title`, all from the content model.

This PR makes a Public Data model change to stash all of these release related fields of a data set version in a new `Release` owned entity owned by `DataSetVersion`. Existing field `DataSetVersion.ReleaseFileId` is also migrated to the new entity.

As well as removing the need for the external call, this change makes getting a data set version compatible with preview tokens in https://github.com/dfe-analytical-services/explore-education-statistics/pull/5133. Attempting to get a data set version with a preview token currently fails due to the Content API not allowing access to unpublished data unless.

This results in one renamed column and three new columns of the `DataSetVersion` database table:

- `ReleaseFileId` -> `Release_ReleaseFileId`
- `Release_DataSetFileId`
- `Release_Slug`
- `Release_Title`

### Other changes

- Fix a bug in `Admin.Tests.Fixture.IntegrationTestFixture` where after clearing test data, tracked entities in `ChangeTracker` were not cleared. Note, it's important to access the context in the fixture with `TestApp.Services.GetRequiredService<PublicDataDbContext>()` rather than `TestApp.GetDbContext<PublicDataDbContext>()` to ensure we get the same Scoped lifetime context instance from the same service provider as used by DI to set up transient services, rather than from a new scope service provider which would create a new context instance where clearing the tracker would have no effect.

- Reformat `Admin.Tests.Services.ReleaseServiceTests`, grouping related tests into separate classes.


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
```diff
@@ -182,12 +182,41 @@ public async Task<Either<ActionResult, DataSetDraftVersionViewModel>> UpdateVers
...
+        }
+    }
+
     private async Task<IReadOnlyDictionary<Guid, ReleaseVersion>> GetReleasesVersionsByDataSetVersionId(
         IReadOnlyList<DataSetVersion> dataSetVersions,
         CancellationToken cancellationToken)
     {
         var dataSetVersionsByReleaseFileId = dataSetVersions
-            .ToDictionary(dsv => dsv.ReleaseFileId);
+            .ToDictionary(dsv => dsv.Release.ReleaseFileId);
```
jack-hive, line 219: I would have expected this to blow up because we didn't do a `.Include(dsv => dsv.Release)` when grabbing the `DataSetVersion` which is passed into this method

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
```diff
@@ -325,7 +354,7 @@ private async Task<ReleaseFile> GetReleaseFile(
     {
         return await contentDbContext.ReleaseFiles
             .AsNoTracking()
-            .Where(rf => rf.Id == dataSetVersion.ReleaseFileId)
+            .Where(rf => rf.Id == dataSetVersion.Release.ReleaseFileId)
```
jack-hive, line 357: I would have expected this to blow up because we didn't do a `.Include(dsv => dsv.Release)` when grabbing the `DataSetVersion` which is passed into this method

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
benoutram: You already messaged me with the answer in a DM but just to resolve this comment, the answer is here:

[Querying owned types](https://learn.microsoft.com/en-us/ef/core/modeling/owned-entities#querying-owned-types)

> When querying the owner the owned types will be included by default. It is not necessary to use the Include method, even if the owned types are stored in a separate table.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
benoutram: That's going to be covered by the answer to the previous comment as well. The `DataSetVersion` passed into this method will have the owned `Release` type included by default.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataSetVersionServiceTests.cs
```diff
@@ -96,12 +100,155 @@ await TestApp.AddTestData<ContentDbContext>(context =>
...
+                context.ReleaseFiles.AddRange(releaseVersion1DataFile, releaseVersion2DataFile);
+            });
+
+            DataSetVersion dataSetVersion1 = DataFixture
+                .DefaultDataSetVersion()
+                .WithDataSet(DataFixture.DefaultDataSet())
+                .WithRelease(DataFixture.DefaultDataSetVersionRelease()
+                    .WithReleaseFileId(releaseVersion1DataFile.Id)
+                    .WithSlug(releaseVersion1.Slug)
+                    .WithTitle(releaseVersion2.Title));
```
jack-hive, line 211: think this should be `WithTitle(releaseVersion1.Title)`

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataSetVersionServiceTests.cs
benoutram: Good spot! Fixed now, thanks. 🙂

PR #5167, opened 2024-08-20
EES-5427 Add file info to data set version summaries in admin

ntsim:
This PR changes data set version summaries in the admin to return file info:

- `DataSetVersionSummary.File.Id` (mapped from `ReleaseFile.File.DataSetFileId`)
- `DataSetVersionSummary.File.Title` (mapped from `ReleaseFile.Name`)

This can then be used in the frontend to link through to the data set page in the public frontend.

## Relevant changes

- Updated `ListVersions` endpoint to use a `dataSetId` query parameter instead of the parameter being inside of the URL. This was previously inconsistent with other data set version endpoints.
- Fixed `ListVersions` endpoint not 404ing if the data set can't be found.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionsControllerTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionsControllerTests.cs
@@ -42,7 +42,7 @@ public abstract class DataSetVersionsControllerTests(
 {
     private const string BaseUrl = "api/public-data/data-set-versions";
 
-    public class ListLiveVersionsTests(
+    public class ListVersionsTests(
         TestApplicationFactory testApp) : DataSetVersionsControllerTests(testApp)
     {
         private static readonly IReadOnlyList<DataSetVersionStatus> PreviouslyPublishedDataSetVersionStatuses =
@@ -54,11 +54,11 @@ public abstract class DataSetVersionsControllerTests(
                 ]
             );
 
-        public static TheoryData<DataSetVersionStatus> PreviouslyPublishedDataSetVersionStatusesData = new(
+        public static readonly TheoryData<DataSetVersionStatus> PreviouslyPublishedDataSetVersionStatusesData = new(
             PreviouslyPublishedDataSetVersionStatuses
         );
 
-        public static TheoryData<DataSetVersionStatus> DraftDataSetVersionStatusesData = new(
+        public static readonly TheoryData<DataSetVersionStatus> DraftDataSetVersionStatusesData = new(
             EnumUtil.GetEnums<DataSetVersionStatus>().Except(PreviouslyPublishedDataSetVersionStatuses)
         );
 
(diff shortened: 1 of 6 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionsControllerTests.cs
```diff
@@ -42,7 +42,7 @@ public abstract class DataSetVersionsControllerTests(
 {
     private const string BaseUrl = "api/public-data/data-set-versions";
 
-    public class ListLiveVersionsTests(
+    public class ListVersionsTests(
```
jack-hive, line 45: Should we rename this back to `ListLiveVersionsTests` as we've kept the logic to only return the live ones?

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Public.Data/DataSetVersionsController.cs
```diff
@@ -18,16 +18,15 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api.Publi
 [Route("api/public-data/data-set-versions")]
 public class DataSetVersionsController(IDataSetVersionService dataSetVersionService) : ControllerBase
 {
-    [HttpGet("{dataSetId:guid}")]
+    [HttpGet]
     [Produces("application/json")]
-    public async Task<ActionResult<PaginatedListViewModel<DataSetLiveVersionSummaryViewModel>>> ListLiveVersions(
+    public async Task<ActionResult<PaginatedListViewModel<DataSetLiveVersionSummaryViewModel>>> ListVersions(
```
jack-hive, line 23: Should we rename this back to `ListLiveVersions` as we've kept the logic to only return the live ones?

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionsControllerTests.cs
ntsim: I made this change on purpose as it felt too specific a name for an endpoint that lists versions.

IMO, filtering on 'live' statuses makes sense as a default, but it'd be a good idea to keep the naming more generic to keep the endpoint open to extension for other statuses (e.g. with a query parameter or something) in the future.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/Public.Data/DataSetVersionsController.cs
ntsim: As above comment

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionsControllerTests.cs
jack-hive: I don't feel too strongly either way. Happy to approve :)

jack-hive (approved):

PR #5168, opened 2024-08-20
EES-5303 Remove ability to add italic text in all content editors & strip italic elements from rendered text content

bennettstuart:
Removes the ability to create italicised text content: removed the "Italic" plugin from the base ckeditor build.
Removes previously created italicised text formatting by stripping `i` & `em` tags from `HtmlBlock` and `MarkdownBlock`.

Also removed references to italic plugin which are now redundant.

Italic text was raised as an accessibility issue in the DAC report.


src/explore-education-statistics-common/src/modules/find-statistics/components/ContentBlockRenderer.tsx
```diff
@@ -46,7 +46,15 @@ const ContentBlockRenderer = ({
 
   switch (type) {
     case 'MarkDownBlock':
-      return <ReactMarkdown className="dfe-content">{body}</ReactMarkdown>;
+      return (
```
amyb-hiveit, line 49: Just for info - we don't allow markdown blocks to be created any more, I think this is still in there in case there are still some old ones lurking. But good to have the check for italics here regardless.

src/explore-education-statistics-admin/src/config/ckEditorConfig.ts
```diff
@@ -74,7 +72,7 @@ export const pluginsConfigLinksOnly: ReadonlySet<PluginName> =
   new Set<PluginName>(['Link']);
 
 export const pluginsConfigSimple: ReadonlySet<PluginName> = new Set<PluginName>(
-  ['Bold', 'Italic', 'Link', 'List'],
+  ['Bold', 'Link', 'List'],
```
amyb-hiveit, line 75: As well as removing it here it can be removed from the PluginName and ToolbarOption types in  https://github.com/dfe-analytical-services/explore-education-statistics/blob/dev/src/explore-education-statistics-admin/src/types/ckeditor.ts

amyb-hiveit (changes requested):
I tried to be critical but I could only find one small thing that needs tidying up! Other than that it all looks good.

src/explore-education-statistics-admin/src/config/ckEditorConfig.ts
bennettstuart: Ahh, good spot! I've now removed those

amyb-hiveit (approved):

PR #5177, opened 2024-08-23
EES-5401 update datablock to use geoJson API call instead of boundaryLevel parameter

bennettstuart:
See ticket: [EES-5401](https://dfedigital.atlassian.net.mcas.ms/browse/EES-5401)

Wired dataBlocks to use the new geoJson endpoint
- Replaces optional `boundaryLevel` parameter being added to `getDataBlockTableData` with a wholly separate call to the recently added geoJson endpoint.

This is the first of two PR's with embedding geoJson API into the frontend, the next will regard [EES-5402](https://dfedigital.atlassian.net.mcas.ms/browse/EES-5402)


src/explore-education-statistics-common/src/services/tableBuilderService.ts
```diff
@@ -275,21 +275,26 @@ const tableBuilderService = {
...
   getFastTrackTableAndReleaseMeta(
     dataBlockParentId: string,
   ): Promise<FastTrackTableAndReleaseMeta> {
     return dataApi.get(`/tablebuilder/fast-track/${dataBlockParentId}`);
   },
+  getGeoJson(
+    releaseId: string,
+    dataBlockParentId: string,
+    boundaryLevelId: number,
+  ): Promise<Dictionary<LocationOption[]>> {
```
ntsim, line 292: Could do a bit better with this type. `LocationOption` has an optional `geoJson` property, but as we're requesting from this endpoint, we know that the `geoJson` is populated.

We could do something like:

```ts
// tableBuilderService.ts

export interface LocationGeoJsonOption extends LocationOption {
  geoJson: GeoJsonFeature;
}


const tableBuilderService = {
  getLocationGeoJson(...): Promise<Dictionary<LocationGeoJsonOption[]>> {
    // ...
  }
}
```

ntsim (changes requested):

src/explore-education-statistics-common/src/modules/charts/components/MapBlock.tsx
```diff
@@ -108,7 +116,36 @@ export const mapBlockDefinition: ChartDefinition = {
...
+  return (
+    <MapBlock
+      {...props}
+      meta={{
+        ...meta,
+        locations: mapLocationOptionsToFilters(geoJson),
+      }}
+    />
+  );
+}
```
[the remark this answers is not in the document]
bennettstuart: Hey, thanks for the review. I agree with your ideas about the implementation. I've given it a stab - please would you take another look and if you feel that this isn't heading in the right direction then let's have a call to get on the same page.

src/explore-education-statistics-common/src/modules/find-statistics/queries/tableBuilderQueries.ts
```diff
@@ -0,0 +1,22 @@
+import tableBuilderService from '@common/services/tableBuilderService';
+import { createQueryKeys } from '@lukemorales/query-key-factory';
+
+const tableBuilderQueries = createQueryKeys('tableBuilder', {
+  getGeoJson(
```
ntsim, line 5: We might not need this if we end up going down the combined query approach (outlined in my other comment), but if we do keep this, we should rename it to `getLocationGeoJson`

src/explore-education-statistics-common/src/modules/find-statistics/queries/tableBuilderQueries.ts
```diff
@@ -0,0 +1,22 @@
...
+    releaseId: string,
+    dataBlockParentId: string,
+    boundaryLevelId: number,
+  ) {
+    return {
+      queryKey: ['geoJson', releaseId, dataBlockParentId, boundaryLevelId],
```
ntsim, line 11: Don't need the `geoJson` string as part of the query key. `createQueryKeys` will automatically include the method name into the query key. It will automatically generate something like:

```ts
['tableBuilder', 'getGeoJson', releaseId, dataBlockParentId, boundaryLevelId]
```

When we don't use `createQueryKeys`, we need to include some unique string (like the method name) to provide enough uniqueness to the query key. 

Currently, we're not using `createQueryKeys` in the public frontend, but now that we've started using this in common, we'll probably want to convert the existing queries to use it at some point down the line.

src/explore-education-statistics-common/src/modules/table-tool/utils/mapFullTableMeta.ts
```diff
@@ -34,15 +34,14 @@ export default function mapFullTableMeta(
...
-   * case.
-   */
+  //  The location hierarchy expects grouping, for example the LAD attribute
+  //  is grouped by Region. However, the data screener
+  //  (https://rsconnect/rsc/dfe-published-data-qa/) does not require the data
+  //  to have all attributes of the hierarchy. When this data is missing the
+  //  backend returns an empty string for the group label. This can cause table
+  //  layout problems so we need to not group the locations when this is the
+  //  case.
+
```
ntsim, line 44: Unnecessary extra line break

src/explore-education-statistics-common/src/modules/find-statistics/hooks/useTableQuery.ts
```diff
@@ -6,15 +6,19 @@ import tableBuilderService from '@common/services/tableBuilderService';
 export default function useTableQuery(
   releaseId: string,
   dataBlockParentId: string,
-  boundaryLevelId?: number,
+  boundaryLevel?: number
```
ntsim, line 9: Shouldn't this be `boundaryLevelId`?

src/explore-education-statistics-common/src/modules/find-statistics/hooks/useTableQuery.ts
```diff
@@ -6,15 +6,19 @@ import tableBuilderService from '@common/services/tableBuilderService';
...
+        releaseId,
+        dataBlockParentId,
+      ),
+      boundaryLevel !== undefined ? tableBuilderService.getLocationGeoJson(releaseId, dataBlockParentId, boundaryLevel): undefined
+    ])
 
-    return mapFullTable(response);
+    
+
+    return mapFullTable({...tableData, subjectMeta: {...tableData.subjectMeta, locations: locationGeoJson ?? tableData.subjectMeta.locations}});
```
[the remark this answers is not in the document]
bennettstuart: Yo, should I also bring in getKeyStat (see **useKeyStateQuery.ts**) into **tableBuildQueries**? Looks like currently `useKeyStatQuery` has a setter, which never seems to be used so I think this can be replaced with a `useQuery` implementation. The reason I ask is because now that I've removed (locally) the old `useTableQuery` this key stat query now needs refactoring.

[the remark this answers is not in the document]
ntsim: Yep that's definitely one option.

Alternatively, it's fine to also keep `useKeyStatQuery` and refactor it. Think it'd be fairly straightforward via the `select` option to do the transform e.g.

```ts
export default function useKeyStatQuery(...): UseQueryResult<KeyStatResult>
  return useQuery({
    ...tableBuilderQueries.getDataBlockTable(...),
    select: data => {
      // Can transform the data in some way here
      return data;
    }
  });
}
```

I don't massively mind which option we go for tbh, so will leave to your discretion.

Btw, it's also worth giving this article on [data transformations](https://tkdodo.eu/blog/react-query-data-transformations) a read. It's by one of the lead maintainers for React Query, so it's got some interesting things you'll want to absorb.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/DataBlockPageTabs.tsx
```diff
@@ -78,12 +78,8 @@ const DataBlockPageTabs = ({
...
     };
 
-    const boundaryLevel = dataBlock.charts[0]?.boundaryLevel ?? undefined;
-    const tableData = await tableBuilderService.getTableData(
-      query,
-      releaseId,
-      boundaryLevel,
-    );
+    const tableData = await tableBuilderService.getTableData(query, releaseId);
+
```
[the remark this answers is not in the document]
bennettstuart: Unless I'm missing something I believe that will be picked up in [EES-5402 the admin side of this work](https://dfedigital.atlassian.net.mcas.ms/browse/EES-5402) - would you agree?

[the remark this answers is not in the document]
ntsim: Not sure as I don't have the full context here. I think you'll want to coordinate with @tomjonesdev.

However, just looking at this change, if you remove the boundary level parameter and don't feed this through to the endpoint, it will stop returning GeoJSON. This would result in all map chart previews being broken in the admin.

I would expect there to be a new endpoint to fetch GeoJSON in a similar way to the how we're doing this in the data blocks. However, if something like this isn't available yet then we need to revert this change and continue to use the current endpoint!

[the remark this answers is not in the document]
bennettstuart: Right you are, I'd stepped over into Admin and made changes there along the way which I think need reverting and any further changes will be added in the subsequent ticket as this aims to focus on using the geoJson endpoint on the frontend

src/explore-education-statistics-common/src/modules/charts/components/MapBlock.tsx
```diff
@@ -120,7 +120,7 @@ export default function MapBlock({
   width,
   height,
   axes,
-}: MapBlockProps) {
+}: Omit<MapBlockProps, 'releaseId' | 'dataBlockParentId' | 'boundaryLevel'>) {
```
ntsim, line 123: These changes should be reverted as `MapBlockProps` hasn't changed

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/__tests__/ChartBuilderPreview.test.tsx
```diff
@@ -184,7 +184,8 @@ describe('ChartBuilderPreview', () => {
     render(
       <ChartBuilderPreview
         chart={produce(testMapChartRenderer, draft => {
-          draft.boundaryLevel = undefined;
+          (draft as Chart & { boundaryLevel?: number }).boundaryLevel =
```
ntsim, line 187: This doesn't seem right. We've now made `boundaryLevel` required in `ChartRenderProps`, so the correct way to set this would be simply:

```ts
draft.boundaryLevel = 0;
```

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/__tests__/ChartBuilderPreview.test.tsx
```diff
@@ -1,12 +1,12 @@
...
-import { render, screen } from '@testing-library/react';
 import ChartBuilderPreview from '@admin/pages/release/datablocks/components/chart/ChartBuilderPreview';
+import { testChartTableData } from '@common/modules/charts/components/__tests__/__data__/testChartData';
 import { ChartRendererProps } from '@common/modules/charts/components/ChartRenderer';
-import { produce } from 'immer';
+import { AxisConfiguration } from '@common/modules/charts/types/chart';
 import { DataSet } from '@common/modules/charts/types/dataSet';
 import mapFullTable from '@common/modules/table-tool/utils/mapFullTable';
-import { testChartTableData } from '@common/modules/charts/components/__tests__/__data__/testChartData';
+import { Chart } from '@common/services/types/blocks';
```
ntsim, line 7: Unneeded import

src/explore-education-statistics-common/src/modules/find-statistics/components/__tests__/DataBlockTabs.test.tsx
```diff
@@ -368,11 +365,21 @@ describe('DataBlockTabs', () => {
...
         'release-1',
         'block-1-parent',
-        1,
+      );
+
+      expect(tableBuilderService.getLocationGeoJson).toBeCalledWith(
+        'release-1',
+        'block-1-parent',
+        testMapConfiguration.type === 'map' &&
+          testMapConfiguration.boundaryLevel,
```
ntsim, line 374: This can just be simplified to `1` (as it was previously)

ntsim (approved):

PR #5178, opened 2024-08-23
EES-5442 Set PublicId when creating filter option and location mappings

benoutram:
This PR fixes a bug where creating new data set versions following v1.0.0 would fail to finalise after completing the mapping stage. 

The failure occurred while inserting new `FilterOptionMetaLink` rows for auto mapped or manually mapped filter options due to violating the unique constraint `IX_FilterOptionMetaLinks_MetaId_PublicId` because every row was being inserted with an empty PublicId.

This was happening because in an earlier stage of processing, the new data set version we were not setting `PublicId` for the source filter options in the `FilterMappingPlan` so these were defaulting to the empty string.

Post-mapping when determining the PublicId mappings , this value is now copied over when creating `PublicIdMappings` for auto mapped and manually mapped filter options.

A similar problem existed for location options which is also fixed.


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionMappingService.cs
```diff
@@ -345,13 +346,13 @@ private LocationMappingPlan CreateLocationMappings(
...
-                            .Select(option => option.ToRow())
+                            .OptionLinks
                             .ToDictionary(
-                                keySelector: MappingKeyFunctions.LocationOptionMetaRowKeyGenerator,
-                                elementSelector: option => new LocationOptionMapping
+                                keySelector: MappingKeyFunctions.LocationOptionMetaLinkKeyGenerator,
+                                elementSelector: link => new LocationOptionMapping
                                 {
-                                    Source = CreateLocationOptionFromMetaRow(option)
+                                    PublicId = link.PublicId,
```
ntsim, line 354: Just wondering if this could have been avoided if we'd made `PublicId` required? 

Seems like there's a few fields that haven't been marked as required on the model, so not sure if there's any particular reason for this. If not, would probably be a good idea to make as many fields required as possible!

Similar thing applies for the filter mapping model as well.

benoutram: Ah yeah, good point. I think you are right and I tried making this change yesterday but it upset a few things and ran out of time getting it sorted. In the mean time though I pushed another change related to the refactoring of MappingKeyFunctions that appeared to have been made before but lost.

duncan-at-hiveit (approved):

PR #5181, opened 2024-08-27
EES-5443 - Verify row data contents inside the 'Draft API datasets' summary table

saicharan2789:


saicharan2789:
image

tests/robot-tests/tests/public_api/public_api_restricted.robot
```diff
@@ -106,6 +106,43 @@ Verify the contents inside the 'Draft API datasets' table
...
+
+User checks row data contents inside the 'Draft API datasets' summary table
+    user checks contents inside the cell value    v1.0                                       xpath://dl[@data-testid="draft-version-summary"]/div/dd[@data-testid='Version-value']/strong
+    user checks contents inside the cell value    Ready                                      xpath:(//div[@data-testid="Status"]//dd[@data-testid="Status-value"]//strong)[2]
+    user checks contents inside the cell value    Financial year 3000-01                     xpath:(//div[@data-testid="Release"]//dd[@data-testid="Release-value"]//a)[1]
+    user checks contents inside the cell value    ${SUBJECT_NAME_1}                          xpath://div[@data-testid="Data set file"]//dd[@data-testid="Data set file-value"]
+    user checks contents inside the cell value    National                                   xpath://div[@data-testid="Geographic levels"]//dd[@data-testid="Geographic levels-value"]
+    user checks contents inside the cell value    2012/13                                    xpath://div[@data-testid="Time periods"]//dd[@data-testid="Time periods-value"]
```
mmyoungman, line 120: Is it possible that some of these selectors could be simplified to just `//dd[@data-testid="Time periods-value"]`?

Same for the selectors below

tests/robot-tests/tests/libs/public-api-common.robot
```diff
@@ -46,3 +46,22 @@ verify status of API Datasets
     user waits for caches to expire
     ${status_value}=    get text    xpath:(//div[@data-testid="Status"]//dd[@data-testid="Status-value"]//strong)[2]
     should be equal as strings    ${status_value}    ${expected_status}
+    
+user checks value inside the column heading in the Draft version table
```
mmyoungman, line 50: If this is checking the status specifically, I think this can be renamed to something like `user checks status in Draft version table`?

tests/robot-tests/tests/libs/public-api-common.robot
```diff
@@ -46,3 +46,22 @@ verify status of API Datasets
...
+    user waits for caches to expire
+    ${status_value}=    get text    xpath:(//div[@data-testid="Status"]//dd[@data-testid="${text}"]//strong)[2]
+    should be equal as strings    ${status_value}    ${expected_status}
+
+user checks contents inside the cell value
+    [Arguments]      ${expected_text}     ${locator}
+     ${status_value}=    get text    ${locator}
+    should be equal as strings    ${status_value}    ${expected_text}
+
+user checks row data heading in the Draft version details table
```
mmyoungman, line 61: I think I'd change this to `user checks table headings for Draft version details table`

mmyoungman (approved):

tests/robot-tests/tests/public_api/public_api_restricted.robot
mmyoungman: I'd be surprised if the selector itself is causing an issue here, but equally I don't feel strongly about this so happy for you to leave it. Assuming there is no issue with the simplified selectors, there is an argument that the time to simplify these selectors for readability isn't worth it anyway.

PR #5183, opened 2024-08-28
EES-5442 - updating file storage path to reflect updates to a next DataSetVersion's version number

duncan-at-hiveit:
This PR:
- adds an additional stage to the final import orchestration of a next DataSetVersion, in which the storage directory path is updated to reflect any changes in the DataSetVersion's version number during its mapping phase.
- if the version number is unchanged from the default version number originally assigned when the next DataSetVersion was created, no path update is necessary.

Thereafter, the import process should continue as usual, using the now-updated file storage path.

I've added a convenience `DataSetVersion.DefautlNextVersion()" method which will give us the expected next version number of a DataSetVersion (using the existing implementation of updating the minor version).  We can then use this method to assign a default version number to a brand new next DataSetVersion, plus use it again to get the default file storage path that may or may not need updating prior to completing the import process.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/ProcessCompletionOfNextDataSetVersionFunction.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/ProcessCompletionOfNextDataSetVersionFunction.cs
@@ -5,6 +5,7 @@ using GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Model;
 using GovUk.Education.ExploreEducationStatistics.Public.Data.Services.Interfaces;
 using Microsoft.Azure.Functions.Worker;
 using Microsoft.DurableTask;
+using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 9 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/ProcessCompletionOfNextDataSetVersionFunction.cs
```diff
@@ -3,6 +3,7 @@
 using GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Model;
 using GovUk.Education.ExploreEducationStatistics.Public.Data.Services.Interfaces;
+using LinqToDB;
```
ntsim, line 6: Whilst this works, probably should be using `SingleAsync` from `Microsoft.EntityFrameworkCore`

duncan-at-hiveit: Aha good spot, updated thanks.

ntsim (approved):

PR #5193, opened 2024-08-30
EES-4723 Add Content API download link for published data files

mmyoungman:
This PR adds a CSV download link for use on Data set details pages. This link will be put on the page alongside guidance of how to use the link in code/scripts to access the data programmatically.

The frontend work will be done separately.


src/GovUk.Education.ExploreEducationStatistics.Content.Services/DataSetFileService.cs
```diff
@@ -246,6 +248,37 @@ public async Task<Either<ActionResult, DataSetFileViewModel>> GetDataSetFile(Gui
...
 
+    public async Task<ActionResult> DownloadDataSetFile(
+        Guid dataSetFileId)
+    {
+        var releaseFile = await _contentDbContext.ReleaseFiles
+            .Include(rf => rf.File)
+            .Where(rf =>
+                rf.File.DataSetFileId == dataSetFileId
+                && rf.ReleaseVersion.Published.HasValue
+                && DateTime.UtcNow >= rf.ReleaseVersion.Published.Value)
```
tomjonesdev, line 259: Is there a situation where something would be "published" in the future? IIRC a scheduled publish uses a different column.

mmyoungman: Not sure to be honest. Without looking it up, my guess would be no - I asked on Slack and one other person agreed. So you could argue this is redundant, but I guess since we absolutely don't want to be displaying releases that shouldn't be published, there is no harm in this.

mmyoungman:
Put Do not merge on this PR - realised I pushed my started integration test here.

PR #5210, opened 2024-09-03
EES-5480 - handling creation of container app managed identity and its role assignments in Bicep

duncan-at-hiveit:
This PR:
- Adds the creation and the role assignments of the Container App managed identity into Bicep rather than being a manual task.

This is made possible now by the work of Mark Nelson to allow our DevOps SPNs to be able to assign a subset of common roles.

### DIFF
--- a/infrastructure/templates/public-api/components/containerRegistryRoleAssignment.bicep
+++ b/infrastructure/templates/public-api/components/containerRegistryRoleAssignment.bicep
@@ -1,14 +1,18 @@
-@description('A Container Registry-specific role to assign')
+@description('Specifies the name of an existing Container Registry to be use as the scope of the role assignment')
+param containerRegistryName string
+
+@description('Specifies the id of the service principals that should inherit this Key Vault policy')
+param principalIds string[]
+
+// A subset of allowed roles is supported here in accordance with role assignment conditions.
+// See https://docs.microsoft.com/azure/role-based-access-control/built-in-roles for possible
+// roles to support here, in conjunction with the limited set of roles that the deploying service
+// principal is allowed to assign.
+@description('The Container Registry-specific role to assign')
 @allowed([
   'AcrPull'
 ])
 param role string
-
-@description('The name of an existing Container Registry to be use as the scope of the role assignment')
-param containerRegistryName string
-
-@description('The name of the Managed Identity that the role will be assigned to')
-param managedIdentityName string
 
 // See https://learn.microsoft.com/en-us/azure/role-based-access-control/built-in-roles#containers for Container-specific built in role ids.
 var rolesToRoleIds = {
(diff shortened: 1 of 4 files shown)


infrastructure/templates/public-api/components/containerRegistryRoleAssignment.bicep
```diff
@@ -0,0 +1,34 @@
...
+
+resource roleAssignment 'Microsoft.Authorization/roleAssignments@2022-04-01' = {
+  name: guid(containerRegistry.id, rolesToRoleIds[role], managedIdentitty.name)
+  scope: containerRegistry
+  properties: {
+    roleDefinitionId: subscriptionResourceId('Microsoft.Authorization/roleDefinitions', rolesToRoleIds[role])
+    principalId: managedIdentitty.properties.principalId
+    principalType: 'ServicePrincipal'
+  }
+}
```
benoutram, line 34: Could add a comment to mention that this is dependent on the deploying service principal having the  `Microsoft.Authorization/roleAssignments/write` permission, such as being an Owner, User Access Administrator, or RBAC Administrator at the scope the role is being assigned, and that AcrPull has been allowed through role assignment conditions? I think there's only a small-ish set of roles been conditionally allowed which can be granted, `AcrPull` being one of them.

infrastructure/templates/public-api/main.bicep
```diff
@@ -228,8 +228,18 @@ module postgreSqlServerModule 'components/postgresqlDatabase.bicep' = if (update
...
+  location: location
+}
+
+module acrPullRoleAssignmentModule 'components/containerRegistryRoleAssignment.bicep' = {
+  name: '${apiContainerAppManagedIdentityName}AcrPullRoleAssignmentDeploy'
+  params: {
+    role: 'AcrPull'
+    containerRegistryName: acrName
+    managedIdentityName: apiContainerAppManagedIdentity.name
+  }
```
benoutram, line 242: This is the Api container's registry access so maybe it should include 'apiContainer' in the name in case another service/identity needs a similar role assignment? Something like `apiContainerAcrPullRoleAssignment` or `apiContainerRegistryAccess` rather than `acrPullRoleAssignmentModule`?

Could also think about grouping everything related to the Api container into a new `app/apiContainer.bicep` in future.

infrastructure/templates/public-api/components/containerRegistryRoleAssignment.bicep
```diff
@@ -0,0 +1,34 @@
...
+// See https://learn.microsoft.com/en-us/azure/role-based-access-control/built-in-roles#containers for Container-specific built in role ids.
+var rolesToRoleIds = {
+  AcrPull: '7f951dda-4ed3-4680-a7ca-43fe172d538d'
+}
+
+resource containerRegistry 'Microsoft.ContainerRegistry/registries@2023-07-01' existing = {
+  name: containerRegistryName
+}
+
+resource managedIdentitty 'Microsoft.ManagedIdentity/userAssignedIdentities@2023-01-31' existing = {
```
benoutram, line 22: Spelling mistake in the name here and where it's referenced below

benoutram (approved):

infrastructure/templates/public-api/components/containerRegistryRoleAssignment.bicep
duncan-at-hiveit: I've added some comments to this effect thanks.  I've also updated `containerAppRoleAssignment.bicep` to behave like `keyVaultRoleAssignment.bicep` for consistency (taking an array of principalIds).

infrastructure/templates/public-api/main.bicep
duncan-at-hiveit: Updated to `apiContainerAppAcrPullRoleAssignmentModule`.

Also, I already have a ticket for better modularising the Bicep as a whole - https://dfedigital.atlassian.net/browse/EES-5479.  I think this should be done prior to deploying to Test.

infrastructure/templates/public-api/components/containerRegistryRoleAssignment.bicep
duncan-at-hiveit: Removed anyhow, but nicely spotted thanks.

PR #5211, opened 2024-09-03
EES-5478 Introduce new indicator unit 'numberstring'

mmyoungman:
This PR adds a new indicator unit "numberstring".

This has been introduced because analysts have wanted to add ids or codes to their data sets and not have those ids/codes comma separated as any number was previously by default - i.e. 123,456,789.


src/explore-education-statistics-common/src/utils/number/formatPretty.ts
```diff
@@ -20,19 +20,53 @@ export default function formatPretty(
   unit?: string,
   decimalPlaces?: number,
 ): string {
+  switch (unit) {
+    case 'numberstring':
+      return String(value);
+    case '£': {
+      let formattedNumber = formatNumber(value, decimalPlaces);
```
bennettstuart, line 27: this came up in the linting error
to use `const` instead of `let` as this variable is never reassigned. Same goes for the rest of the _cases_ below

bennettstuart (changes requested):
Looks good! just need to fix the pipeline

bennettstuart (approved):

PR #5219, opened 2024-09-04
EES-5422 Change filters and indicators to use sqids for public IDs

ntsim:
## :warning: Breaking changes :warning: 

This PR does a full re-baseline of the public API DB and introduces backwards incompatible changes, it requires a full clean down of dev once merged:

- Public API DB - All tables and sequences should be deleted
- Content DB - `PublicApiDataSetId` and `PublicApiDataSetVersion` in `ReleaseFiles` table should all be set to null
- Public API file store - All files should be deleted

## Summary

This PR changes the public ID implementations for filter and indicator meta to use sqids like other meta types. Previously, we were just using the column names as their public IDs, but this meant that changes to filter or indicator columns would always need to be classed as major changes.

With this change, we will be able to implement mapping operations for filter and indicator columns, allowing only minor version bumps for these.

As part of this, various parts of the public API now use sqid-based public IDs in requests and responses e.g.

- Query requests now use sqids for `indicators`
- Query results `filters` now have the filter sqids as keys (previously used column names)
- Query results `values` now have the indicator sqids as keys (previously used column names)
- Enabling `debug` mode for queries displays the filter / indicator column name in the respective keys

## Re-baselined database and column constraints

As part of this change, breaking changes were unavoidable so we've gone with fully re-baselining the database for ease.

To prevent invalid or undesirable values making it into the public API DB, we've also taken this opportunity to add extra column constraints across the various models. This is mostly in the form of max character lengths e.g.

- Location codes - 20 or 30 characters
- Indicator / filter / location labels - 80 to 120 characters
- Indicator / filter columns - 50 characters

These max lengths have all been agreed with Cam. If required in the future, we can raise these on a case-by-case basis.


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Scripts/Commands/SeedDataCommand.cs
```diff
@@ -58,7 +58,8 @@ public async ValueTask ExecuteAsync(IConsole console)
             DataSetSeed.SpcEthnicityLanguage,
             DataSetSeed.SpcYearGroupGender,
             DataSetSeed.Nat01,
-            DataSetSeed.Qua01,
+            // QUA01 has filter option labels that are too long
```
duncan-at-hiveit, line 61: Is it worth just getting rid of this if we can't use it?

I'm sure you've discussed with Cam, but if their filter option labels are too long in this example file, doesn't this suggest that teams out there will be requiring labels of a greater length than we support?  If so, I guess that there will need to be communications about this sent out to the teams.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Scripts/Commands/SeedDataCommand.cs
```diff
@@ -387,59 +388,88 @@ private async Task CreateIndicatorMeta(
...
+                currentId - 1,
+                _cancellationToken
+            );
         }
 
         private async Task CreateFilterMetas(
             IEnumerable<MetaFileRow> metaFileRows,
             IReadOnlySet<string> allowedColumns)
         {
+            var currentMetaId = await _dbContext.NextSequenceValue(
```
duncan-at-hiveit, line 431: Could just call this `currentId` as you have done in other places?

src/explore-education-statistics-admin/src/pages/release/data/components/ApiDataSetMappableFilterColumnOptionsModal.tsx
```diff
@@ -3,45 +3,53 @@ import Modal from '@common/components/Modal';
...
-const itemsPerPage = 10;
-
 interface Props {
-  id: string;
+  column: string;
   label: string;
-  modalLabel: string;
   options: string[];
+  pageSize?: number;
+  publicId?: string;
```
duncan-at-hiveit, line 15: Is there a chance of publicId being null?  Assuming that the column, label, publicId etc are all from a *source* Filter, we should always have a publicId in the mapping data.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/DataSetQueryLocation.cs
```diff
@@ -214,8 +213,7 @@ public record DataSetQueryLocationLocalAuthorityOldCode : DataSetQueryLocation
 {
     /// <summary>
     /// The old code (previously the LEA code) of the local authority.
-    /// This should be a 3 digit number (e.g. `318`) or be
-    /// a combination of two codes (e.g. `314 / 318`).
+    /// This is typically a 3-digit number (e.g. `318`).
```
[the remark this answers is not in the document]
ntsim: I wasn't too sure about this. I backpedalled on having this comment because in the future they may have further variations like combining three or more codes together. Whilst unlikely, it's not impossible so the current comment didn't feel right.

I've decided to change the wording up a bit more to: 

> This is typically a 3-digit number e.g. `318`, but may be a combination of multiple codes e.g. `318 / 319`.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Migrations/20240902104511_InitialMigration.cs
```diff
@@ -54,7 +54,7 @@ IF EXISTS (SELECT 1 FROM pg_roles WHERE rolname = '{PublicDataDbContext.PublicDa
                 {
                     Id = table.Column<int>(type: "integer", nullable: false)
                         .Annotation("Npgsql:ValueGenerationStrategy", NpgsqlValueGenerationStrategy.IdentityByDefaultColumn),
-                    Label = table.Column<string>(type: "text", nullable: false),
+                    Label = table.Column<string>(type: "character varying(120)", maxLength: 120, nullable: false),
```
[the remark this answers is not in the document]
ntsim: Think this is just a Postgres alias for varchar. It's automatically generated for us by EF anyway :wink:

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Scripts/Commands/SeedDataCommand.cs
ntsim: I actually kept this in because I was planning on doing some performance testing with it later on, which would modifying the script a little to trim down the labels. I can do this regardless so I've just removed it as you've suggested

I did indeed discuss the length constraints with Cam and the examples of really long filter option labels were qualification titles like:

- Industry placement (315+ hours, aged 16 – 18, level 2 or 3 with a vocational or technical core aim, occupationally specific, must be studying towards a type of qualification that may align to T levels in future and includes an external wo...
- Understand the range of service provision and roles within health and social care adults and children and young people, early years and childcare
- Non regulated Adult skills formula funded provision, Entry Level, Languages, Literature and Culture of the British Isles, 7 to 12 hrs, PW A
- Non regulated Adult skills formula funded provision, Entry Level, Languages, Literature and Culture of the British Isles, Up to 2 hrs, PW A
- Non regulated Adult skills formula funded provision, Entry Level, Nursing and Subjects and Vocations Allied to Medicine, 7 to 12 hrs, PW 

These are unnecessarily long and seemed to be the FE team labelling things badly or trying to skirt around best practice. For example, there are a bunch of 'Non regulated Adult skills formula funded provision...' options that clearly contain filters that they've  not added to the data set as columns.

I think we agreed that the problem lies with FE and that the onus should be on them to sort out, rather than let these bad labels make it into the public API. I think we'll probably try and communicate to the various stats team that public API data is being held to a higher standard and should conform to best practice to make it onto the API.

We've already agreed that a max label length of 120 chars should be more than sufficient if things are being labelled correctly. However, we're open to increasing this in the future if necessary. It's easier to increase the max than it is to lower it later on!

I think we also discussed that Cam (or whoever) would implement the various max length constraints in the screener as well to stop excessively long stuff making it into the platform at all.

src/explore-education-statistics-admin/src/pages/release/data/components/ApiDataSetMappableFilterColumnOptionsModal.tsx
ntsim: This modal is displayed for both filter columns that have been added or deleted. Where a filter column has been added, it doesn't actually have a public ID yet so it will be undefined as we won't have set it - see `ApiDataSetNewFilterColumnsTable`.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/DataSetQueryLocation.cs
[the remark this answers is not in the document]
duncan-at-hiveit: Cool sounds good!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Scripts/Commands/SeedDataCommand.cs
duncan-at-hiveit: Brill sounds like a good plan

PR #5222, opened 2024-09-05
Ees 5479 modularise bicep

duncan-at-hiveit:
This PR:
- breaks the big `main.bicep` file up into separate modules, one for each high-level resource being managed or referenced from the Public API infrastructure deploy.

Resources are created in separate modules, and these modules are in turn separated between "shared" and "public-api" modules.  Shared modules represent resources that can be used beyond the scope of the Public API e.g. generic Log Analytics Workspaces, or Container App Environments.  They also house the ARM-template controlled resources for lookup purposes.

Resource names are now defined in each module using appropriate prefixes, and these are handed down to the Bicep modules in the `components` folder rather than being constructed within each of the `components` modules.  This has led to much more clarity in terms of the names that all of the various resources will take on in the deployment, rather than being buried down in low-level modules.


infrastructure/templates/public-api/application/public-api/publicApiApp.bicep
```diff
@@ -0,0 +1,154 @@
...
+@description('Specifies the Application (Client) Id of the App Registration used to represent the API Container App.')
+param apiAppRegistrationClientId string
+
+@description('Specifies the connection string template for the Public API PSQL database.')
+param publicApiDbConnectionStringTemplate string
+
+@description('Specifies a set of tags with which to tag the resource in Azure.')
+param tagValues object
+
+var apiAppName = '${publicApiResourcePrefix}-ca-api'
```
benoutram, line 37: You could consider adding an `abbreviations.json` file with the set of all abbreviations, which you can take out of one of the sample Azure template projects e.g. https://github.com/Azure-Samples/todo-csharp-sql/blob/main/infra/abbreviations.json, and then using that to reference abbreviations like `ai-`, `ca-`, `cae-`, etc, so that they are always consistent.

You would need to load it though in each file, using `var abbrs = loadJsonContent('./abbreviations.json')` , which is going to get a bit repetitive unless more names are worked out at higher level in main.bicep and passed down.

infrastructure/templates/public-api/application/shared/postgreSqlFlexibleServer.bicep
```diff
@@ -0,0 +1,62 @@
...
+param postgreSqlStorageSizeGB int = 32
+
+@description('Autogrow setting.')
+param postgreSqlAutoGrowStatus string = 'Disabled'
+
+@description('Firewall rules.')
+param postgreSqlFirewallRules {
+  name: string
+  cidr: string
+}[] = []
```
benoutram, line 27: Do a lot of the params need to be prefixed with `postgreSql`? E.g. could be `storageSizeGb` instead of `postgreSqlStorageSizeGB`.

duncan-at-hiveit: Sorted ta

infrastructure/templates/public-api/application/public-api/publicApiApp.bicep
duncan-at-hiveit: As per the other PR - added as a new Bicep file along with types

infrastructure/templates/public-api/components/functionApp.bicep
```diff
@@ -124,10 +125,10 @@ resource appServicePlan 'Microsoft.Web/serverfarms@2023-12-01' = {
 // Configure a single shared storage account for access key storage, and 2 individual storage accounts to be split
 // between the production slot and staging slot for reliable execution. See
 // https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-zero-downtime-deployment#status-check-with-slot
-var sharedStorageAccountName = replace('${subscription}eessa${functionAppName}mg', '-', '')
-var slot1StorageAccountName = replace('${subscription}eessa${functionAppName}s1', '-', '')
-var slot2StorageAccountName = replace('${subscription}eessa${functionAppName}s2', '-', '')
-var functionAppCodeFileShareName = '${fullFunctionAppName}-fs'
+var sharedStorageAccountName = '${storageAccountsNamePrefix}mg'
+var slot1StorageAccountName = '${storageAccountsNamePrefix}s1'
+var slot2StorageAccountName = '${storageAccountsNamePrefix}s1'
```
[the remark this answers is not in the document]
duncan-at-hiveit: Nice spot thanks 😓  Fixed!

infrastructure/templates/public-api/components/containerRegistry.bicep
```diff
@@ -1,14 +1,9 @@
-@description('Specifies the Resource Prefix')
-param resourcePrefix string
+@description('Specifies the name of the Container Registry')
```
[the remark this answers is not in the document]
duncan-at-hiveit: Done ta

PR #5226, opened 2024-09-06
EES-4771 Use Azure.Data.Tables dependency in Admin and Notifier

mmyoungman:
This is a PR to review the work done so far on moving to Azure.Data.Tables.

This work is just being merged to a temporary branch, `EES-4771-reviewed` - not to `dev`.

This should cover all uses of Azure Storage in Admin and Notifier.


mmyoungman:
Closing this as we agreed to have a PR for the whole dependency switch rather than do it piecemeal.

PR #5236, opened 2024-09-10
EES-5464 Download data-catalogue CSV with code UI

bennettstuart:
PR to add UI to that shows users how to use our API to directly access CSV data from our new API endpoint (See ticket: [EES-5463](https://dfedigital.atlassian.net/browse/EES-5463).
Some notable changes:
- Added CodeBlock preview component
- Minor tweaks to CopyTextButton component styling

image


src/explore-education-statistics-common/src/components/Code.module.scss
```diff
@@ -0,0 +1,36 @@
...
+  border: 1px solid #bfc1c3;
+  background: #f9f9f8;
+  max-width: 100%;
+  overflow-x: auto;
+  padding: 60px 15px 15px !important;
+
+  &:focus {
+    border: $govuk-focus-width solid #000;
+    outline: $govuk-focus-colour solid $govuk-focus-width;
+    padding: 58px 13px 13px !important; // counter border width to avoid content shift
```
tomjonesdev, line 34: Can `box-sizing: border-box` be used to counteract content shift?

src/explore-education-statistics-common/src/components/Code.module.scss
```diff
@@ -0,0 +1,36 @@
...
+    top: 12px;
+  }
+}
+
+.pre {
+  border: 1px solid #bfc1c3;
+  background: #f9f9f8;
+  max-width: 100%;
+  overflow-x: auto;
+  padding: 60px 15px 15px !important;
```
tomjonesdev, line 29: What predefined style is being overridden here? Maybe check with a more dedicated FE, but I typically try to avoid using `!important` unless absolutely necessary, opting instead to rely on the cascade, or specificity.

tomjonesdev (approved):

src/explore-education-statistics-common/src/components/Code.module.scss
bennettstuart: Great suggestion, but I've since removed this line after removing the ability to focus code snippets

src/explore-education-statistics-common/src/components/Code.module.scss
bennettstuart: Unfortunately the package is setting an inline `style` on the element, so this is the only way overwrite that.  
There may be a way to configure the styles, but I think that may come out in the wash when we want to render more code snippets (more than just 1 or 2 lines)

PR #5237, opened 2024-09-11
EES-5494 Update Swagger documentation to work with public API docs site

ntsim:
This PR primarily updates the public API's Swagger documentation to be compatible with the public API docs site generator.

The following changes have been made:

- Added examples and missing docblocks for all request / view models in the public API.
- Fixed various issues with the `allOf` and `anyOf` schemas that were being generated by Swashbuckle when certain options were enabled. These have now been disabled to allow the generated schema to roughly work as we'd expect.
- Fixed incorrect `application/+json` media type being used for the `POST /query` endpoint.
- Updated `SwaggerEnum` to a reference to the enum schema using `allOf` as the default. This allows us to link directly to the enum schema in the final API docs.
- Changed `DataSetQueryCriteria` to be an interface so it no longer generates a schema.
- Aliased `GeographicLevel` as `GeographicLevelCode` and `TimeIdentifier` as `TimePeriodCode` to better align with their usage in the public API and documentation.
- Updated `GeographicLevelCode` and `TimePeriodCode` enum schemas to be naturally ordered so they are more human-readable compared to default ASCII-based ordering.

## Other changes

- Fixed incorrect JSON property ordering in responses for meta view models. The default order was placing `options` first meaning that more important information was not immediately visible in the `/meta` response.
- Allow comments in JSON requests for the public API to make it easier for users to create data set queries.


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Swagger/SwaggerEnumSchemaFilter.cs
```diff
@@ -23,6 +24,44 @@ public void Apply(OpenApiSchema schema, SchemaFilterContext context)
                 $"Must use an enum type for '{nameof(SwaggerEnumAttribute)}'");
         }
 
+        var schemaType = schema.Type == "array" ? schema.Items.Type : schema.Type;
```
duncan-at-hiveit, line 27: I'm going to assume that you know exactly what is happening here :)

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Extensions/StringExtensionTests.cs
```diff
@@ -517,5 +517,64 @@ public void AsciiEncoding_Utf8Chars()
...
+                    "z1",
+                    "z3",
+                    "abc",
+                    "z20",
+                    "z5",
+                    "abcd",
+                    "z11",
+                    "z22"
+                ];
+                string[] expected =
```
duncan-at-hiveit, line 559: Nice addition - I wonder if there are other places around the codebase that could use this.  Does the front end have any concept of natural ordering?

duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetsControllerPostQueryTests.cs
```diff
@@ -779,7 +779,7 @@ public async Task AllComparatorsInvalid_Returns400()
 
             var validationProblem = response.AssertValidationProblem();
 
-            Assert.Equal(8, validationProblem.Errors.Count);
+            Assert.Equal(7, validationProblem.Errors.Count);
```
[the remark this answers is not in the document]
ntsim: I removed the `NotEmpty` validation for `IDataSetQueryLocation.Code` as it's unnecessary due to the use of the `AllowValue` validation. This meant the total number of errors dropped by 1.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Swagger/SwaggerEnumSchemaFilter.cs
ntsim: Yep, in where we have an array of items, we're getting the underlying `Type` for each array item. Or, if we don't have a list, we just use the schema `Type` as-is.

PR #5246, opened 2024-09-13
EES-4489: Add endpoint to remove all themes created by UI tests

tomjonesdev:
Robot UI tests generate themes and topics as part of their process, which are then removed during the teardown phase. This phase uses an existing endpoint which accepts an ID to target the removal of an individual resource.

UI test execution occasionally fails, or is cancelled prematurely, meaning that the teardown phase isn't triggered. This results in test resources remaining in the system and causing bloat.

This PR adds an endpoint which removes all themes and associated topics created during UI tests in one go, using the same logic to identify tests as is used for individual removal. The intended use for this endpoint is as part of a manual clean-up process when a notification is received regarding test execution failure, or eventually implemented into an automated process.

_The PR also includes a few minor recommended code style amendments._

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ThemeServicePermissionTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ThemeServicePermissionTests.cs
@@ -1,16 +1,17 @@
-using System;
-using System.Threading.Tasks;
 using AutoMapper;
 using GovUk.Education.ExploreEducationStatistics.Admin.Security;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Admin.ViewModels;
+using GovUk.Education.ExploreEducationStatistics.Common.Services;
 using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces.Security;
 using GovUk.Education.ExploreEducationStatistics.Common.Utils;
 using GovUk.Education.ExploreEducationStatistics.Content.Model;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
 using Microsoft.Extensions.Configuration;
 using Moq;
+using System;
+using System.Threading.Tasks;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.MapperUtils;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.PermissionTestUtil;
 using static GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils.MockUtils;
@@ -19,7 +20,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
 {
     public class ThemeServicePermissionTests
     {
-        private readonly Theme _theme = new Theme
+        private readonly Theme _theme = new()
         {
             Id = Guid.NewGuid()
         };
@@ -224,14 +225,23 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
                 );
         }
 
-        private static Mock<IConfiguration> DefaultConfigurationMock()
+        [Fact]
+        public async Task DeleteUITestThemes()
         {
-            var mock = new Mock<IConfiguration>(MockBehavior.Strict);
-            mock
-                .Setup(x => x.GetSection(It.IsAny<String>()))
-                .Returns(new Mock<IConfigurationSection>().Object);
-            return mock;
+            await PolicyCheckBuilder()
+                .SetupCheck(SecurityPolicies.CanManageAllTaxonomy, false)
+                .AssertForbidden(
+                    async userService =>
+                    {
+                        var service = SetupThemeService(userService: userService.Object);
+
+                        return await service.DeleteUITestThemes();
+                    }
+                );
         }
+
+        private static Mock<IConfiguration> DefaultConfigurationMock()
+            => CreateMockConfiguration(CollectionUtils.TupleOf("enableThemeDeletion", "true"));
 
         private ThemeService SetupThemeService(
             ContentDbContext context = null,
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ThemeServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ThemeServiceTests.cs
@@ -1,8 +1,4 @@
 #nullable enable
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Threading.Tasks;
 using AutoMapper;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces;
@@ -13,10 +9,14 @@ using GovUk.Education.ExploreEducationStatistics.Common.Tests.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Utils;
 using GovUk.Education.ExploreEducationStatistics.Content.Model;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
+using Microsoft.EntityFrameworkCore;
 using Moq;
-using static GovUk.Education.ExploreEducationStatistics.Admin.Validators.ValidationErrorMessages;
+using System;
+using System.Linq;
+using System.Threading.Tasks;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.DbUtils;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.MapperUtils;
+using static GovUk.Education.ExploreEducationStatistics.Admin.Validators.ValidationErrorMessages;
 using static GovUk.Education.ExploreEducationStatistics.Common.Services.CollectionUtils;
 using static GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils.MockUtils;
 using static Moq.MockBehavior;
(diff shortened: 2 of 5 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ThemeService.cs
mmyoungman, line 1: I believe the UI test are currently creating new topics for every test run. Here is an example of a topic on dev created last Friday and not cleaned up:
image

The code for this is in `robot-tests/admin_api.py` in `create_test_topic`.

I don't think changing this should be too difficult though, as I believe all publications we create include `%{RUN_IDENTIFIER}` and so should be unique for each run (if a test suite didn't include the run id in the name, you could add it) so all created publications for all test runs can live under the same topic.

In `robot-tests/admin_api.py`, I think you need to change `topic_name` so it doesn't include the `run_id`. I'd be inclined to set it as a global variable `test_topic_name = "UI test topic` alongside `test_theme_name` at the top of the file.

You could then copy the pattern for the theme, where it checks for a `test_theme_id` first and creates it if it doesn't exist.

I'd also change the comment in `run_tests.py` where we call `create_test_topic` so it reflects that we're just making sure the theme/topic are created - not creating a new one each test run.

mmyoungman (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ThemeService.cs
```diff
@@ -159,6 +160,40 @@ public async Task<Either<ActionResult, Unit>> DeleteTheme(Guid themeId)
                 });
         }
 
+        public async Task<Either<ActionResult, Unit>> DeleteUITestThemes(CancellationToken cancellationToken = default)
+        {
+            return await _userService.CheckCanManageAllTaxonomy()
+                .OnSuccess(_ => _context.Themes
+                    .Where(theme => theme.Title.StartsWith("UI test theme") || theme.Title.StartsWith("Seed theme"))
```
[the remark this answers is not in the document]
tomjonesdev: I'm not super thrilled about deleting things based on the _expectation_ that they'll match a specifically formatted `Title`, however this behaviour has remained unchanged, and I'd suggest that it would be better to remove outliers manually, rather than muddying the code with more string-based conditions.

Presumably these outliers were created either by an error or a non-standard approach, which has since been rectified, as the recent examples match the expected format. The ticket includes a criteria to tidy up dev, and I'm happy to remove these outliers manually.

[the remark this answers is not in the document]
tomjonesdev: Actually I've just realised that the existing condition, `UI test theme`, matches almost no themes on `dev`, as the majority use the format "UI Test - Test theme - <timestamp> <some id>".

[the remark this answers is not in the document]
mmyoungman: I think we allowed deleting of a Seed theme in case they get messed up or something. But we wouldn't necessarily want to be forced to remove them alongside the UI test themes. 

So in the context of removing a specific theme we want to be able to remove any theme related to the UI tests or Seed data. But I think when cleaning up the UI test themes specifically, we won't want to remove the seed themes - just all the UI test themes.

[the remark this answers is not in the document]
mmyoungman: So my views on this changed overnight :)

I think we push the cleaning up of random junk themes on dev not related to cleaning up UI test themes to another ticket.

`Actually I've just realised that the existing condition, UI test theme, matches almost no themes on dev`

I think matching by StartsWith `UI test theme` is actually fine. 

I think maybe for now we ignore the state of dev and just focus on what themes the UI tests will create once all the removing-topics work is done. If we decide now that the UI tests will create themes staring with `UI test theme`, then the issue is resolved. It just needs to be clear to the person who does the work to change the UI tests to create one theme per test run to name the theme created per run `UI test theme - {runId}`.

The theme naming pattern you mention (`UI Test - Test theme - <timestamp> <some id>`), AFAIK, isn't being created by the UI tests anymore, so I think we just ignore that for now.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ThemeService.cs
mmyoungman: I now think this comment is wrong. When the work to remove topics has been done, we will be creating one theme per test run, I believe. I think it's ok for now for the UI tests to create one topic per run - it won't be especially hard to change this to one theme per run.

So feel free to ignore all this.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ThemeService.cs
mmyoungman, line 167: No action to be taken on this - but just noting that the performance tests also create a theme - but it is named "UI test theme - Performance tests" (found in `tests/performance-tests/src/tests/testData.ts`) and so would be cleaned up by this.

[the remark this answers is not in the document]
mmyoungman: Also discussed this on a call - I don't think there is a need to change this now. We just need to ensure that whoever does the work on the UI tests after topics are removed knows to name test themes created by UI test runs `UI test theme - {blah}`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ThemeService.cs
```diff
@@ -159,6 +160,40 @@ public async Task<Either<ActionResult, Unit>> DeleteTheme(Guid themeId)
...
+                        _context.Themes.Remove(theme);
+                    }
+
+                    await _context.SaveChangesAsync();
+                    await _publishingService.TaxonomyChanged();
+                });
+        }
+
+        private async Task<Either<ActionResult, Unit>> CheckCanDeleteThemes(IEnumerable<Theme> themes)
+        {
```
ntsim, line 186: Mentioned in previously Slack - we'll want to do something similar to the current theme deletion endpoint where this endpoint can only be ran if the environment has been configured to do so.

See the `ThemeService` for how we're doing this.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ThemeService.cs
```diff
@@ -159,6 +160,27 @@ public async Task<Either<ActionResult, Unit>> DeleteTheme(Guid themeId)
                 });
         }
 
+        public async Task<Either<ActionResult, Unit>> DeleteUITestThemes(CancellationToken cancellationToken = default)
+        {
+            return await _userService.CheckCanManageAllTaxonomy()
+                .OnSuccess(_ => _context.Themes
+                    .Where(theme => theme.Title.Contains("UI test"))
+                    .Include(theme => theme.Topics))
+                .OnSuccessVoid(async themes =>
```
mmyoungman, line 169: I think we need to check `_themeDeletionAllowed` here to prevent people removing themes on prod - the same as in `CheckCanDeleteThemes`.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ThemeServiceTests.cs
mmyoungman, line 1: We also want a test adding for `DeleteUITestThemes` in `ThemeServicePermissionTests.cs`.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ThemeServiceTests.cs
```diff
@@ -485,7 +485,71 @@ public async Task DeleteTheme_OtherThemesUnaffected()
                 VerifyAllMocks(topicService);
 
                 result.AssertRight();
-                Assert.Equal(otherTheme.Id, context.Themes.AsQueryable().Select(t => t.Id).Single());
+                Assert.Equal(otherTheme.Id, await context.Themes.AsQueryable().Select(t => t.Id).SingleAsync());
+            }
+        }
+
+        [Fact]
+        public async Task DeleteUITestThemes()
```
mmyoungman, line 493: I'd create an additional test `DeleteUITestThemes_DisallowedByConfiguration` - like the test for `DeleteTheme`.

mmyoungman (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ThemeServicePermissionTests.cs
```diff
@@ -224,11 +224,25 @@ public void DeleteTheme()
                 );
         }
 
+        [Fact]
+        public async Task DeleteUITestThemes()
+        {
+            await PolicyCheckBuilder()
+                .AssertForbidden(
```
mmyoungman, line 231: You also need the `SetupCheck` here like the test above.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ThemeService.cs
```diff
@@ -159,6 +160,29 @@ public async Task<Either<ActionResult, Unit>> DeleteTheme(Guid themeId)
                 });
         }
 
+        public async Task<Either<ActionResult, Unit>> DeleteUITestThemes(CancellationToken cancellationToken = default)
+        {
+            return !_themeDeletionAllowed
+                ? (Either<ActionResult, Unit>)new ForbidResult()
```
mmyoungman, line 166: Not asking for a change here - just spitballing - but it did take me a sec to realise where this check was taking place as I didn't expect it to be here - that could just be me. My style would be to put an if at the top to really shove it in the readers face, and remove a level of indenting from the below code - but I accept I'd then lose concision in return. I'm interested in what other devs prefer here.

I think checking `_themeDeletionAllowed` first thing makes more sense than what `DeleteTheme` does.

mmyoungman (changes requested):

mmyoungman (approved):

PR #5253, opened 2024-09-17
EES-5464 improvements

bennettstuart:
See [ticket 5464](https://dfedigital.atlassian.net/browse/EES-5464)  and [ticket 4723](https://dfedigital.atlassian.net/browse/EES-4723)
Following previous merge #5236,  
Notable changes include
- Add code theme to improve accessibility
- Add the previously forgotten rewrites to allow for nicer file download links

image

### DIFF
--- a/src/explore-education-statistics-frontend/src/modules/data-catalogue/components/DataSetFileUsage.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/data-catalogue/components/DataSetFileUsage.tsx
@@ -24,7 +24,7 @@ export default function DataSetFileUsage({
   tableToolLink,
   onDownload,
 }: Props) {
-  const downloadLink = `${process.env.CONTENT_API_BASE_URL}/data-set-files/${dataSetFileId}/download`;
+  const downloadLink = `${process.env.PUBLIC_URL}data-catalogue/data-set/${dataSetFileId}/csv`;
 
   return (
     <DataSetFilePageSection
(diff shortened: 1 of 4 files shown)


src/explore-education-statistics-frontend/src/modules/data-catalogue/components/DataSetFileUsage.tsx
```diff
@@ -24,7 +24,7 @@ export default function DataSetFileUsage({
   tableToolLink,
   onDownload,
 }: Props) {
-  const downloadLink = `${process.env.CONTENT_API_BASE_URL}/data-set-files/${dataSetFileId}/download`;
+  const downloadLink = `${process.env.PUBLIC_URL}data-catalogue/data-set/${dataSetFileId}/csv`;
```
mmyoungman, line 27: Just checking that we can rely on `PUBLIC_URL` to always have the slash - I know there are cases with env variables where sometimes they do and sometimes they don't.

All that said, might be easiest to let it be tested as it's an easy fix if it isn't right. The only way we get stung by this is if prod is different to every other environment, which you'd hope isn't the case.

mmyoungman (approved):

src/explore-education-statistics-frontend/src/modules/data-catalogue/components/DataSetFileUsage.tsx
bennettstuart: Good point, I've cleaned that up to ensure we have correct slashes when creating the URL

PR #5270, opened 2024-09-20
EES-5518 Add server URL to OpenAPI docs

ntsim:
This PR adds the server URL to the OpenAPI docs so that it can be used independently of where it's hosted. The URL is passed into the docs via the `AppSettings.HostUrl` app setting.

## Other changes

- Simplified `ApplicationInsightsOptions` to `AppInsightsOptions`
- Refactored public API `Startup` to use typed options instead of getting values directly from `IConfiguration`.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
@@ -173,6 +173,8 @@ public class Startup(IConfiguration configuration, IHostEnvironment hostEnvironm
 
         // Options
 
+        services.AddOptions<AppSettingsOptions>()
+            .Bind(configuration.GetRequiredSection(AppSettingsOptions.Section));
         services.AddOptions<ContentApiOptions>()
             .Bind(configuration.GetRequiredSection(ContentApiOptions.Section));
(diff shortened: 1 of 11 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
```diff
@@ -37,14 +37,21 @@ namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api;
...
 {
-    private readonly IConfiguration _miniProfilerConfig =
-        configuration.GetRequiredSection(MiniProfilerOptions.Section);
-    private readonly IConfiguration _openIdConnectConfig =
-        configuration.GetRequiredSection(OpenIdConnectOptions.Section);
-    private readonly IConfiguration _requestTimeoutConfig =
-        configuration.GetSection(RequestTimeoutOptions.Section);
-    private readonly IConfiguration _applicationInsightsConfig =
-        configuration.GetSection(ApplicationInsightsOptions.Section);
+    private readonly MiniProfilerOptions _miniProfilerOptions = configuration
```
duncan-at-hiveit, line 40: Nice simplification

duncan-at-hiveit (approved):

PR #5277, opened 2024-09-23
EES-XXXX - fixed failing Public API test which was expecting an inval…

duncan-at-hiveit:
image

This PR:
- fixes a failing Public API UI test by re-introducing a data file that is invalid for Public API import.
- refactors existing data file import keywords to reuse as much logic as possible.
- renames some test names in `public_api_restricted.robot` to avoid warnings about duplicate test names within a single suite.


N-moh (approved):

tests/robot-tests/tests/files/public-api-data-files/invalid-data-set.csv
```diff
@@ -0,0 +1,2 @@
+time_period,time_identifier,geographic_level,country_code,country_name,region_name,region_code,URN,old_la_code,la_name,Estab,school_name,LAEstab,OpenDate,sex_of_school,Phase-type_grouping,type_of_esstablishment,middle_school,all_through_school,hospital_school,Denomination,Denomination_Detail,admissions_policy,district_administrative_code,district_administrative_name,parlc_code,parlc_name,ward_code,ward_name,urban_rural,school_postcode,new_la_code,ethnicity,FSM,Language,headcount,percent_of_pupils
+201920,Academic year,Local authority,E92000001,England,East Midlands,E12000004,,856,Leicester,,,,,,Pupil referral unit,,,,,,,,,,,,,,,,E06000016,Asian - Pakistani,Total,Total,0,0
```
benoutram, line 2: Is it necessary to have all of the columns of this data set. I think the more columns there are, the less obvious it is to someone looking at it to realise why it's 'invalid'.

I think at the moment that this is a cut down version of the data set @saicharan2789 added originally, which has multiple issues with it, and while it could be possible that an analyst attempts to create an API data set from something like this, if we are trying to force the processing to fail, why don't we add a small dataset with just one known cause of failure into it?

mmyoungman: Might also be good to make it clear what the issue with data set is in the name of the file

duncan-at-hiveit: As discussed on Slack, this'll be tackled in a subsequent follow-up piece of work which'll allow us to remove the CSV

PR #5286, opened 2024-09-26
EES-5508 Add public API and docs URLs

ntsim:
This PR adds URLs for the public API and the docs site to the admin and public frontend.

## Other changes

- Simplified naming of `ConfigurationController` to `ConfigController` (and its respective view model).


infrastructure/parameters/dev.parameters.json
```diff
@@ -59,6 +59,9 @@
     "publicApiUrl": {
       "value": "dev.statistics.api.education.gov.uk"
     },
+    "publicApiDocsUrl": {
```
duncan-at-hiveit, line 62: Do you envision this changing per environment?  This could be a default value which can be overridden if necessary.

duncan-at-hiveit (approved):

infrastructure/parameters/dev.parameters.json
ntsim: I think we're likely to just have 2 environments for the docs site - dev and prod. 

Dev would be the GitHub hosted page and prod will be an Azure static site (when we get around to it).

I'll go with your suggestion of setting the GitHub page as the default anyway.

PR #5288, opened 2024-09-26
EES-2468 Added Ability To Delete User

rianski:
PR changes:
- Added ability to delete user on user management page
- Added frontend tests to BAU User Page (backend tests were already implemented for deleting a user)
- Updated `CreatedById` column in UserInvites table to set to null on deletion

delete user screenshot

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserManagementServicePermissionTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserManagementServicePermissionTests.cs
@@ -131,36 +131,12 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
             public async Task Success()
             {
                 await PolicyCheckBuilder<SecurityPolicies>()
+                    .SetupCheck(CanManageUsersOnSystem)
                     .AssertSuccess(async userService =>
                         {
                             var service = SetupUserManagementService(
-                                userService: userService.Object,
-                                enableDeletion: true);
+                                userService: userService.Object);
                             return await service.DeleteUser("ees-test.user@education.gov.uk");
-                        }
-                    );
-            }
-
-            [Fact]
-            public async Task Forbidden_EnvironmentConfiguration()
-            {
-                await PolicyCheckBuilder<SecurityPolicies>()
-                    .AssertForbidden(async userService =>
-                        {
-                            var service = SetupUserManagementService(userService: userService.Object);
-                            return await service.DeleteUser("ees.test-user@education.gov.uk");
-                        }
-                    );
-            }
-
-            [Fact]
-            public async Task Forbidden_InvalidEmailAddressFormat()
-            {
-                await PolicyCheckBuilder<SecurityPolicies>()
-                    .AssertForbidden(async userService =>
-                        {
-                            var service = SetupUserManagementService(userService: userService.Object);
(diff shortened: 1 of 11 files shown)


rianski:
It's not specified in the ticket but we potentially want the delete user ability to have a confirmation modal just in case delete is clicked on by mistake by a user.

src/explore-education-statistics-admin/src/services/userService.ts
```diff
@@ -129,6 +134,9 @@ const userService: UsersService = {
   updateUser(userId: string, update: UserUpdate): Promise<boolean> {
     return client.put(`/user-management/users/${userId}`, update);
   },
+  deleteUser(email: string): Promise<RemoveUser> {
```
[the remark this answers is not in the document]
rianski: Agree with you that there's not much bother in creating a new endpoint so removed the check. Good spot on that it's something I overlooked.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
```diff
@@ -474,12 +474,6 @@ private async Task<Either<ActionResult, Unit>> CheckCanDeleteUser(string email)
                 return new ForbidResult();
             }
```
mmyoungman, line 475: You will also need to delete this if statement - github is only letting me highlight the bottom two lines. This will prevent users from being deleted on prod.

`_userDeletionAllowed` basically the variable `enableThemeDeletion` renamed. `enableThemeDeletion` defaults to false in our main IoC / ARM template `infrastructure/templates/template.json`, but is set to be true for dev/test/preprod environments in `infrastructure/parameters/{env}.parameters.json`.

mmyoungman (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
rianski: Ah I see - is it worth just removing this method entirely and replacing the check with `CheckCanManageAllUsers`?

mmyoungman (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserManagementServicePermissionTests.cs
```diff
@@ -131,39 +131,15 @@ public class DeleteUserTests
...
-                    .AssertForbidden(async userService =>
-                        {
-                            var service = SetupUserManagementService(userService: userService.Object);
-                            return await service.DeleteUser("invalid-email-to-delete@education.gov.uk");
-                        }
-                    );
-            }
         }
 
         private static UserManagementService SetupUserManagementService(
```
mmyoungman, line 145: Now `appOptions` is gone, we need to remove it from here.

mmyoungman (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserManagementServicePermissionTests.cs
rianski: weirdly didn't complain about this locally when running the tests, it's gone now.

mmyoungman:
I just squashed the commits and put EES-2468 into the commit message

mmyoungman (approved):

PR #5295, opened 2024-09-30
EES-5539 Fix excessively large `CodeBlock` dependency bundle

ntsim:
This PR fixes the `CodeBlock` component importing the entirety of Highlight.js and creating excessively large bundles.

We've now stripped this down to the bare minimum and only import in the languages that we need (i.e. R and Python).

As part of this, we've also swapped out the current colour theme for the same one that's used by the GOV.UK Design System (which appears to be based on GitHub's). We've also made some improvements to make most of the colours WCAG AAA colour contrast compliant.

The colours now look like:

image

image

## Related changes

- Minor tweaks to `CodeBlock` props to improve type-safety and replaced `code` prop with `children` to be more idiomatic.
- Added `build:analyze` script that can be ran from the frontend project to analyze its bundles using `@next/bundle-analyzer`.


bennettstuart (approved):
Looks good, just check with Nusrath before merging (today)

PR #5296, opened 2024-10-01
EES-5402 - ability to select geoJson per dataset

bennettstuart:
[EES-5402](https://dfedigital.atlassian.net/browse/EES-5402)

This branch is to be the main trunk for this feature, allowing multiple branches to be based from this and to be worked on separately.


bennettstuart:
to be merged with EES-5541

PR #5301, opened 2024-10-01
EES-5540 - fixed bug with Pabot and the Robot --rerunfailedsuites fla…

duncan-at-hiveit:
This PR:
- Fixes a bug between Pabot and Robot when using the `--rerunfailedsuites` option.
- Improves the ability to see failures on different rerun attempts rather than the merged  report which loses details.
- Makes the test run process less verbose by decreasing the severity of several logging points in the Python code.
- Makes finding failure screenshots easier to see in the test reports by only having to open up 1 element in the Robot test report at the failure point.
- Swaps the behaviour for failing subsequent tests to a first failing test in a suite for skipping the tests instead, to make locating the first failure in the chain more obvious.


N-moh (approved):

tests/robot-tests/test_runners.py
```diff
@@ -0,0 +1,110 @@
...
+    # We want to add arguments on the first rerun attempt, but on subsequent attempts, we just want
+    # to change rerunfailedsuites xml file we use
+    robot_args += ["--output", "output.xml"]
+
+    return robot_args
+
+
+def process_includes_and_excludes(robot_args: [], arguments: argparse.Namespace):
+    if arguments.reseed:
+        robot_args += ["--include", "SeedDataGeneration"]
```
luke-h1, line 62: **question** just wondering what `SeedGeneration` is responsible for? does it generate test data for test suites etc. ?

tests/robot-tests/test_runners.py
```diff
@@ -0,0 +1,110 @@
...
+        if path_to_previous_report_file is not None:
+            robot_args += ["--rerunfailedsuites", path_to_previous_report_file]
+        
+        robot_args += [arguments.tests]
+
+        robot.run_cli(robot_args, exit=False)
+
+    elif arguments.interp == "pabot":
+
+        logger.info('Performing test run with Pabot')
```
luke-h1, line 94: **question** not a requirement of course but just wondering if it's possible to display the arguments that `pabot` was instantiated with here? (for example: number of processes, concurrency etc.)  - imagine this could be useful in CI environments

tests/robot-tests/tests/libs/fail_fast.py
```diff
@@ -46,32 +64,37 @@ def record_failing_test_suite():
...
+
+
+def _capture_html():
+    html = sl().get_source()
+    current_time_millis = round(datetime.datetime.timestamp(datetime.datetime.now()) * 1000)
+    output_dir = BuiltIn().get_variable_value('${OUTPUT DIR}')
+    html_file = open(f"{output_dir}{os.sep}captured-html-{current_time_millis}.html", "w", encoding="utf-8")
+    html_file.write(html)
+    html_file.close()
+    logger.info(f"Captured HTML of {sl().get_location()}      HTML saved to file://{os.path.realpath(html_file.name)}")
```
luke-h1, line 89: **nitpick** Just wondering what the spacing between `{sl().get_location()` to `HTML` is for here? I'm assuming it's for CI but not too sure :)

tests/robot-tests/tests/libs/fail_fast.py
```diff
@@ -46,32 +64,37 @@ def record_failing_test_suite():
...
+
 
+def _prompt_to_continue():
+    logger.warn("Continue? (Y/n)")
+    choice = input()
+    if choice.lower().startswith("n"):
```
luke-h1, line 95: **question** Do we need to add a condition for when people press `y` here?

tests/robot-tests/test_runners.py
```diff
@@ -0,0 +1,110 @@
...
+    if arguments.tags:
+        robot_args += ["--include", arguments.tags]
+    if arguments.print_keywords:
+        robot_args += ["--listener", "listeners/KeywordListener.py"]
+    if arguments.ci:
+        # NOTE(mark): Ensure secrets aren't visible in CI logs/reports
+        robot_args += ["--removekeywords", "name:operatingsystem.environment variable should be set"]
+        robot_args += ["--removekeywords", "name:common.user goes to url"]  # To hide basic auth credentials
+    
+    process_includes_and_excludes(robot_args, arguments)
```
luke-h1, line 40: **nitpick** super small nitpick but just wondering if you think something such as `process_environment_args` or something along those lines could be a bit more descriptive for consumers?

tests/robot-tests/reports.py
```diff
@@ -0,0 +1,109 @@
...
+        suite_stats = report.find_all('stat', recursive=True)
+        [suite_stat.extract() for suite_stat in suite_stats if suite_stat.get('id') in passing_suite_ids]
+    
+        with open(path_to_filtered_report, "a") as filtered_file:
+            filtered_file.write(report.prettify())
+
+        
+def _get_passing_suite_ids_from_report(report: BeautifulSoup) -> []: 
+    suite_results = report.find('statistics').find('suite').find_all('stat')
+    passing_suite_results = [suite_result for suite_result in suite_results if int(suite_result.get('fail')) == 0]
```
luke-h1, line 99: **praise** very nice 😎

tests/robot-tests/test_runners.py
duncan-at-hiveit: Why hello there @luke-h1 !

It does indeed.  Lets us generate consistent seed data in each environment, and regenerate as needed.  It's a manual job currently to reseed an environment as we don't need it often.

tests/robot-tests/tests/libs/fail_fast.py
duncan-at-hiveit: Yeah, it's basically that!  Separate a large HTTP URL from another large file URL in the logs to make it easier to read.

tests/robot-tests/reports.py
duncan-at-hiveit: Why thank you thank you :D  Handy to have all that pre-existing BeautifulSoup code to use as a reference of course!

tests/robot-tests/tests/libs/fail_fast.py
duncan-at-hiveit: No, so long as they don't say no to continuing, the code will then stop blocking at this point and resume execution.

tests/robot-tests/test_runners.py
duncan-at-hiveit: Hey, so if you're referring to the `process_includes_and_excludes` method name, this is purely for adding the arguments for which tests to include and which to exclude, using the `--include` and `--exclude` options only, hence the name.  I could've left it inline as I only extracted it into its own method for testing some behaviour, but I decided to leave it separate as it does have quite complex logic that I felt was OK to be off on its own.

I will rename it to `_create_include_and_exclude_args` though to make it more private, as it shouldn't be getting used outside of this file really, and got it to return an array of args rather than amending the passed-in args directly.  Thus:

```
    robot_args += _create_include_and_exclude_args(arguments)
```

duncan-at-hiveit: And also put the:

```
if arguments.tags:
        robot_args += ["--include", arguments.tags]
```

part in there too.

duncan-at-hiveit: And:

```
        "--exclude",
        "Failing",
        "--exclude",
        "UnderConstruction",
        "--exclude",
        "VisualTesting",
```

tests/robot-tests/test_runners.py
duncan-at-hiveit: Nice suggestion.  Just including processes for now:

```
2024-10-02 10:32:40,653: Performing test run with Pabot (8 processes)
```

tests/robot-tests/run_tests.py
```diff
@@ -217,16 +141,26 @@ def run():
...
+                # If re-running failed suites, get the appropriate previous report file from which to determine which suites failed.
+                # If this is run 2 or more when using the `--rerun-attempts n` option, the path to the previous report will be in the
+                # previous run attempt's folder.
+                # If this is a rerun using the `--rerun-failed-suites` option, the path to the previous report will be in the main 
+                # test results folder directly.
                 if rerunning_failed_suites:
-                    logger.info(f"Merging results from test run {test_run_index + 1} with previous run's report")
-                    merge_test_reports()
+                    if test_run_index == 0:
+                        previous_report_file = f"{robot_tests_folder}{os.sep}{main_results_folder}{os.sep}output.xml"
```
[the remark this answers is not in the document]
duncan-at-hiveit: Discussed on a huddle and thinking that it's not the same issue here

PR #5302, opened 2024-10-01
EES-5497 Add API data set preview token guidance

ntsim:
This PR adds the preview token guidance for API data sets. This looks like:

image

## Other changes

- Fixed various issues with `UrlContainer` and `CopyTextButton`:
  - Makes `label` required for `UrlContainer` so that the copy input is labelled correctly by the consumer.
  - Allowes inline styling to be correctly toggled using the `inline` prop for `UrlContainer`.
  - Changes `UrlContainer` and `CopyTextButton` to fill the parent's space as much as possible instead of being limited to 500px.
  - Fixed incorrect flex styling causing the input in `CopyTextButton` to not take up the correct amount of space.
  - Adds the GOV.UK focus style to the `UrlContainer` input.
  - Improves how `CopyTextButton` responds at mobile / tablet.
  - Updates `CopyTextButton` to show the `confirmMessage` in the button itself rather than in a separate line. 
- Updated `CopyLinkModal` to be wider to show of the link

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/data/ReleaseApiDataSetPreviewTokenPage.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/data/ReleaseApiDataSetPreviewTokenPage.tsx
@@ -11,11 +11,14 @@ import previewTokenQueries from '@admin/queries/previewTokenQueries';
 import apiDataSetQueries from '@admin/queries/apiDataSetQueries';
 import previewTokenService from '@admin/services/previewTokenService';
 import { useLastLocation } from '@admin/contexts/LastLocationContext';
+import CodeBlock from '@common/components/CodeBlock';
 import LoadingSpinner from '@common/components/LoadingSpinner';
 import Button from '@common/components/Button';
 import CopyTextButton from '@common/components/CopyTextButton';
 import FormattedDate from '@common/components/FormattedDate';
 import ModalConfirm from '@common/components/ModalConfirm';
+import Tabs from '@common/components/Tabs';
+import TabsSection from '@common/components/TabsSection';
 import ApiDataSetQuickStart from '@common/modules/data-catalogue/components/ApiDataSetQuickStart';
 import { useQuery } from '@tanstack/react-query';
 import { generatePath, useHistory, useParams } from 'react-router-dom';
@@ -75,6 +78,8 @@ export default function ReleaseApiDataSetPreviewTokenPage() {
     );
   };
 
+  const tokenExampleUrl = `${publicApiUrl}/api/v1.0/data-sets/${dataSet?.draftVersion?.id}`;
+
   return (
     <>
       <Link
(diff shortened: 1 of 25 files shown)


src/explore-education-statistics-admin/src/pages/release/data/ReleaseApiDataSetPreviewTokenPage.tsx
```diff
@@ -92,51 +97,93 @@ export default function ReleaseApiDataSetPreviewTokenPage() {
...
-                  text={previewToken.id}
-                />
+
+                <h3>Using the preview token</h3>
+
                 <p>
-                  Please revoke the token as soon as you have finished checking
-                  the API data set.
+                  To use the preview token, add it to your request using a{' '}
+                  <code>Preview-Token</code> header.
```
bennettstuart, line 127: I'm not sure about the use of the `<code>` here, as while this is referencing a code related property name it isn't actually code. The following section demonstrates how to use this header too so it may be redundant. My main concern is that there isn't a clear rule when to and when not to use this markup when referencing code variables, which may lead to some inconsistencies.

This is just my opinion, happy for you to take the steer on this just wanted to raise the potential inconsistency issue

src/explore-education-statistics-common/src/components/CopyTextButton.tsx
```diff
@@ -3,89 +3,70 @@ import styles from '@common/components/CopyTextButton.module.scss';
...
+      <Button
+        className={classNames(styles.button, {
+          'govuk-!-margin-right-2': !inline,
+        })}
+        onClick={async () => {
+          await navigator.clipboard.writeText(text);
+          toggleCopied.on();
+        }}
+      >
+        <span aria-live="polite">{copied ? confirmMessage : buttonText}</span>
```
bennettstuart, line 68: I think only the confirmMessage should be wrapped with an aria-live, as buttonText need not be notifed to users, I think the current setup would announce the confirm message, then announce the buttonText 5 seconds later

bennettstuart (approved):

src/explore-education-statistics-admin/src/pages/release/data/ReleaseApiDataSetPreviewTokenPage.tsx
ntsim: Think it's acceptable to use this here as we're referencing a technical term that needs to stand out.

We've only just added styling for `code` in the last couple of months so there's certainly going to be inconsistencies as we start to use it in more places. 

I think the general rule of thumb should be:

- is this a technical term?
- does it need to stand out to the user?

These are criteria that are being met a lot more by the public API work as it's much closer to the metal than other areas of the service.

src/explore-education-statistics-common/src/components/CopyTextButton.tsx
ntsim: Good point! I've updated this to use `ScreenReaderMessage` instead (which should handle the behaviour correct:

image

src/explore-education-statistics-admin/src/pages/release/data/ReleaseApiDataSetPreviewTokenPage.tsx
bennettstuart: Fair point!

PR #5308, opened 2024-10-04
EES-5541 - Boundary level per dataset form changes

bennettstuart:
## Overview
Most changes are hidden behind a feature flag.

This work introduces the concept of boundaryLevels on dataSetConfigs on maps, this is to allow maps to show different geoJson per dataset as to produce a more accurate representation of data alongside their boundaries on the map.

- Reducer actions altered to cater for dataset boundary levels
- Separated Chart boundary levels form data setup and boundary levels form into their own components for simplicity.
- (hidden for now) data set table with boundary level select, below existing boundary level select, to allow users to select boundary levels per data set
- Refactor to chartBuilderReducer:
  - Hoists initialization of map dataSetConfigs into chartBuilderReducer
  - Refactors form types for boundary level and data grouping tabs
  - Fixes boundary level tab not showing invalid form errors for other tabs
  - Refactors out new `getMapDataSetCategoryConfigs` function from `getDataSetCategoryConfigs`

### Additional
- Added `onChange` prop to our Form component
  - `<Effect value={formValues} onChange={props.onChange} />` no longer the only option to onChange form updates (and it doesn't work with form array field)
- Fixed a bug on Chart data groupings tab not allowing modal to close

image


ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/__tests__/ChartDataGroupingsConfiguration.test.tsx
```diff
@@ -337,29 +259,10 @@ describe('ChartDataGroupingsConfiguration', () => {
     ).toBeInTheDocument();
   });
 
-  test('renders correctly when data sets are removed', () => {
```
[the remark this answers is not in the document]
bennettstuart: I've added: ` test('making a form change and submitting the form, ...` to the bottom of the file

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/__tests__/ChartBoundaryLevelsConfiguration.test.tsx
```diff
@@ -144,62 +150,79 @@ describe('ChartBoundaryLevelsConfiguration', () => {
...
+        if (optionIndex === 0) {
+          expect(option).toHaveTextContent('Use default');
+          expect(option).toHaveValue('');
+        } else {
+          const { id, label } = testMeta.boundaryLevels[optionIndex - 1];
+          expect(option).toHaveTextContent(label);
+          expect(option).toHaveValue(String(id));
+        }
+      });
+    });
```
ntsim, line 175: Wouldn't advise using for loops with conditional logic like this in test cases. We want things to be as clear as possible even at the expense of a bit of copy and paste.

Keep things simple and just check things using indices.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/__tests__/ChartBoundaryLevelsConfiguration.test.tsx
```diff
@@ -119,23 +120,28 @@ describe('ChartBoundaryLevelsConfiguration', () => {
...
+    defaultBoundaryLevelOptions.forEach((option, index) => {
+      if (index === 0) {
+        expect(option).toHaveTextContent('Please select');
+        expect(option).toHaveValue('');
+      } else {
+        const { id, label } = testMeta.boundaryLevels[index - 1];
+        expect(option).toHaveTextContent(label);
+        expect(option).toHaveValue(String(id));
+      }
+    });
```
ntsim, line 139: Wouldn't advise using for loops with conditional logic like this in test cases. We want things to be as clear as possible even at the expense of a bit of copy and paste.

Keep things simple and just check things using indices.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/__tests__/ChartBoundaryLevelsConfiguration.test.tsx
```diff
@@ -64,6 +64,18 @@ describe('ChartBoundaryLevelsConfiguration', () => {
       timePeriod: '2015_AY',
     },
   ];
+  const testDefaultMap = {
```
ntsim, line 67: Would add an explicit type to avoid any ambiguity about what this should be.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/__tests__/ChartBoundaryLevelsConfiguration.test.tsx
```diff
@@ -144,62 +150,79 @@ describe('ChartBoundaryLevelsConfiguration', () => {
...
         }}
         onChange={noop}
         onSubmit={noop}
         hasDataSetBoundaryLevels
       />,
     );
 
-    expect(screen.getByLabelText('Default boundary level')).toHaveValue('2');
+    expect(screen.getByLabelText('Default boundary level')).toHaveValue(
+      String(testMeta.boundaryLevels[2].id),
```
ntsim, line 207: Nitpick, but you could use the `toString` method to convert this to a string. It personally reads a bit nicer, but it doesn't massively make a difference in this case.

Similar could be done below

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/__tests__/ChartDataGroupingsConfiguration.test.tsx
```diff
@@ -321,4 +321,57 @@ describe('ChartDataGroupingsConfiguration', () => {
...
+    await waitFor(() => {
+      expect(handleChange).toHaveBeenCalledTimes(1);
+    });
+
+    await user.click(
+      screen.getByRole('button', { name: 'Save chart options' }),
+    );
+
+    await waitFor(() => {
+      expect(handleSubmit).toHaveBeenCalledTimes(1);
```
ntsim, line 374: We should check what `handleSubmit` has been called with

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/__tests__/ChartDataGroupingsConfiguration.test.tsx
```diff
@@ -321,4 +321,57 @@ describe('ChartDataGroupingsConfiguration', () => {
...
+    await user.click(screen.getAllByRole('button', { name: 'Edit' })[0]);
+
+    expect(await screen.findByRole('dialog')).toBeInTheDocument();
+    expect(screen.getByText('Edit groupings')).toBeInTheDocument();
+
+    await user.click(screen.getByLabelText('Quantiles'));
+    await user.click(screen.getByRole('button', { name: 'Done' }));
+
+    await waitFor(() => {
+      expect(handleChange).toHaveBeenCalledTimes(1);
```
ntsim, line 366: We should check `handleChange` has been called with

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/reducers/__tests__/chartBuilderReducer.test.ts
```diff
@@ -625,14 +614,106 @@ describe('chartBuilderReducer', () => {
...
+          data: [],
+          meta: testFullTable.subjectMeta,
+          tableTitle: undefined,
+        }),
+      )(initialState, action);
+
+      expect(nextState.options?.boundaryLevel).toEqual(2);
+      expect(nextState.map?.dataSetConfigs[0]?.boundaryLevel).toEqual(1);
+    });
+  });
```
ntsim, line 657: It'd be a good idea to have another test case that checks data set boundary levels don't overwrite any other other part of the data set (i.e. the data grouping)

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/reducers/__tests__/chartBuilderReducer.test.ts
```diff
@@ -625,14 +614,106 @@ describe('chartBuilderReducer', () => {
...
+          meta: testFullTable.subjectMeta,
+          tableTitle: undefined,
+        }),
+      )(initialState, action);
+
+      expect(nextState.map?.dataSetConfigs[0]?.dataGrouping).toEqual(
+        testDataGrouping,
+      );
+    });
+  });
```
ntsim, line 708: Similar to `UPDATE_MAP_BOUNDARY_LEVELS`, we should have another test case to check that the update doesn't overwrite any other part of the data set (i.e. the boundary level)

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/reducers/__tests__/chartBuilderReducer.test.ts
```diff
@@ -1071,5 +1152,74 @@ describe('chartBuilderReducer', () => {
...
+        type: 'map',
+        height: 300,
+        title: 'Chart title',
+        alt: '',
+        boundaryLevel: 1,
+      };
+
+      const { result } = renderHook(() =>
+        useChartBuilderReducer({
+          data: [],
```
ntsim, line 1183: We're currently not testing anything of value as we're not generating any `map.dataSetConfigs`.

Looks like we need to provide some data so that we can generate the initial map data set configs and assert that these are as expected.

src/explore-education-statistics-common/src/components/form/__tests__/Form.test.tsx
```diff
@@ -141,6 +142,38 @@ describe('Form', () => {
...
+          <FormFieldTextInput label="first name" name="firstName" />
+          <button type="submit">Submit</button>
+        </Form>
+      </FormProvider>,
+    );
+
+    const firstName = 'first name';
+
+    await user.type(
+      screen.getByRole('textbox', { name: 'first name' }),
```
ntsim, line 168: Don't need to use a role selector - `getByLabel` will do and is faster.

src/explore-education-statistics-common/src/components/form/__tests__/Form.test.tsx
```diff
@@ -141,6 +142,38 @@ describe('Form', () => {
...
+        initialValues={{
+          firstName: 'Firstname',
+        }}
+        validationSchema={Yup.object({
+          firstName: Yup.string().defined(),
+        })}
+      >
+        <Form id="test-form" onChange={handleChange} onSubmit={jest.fn()}>
+          The form
+          <FormFieldTextInput label="first name" name="firstName" />
```
ntsim, line 159: Nitpick, but would just capitalise 'First name'

src/explore-education-statistics-common/src/components/form/__tests__/Form.test.tsx
```diff
@@ -141,6 +142,38 @@ describe('Form', () => {
...
+    const firstName = 'first name';
+
+    await user.type(
+      screen.getByRole('textbox', { name: 'first name' }),
+      firstName,
+    );
+
+    await waitFor(() => {
+      expect(handleChange).toHaveBeenCalledTimes(firstName.length);
+    });
```
ntsim, line 174: Can we do a final check that the `handleChange` has been called with the expected value? Probably makes sense to do this after the wait.

ntsim (changes requested):

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/__tests__/ChartDataGroupingsConfiguration.test.tsx
bennettstuart: Added test `Data groupings tab edits submit with correct values` to **ChartBuilder.test.tsx**

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/__tests__/ChartBoundaryLevelsConfiguration.test.tsx
bennettstuart: will bare in mind for future!

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/__tests__/ChartBoundaryLevelsConfiguration.test.tsx
bennettstuart: I've simplified the code to be more obvious, stating that the options should match up with the supplied boundary levels in the data - I've removed the conditional in the loop. I'll take this forward, keeping tests simple and obvious, avoiding loops with the idea saving lines isn't worth the complexity in tests.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/__tests__/ChartBoundaryLevelsConfiguration.test.tsx
bennettstuart: as above 
I've simplified the code to be more obvious, stating that the options should match up with the supplied boundary levels in the data - I've removed the conditional in the loop. I'll take this forward, keeping tests simple and obvious, avoiding loops with the idea saving lines isn't worth the complexity in tests.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/__tests__/ChartBoundaryLevelsConfiguration.test.tsx
```diff
@@ -63,75 +90,180 @@ describe('ChartBoundaryLevelsConfiguration', () => {
...
+      expect(option).toHaveValue(String(id));
+    });
+
+    expect(
+      screen.getByText('Set boundary levels per data set'),
+    ).toBeInTheDocument();
+    expect(screen.queryByRole('table')).toBeInTheDocument();
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [headerRow, ...dataRows] = screen.getAllByRole('row');
```
ntsim, line 148: Don't destructure like this. It's not good code hygiene to leave unused variables as we're unnecessarily assigning variables to memory for no reason, hence the lint error.

Seems like part of the problem is that we're still trying to be too smart with the for loop at the end. Similar to above comment, I wasn't suggest that we keep the for loop.

Would suggest just ditching the destructuring completely and doing something like this:

```ts
const rows = screen.getAllByRole('row');

expect(rows).toHaveLength(3);

const row1Cells = within(rows[1]).getAllByRole('cell');

expect(row1Cells[0]).toHaveTextContent(
  'Number of authorised absence sessions (Ethnicity Major Chinese, State-funded primary, All locations, 2014/15)',
);
expect(within(row1Cells[1]).getByRole('combobox')).toHaveValue('');

const row1BoundaryLevels = within(rows1Cells[1]).getAllByRole('option');
expect(row1BoundaryLevels[0]).toHaveTextContent('Use default');
expect(row1BoundaryLevels[0]).toHaveValue('');
expect(row1BoundaryLevels[1]).toHaveTextContent('Boundary level 1');
expect(row1BoundaryLevels[1]).toHaveValue('1');
expect(row1BoundaryLevels[2]).toHaveTextContent('Boundary level 2');
expect(row1BoundaryLevels[2]).toHaveValue('2');
expect(row1BoundaryLevels[3]).toHaveTextContent('Boundary level 3');
expect(row1BoundaryLevels[3]).toHaveValue('3');

const row2Cells = within(rows[2]).getAllByRole('cell');

// etc for row 2...
```

Again, this is a bit of copy and paste - but it's clear and precise.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/reducers/__tests__/chartBuilderReducer.test.ts
```diff
@@ -966,5 +1152,371 @@ describe('chartBuilderReducer', () => {
...
+            ],
+            referenceLines: [],
+            visible: true,
+            label: {
+              text: '',
+            },
+          },
+        },
+        definition: mapBlockDefinition,
+        map: { dataSetConfigs: [] },
```
ntsim, line 1209: Left a previous comment that we're not testing anything of value currently because we're not actually calculating any of the map's initial `dataSetConfigs`. Seems like that's still the case.

Seems like the problem might be because the data we're supplying doesn't actually match up to the data sets we've configured in `axes.major`.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/reducers/__tests__/chartBuilderReducer.test.ts
```diff
@@ -966,5 +1152,371 @@ describe('chartBuilderReducer', () => {
...
+            },
+            boundaryLevel: 30,
+          },
+        ],
+      });
+      rerender();
+
+      expect(result.current.state).toEqual<ChartBuilderState>({
+        ...expectedChartBuilderStateProps,
+        map: {
```
ntsim, line 1350: Would suggest to simplify this test and make it quicker to read, we don't re-check the entire chart builder state.

We can just check the specific parts of the state we're actually interested in:

```ts
expect(result.current.state.options.boundaryLevel).toBe(...);
expect(result.current.state.map).toEqual<MapConfig>({
  // ...
});
```

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/reducers/__tests__/chartBuilderReducer.test.ts
```diff
@@ -966,5 +1152,371 @@ describe('chartBuilderReducer', () => {
...
+        },
+        options: {
+          alt: '',
+          height: 300,
+          subtitle: '',
+          title: 'Chart title',
+          titleType: 'alternative',
+          boundaryLevel: 10,
+        },
+      };
```
ntsim, line 1314: Was going to suggest moving this closer to where the assertion is, but as per other comment - we could just ditch this entirely in favour of checking the specific parts of state we're interested in.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/reducers/__tests__/chartBuilderReducer.test.ts
```diff
@@ -966,5 +1152,371 @@ describe('chartBuilderReducer', () => {
...
+        },
+        options: {
+          alt: '',
+          height: 300,
+          subtitle: '',
+          title: 'Chart title',
+          titleType: 'alternative',
+          boundaryLevel: 1,
+        },
+      };
```
ntsim, line 1463: As above - don't need this. Just check the specific parts of state we're interested in.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/reducers/__tests__/chartBuilderReducer.test.ts
```diff
@@ -966,5 +1152,371 @@ describe('chartBuilderReducer', () => {
...
+                timePeriod: '2014_AY',
+              },
+              dataGrouping: { type: 'Custom', customGroups: [] },
+              boundaryLevel: 30,
+            },
+          ],
+        },
+      });
+    });
+    test('setting data groupings does not change boundary levels', () => {
```
ntsim, line 1374: Line break between test cases

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/reducers/__tests__/chartBuilderReducer.test.ts
```diff
@@ -100,13 +103,25 @@ describe('chartBuilderReducer', () => {
         payload: testChartDefinition,
       };
 
-      const nextState = produce(chartBuilderReducer)(initialState, action);
+      const nextState = produce(
+        chartBuilderReducer({
+          data: [],
+          meta: testFullTable.subjectMeta,
+          tableTitle: undefined,
```
ntsim, line 110: Similar to an earlier comment about a similar property - we don't need to set `tableTitle` to `undefined` as it's optional.

There's a bunch of other places throughout this test suite we can remove this from as well.

ntsim (approved):
LGTM!

PR #5315, opened 2024-10-08
Ees 5526 part 2 - making date time handling code in Robot tests easier

duncan-at-hiveit:
# UI test results

## Local run (BST)

image

## Pipeline run (UTC)

image

The "Public Suite" tests are failing for an unrelated reason, because the back button in the table tool in the public site isn't doing what is expected. 

# Overview

This PR:
- simplifies the datetime wrangling code that we use in the Robot tests, by offering more targeted keywords.
- moves the default timezone of dates and times that are used for checking dates and times shown in the browser to be Europe/London by default, as that is the default timezone that EES displays dates in.
- adds specific keywords for getting individual date portions e.g. day, month, month word, year.
- splits keywords into "get london" and "get local browser" keywords, depending on whether we expect them to be displayed in Europe/London time or in time that is local to the browser.
- removes the need to use datetimes in the Robot tests when all we're interested in getting is date parts.
- removes a trailing forward slash from the UI test pipeline yml that was causing the file download name of UI test report failures to be invalid in the pipeline.

### DIFF
--- a/tests/robot-tests/tests/libs/dates_and_times.py
+++ b/tests/robot-tests/tests/libs/dates_and_times.py
@@ -1,6 +1,5 @@
 import datetime
 import os
-
 import pytz
 from tests.libs.selenium_elements import sl
 
--- a/tests/robot-tests/tests/libs/dates_and_times.py
+++ b/tests/robot-tests/tests/libs/dates_and_times.py
@@ -1,0 +1,50 @@
+import datetime
+import os
+
+import pytz
+from tests.libs.selenium_elements import sl
+
+
+def get_london_date(offset_days: int = 0, format_string: str = "%-d %B %Y") -> str:
+    return _get_date_and_time(offset_days, format_string, "Europe/London")
+
+
+def get_london_date_and_time(offset_days: int = 0, format_string: str = "%-d %B %Y") -> str:
+    return _get_date_and_time(offset_days, format_string, "Europe/London")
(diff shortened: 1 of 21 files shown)


tests/robot-tests/tests/libs/dates_and_times.py
```diff
@@ -0,0 +1,49 @@
...
+def get_london_month_date(offset_days: int = 0) -> str:
+    return get_london_date_and_time(offset_days, "%-m")
+
+
+def get_london_month_word(offset_days: int = 0) -> str:
+    return get_london_date_and_time(offset_days, "%B")
+
+
+def get_london_year(offset_days: int = 0) -> str:
+    return get_london_date_and_time(offset_days, "%Y")
```
mmyoungman, line 32: Not suggesting a change, but would be interested in discussing the pros and cons of this kind of approach with you at some point. 

I think my general inclination is the reverse of yours, which isn't to wrap things up like this, but in this case I can see it makes sense. But I've seen other instances where more complex things than this gets wrapped up with several layers of abstraction. If it breaks or needs updating, understanding what it is doing can be quite difficult, especially if it's already been tweaked a couple of times previously. In those cases I've tended to inline all the abstraction so I could understand it better, and I've frequently found that has allowed me to simplify things a lot.

mmyoungman (approved):

PR #5317, opened 2024-10-08
EES-5502: Replace WYSIWYG link text validation with pre-defined list of banned words.

tomjonesdev:
This PR updates the validation by removing the single-word rule, and updating the "click here" rule to instead check against a larger list of words which would be considered unacceptable from an accessibility standpoint.

image

### DIFF
--- a/src/explore-education-statistics-admin/src/components/editable/utils/getInvalidContent.ts
+++ b/src/explore-education-statistics-admin/src/components/editable/utils/getInvalidContent.ts
@@ -3,9 +3,8 @@ import parseNumber from '@common/utils/number/parseNumber';
 
 export interface InvalidContentError {
   type:
-    | 'clickHereLinkText'
     | 'repeatedLinkText'
-    | 'oneWordLinkText'
+    | 'badLinkText'
     | 'urlLinkText'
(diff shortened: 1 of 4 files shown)


src/explore-education-statistics-admin/src/components/editable/utils/getInvalidContent.ts
```diff
@@ -110,13 +102,64 @@ export default function getInvalidContent(
...
+      'learn more',
+      'link to',
+      'read more',
+      'this page',
+      'visit this',
+      'web page',
+      'web site',
+    ];
+
+    if (badLinkTextWords.includes(link.data?.toLowerCase()!)) {
```
bennettstuart, line 160: Unless this is already accounted for, we may also want to `trim` the link text to avoid allowing [ click here ]() being valid (with a leading or trailing space)

bennettstuart (approved):

src/explore-education-statistics-admin/src/components/editable/utils/getInvalidContent.ts
tomjonesdev: Good catch 👍

PR #5319, opened 2024-10-09
Ees 5560 adding public guidance notes to public page

duncan-at-hiveit:
# UI test results

image

# Overview

This PR:
- adds the public guidance notes for the changelog into the public dataset landing page in the changelog section.
- reworks a lot of the xpath text selectors in `common.robot` to be either exact or fuzzy based on `exact_match` argument.
- amends run_tests.py to report failing suites based on final `output.xml` rather than our own `.failing-suites` file.

## Reworking text selectors

During my aborted UI test writing for the public Changelog section (waiting for EES-5559 to go in), I ran into issues around not having a fuzzy text selector for some elements.  I needed `text()="Blah"` xpath to instead be `contains(., "Blah")` instead. On changing this though, this broke a whole lot of other scenarios instead of course!

I saw that we had a right old mix of styles in the common keywords and so I added an argument to allow the caller to specify the desired behaviour instead, which produces either a `text()="Blah"` or `contains(., "Blah")` depending on the choice.

I defaulted the behaviour almost entirely to be fuzzy, with the exception of checkboxes where we tend to be able to rely on exact text with little to no decoration (happy to be pushed back on this but it would mean a lot of checkbox assertions would need updating in the tests and would make them a bit more verbose).

## run_tests.py updates to use output.xml

During my battle with the text selector changes, I broke the Robot code a few times but noticed that misconfigured keywords and errors within our own keyword code didn't cause our tests to fail fast (nothing new, as the `run_on_failure` that we use to trigger failures only kicks in within SeleniumLibrary's own keywords as has always been the case).  This would mean that some suites ultimately failed but didn't get recorded by our fail-fast mechanism.

They were however captured in output.xml and so I changed the code up to use this as the ultimate source of truth instead, should we accidentally introduce keyword changes that break them but don't get caught by our fail-fast mechanism.

Ultimately to fix the fail-fast issue we could do with setting `run_on_failure=None` to prevent Selenium-only keywords failing suites fast, and instead in each test suite specify `Test Teardown    Run Keyword On Test Failure    Record failing suite` - the only issue here being that it doesn't catch failures in `Suite Setup`, so we'd need another mechanism for that, potentially with listeners.

### DIFF
--- a/src/explore-education-statistics-frontend/src/modules/data-catalogue/components/DataSetFileApiChangelog.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/data-catalogue/components/DataSetFileApiChangelog.tsx
@@ -6,13 +6,13 @@ import React from 'react';
 
 interface Props {
   changes: ApiDataSetVersionChanges;
-  publicGuidanceNotes?: string;
+  guidanceNotes: string;
   version: string;
 }
 
 export default function DataSetFileApiChangelog({
   changes,
-  publicGuidanceNotes,
+  guidanceNotes,
   version,
 }: Props) {
   const { majorChanges, minorChanges } = changes;
@@ -26,8 +26,8 @@ export default function DataSetFileApiChangelog({
       heading={pageSections.apiChangelog}
       id="apiChangelog"
     >
-      {publicGuidanceNotes && (
-        <p data-testid="public-guidance-notes">{publicGuidanceNotes}</p>
+      {guidanceNotes.length && (
+        <p data-testid="public-guidance-notes">{guidanceNotes}</p>
       )}
       <ApiDataSetChangelog
         majorChanges={changes.majorChanges}
--- a/src/explore-education-statistics-frontend/src/modules/data-catalogue/components/DataSetFileApiChangelog.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/data-catalogue/components/DataSetFileApiChangelog.tsx
@@ -6,10 +6,15 @@ import React from 'react';
 
 interface Props {
   changes: ApiDataSetVersionChanges;
+  publicGuidanceNotes?: string;
   version: string;
 }
 
-export default function DataSetFileApiChangelog({ changes, version }: Props) {
+export default function DataSetFileApiChangelog({
+  changes,
+  publicGuidanceNotes,
+  version,
+}: Props) {
   const { majorChanges, minorChanges } = changes;
 
   if (!Object.keys(majorChanges).length && !Object.keys(minorChanges).length) {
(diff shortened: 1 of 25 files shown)


src/explore-education-statistics-frontend/src/modules/data-catalogue/components/DataSetFileApiChangelog.tsx
```diff
@@ -6,10 +6,15 @@ import React from 'react';
...
 interface Props {
   changes: ApiDataSetVersionChanges;
+  publicGuidanceNotes?: string;
   version: string;
 }
 
-export default function DataSetFileApiChangelog({ changes, version }: Props) {
+export default function DataSetFileApiChangelog({
+  changes,
+  publicGuidanceNotes,
```
ntsim, line 15: Really minor, but would personally simplify this to just `guidanceNotes` as the 'public' part is fairly redundant. Can leave it though as well.

src/explore-education-statistics-frontend/src/services/apiDataSetService.ts
```diff
@@ -33,7 +33,7 @@ export interface ApiDataSetVersion {
   status: ApiDataSetStatus;
   published: string;
   withdrawn?: Date;
-  notes: string;
+  notes?: string;
```
ntsim, line 36: This isn't optional according to the backend view models? If there aren't any notes, this will just be an empty string.

Think this should be reverted?

tests/robot-tests/args_and_variables.py
```diff
@@ -82,12 +82,7 @@ def create_argument_parser() -> argparse.ArgumentParser:
         action="store_true",
         help="rerun failed test suites and merge results into original run results",
     )
-    parser.add_argument(
-        "--rerun-attempts",
-        dest="rerun_attempts",
-        type=int,
-        default=0,
-        help="Number of rerun attempts")
+    parser.add_argument("--rerun-attempts", dest="rerun_attempts", type=int, default=0, help="Number of rerun attempts")
```
ntsim, line 85: Seems like we've ended up making a tonne of unrelated changes (formatting and logic) across the RF / Python code in this PR. There's _a lot_ of unrelated stuff on top of what is just a simple change to add some guidance notes and is making it a lot longer / harder to read. 

I'm particularly unsure about the Python stuff, particularly as it's also interlaced with formatting changes so I'm just going to have to trust it's correct...

Can we avoid doing this in future PRs? We've previously discussed in retros and dev team meetings to avoid doing stuff like this to keep the scope as narrow as possible.

ntsim (changes requested):

PR #5320, opened 2024-10-10
EES-5559 Fix filters and indicators not re-using public IDs for new versions

ntsim:
This PR fixes filters and indicators not re-using the public IDs of existing filters / indicators  when preparing new API data set versions. This was due to the `ImportMetadataFunction` code only really handling public ID mappings for filter options and location options.

As the public IDs were not being re-used, this was leading to the changelog reporting that _all_ filters and indicators were being deleted and re-added. By re-using the public IDs correctly, we now no longer see this.

We've importantly added a lot more tests for `ImportaMetadataFunction`, however, we still need to cover the other meta types more comprehensively. In the interest of time, this will be handled in a subsequent ticket - EES-5560.

## Relevant change

- Added `CloneExtensions` for shallow / deep clone extension methods. This uses [ObjectCloner](https://github.com/marcelltoth/ObjectCloner) internally.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/CloneExtensions.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/CloneExtensions.cs
@@ -1,0 +1,15 @@
+#nullable enable
+namespace GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+
+public static class CloneExtensions
+{
+    public static T ShallowClone<T>(this T obj)
+    {
+        return ObjectCloner.ObjectCloner.ShallowClone(obj);
+    }
+
+    public static T DeepClone<T>(this T obj)
+    {
+        return ObjectCloner.ObjectCloner.DeepClone(obj);
+    }
+}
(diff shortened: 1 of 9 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/FilterMetaRepository.cs
```diff
@@ -53,17 +54,18 @@ public async Task CreateFilterMetas(
         IReadOnlySet<string> allowedColumns,
         CancellationToken cancellationToken = default)
     {
+        var publicIdMappings = await CreatePublicIdMappings(dataSetVersion, cancellationToken);
```
duncan-at-hiveit, line 57: Aha cool.  I see the original problem now - shame we didn't include mapping filters and indicators in MVP cos we would have spotted this by now otherwise!

duncan-at-hiveit (approved):

src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/CloneExtensions.cs
```diff
@@ -0,0 +1,16 @@
+
```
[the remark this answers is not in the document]
ntsim: As discussed in Slack - gonna leave this in for now in case we need it elsewhere

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Repository/FilterMetaRepository.cs
ntsim: Yep indeed! A bit of a blind spot unfortunately. Will hopefully be sorting more of the tests out in a follow up ticket as well!

PR #5327, opened 2024-10-10
EES-XXXX - amending run_tests.py to detect failed suites from output.xml

duncan-at-hiveit:
image

During some battles with text selector changes, I broke the Robot code a few times but noticed that misconfigured keywords and errors within our own keyword code didn't cause our tests to fail fast (nothing new, as the `run_on_failure` that we use to trigger failures only kicks in within SeleniumLibrary's own keywords as has always been the case). This would mean that some suites ultimately failed but didn't get recorded by our fail-fast mechanism.

They were however captured in output.xml and so I changed the code up to use this as the ultimate source of truth instead, should we accidentally introduce keyword changes that break them but don't get caught by our fail-fast mechanism.

Ultimately to fix the fail-fast issue we could do with setting `run_on_failure=None` to prevent Selenium-only keywords failing suites fast, and instead in each test suite specify `Test Teardown    Run Keyword On Test Failure    Record failing suite` - the only issue here being that it doesn't catch failures in `Suite Setup`, so we'd need another mechanism for that, potentially with listeners.

## Additional console logging details

There is now a direct link to the last run attempt's report `log.html` as well as the merged report, meaning you can jump to the final run's failures without all the noise of the full test suite results.

```
2024-10-10 14:20:40,753: *************************************
2024-10-10 14:20:40,753: FINAL REPORT
2024-10-10 14:20:40,753: Log available at: file:///home/dwatson/development/projects/dfe/explore-education-statistics/tests/robot-tests/test-results/log.html
2024-10-10 14:20:40,753: Report available at: file:///home/dwatson/development/projects/dfe/explore-education-statistics/tests/robot-tests/test-results/report.html
2024-10-10 14:20:40,753: *************************************

2024-10-10 14:20:40,753: *************************************
2024-10-10 14:20:40,753: LAST RUN ATTEMPT REPORT
2024-10-10 14:20:40,753: Log available at: file:///home/dwatson/development/projects/dfe/explore-education-statistics/tests/robot-tests/test-results/run-1/log.html
2024-10-10 14:20:40,753: Report available at: file:///home/dwatson/development/projects/dfe/explore-education-statistics/tests/robot-tests/test-results/run-1/report.html
2024-10-10 14:20:40,753: *************************************

2024-10-10 14:20:40,753: Number of test run attempts: 1
2024-10-10 14:20:40,753: 
All tests passed!
```

### DIFF
--- a/tests/robot-tests/reports.py
+++ b/tests/robot-tests/reports.py
@@ -1,7 +1,7 @@
-import os
 import glob
+import os
 import shutil
-import time
+
 from bs4 import BeautifulSoup
 from robot import rebot_cli as robot_rebot_cli
 from tests.libs.logger import get_logger
@@ -13,21 +13,19 @@ logger = get_logger(__name__)
 
 # Merge multiple Robot test reports and assets together into the main test results folder.
 def merge_robot_reports(first_run_attempt_number: int, number_of_test_runs: int):
-
-    first_run_folder=f"{main_results_folder}{os.sep}run-{first_run_attempt_number}"
+    first_run_folder = f"{main_results_folder}{os.sep}run-{first_run_attempt_number}"
 
     logger.info(f"Merging test run {first_run_attempt_number} results into full results")
-        
+
     for file in os.listdir(first_run_folder):
         _copy_to_destination_folder(first_run_folder, file, main_results_folder)
 
     for test_run in range(first_run_attempt_number + 1, number_of_test_runs + 1):
-        
         logger.info(f"Merging test run {test_run} results into full results")
-        
+
         test_run_foldername = f"{main_results_folder}{os.sep}run-{test_run}"
         _merge_test_reports(test_run_foldername)
-        
+
         for file in glob.glob(rf"{test_run_foldername}{os.sep}*screenshot*"):
             _copy_to_destination_folder(test_run_foldername, file.split(os.sep)[-1], main_results_folder)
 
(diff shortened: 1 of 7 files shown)


tests/robot-tests/reports.py
```diff
@@ -79,36 +105,48 @@ def _merge_test_reports(test_results_folder):
...
+    suite_results = report.find("statistics").find("suite").find_all("stat")
+    failing_suite_results = [suite_result for suite_result in suite_results if int(suite_result.get("fail")) > 0]
+    suite_ids = [failing_suite.get("id") for failing_suite in failing_suite_results]
+    leaf_suite_ids = []
+    for suite_id in suite_ids:
+        other_suite_ids_containing_suite_id = [
+            other_suite_id for other_suite_id in suite_ids if other_suite_id != suite_id and suite_id in other_suite_id
+        ]
+        if len(other_suite_ids_containing_suite_id) == 0:
+            leaf_suite_ids += [suite_id]
```
mmyoungman, line 143: Is this to remove duplicates from `suite_ids`? I think you could do `suite_ids_no_dups = list(set(suite_ids))`

tests/robot-tests/reports.py
```diff
@@ -79,36 +105,48 @@ def _merge_test_reports(test_results_folder):
...
-    passing_suite_results = [suite_result for suite_result in suite_results if int(suite_result.get('fail')) == 0]
-    return [passing_suite.get('id') for passing_suite in passing_suite_results]
+
+def _get_passing_suite_ids_from_report(report: BeautifulSoup) -> []:
+    suite_results = report.find("statistics").find("suite").find_all("stat")
+    passing_suite_results = [suite_result for suite_result in suite_results if int(suite_result.get("fail")) == 0]
+    return [passing_suite.get("id") for passing_suite in passing_suite_results]
+
+
+def _get_failing_leaf_suite_ids_from_report(report: BeautifulSoup) -> []:
```
mmyoungman, line 133: I'm not sure what you mean by `leaf` here

mmyoungman (approved):

tests/robot-tests/tests/libs/fail_fast.py
```diff
@@ -93,8 +94,8 @@ def _prompt_to_continue():
     logger.warn("Continue? (Y/n)")
     choice = input()
     if choice.lower().startswith("n"):
-        raise_assertion_error("Tests stopped!")
-        
+        _raise_assertion_error("Tests stopped!")
+
 
 def _raise_assertion_error(err_msg):
```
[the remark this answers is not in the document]
duncan-at-hiveit: Oh no, it still is:

```
def _raise_assertion_error(err_msg):
    sl().failure_occurred()
    raise AssertionError(err_msg)
```

This is different code than the skipping of tests :)

tests/robot-tests/reports.py
duncan-at-hiveit: This is to remove non-leaf suites.  The error report xml includes stats reporting at multiple levels, so if you have:

Admin:Admin And Public:Some Leaf Suite Here

duncan-at-hiveit: Given it wasn't very obvious, I've added a comment to this effect in the code.

tests/robot-tests/reports.py
duncan-at-hiveit: See above :)

tests/robot-tests/tests/libs/fail_fast.py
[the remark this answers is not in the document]
mmyoungman: My bad - I didn't see the 3rd line...

PR #5330, opened 2024-10-11
EES-5565 Fix UI test failure due to existence of other prerelease users

mmyoungman:
This PR fixes a UI test failure we were seeing due to an additional prerelease user being created on an environment alongside `ees-prerelease1` and `ees-prerelease2`.

Previously the tests passed as this would happen:

- `user_resets_user_roles_via_api_if_required` is called. This removes the global prerelease role for both ees-prerelease1 and ees-prerelease2. This means both users no longer appear in the list of prerelease users, but in the list of all other users.

- `_get_prerelease_user_details_via_api` is called. This results in `admin_client.get("/api/user-management/pre-release")` returning a 404 (as they are no other prerelease users) and so it throws an exception when `.json()` is called on the response.

- This 404 is caught and in the `except` block we go on to call `_get_user_details_via_api`, the prerelease users are found and the UI test continues as expected

However, if `user-management/pre-release` returned a 200 - because another prerelease user existed -  the response would be successfully parsed into json, resulting in us triggering `assert matching_users, f"Could not find user with email {user_email}"`.

I've removed the checks for exceptions and instead rely on checking the response status code directly to see if the user has been found amongst prerelease users or general users.

I also changed the backend so the user-management/pre-release endpoint no longer returns a 404 if there are no pre-release users. It now returns an JSON empty list.

### DIFF
--- a/tests/robot-tests/tests/libs/admin_api.py
+++ b/tests/robot-tests/tests/libs/admin_api.py
@@ -2,7 +2,6 @@ import os
 from datetime import datetime
 
 import requests
-from robot.libraries.BuiltIn import BuiltIn
 from tests.libs import local_storage_helper
 
 # To prevent InsecureRequestWarning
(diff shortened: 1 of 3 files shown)


tests/robot-tests/tests/libs/admin_api.py
```diff
@@ -219,21 +215,41 @@ def delete_test_user(email: str):
 
 
 def _get_user_details_via_api(user_email: str):
-    users = admin_client.get("/api/user-management/users").json()
+    response = admin_client.get("/api/user-management/users")
+
+    if response.status_code == 404:
```
duncan-at-hiveit, line 220: I don't think we should be expecting a 404 here - just an empty list.

tests/robot-tests/tests/libs/admin_api.py
```diff
@@ -219,21 +215,41 @@ def delete_test_user(email: str):
...
+    assert len(matching_users) == 1, f"Should only have found one user with email {user_email}"
 
     return matching_users[0]
 
 
 def _get_prerelease_user_details_via_api(user_email: str):
-    users = admin_client.get("/api/user-management/pre-release").json()
+    response = admin_client.get("/api/user-management/pre-release")
+
+    if response.status_code == 404:
```
duncan-at-hiveit, line 240: Again with the 404 vs empty list

duncan-at-hiveit (approved):

tests/robot-tests/tests/libs/admin_api.py
mmyoungman: See other comment

tests/robot-tests/tests/libs/admin_api.py
mmyoungman: I changed my mind and changed the pre-release users endpoint to return an empty list instead of a 404. So I've also removed the 404 check from the python code.

PR #5331, opened 2024-10-13
EES-5570 Cleanup various tests related to `DataSetVersionMappings`

ntsim:
This PR performs various cleanup across two test suites relating to `DataSetVersionMappings`.

This primarily does things like:

- Refactoring common test methods for assertion and getting things from the DB
- Renaming plural `mappings` variable to `mapping` singular


duncan-at-hiveit (approved):
Very nice very nice, and thanks for sorting out the pluralisation!  Was on my list of TODOs for some time in the future that would never come.

PR #5332, opened 2024-10-13
EES-5563 Fix not all meta deletions being considered as major version updates

ntsim:
:warning: Depends on #5331

---

This PR primarily fixes deletions of indicator, geographic level and time periods previously not being considered by the data set's version number as a major update.

We have implemented this as new flag columns on `DataSetVersionMapping` in the form of:

- `HasDeletedIndicators`
- `HasDeletedGeographicLevels`
- `HasDeletedTimePeriods`

This is not necessarily an optimum way of implementing this as a better solution may be to structure this information in a similar to the mappings for filters and locations. Unfortunately, we're quite time constrained to get this working for private testing and have gone with this as a simpler solution for now.

In the future, we'll probably need to replace these columns if we want to implement things like indicator mappings.

## Related changes

- Fixed the read-only `FilterMetaRepository.ReadFilterMetas` method incorrectly incrementing the meta ID sequence in the database.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionMappingControllerTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionMappingControllerTests.cs
@@ -416,7 +416,7 @@ public abstract class DataSetVersionMappingControllerTests(
         }
 
         [Fact]
-        public async Task Success_DeletedLevel_MajorUpdate()
+        public async Task Success_MappedLocation_HasDeletedLocationLevels_MajorUpdate()
         {
             var (initialDataSetVersion, nextDataSetVersion) = await CreateInitialAndNextDataSetVersion();
 
@@ -482,7 +482,7 @@ public abstract class DataSetVersionMappingControllerTests(
         }
 
         [Fact]
-        public async Task Success_AddedLevel_MinorUpdate()
+        public async Task Success_MappedLocation_HasDeletedIndicators_MajorUpdate()
         {
             var (initialDataSetVersion, nextDataSetVersion) = await CreateInitialAndNextDataSetVersion();
 
@@ -504,16 +504,9 @@ public abstract class DataSetVersionMappingControllerTests(
                                     .WithNoMapping())
                             .AddCandidate(
                                 targetKey: "target-location-1-key",
-                                candidate: DataFixture.DefaultMappableLocationOption()))
-                    // Country level has been added and only has candidates.
-                    .AddLevel(
-                        level: GeographicLevel.Country,
-                        mappings: DataFixture
-                            .DefaultLocationLevelMappings()
-                            .AddCandidate(
-                                targetKey: "source-location-1-key",
-                                candidate: DataFixture
-                                    .DefaultMappableLocationOption())));
+                                candidate: DataFixture.DefaultMappableLocationOption())))
+                // There are deleted indicators that cannot be mapped.
+                .WithHasDeletedIndicators(true);
 
             await TestApp.AddTestData<PublicDataDbContext>(context => context.DataSetVersionMappings.Add(mapping));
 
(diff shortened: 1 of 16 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Models/DataSetVersionMappingMeta.cs
```diff
@@ -0,0 +1,18 @@
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Models;
+
+public record DataSetVersionMappingMeta
```
duncan-at-hiveit, line 5: Nice spot on the rename!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionMappingService.cs
```diff
@@ -63,6 +63,21 @@ public async Task<Either<ActionResult, Unit>> CreateMappings(
             sourceFilterMeta,
             nextVersionMeta.Filters);
 
+        var hasDeletedIndicators = await HasDeletedIndicators(
+            liveVersion.Id,
+            nextVersionMeta.Indicators,
+            cancellationToken);
+
+        var hasDeletedGeographicLevels = await HasDeletedGeographicLevels(
```
duncan-at-hiveit, line 71: I think we can infer this one currently from the fact that we have a LocationLevelMapping for a particular location level in the plan that has no Candidates.  I'm not against storing it explicitly for time-saving purposes, but just wondering it there was something I missed?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionMappingService.cs
```diff
@@ -63,6 +63,21 @@ public async Task<Either<ActionResult, Unit>> CreateMappings(
             sourceFilterMeta,
             nextVersionMeta.Filters);
 
+        var hasDeletedIndicators = await HasDeletedIndicators(
```
duncan-at-hiveit, line 66: I guess that post-MVP, when we have the chance to map Indicators, we can retire this column.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionMappingService.cs
duncan-at-hiveit: In fact, you have it in the code for `IsMajorVersionUpdate()`:

```
        var hasDeletedLocationLevels = mapping.LocationMappingPlan
            .Levels
            .Any(level => level.Value.Candidates.Count == 0);
```

So I was wondering what the benefit of having it in a separate column is, unless it doesn't represent the exact same thing?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionMappingControllerTests.cs
```diff
@@ -504,16 +504,125 @@ public async Task Success_AddedLevel_MinorUpdate()
...
+
+            Assert.True(updatedMapping.LocationMappingsComplete);
+
+            // This update completes the mapping and would normally be a minor version
+            // update, but the deleted indicators mean this is still a major version update.
+            await AssertCorrectDataSetVersionNumbers(updatedMapping, "2.0.0");
+        }
+
+        [Fact]
+        public async Task Success_MappedLocation_HasDeletedGeographicLevels_MajorUpdate()
```
duncan-at-hiveit, line 543: So what is the difference between `Success_MappedLocation_HasDeletedLocationLevels_MajorUpdate` and `Success_MappedLocation_HasDeletedGeographicLevels_MajorUpdate`?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionMappingService.cs
ntsim: Location levels and geographic levels aren't the same thing though:

- location levels correspond to CSV columns
- geographic levels correspond to the values specifically in the `geographic_level` column

Whilst similar looking, these aren't the same thing. For example, a data set may have location levels for national (country), regional and local authority, but may only contain data where the `geographic_level` column is `Local authority` (i.e. only local authority level data).

This means there are:

- location levels - NAT, REG, LA
- geographic levels - LA

This is a really subtle difference, but is entirely possible and needs to be encoded correctly. The location mapping plan only considers location levels (i.e. `LocationMeta` and `LocationOptionMeta`), whereas we have a separate `GeographicLevelMeta` that handles geographic levels.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionMappingControllerTests.cs
ntsim: Think this relates to my earlier reply: https://github.com/dfe-analytical-services/explore-education-statistics/pull/5332#discussion_r1799207660

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionMappingService.cs
ntsim: Yep I would think so :smile:

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Services/DataSetVersionMappingService.cs
duncan-at-hiveit: Gotcha, makes sense now, and at some point I'm going to get this concept...

duncan-at-hiveit (approved):

PR #5342, opened 2024-10-16
EES-5568 Adding modal for admins deleting users

jack-hive:
When BAU users try to delete users from Admin, we need to warn users to confirm this as it's a fairly big action.
This PR adds a confirmation modal to confirm the deletion before they proceed.

### Notes for Reviewer
- I'm not sure if I've put the modal component in the right place. Nor am I sure if there was another component I should have re-used instead. Please let me know if there was something different I should have done here.
- The code is commented out as we still want to hide this Delete button for now until the issue in [EES-5573](https://dfedigital.atlassian.net/browse/EES-5573) is resolved.
- Please uncomment my code, add the necessary imports, and re-test it for me. I'm new to React so I may have missed something.
- Thank you


src/explore-education-statistics-admin/src/pages/bau/components/DeleteUserModal.tsx
```diff
@@ -0,0 +1,26 @@
...
+import React, { ReactNode } from 'react';
+
+interface Props {
+  triggerButton: ReactNode;
+  onConfirm: () => void;
+}
+
+const DeleteUserModal = ({ triggerButton, onConfirm }: Props) => {
+  return (
+    <ModalConfirm
```
bennettstuart, line 11: Hey! This is absolutely fine to do, and works perfectly well. I just wanted to add my personal opinion to give some food for thought on your frontend journey!

We didn't have a **DeleteUserButton** previously, I think for the same reasons I would not create a **DeleteUserModal**. As the **DeleteUserModal** doesn't do much other than simply return another component, I think I would avoid creating a new component and use the **ModalConfirm** in **BauUsersPage** directly. Doing so will have less lines added and less files added. However, if this modal is to appear across the service, then I'd opt for the component separation for easier tweaking later on. Hope this helps!

bennettstuart (approved):
Looks good, left an opinion comment to think about but implementation is all gooood!

src/explore-education-statistics-admin/src/pages/bau/components/DeleteUserModal.tsx
jack-hive: Thanks Stuart!

I completely agree, that sounds like a more sensible approach to me! I think I did it like this because I saw that we split out a custom component for the same thing elsewhere, which was also used by only one other component. I was torn at the time, but your comment has now confirmed in my mind that it's better to **not** split out a custom component for this, as it is only used in one place :)

PR #5343, opened 2024-10-17
EES-5582 Fix snapshots script and update snapshots

mmyoungman:
This PR fixes an issue where the snapshots for the table tool was returning no publications. This was fixed by updating a CSS selector.

I also fixed an issue where if a snapshot file didn't exist, the script would fail. I've changed it so now the script will generate a snapshot file if none exists.

I also updated the snapshots.


cjrace (approved):
Happy with the snapshot changes. Not that familiar with Python but those changes also look fine to me

PR #5344, opened 2024-10-17
EES-4863 accessible reorderable list component

amyb-hiveit:
Adds a new reusable `ReorderableList` component for accessible reordering that renders a simplified ordered list. 
- includes move up and down buttons to satisfy the new WCAG requirement (note that these can't be nested in the draggable because it has the button role).
- moves focus to the first item when start reordering and keeps focus on the item you've moved.
- moves the confirm button to be consistently under the reorderable list.
- instead of updating the existing table / list / whatever to make it reorderable I'm replacing it with the new ReorderableList component when reordering is enabled. This is to make it simpler to use, plus it should help keep the reordering UIs more consistent.

I've implemented this new component for reordering:
- footnotes 
- releases (in the legacy releases tab)
- featured tables

### DIFF
--- a/src/explore-education-statistics-common/src/components/ReorderableItem.tsx
+++ b/src/explore-education-statistics-common/src/components/ReorderableItem.tsx
@@ -1,0 +1,103 @@
+import DragIcon from '@common/components/DragIcon';
+import styles from '@common/components/ReorderableItem.module.scss';
+import VisuallyHidden from '@common/components/VisuallyHidden';
+import { ArrowLeft, ArrowRight } from '@common/components/ArrowIcons';
+import Button from '@common/components/Button';
+import ButtonGroup from '@common/components/ButtonGroup';
+import { mergeRefs } from '@common/utils/mergeRefs';
+import classNames from 'classnames';
+import React, { ReactNode, useEffect, useRef } from 'react';
+import { DraggableProvided, DraggableStateSnapshot } from '@hello-pangea/dnd';
+
+export interface ReorderableListItem {
+  id: string;
+  label: ReactNode | string;
+}
+
+interface Props {
+  draggableProvided: DraggableProvided;
+  draggableSnapshot: DraggableStateSnapshot;
+  dropAreaActive: boolean;
+  focusItem: boolean;
+  index: number;
+  isLastItem: boolean;
+  item: ReorderableListItem;
+  onMoveItem: ({
+    prevIndex,
+    nextIndex,
+  }: {
+    prevIndex: number;
+    nextIndex: number;
+  }) => void;
+}
+
+export default function ReorderableItem({
+  draggableProvided,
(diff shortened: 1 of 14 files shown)


src/explore-education-statistics-common/src/components/ReorderableItem.tsx
```diff
@@ -0,0 +1,100 @@
...
+        ref={mergeRefs(draggableProvided.innerRef, itemRef)}
+        style={draggableProvided.draggableProps.style}
+        data-testid="reorderable-item"
+      >
+        <div className={styles.itemLabel}>
+          <DragIcon className={styles.dragIcon} />
+          <span>{item.label}</span>
+        </div>
+      </div>
+      {!draggableSnapshot.isDragging && (
```
bennettstuart, line 70: image

In the image above, you'll notice the UP button from footnote 3 has moved up into the empty drop area - I wonder if we might hide move buttons for all items during drag action

src/explore-education-statistics-common/src/components/ReorderableItem.tsx
```diff
@@ -0,0 +1,100 @@
...
+        {...draggableProvided.dragHandleProps}
+        className={classNames('govuk-!-margin-bottom-0', styles.draggable, {
+          [styles.isDragging]: draggableSnapshot.isDragging,
+          [styles.isDraggedOutside]:
+            draggableSnapshot.isDragging && !draggableSnapshot.draggingOver,
+        })}
+        ref={mergeRefs(draggableProvided.innerRef, itemRef)}
+        style={draggableProvided.draggableProps.style}
+        data-testid="reorderable-item"
+      >
```
bennettstuart, line 64: image

An interesting state where one item is focussed and another is then dragged, resulting in a double highlight

bennettstuart: Also, the crossed arrows on the left a quite close to the border, some padding might make it more aesthetic

bennettstuart (changes requested):
Noted some possible points for improving the usability, the code looks fine to me.

src/explore-education-statistics-common/src/components/ReorderableItem.tsx
amyb-hiveit: Good spot, that's sorted now - the buttons are hidden when something is being dragged.

src/explore-education-statistics-common/src/components/ReorderableItem.tsx
amyb-hiveit: Sorted now - the focus style is removed if it's not the item that's being dragged and I've added a bit of padding.

amyb-hiveit:
@bennettstuart I've added the reverse order option and some tests now.

bennettstuart (approved):

PR #5345, opened 2024-10-21
EES-5546: Allow bulk imports to start replacements

tomjonesdev:
> PR raised on behalf of @mmyoungman in his absence

Currently, a bulk import will fail if one the data sets contained in the zip has the same title as an already-imported data set. These changes allow users to initiate a data replacement for those data sets that have the same title as an existing data set.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Models/ArchiveDataSetFile.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Models/ArchiveDataSetFile.cs
@@ -1,4 +1,6 @@
 #nullable enable
+using GovUk.Education.ExploreEducationStatistics.Content.Model;
+
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Models;
 
 public record ArchiveDataSetFile(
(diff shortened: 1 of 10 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Models/ArchiveDataSetFile.cs
```diff
@@ -1,9 +1,12 @@
...
+
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Models;
 
 public record ArchiveDataSetFile(
     string Title,
     string DataFilename,
     string MetaFilename,
     long DataFileSize = 1048576,
-    long MetaFileSize = 1024);
+    long MetaFileSize = 1024,
```
tomjonesdev, line 11: Appreciate you didn't add these, but what's the purpose of the default values here?

tomjonesdev: Ben advised that this may be to avoid repeating values in unit tests, and recommended creating a test data generator instead - made a note to create one and remove these defaults in my current branch.

N-moh (approved):

PR #5348, opened 2024-10-21
EES-5491 delete public api data with UI test data teardown

duncan-at-hiveit:
# UI test results

## Local run

image

## Dev run

image

# Overview

This PR:
- combines the 2 existing "Delete ReleaseVersion" code paths into 1.
- hooks the existing "Bulk Delete Data Set Versions" Processor endpoint into the UI test teardown to add removal of Public API data into teardown process.
- amends the deletion order of ReleaseVersions to ensure we're deleting ReleaseVersions that contain later versions of individual Data Sets before others that contain previous versions.
- plugs up some additional leaks in our teardown code.
- adds some tidy-up to our UI tests so they clear up some data that is not covered by teardown of specific Themes / Publications / Releases etc.

# Changes

## Combining Delete ReleaseVersion code paths

We already had a mature code path for hard-deleting ReleaseVersions for a long time, specifically to support test data teardown.

Recently, we added the ability for a BAU user to hard-delete the first versions of ReleaseVersions (i.e. not Amendments), and tagged it onto the existing code for soft-deleting Amendments.

Part of the problem above whereby Public API test data was being deleted for real ReleaseVersions but not test ReleaseVersions was because it was added to one (production) code path but not the other (test) path.  This PR removes the need to maintain an additional code path by removing the test one and delegating to the production one during UI test teardown.

In order to maintain compatibility with existing behaviour of the different teardowns, I wrote out the steps that each code path took and ensured that behaviour was retained after the code paths were merged.  Test data teardowns effectively worked just like the hard-delete code path in the production code, but would run against ALL ReleaseVersions, not just non-Amendments.  Therefore not much needed to change when switching teardown over to use the production hard-delete code path rather than the test one, other than just tighten up some parts of the hard-delete process whereby we were assuming the ReleaseVersion being deleted *had* to be a non-Amendment.

### Permissions

Rather than confuse the existing `DeleteSpecificReleaseVersion` authorization code with a lot of additional clauses that checked if it was test data or not, I decided to add completely separate `DeleteTestReleaseVersion` code that is only called during teardown.  This auth code ensures that the user is BAU, that the ReleaseVersion being deleted is seed data or a UI test ReleaseVersion, and that the environment supports data teardown (`EnableThemeDeletion`).

## Hooking in "Bulk Delete Public API Data Set Versions"

Because the test data teardown code was now using the production code path for hard-deleting ReleaseVersions, this meant that we were now also calling the Public API Processor endpoint for deleting Public API data sets.

I then had to enable the Public API delete code to also allow deletion of test data, as it was currently coded to not allow teardown if data was in specific states.

In order to do this, I added an optional `forceDeleteAll` flag to the Bulk Delete Function that, if set to true, would allow the endpoint to also delete protected data if it was test data and the environment allows it.  Only by supplying this flag as true, the targeted data set versions belonging to test or seed ReleaseVersions, and the environment allowing theme deletion with the `EnableThemeDeletion` flag being set to true will allow this protected data to be torn down.

## Updating delete order of ReleaseVersions to take into account Data Set versions

The delete order of ReleaseVersions has been updated to ensure that any ReleaseVersion containing a later version of a particular Data Set is torn down prior to tearing down the others.  This ensures that we don't try to remove a ReleaseVersion that is used by a Data Set Version "in the middle" of a chain of versions, which would cause referential integrity issues due to the Changeset tables referencing "Previous" and "Next" changes.

## Plugging additional gaps in our teardown process

In order to identify other areas where we were accidentally failing to delete test data, I took counts of all tables in all databases (Content, Stats and Public) before a test run, ran the UI tests fully, and then counted again afterwards.  The queries are:

```sql
SELECT t.name       AS table_name
      ,s.row_count AS row_count
FROM   sys.tables t
JOIN   sys.dm_db_partition_stats s
 ON t.OBJECT_ID = s.OBJECT_ID
AND t.type_desc = 'USER_TABLE'
AND t.name NOT LIKE '%dss%' --Exclude tables created by SQL Data Sync for Azure.
AND s.index_id IN (0, 1)
ORDER  BY table_name;
```

for Content and Stats, and:

```sql
SELECT schemaname,relname as table_name
      , n_live_tup as row_count
FROM pg_stat_user_tables
ORDER BY schemaname,relname;
```

for the Public API DB.

Looking at these before and afters showed that we were leaving orphaned data in the following tables:

*content.Contacts*

When Publications were being deleted, they were leaving their Contacts behind - legacy domain model!  To fix this, I've added an explicit deletion of their Contact records when a test Publication is deleted.

*statistics.Subject*

We had some code for hard-deleting Subjects belonging to test ReleaseVersions:

```
context.Subject.FromSqlInterpolated($"DELETE Subject WHERE Id = {subjectId}");
``` 

This however doesn't actually execute any SQL (and the SQL is wrong too!).  So I have changed it to:

```
context.Database.ExecuteSqlInterpolated($"DELETE FROM Subject WHERE Id = {subjectId}");
```

This now cascade-deletes the target Subjects successfully.

#### Soft-deleted statistics.Subjects

We also have leftover soft-deleted Subjects from cancelled Amendments, but they are cleaned up by the Data Factory cleanup job.

#### statistics.Locations

We reuse Locations between Subjects and we don't clear out orphans.  Because our test data files don't introduce new Locations, they simply reuse any orphans left in the Locations table and don't create new ones on subsequent runs.

### Public Data LocationOptionMeta, FilterOptionMeta, XOptionMeta tables

Similar to statistics.Locations, orphans aren't removed but simply reused on subsequent UI test runs, without creating new ones.

## Adding some tidy-up to our UI test suite teardowns

Every test run, we are adding more user invites to the target environment without then cleaning up after ourselves.  This PR adds a new Python keyword for deleting user invites via the API, and we invoke this in test suites that set up new user invites.

# Manual clear-up needed on Dev following this PR

* Clear out an orphaned Public API data
* Remove orphaned content.Contacts
* Remove orphaned, non-Soft Deleted Subjects
* Remove user invites left over from UI tests


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ThemeService.cs
```diff
@@ -32,46 +30,35 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Services
     public class ThemeService : IThemeService
     {
         private readonly ContentDbContext _contentDbContext;
-        private readonly StatisticsDbContext _statisticsDbContext;
+        private readonly PublicDataDbContext _publicDataDbContext;
```
mmyoungman, line 33: I *think* that adding this means that creating a ContentDb migration now fails - although it's possible I've got this wrong:

```
An error occurred while accessing the Microsoft.Extensions.Hosting services. Continuing without the application service provider. Error: Some services are not able to be constructed (Error while validating the service descriptor 'ServiceType: GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api.ThemeController Lifetime: Transient ImplementationType: GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api.ThemeController': Unable to resolve service for type 'GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Database.PublicDataDbContext' while attempting to activate 'GovUk.Education.ExploreEducationStatistics.Admin.Services.ThemeService'.) (Error while validating the service descriptor 'ServiceType: GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.IThemeService Lifetime: Transient ImplementationType: GovUk.Education.ExploreEducationStatistics.Admin.Services.ThemeService': Unable to resolve service for type 'GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Database.PublicDataDbContext' while attempting to activate 'GovUk.Education.ExploreEducationStatistics.Admin.Services.ThemeService'.)
```

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/Cache/MethodologyCacheServiceTests.cs
```diff
@@ -192,7 +192,8 @@ public async Task GetSummariesByPublication_PublicationNotFound()
 
         VerifyAllMocks(PublicBlobCacheService);
 
-        result.AssertRight([]);
+        var viewModels = result.AssertRight();
+        Assert.Empty(viewModels);
```
mmyoungman, line 196: I had to make the same change on my branch! It did compile but Rider didn't like it for whatever reason. I'm guessing dotnet version mismatch or something

tests/robot-tests/tests/seed_data/seed_data_generation_mechanism.robot
```diff
@@ -22,13 +22,13 @@ Force Tags          Local    Dev
 
 
 *** Variables ***
-${PUBLICATION_NAME}     Seed data mechanism publication %{RUN_IDENTIFIER}
+${PUBLICATION_NAME}     UI tests - seed data mechanism publication %{RUN_IDENTIFIER}
```
mmyoungman, line 25: So I think adding "UI tests" here means that this will be deleted when we trigger the endpoint to clean up test themes?

Pretty sure this is deliberate, but just double checking.

mmyoungman (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ThemeService.cs
```diff
@@ -408,4 +334,60 @@ private async Task<List<Theme>> GetUserThemes()
...
+                {
+                    return releaseVersion1.SoftDeleted ? 1 : 0;
+                }
+
+                return -releaseVersion1.Created.CompareTo(releaseVersion2.Created);
+            }
+
+            // If one ReleaseVersion contains a later version of a Public API DataSet than the other, order it
+            // towards the top of the list so that it is deleted prior to a previous version of that DataSet
+            // being deleted.
```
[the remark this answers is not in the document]
mmyoungman: actually i don't think that would cause an issue - so ignore me :)

README.md
```diff
@@ -718,10 +718,13 @@ assert that AOP code is executed only once.
...
 
 Due to the resource requirements of the integration tests, Linux users need to ensure that the system running the tests
-is capable of doing so. The `max_user_watches` setting must be set to a high enough limit, for example by running:
+is capable of doing so. The `max_user_watches` and `max_user_instances` settings must be set to a high enough limit,
+for example by running:
 
 ```sh
 echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf
+echo fs.inotify.max_user_instances=524288 | sudo tee -a /etc/sysctl.conf
+sudo sysctl -p
```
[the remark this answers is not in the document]
duncan-at-hiveit: oh yes - for about the 3rd time!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ThemeService.cs
```diff
@@ -215,145 +214,72 @@ private async Task<Either<ActionResult, Unit>> DeleteReleasesForPublications(IEn
...
+                        ReleaseVersion: rv,
+                        DataSetVersions: dataSetVersions);
+                })
+                .ToListAsync();
 
-            // Delete release entries in the Azure Storage ReleaseStatus table - if not it will attempt to publish
-            // deleted releases that were left scheduled
-            await _releasePublishingStatusRepository.RemovePublisherReleaseStatuses(releaseVersionIdsInDeleteOrder);
+            var releaseVersionIdsInDeleteOrder = releaseVersionsAndDataSetVersions
+                .Order(new DependentReleaseVersionDeleteOrderComparator())
```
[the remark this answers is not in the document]
duncan-at-hiveit: The main reason that this needed to be reworked like this is so that we ensure we delete DataSetVersions in the correct reverse order.  We can't delete a V2 before a V3 for instance because the V3 version has a set of changelog table entries that refer to entities that the V2 DataSetVersion owns and will attempt to remove via cascade delete when the DSV itself is deleted.

The original code that deleted in a certain order I think is actually potentially not that important, but I didn't see the harm of retaining the behaviour to delete in a "realistic" order.

Also regarding your point with preferring `Version` over `PreviousVersionId` to determine ReleaseVersion order, I 100% agree, and in fact you'll see that I have switched to using Version instead in the new ordering code.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/Cache/MethodologyCacheServiceTests.cs
duncan-at-hiveit: Haha yeah!

tests/robot-tests/tests/seed_data/seed_data_generation_mechanism.robot
duncan-at-hiveit: Yes that's right!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ThemeService.cs
duncan-at-hiveit: Hmm, seems to work OK for me:

Running:

```bash
cd src/GovUk.Education.ExploreEducationStatistics.Admin
dotnet ef migrations add EES1234_MigrationNameHere --context ContentDbContext --output-dir Migrations/ContentMigrations -v
```

I get:

image

mmyoungman: Fixed

mmyoungman (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Theme.cs
```diff
@@ -8,14 +9,16 @@ public class Theme
...
+        public Guid Id { get; set; } = Guid.Empty;
 
-        public string Slug { get; set; }
+        public string Slug { get; set; } = string.Empty;
 
-        [Required] public string Title { get; set; }
+        [Required] public string Title { get; set; } = string.Empty;
 
-        public string Summary { get; set; }
+        public string Summary { get; set; } = string.Empty;
```
mmyoungman, line 18: The Slug and Summary changes have led to a migration change:
```
                     b.Property<string>("Slug")
+                        .IsRequired()
                         .HasColumnType("nvarchar(max)");
 
                     b.Property<string>("Summary")
+                        .IsRequired()
                         .HasColumnType("nvarchar(max)");
```

mmyoungman (changes requested):

mmyoungman (approved):

PR #5351, opened 2024-10-22
EES-5516 added ability to reverse table order

rianski:
Changes made:
Added button to reverse table tool header orders to descending/ascending

reversed

unreversed


src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersGroupControls.tsx
```diff
@@ -160,14 +161,23 @@ const TableHeadersGroupControls = ({
...
-                  className="govuk-!-width-one-half"
+                  className="govuk-!-width-one-third"
                   disabled={!!disableControls}
                   onClick={() => {
                     toggleMoveControlsActive(id);
                   }}
                 >
                   Move<VisuallyHidden> {legend}</VisuallyHidden>
                 </Button>
+                <Button
```
amyb-hiveit, line 172: I think it might be better to show this button when you've clicked the 'reorder' button instead of here as buttons here relate to the group rather than the items in it.

amyb-hiveit (changes requested):

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersReadOnlyList.tsx
```diff
@@ -20,14 +20,20 @@ const TableHeadersReadOnlyList = ({
...
+    useTableHeadersContext();
   const isExpanded = expandedLists.includes(id);
 
   const { getValues } = useFormContext();
   const list: Filter[] = getValues(name);
 
   const displayItems = isExpanded ? list.length : defaultNumberOfItems;
 
+  if (reverseOrderIds.includes(name) && isReverse) {
+    list.reverse();
```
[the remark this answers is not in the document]
rianski: It didn't cause a problem but you are right, creating a copy is better practise. I had to include the reversed copy within the setValue state as well as the saved values were still catching the originals instead of the reversed ones.

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersGroupControls.tsx
rianski: I had a chat with Cam about this and we have decided to keep it where it is at least for now, as the acceptance criteria wanted it where it currently is. If we decide down the line it is better elsewhere we can make the appropriate changes.

src/explore-education-statistics-common/src/modules/table-tool/contexts/TableHeadersContext.tsx
```diff
@@ -78,6 +83,14 @@ export const TableHeadersContextProvider = ({
           : [...current, groupId],
       );
 
+    const toggleReverseOrder = (groupId: string) => {
```
amyb-hiveit, line 86: I don't really get why we need to store a list of the groups that have been reversed, especially as `reverseOrder` doesn't seem to be used anywhere.

I think that instead you can make sure that the read only list component updates when the values change by changing it to use `watch` instead of `getValues` (see https://www.react-hook-form.com/api/useform/watch/)

src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersGroupControls.tsx
```diff
@@ -49,6 +52,17 @@ const TableHeadersGroupControls = ({
   const displayItems = isExpanded ? totalItems : defaultNumberOfItems;
   const hasMoreItems = isExpanded || totalItems > displayItems;
   const disableControls = activeGroup && activeGroup !== id;
+  const { getValues, setValue } = useFormContext();
+  const list: Filter[] = getValues(groupName);
+  const [isReverse, setIsReverse] = useState(false);
+
+  useEffect(() => {
```
amyb-hiveit, line 59: I don't think this needs to be in a `useEffect`, you could just trigger it on click directly, like:
```
  const handleReverse = () => {
    setValue(groupName, list.toReversed());
    toggleReverseOrder(id);
  };
  ```

**Review by amyb-hiveit — CHANGES_REQUESTED** [review]

File: src/explore-education-statistics-common/src/modules/table-tool/components/TableHeadersReadOnlyList.tsx [hunk 5]
```diff
@@ -22,10 +22,9 @@ const TableHeadersReadOnlyList = ({
 }: Props) => {
   const { expandedLists } = useTableHeadersContext();
   const isExpanded = expandedLists.includes(id);
+  const { watch } = useFormContext();
 
-  const { getValues } = useFormContext();
-  const list: Filter[] = getValues(name);
-
+  const list = watch(name) as Filter[];
```
amyb-hiveit, line 27: Does it definitely need to use `as Filter[]` instead of `const list: Filter[]`?

amyb-hiveit (approved):

PR #5361, opened 2024-10-29
EES-5573 Soft delete users

mmyoungman:
This PR allows users to be soft deleted (ish) rather than hard deleted. 

This work was mainly motivated by banging into referential integrity issues when hard deleting a user. If a user was referenced in one of the many `CreatedById`, `UpdatedById`, or `DeletedById` columns across the tables we have, the user would not be deleted.

To avoid this, we now soft delete a user's ContentDb Users table entry. All other database entries associated with the user are still hard deleted. Due to soft-deleting, we needed a little extra logic to handle the case of a soft-deleted user being readded to the service.

`HasQueryFilter` wasn't used (as we have with other entities in `ContentDbContext.cs`) as we still want to display the "deleted" user's info that is attached to various still-existing data sets they've uploaded, release status changes they've made, etc.

As part of this work, I've uncommented the `Delete` button/modal on Platform admin's User page and Pre-release user page.


src/GovUk.Education.ExploreEducationStatistics.Content.Model/User.cs
```diff
@@ -12,6 +12,8 @@ public class User
 
         public string Email { get; set; }
 
+        public DateTime? Deleted { get; set; }
```
tomjonesdev, line 15: What's the relevance of this being a `DateTime` as opposed to a boolean? It looks like the conditions used with this property only check for nulls.

mmyoungman: It's for audit purposes - since it's cheap to make it a date instead of a boolean, we get the additional info of when the deletion happened should we ever need it.

To be honest, I wouldn't have any problem with this being a boolean, but generally in this kind of situation on EES we have used nullable datetimes. I think this might actually originated from a PR comment on one of my PRs, where I had used a boolean I got the request to turn it into a datetime.

benoutram, line 15: I don't think it would have been much more effort to add in `DeletedById` here as well. Where we've got `Deleted` or `SoftDeleted` attributes I think we've also got `DeletedById` where the value comes from `userService.GetUserId()`

benoutram: Agree with Mark, we've got another example of this in `ResourceRole`. https://github.com/dfe-analytical-services/explore-education-statistics/blob/c7410430c26672e8a97fc73688ed8d631a08366a/src/GovUk.Education.ExploreEducationStatistics.Content.Model/ResourceRole.cs#L35.

I'm going to leave another comment elsewhere about this though as it looks like the deletion request can be repeated and that in that case nothing is checking the user entity is not already flagged as deleted, in which case the date would just be updated, rather than an error status being returned.

By the way, we've also got this interface which if used and the entity is marked for deletion, the value will be set rather than the entity being deleted. Not sure if this approach was originally considered or not?

https://github.com/dfe-analytical-services/explore-education-statistics/blob/8f8f4d080b97bcc8334205bb971ee69b212c36b3/src/GovUk.Education.ExploreEducationStatistics.Common/Model/Timestamps.cs#L15-L18

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
```diff
@@ -448,10 +448,13 @@ public async Task<Either<ActionResult, Unit>> DeleteUser(string email)
...
 
-                    // Delete the internal EES user, if found.
+                    // We don't remove the ContentDb's Users entry. All other stuff related to the
+                    // user is removed. This is done so if that user's Id exists in other table's
+                    // Created/Updated/DeletedById columns, that info isn't lost and we don't have to worry about
+                    // referential integrity (which we would if we removed the User table entry).
                     if (internalUser != null)
                     {
-                        _contentDbContext.Users.Remove(internalUser);
+                        internalUser.Deleted = DateTime.UtcNow;
```
benoutram, line 457: Shouldn't we also do a check above to only find an internal user that's not already flagged as deleted, and return an error if no user is found?

At present,
1. This endpoint could be run for an email that's never existed and it will return 204 as though it's been successful.
2. This endpoint could be run for an email which has already deleted and it will 'refresh' the Deleted date with the current time, overwriting the date it was actually deleted, as well as returning 204.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
```diff
@@ -448,10 +448,13 @@ public async Task<Either<ActionResult, Unit>> DeleteUser(string email)
                         .Where(invite => invite.Email.ToLower() == email.ToLower())
                         .ToListAsync();
 
-                    // Delete the internal EES user, if found.
+                    // We don't remove the ContentDb's Users entry. All other stuff related to the
+                    // user is removed. This is done so if that user's Id exists in other table's
```
benoutram, line 452: I think you have missed deleting `UserPublicationRole` and `UserReleaseRole` entries.

They extend from `ResourceRole` which has a `User` attribute.

Methods like this https://github.com/dfe-analytical-services/explore-education-statistics/blob/1b7bb64f7649b33fb895639065592a54e9af8b7e/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserReleaseRoleRepository.cs#L91 are returning data from those tables joined with the ContentDb's user.

I can see a bug where views like the Publication team access tab are listing users which have been deleted.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Model/User.cs
mmyoungman: I wasn't aware that `ISoftDeletedTimestamp` existed. I'll switch to using it - seems preferable in the case that someone tries to remove a `User` entry somewhere else.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
mmyoungman: I've updated the code so the `SoftDeleted` date cannot be refreshed.

The existing behaviour was that if someone tried to delete a user that didn't exist, it would return a 204, and I went along with that. I assume this was done as I believe we've had cases where the Identity Framework user existed and the app user didn't exist, and where the app user did exist, and the Identity Framework user didn't. In cases where the data got in a mess, we wanted delete user to clean up any instances of a user for us.

I've added a case where we return not found if no identity user *and* no app user are found.

Happy to do something else if you're were thinking of something different.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/User.cs
mmyoungman: Since I've added DeletedById, I've now backed out of this change.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/User.cs
```diff
@@ -12,6 +13,10 @@ public class User
 
         public string Email { get; set; }
 
+        public Guid DeletedById { get; set; }
```
benoutram, line 16: I think there's a few problems here. The database migration for this new `DeletedById` attribute is missing.

Shouldn't `DeletedById` be optional? Note, nullable context is also missing on this type.

For the relationship to be discoverable without needing explicit configuration, I think we need to add 

`public User? DeletedBy { get; set; } = null!;`

See https://learn.microsoft.com/en-us/ef/core/modeling/relationships/conventions

src/GovUk.Education.ExploreEducationStatistics.Content.Model/User.cs
```diff
@@ -12,6 +13,10 @@ public class User
...
+
+        public DateTime? SoftDeleted { get; set; }
```
benoutram, line 18: I think that with the recent changes renaming `Deleted` to `SoftDeleted` the database migration needs updating.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/User.cs
```diff
@@ -1,4 +1,5 @@
 using System;
+using GovUk.Education.ExploreEducationStatistics.Common.Model;
```
benoutram, line 2: I think with the changes made that this import wasn't necessary.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserManagementServiceTests.cs
```diff
@@ -1017,6 +1017,135 @@ public async Task CancelInvite_InviteNotFound()
...
+                Assert.Empty(releaseRoles);
+
+                var publicationRoles = contentDbContext.UserPublicationRoles.ToList();
+                Assert.Empty(publicationRoles);
+
+                var dbInternalUserList = contentDbContext.Users.ToList();
+                var dbInternalUser = Assert.Single(dbInternalUserList);
+                Assert.Equal(email, dbInternalUser.Email);
+                Assert.NotNull(dbInternalUser.SoftDeleted);
+                Assert.InRange(DateTime.UtcNow.Subtract(dbInternalUser.SoftDeleted.Value).Milliseconds, 0, 1500);
```
benoutram, line 1125: We've got a convenience extension `userInvite.Created.AssertUtcNow();` which can be used instead of this.

benoutram (changes requested):

benoutram (approved):

PR #5370, opened 2024-10-31
EES-5586 use reordeable list in chart data sets and key stats

amyb-hiveit:
Changes chart data sets and key stats to use the `ReorderableList` component.


bennettstuart (approved):
Nice to see it coming together!

PR #5374, opened 2024-11-05
EES-5617 Add public API docs to App Gateway via `/docs` route

ntsim:
This PR adds the public API docs app to the App Gateway via a `/docs` route.

To enable this, we've made significant modifications to the App Gateway to enable us to configure all the required parts i.e. path rules, rewrites and redirects.

We've also moved the private DNS zone (and records) out of the `appGateway` module and into the `containerApp` module. The private DNS configuration was previously necessary for Container Apps to make them accessible to the App Gateway without opening them to the public internet. However, a private DNS is not necessary (or possible) with a Static Web App, which is used for the public API docs, and did not fit the previous abstraction.

## Related changes

- Updated Swagger endpoints in public API to `/swagger` instead of `/docs` (which conflicts with the API docs app)
- Fixed `corsPolicy` type error in `containerApp` module

## Deployment

The new infrastructure changes will fail to deploy over the existing infrastructure due to existing resources that can't be automatically deleted.

To deploy the infrastructure changes, we need to manually the delete the following resources beforehand:

- App Gateway - `s101d01-ees-agw-01`
- Public IP - `s10dt01-ees-papi-ca-api-pip`

### DIFF
--- a/infrastructure/templates/public-api/components/containerApp.bicep
+++ b/infrastructure/templates/public-api/components/containerApp.bicep
@@ -137,7 +137,7 @@ resource containerApp 'Microsoft.App/containerApps@2023-11-02-preview' = {
         external: true
         targetPort: containerAppTargetPort
         allowInsecure: false
-        corsPolicy: corsPolicy
+        corsPolicy: any(corsPolicy)
         traffic: [
           {
             latestRevision: true
--- a/infrastructure/templates/public-api/components/containerApp.bicep
+++ b/infrastructure/templates/public-api/components/containerApp.bicep
@@ -75,8 +75,17 @@ param dockerPullManagedIdentityClientId string
 @secure()
 param dockerPullManagedIdentitySecretValue string
 
-@description('Id of the owning Container App Environment')
-param managedEnvironmentId string
+@description('The id of the owning Container App Environment')
+param environmentId string
+
+@description('The IP address of the Container App Environment')
+param environmentIpAddress string
+
+@description('Deploy private DNS zone and records so other resources can communicate with Container App over private network using FQDNs')
+param deployPrivateDns bool = true
+
+@description('The vnet that the Container App will be connected to')
+param vnetName string
 
 @description('Volumes to mount within Containers - used in conjunction with "volumeMounts"')
 param volumes {
@@ -114,7 +123,7 @@ resource containerApp 'Microsoft.App/containerApps@2023-11-02-preview' = {
     }
   }
   properties: {
-    managedEnvironmentId: managedEnvironmentId
+    managedEnvironmentId: environmentId
     configuration: {
       secrets: [
         {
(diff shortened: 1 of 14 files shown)


infrastructure/templates/public-api/main.bicep
```diff
@@ -280,21 +280,82 @@ module docsModule 'application/public-api/publicApiDocs.bicep' = {
...
-    publicApiContainerAppSettings: {
-      resourceName: apiAppModule.outputs.containerAppName
-      backendFqdn: apiAppModule.outputs.containerAppFqdn
-      backendDomainName: substring(apiAppModule.outputs.containerAppFqdn, indexOf(apiAppModule.outputs.containerAppFqdn, '.') + 1)
-      backendIpAddress: containerAppEnvironmentModule.outputs.containerAppEnvironmentIpAddress
-      publicFqdn: replace(publicUrls.publicApi, 'https://', '')
-      certificateName: '${apiAppModule.outputs.containerAppName}-certificate'
-      healthProbeRelativeUrl: apiAppModule.outputs.containerAppHealthProbeRelativeUrl
-    }
+    sites: [
```
duncan-at-hiveit, line 291: Wish I'd looked here right at the start of the review - instantly more understandable the changes you needed to make :)

infrastructure/templates/public-api/types.bicep
```diff
@@ -64,14 +64,136 @@ type entraIdAuthenticationType = {
...
+  @description('Path that the health probe should periodically ping e.g. /')
+  healthProbePath: string
+}
+
+@export()
+type AppGatewayRoute = {
+  name: string
+
+  @description('Name of the site')
+  siteName: string
```
duncan-at-hiveit, line 95: Is this a generic name, or is it the name of the target Azure resource itself?  Might be worth adding the distinction into the param descriptions now that variables have been renamed in some places that previously were more explicit (e.g. "resourceName" changing to "name" in some places now makes it a bit less obvious if something is just a descriptive name or if it's more defined than that).

duncan-at-hiveit (approved):

infrastructure/templates/public-api/components/containerApp.bicep
```diff
@@ -137,7 +137,7 @@ resource containerApp 'Microsoft.App/containerApps@2023-11-02-preview' = {
         external: true
         targetPort: containerAppTargetPort
         allowInsecure: false
-        corsPolicy: corsPolicy
+        corsPolicy: any(corsPolicy)
```
duncan-at-hiveit, line 140: Haha nice one - I'd previously tried to fix this and was rejected by the deploy process for my changes, so this is a good trick!

ntsim: Not ideal though! I think the type we've used is actually correct, but if you try and set empty arrays as defaults, ARM will error out saying that it's invalid! The type checker seems to be lying to us by suggesting the properties aren't optional.

duncan-at-hiveit: Yeah I remember!

infrastructure/templates/public-api/types.bicep
duncan-at-hiveit: Nice one!

PR #5377, opened 2024-11-07
EES-5547: Display info modal for bulk imports before the actual imports start

tomjonesdev:
Admin users can currently upload data sets as a bulk ZIP archive, but there's no indication of whether any of their files will replace existing data until after the files have been imported.

This PR adds a modal in between the form submission and the actual import which summarises the files to be imported, and warns of any potential file replacements. This is to provide more assurance to users that the bulk upload is doing the right thing.

image

The first part (**upload**) links in with the existing form and targets an endpoint which performs archive validation, extracts the files, and uploads them individually into a temporary blob container. The data returned from this action contains the aforementioned detail.

The second part (**import**) occurs when the user confirms the modal, which targets another endpoint to trigger the transfer of the previously uploaded files from temporary storage into the standard releases container. The data returned to the frontend is the same as before, so the file replacement process which may be triggered manually after the upload is unchanged.

### Notes

- .zip file records are no longer added to the database, just the files contained within. The archive records were initially intended as a way to trace the source of the data set, but were later deemed irrelevant
- Stream reading is currently duplicated in `ReleaseDataFileService.ValidateAndUploadBulkZip` and `DataArchiveValidationService.ValidateBulkDataArchiveFile` (which is called by the former). I tried to wrap the dependant actions in a single stream read but was met with the exception `Cannot access a disposed object`. I'm not yet clear on the impacts of having this duplication - my impression being that it will affect performance and latency when reading large files - but I have a stash of the attempt to resolve this which has so far has been unsuccessful if anyone has a potential solution or would like to have a look
- The ticket was sub-tasked into EES-5583 (backend) and EES-5584 (frontend).


bennettstuart (changes requested):

src/explore-education-statistics-admin/src/pages/release/data/components/DataFileUploadForm.tsx
```diff
@@ -294,9 +344,92 @@ export default function DataFileUploadForm({
...
+                                      >
+                                        !
+                                      </span>
+                                      <strong className="govuk-warning-text__text">
+                                        <span className="govuk-visually-hidden">
+                                          Warning
+                                        </span>
+                                        Upload will initiate a file replacement
+                                      </strong>
+                                    </div>
```
bennettstuart, line 418: We have a WarningMessage component we can use here instead

```jsx
<WarningMessage>Upload will initiate a file replacement</WarningMessage>

```

bennettstuart (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -428,89 +432,144 @@ public async Task<Either<ActionResult, DataFileInfo>> UploadAsZip(Guid releaseVe
...
+                })
+                .OnSuccess(async _ => await _dataArchiveValidationService.ValidateBulkDataArchiveFile(releaseVersionId, zipFile))
+                .OnSuccess(async dataSetFiles =>
+                {
+                    await using var stream = zipFile.OpenReadStream();
+                    var archive = new ZipArchive(stream);
 
-                    // each data/meta file pair are extracted to blob storage by _dataImportService.Import
-                    await UploadFileToStorage(bulkZipFile, bulkZipFormFile);
+                    foreach (var file in dataSetFiles)
```
mmyoungman, line 458: I was expecting us to be uploading just the zip file to temporary storage. I think that probably simplifies things a little as we can rely on `dataImportService#Import` to extract each data set from the zip.

I think the question is how much does this all-files unzipping hurt performance of the `UploadBulkZipDataSetsToTempStorage` endpoint, because I believe right now we're essentially fully decompressing the zip before a response will be returned. If the zip file ends up being in the gigbytes (which I think they will be) then that could be a big performance hit. This could also negatively effect other requests by monopolising the resources of Admin servers.

Previously, the files were being partially decompressed by Admin for validation. I believe that doesn't fully decompress the files - just a sample of each file check the csv files are valid in FileTypeService#IsValidCsvFile.

The actual full data set files decompression was performed by the Data Processor via DataImportService#Import with a `sourceZipFile` specified, which is preferable as it's done by an Azure Function.

I realise all this could represent a fair amount of refactoring, so would be happy to leave this as it is for now if performance isn't a problem - but I suspect it might be with larger zip files.

src/explore-education-statistics-admin/src/pages/release/data/components/DataFileUploadForm.tsx
```diff
@@ -114,15 +121,56 @@ function baseErrorMappings(
...
   dataFiles,
   isDataReplacement = false,
+  releaseId,
   onSubmit,
 }: Props) {
   const [selectedFileType, setSelectedFileType] = useState<FileType>('csv');
+  const [showModal, toggleShowModal] = useToggle(false);
+  const [bulkUploadPlan, setBulkUploadPlan] = useState<ArchiveDataSetFile[]>();
+
+  async function getBulkUploadPlan(zipFile: File | null | undefined) {
```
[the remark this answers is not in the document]
tomjonesdev: I think we added the union type here because `DataFileUploadFormValues.bulkZipFile` is optional, and possibly because a file input can be null/undefined.
image

src/GovUk.Education.ExploreEducationStatistics.Admin/Models/ArchiveDataSetFileImportRequest.cs
```diff
@@ -0,0 +1,14 @@
+#nullable enable
+using System;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Models;
+
+public record ArchiveDataSetFileImportRequest(
+    string Title,
+    string DataFilename,
+    string DataFilePath,
```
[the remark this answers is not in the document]
tomjonesdev: I did, but hesitated due to the path reconstruction you mention - I agree it's worth changing for the data integrity/security implication though.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
tomjonesdev: Only saving the archive was the initial plan, but another requirement/preference was to stop referencing the uploaded ZIP files, and only reference the files contained within. The logical step seemed to be to isolate archive file extraction to a single place and only handle the individual files from then onwards.

I'm not sure how to identify or measure if a file is only being partially extracted as you mention, my understanding is that you can't properly read or access a file until it has been fully extracted from an archive, though I may be wrong as it's not something I've looked into before.

The previous extraction behaviour was handled as a separate function, but this still happened as part of the same overall process, i.e. when the user uploads the data file, so they're still sitting around waiting regardless of which backend process is handling the extraction work - I appreciate that the Azure function would have a separate resource allocation which may improve performance though.

mmyoungman: > Only saving the archive was the initial plan, but another requirement/preference was to stop referencing the uploaded ZIP files, and only reference the files contained within

I think that could still be achieved - i.e. not adding the zip file to the Files table - while still push the decompressing to the Data Processor: we'd need to change `DataImportService#Import` to alternatively read a zip from temporary storage.

> my understanding is that you can't properly read or access a file until it has been fully extracted from an archive

I would hope that when we call `await stream.ReadAsync(sampleBuffer, 0, 1024);` in `IsValidCsvFile` on the archive's stream, it only decompresses those 1024 bytes, not the whole file. But I'm not certain that's correct and could be wrong.

This page seems to suggest that partially extracting a file is possible - although not sure if the ZipArchive implementation will do that: https://superuser.com/questions/1561131/is-it-possible-to-partially-extract-a-zipped-file

I''d be inclined to test your local env with a large bulk zip file - like at least 1 GB, and see how it does.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -428,89 +432,144 @@ public async Task<Either<ActionResult, DataFileInfo>> UploadAsZip(Guid releaseVe
...
+                    {
+                        file.DataFilePath = $"{FileStoragePathUtils.FilesPath(releaseVersionId, FileType.Data)}{Guid.NewGuid()}";
+                        var dataFileEntry = archive.GetEntry(file.DataFilename.ToLower());
 
-                        var results = new List<DataFileInfo>();
+                        if (dataFileEntry is null)
+                        {
+                            Common.Validators.ValidationUtils.ValidationResult(ValidationMessages.GenerateErrorFileNotFoundInZip(file.DataFilename, FileType.Data));
+                        }
```
mmyoungman, line 466: I think this check that can removed. If the file wasn't there, then it would have failed validation.

Same for the metaFileEntry check below.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -428,89 +432,144 @@ public async Task<Either<ActionResult, DataFileInfo>> UploadAsZip(Guid releaseVe
...
+                {
+                    await using var stream = zipFile.OpenReadStream();
+                    var archive = new ZipArchive(stream);
 
-                    // each data/meta file pair are extracted to blob storage by _dataImportService.Import
-                    await UploadFileToStorage(bulkZipFile, bulkZipFormFile);
+                    foreach (var file in dataSetFiles)
+                    {
+                        file.DataFilePath = $"{FileStoragePathUtils.FilesPath(releaseVersionId, FileType.Data)}{Guid.NewGuid()}";
+                        var dataFileEntry = archive.GetEntry(file.DataFilename.ToLower());
```
mmyoungman, line 461: I think I'd remove the `ToLower` here. With it, I think with it any original file containing capital letters won't be found.

Same when fetching `metaFileEntry` below.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -428,89 +432,144 @@ public async Task<Either<ActionResult, DataFileInfo>> UploadAsZip(Guid releaseVe
...
-                        ContentType = bulkZipFormFile.ContentType,
-                        Filename = bulkZipFormFile.FileName.ToLower(),
-                        Type = BulkDataZip,
-                    };
-                    _contentDbContext.Files.Add(bulkZipFile);
-                    await _contentDbContext.SaveChangesAsync();
+                        // TODO: Check this returns
+                        Common.Validators.ValidationUtils.ValidationResult(errors);
+                    }
+                })
```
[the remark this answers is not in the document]
tomjonesdev: Don't recall exactly now, but I imagine it was to help split up a large function, and allow potential future reusability of a relatively generic bit of logic.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -428,89 +432,146 @@ public async Task<Either<ActionResult, DataFileInfo>> UploadAsZip(Guid releaseVe
                 });
         }
 
-        public async Task<Either<ActionResult, List<DataFileInfo>>> UploadAsBulkZip(
+        public async Task<Either<ActionResult, List<ArchiveDataSetFile>>> ValidateAndUploadBulkZip(
```
mmyoungman, line 435: Need a test in `ReleaseDataFileServicePermissionTests` for this.

The existing permission test also needs to be renamed to match the new method name.

mmyoungman (changes requested):

src/explore-education-statistics-admin/src/pages/release/data/components/DataFileUploadForm.tsx
```diff
@@ -114,15 +121,56 @@ function baseErrorMappings(
...
+    if (releaseId && zipFile) {
```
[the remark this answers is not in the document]
tomjonesdev: I've acknowledged `releaseId`, but see https://github.com/dfe-analytical-services/explore-education-statistics/pull/5377#discussion_r1832849012 regarding `zipFile`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
tomjonesdev: After discussions, the process will remain as-is.

src/explore-education-statistics-admin/src/pages/release/data/components/DataFileUploadForm.tsx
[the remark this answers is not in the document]
bennettstuart: Happy to go with your preference - I think we're splitting hairs on where the `if` goes..
I'm suggesting something like
```ts
if (order) getMyFood(order) // getMyfood(order: Order)
```
in comparison to the implementation
```ts
getMyFood(order: Order | undefined) {
  if(!order) return // do nothing
  // ...
}
```

src/explore-education-statistics-admin/src/pages/release/data/components/DataFileUploadForm.tsx
```diff
@@ -114,15 +121,56 @@ function baseErrorMappings(
...
+          zipFile,
+        ),
+      );
+    }
+  }
+
+  async function confirmBulkUploadPlan(
+    archiveDataSetFiles: ArchiveDataSetFile[] | undefined,
+  ) {
+    if (releaseId && archiveDataSetFiles) {
```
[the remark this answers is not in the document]
bennettstuart: We still have a redundant `if` here - our types suggest we will always have an array so no need to check for an array at this point.
```tsx
async function confirmBulkUploadPlan(
    archiveDataSetFiles: ArchiveDataSetFile[],
  ) {
    if (archiveDataSetFiles) {
```

src/explore-education-statistics-admin/src/pages/release/data/components/DataFileUploadForm.tsx
```diff
@@ -114,15 +121,30 @@ function baseErrorMappings(
...
+  const [bulkUploadPlan, setBulkUploadPlan] = useState<ArchiveDataSetFile[]>();
+
+  async function confirmBulkUploadPlan(
+    archiveDataSetFiles: ArchiveDataSetFile[],
+  ) {
+    if (archiveDataSetFiles) {
+      await releaseDataFileService.importBulkZipDataFile(
+        releaseId,
+        archiveDataSetFiles,
+      );
```
bennettstuart, line 145: when I previously made a comment here, the code was transforming the archivedataSetFiles - I just want to make sure we're not missing a `replacingFileId` property as this is no longer transforming the incoming files, but the type required by `importBulkZipDataFile` remains unchanged.

tomjonesdev: After some other related changes, the two models ended up being exactly the same so I combined them into one (i.e. no transformation needed).

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseDataFileService.cs
```diff
@@ -428,89 +432,155 @@ public async Task<Either<ActionResult, DataFileInfo>> UploadAsZip(Guid releaseVe
...
+                    var viewModels = new List<ArchiveDataSetFileViewModel>();
+
+                    foreach (var file in dataSetFiles)
+                    {
+                        var dataFileEntry = archive.GetEntry(file.DataFilename);
+                        var metaFileEntry = archive.GetEntry(file.MetaFilename);
 
-                        foreach (var archiveFile in dataArchiveFiles)
+                        await UploadFileToTempStorage(file.DataFilePath, dataFileEntry!, cancellationToken);
+                        await UploadFileToTempStorage(file.MetaFilePath, metaFileEntry!, cancellationToken);
```
mmyoungman, line 467: I think we can generate Guids for the data and meta file here. The path can be generated inside `UploadFileToTempStorage` by passing it extra arguments so we can do `$"{FileStoragePathUtils.FilesPath(releaseVersionId, fileType)}{fileId}}"

Then can also remove `DataFilePath` and `MetaFilePath` from `ArchiveDataSetFile` and `GetFileIdFromPath`.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataArchiveValidationServiceTests.cs
```diff
@@ -517,7 +380,65 @@ public async Task ValidateBulkDataArchiveFile_Fail_UnusedFilesInZip()
...
+
+            IFormFile? archive = null;
+
+            // Act
+            var result = await service.IsValidZipFile(archive!);
+
+            // Assert
+            Assert.Single(result);
+            Assert.Equal("FileIsNull", result[0].Code);
+            Assert.Equal("Argument provided for parameter 'zipFile' is null.", result[0].Message);
```
mmyoungman, line 403: This needs to be updated to match the new error message

Could also utilise `AssertHasErrors` here like I suggest in another comment

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseDataFileServiceTests.cs
```diff
@@ -2943,10 +2962,179 @@ public async Task UploadAsBulkZip()
...
+
+                //privateBlobStorageService.Setup(mock =>
+                //    mock.UploadStream(
+                //        PrivateReleaseTempFiles,
+                //        It.Is<string>(path => path.Contains(FilesPath(releaseVersion.Id, FileType.Data, It.IsAny<Guid>()))),
+                //        It.IsAny<Stream>(),
+                //        ContentTypes.Csv,
+                //        null,
+                //        default
+                //    )).Returns(Task.CompletedTask);
```
mmyoungman, line 3081: Forgotten comment

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseDataFileServiceTests.cs
```diff
@@ -2943,10 +2962,179 @@ public async Task UploadAsBulkZip()
...
+
+                Assert.Equal("Two", result[1].Title);
+                Assert.Equal("two.csv", result[1].DataFilename);
+                Assert.Equal("two.meta.csv", result[1].MetaFilename);
+            }
+        }
+
+        // TODO: Refactor 2x duplicate functions in DataArchiveValidationServiceTests
+        // Possibly in MockFormTestUtils
+        private static IFormFile CreateFormFileFromResource(string fileName, string? newFileName = null)
```
mmyoungman, line 3113: Found another :)

There is another lower down too, but I won't leave a comment for any more I find

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseDataFileServiceTests.cs
tomjonesdev: Not forgotten, tests are still a WIP :P

bennettstuart (approved):

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/MockFormTestUtils.cs
```diff
@@ -34,5 +36,27 @@ public static Mock<IFormFile> CreateFormFileMock(string filename, string content
 
             return formFile;
         }
+
+        public static IFormFile CreateFormFileFromResource(string fileName, string? newFileName = null)
+        {
+            var filePath = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)!,
+                "Resources" + Path.DirectorySeparatorChar + fileName);
```
mmyoungman, line 43: Was half expecting this to break the tests, but obviously not. Seems that "GetExecutingAssembly().Location" is refers not to Common.Tests, but the project from which this method is executed.

mmyoungman (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseDataFileServiceTests.cs
```diff
@@ -1839,14 +1842,14 @@ public async Task Upload()
                 privateBlobStorageService.Setup(mock =>
                     mock.UploadFile(PrivateReleaseFiles,
                         It.Is<string>(path =>
-                            path.Contains(FilesPath(releaseVersion.Id, FileType.Data))),
+                            path.Contains(FilesPath(releaseVersion.Id, FileType.Data, null))),
```
benoutram, line 1845: Just questioning the change to `FilesPath`  in this PR which gives the folder path to a particular type of files, and is still used for this in the majority of cases but now has an additional optional param `filePath` which turns it into the path to a file when supplied.

It seems a bit strange that the method `FilesPath` (plural) is used to give the path to a file (singular).

I think rather than change this method you could have used the method which Mark already added `GovUk.Education.ExploreEducationStatistics.Content.Model.Extensions.FileExtensions.Path`

```csharp
        public static string Path(Guid rootPath, FileType type, Guid fileId)
        {
            return $"{FilesPath(rootPath, type)}{fileId}";
        }
```

and then these changes to supply the `null` param in all the existing cases where it's used for the folder path wouldn't be necessary. Appreciate that it's not an extension method even though it's in the extension namespace so that's something we could look into sorting out later on.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Services/BlobStorageServiceTests.cs
```diff
@@ -634,6 +634,62 @@ await service.DeleteBlobs(
...
+            var blobServiceClient = MockBlobServiceClient(blobContainerClient);
+
+            var service = SetupTestBlobStorageService(blobServiceClient.Object);
+
+            // Act
+            var result = await service.MoveBlob(PrivateReleaseTempFiles, sourcePath, destinationPath);
+
+            // Assert
+            Assert.False(result);
+        }
```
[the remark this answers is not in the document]
tomjonesdev: I tried, but couldn't figure out how to handle the lease behaviour so I removed the attempt - any suggestions?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/ReleasesControllerTests.cs
```diff
@@ -448,6 +448,72 @@ public async Task CreateReleaseAmendment()
...
+                Guid.NewGuid(),
+                importRequests,
+                default);
+
+            // Assert
+            VerifyAllMocks(releaseDataFileService);
+
+            result.AssertOkResult(dataFileInfo);
+        }
+
```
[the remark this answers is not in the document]
tomjonesdev: I could remove them, but I was just following the existing pattern in this class. Would you agree that they provide at least _some_ benefit, and removing them would be a detriment, albeit minor?

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/MockFormTestUtils.cs
```diff
@@ -34,5 +36,27 @@ public static Mock<IFormFile> CreateFormFileMock(string filename, string content
...
+
+            return CreateFormFileFromResourceWithPath(filePath, newFileName ?? fileName);
+        }
```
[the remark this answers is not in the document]
tomjonesdev: The existing method uses the `fileName` parameter to reference a real file in the `Resources` directory of the project. The optional parameter allowed testing an unconventionally long filename for an existing test case, without having to actually store one in the project.

NB. I didn't create this method, I just moved it out of a test class into a shared utility, and modified it it so it could be reused.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/ReleasesControllerTests.cs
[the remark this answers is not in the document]
benoutram: I agree they provide some benefit, yes, so leave them in if you want. They check the service call is actually made and that service response is returned and that the input/output is not modified.

I don't think anyone would mind though if you were to go in and delete `ReleasesControllerTests.cs` entirely!

The following failure test cases in that class test a little bit more i.e. that different failure return types from the service layer are handled as bad requests / not found, etc, but this should be generic behaviour that's tested once elsewhere.

```
DeleteDataFiles_Fail_UnableToFindMetaFileToDelete
DeleteReleaseVersion_Returns_NotFound
DeleteReleaseVersion_Returns_ValidationProblem
```

I said before in the previous comment that the handling of success/failure is already unit tested but I might be wrong about that - I was expecting to see that `EitherTaskExtensions` is covered by unit tests but it looks like it might not be.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/MockFormTestUtils.cs
```diff
@@ -34,5 +36,27 @@ public static Mock<IFormFile> CreateFormFileMock(string filename, string content
...
+                .Setup(f => f.OpenReadStream())
+                .Returns(() => System.IO.File.OpenRead(filePath));
+
+            formFile
+                .Setup(f => f.FileName)
+                .Returns(() => fileName);
+
+            return formFile.Object;
+        }
     }
```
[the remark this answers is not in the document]
tomjonesdev: I didn't write this method (or its caller), I just moved them from private functions to allow reuse ([original location](https://github.com/dfe-analytical-services/explore-education-statistics/blob/cbf630445257039de9650fbcae641de57c0fa796/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataArchiveValidationServiceTests.cs#L531))

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/MockFormTestUtils.cs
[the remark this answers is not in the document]
benoutram: Ok thanks for the explanation, I didn't look into where it was being used.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataArchiveValidationServiceTests.cs
```diff
@@ -517,29 +380,60 @@ public async Task ValidateBulkDataArchiveFile_Fail_UnusedFilesInZip()
...
-            var formFile = new Mock<IFormFile>();
-            formFile
-                .Setup(f => f.OpenReadStream())
-                .Returns(() => File.OpenRead(filePath));
+            // Arrange
+            var contentDbContextId = Guid.NewGuid().ToString();
+            await using var contentDbContext = DbUtils.InMemoryApplicationDbContext(contentDbContextId);
+
+
+            var longFilename = "loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooongfilename.csv";
```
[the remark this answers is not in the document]
tomjonesdev: Good idea, and if it were a much larger string I'd probably change it, but I feel that it makes the variable declaration more complicated than it needs to be in this situation.
```
var longFilename = $"{new string('A', count: 151)}.csv";
```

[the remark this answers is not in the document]
tomjonesdev: PS. I also didn't create this test, just moved it (validation changed around a bit).

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseDataFileServiceTests.cs
tomjonesdev: Cheers, must have missed that method, I'll update.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Services/BlobStorageServiceTests.cs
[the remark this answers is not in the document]
benoutram: Ok fair enough. If it's tricky to add because of the lease behaviour its probably not worth adding in right now.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataArchiveValidationServiceTests.cs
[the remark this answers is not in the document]
benoutram: Ok yeah I hadn't really considered that you need to retain the file extension as a suffix.

If the extension wasn't needed I would recommend this approach as its not obvious right now how many characters are in the long string without counting them and your less likely to associate that count with the value of the field limit.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/MockFormTestUtils.cs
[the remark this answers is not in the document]
benoutram: Ok cool, I hadn't spotted they were moved from that other test class.

benoutram (approved):

PR #5379, opened 2024-11-07
EES-5597 Ensuring that `mailto:` links DO NOT HAVE '(opens in new tab)' added to them

jack-hive:
We currently add '(opens in new tab) 'to any external links added in the admin system. However this is also treating `mailto:` links as external, and while that is partially true, it’s not technically true, and is not an expected behaviour for users.

This PR prevents `mailto:` links from having the '(opens in new tab)' text appended to them inside the release content.

### DIFF
--- a/src/explore-education-statistics-common/src/components/ContentHtml.tsx
+++ b/src/explore-education-statistics-common/src/components/ContentHtml.tsx
@@ -87,7 +87,8 @@ export default function ContentHtml({
         return !url?.includes('explore-education-statistics.service.gov.uk') &&
           typeof node.attribs['data-featured-table'] === 'undefined' ? (
           <a href={url} target="_blank" rel="noopener noreferrer">
-            {text} (opens in a new tab)
+            {text}
+            {url.startsWith('mailto:') ? '' : ' (opens in a new tab)'}
           </a>
         ) : (
           <a href={url}>{text}</a>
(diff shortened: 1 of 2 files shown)


src/explore-education-statistics-common/src/components/ContentHtml.tsx
```diff
@@ -87,7 +87,8 @@ export default function ContentHtml({
         return !url?.includes('explore-education-statistics.service.gov.uk') &&
           typeof node.attribs['data-featured-table'] === 'undefined' ? (
           <a href={url} target="_blank" rel="noopener noreferrer">
-            {text} (opens in a new tab)
+            {text}
+            {url.startsWith('mailto:') ? '' : ' (opens in a new tab)'}
```
tomjonesdev, line 91: I wonder if this condition may be better as part of the parent condition?

If it's an external link or NOT a `mailto` (first part of the ternary), render the external link markup which shows "(opens in a new tab)" otherwise, just show the link text.

tomjonesdev (approved):

src/explore-education-statistics-common/src/components/ContentHtml.tsx
jack-hive: Thanks. Only problem with that is I'd have to duplicate the `<a>` tag. If I've interpreted what you're saying correctly?

tomjonesdev: It's already duplicated by the ternary's negative path on line 94. This is what I had in mind:

```
!url?.includes('explore-education-statistics.service.gov.uk' &&
!url.startsWith('mailto:') &&
typeof node.attribs['data-featured-table'] === 'undefined')
    ? <a href={url} target="_blank" rel="noopener noreferrer">{text}</a>
    : <a href={url}>{text}</a>
```

PR #5380, opened 2024-11-07
EES-5588 update reorderable accordions

amyb-hiveit:
Changes the reorderable accordions in the following places to use the new `ReorderableList` component:

- reordering content sections and blocks on the release content page
- reordering content sections and blocks on the methodology content page
- reordering data files on the data uploads page


src/explore-education-statistics-admin/src/components/editable/EditableSectionBlocks.tsx
```diff
@@ -59,32 +55,33 @@ const EditableSectionBlocks = <T extends EditableBlock = EditableBlock>({
     return <InsetText>There is no content for this section.</InsetText>;
   }
 
+  if (isReordering) {
+    return (
+      <ReorderableList
+        id="reorder-sections"
+        list={blocks.map(block => ({
+          id: block.id,
+          label: <div className={styles.draggable}>{renderBlock(block)}</div>,
```
bennettstuart, line 64: Works great!  
One UX point: As the only way to add a text block in-between previous blocks is to add a new one at the end of the list and then reorder it into position, I wonder if we can improve the visual look of the draggable and empty text block, here it is quite slim there isn't much indication as to what it represents.
image
It would be great if we could alter the look of the empty "preview" text block to just make it a little more clear, like we do for the empty editable text block
image

amyb-hiveit: Good spot, I've changed it to add a min height and the 'this section is empty' text.

Screenshot 2024-11-14 153909

bennettstuart (approved):
Ace!

PR #5387, opened 2024-11-12
EES-5631 Adding `ReleaseRedirects` table to the Content DB

jack-hive:
This PR adds a `ReleaseRedirects` table to the Content DB.

With the introduction of a new `Label` concept in [EES-5626](https://dfedigital.atlassian.net/jira/software/c/projects/EES/boards/105?selectedIssue=EES-5626), and the ability to edit that `Label` in [EES-5637](https://dfedigital.atlassian.net/jira/software/c/projects/EES/boards/105?selectedIssue=EES-5637), we will now be introducing the ability to change the `Slug` for a Release, which necessitates the introduction of Release Redirects.

The `ReleaseRedirects` schema is as the following:
- `Slug` - the previous slug of the Release
- `ReleaseId` - a foreign key to the `Id` primary key of the `Releases` table. Used to redirect to the **new** `Slug`
- `Created` - timestamp of when this redirect was created

We follow the same pattern as for the existing Publication and Methodology redirects.


tomjonesdev:
I may have missed/misunderstood something from the tickets, but since a redirect is linked to a specific `Release` (rather than a `ReleaseVersion`), I gather we only retain _a single previous item_ in the `Slug` field? If so, what happens if the slug is changed more than once?

Or is the idea that a new record is created each time the slug changes, and the `ReleaseId` is used alongside the `Created` property to determine the newest? If so, was a JSON field containing an array of redirects considered?

tomjonesdev (approved):

PR #5391, opened 2024-11-14
EES 5632 - Return the Release redirects from the `api/redirects` endpoint

jack-hive:
⚠️ This PR depends on https://github.com/dfe-analytical-services/explore-education-statistics/pull/5409 ⚠️ 

In this PR we add the release redirects to the `RedirectsViewModel`, to be returned in the Content API's `GET api/redirects` endpoint.

We also add some integration tests for the `api/redirects` endpoint - testing all types of redirects and the caching. No tests previously existed.

We have **TEMPORARILY** added a `Slug` field to the `Release` model as a necessity so that we can create the release redirect view models. This **IS** where the `Slug` field belongs, but is actually going to be introduced as part of [EES-5625](https://dfedigital.atlassian.net/jira/software/c/projects/EES/boards/105?selectedIssue=EES-5625). So, this PR **IS BLOCKED** by EES-5625, and the `Slug` has only been added to the `Release` here as a temporary measure to make the code compile and the tests succeed _[the migration has been purposefully excluded]_.

Once EES-5625 is merged, then we should rebase this branch and resolve any conflicts with that `Slug` field.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/RedirectsControllerTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/RedirectsControllerTests.cs
@@ -1,0 +1,252 @@
+#nullable enable
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Threading.Tasks;
+using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces;
+using GovUk.Education.ExploreEducationStatistics.Common.Tests.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Content.Api.Tests.Fixtures;
+using GovUk.Education.ExploreEducationStatistics.Content.Model;
+using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
+using GovUk.Education.ExploreEducationStatistics.Content.Model.Tests.Fixtures;
+using GovUk.Education.ExploreEducationStatistics.Content.Services.Cache;
+using GovUk.Education.ExploreEducationStatistics.Content.ViewModels;
+using Microsoft.AspNetCore.Mvc.Testing;
+using Microsoft.Extensions.DependencyInjection;
+using MockQueryable.Moq;
+using Moq;
+using Xunit;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Api.Tests.Controllers;
+
+public abstract class RedirectsControllerTests(TestApplicationFactory testApp) : IntegrationTestFixture(testApp)
+{
+    public class ListTests(TestApplicationFactory testApp) : RedirectsControllerTests(testApp)
+    {
+        public override async Task InitializeAsync() => await TestApp.StartAzurite();
+
+        [Fact]
+        public async Task RedirectsExist_Returns200WithRedirects()
+        {
+            var publicationRedirects = DataFixture.DefaultPublicationRedirect()
+                .WithPublication(DataFixture.DefaultPublication())
+                .GenerateList(3);
+
+            var releaseRedirects = DataFixture.DefaultReleaseRedirect()
+                .WithRelease(DataFixture.DefaultRelease())
+                .GenerateList(3);
+
+            var methodologyVersions = DataFixture.DefaultMethodologyVersion()
+                .ForIndex(3, s => s.SetPublished(DateTime.UtcNow))
+                .GenerateList(4);
+
+            Methodology methodology = DataFixture.DefaultMethodology()
+                .WithMethodologyVersions(methodologyVersions)
+                .WithLatestPublishedVersion(methodologyVersions.Last());
+
+            var methodologyRedirects = DataFixture.DefaultMethodologyRedirect()
+                .ForIndex(0, s => s.SetMethodologyVersion(methodologyVersions[0]))
+                .ForIndex(1, s => s.SetMethodologyVersion(methodologyVersions[1]))
+                .ForIndex(2, s => s.SetMethodologyVersion(methodologyVersions[2]))
+                .GenerateList(3);
+
+            var contentDbContext = ContentDbContextMock(
+                publicationRedirects: publicationRedirects,
+                releaseRedirects: releaseRedirects,
+                methodologiesRedirects: methodologyRedirects);
+
+            var client = BuildApp(contentDbContext.Object).CreateClient();
+
+            var response = await ListRedirects(client);
+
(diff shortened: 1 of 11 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Services/RedirectsService.cs
```diff
@@ -64,8 +64,18 @@ public async Task<Either<ActionResult, RedirectsViewModel>> List()
             .Distinct()
             .ToList();
 
+        var releaseRedirectViewModels = await _contentDbContext.ReleaseRedirects
+            .Where(rr => rr.Slug != rr.Release.Slug) // don't use redirects to the current live slug
+            .Distinct()
+            .Select(rr => new RedirectViewModel(
+                rr.Slug,
+                rr.Release.Slug
```
benoutram, line 72: I think labelling the attributes `fromSlug:` and `toSlug:` would help readability here.

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Fixtures/ReleaseRedirectGeneratorExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+    public static InstanceSetters<ReleaseRedirect> SetRelease(
+        this InstanceSetters<ReleaseRedirect> setters,
+        Release release)
+        => setters.Set(rr => rr.Release, release)
+            .SetReleaseId(release.Id);
+
+    public static InstanceSetters<ReleaseRedirect> SetReleaseId(
+        this InstanceSetters<ReleaseRedirect> setters,
+        Guid releaseId)
+        => setters.Set(rr => rr.ReleaseId, releaseId);
```
benoutram, line 38: Could also add `WithSlug` / `SetSlug` for completeness.

Same comment also applies to `PublicationRedirectGeneratorExtensions` and `MethodologyRedirectGeneratorExtensions`.

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Fixtures/PublicationRedirectGeneratorExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+    public static InstanceSetters<PublicationRedirect> SetPublication(
+        this InstanceSetters<PublicationRedirect> setters,
+        Publication publication)
+        => setters.Set(pr => pr.Publication, publication)
+            .SetPublicationId(publication.Id);
+
+    public static InstanceSetters<PublicationRedirect> SetPublicationId(
+        this InstanceSetters<PublicationRedirect> setters,
+        Guid publicationId)
+        => setters.Set(rr => rr.PublicationId, publicationId);
```
benoutram, line 38: You could use `pr` for `PublicationRedirect` or `p` for `Property` here. I think `rr` is an abbreviation of `ReleaseRedirect` probably copied over from `RedirectGeneratorExtensions`.

src/GovUk.Education.ExploreEducationStatistics.Content.ViewModels/RedirectViewModels.cs
```diff
@@ -2,7 +2,8 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.ViewModels;
 
 public record RedirectsViewModel(
     List<RedirectViewModel> Publications,
-    List<RedirectViewModel> Methodologies);
+    List<RedirectViewModel> Methodologies,
+    List<RedirectViewModel> Releases);
```
benoutram, line 6: I think this change has caused some build failures in other tests.

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/RedirectsControllerTests.cs
```diff
@@ -0,0 +1,252 @@
...
+                .Returns((publicationRedirects ?? Array.Empty<PublicationRedirect>()).AsQueryable().BuildMockDbSet().Object);
+
+            contentDbContext.Setup(context => context.ReleaseRedirects)
+                .Returns((releaseRedirects ?? Array.Empty<ReleaseRedirect>()).AsQueryable().BuildMockDbSet().Object);
+
+            contentDbContext.Setup(context => context.MethodologyRedirects)
+                .Returns((methodologiesRedirects ?? Array.Empty<MethodologyRedirect>()).AsQueryable().BuildMockDbSet().Object);
+
+            return contentDbContext;
+        }
```
benoutram, line 229: Not sure that we really want to do this here, as most other integration tests are using the EF Core In-Memory database for `ContentDbContext`.

I think you probably want to be using `await TestApp.AddTestData<ContentDbContext>` instead.

This pattern looks like it could have been copied from `DataSetFilesControllerTests`. The reason that class uses `ContentDbContextMock` is because `ReleaseFilesFreeTextTable` used by it isn't supported by EF Core In-Memory.

benoutram (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Services/RedirectsService.cs
jack-hive: Unfortunately I get a `CS0853: An expression tree may not contain a named argument specification` error :(

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Fixtures/ReleaseRedirectGeneratorExtensions.cs
jack-hive: good catch, I'll add those in :)

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Fixtures/PublicationRedirectGeneratorExtensions.cs
jack-hive: Yep, good catch! I've changed this now :)

src/GovUk.Education.ExploreEducationStatistics.Content.ViewModels/RedirectViewModels.cs
jack-hive: Yep, you're absolutely right! I was meant to come back to this and totally forgot going into the weekend.

benoutram (approved):

PR #5415, opened 2024-11-26
EES-5691 Create new Files.DataSetFileMeta column

mmyoungman:
This PR created a new Files.DataSetFileMeta column. The old column is renamed to `DataSetFileMetaOld`

This is being done to allow the JSON in DataSetFileMeta to be queryable, such that we can filter data sets on the Data Catalogue page by geographic level (in EES-5598).

A migration endpoint has been created that will need to be manually hit on each environment after this work has been deployed: see `DataSetFileMetaMigrationController`.

The changes to `ContentDbContext` required lots of finessing to make it work. I mention this because I wouldn't be surprised if Entity Framework break backwards compatibility and this feature fails to work next time we update EF.

As the tests relied on the in-memory DB, and that didn't support EF8 JSON columns, we had a couple of failing tests. To fix them, I made the saving of changes to the database mockable.

The final step of removing `DataSetFileMetaOld` and switching to using DataSetFileMeta everywhere will be done in EES-5693

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/DataSetFileMetaMigrationController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/DataSetFileMetaMigrationController.cs
@@ -1,0 +1,94 @@
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using GovUk.Education.ExploreEducationStatistics.Admin.Models;
+using GovUk.Education.ExploreEducationStatistics.Content.Model;
+using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Controllers.Api;
+
+[Route("api")]
+[ApiController]
+[Authorize(Roles = GlobalRoles.RoleNames.BauUser)]
+public class DataSetFileMetaMigrationController(ContentDbContext contentDbContext) : ControllerBase
+{
+    public class DataSetFileMetaMigrationResult
+    {
+        public bool IsDryRun;
+        public int Processed;
+    }
+
+    [HttpPut("bau/migrate-datasetfilemeta")]
+    public async Task<DataSetFileMetaMigrationResult> MigrationDataSetFileMeta(
+        [FromQuery] bool dryRun = true,
+        [FromQuery] int? num = null,
+        CancellationToken cancellationToken = default)
+    {
+        var queryable = contentDbContext.Files
+            .Where(f => f.DataSetFileMeta == null
+                && f.DataSetFileMetaOld != null);
+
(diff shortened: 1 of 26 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/DataSetFileMetaMigrationController.cs
```diff
@@ -0,0 +1,94 @@
...
+    public class DataSetFileMetaMigrationResult
+    {
+        public bool IsDryRun;
+        public int Processed;
+    }
+
+    [HttpPut("bau/migrate-datasetfilemeta")]
+    public async Task<DataSetFileMetaMigrationResult> MigrationDataSetFileMeta(
+        [FromQuery] bool dryRun = true,
+        [FromQuery] int? num = null,
```
tomjonesdev, line 26: If the intended purpose of this parameter is to manage a large number of items as separate calls, should there not also be some ordering and an implementation of `Skip` prior to `Take`? Appreciate the `Where` clauses would exclude records which have already been populated, so just curious on the approach as it seems like quite a manual process for a post-deployment task (i.e. you'd need to hit the endpoint a variable number of times per environment).

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/DataSetFileMetaMigrationController.cs
```diff
@@ -0,0 +1,94 @@
...
+        CancellationToken cancellationToken = default)
+    {
+        var queryable = contentDbContext.Files
+            .Where(f => f.DataSetFileMeta == null
+                && f.DataSetFileMetaOld != null);
+
+        if (num != null)
+        {
+            queryable = queryable.Take(num.Value);
```
tomjonesdev, line 35: What happens if `num` exceeds the remaining number of records, does it throw or just return whatever remains?

src/GovUk.Education.ExploreEducationStatistics.Common/Services/DbContextService.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace GovUk.Education.ExploreEducationStatistics.Data.Processor.Services;
+
+public class DbContextService : IDbContextService
+{
+    public DbContextService()
+    {
+    }
+
+    // Created to be mocked by tests for cases when the in-memory db doesn't emulate the real
+    // db, e.g. with EF8 JSON columns
```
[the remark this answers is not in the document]
tomjonesdev: Sorry I thought the comment above meant we weren't using EF8 yet but it looks like we are - is using the SQLite provider a valid solution here?

See:
- https://learn.microsoft.com/en-us/ef/core/what-is-new/ef-core-8.0/whatsnew#json-columns-for-sqlite
- https://learn.microsoft.com/en-us/ef/core/providers/sqlite/?tabs=dotnet-core-cli

mmyoungman:
Going with a different approach - see jira ticket

PR #5419, opened 2024-11-27
EES 5633 - FE - Redirect the user to the latest version of a page if using an old Release slug in the URL

jack-hive:
⚠️ This PR depends on some more BE changes we need to make on top of those in https://github.com/dfe-analytical-services/explore-education-statistics/pull/5391. The new changes will be made in another PR under [EES-5632](https://dfedigital.atlassian.net/jira/software/c/projects/EES/boards/105?selectedIssue=EES-5632). Please see https://github.com/dfe-analytical-services/explore-education-statistics/pull/5446 to see those BE changes. [THIS PR IS BLOCKED BY [#5446](https://github.com/dfe-analytical-services/explore-education-statistics/pull/5446) ]⚠️ 

This PR consists of the frontend changes necessary to allow **Release** redirects to work.

Previously, we only had the ability to be able to redirect the user when we detected an old **methodology/publication** slug in the URL. But now, with the introduction of `ReleaseRedirects` in [EES-5631](https://dfedigital.atlassian.net/jira/software/c/projects/EES/boards/105?selectedIssue=EES-5631) & [EES-5632](https://dfedigital.atlassian.net/jira/software/c/projects/EES/boards/105?selectedIssue=EES-5632), we need to be able to handle redirects for **releases** too.

During this work, I had to refactor **how** the redirects worked on the frontend. Previously, we just looked for a prefix of either `/methodology` or `/find-statistics` in the URL and then checked if the next segment of the URL contained an **old** slug for either a methodology or publication, respectively, and then proceeded to redirect **ANY** child route.

However, this no longer works with the introduction of release redirects, because the current pages which contain a release slug in their URL also contain a publication slug. For this reason, if we only checked for the occurrence of a prefix in the URL before redirecting the user on any child route, then the release slug wouldn't be taken into account if we found a redirect for the publication. Equally, if we introduce any further routes in the future which contain multiple possible redirect slugs in their URL, they will be impacted too.

For this reason, I have refactored the redirect logic to only check for very **explicit** patterns in the URL, when trying to determine whether or not the URL looks like one that can be redirected. I've split out the _'URL Patterns'_ into a list, which can be iterated through and easily added to. We use the [URL Pattern Standard](https://urlpattern.spec.whatwg.org/) to implement the matching logic, and use Regex to check for the various valid alternatives for a pattern (including checking for trailing slashes, which should be allowed). I have intentionally only included **existing** routes that need to be checked for redirects, and have not included subsequent child-routes that don't exist yet. Any new routes in the future, that have the potential to be redirected, can easily be added to the list of URL Patterns we check against.

Every (**current**) possible URL Pattern has been tested via Jest unit tests.

# Other Related Changes
---

- Previously, the unit tests were importing and using the same instance of the `redirectPages` module throughout each test. However, due to the `redirectPages` module containing state in the form of the cached redirects (see the `cachedRedirects` property), this was being carried over between tests and potentially interfering. This hasn't been detected until now, because the previous tests, although mocking the response from the `redirectService.list` method, always returned the same mocked data from that method. So, the caching had no consequence to whether the test passed or failed. Now we have multiple test cases for each test, corresponding to each unique URL Pattern, it was necessary to use Jest's [`isolateModulesAsync`](https://jestjs.io/docs/29.4/jest-object#jestisolatemodulesasyncfn) to isolate specific modules for every test so that local module state doesn't conflict between tests (which was being flagged when doing the `expect(redirectService.list).toHaveBeenCalledTimes(1);` checks in the tests)

### DIFF
--- a/src/explore-education-statistics-frontend/src/middleware/pages/__tests__/redirectPages.test.ts
+++ b/src/explore-education-statistics-frontend/src/middleware/pages/__tests__/redirectPages.test.ts
@@ -299,6 +299,33 @@ describe('redirectPages', () => {
       nonRedirectedCases: findStatisticsReleasePageTestData.nonRedirectedCases,
     };
 
+  const dataTablesPublicationPageTestData: RoutePatternTestCases = {
+    routePattern: `data-tables/${publicationSlugPlaceholder}`,
+    redirectedCases: [
+      {
+        oldSlugNewSlugPairsByPlaceholder: {
+          [publicationSlugPlaceholder]: {
+            oldSlug: 'original-publication-slug-1',
+            newSlug: 'updated-publication-slug-1',
+          },
+        },
+      },
+    ],
+    nonRedirectedCases: [
+      {
+        oldSlugsByPlaceholder: {
+          [publicationSlugPlaceholder]: 'original-publication-slug-2',
+        },
+      },
+      // Matches redirect slugs from another redirect type
+      {
+        oldSlugsByPlaceholder: {
+          [publicationSlugPlaceholder]: 'original-methodology-slug-1',
+        },
+      },
+    ],
+  };
+
   const dataTablesReleasePageTestData: RoutePatternTestCases = {
     routePattern: `data-tables/${publicationSlugPlaceholder}/${releaseSlugPlaceholder}`,
     redirectedCases: mixedPublicationReleasePageRedirectedCases,
@@ -313,6 +340,7 @@ describe('redirectPages', () => {
     findStatisticsReleasePageTestData,
     findStatisticsReleaseDataGuidancePageTestData,
     findStatisticsReleasePrereleaseAccessListPageTestData,
+    dataTablesPublicationPageTestData,
     dataTablesReleasePageTestData,
   ];
 
--- a/src/explore-education-statistics-frontend/src/middleware/pages/__tests__/redirectPages.test.ts
+++ b/src/explore-education-statistics-frontend/src/middleware/pages/__tests__/redirectPages.test.ts
@@ -134,6 +134,45 @@ describe('redirectPages', () => {
       },
     ];
 
+  const testRedirectDataWithChildRoute: TestDataWithoutRedirect[] = [
+    {
+      routePattern: `methodology/${slugPlaceholder}/child-route`,
+      fromSlug: testRedirects.methodologies[0].fromSlug,
+    },
+    {
+      routePattern: `find-statistics/${slugPlaceholder}/data-guidance/child-route`,
+      fromSlug: testRedirects.publications[0].fromSlug,
+    },
+    {
+      routePattern: `find-statistics/${slugPlaceholder}/prerelease-access-list/child-route`,
+      fromSlug: testRedirects.publications[0].fromSlug,
+    },
+    {
+      routePattern: `find-statistics/publication-slug/${slugPlaceholder}/child-route`,
+      fromSlug: testRedirects.releases[0].fromSlug,
+    },
+    {
+      routePattern: `find-statistics/publication-slug/${slugPlaceholder}/data-guidance/child-route`,
+      fromSlug: testRedirects.releases[0].fromSlug,
+    },
+    {
+      routePattern: `find-statistics/publication-slug/${slugPlaceholder}/prerelease-access-list/child-route`,
+      fromSlug: testRedirects.releases[0].fromSlug,
+    },
+    {
+      routePattern: `data-tables/${slugPlaceholder}/fast-track/child-route/child-route`,
+      fromSlug: testRedirects.publications[0].fromSlug,
+    },
+    {
+      routePattern: `data-tables/${slugPlaceholder}/permalink/child-route/child-route`,
+      fromSlug: testRedirects.publications[0].fromSlug,
+    },
+    {
+      routePattern: `data-tables/publication-slug/${slugPlaceholder}/child-route`,
+      fromSlug: testRedirects.releases[0].fromSlug,
+    },
+  ];
+
   let redirectSpy: jest.SpyInstance;
   const nextSpy = jest.spyOn(NextResponse, 'next');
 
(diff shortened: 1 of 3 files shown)


src/explore-education-statistics-frontend/src/middleware/pages/__tests__/redirectPages.test.ts
```diff
@@ -11,317 +11,457 @@ const redirectService = _redirectService as jest.Mocked<
...
-      { fromSlug: 'original-slug-1', toSlug: 'updated-slug-1' },
-      { fromSlug: 'original-slug-2', toSlug: 'updated-slug-2' },
-    ],
-    publications: [
-      { fromSlug: 'original-slug-3', toSlug: 'updated-slug-3' },
-      { fromSlug: 'original-slug-4', toSlug: 'updated-slug-4' },
-    ],
+    methodologies: [{ fromSlug: 'original-slug-1', toSlug: 'updated-slug-1' }],
+    publications: [{ fromSlug: 'original-slug-2', toSlug: 'updated-slug-2' }],
+    releases: [{ fromSlug: 'original-slug-3', toSlug: 'updated-slug-3' }],
```
mmyoungman, line 28: I was expecting something different for the format of `RedirectsViewModel`. 

I was expecting each release directly to be directly tied to a specific publication. If they're not tied together, then if you have a redirect for release slug `2000`, I think that'll effect all publication's with a release `2000`? So analysts could introduce a new release slug, but then unintentionally change the release slug for a different publication?

src/explore-education-statistics-frontend/src/middleware/pages/__tests__/redirectPages.test.ts
mmyoungman, line 1: The title of this PR is 5663 when I think it should be 5633 - worth fixing so it's searchable in the future.

src/explore-education-statistics-frontend/src/middleware/pages/__tests__/redirectPages.test.ts
jack-hive: We have decided to do the BE changes as part of another PR for [EES-5632](https://dfedigital.atlassian.net/jira/software/c/projects/EES/boards/105?selectedIssue=EES-5632).

This PR will be blocked for merging until that work is done. I have updated the description of this PR to reflect this.

benoutram (commented):
I found it pretty tricky to review all of the different permutations that need to be tested but they seem to have all been covered and fall into the right categories! @jack-hive I'll just wait for you to review the additional comments I left before approving...

src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
```diff
@@ -10,14 +9,30 @@ interface CachedRedirects {
...
+  }),
+  new URLPattern({
+    pathname: `/find-statistics/:${publicationSlugKey}/:${releaseSlugKey}/(data-guidance/?|prerelease-access-list/?|.{0})?`,
+  }),
+  new URLPattern({
+    pathname: `/data-tables/:${publicationSlugKey}/(fast-track/[^/]*/?|permalink/[^/]*/?|.{0})?`,
+  }),
+  new URLPattern({
+    pathname: `/data-tables/:${publicationSlugKey}/:${releaseSlugKey}{/}?`,
+  }),
```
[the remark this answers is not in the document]
jack-hive: You are correct!

I've removed this pattern and the associated tests

src/explore-education-statistics-frontend/src/middleware/pages/__tests__/redirectPages.test.ts
```diff
@@ -11,317 +12,758 @@ const redirectService = _redirectService as jest.Mocked<
...
+        [publicationSlugPlaceholder]: 'original-publication-slug-1',
+        [releaseSlugPlaceholder]: 'original-release-slug-1',
+      },
+    },
+    {
+      routePattern: `data-tables/${publicationSlugPlaceholder}/fast-track/child-route/child-route`,
+      oldSlugsByPlaceholder: {
+        [publicationSlugPlaceholder]: 'original-publication-slug-1',
+      },
+    },
```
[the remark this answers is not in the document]
jack-hive: You are correct!

I've removed this pattern and the associated tests

src/explore-education-statistics-frontend/src/middleware/pages/__tests__/redirectPages.test.ts
```diff
@@ -11,317 +12,758 @@ const redirectService = _redirectService as jest.Mocked<
...
+    {
+      routePattern: `data-tables/${publicationSlugPlaceholder}/permalink/child-route/child-route`,
+      oldSlugsByPlaceholder: {
+        [publicationSlugPlaceholder]: 'original-publication-slug-1',
+      },
+    },
```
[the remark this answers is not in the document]
jack-hive: You are correct!

I've removed this pattern and the associated tests

src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
```diff
@@ -26,9 +26,6 @@ const redirectPatterns: URLPattern[] = [
   new URLPattern({
     pathname: `/find-statistics/:${publicationSlugKey}/:${releaseSlugKey}/(data-guidance/?|prerelease-access-list/?|.{0})?`,
   }),
-  new URLPattern({
-    pathname: `/data-tables/:${publicationSlugKey}{/}?`,
-  }),
```
benoutram: I'm wondering if removing this is correct because this looks like it's covering the route `data-tables/{publicationSlug}` AFTER you've selected a publication in step 1 of the table tool, which I guess could be shared/bookmarked.

Or is this still covered by the pattern below?

src/explore-education-statistics-frontend/src/middleware/pages/__tests__/redirectPages.test.ts
```diff
@@ -299,33 +299,6 @@ describe('redirectPages', () => {
...
-        },
-      },
-      // Matches redirect slugs from another redirect type
-      {
-        oldSlugsByPlaceholder: {
-          [publicationSlugPlaceholder]: 'original-methodology-slug-1',
-        },
-      },
-    ],
-  };
```
benoutram: I'm wondering if removing this is correct because this looks like it's covering the route data-tables/{publicationSlug} AFTER you've selected a publication in step 1 of the table tool.

src/explore-education-statistics-frontend/src/middleware/pages/redirectPages.ts
jack-hive: You're right, I've added this back in :)

src/explore-education-statistics-frontend/src/middleware/pages/__tests__/redirectPages.test.ts
jack-hive: You're right, I've added this back in :)

benoutram (approved):

benoutram (approved):

PR #5425, opened 2024-12-02
Ees 5446 prevent function app slot swap until orchestrations complete

duncan-at-hiveit:
# Overview

This PR:
- adds in an additional step into the Data Processor deploy process to delay slot swapping until all active orchestrations in the production slot are complete.

Additionally,this PR:
- adds additional `deploySharedPrivateDnsZones` parameter to the pipeline in order to allow for Bicep template deployment time savings of ~2 minutes for a rarely changing common resource.
- adds additional `deployContainerApp` and `deployDataProcessor` parameters to provide further time saving opportunities if performing focused deployment activities on specific resources (and countless hours of savings if doing multiple deploys as pat of infra work).
- adds a new `IpRange` Bicep type that is differentiated from the existing `FirewallRule` type, as many Azure resources only support very simple properties (e.g. name, rangeStart, rangeEnd) for their restriction configuration, whereas others support richer details (e.g. description, priority, action).
- Removes the staging slot's healthcheck call to eliminate noise in App Insights from legit deploy activities - see the `FileNotFoundException` bulletpoint on https://dfedigital.atlassian.net/browse/EES-5232.

## Summary of the work carried out in this PR

* Adding a new endpoint to the Data Processor to report the number of active orchestrations.
* Enabling the pipeline to have network access to the new Data Processor endpoint.
* Enabling the pipeline to have EasyAuth access to the new Data Processor endpoint.
* Adding additional steps into the pipeline to wait for services to be healthy in slots prior to commencing to next step in pipeline.
* Adding additional steps into the pipeline to wait for active orchestrations to complete prior to slot swapping.

image

# Adding in a waiting step to allow orchestrations to complete

The following diagram illustrates the step that we were missing:

image

(from https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-zero-downtime-deployment)

## Adding a "StatusCheck" Function to report active orchestrations

I added a new HTTP Trigger Function which we can call that reports the current number of active / pending orchestrations on the instance that's being called e.g. the Data Processor in the production slot.  The implementation was mostly borrowed from https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-zero-downtime-deployment.  I kept the name of the Function the same as in the docs ("StatusCheck"), as I saw other mentions of it elsewhere around the internet and thought it'd probably be worth keeping the same for future Googling purposes!

### New endpoint calls

I added 2 new task types to the pipeline.  One is a generic call to an endpoint waiting for a success response (200 to 204).  The other is a specific call to an "active orchestrations" endpoint, which waits for an active orchestration count of zero to be reached.  Both of these tasks poll their endpoints for a certain number of attempts before timing out or reaching their success state.

After we have deployed new code and app settings to the *staging slot*, we firstly wait until it is back up and healthy again, polling the `/api/HealthCheck` endpoint until it returns 200.

We then proceed to call the active orchestrations endpoint on the *production slot* until it returns zero active orchestrations.

We are then free to initiate the slot swap.

Finally for our own visibility and assurance, we poll the `/api/HealthCheck` endpoint in the *production slot* until it returns 200 following the slot swap.  This _should_ be immediate but we will see!

### Network changes to allow CI to call Data Processor endpoints

The Data Processor was previously locked down to VNet traffic only.  This meant that the DevOps pipeline runners were unable to call it due to network restrictions.

In order to allow calls through without VNet peering (which given that we're in s101 is infinitely harder!) we instead enable public network access but restrict to specific IPs.  As the work of assigning the DevOps runners a static IP address is ongoing, for the purposes of this PR we are whitelisting the "AzureCloud" service tag for now.  This in conjunction with the EasyAuth work (detailed below) gives enough security while we wait for the runners to have a whitelisted IP address that we can rely on.

The Admin App's internal subnet range is whitelisted also, and the Admin App uses the internal range because the private endpoint connecting the Data Processor to the VNet allows VNet-connected resources to communicate without the need to go out over the public internet.

We also whitelist the generic maintenance IP address ranges so that we can more easily support the Function App when in production.

Because of these changes, we no longer need the steps in the pipeline that made the staging slot temporarily publicly visible and then invisible again, giving us a bonus speed boost.

### EasyAuth changes to whitelist the DevOps SPN

In order for the pipeline to make authenticated and authorised calls to the Data Processor, we need to add the SPN's client ID (application ID) to the whitelist of applications that are allowed to make calls to the Data Processor.  The Admin App Service is already whitelisted, and we add the SPN to this same list of allowed applications.  I also realised that the Admin App Service's managed identity was a whitelisted user as well as the application being whitelisted, which wasn't required.

We get the application ID from the pipeline run by using the `addSpnToEnvironment` option which then exposes a number of environment variables for pipelines to consume. 

To make use of this new whitelisting, the pipeline needs to acquire an access token for the default scope of the Data Processor's app registration, which it does in both of the new endpoint-polling tasks by using `az account get-access-token`.  It then uses this as a bearer token to call the protected endpoints.

### New "awaitActiveOrchestrations" pipeline parameter

Because the first time this code is deployed there will be no `/api/StatusCheck` endpoint in the production slot, we need to be able to bypass this step in the pipeline for its first successful deploy.  This will need to be done manually, with this setting set to false.  By default it will be true.

## Testing pipeline changes with long-running orchestration

In order to test the additional step in the pipeline, I added a dummy `LongRunningOrchestration` to the Data Processor.  This can be used to create - you guessed it - a long-running orchestration.  We can then see the effects of triggering one of these in the pipeline:

image

The task polls until the orchestration is complete and then the deploy continues successfully.

I've chosen to leave this orchestration in the codebase for the purposes of testing this feature, as our current real orchestrations are just too fast!

# Future work suggestions

- Whitelist the DevOps runners' static IP once it has been established by the CIP team how best to achieve this.
- Remove or restrict to the Dev environment the `LongRunningOrchestration` code.
- Rename `functionApp.bicep` to `durableFunctionApp.bicep` as there is so much in there that is specific to Durable Function Apps only.
- Break out a `deploy-durable-function-app.yml` template from `deploy-data-processor.yml` for use in other future Durable Function Apps in the future, as it is all steps that we would wish to carry out on others too.
- Enable access restrictions to SCM site - I have included the default `scmIpSecurityRestrictions` element to `functionApp.bicep` which is what currently exists on our Data Processor.  It would be good to lock this down as well to our maintenance IP ranges and to DevOps runners.


infrastructure/templates/public-api/main.bicep
```diff
@@ -360,15 +379,38 @@ module appGatewayModule 'application/shared/appGateway.bicep' = if (deployContai
   }
 }
 
-module dataProcessorModule 'application/public-api/publicApiDataProcessor.bicep' = {
+var adminSubnetFirewallRule = {
+  name: 'Admin App Service subnet range'
+  cidr: vNetModule.outputs.adminAppServiceSubnetCidr
+  tag: 'Default'
+  priority: 100
+}
```
ntsim, line 387: Does this need to be a separate variable? Would be tempted to inline it.

If it does need to be a variable, perhaps it should be grouped with `maintenanceFirewallRules` so they're next to each other.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/LongRunningTriggerFunction.cs
```diff
@@ -0,0 +1,49 @@
...
+    public async Task<IActionResult> TriggerLongRunningOrchestration(
+        [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = nameof(TriggerLongRunningOrchestration))]
+        HttpRequest httpRequest,
+        [DurableClient] DurableTaskClient client,
+        CancellationToken cancellationToken)
+    {
+        var instanceId = Guid.NewGuid();
+
+        var durationSeconds =
+            httpRequest.GetRequestParamInt(paramName: "durationSeconds", 60);
```
ntsim, line 25: Minor, but might be clearer to declare this like `GetRequestParamInt("durationInSeconds", defaultValue: 60)`. From a PR perspective, it's unclear what 60 is supposed to mean.

src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/HttpRequestExtensions.cs
```diff
@@ -114,4 +114,13 @@ public static bool GetRequestParamBool(
         var paramValue = GetRequestParam(httpRequest, paramName, defaultValue + "");
         return bool.Parse(paramValue);
     }
+
+    public static int GetRequestParamInt(
+        this HttpRequest httpRequest,
+        string paramName,
+        int defaultValue)
+    {
+        var paramValue = GetRequestParam(httpRequest, paramName, defaultValue + "");
```
ntsim, line 123: Could just do `defaultValue.ToString()`?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/LongRunningFunctions.cs
```diff
@@ -0,0 +1,25 @@
+using System.Diagnostics;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Model;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Extensions.Logging;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Functions;
+
+public class LongRunningFunctions(ILogger<LongRunningFunctions> logger)
```
ntsim, line 8: Do the orchestration and trigger functions need to be in separate classes? Given we're just using this for testing, perhaps it'd be simpler to keep everything together in the same class?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/LongRunningTriggerFunction.cs
```diff
@@ -0,0 +1,49 @@
...
+using Microsoft.DurableTask.Client;
+using Microsoft.Extensions.Logging;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Processor.Functions;
+
+public class LongRunningTriggerFunction(
+    ILogger<LongRunningTriggerFunction> logger)
+{
+    [Function(nameof(TriggerLongRunningOrchestration))]
+    public async Task<IActionResult> TriggerLongRunningOrchestration(
```
ntsim, line 16: Given that it looks like we've proven that the status check works, is there any value anymore in having this code?

Would probably argue that we could just delete all this long-running function code as it's just cluttering up the project.

If we think there's a reason to keep it, would suggest that we at least [disable the trigger](https://learn.microsoft.com/en-us/azure/azure-functions/disable-function?tabs=portal) to prevent it from being triggered accidentally when we don't need it.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/StatusCheckFunction.cs
```diff
@@ -0,0 +1,33 @@
...
+
+public class StatusCheckFunction
+{
+    private static readonly OrchestrationQuery ActiveOrchestrationsQuery = new()
+    {
+        Statuses = new List<OrchestrationRuntimeStatus>
+        {
+            OrchestrationRuntimeStatus.Pending,
+            OrchestrationRuntimeStatus.Running
+        }
```
ntsim, line 16: Nitpick, but could use a collection expression?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/StatusCheckFunction.cs
```diff
@@ -0,0 +1,33 @@
...
+    };
+
+    [Function("StatusCheck")]
```
ntsim, line 19: Could use `nameof`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/StatusCheckFunction.cs
```diff
@@ -0,0 +1,33 @@
...
+    public static async Task<IActionResult> StatusCheck(
+        [HttpTrigger(AuthorizationLevel.Anonymous, "get")]
+#pragma warning disable IDE0060
+        HttpRequest request,
+#pragma warning restore IDE0060
+        [DurableClient] DurableTaskClient client)
+    {
+        var activeOrchestrations = await client
+            .GetAllInstancesAsync(filter: ActiveOrchestrationsQuery)
+            .ToListAsync();
```
ntsim, line 30: Could `CountAsync`?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/StatusCheckFunction.cs
```diff
@@ -0,0 +1,33 @@
...
+        return new OkObjectResult(new { ActiveOrchestrations = activeOrchestrations.Count });
```
ntsim, line 31: Would it be better to return a 5xx status code if there are active orchestrations?

The main benefit is that we could reuse the `wait-for-endpoint-success.yml` task rather than need the additional `wait-for-orchestrations-to-complete.yml` task (which basically does a similar thing).

We'd probably want to also rename the function to something more intuitive for these semantics e.g. `DeployStatusCheck` (checks if the function is ready for a new deployment).

ntsim:
Seems like there's a bunch of merge conflicts from the alerts work going in!

infrastructure/templates/public-api/ci/azure-pipelines.yml
```diff
@@ -1,12 +1,21 @@
...
-  - name: deployContainerApp
-    displayName: Can we deploy the Container App yet? This is dependent on the PostgreSQL Flexible Server being set up and having users manually added.
-    default: true
-  - name: updatePsqlFlexibleServer
+  - name: deploySharedPrivateDnsZones
+    displayName: Do the shared Private DNS Zones need creating or updating?
+    default: false
+  - name: deployPsqlFlexibleServer
     displayName: Does the PostgreSQL Flexible Server require any updates? False by default to avoid unnecessarily lengthy deploys.
     default: false
```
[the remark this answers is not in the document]
duncan-at-hiveit: Hey yeah, I have investigated in the past, especially around PSQL which is a real time drain.  Didn't come away with anything concrete - apart from the tags, an identical subsequent deploy will post all of the same properties but result in a lot of churn.

Seems to be that according to the docs, it's not ARM's responsibility to ensure that subsequent updates to already-deployed resources don't result in a bunch of rework.  ARM just orchestrates stuff and then sends all of the deployment parameters off to the relevant resource provider (e.g. PostgreSQL Flexible Server).  It's up to the individual resource providers to then determine what to do with those.  An example is here: https://github.com/Azure/bicep/discussions/7001.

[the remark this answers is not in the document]
ntsim: Ah thought it'd be something like that. Bit of a shame, but I guess we're stuck with it. 

Does put a bit of a case forward for using Terraform, which maintains state and would be able to avoid making unnecessary changes. Guess we're not changing to that anytime soon though!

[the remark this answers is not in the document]
duncan-at-hiveit: Haha yes it would be nice for sure - I guess that the resource providers'll only improve over time... right...?

infrastructure/templates/public-api/components/functionApp.bicep
```diff
@@ -178,6 +182,14 @@ resource slot2FileShare 'Microsoft.Storage/storageAccounts/fileServices/shares@2
   ]
 }
 
+var firewallRules = [for (firewallRule, index) in functionAppEndpointFirewallRules: {
+  name: firewallRule.name
+  ipAddress: firewallRule.cidr
+  action: 'Allow'
+  tag: firewallRule.tag != null ? firewallRule.tag : 'Default'
+  priority: firewallRule.priority != null ? firewallRule.priority : 100 + index
```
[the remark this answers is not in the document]
duncan-at-hiveit: Yup!

infrastructure/templates/public-api/components/functionApp.bicep
```diff
@@ -195,9 +207,20 @@ var commonSiteProperties = {
...
+    ipSecurityRestrictionsDefaultAction: publicNetworkAccessEnabled && length(firewallRules) > 0 ? 'Deny' : 'Allow'
+    scmIpSecurityRestrictions: [
+      {
+        ipAddress: 'Any'
+        action: 'Allow'
+        priority: 2147483647
+        name: 'Allow all'
+        description: 'Allow all access'
+      }
+    ]
```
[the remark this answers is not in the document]
duncan-at-hiveit: I believe it's IP restrictions to determine who's able to see the part of the stack that's responsible for performing deploys to the Function App / App Service.  So ideally when we have static IPs for our runners, we'd be able to whitelist them so that only the pipeline would be able to deploy code to these resources.  Might also control who's able to access the Kudu site that accompanies each Function App / App Service host (the one you can access through `Advanced tools` in Azure Portal.

I've stuck a comment on:

```
    // TODO EES-5446 - this setting controls access to the deploy site for the Function App.
    // This is currently the default value, but ideally we would lock this down to only be accessible
    // by our runners and certain other whitelisted IP address ranges (e.g. trusted VPNs).
```

infrastructure/templates/public-api/main.bicep
duncan-at-hiveit: Rather than retain it as a variable, I've just made it inline within the `functionAppFirewallRules` property:

```
    functionAppFirewallRules: union([
      {
        name: 'Admin App Service subnet range'
        cidr: vNetModule.outputs.adminAppServiceSubnetCidr
        tag: 'Default'
        priority: 100
      }
      // TODO EES-5446 - remove service tag whitelisting when runner scale set IP range reinstated 
      {
        cidr: 'AzureCloud'
        tag: 'ServiceTag'
        priority: 101
        name: 'AzureCloud'
      }
```

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/LongRunningTriggerFunction.cs
duncan-at-hiveit: Yup good point, updated ta

src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/HttpRequestExtensions.cs
duncan-at-hiveit: Done

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/LongRunningTriggerFunction.cs
duncan-at-hiveit: Good idea, and I have disabled by adding into `deploy-data-processor.yml`:

```
"AzureWebJobs.TriggerLongRunningOrchestration.Disabled=true" \                  
```

I'd suggest we keep it for allowing the testing of this ticket, upon which it can be enabled in Azure Portal and triggered when needed.

I've also moved the `App__MetaInsertBatchSize` out of Bicep and into `deploy-data-processor.yml` as well, so now all app-specific settings are controlled in the yml.

src/GovUk.Education.ExploreEducationStatistics.Common/Extensions/HttpRequestExtensions.cs
duncan-at-hiveit: And for the bool equivalent also.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/StatusCheckFunction.cs
duncan-at-hiveit: All nitpicks welcome here!  Done - I'm almost surprised that Rider didn't suggest it also :)

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/StatusCheckFunction.cs
duncan-at-hiveit: Good spot - done ta

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/StatusCheckFunction.cs
duncan-at-hiveit: Ooh good suggestion, done ta

duncan-at-hiveit:
> Seems like there's a bunch of merge conflicts from the alerts work going in!

Yup, it was unpleasant but all resolved now thanks!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/StatusCheckFunction.cs
duncan-at-hiveit: I see what you're saying, but I'm slightly hesitant to do this because I feel that the details you get form this endpoint could be useful, and also there's no easy way to distinguish a planned 500 from an ISE 500 if one should arise for whatever reason.  

The other reason why having the count and the specific pipeline job that's asserting on the count is useful I think is that you can see during deploy in the DevOps logging how many active orchestrations are currently holding up the deploy, which gives you some indication of how long it'll take to get through.

It's not necessarily so desperately urgent at the mo because we're expecting orchestrations to be relatively quick to run, but if we end up reusing for another durable function app or we include longer orchestrations here in the future, I think it would be beneficial to have that visibility.

What do you reckon?

infrastructure/templates/public-api/components/functionApp.bicep
```diff
@@ -39,6 +39,9 @@ param privateEndpointSubnetId string?
 @description('Specifies whether this Function App is accessible from the public internet')
 param publicNetworkAccessEnabled bool = false
 
+@description('IP address ranges that are allowed to access the Function App endpoints. Dependent on "publicNetworkAccessEnabled" being true.')
+param functionAppEndpointFirewallRules FirewallRule[] = []
```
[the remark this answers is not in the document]
duncan-at-hiveit: Yup good idea - updated ta.

infrastructure/templates/public-api/components/functionApp.bicep
```diff
@@ -195,9 +207,20 @@ var commonSiteProperties = {
     netFrameworkVersion: '8.0'
     linuxFxVersion: appServicePlanOS == 'Linux' ? 'DOTNET-ISOLATED|8.0' : null
     keyVaultReferenceIdentity: keyVaultReferenceIdentity
+    publicNetworkAccess: publicNetworkAccessEnabled ? 'Enabled' : 'Disabled'
+    ipSecurityRestrictions: publicNetworkAccessEnabled && length(firewallRules) > 0 ? firewallRules : null
+    ipSecurityRestrictionsDefaultAction: publicNetworkAccessEnabled && length(firewallRules) > 0 ? 'Deny' : 'Allow'
```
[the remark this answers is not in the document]
duncan-at-hiveit: I *think* that you're right conceptually, although I'm not sure it makes much difference either way in reality.  If `publicNetworkAccessEnabled` is true and there are no firewall rules, that means it's open to the world anyway regardless of whether this is set to 'Allow' or 'Deny'.

But thinking about it now, I don't think that there's any reason to set this to anything but 'Deny' here, as we're already explicitly saying that each of our firewall rules are `Allow` rules in this same file:

```
var firewallRules = [for (firewallRule, index) in functionAppFirewallRules: {
  name: firewallRule.name
  ipAddress: firewallRule.cidr
  action: 'Allow'
  tag: firewallRule.tag != null ? firewallRule.tag : 'Default'
  priority: firewallRule.priority != null ? firewallRule.priority : 100 + index
}]
```

So I have simply set it to 'Deny' for best safety.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/StatusCheckFunction.cs
duncan-at-hiveit: (see the screenshot under `Testing pipeline changes with long-running orchestration` in the PR description for an idea of what I mean)

infrastructure/templates/public-api/components/functionApp.bicep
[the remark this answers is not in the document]
ntsim: Nice, yep that makes sense

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Processor/Functions/StatusCheckFunction.cs
ntsim: Hmm, there's potentially some alterations we could do to make this work i.e. we could change `wait-for-endpoint-success.yml` to output the body of non-200 responses. This would output the number of active orchestrations in the response JSON.

I'd also probably lean towards using 503 instead of 500 as this might be better for communicating the service isn't available to be deployed.

Regardless, it's not a big deal either way. If you're happy enough with the current approach, happy for this to go in as-is!

PR #5442, opened 2024-12-06
EES-5656 Rename Admin 'Legacy releases' to 'Release order'

benoutram:
This PR renames the Admin 'Legacy releases' page to 'Release order'.

This is being done to make it more obvious that this page allows reordering releases which will become more commonly required when 'Release labelling' is made available by EES-5626.

When release labelling is introduced, we will allow a publication to contain multiple releases with the same year and time identifier.

Our traditional way of ordering releases within a publication in reverse chronological order by year and then by time period will no longer work after we lose this time period uniqueness. There will be no secondary sorting by label and analysts will decide which release should appear first or last.

The same possibility applies to the 'latest' release for a publication which is currently determined automatically by the first release in the reverse chronological order. In future analysts will be able to decide which release it should be by ordering that release as the top entry on this page.

This change to the title is being made separately ahead of other changes to the page as part of EES-5656.

Before:
before

After:
after

### UI test result

image

### DIFF
--- a/tests/robot-tests/tests/admin_and_public/bau/release_reordering.robot
+++ b/tests/robot-tests/tests/admin_and_public/bau/release_reordering.robot
@@ -31,14 +31,14 @@ Create new publication via api
 
 Validate publication release order table is empty
     user navigates to publication page from dashboard    ${PUBLICATION_NAME}
-    user clicks link    Legacy releases
-    user waits until h2 is visible    Legacy releases
+    user clicks link    Release order
+    user waits until h2 is visible    Release order
     user checks page contains    No releases for this publication.
 
 Create first legacy release
     user navigates to publication page from dashboard    ${PUBLICATION_NAME}
-    user clicks link    Legacy releases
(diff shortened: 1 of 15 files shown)


tests/robot-tests/tests/admin_and_public/bau/release_reordering.robot
mmyoungman, line 1: I think there are more UI tests that refer to the old `Legacy releases` Admin link? 
`analyst_absence_permissions.robot`
`publication_approver_ui_permissions.robot`
`publication_owner_ui_permissions.robot`
`release_approver_ui_permissions.robot`
`release_contributtor_ui_permissions.robot`
`generate_seed_data_theme_1.robot`

I think that's it

benoutram: Thanks, good spot!
Also found the same changes needed in `release_contributor_ui_permissions.robot`.

mmyoungman (approved):

PR #5444, opened 2024-12-09
EES-5656 Release series tidy up

benoutram:
This PR contains mainly backend changes, separating out a few bits of tidy up ahead of the main changes in EES-5656.

There are two main changes amongst some other smaller improvements.

1. I attempted to remove `Id` from `ReleaseSeriesItem`, thinking that in cases where a publication has no legacy releases it would be nice if the `ReleaseSeries` field only contained an array of release id's. I backtracked on removing it entirely when I realised it's needed to be able to safely select the chosen legacy release from the release series array in the Admin UI when choosing to 'Edit' a legacy link. I've left in a lot of the changes where it doesn't need to be present though. For example, it's removed from `ReleaseSeriesItemViewModel` so it's no longer exposed to the public frontend. It's also removed from `ReleaseSeriesItemUpdateRequest` where supplying with each entry in the update made no difference, since we replace all the entries in the old value of `ReleaseSeries` with an entirely new set of entries.

2. I changed the building of the view model in `PublicationService.GetReleaseSeries` to be built from a `Release` rather than `ReleaseVersion` which is possible now the fields `Title` and `Slug` have moved to `Release`. This also makes things a bit clearer in test code where we can compare fields in the view model with expected releases, rather than release versions.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/ReleaseSeriesItemRequests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/ReleaseSeriesItemRequests.cs
@@ -1,21 +1,22 @@
 #nullable enable
-
 using System;
+using System.ComponentModel.DataAnnotations;
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Requests;
 
-public class ReleaseSeriesLegacyLinkAddRequest
+public record ReleaseSeriesLegacyLinkAddRequest
 {
-    public string Description { get; set; } = string.Empty;
-    public string Url { get; set; } = string.Empty;
+    [Required]
+    public required string Description { get; init; }
+
+    [Required]
+    public required string Url { get; init; }
 }
 
-public class ReleaseSeriesItemUpdateRequest
+public record ReleaseSeriesItemUpdateRequest
 {
-    public Guid Id { get; set; }
-    public Guid? ReleaseId { get; set; }
+    public Guid? ReleaseId { get; init; }
 
-    public string? LegacyLinkDescription { get; set; }
-    public string? LegacyLinkUrl { get; set; }
+    public string? LegacyLinkDescription { get; init; }
+    public string? LegacyLinkUrl { get; init; }
 }
-
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/ReleaseSeriesItemRequests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/ReleaseSeriesItemRequests.cs
@@ -1,15 +1,12 @@
 #nullable enable
 using System;
-using System.ComponentModel.DataAnnotations;
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Requests;
 
 public record ReleaseSeriesLegacyLinkAddRequest
 {
-    [Required]
     public required string Description { get; init; }
 
-    [Required]
     public required string Url { get; init; }
 }
 
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
@@ -22,11 +22,10 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
+using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Validators.ValidationErrorMessages;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Validators.ValidationUtils;
 using ExternalMethodologyViewModel = GovUk.Education.ExploreEducationStatistics.Admin.ViewModels.ExternalMethodologyViewModel;
-using IPublicationRepository = GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.IPublicationRepository;
-using IPublicationService = GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.IPublicationService;
 using IReleaseVersionRepository = GovUk.Education.ExploreEducationStatistics.Content.Model.Repository.Interfaces.IReleaseVersionRepository;
 using PublicationViewModel = GovUk.Education.ExploreEducationStatistics.Admin.ViewModels.PublicationViewModel;
 using ReleaseSummaryViewModel = GovUk.Education.ExploreEducationStatistics.Admin.ViewModels.ReleaseSummaryViewModel;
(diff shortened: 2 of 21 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -538,56 +519,48 @@ public async Task<Either<ActionResult, List<ReleaseSeriesTableEntryViewModel>>>
...
 
                     // Check all publication releases are included in updatedReleaseSeriesItems
-                    var publicationReleaseIds = publication.ReleaseVersions
-                        .Select(rv => rv.ReleaseId)
-                        .Distinct()
-                        .ToList();
+                    var publicationReleaseIds = await _context.Releases
+                        .Where(r => r.PublicationId == publicationId)
+                        .Select(r => r.Id)
+                        .ToListAsync();
```
mmyoungman, line 536: I think this means rather than one query with a join, we now do two separate queries? I'm not fussed either way, but interested why you decided to make the change

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -450,43 +449,28 @@ public async Task<Either<ActionResult, List<ReleaseSeriesTableEntryViewModel>>>
...
                         else
                         {
-                            // prefer getting the latest published version over an unpublished amendment
-                            var latestVersion = await _context.ReleaseVersions
-                                .LatestReleaseVersion(seriesItem.ReleaseId!.Value, publishedOnly: true)
-                                .SingleOrDefaultAsync();
+                            var latestReleaseVersion = await _context.ReleaseVersions
+                                .Include(rv => rv.Release)
+                                .LatestReleaseVersion(releaseId: seriesItem.ReleaseId!.Value)
+                                .SingleAsync();
```
mmyoungman, line 462: What happens if we grab an unpublished amendment with an updated slug, but that new slug hasn't gone live yet? If release redirects are enabled, I think this will be possible?

Previously I think we specifically grabbed the latest published release to ensure we put the correct slug in the returned view model.

benoutram: It shouldn't be possible to have release versions belonging to the same release with differing slugs after we put in a change which prevents changing the year and time identifier on amendment versions. It will be more obvious that this isn't allowed when EES-5679 makes a change to the UI. It's currently only blocked by the fact the BE no longer allows it.

Release redirects are necessary because in future release label will become part of the release slug and we will allow changing the label and add a redirect to the slug containing the old label.

I've just made another change here though because I had failing unit tests. In this change it becomes a bit clearer that the slug returned in the view model is from the release relating to the release series item, and nothing to do with the release version. The `Slug` field will be removed from `ReleaseVersion` by EES-5658.

Before

```csharp
var latestReleaseVersion = await _context.ReleaseVersions
    .Include(rv => rv.Release)
    .LatestReleaseVersion(releaseId: seriesItem.ReleaseId!.Value)
    .SingleAsync();

var release = latestReleaseVersion!.Release;

result.Add(new ReleaseSeriesTableEntryViewModel
{
    Id = seriesItem.Id,
    ReleaseId = release.Id,
    Description = release.Title,
    ReleaseSlug = release.Slug,
    IsLatest = latestReleaseVersion.Id == publication.LatestPublishedReleaseVersionId,
    IsPublished = latestReleaseVersion.Version > 1 ||
                  latestReleaseVersion is { Version: 0, Live: true }
});
```

After
```csharp
var release = await _context.Releases
    .SingleAsync(r => r.Id == seriesItem.ReleaseId);

var latestPublishedReleaseVersion = await _context.ReleaseVersions
    .LatestReleaseVersion(releaseId: seriesItem.ReleaseId!.Value, publishedOnly: true)
    .SingleOrDefaultAsync();

result.Add(new ReleaseSeriesTableEntryViewModel
{
    Id = seriesItem.Id,
    ReleaseId = release.Id,
    Description = release.Title,
    ReleaseSlug = release.Slug,
    IsLatest = latestPublishedReleaseVersion?.Id == publication.LatestPublishedReleaseVersionId,
    IsPublished = latestPublishedReleaseVersion != null
});
```

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
benoutram: I'm not sure I follow where the query before with the join was or where the two separate queries come from.

The code in the comment here should be generating something equivalent to `SELECT Id fFROM Releases WHERE PublicationId = <publicationId>`

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
mmyoungman: Discussed this on Slack.

I was thinking that an unpublished amendment could change the publication slug, and that would only go live when the amendment is published, but this is incorrect. A publication slug, when changed, always goes live immediately.

I was getting confused with methodology slugs, and those can changed without going live when attached to an unpublished amendment.

TLDR the above code is fine.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
benoutram: Ah I see your point now, yes you are correct, there was a join between `Publications` and `ReleaseVersions` before but including the `ReleaseVersions` collection in that way that meant every column of `ReleaseVersions` would be returned for every publication in the result of the query. This was a lot more data about release versions than what was needed which is just the release id.

Now we only select the id's in the second query. We've also changed this to get the Id's from `Releases` directly by publication id without `ReleaseVersions` meaning there's no longer any need for a distinct.

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/ReleaseSeriesItemRequests.cs
```diff
@@ -1,21 +1,22 @@
...
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Requests;
 
-public class ReleaseSeriesLegacyLinkAddRequest
+public record ReleaseSeriesLegacyLinkAddRequest
 {
-    public string Description { get; set; } = string.Empty;
-    public string Url { get; set; } = string.Empty;
+    [Required]
+    public required string Description { get; init; }
```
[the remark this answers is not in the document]
benoutram: They have different effects, yes. 

The `required` keyword was new in C#11 and means the field must be initialized at the same time the object is constructed and is enforced at compile time. For example, if you look in the changes to the 

we can no longer do this

```csharp
new ReleaseSeriesLegacyLinkAddRequest()
````

which we wouldn't really want to because it doesn't make sense to have a request with no values, and instead must do something like this

```csharp
new ReleaseSeriesLegacyLinkAddRequest
{
    Description = "Test description",
    Url = "https://test.url"
};
```

The `RequiredAttribute` used with `[Required] provides validation that the field has a value in the request.

I imagine that if you don't use the `[Required]` annotation but still use `required`, and the value is not present in the request, that the request won't fail validation and then there might be an error attempting to deserialise the JSON into the object due to the field marked as `required` not being initialised. I'd have to try this out to see if this is true. I think you are right to question it though because if this is true you might expect that type of error to be returned as a validation error, but I don't think that's currently how it works.

mmyoungman (approved):

PR #5445, opened 2024-12-09
EES-5396 Sync public API OpenAPI changes with API docs

ntsim:
This PR adds a syncing mechanism to ensure that OpenAPI changes in the public API are documented in the API docs.

To enable this:

1. The public API now generates its OpenAPI documents at build time using the Swashbuckle CLI.
2. We use [oasdiff](https://github.com/Tufin/oasdiff) to diff between the OpenAPI documents used by the API docs and those generated by the public API.
3. If there are any changes in the diff, we fail the build pipeline as this means there are **undocumented** changes.
4. To resolve any undocumented changes, developers need to copy over any changes from the public API to the relevant OpenAPI document in the API docs.

Once this diff has been resolved, the build can pass and deployment can proceed.

The API docs deployment has now been adjusted so that they are built during the build itself. We need to do this as 

## Related changes

- Removed `/api` prefix from `openapi-v1.json` in the API docs.
- Moved Swagger config into public API's `Startup` to enable the Swashbuckle CLI to work.

## Other changes

- Disabled Swagger UI in non-dev environments (i.e. test, pre-prod, prod). This is consistent with our other backend APIs and avoids any problems from hosting the Swagger UI (it's expensive to generate).

### DIFF
--- a/azure-pipelines-main.yml
+++ b/azure-pipelines-main.yml
@@ -15,6 +15,8 @@ variables:
   NodeVersion: '20.16.0'
   RubyVersion: '3.2'
   AcrServiceConnection: s101d-datahub-spn-ees-dfe-gov-uk-docker-managed-service-connection
+  PublicApiDir: GovUk.Education.ExploreEducationStatistics.Public.Data.Api
+  PublicApiDocsDir: explore-education-statistics-api-docs
 
 trigger:
   branches:
@@ -71,6 +73,25 @@ jobs:
           projects: src/GovUk.Education.ExploreEducationStatistics.sln
 
       # TODO: Wrap the above three tasks up into a single `dotnet format` task once all TODOs are done
+
+      - task: DotNetCoreCLI@2
+        displayName: Build Public API
+        inputs:
+          projects: '**/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.csproj'
+          arguments: --configuration $(BuildConfiguration)
+
+      - task: Bash@3
+        displayName: Diff Public API OpenAPI docs
+        inputs:
+          targetType: inline
+          script: |
+            docker run --rm -t \
+              -v $(System.DefaultWorkingDirectory)/src/$(PublicApiDir)/bin/$(BuildConfiguration)/net8.0:/api \
+              -v $(System.DefaultWorkingDirectory)/src/$(PublicApiDocsDir):/api-docs \
+              -v $(System.DefaultWorkingDirectory)/ci/scripts:/scripts \
+              --entrypoint /bin/sh \
+              tufin/oasdiff \
+              /scripts/public-api-openapi-diff.sh
 
   - job: BackendTest
     pool: ees-ubuntu2204-xlarge
--- a/azure-pipelines-main.yml
+++ b/azure-pipelines-main.yml
@@ -163,7 +163,6 @@ jobs:
           projects: '**/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.csproj'
           arguments: -r linux-musl-x64 --configuration $(BuildConfiguration) --output $(Build.ArtifactStagingDirectory)/public-api
           zipAfterPublish: False
-
       - task: Docker@2
         displayName: Build Public API Docker image
         condition: and(succeeded(), eq(variables.IsBranchDeployable, true))
@@ -315,7 +314,7 @@ jobs:
           projects: '**/GovUk.Education.ExploreEducationStatistics.Admin.csproj'
           arguments: --self-contained true -r win-x64 --configuration $(BuildConfiguration) --output $(Build.ArtifactStagingDirectory)
 
-      - task: PublishPipelineArtifact@0
+      - task: PublishPipelineArtifact@1
         condition: and(succeeded(), eq(variables.IsBranchDeployable, true))
         displayName: Publish Admin artifact
         inputs:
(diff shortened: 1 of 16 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
```diff
@@ -152,6 +154,14 @@ public void ConfigureServices(IServiceCollection services)
         services.AddApiVersioning().AddMvc().AddApiExplorer();
         services.AddEndpointsApiExplorer();
 
+        // Swagger
+
+        if (!hostEnvironment.IsIntegrationTest())
```
duncan-at-hiveit, line 172: I'm guessing that the need to move all of these things is that `Program.cs` is only invoked when fully starting the service in a hosted environment, whereas the Swashbuckle CLI need only fire up the DI container?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.csproj
```diff
@@ -53,5 +53,26 @@
     <ProjectReference Include="..\GovUk.Education.ExploreEducationStatistics.Public.Data.Services\GovUk.Education.ExploreEducationStatistics.Public.Data.Services.csproj"/>
     <ProjectReference Include="..\GovUk.Education.ExploreEducationStatistics.Public.Data.Utils\GovUk.Education.ExploreEducationStatistics.Public.Data.Utils.csproj"/>
   </ItemGroup>
+  
+  <Target Name="BuildOpenApiFiles" AfterTargets="Build">
```
duncan-at-hiveit, line 57: I had no idea you could do this!  I figured it'd be something invoked in the pipeline.  Nice.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.csproj
```diff
@@ -53,5 +53,26 @@
...
+    <!-- Add new API versions to this group -->
```
duncan-at-hiveit, line 58: I don't quite understand what the purpose of this block is?

infrastructure/templates/public-api/application/public-api/publicApiApp.bicep
```diff
@@ -33,6 +33,9 @@ param appInsightsConnectionString string
 @description('Whether to create or update Azure Monitor alerts during this deploy')
 param deployAlerts bool
 
+@description('Enable the Swagger UI')
+param enableSwagger bool = false
```
duncan-at-hiveit, line 37: I've taken to removing default values from Bicep files that are only invoked once (e.g. this file is only ever called from `main.bicep`, and `main.bicep` is always providing a value for `enableSwagger`.  And if for some reason it *isn't* providing anything when it really should be, something has gone wrong and we'll find out about it at deploy time.

For lower-level Bicep templates that stand more chance of being reused in various other contexts e.g. `storageAccount.bicep`, then it totally *does* make sense to provide more default values though.

Are you happy with this convention?

azure-pipelines-main.yml
```diff
@@ -31,36 +32,28 @@ pr:
   - test
 
 jobs:
-  - job: Backend
-    pool: ees-ubuntu2204-xlarge
+  - job: BackendVerify
```
duncan-at-hiveit, line 36: Been looking forward to these coming in!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
ntsim: I _think_ the Swashbuckle CLI can technically detect `Program` and run that (e.g. for minimal APIs), but because we have both `Program` and `Startup`, it seems to preferentially detect `Startup` and boots up an application using that. Consequently none of the Swagger config was being picked up when it was in `Program`, hence having to move it.

This has helped me realise though that this boolean was supposed to be `if (_appOptions.EnableSwagger)`, so I've changed that now :smile:

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.csproj
ntsim: In the future, when we want to publish v1+ API versions, we would add extra lines for each version e.g.

```xml
<ItemGroup>
  <Version Include="1" />
  <Version Include="2" />
</ItemGroup>
```

Unfortunately, the Swashbuckle CLI doesn't have an option to generate all the available OpenAPI docs automatically. You have to manually provide the name of each document (i.e. the API version number) to the CLI.

After thinking about it some more, I went down a rabbit hole trying to create a custom MSBuild task that could fetch the API version numbers for us and generate the docs. 

Unfortunately, after trying a few things, I haven't been able to get the custom task working. I think in the interest of time, whilst not ideal, we stick with the current solution of adding these version numbers manually.

infrastructure/templates/public-api/application/public-api/publicApiApp.bicep
ntsim: Cool, that's fine with me! Removed the default now

ci/scripts/public-api-openapi-diff.sh
```diff
@@ -0,0 +1,60 @@
+#/usr/bin/env sh
+
+set -u
```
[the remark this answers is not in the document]
ntsim: Originally, I was using the standard `set -eux` that's typically recommended for scripts, however, `-e` was causing issues with aggregating all the errors from the `for` loop that runs `oasdiff`. 

I was messing around with the flags to get it working, but just ended up forgetting about it once I got the main logic working.

Setting `-u` doesn't really cause any issues, but to keep it simple, I've now removed it

infrastructure/templates/public-api/ci/jobs/deploy-api-docs.yml
```diff
@@ -12,20 +12,43 @@ jobs:
...
+              displayName: Install Gems
               inputs:
+                workingDirectory: $(docsPath)
                 targetType: inline
-                script: |
-                  set -e
-                  pip install --force-reinstall azure-cli==2.64.0
+                script: bundle install
+
+            - task: Bash@3
```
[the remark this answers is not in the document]
ntsim: We did previously, but now that we're using separate docs sites for each environment, we need to build these separately for each environment to inject different environment variables (e.g. the API URLs). It's an unfortunate limitation of static site as we can't change these server-side at runtime!

infrastructure/templates/public-api/application/public-api/publicApiApp.bicep
duncan-at-hiveit: Ta!

ci/scripts/public-api-openapi-diff.sh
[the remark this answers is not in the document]
duncan-at-hiveit: Nice one - will remember the `set -eux` for another day.  FYI `set -e` was causing me errors elsewhere too as a curl error response was also causing pipeline tasks to fail immediately whereas I just wanted them to retry again.  I suspect that there's a way to get it to behave but didn't have too much time to sink into it.

infrastructure/templates/public-api/ci/jobs/deploy-api-docs.yml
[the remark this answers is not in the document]
duncan-at-hiveit: Nice one - that makes a lot of sense!  Need to get our Bicep files validated in the build too for that early feedback (and maybe our pipeline yml files too).

duncan-at-hiveit (approved):

PR #5456, opened 2024-12-12
EES-5598 Add Geographic lvl filter to Data Catalogue page

mmyoungman:
This PR adds the frontend code to filter by Geographic Level on the Data Catalogue page.

Most of the backend work was completed in EES-5691.

There are a few backend changes made to allow the frontend to provide a geographic level code to the frontend - the backend turns the code into a `GeographicLevel` via `DataSetFileListRequest.GeographicLevelEnum`.

We also add a generator for the new `DataSetFileMeta`, but set it to null. We do this because the InMemory DB does not support EF8 JSON columns, so any test that tries to save a File with DataSetFileMeta set will fail. Our workaround for this issue, if a test needs to set DataSetFileMeta, is to mock it out via a repository - this was done in EES-5691. 

:rotating_light: This PR should be merged into the EES-5691 branch before merging the EES-5691 PR itself :rotating_light:

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/FunctionsIntegrationTest.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/FunctionsIntegrationTest.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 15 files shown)


mmyoungman:
Going with a different approach - see jira ticket

PR #5457, opened 2024-12-12
EES-5627 Adding `Label` to the `Release`

jack-hive:
This PR adds a new **optional** `Label` database column to the `Release`.

It has been set to have a max character limit of `50`.

The plan is, in a subsequent PR, to change how we calculate the `Slug` for a `Release` such that it appends this `Label` to the end, prefixed with a `-`. For this reason, the character limit of the `Slug` has been **extended** by `51`.

As part of this, we have changed the existing unique index that exists on the `PublicationId`/`Year`/`TimePeriodCoverage` columns, and have added the `Label` to the mix.

## Other Changes
- We have added the support to change the `Label` in the `ReleaseGeneratorExtensions` in the spirit of avoiding conflicts in some further work that might be coming along at the same time (namely, EES-5628 and EES-5634)

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20241212105726_EES5627_AddingLabelToRelease.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20241212105726_EES5627_AddingLabelToRelease.cs
@@ -1,0 +1,70 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Migrations.ContentMigrations
+{
+    /// <inheritdoc />
+    public partial class EES5627_AddingLabelToRelease : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropIndex(
+                name: "IX_Releases_PublicationId_Year_TimePeriodCoverage",
+                table: "Releases");
+
+            migrationBuilder.AlterColumn<string>(
+                name: "Slug",
+                table: "Releases",
+                type: "nvarchar(81)",
+                maxLength: 81,
+                nullable: false,
+                oldClrType: typeof(string),
+                oldType: "nvarchar(30)",
+                oldMaxLength: 30);
+
+            migrationBuilder.AddColumn<string>(
+                name: "Label",
+                table: "Releases",
+                type: "nvarchar(50)",
+                maxLength: 50,
+                nullable: true);
+
+            migrationBuilder.CreateIndex(
+                name: "IX_Releases_PublicationId_Year_TimePeriodCoverage_Label",
+                table: "Releases",
+                columns: new[] { "PublicationId", "Year", "TimePeriodCoverage", "Label" },
+                unique: true,
+                filter: "[Label] IS NOT NULL");
+        }
+
+        /// <inheritdoc />
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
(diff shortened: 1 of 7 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Model/Release.cs
```diff
@@ -24,6 +24,8 @@ public class Release : ICreatedUpdatedTimestamps<DateTime, DateTime?>
 
     public required int Year { get; set; }
 
+    public string? Label { get; set; }
```
benoutram, line 27: You could consider adding a change to ReleaseGeneratorExtensions to support `Label`. It might help avoid merge conflicts if both EES-5628 and EES-5634 both come along at the same time and need it.

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20241212105726_EES5627_AddingLabelToRelease.cs
```diff
@@ -0,0 +1,70 @@
...
+                type: "nvarchar(50)",
+                maxLength: 50,
+                nullable: true);
+
+            migrationBuilder.CreateIndex(
+                name: "IX_Releases_PublicationId_Year_TimePeriodCoverage_Label",
+                table: "Releases",
+                columns: new[] { "PublicationId", "Year", "TimePeriodCoverage", "Label" },
+                unique: true,
+                filter: "[Label] IS NOT NULL");
```
benoutram, line 39: Out of interest, does this unique constraint still take effect for the combination of PublicationId, Year and TimePeriodCoverage when Label IS null? We'd hope it does! 🙂

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Release.cs
jack-hive: Good shout. I'll add it

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20241212105726_EES5627_AddingLabelToRelease.cs
jack-hive: I would hope so! I'll double check

jack-hive: I've fixed this with adding `HasFilter(null)` to the configuration for the index

benoutram: I wasn't sure how the null filter applied by the code `HasFilter(null)` is getting related specifically to the `Label` column.

I think the documentation [here](https://learn.microsoft.com/en-us/ef/core/modeling/indexes?tabs=data-annotations#index-filter) covers this. It says 

> When using the SQL Server provider EF adds an 'IS NOT NULL' filter for all nullable columns that are part of a unique index. To override this convention you can supply a null value.

So I think the answer is that it's probably not exclusive to the `Label` column, and would be applicable to other nullable columns if there were any.

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20241212105726_EES5627_AddingLabelToRelease.cs
jack-hive: That's correct yes, that's also what I found :)

PR #5459, opened 2024-12-12
EES-5544 Change data set import page accordions into a table

bennettstuart:
See [Ticket EES-5544](https://dfedigital.atlassian.net/browse/EES-5544)

image

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/data/components/DataFilesTable.module.scss
+++ b/src/explore-education-statistics-admin/src/pages/release/data/components/DataFilesTable.module.scss
@@ -20,17 +20,12 @@
 }
 
 .actions {
+  --verticalSpacing: #{govuk-spacing(1)};
   margin-bottom: 0;
 
   @include govuk-media-query($until: desktop) {
     align-items: start;
     flex-direction: column;
-
-    a,
-    button {
-      margin-bottom: 1rem;
-      margin-left: 0;
-      white-space: nowrap;
-    }
+    white-space: nowrap;
   }
 }
--- a/src/explore-education-statistics-admin/src/pages/release/data/components/DataFilesTable.module.scss
+++ b/src/explore-education-statistics-admin/src/pages/release/data/components/DataFilesTable.module.scss
@@ -2,21 +2,35 @@
 
 .table {
   td {
-    padding-bottom: govuk-spacing(2);
-    padding-top: govuk-spacing(2);
     vertical-align: middle;
   }
+}
 
-  .fileSize {
-    width: 5rem;
-    text-align: right;
-  }
+.title {
+  max-width: 500px;
+}
 
-  .fileStatus {
-    max-width: 164px;
-  }
+.fileSize {
+  width: 5rem;
+  text-align: right;
+}
 
-  .actions {
-    margin-bottom: 0;
+.fileStatus {
+  max-width: 126px;
+}
+
+.actions {
+  margin-bottom: 0;
+
+  @include govuk-media-query($until: desktop) {
+    align-items: start;
+    flex-direction: column;
+
+    a,
+    button {
+      margin-bottom: 1rem;
+      margin-left: 0;
+      white-space: nowrap;
+    }
   }
 }
--- a/src/explore-education-statistics-admin/src/pages/release/data/components/DataFilesTable.module.scss
+++ b/src/explore-education-statistics-admin/src/pages/release/data/components/DataFilesTable.module.scss
@@ -1,0 +1,22 @@
+@import '~govuk-frontend/dist/govuk/base';
+
+.table {
+  td {
+    padding-bottom: govuk-spacing(2);
+    padding-top: govuk-spacing(2);
+    vertical-align: middle;
+  }
+
+  .fileSize {
+    width: 5rem;
+    text-align: right;
+  }
+
+  .fileStatus {
+    max-width: 164px;
+  }
+
+  .actions {
+    margin-bottom: 0;
+  }
+}
--- a/src/explore-education-statistics-admin/src/pages/release/data/components/DataFilesTable.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/data/components/DataFilesTable.tsx
@@ -189,7 +189,7 @@ export default function DataFilesTable({
                           showClose
                           title="Cannot delete files"
                           triggerButton={
-                            <ButtonText className="govuk-!-margin-left-3">
+                            <ButtonText variant="warning">
                               Delete files
                             </ButtonText>
                           }
--- a/src/explore-education-statistics-admin/src/pages/release/data/components/DataFilesTable.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/data/components/DataFilesTable.tsx
@@ -96,11 +96,11 @@ const DataFilesTable = ({
       <tbody>
         {dataFiles.map(dataFile => (
           <tr key={dataFile.title}>
-            <td>{dataFile.title}</td>
-            <td className={styles.fileSize}>
+            <td data-testid="Subject title">{dataFile.title}</td>
+            <td data-testid="Data file size" className={styles.fileSize}>
               {dataFile.fileSize.size.toLocaleString()} {dataFile.fileSize.unit}
             </td>
-            <td>
+            <td data-testid="Status">
               <ImporterStatus
                 className={styles.fileStatus}
                 releaseId={releaseId}
@@ -108,7 +108,7 @@ const DataFilesTable = ({
                 onStatusChange={onStatusChange}
               />
             </td>
-            <td>
+            <td data-testid="Actions">
               <ButtonGroup className={styles.actions}>
                 <Modal
                   showClose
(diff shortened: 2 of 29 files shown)


src/explore-education-statistics-admin/src/pages/release/data/components/DataFilesTable.tsx
```diff
@@ -0,0 +1,239 @@
...
+  onCancelReordering: () => void;
+  onConfirmReordering: (nextSeries: DataFile[]) => void;
+  onDeleteFile: (dataFile: DataFile) => Promise<void>;
+  onStatusChange: (
+    dataFile: DataFile,
+    { totalRows, status }: DataFileImportStatus,
+  ) => Promise<void>;
+}
+
+const DataFilesTable = ({
```
ntsim, line 43: Could change to use `export default function`

src/explore-education-statistics-admin/src/pages/release/data/components/DataFilesTable.tsx
```diff
@@ -0,0 +1,239 @@
...
+              {dataFile.fileSize.size.toLocaleString()} {dataFile.fileSize.unit}
+            </td>
+            <td data-testid="Status">
+              <ImporterStatus
+                className={styles.fileStatus}
+                releaseId={releaseId}
+                dataFile={dataFile}
+                onStatusChange={onStatusChange}
+              />
+            </td>
```
ntsim, line 110: Currently looks a bit horrible when a data replacement is in progress as we get a horrible big blue box. 

Think we could get rid of the 'Data' part of the replacement label so we don't get wrapping over 3 lines:

image

src/explore-education-statistics-admin/src/pages/release/data/components/DataFilesTable.tsx
```diff
@@ -0,0 +1,239 @@
...
+                            )}
+                          >
+                            Edit title
+                          </Link>
+                          {dataFile.publicApiDataSetId ? (
+                            <Modal
+                              showClose
+                              title="Cannot replace data"
+                              triggerButton={
+                                <ButtonText>Replace data</ButtonText>
```
ntsim, line 146: At smaller screens the word 'Replace' breaks across multiple lines and it looks quite weird:

image

src/explore-education-statistics-admin/src/pages/release/data/components/ReleaseDataUploadsSection.tsx
```diff
@@ -140,6 +117,17 @@ const ReleaseDataUploadsSection = ({
...
 
+  const handleDeleteFile = async (dataFile: DataFile) => {
+    releaseDataFileService
+      .getDeleteDataFilePlan(releaseId, dataFile)
+      .then(plan => {
+        setDeleteDataFile({
+          plan,
+          file: dataFile,
+        });
+      });
```
ntsim, line 128: Can rewrite to use awaits

tests/robot-tests/tests/admin/bau/upload_files.robot
```diff
@@ -35,23 +35,19 @@ Upload a ZIP file subject
...
     # To ensure "Data file size" and "Number of rows" will be filled
     user waits until page does not contain    Queued    %{WAIT_MEDIUM}
-
-    user checks headed table body row contains    Subject title    Absence in PRUs    ${section}
-    user checks headed table body row contains    Data file    absence_in_prus.csv    ${section}
-    user checks headed table body row contains    Metadata file    absence_in_prus.meta.csv    ${section}
-    user checks headed table body row contains    Data file size    141 Kb    ${section}
-    user checks headed table body row contains    Number of rows    612    ${section}
-    user checks headed table body row contains    Status    Complete    ${section}    %{WAIT_DATA_FILE_IMPORT}
+    user checks table cell contains    row=1    column=1    expected=Absence in PRUs    parent=testid:Data files table
```
ntsim, line 42: Why use named parameters? This is pretty inconsistent with how we typically use keywords. Including the parameter names makes the lines much more noisy and hinders readability.

In generally, we use named parameters sparingly throughout the tests e.g. to set optional parameters. We don't usually use them for required parameters like this.

Would revert back to using positional arguments only. Same applies for rest of Robot tests where we've made this change.

tests/robot-tests/tests/libs/admin-common.robot
```diff
@@ -575,25 +574,44 @@ user uploads subject
...
     END
 
-    user checks headed table body row contains    Status    ${IMPORT_STATUS}    ${section}    %{WAIT_DATA_FILE_IMPORT}
+    user waits until page contains data uploads table
+    ${row}=    user gets table row    row_cell_text=${SUBJECT_NAME}    parent=testid:Data files table
+    ...    wait=%{WAIT_SMALL}
+    ${status_cell}=    user gets testid element    id=Status    wait=%{WAIT_SMALL}    parent=${row}
+    user scrolls to element    ${status_cell}
+    user waits until element contains    element=${status_cell}    text=${IMPORT_STATUS}
+    ...    wait=%{WAIT_DATA_FILE_IMPORT}
```
ntsim, line 589: Similar to earlier comment - would switch back to using positional arguments for required parameters.

tests/robot-tests/tests/libs/tables-common.robot
```diff
@@ -23,9 +23,10 @@ user checks table row heading contains
     user waits until element contains    xpath://table/tbody/tr[${row}]/th[${column}]    ${expected}
 
 user checks table cell contains
-    [Arguments]    ${row}    ${column}    ${expected}    ${parent}=css:table
+    [Arguments]    ${row}    ${column}    ${expected}    ${parent}=css:table    ${wait}=%{WAIT_SMALL}
```
ntsim, line 26: Does this need a wait? Typically, we could use some other keyword to check that the table (or something in it) has rendered before checking the table cells.

Most 'check' keywords don't include waits for this reason, so this is fairly inconsistent.

tests/robot-tests/tests/libs/tables-common.robot
```diff
@@ -58,7 +59,8 @@ user gets table row with heading
     [Return]    ${elem}
 
 user gets table row
-    [Arguments]    ${row_cell_text}    ${parent}=css:table
+    [Arguments]    ${row_cell_text}    ${parent}=css:table    ${wait}=%{WAIT_SMALL}
```
ntsim, line 62: As above comment - does this need a wait? Similarly, we should be able to use some other keyword to wait for the table (or something in it) before checking table rows.

ntsim (changes requested):

tests/robot-tests/tests/admin/bau/upload_files.robot
bennettstuart: I feel it makes the test easier to understand, instead of seemingly random strings appended to the end of user created keywords/functions - it's not immediately obvious what the function is doing in my opinion (which is a robot newb's opinion)

I'll remove the names

tests/robot-tests/tests/libs/tables-common.robot
bennettstuart: I suppose I'm exposing the wait param of the underlying `user waits until parent contains element`, am I to rename this to `user waits until table cell contains`?

bennettstuart: Was added due to table updating asynchronously to the form submit, but have abstracted the waits into separate keywords

src/explore-education-statistics-admin/src/pages/release/data/components/DataFilesTable.module.scss
```diff
@@ -0,0 +1,36 @@
...
+.actions {
+  margin-bottom: 0;
+
+  @include govuk-media-query($until: desktop) {
+    align-items: start;
+    flex-direction: column;
+
+    a,
+    button {
+      margin-bottom: 1rem;
```
ntsim, line 31: Seems like the spacing is a bit excessive now:

image

Additionally, we should be using `govuk-spacing` function for this.

1 spacing unit:

image

2 spacing units:

image

Feels like 1 spacing unit might actually be fine

src/explore-education-statistics-admin/src/pages/release/data/components/DataFilesTable.tsx
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  return (
+    <table className={styles.table} data-testid="Data files table">
+      <thead>
+        <tr>
+          <th scope="col">Title</th>
+          <th scope="col">Size</th>
+          <th scope="col">Status</th>
+          <th scope="col">Actions</th>
```
ntsim, line 92: Seems like we're missing a bunch of columns that appear in the ticket and acceptance criteria? Is this on purpose? If so, the ticket needs to be updated. 

If not, looks like we need to add those in. We'll need to update the tests accordingly as well!

bennettstuart: With the addition of the "View details" modal, we've opted to remove some columns to make the rows more concise at first glance wit the added option to see more information. [Cam and I have discussed briefly here](https://dfedigital.atlassian.net/browse/EES-5544?focusedCommentId=100335) in the ticket comments

ntsim: The ticket has been updated accordingly now

ntsim (approved):

PR #5462, opened 2024-12-12
EES-5587 refactor reordering filters and indicators

amyb-hiveit:
Adds a new component for reordering nested lists which uses the accessible `ReorderableList` component.

Implements it for reordering filters and indicators.

Screenshot 2024-12-12 162858


bennettstuart (approved):
That's ace!

One thought I did have is around general list items, as to whether the move up/down buttons should be enabled/disabled instead of conditionally rendered. I kept making mistakes when clicking the move buttons because the position of the button would be different after moving an item out of the first/last position - hope this makes sense, be happy to hear your thoughts.

amyb-hiveit:
> That's ace!
> 
> One thought I did have is around general list items, as to whether the move up/down buttons should be enabled/disabled instead of conditionally rendered. I kept making mistakes when clicking the move buttons because the position of the button would be different after moving an item out of the first/last position - hope this makes sense, be happy to hear your thoughts.

Yeah I get what you mean, there are accessibility issues with disabled buttons though and GDS say:

> Disabled buttons have poor contrast and can confuse some users, so avoid them if possible. Only use disabled buttons if research shows it makes the user interface easier to understand.
https://design-system.service.gov.uk/components/button/

So it might be good to get some feedback from users on this @cjrace

PR #5468, opened 2024-12-13
EES-5738 Add filtering by geographic level for Data Catalogue page

mmyoungman:
This PR allows uses of the public Data Catalogue page to filter data sets by geographic level.

:rotating_light: There is an Admin migration endpoint to hit (`PUT /bau/migrate-datasetfile-geographiclevels?isDryRun=false`) as this is deployed across environments. :rotating_light:

The original approach for this feature was to use EF8 JSON column/query support to make the Files.DataSetFileMeta column queryable. While this approach actually worked, we couldn't test it due to not having any support with the InMemory DB we use in backend testing. For this reason we switched approach to duplicating DataSetFileMeta's geographic levels into a table to make them queryable for the Data Catalogue page.

It's worth additionally mentioning that while the EF JSON column approach did work, it required a lot of finessing to make it work. That, along with no support from the InMemory DB, made me think that persisting with that approach would only lead to more future problems and that it was better to go with a simpler approach.

The new table is named `DataSetFileGeographicLevels`. This is because we intend to pull data and meta files out of the Files table and into new `DataSetFile`/`DataSetFileVersion` tables. This work is going to be done after this work because the priority was to provide this feature to end users first.

We also have work planned (also to happen in the near future) to remove `Files.DataSetFileMeta` and make all aspects of the meta queryable by bringing the meta data out of JSON and into tables. This PR takes one step in that direction.


src/GovUk.Education.ExploreEducationStatistics.Content.Api/Controllers/DataSetFilesController.cs
```diff
@@ -39,6 +40,7 @@ public async Task<ActionResult<PaginatedListViewModel<DataSetFileSummaryViewMode
                 themeId: request.ThemeId,
                 publicationId: request.PublicationId,
                 releaseVersionId: request.ReleaseId,
+                geographicLevel: request.GeographicLevelEnum,
```
benoutram, line 43: Just wondering if from an API point of view you considered implementing this to allow an array of levels rather than only allowing one value? I think it would have provided more flexibility and been a similar amount of work (although I might be wrong about that!)

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/DataSetFilesControllerTests.cs
```diff
@@ -95,6 +95,51 @@ await TestApp.AddTestData<ContentDbContext>(context =>
...
             }
 
+            [Fact]
+            public async Task FilterByGeographicLevel_Success()
+            {
+                var (publication1, publication2) = _fixture
+                    .DefaultPublication()
+                    // Publications each have a published release version
+                    .WithReleases(_fixture.DefaultRelease(publishedVersions: 1)
+                        .Generate(1))
```
benoutram, line 105: You can shorten this as 

```csharp
.WithReleases([_fixture.DefaultRelease(publishedVersions: 1)])
```

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/DataSetFilesControllerTests.cs
```diff
@@ -95,6 +95,51 @@ await TestApp.AddTestData<ContentDbContext>(context =>
...
+                    .GenerateTuple2();
+
+                var publication1Release1Version1Files = _fixture.DefaultReleaseFile()
+                    .WithReleaseVersion(publication1.ReleaseVersions[0])
+                    .WithFiles(_fixture.DefaultFile(FileType.Data)
+                    .WithDataSetFileMeta(_fixture.DefaultDataSetFileMeta()
+                        .WithGeographicLevels(
+                            [GeographicLevel.Country, GeographicLevel.LocalAuthority, GeographicLevel.Institution]))
+                    .WithDataSetFileGeographicLevels(
+                        [GeographicLevel.Country, GeographicLevel.LocalAuthority, GeographicLevel.Institution])
```
benoutram, line 116: I think this is an example of where it can get a bit confusing keeping geographic levels in `DataSetFileMeta` when it's now also in the collection of `DataSetFileGeographicLevel`. Presumably only one set of levels needs to be set up here?

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/DataSetFilesControllerTests.cs
```diff
@@ -95,6 +95,51 @@ await TestApp.AddTestData<ContentDbContext>(context =>
...
+                    .GenerateList(1))
+                    .GenerateList();
```
benoutram, line 118: This doesn't look quite right and from the test result looks like there's only one file being generated here anyway.

Rather than `var publication1Release1Version1Files` could use `ReleaseFile publication1Release1Version1File` and remove both of these lines, or give more files to publication 1 release version 1 with some different levels.

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Fixtures/DataImportGeneratorExtensions.cs
```diff
@@ -19,8 +19,15 @@ public static Generator<DataImport> WithSubjectId(
...
-        string dataFileName)
-        => generator.ForInstance(s => s.SetDefaultFiles(dataFileName));
+        string dataFileName,
+        bool metaSet = true)
+    {
+        if (metaSet)
+        {
+            return generator.ForInstance(s => s.SetDefaultFiles(dataFileName));
+        }
+        return generator.ForInstance(s => s.SetDefaultFilesWithoutMeta(dataFileName));
```
benoutram, line 29: Just wondering when would there be a case of wanting to use this with `false`?

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
```diff
@@ -110,6 +111,7 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             ConfigureReleaseStatus(modelBuilder);
             ConfigureReleaseFile(modelBuilder);
             ConfigureFile(modelBuilder);
+            ConfigureDataSetFileGeographicLevel(modelBuilder);
```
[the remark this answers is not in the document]
benoutram: Happy for you to leave this one.

I already thought along the same lines, but was just thinking that in this case, new tables could use the new approach, meaning less work for someone in future that has to move all of the configuration in one go. I imagine that will be a fairy big change anyway.

Anyone on the project touching the Public API already needs to be aware of the new approach.

src/GovUk.Education.ExploreEducationStatistics.Content.Model/Database/ContentDbContext.cs
```diff
@@ -452,13 +454,28 @@ private static void ConfigureFile(ModelBuilder modelBuilder)
...
+                entity.HasKey(gl => new
+                {
+                    gl.DataSetFileVersionId,
+                    gl.GeographicLevel
+                });
+
+                entity.Property(gl => gl.GeographicLevel)
+                    .HasConversion(new EnumToEnumValueConverter<GeographicLevel>());
+            });
+        }
```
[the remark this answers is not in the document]
mmyoungman: Discussed on Slack. `HasKey` is preventing duplicates so I will resolve this

src/GovUk.Education.ExploreEducationStatistics.Content.Model/DataSetFileGeographicLevels.cs
```diff
@@ -0,0 +1,14 @@
+﻿#nullable enable
+using System;
+using GovUk.Education.ExploreEducationStatistics.Common.Model.Data;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Model;
+
+public class DataSetFileGeographicLevel
```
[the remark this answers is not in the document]
mmyoungman: I didn't - good point. I'll add "version"

src/GovUk.Education.ExploreEducationStatistics.Content.Requests/DataSetFileListRequest.cs
```diff
@@ -1,12 +1,16 @@
...
+using GovUk.Education.ExploreEducationStatistics.Common.Utils;
+using GovUk.Education.ExploreEducationStatistics.Common.Validators;
 
 namespace GovUk.Education.ExploreEducationStatistics.Content.Requests;
 
 public record DataSetFileListRequest(
     Guid? ThemeId = null,
     Guid? PublicationId = null,
     Guid? ReleaseId = null,
+    string? GeographicLevel = null,
```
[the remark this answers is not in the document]
mmyoungman: Yeah I did try to do this with `GeographicLevel`, but discussed this with Nick and ended up on this approach with `GeographicLevelEnum`.

I'm inclined to leave it - I already spent a fair bit of time on this - but if you feel strongly about it I could have another crack at making it work

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/DataSetFilesControllerTests.cs
mmyoungman: It uses DataSetFileVersionGeographicLevel to filter in the db query, but then uses the DataSetFileMeta to populate the response.

This will get sorted as part of EES-5666 which I'll be picking up after the filter hierarchies work

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20241213123442_EES5738_CreateDataSetFileGeographicLevelsTable.cs
```diff
@@ -0,0 +1,43 @@
...
+                    table.ForeignKey(
+                        name: "FK_DataSetFileGeographicLevels_Files_DataSetFileVersionId",
+                        column: x => x.DataSetFileVersionId,
+                        principalTable: "Files",
+                        principalColumn: "Id",
+                        onDelete: ReferentialAction.Cascade);
+                });
+
+            migrationBuilder.Sql("GRANT SELECT ON dbo.DataSetFileGeographicLevels TO [content];");
+            migrationBuilder.Sql("GRANT INSERT ON dbo.DataSetFileGeographicLevels TO [importer];");
```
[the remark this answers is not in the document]
mmyoungman: I've added select permission for data and publisher users

benoutram (approved):

amyb-hiveit (dismissed):
The FE code looks fine, but it's missing:

- the button to remove the filter, it should be above the list of data sets like when you filter by theme etc
- UI test (probably quick to add to `data_catalogue.robot`)

Also a couple of queries:

- should we show levels in the dropdown that there aren't data sets for?
- there's a discrepancy in labelling for National / Country, can it be made consistent?
Screenshot 2024-12-19 125950

mmyoungman:
> should we show levels in the dropdown that there aren't data sets for?

Discussed this with Rich and we've create another ticket to do this work, EES-5746

mmyoungman:
> the button to remove the filter, it should be above the list of data sets like when you filter by theme etc

:+1:

mmyoungman:
> UI test (probably quick to add to data_catalogue.robot)

:+1:

tests/robot-tests/tests/admin_and_public/bau/data_catalogue.robot
```diff
@@ -212,6 +212,22 @@ Remove release filter
     user checks page does not contain button    ${PUPIL_ABSENCE_RELEASE_NAME}
     user checks selected option label    id:filters-form-release    All releases
 
+Filter by geographic level
+    user wait for option to be available and select it    css:select[id="filters-form-geographic-level"]
+    ...    Local Authority District
+
+    user checks page contains button    Local Authority District
+    user checks testid element contains    total-results    1 data set
```
benoutram, line 220: Just wondering how resilient this is going to be when run in parallel with other test suites where other files may be uploaded containing the same geographic level data. I think there's no filtering by theme or publication applied by this stage so it's operating on all published data data set files. Isn't there a possibility of finding more than one result, especially when run against the Dev environment?

Wouldn't it be better to be filtered by the suite's publication by this stage and check the count of files filtered by geographic level against what we expect, out of the four data set files which were uploaded at the beginning?

We could also check the results are those expected by title or at least have the expected level value displayed in  'Geographic level' in the 'Show more details' section.

mmyoungman: >  I think there's no filtering by theme or publication applied by this stage

There is still filtering for the pupils and schools theme and pupil absence publication at this point. The previous test case (Remove release filter) just removes the release filter, but leaves the theme and publication filter. So I think this is fine, as the publication is seed data and we'd not expect anyone to add data sets to it.

I will check the the Geographic level of the data set that is returned.

benoutram: > There is still filtering for the pupils and schools theme and pupil absence publication at this point. The previous test case (Remove release filter) just removes the release filter, but leaves the theme and publication filter.

Cool, I missed that 🙂 . Change looks good.

benoutram (approved):

PR #5455, opened 2024-12-12
EES-5656 Allow custom ordering of a publication's latest release when reordering releases

benoutram:
This PR builds upon the changes already made for EES-5656 in #5441, #5442 and #5444.

It allows custom ordering of a publication's latest release in the Admin's 'Release order' page when reordering releases.

With this change it's now possible to alter which the latest published release for a publication should be by ordering that release as the top release in the list after any unpublished or legacy releases.

A backend change has been made so that when reordering the releases, the publication's `LatestPublishedReleaseVersionId` value is worked out to be the latest published version of the first release in the reordered list of releases which has a published version (if any).

The remaining work in EES-5656 (to follow) will alter the publishing workflow to change `LatestPublishedReleaseVersionId` accordingly if a release version being published belongs to the latest release in the publication's release order.

Until that change has been it's possible that publishing another release version can revert an update made to the publication by the 'Release order' page so I'm marking this PR as 'Do not merge' for now.

ℹ️ ℹ️  Update 20/12/24: #5474 has now been merged into this change. ℹ️ ℹ️

### Other changes

- Changed `PublicationService` to use primary constructor.

### UI test result

release_reordering.robot:

image

PR #5474, opened 2024-12-18
EES-5656  Publishing change

benoutram:
This PR follows up the changes already made for EES-5656 in https://github.com/dfe-analytical-services/explore-education-statistics/pull/5455.

It also builds upon https://github.com/dfe-analytical-services/explore-education-statistics/pull/5426 which removed `IReleaseVersionRepository.ListLatestReleaseVersions(CancellationToken cancellationToken)` which returned release versions in reverse chronological order for User Management pages.

Now on publishing a release version, we change the `LatestPublishedReleaseVersionId` of a publication based on the publication's release series. If the release version being published is within the same release as the latest release in the series then we will replace the `LatestPublishedReleaseVersionId` with the current one being published. If the release version being published is not within the same release then it's left unchanged.

This replaces the previous behaviour which replaced the value if the release version belonged to the publication's latest release in reverse chronological order according to time period and year.

It  updates methods `IReleaseVersionRepository.GetLatestReleaseVersion` to use the latest release in the release series when determining the latest version, and `IReleaseVersionRepository.ListLatestReleaseVersions` to use the release series to order the release versions by.

### Other changes

- Remove `IReleaseVersionRepository.GetLatestPublishedRelease(Guid publicationId)` - This method which previously calculated the latest release version of a publication's latest release in reverse chronological order according to time period and year is removed in favour of using `Publication.LatestPublishedReleaseVersion` directly.

- Consolidated methods `IReleaseVersionRepository.ListLatestReleaseVersionIds/ListLatestPublishedReleaseVersionIds` and `IReleaseVersionRepository.ListLatestReleaseVersions/ListLatestPublishedReleaseVersions` with new `publishedOnly` flag.

### UI test report

image


src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Repository/ReleaseVersionRepositoryTests.cs
```diff
@@ -210,27 +118,26 @@ public class GetLatestReleaseVersionTests : ReleaseVersionRepositoryTests
         [Fact]
         public async Task Success()
         {
-            var publications = _dataFixture
+            var (publication1, publication2) = _dataFixture
```
jack-hive, line 121: wonder if to bother adding two publications? Maybe just add 1 test that checks to see that the publication Guid passed into the method we're testing grabs the right one, where we do set-up 2 publications, but then for the other tests just leave it as 1 publication to make it leaner?

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Repository/ReleaseVersionRepositoryTests.cs
```diff
@@ -22,186 +21,95 @@ public class GetLatestPublishedReleaseVersionTests : ReleaseVersionRepositoryTes
...
-            var repository = BuildRepository(contentDbContext);
-
-            Assert.Null(await repository.GetLatestPublishedReleaseVersion(Guid.NewGuid()));
-        }
+                .WithReleases(_ =>
+                [
+                    _dataFixture.DefaultRelease(publishedVersions: 1, year: 2022),
+                    _dataFixture.DefaultRelease(publishedVersions: 2, draftVersion: true, year: 2021)
+                ])
+                .GenerateTuple2();
```
jack-hive, line 31: Wonder if we should do the same thing as `GetLatestReleaseVersionTests.Success` here, and set the `ReleaseSeries` explicitly. Purposefully make `2021` go first in the series, but give it **zero** published versions. And that way we can check that the `2021` release version should not be the one fetched back.

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
```diff
@@ -175,41 +166,41 @@ public async Task Success()
...
+                    Assert.Equal(release2020.Slug, publicationViewModel.Releases[1].Slug);
+                    Assert.Equal(release2020.Title, publicationViewModel.Releases[1].Title);
 
-                    Assert.Equal(expectedReleaseVersion2.Id, publicationViewModel.Releases[1].Id);
-                    Assert.Equal(expectedReleaseVersion2.Slug, publicationViewModel.Releases[1].Slug);
-                    Assert.Equal(expectedReleaseVersion2.Title, publicationViewModel.Releases[1].Title);
+                    var releaseSeries = publicationViewModel.ReleaseSeries;
 
-                    Assert.Equal(3, publicationViewModel.ReleaseSeries.Count);
+                    Assert.Equal(3, releaseSeries.Count);
```
jack-hive, line 182: where did `2021` go? I've probably missed something

jack-hive (changes requested):
Great work! Can't spot anything too out of place. Only jotted a few things down. I've not ran the tests or checked if it builds or anything.

src/GovUk.Education.ExploreEducationStatistics.Publisher/PublisherHostBuilderExtensions.cs
```diff
@@ -123,6 +123,7 @@ public static IHostBuilder ConfigurePublisherHostBuilder(this IHostBuilder hostB
                     .AddScoped<IIndicatorRepository, IndicatorRepository>()
                     .AddScoped<IPublishingCompletionService, PublishingCompletionService>()
                     .AddScoped<IPublicationRepository, PublicationRepository>()
+                    .AddScoped<IReleaseRepository, ReleaseRepository>()
```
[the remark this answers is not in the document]
benoutram: I didn't realise `IReleaseRepository` was needed at first, but it's because it's used indirectly by `Content.Services.PublicationCacheService` -> `Content.Services.IPublicationService` -> `IReleaseRepository` which is why I added this in.

I also checked `IReleaseVersionRepository` is still needed, and that one is used by `Publisher.Services.ReleaseService` which uses method `GetPublishedDate`.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/PublicationServiceTests.cs
benoutram: The data that's setup contains a publication with 3 releases.

2020 has 1 published version
2021 has 0 published versions and 1 draft
2022 has 2 published versions and 1 draft

So slightly above this, `Assert.Equal(2, publicationViewModel.Releases.Count);` and the assertions following it check that `publicationViewModel.Release` contain `2022` and `2020`.

Then here `Assert.Equal(3, releaseSeries.Count)` checks that `publicationViewModel.ReleasesSeries` contains `2022`, `2020`, AND the legacy link which is also set up. See `.WithLegacyLinks([legacyLink])`.

`2021` is missing from both `publicationViewModel.Releases` and `publicationViewModel.ReleaseSeries` because this is a public facing method and it filters out unpublished releases.

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Repository/ReleaseVersionRepositoryTests.cs
benoutram: Great suggestion, that certainly leans things up quite a bit!

I did this everywhere so now each inner class of tests has one method beginning with the name `MultiplePublications_` which still uses the `GenerateTuple2()` to create two publications, and tests the data comes back from the specified publication.

In all other cases the tuple of publications is removed.

I also means we get rid of all the `.ForIndex(0,`, `.ForIndex(1,` etc. 😄

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Repository/ReleaseVersionRepositoryTests.cs
benoutram: It's not super obvious, but the only overload of `GetLatestPublishedReleaseVersion` which remains now is the variant with parameters `Guid publicationId, string releaseSlug`. 

There can only ever be one release in the publication with the specified slug, so the release series order is not relevant. It's still returned even if it's not the latest release for the publication, because the method is used to retrieve the latest version of a specific release. Up until the work that @rianski is doing in EES-5219 this would usually have been a non-latest release.

We also have a  test `PublicationHasNoPublishedReleaseVersions_ReturnsNull` where `2021` is given zero published versions and checks there's no result.

The overload with just  `Guid publicationId` was removed in favour of replacing all occurrences to access it directly from field `Publication.LatestPublishedReleaseVersion` instead of having a method which calculated it based on the release series order and working out the latest version for the latest release.

To make things a bit clearer I've renamed the method from `GetLatestPublishedReleaseVersion` to `GetLatestPublishedReleaseVersionByReleaseSlug` in a new commit, and renamed the inner test class too.

I also change the implementation of that method to use `SingleOrDefaultAsync` rather than `FirstOrDefaultAsync`.

I think your comment might be valid for other methods though and this has reminded me that we should be checking the order of results so I'm going to review the other tests.

src/GovUk.Education.ExploreEducationStatistics.Publisher/PublisherHostBuilderExtensions.cs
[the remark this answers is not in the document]
jack-hive: ah ok, makes sense :)

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Repository/ReleaseVersionRepositoryTests.cs
```diff
@@ -469,11 +471,11 @@ public async Task Success()
 
                 var result = await repository.ListLatestReleaseVersionIds(publication.Id, publishedOnly: PublishedOnly);
```
jack-hive, line 472: I wonder if it would be better to create a private method `ListLatestReleaseVersionIds`, that feeds a value of `false` for the `publishedOnly` parameter into the method call, rather than storing the `publishedOnly` parameter in a class constant.

so you call it like `var result = await ListLatestReleaseVersionIds(repository, publication.Id)`, and you can forget about the `publishedOnly` flag.

Similar for the other test classes

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Content.Model.Tests/Repository/ReleaseVersionRepositoryTests.cs
jack-hive: Changes look good! :)

jack-hive (approved):

PR #5488, opened 2024-12-31
EES-5756 Set DataSetFileMeta geographic levels to null

mmyoungman:
This PR sets DataSetFileMeta geographic levels to null. We do this because the same geographic levels are now cached in the new `DataSetFileVersionGeographicLevels` table.

This PR also makes changes to use the new `DataSetFileVersionGeographicLevels` table entries over `DataSetFileMeta.GeographicLevels`.

We set the geog lvls in JSON to null using a migration endpoint. There is no rush on using the endpoint after deploying. 

Unfortunately, we also needed a new temporary column in the Files table to record what entries have been migrated. This will be removed along with `DataSetFileMeta.GeographicLevels` in EES-5750.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/DataSetFileMetaMigrationController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/DataSetFileMetaMigrationController.cs
@@ -1,9 +1,7 @@
-using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using GovUk.Education.ExploreEducationStatistics.Admin.Models;
-using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Model;
 using GovUk.Education.ExploreEducationStatistics.Content.Model;
 using GovUk.Education.ExploreEducationStatistics.Content.Model.Database;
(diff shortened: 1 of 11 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/DataSetFileMetaMigrationController.cs
```diff
@@ -42,30 +41,21 @@ public async Task<MigrationResult> DataSetFileVersionGeographicLevelsMigration(
...
+                file.DataSetFileMeta = new DataSetFileMeta
                 {
-                    DataSetFileVersionId = file.Id,
-                    GeographicLevel = gl,
-                })
-                .ToList();
-
-            contentDbContext.DataSetFileVersionGeographicLevels.AddRange(
-                dataSetFileVersionGeographicLevels);
+                    GeographicLevels = null,
```
tomjonesdev, line 51: More of a general query, but what's stopping you from simply removing this column/property and the associated data now, rather than nulling and removing later? Failsafe?

tomjonesdev (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/DataSetFileMetaMigrationController.cs
mmyoungman: I guess it doesn't really matter to me. My thought is if anyone else looks at this code they understand this isn't being skipped intentionally. Had a few instances where things like this end up living in the codebase longer than I expected - like months or even years - and I wouldn't want to waste a dev's time if they see geog lvls aren't being set and starting hunting for why.

That's as strong of a justification as I can muster :)

PR #5492, opened 2025-01-02
EES-5750 Clean up after removing geographic levels from Files.DataSetFileMeta JSON

mmyoungman:
This PR takes the final steps after removing geographic levels from Files.DataSetFileMeta JSON.


mmyoungman:
Ran into a bug so we're changing approach

PR #5498, opened 2025-01-06
EES-5190 - added Recovery Services Vault to environment, and registered Public API file share with a daily backup policy

duncan-at-hiveit:
This PR:
- adds backup facilities for the Public API file share.
- adds a deploy flag to optionally skip setting up the Recovery Services Vault and policies, and registration of Public API storage with the backup policy, for quicker deploys.

# Backup process for Azure File Shares

## Recovery Services Vault

We are making use of a suite of services called `Azure Backup`, and in particular a component of the suite called `Recovery Services Vault`.  A Recovery Services Vault gives you a single place to manage backups of various Azure datasources, including in this case File Shares.  Other datasources are covered by other components from Azure Backup, for example Blob Storage and PSQL being covered by `Backup Vault` rather than Recovery Services Vault, but Backup Vault does not cover File Shares.

In this PR, we are creating a generic Recovery Services Vault that can be used to backup any EES datasources and not just Public API ones.

Our Recovery Services Vault has been set up so that:
* it is immutable but not locked - this means that by default backups cannot be manually removed, but it is not yet locked which means that it can still be changed so that we can delete backups, change various configuration aspects if we need to, delete the RSV if necessary.  Making it immutable and then locking it means that it's not possible to ever delete the vault without Azure support, so I've opted to exclude this option for now.
* geo-redundant - this is in keeping with our Azure SQL and PSQL backups which are also geo-redundant, meaning that they can be recovered into their paired region (North Europe) if necessary.
* "always on" soft-delete with a retention of 30 days - this is the preferred soft-delete option and garners a security score of "Good" (the 2nd highest security score for RSV without also introducing multi-user authorization for deleting backups).     

### The Recovery Services Vault after deploying

image

## Recovery Services Vault backup policy

Out of the box, RSV comes with a set of policies which can be manually configured.  In this case, we are creating a customised daily backup policy with Bicep, which stops RSV from creating its own OOTB daily policy, so we have full control through automation.

Our daily policy is set to take backups at `midnight UTC` every day.  It is set to retain daily backups for 30 days, to retain the `Monday midnight` backup of each week for 8 weeks, and to retain the `first week's Monday midnight` backup of each month for 12 months.  The weekly backup retaining for 8 weeks has some overlap with the daily retention backups, and likewise the monthly backup has some overlap with the weekly backups, but the spread of retained backups gives us good short- and long-term recovery options should the worst happen and we need to recover to a certain point in time.

### The DailyPolicy configuration after deploying

image

## Backup policy registration for the Public API Storage Account and File Share

The Bicep deploy registers both the Storage Account (a prerequisite for registering the File Share) and the File Share itself with the daily policy set up above. 

### Storage account registration after deploying

image

### File share registration and backup status after deploying

image


infrastructure/templates/public-api/application/shared/recoveryVault.bicep
```diff
@@ -0,0 +1,55 @@
...
+    daysToRetainDailyBackups: 30
+    weeklySnapshotRetention: {
+      targetBackupDay: 'Monday'
+      weeksToRetain: 8
+    }
+    monthlySnapshotRetention: {
+      targetBackupDay: 'Monday'
+      targetBackupWeek: 'First'
+      monthsToRetain: 12
+    }
```
benoutram, line 49: I didn't see anything in the config about incremental vs. full backups. Do you know if incremental backups are an option and did you consider them?

I was also wondering if you thought about how the size of the file share might grow and looked at the cost of storage? I think storage is cheap anyway, so hopefully should not be an issue! 🙂

benoutram, line 49: I noticed that a yearly retention schedule isn't defined here but support has been added in elsewhere for configuring one should we need it. I'm guessing that you don't think retaining a yearly backup is necessary?

Tbh, I can't think why we'd need to back that far, and it might only ever be out of curiosity that we'd need to peek into one that old anwyay, rather than to restore from.

benoutram (approved):
I looked over the changes which look good to me.

Without knowing anything about Azure Backup or Recovery Services or getting too invested in it I found it difficult to provide an in-depth review. I haven't really gone over the config in a lot of detail although I did cross check a few things with the docs.

The description of the PR helped with understanding what we are adding, thank you.

infrastructure/templates/public-api/application/shared/recoveryVault.bicep
duncan-at-hiveit: Yeah exactly that!  I don't normally add in support for EVERY ARM feature to our Bicep templates, but normally limit it to the ones that we're using or are likely to use.  In this case, although I know we're not using it now, it seemed such little effort to add support for yearly too so I just added it in as well for the future!

PR #5499, opened 2025-01-07
EES-5628 - Adding the ability to add an optional `Label` when creating a new Release (BE part)

jack-hive:
As part of [EES-5351 - Release Labelling](https://dfedigital.atlassian.net/jira/software/c/projects/EES/boards/105?selectedIssue=EES-5351), we want to give the user the ability to give any Release a `Label`. This `Label` will be used to distinguish the Release from another Release with the same year and time period.

This PR is the back-end work for the associated [EES-5626](https://dfedigital.atlassian.net/jira/software/c/projects/EES/boards/105?selectedIssue=EES-5626) ticket, whereby we are adding the functionality to add and edit this `Label` for any **NEW** Release. Adding the ability to add/edit this `Label` for existing Releases will come later in another ticket; namely, [EES-5637](https://dfedigital.atlassian.net/jira/software/c/projects/EES/boards/105?selectedIssue=EES-5637).

## Other Changes
- When writing the integration tests for these changes, I realised that we had an issue running the tests due to the operations wrapped in a **transaction** [here](https://github.com/dfe-analytical-services/explore-education-statistics/pull/5499/files#diff-eacc06f6f6c99307be5d72ea5e42f2b6a9273f63a2db6115fd069825b821f015:~:text=await%20_context.RequireTransaction,releaseVersion). I reverted my changes and wrote an integration test against the old code, and still had the same issue. So it wasn't a newly introduced issue from my changes. I span up the admin app and tested the endpoint manually using Postman, and had no issues. So it was purely a test environment specific issue. 
The issue boiled down to using the `ContentDbContext` [here](https://github.com/dfe-analytical-services/explore-education-statistics/pull/5499/files#diff-eacc06f6f6c99307be5d72ea5e42f2b6a9273f63a2db6115fd069825b821f015:~:text=await%20_context.RequireTransaction,releaseVersion) to open up a new transaction scope, but only having set up the `PublicDataDbContext` with a retry policy [here](https://github.com/dfe-analytical-services/explore-education-statistics/pull/5499/files#diff-b4b68e42d1455ecfb4629bcc07844bc6f339ec54682720940171f1981cdf29d2:~:text=psqlOptions%20%3D%3E%20psqlOptions.EnableRetryOnFailure()), and not setting up a retry policy for the `ContentDbContext`. The error was due to opening up the transaction scope with a different context to the one which had a retry policy set up. We couldn't, however, set up a retry policy for the `ContentDbContext` (which it **does** have in the production environment) due to it being an in-memory database. So, in order to fix the issue, I removed the retry policy for the `PublicDataDbContext` in the test environment. This didn't look to break any existing tests, and I don't see the need for it unless we decide to write tests for the retry policy.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/ReleasesControllerTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/ReleasesControllerTests.cs
@@ -36,7 +36,7 @@ using ValidationUtils = GovUk.Education.ExploreEducationStatistics.Common.Valida
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Controllers.Api
 {
-    public abstract class ReleasesControllerUnitTests
+    public class ReleasesControllerUnitTests
     {
         private readonly Guid _releaseVersionId = Guid.NewGuid();
         private readonly Guid _publicationId = Guid.NewGuid();
@@ -882,7 +882,9 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Controllers.Api
 
                 var validationProblem = response.AssertValidationProblem();
 
-                var error = Assert.Single(validationProblem.Errors); // Need to change the code to stop throwing an error and return a Validation Result instead
+                var error = Assert.Single(validationProblem.Errors);
+
+                Assert.Equal(UpdateRequestForPublishedReleaseInvalid.ToString(), error.Code);
             }
 
             [Fact]
@@ -908,7 +910,9 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Controllers.Api
 
                 var validationProblem = response.AssertValidationProblem();
 
-                var error = Assert.Single(validationProblem.Errors); // Need to change the code to stop throwing an error and return a Validation Result instead
+                var error = Assert.Single(validationProblem.Errors);
+
+                Assert.Equal(UpdateRequestForPublishedReleaseInvalid.ToString(), error.Code);
             }
 
             [Fact]
@@ -934,7 +938,9 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Controllers.Api
 
                 var validationProblem = response.AssertValidationProblem();
 
-                var error = Assert.Single(validationProblem.Errors); // Need to change the code to stop throwing an error and return a Validation Result instead
+                var error = Assert.Single(validationProblem.Errors);
+
+                Assert.Equal(UpdateRequestForPublishedReleaseInvalid.ToString(), error.Code);
             }
 
             [Fact]
@@ -1036,11 +1042,12 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Controllers.Api
                 string existingReleaseSlug)
             {
                 Publication publication = DataFixture.DefaultPublication()
-                    .WithReleases([
+                    .WithReleases(
                         DataFixture
                             .DefaultRelease(publishedVersions: 0, draftVersion: true)
                             .WithSlug(existingReleaseSlug)
-                        ]);
+                            .GenerateList(2)
+                        );
 
                 await TestApp.AddTestData<ContentDbContext>(
                     context => context.Publications.Add(publication));
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/ReleasesControllerTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/ReleasesControllerTests.cs
@@ -713,7 +713,27 @@ namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Controllers.Api
                 Assert.Equal(SlugNotUnique.ToString(), error.Code);
             }
 
-            // Add test for character limit on Label of 50
+            [Fact]
+            public async Task LabelOver50Characters()
+            {
+                Publication publication = DataFixture.DefaultPublication();
+
+                await TestApp.AddTestData<ContentDbContext>(
+                    context => context.Publications.Add(publication));
+
+                var response = await CreateRelease(
+                    publicationId: publication.Id,
+                    year: 2020,
+                    timePeriodCoverage: TimeIdentifier.AcademicYear,
+                    label: new string('a', 51));
+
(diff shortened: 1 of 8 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/ReleasesControllerTests.cs
```diff
@@ -22,10 +32,11 @@
 using static GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils.MockUtils;
 using static Moq.MockBehavior;
 using ErrorViewModel = GovUk.Education.ExploreEducationStatistics.Common.ViewModels.ErrorViewModel;
+using ValidationUtils = GovUk.Education.ExploreEducationStatistics.Common.Validators.ValidationUtils;
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Controllers.Api
 {
-    public class ReleasesControllerTests
+    public class ReleasesControllerUnitTests
```
benoutram, line 39: Good call on renaming this to separate out integration tests while retaining the existing unit tests.

Looking at the relatively simple implementation of the controller methods I suspect the unit tests aren't doing much more than an exercise of matching up the params in the mock service behaviour with expectations, and that we could get rid of them, but it would be time consuming to go through each one and determine if it's worth keeping or not.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/ReleasesControllerTests.cs
```diff
@@ -543,4 +554,591 @@ private static ReleasesController BuildController(
...
+            {
+                client ??= BuildApp().CreateClient();
+
+                var request = new
+                {
+                    Type = type,
+                    Year = year,
+                    TimePeriodCoverage = new { Value = timePeriodCoverage.GetEnumValue() },
+                    Label = label,
+                };
```
benoutram, line 760: Just wondering why you don't create this as a new `ReleaseCreateRequest`?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/ReleasesControllerTests.cs
```diff
@@ -543,4 +554,591 @@ private static ReleasesController BuildController(
...
+                    year: 2020,
+                    timePeriodCoverage: TimeIdentifier.AcademicYear,
+                    label: new string('a', 51));
+
+                var validationProblem = response.AssertValidationProblem();
+
+                var error = Assert.Single(validationProblem.Errors);
+
+                Assert.Equal($"The field {nameof(ReleaseCreateRequest.Label)} must be a string or array type with a maximum length of '50'.", error.Message);
+                Assert.Equal(nameof(ReleaseCreateRequest.Label), error.Path);
```
benoutram, line 735: I think you might be able to do something like this here:

```csharp
validationProblem.AssertHasMaximumLengthError(nameof(ReleaseCreateRequest.Label).ToLowerFirst(),maxLength: 50);
```

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseService.cs
```diff
@@ -838,26 +807,62 @@ private async Task<Either<ActionResult, ReleaseFile>> CheckReleaseDataFileExists
...
+
+            return slugAlreadyExists 
+                ? ValidationActionResult(SlugNotUnique) 
+                : Unit.Instance;
+        }
 
-            if (releaseVersions.Any(release =>
-                    release.Slug == slug &&
-                    release.Id != releaseVersionId &&
-                    IsLatestVersionOfRelease(releaseVersions, release.Id)))
```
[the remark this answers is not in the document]
jack-hive: yeah, bit tidier :D

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/ReleasesControllerTests.cs
jack-hive: I'll take a quick look to see if any can be removed

jack-hive: just looked, and although they're **nearly** useless, they're not completely useless. They still technically have some value in checking that the right thing gets returned from the controller. So I think maybe we should leave them in for now?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/ReleasesControllerTests.cs
jack-hive: I did initially, but it turns out we have a `[JsonConverter(typeof(TimeIdentifierJsonConverter))]` on the `TimePeriodCoverage` field on the `ReleaseCreateRequest`. And it essentially requires the object to look like:
```json
{
  ...
  "timePeriodCoverage": {
    "value": "AYQ1"
  }
  ...
}
```

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/ReleasesControllerTests.cs
jack-hive: I don't think I can in this case unfortunately... I think it's because we're using data annotations for the max length check, rather than fluent validation

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/ReleasesControllerTests.cs
benoutram: Ok good to know, thanks!

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/ReleasesControllerTests.cs
benoutram: Ah right, that makes sense. I think when I saw `ReleaseCreateRequestValidator.Validate` in the service I automatically assumed that was fluent validation, but now I can see it's not. Thanks!

benoutram (approved):

PR #5501, opened 2025-01-07
EES-5687 - restructuring staticMetricAlert.bicep to be in line with newly restructured dynamicMetricAlert.bicep

duncan-at-hiveit:
This PR:
- restructures `staticMetricAlert.bicep` to be in line with recent changes to `dynamicMetricAlert.bicep`, including using better type checking for resourceType / metric combinations and ability to reuse common alert criteria for consistency and less boilerplate.
- removes now-defunct `ResourceType` and `MetricName` types from `alerts/types.bicep`.

## Also included in this PR

* Removal of `alerts/flexibleServers/databaseAlive.bicep` file that was accidentally left in.
* Upping threshold of the App Gateway's "UnhealthyHostCount" alert to "0.05" in order to reduce noise in the #alerts-dev Slack channel.  I have raised https://dfedigital.atlassian.net/browse/EES-5776 to investigate the cause of all these alerts further on down the line.

## Follow-up work not included in this PR

As part of this PR, I was also going to split `alerts/config.bicep` into `staticAlertConfig.bicep` and `dynamicAlertConfig.bicep`, and move the `StaticAlertConfig` and `DynamicAlertConfig` types out of `types.bicep` into these new files respectively, but that would have made this PR much bigger!  Happy to immediately follow this one up with that change though.

Also not included in this PR is removing any of the resource-specific alert Bicep templates and replacing with direct calls to `staticMetricAlert.bicep` from the component files directly as again that would cause much bigger changes.  Jury is also out as to whether or not this is the best path to take, as I'm personally on the fence.


infrastructure/templates/public-api/components/alerts/fileServices/availabilityAlert.bicep
```diff
@@ -1,32 +1,27 @@
...
-      aggregation: 'Average'
-      operator: 'LessThan'
-      threshold: 100
     }
-    evaluationFrequency: 'PT1M'
-    windowSize: 'PT5M'
-    severity: severity
+    config: {
+      ...staticAverageLessThanHundred
+      nameSuffix: 'fileservice-availability'
```
ntsim, line 23: Nitpick - should this be `file-service-availability`?

ntsim (commented):
Looks good!

Would agree with moving the configs into their own respective files and removing the resource-specific alerts.

Like the dynamic alerts, doesn't seem like there's much benefit to keeping the resource-specific alerts? Removing them would be more consistent with what we've done for dynamic alerts so feels like it'd be more confusing to leave these in.

ntsim (approved):

infrastructure/templates/public-api/components/alerts/fileServices/availabilityAlert.bicep
duncan-at-hiveit: So long as it doesn't take us over the frustratingly limited resource name length limits, I'll update this thanks :joy:

PR #5512, opened 2025-01-09
EES-5458 fix table headers multi select

amyb-hiveit:
Fixes the styling when reordering table headers to make it obvious which items you've selected when moving multiple items.

Also fixes a bug where moving multiple items was sometimes moving them to the wrong place.

Screenshot 2025-01-09 125424


bennettstuart (approved):
LGTM!

PR #5516, opened 2025-01-09
EES-5707 Remove geojson cache file after updating data block with map

mmyoungman:
This PR removes associated geo json cache files (controlled by `LocationsForDataBlockCacheKey`) after a data block is updated.

This is to prevent a bug where if someone caches geo json for a data set that has two locations with the same name, and then update the data block to include both locations of the same name, the user won't be able to look at the data block chart without getting a Sorry page. This would continue until the cache file had been removed. See EES-5704 for details.

This hasn't solved the issue entirely - if you update a data block the Content page may still lead to a Sorry page, but refreshing the page will fix this. This has been deemed acceptable for now.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using AutoMapper;
+using GovUk.Education.ExploreEducationStatistics.Admin.Cache;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.Cache;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.Security;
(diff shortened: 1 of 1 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
```diff
@@ -229,6 +230,14 @@ await _releaseFileService.Delete(releaseVersionId: dataBlockVersion.ReleaseVersi
                             fileId: new Guid(infographicChart.FileId));
                     }
 
+                    // If has map chart, remove geojson cache
+                    var mapChart = dataBlockVersion.Charts.OfType<MapChart>().FirstOrDefault();
+                    if (mapChart != null)
+                    {
+                        await _cacheService.DeleteItemAsync(
+                            new LocationsForDataBlockCacheKey(dataBlockVersion, mapChart.BoundaryLevel));
```
benoutram, line 238: Just wondering if there are any other places we should be invalidating the geojson cache, like if the data is replaced?

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataBlockService.cs
mmyoungman: Discussed this with Ben and Cam on Slack. We think the geojson cache isn't being invalidated for replacements, but we think that probably isn't an issue. We decided to not do anything about replacements immediately and highlight this case for testing.

PR #5520, opened 2025-01-10
EES-5787 Allow request Chart.Title to be null

mmyoungman:
This fixes a bug introduced in EES-5731. A chart couldn't be created because the backend validation didn't allow a null chart title.

When creating a chart, the chart title is only provided in the request if the user is providing an alternative chart title. Otherwise, the chart title is generated by the backend (I believe).

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/DataBlockRequests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/DataBlockRequests.cs
@@ -70,8 +70,7 @@ public record DataBlockUpdateRequest
             {
                 chart
                     .RuleFor(request => request.Title)
-                    .NotEmpty()
-                    .MaximumLength(220);
+                    .Length(1, 220);
 
                 chart.RuleFor(request => request.Alt)
                     .NotEmpty()


src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/DataBlockRequests.cs
```diff
@@ -70,7 +70,7 @@ public Validator()
             {
                 chart
                     .RuleFor(request => request.Title)
-                    .NotEmpty()
+                    .NotEmpty().When(request => request.Title != null)
```
duncan-at-hiveit, line 73: Hey @mmyoungman - could this not just be:

```
                chart
                    .RuleFor(request => request.Title)
                    .MaximumLength(220);
```

mmyoungman: I think this would allow title to be an empty string?

I've switched to 
```
  chart
                    .RuleFor(request => request.Title)
                    .Length(1, 220);
```

duncan-at-hiveit (approved):

N-moh (approved):

PR #5522, opened 2025-01-13
EES-5784 Switch BE view-models to use Release.Title instead of ReleaseVersion.Title

benoutram:
This PR switches all back end view-models to use `Release.Title` instead of `ReleaseVersion.Title` and removes the calculated property `ReleaseVersion.Title`. This has been done in conjunction with #5519  and #5499 which have added the optional ability to add a label to `Release`.

### UI Test result

image


src/GovUk.Education.ExploreEducationStatistics.Content.Services/ReleaseService.cs
```diff
@@ -76,6 +76,7 @@ public async Task<Either<ActionResult, ReleaseCacheViewModel>> GetRelease(
 
             var releaseVersion = _contentDbContext
                 .ReleaseVersions
+                .Include(rv => rv.Release)
```
jack-hive, line 79: Almost thought this wasn't doing anything, but then saw your auto-mapper config and realised all is good! :D

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/ManageContentPageService.cs
```diff
@@ -159,11 +159,11 @@ private IQueryable<ReleaseVersion> HydrateReleaseQuery(IQueryable<ReleaseVersion
             // performant as running individual queries, and revert this if required.
             return queryable
                 .AsSplitQuery()
-                .Include(rv => rv.Publication)
+                .Include(rv => rv.Release)
+                .ThenInclude(release => release.Publication)
```
jack-hive, line 163: should we just abbreviate to `r => r.Publication` here, and same for some others below?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseService.cs
```diff
@@ -578,20 +578,25 @@ await _releaseCacheService.UpdateRelease(
 
         public async Task<Either<ActionResult, IdTitleViewModel>> GetLatestPublishedRelease(Guid publicationId)
         {
-            return await _persistenceHelper
-                .CheckEntityExists<Publication>(publicationId, queryable =>
-                    queryable.Include(rv => rv.LatestPublishedReleaseVersion))
+            return await _context.Publications
+                .SingleOrNotFoundAsync(p => p.Id == publicationId)
```
jack-hive, line 582: Should we `.Include(p => p.LatestPublishedReleaseVersion)` here, just to save doing a 2nd round-trip on line `591`?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PreReleaseUserServiceTests.cs
jack-hive, line 1: I think the title of this commit is accidentally wrong. Looks like the change to `SendPreReleaseInviteEmail` was done in another commit

jack-hive (changes requested):

jack-hive (commented):
Nice work!

Couldn't spot very much, only a few comments :)

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/EmailTemplateService.cs
```diff
@@ -101,20 +101,18 @@ public Either<ActionResult, Unit> SendReleaseRoleEmail(
             return emailService.SendEmail(email, template, emailValues);
         }
 
-        public Either<ActionResult, Unit> SendReleaseHigherReviewEmail(string email,
+        public Either<ActionResult, Unit> SendReleaseHigherReviewEmail(
```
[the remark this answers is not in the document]
benoutram: Great spot!

Whilst fixing this I noticed that there's still quite a few other places using `.Include(rv => rv.Publication)` which could be a good check to do again when we switch over all of the other fields and changed the navigation to go via `Release`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ManageContent/ManageContentPageService.cs
benoutram: Yes that makes sense 👍.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/UserManagementService.cs
```diff
@@ -326,19 +326,21 @@ await _userPublicationInviteRepository.CreateManyIfNotExists(
...
-                .OnSuccess(userInvite =>
+                .OnSuccess(async userInvite =>
                 {
-                    var userReleaseInvites = _contentDbContext
+                    var userReleaseInvites = await _contentDbContext
                         .UserReleaseInvites
-                        .Include(invite => invite.ReleaseVersion.Publication)
-                        .Where(invite => invite.Email.ToLower() == sanitisedEmail)
-                        .ToList();
+                        .Include(uri => uri.ReleaseVersion)
```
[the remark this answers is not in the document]
benoutram: Yes I agree it would be better now especially as https://github.com/dfe-analytical-services/explore-education-statistics/pull/5426 updated user management for release invites to work with release id's. Internally it still has to work out the latest release version for a release because it was going to be a big change to migrate `UserReleaseInvite` to swap the navigation property from `ReleaseVersion` to `Release`.

Like you say, I think it's out of scope for this work. Cam has suggested that release invites could be removed in favour of using publication invites which would simplify things quite a bit. They'd still need to be kept around for Pre-release access I think.

Maybe both changes could be done at the same time.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseService.cs
benoutram: Ah yeah, it was like this originally before my changes but I added the second query because I wasn't sure if we could do this in the first query:

```
.Include(p => p.LatestPublishedReleaseVersion)
.ThenInclude(rv => rv.Release)
```

There's a compiler warning that `rv` could be null because `p.LatestPublishedReleaseVersion` is optional.

I researched the compiler warning for `ThenInclude` with the optional and there's an open issue about it. The warning is safe to ignore so I've gone with your suggestion now with the addition of `ThenInclude`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/EmailTemplateService.cs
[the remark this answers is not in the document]
jack-hive: Yeah good point! I'll try to store that in my brain and remember to think about it when/if I get onto making similar changes

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseVersionRepository.cs
```diff
@@ -37,39 +37,40 @@ public async Task<List<ReleaseVersion>> ListReleases(
...
+            Guid userId,
             params ReleaseApprovalStatus[] releaseApprovalStatuses)
         {
-            var userReleaseVersionIds = await _contentDbContext
+            var releaseRoleReleaseVersionIds = await _contentDbContext
                 .UserReleaseRoles
-                .AsQueryable()
-                .Where(r => r.UserId == userId && r.Role != ReleaseRole.PrereleaseViewer)
+                .Where(urr => urr.Role != ReleaseRole.PrereleaseViewer)
+                .Where(urr => urr.UserId == userId)
```
jack-hive, line 47: Wonder if we should switch the order of these to make for a more efficient query? I think `.Where(urr => urr.UserId == userId)` will filter the results down to the fewer results, before then applying `.Where(urr => urr.Role != ReleaseRole.PrereleaseViewer)`?

I may be wrong though! And it may be a negligible effect.

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseVersionRepository.cs
benoutram: Good point. I'm not sure how this would be evaluated by the database when they are applied together.

I'd like to think that when the database builds the query execution plan it will have the order optimised to only fetch the minimal amount of data, and that the order won't matter here. Not 100% sure though!

I've swapped the order now anyway.

jack-hive: I was told years ago that it isn't that smart, and will build the query in the order you wrote it. But that may have all changed now!

PR #5524, opened 2025-01-15
Arrange the dotnet projects into solution folders.

leeoades:
All of the dotnet projects existed at the root of the solution. In order to aid understanding of which parts belong to which parts of the system, I've arranged the projects into solution folders. This is simply a logical view and does not affect the location of the files on the disk.

I believe it is worth considering dividing up the codebase, if not into separate repos, perhaps separate solution files - one per service or component.


src/GovUk.Education.ExploreEducationStatistics.sln
```diff
@@ -94,6 +94,40 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "GovUk.Education.ExploreEduc
...
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Utils", "Utils", "{5FD84034-5CD0-4F43-9B6D-6BEFF66864FF}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "PublicAPI", "PublicAPI", "{A463EAD4-4E6F-4E26-B519-2EA645E60E66}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "API", "API", "{0F77425C-2F87-481E-94EB-4FA2B47EF53D}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "ContentServices", "ContentServices", "{B980CD0D-48D4-485F-933B-0681D81E408E}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "PublicDataProcessor", "PublicDataProcessor", "{82AD39EC-E1A5-44A0-9280-EB6F7ABB0112}"
```
benoutram, line 129: I started off by thinking some top level groups for each service or component such as

```
Admin
Content
Data
Notifier
Processor
Publisher
PublicApi
Shared
```

to group the projects might work better and avoid having to pigeonhole them into one of Admin, Public, Shared, or Global, but I think this approach is fine!

benoutram (approved):

PR #5529, opened 2025-01-15
EES-5678: RebuildIndexes procedure calls update statistics regardless of timeout

tomjonesdev:
This PR adds a migration which updates the behaviour of the `RebuildIndexes` procedure on the statistics database. The changes consist of:

1. extracting the logic which updates statistics into a separate procedure. This streamlines the main procedure, and avoids duplicate code for when the timeout condition _isn't_ triggered
2. dropping the `RebuildIndexesUpdateStatistics` table, as this was no longer deemed useful
3. removing messaging when a re-index action is skipped, as this was deemed too verbose
4. ensuring that the `RebuildIndexes` procedure updates the statistics regardless of timeout.


src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_UpdateStatistics.sql
```diff
@@ -0,0 +1,40 @@
+CREATE OR ALTER PROCEDURE UpdateStatistics
+	@RunId INT
+AS
+BEGIN
+	IF OBJECT_ID('tempdb..#TableList') IS NULL
+    THROW 51000, 'UpdateStatistics expects a temporary table named #TableList to already exist.', 1;
```
benoutram, line 6: This looks like bad practice to me, to depend on a temp table created by another stored procedure.

It's also got another dependency below on `__Log_RebuildIndexesAlterIndexes` so any changes to that in `RebuildIndexes` will break `UpdateStatistics`.

I would have a table value parameter of the tables to update statistics for, iterate through them calling `UPDATE STATISTICS` for each one, and then return.

In the calling SP you could create a variable like `@ModifiedTablesList` and select into it the distinct list of table names which have been modified.

You would then call this SP with something like `EXEC UpdateStatistics @ModifiedTablesList`.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/Current_Routine_RebuildIndexes.sql
```diff
@@ -27,7 +33,8 @@ BEGIN
         @ActionRequired NVARCHAR(MAX),
 
         -- Used when iterating over tables
-        @CurrentTable NVARCHAR(MAX) = ''
+        @CurrentTable NVARCHAR(MAX) = '',
+        @ModifiedTables AS ModifiedTablesType;
```
benoutram, line 37: I don't think you need `AS` here when declaring variables?

tomjonesdev: Normally I don't think so, but the example implied it may be expected for table types.

image
[Source](https://learn.microsoft.com/en-us/sql/relational-databases/tables/use-table-valued-parameters-database-engine?view=sql-server-ver16#Example)

src/GovUk.Education.ExploreEducationStatistics.Data.Api/appsettings.Development.json
```diff
@@ -3,7 +3,7 @@
   "PublicStorage": "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://data-storage:10000/devstoreaccount1;TableEndpoint=http://data-storage:10002/devstoreaccount1;",
   "ConnectionStrings": {
     "ContentDb": "Server=db;Database=content;User=data;Password=Your_Password123;TrustServerCertificate=True;",
-    "StatisticsDb": "Server=db;Database=statistics;User=data;Password=Your_Password123;TrustServerCertificate=True;"
+    "StatisticsDb": "Server=db;Database=statistics;User=sa;Password=Your_Password123;TrustServerCertificate=True;"
```
benoutram, line 6: Is this change intentional? I don't think we want to do this otherwise we won't be aware if the Data API uses resources (tables etc) that it hasn't been granted with the user role that it uses in Azure environments. This has just saved me from that scenario today on a different ticket.

I think what you might need is something like `GRANT EXECUTE ON TYPE::ModifiedTablesType TO [data];` added to the migration, so that Azure environments using the `data` role are are able to use it.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20250108173703_EES5678_RebuildIndexesTimeoutDoesNotSkipTableStatsUpdate.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        private const string PreviousRebuildIndexesMigrationId = EES5368_UpdateRebuildIndexesSPToWriteRunDataToDB.MigrationId;
+        internal const string MigrationId = "20250108173703";
+
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql("DROP TABLE IF EXISTS [dbo].[__Log_RebuildIndexesUpdateStatistics]");
+
+            migrationBuilder.Sql("CREATE TYPE ModifiedTablesType AS TABLE (Id INT IDENTITY(1,1), TableName VARCHAR(100));");
```
benoutram, line 19: I think we've got all of table types following the same `Current_TableType_x.sql` pattern to track changes to them. 

image


Historically we had a lot more, like I can see in https://github.com/dfe-analytical-services/explore-education-statistics/commit/edc998e63016cac75eca8f8c90eb92af15247a52 we had loads.

I think if you were to do the same then we are less likely to forget about this one existing if we come to squash all of the migrations.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/appsettings.Development.json
tomjonesdev: Didn't mean to commit that, just changed it for testing so I'll revert.

In regards to the TYPE permission, the issue which needed me to use SA was due to creating a table (migrations). I didn't encounter a problem with a type permission, but that may still happen but have been masked by the fact I was using `sa`, so I'll add the `GRANT` as well.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20250108173703_EES5678_RebuildIndexesTimeoutDoesNotSkipTableStatsUpdate.cs
tomjonesdev: I'll add and reference a version prefixed with the migration ID to match what we've done with the routines.

src/GovUk.Education.ExploreEducationStatistics.Data.Api/appsettings.Development.json
tomjonesdev: As discussed, `GRANT EXECUTE ON TYPE` may not be required here so I'll exclude.

benoutram (approved):

benoutram (approved):

PR #5536, opened 2025-01-20
EES-5754 add data symbols modal and table

rianski:
Changes made:
Added definitions for common data symbols into the table tool

**data symbols link:**

data symbols 1

**data symbols modal:**

data symbols 2


cjrace:
Not a review as assuming it'll want another dev to check it over, though a quick comment on the content of the modal @rianski - could we change the table so that it matches the simplified version in the ticket, and has the line of text above? I think you've taken the full version from the guidance website, but that's got information that end users don't really need to see. This is the simplified one

image

src/explore-education-statistics-common/src/components/DataSymbolsModal.tsx
```diff
@@ -0,0 +1,21 @@
...
+import ButtonText from './ButtonText';
+
+export default function DataSymbolsModal() {
+  return (
+    <Modal
+      showClose
+      title="Data symbols"
+      triggerButton={
+        <ButtonText>
+          Data symbols <InfoIcon />
```
mmyoungman, line 14: Not sure this is right, but does `InfoIcon` need a `description` attribute for accessibility?

mmyoungman (approved):

PR #5538, opened 2025-01-20
EES-4735 Allow analysts to assign default/auto select filter items for data sets

mmyoungman:
This PR allows analysts to choose a custom default / auto select filter item for filters. This means that, in the table tool, if a user doesn't select any filter items for a particular filter, the default / auto select item will automatically be selected when they try to create a table.

:rotating_light: This PR will require a faffy deploy to set Filter.AutoSelectFilerItemId for existing data sets with a "Total" item. :rotating_light: 

Before this work, the functionality was such that filter items labelled "Total" inside a "Total" filter group would be the auto select filter item. After this work, the previous behaviour mostly remains unchanged, but analysts can choose their own auto select item for a filter by adding a `filter_default` column to the meta file by entering the item label they want for the item against the appropriate filter. 

We have relaxed the need for a "Total" group for an item - instead we check to ensure there is only one Total item for the filter. If there is more than one item with a "Total" label, no auto select item is set. Matching previous behaviour, there is no need to put "Total" as the "filter_default" in the meta file to make it the auto select item: that happens automatically when a data set is imported.

Example meta file:
image

Example after clicking "Create table" if the user hasn't selected any filter items:
image

All the logic for this happens during the import of a data set to set of the `Filter.AutoSelectFilterItemId` column. Ideally, we'd have only added a `AutoSelectFilterItemId` column, but because `Filter`s are created at a different time to the creation of `FilterItem`s during import, when we create the Filters we don't have access to `FilterItem` Ids as they don't exist yet. For this reason we store the label in `Filter.AutoSelectFilterLabel` for later use, allowing us to set `Filter.AutoSelectFilterItemId` when the `FilterItem`s are created.

Because `Filter.AutoSelectFilterItemId` is the only thing that determines whether a filter auto selects an item, we need to set the `AutoSelectFilterItemId` for preexisting data sets with Total items. A migration endpoint has been created to do this.

### Other things of note:
- We originally called the auto select item "DefaultFilterItem", but this clashed with existing variables in two places, so we instead decided to switch to using "AutoSelectFilterItem" internally in the codebase, even though the meta file column is `filter_default`. I also renamed one of the other preexisting "DefaultFilterItem/Group" labels to "NotSpecifiedFilterItem/Group" in an attempt to avoid any future confusion.
- The migration sets all existing filter item "total"s to have the casing "Total". This is purely for neatness - our Stats DB is case agnostic when searching for specific text in a column.
- There is basically no frontend work done here as it already took a item id to determine which item to auto select. I just changed some variable names to use the new auto select term for consistency.
- I've added a test data set which is used in backend tests and in the `small-files` directory for general use.

UPDATE:
I initially missed that as part of this PR, we also had `isAggregate` which is used by the Public API. There was some discussion about whether we should keep this, but for now I have and have implemented similar `isAutoSelect` functionality as I've implemented in the Processor/Content projects.

Before this work, there were differences in behaviour between the Public API and elsewhere - we believe duplicate filter item labels are possible in the Stats DB, but I don't believe this is possible in the Public API. We also had validation to ensure that "Total" items in the Stats DB belonged to a "Total" filter group, while in the Public API any "Total" filter option was considered an aggregate.

I've now removed the "IsAggregate" bool column from the PublicData DB's `FilterOptionMetas` table and added a `AutoSelectLabel` column to `FilterMetas`, and we use that to work out the `IsAutoSelect` (previously `IsAggregate`) bool returned in view models.

Determining whether preexisting data sets have a "Total" is much easier in the PublicData DB, so FilterMeta.AutoSelectLabel is set if appropriate via an SQL query in the migration.


src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/AutoSelectFilterItemMigrationController.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        foreach (var filterId in filterIds)
+        {
+            var autoSelectFilterItemId = statisticsDbContext.FilterItem
+                .Where(fi => fi.FilterGroup.FilterId == filterId
+                             && fi.Label == "Total")
+                .Select(fi => fi.Id)
+                .SingleOrDefault();
+
+            if (autoSelectFilterItemId == default)
```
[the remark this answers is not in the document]
mmyoungman: Originally I had done a null check here and then during a quick test found it wasn't working because it wasn't returning null.

I don't think I need to check against null because I don't believe `SingleOrDefault` will ever return null here, since we're returning a `Guid`.

I don't agree about switching to `Guid.Empty` - the flow from `SingleOrDefault` into checking for the `default` is easy to follow to me. I could add a comment saying something like `// or == Guid.Empty`?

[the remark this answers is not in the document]
mmyoungman: Oh it's the migration controller - so maybe it's not worth the bother since all this code will get deleted anyway.

Kind of interested in whether using `Single` rather than `SingleOrDefault` would lead to this blowing up or if it would return null as is stored in the DB. (Edit: actually thinking about it this would blow up because it's getting the filter item id which cannot be null)

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20250117112425_EES4735_AddAutoSelectFilterItemColToFiltersTable.cs
```diff
@@ -0,0 +1,66 @@
...
+            migrationBuilder.AddColumn<Guid>(
+                name: "AutoSelectFilterItemId",
+                table: "Filter",
+                type: "uniqueidentifier",
+                nullable: true);
+
+            migrationBuilder.AddColumn<string>(
+                name: "AutoSelectFilterItemLabel",
+                table: "Filter",
+                type: "nvarchar(max)",
```
tomjonesdev, line 23: Not sure what volume of records are like on this table, but I tend to encourage specifying sensible maximums, assuming one can be established/agreed upon. IIRC there are performance and index-related benefits to specifying a limit as opposed to `max`, but they may well be negligible in this scenario.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/AutoSelectFilterItemMigrationController.cs
```diff
@@ -0,0 +1,72 @@
...
+                continue;
+
+            await statisticsDbContext.Filter
+                .Where(f => f.Id == filterId)
+                .ExecuteUpdateAsync(setters => setters
```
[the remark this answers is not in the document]
mmyoungman: Think they're a fairly new addition - saves having to pull objects into memory for an update

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20250117112425_EES4735_AddAutoSelectFilterItemColToFiltersTable.cs
```diff
@@ -0,0 +1,66 @@
...
+                column: "AutoSelectFilterItemId",
+                principalTable: "FilterItem",
+                principalColumn: "Id");
+
+            // Fix casing all all preexisting totals to "Total" - WHERE is case agonostic (i.e. matches any casing of "total")
+            migrationBuilder.Sql("""
+                                 UPDATE FilterItem
+                                 SET Label = 'Total'
+                                 WHERE Label = 'total';
+                                 """);
```
tomjonesdev, line 43: Do users see inconsistencies of "total"/"Total" on the frontend, or is the casing adjusted in some way before being displayed? i.e. is the tidy-up purely for the benefit of someone looking at the raw data, or does it have a user benefit?

Is there any casing enforcement/manipulation for values in the new CSV column?

src/GovUk.Education.ExploreEducationStatistics.Data.Processor.Tests/Functions/ProcessorStage2Tests.cs
```diff
@@ -316,6 +321,37 @@ private async Task AssertStage2ItemsImportedCorrectly(
...
+
+                var autoSelectFilterValue = autoSelectFilterValues.SingleOrDefault(autoSelect =>
+                    autoSelect.FilterLabel == filter.Label);
+
+                if (autoSelectFilterValue == null)
+                {
+                    Assert.Null(filter.AutoSelectFilterItemId);
+                    Assert.Null(filter.AutoSelectFilterItemLabel);
+
+                    // There should either be no Total items, or more one than (in different groups).
```
tomjonesdev, line 339: Nitpick: typo? (`more one than` > `more than one`)

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
```diff
@@ -162,14 +162,52 @@ await _dataImportService.UpdateStatus(dataImport.Id,
...
+
+            var filterIdToAutoSelectFilterItem = filterIds
+                .ToDictionary(
+                    filterId => filterId,
+                    filterId =>
+                    {
+                        var autoSelectFilterItemLabel = context.Filter
+                            .Where(f => f.Id == filterId)
+                            .Select(f => f.AutoSelectFilterItemLabel)
+                            .SingleOrDefault() ?? "Total"; // If meta file didn't specify a default, look for "Total"
```
tomjonesdev, line 182: I believe the overload for `...OrDefault` methods lets you specify the default as the argument (unless I've misunderstood what's going on here).

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/__data__/tableHeadersConfig.data.ts
```diff
@@ -47,14 +47,14 @@ export const testCategoryFilters: CategoryFilter[] = [
     value: 'category-1-value',
     label: 'Category 1',
     group: 'Default',
-    isTotal: false,
+    isAutoSelect: false,
```
tomjonesdev, line 50: Not crucial, but the acceptance criteria references names of `IsAggregate` / `IsDefault` which arguably may be more descriptive/meaningful to devs less familiar with this area of the site/codebase.

src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/AutoSelectFilterItemMigrationController.cs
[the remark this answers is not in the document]
tomjonesdev: The suggestion for `Guid.Empty` was because, at a glance, you can't tell if the ID is a `string`, `int`/`long` or `Guid` without looking elsewhere for that information - you can easily tell in your IDE by hovering, but that's not always an option (e.g. a reviewer on Github 😉).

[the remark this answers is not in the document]
mmyoungman: Ah makes sense. To be honest, I just assume that Id is always a Guid, which I think is basically true in this codebase, but probably best not to assume these kinds of thing.

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20250117112425_EES4735_AddAutoSelectFilterItemColToFiltersTable.cs
mmyoungman: Public users will see the inconsistencies on the frontend. This fixe is mostly for that reason. To be honest, I could have probably skipped this but I wanted the casing for the label and entries in `AutoSelectFilterItemLabel` to be consistent, which they wouldn't be if I didn't do this. And there was at least one instance of "total" on prod so that gets cleaned up on the public site too.

> Is there any casing enforcement/manipulation for values submitted in the new CSV column? Should there be?

There isn't, but I'm inclined to think this is ok. 

Analysts should run their data sets through the screener (created by Cam and others I believe) before they import them via admin, and that does the bulk of the validation. There is a ticket to run the screener as part of our import process.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
mmyoungman: DIdn't know you could do that - you learn something new every day :)

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/__data__/tableHeadersConfig.data.ts
mmyoungman: I discussed the change to "auto select" with Cam and he was happy with the change.

But there is a requirement to change `isAggregate` in Public API code, so I'll change that to `isAutoSelect`

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
mmyoungman: Turns out LINQ cannot turn this into SQL, so I've switched back:
```
.SingleOrDefault(__p_1)' could not be translated. Either rewrite the query in a form that can be translated, or switch to client evaluation explicitly by inserting a call to 'AsEnumerable', 'AsAsyncEnumerable', 'ToList', or 'ToListAsync'.```

src/GovUk.Education.ExploreEducationStatistics.Data.Model/Migrations/20250117112425_EES4735_AddAutoSelectFilterItemColToFiltersTable.cs
mmyoungman: We discussed this with the other devs in a dev meeting. A decision was made that new columns should set a limit but existing columns without a limit should be left as they are for now.

That said, even though this is a new column, to put a limit on this column we really should also put a limit on Filter.Label (as they're essentially the same). If we did enforce a limit on it, we could get weird behaviour where a data set import partially fails due to an AutoSelectFilterLabel being too long, even though the FilterItem.Label is considered fine. 

So, if I have Tom's blessing, I'm going to leave this.

src/explore-education-statistics-common/src/modules/table-tool/components/__tests__/__data__/tableHeadersConfig.data.ts
mmyoungman: So after looking at this I realised that I needed to update the Public API with the same functionality - more detail the updated PR description.

tomjonesdev (approved):

PR #5549, opened 2025-01-23
EES-5768: Remove time period AcademicYearNationalTutoringProgramme

tomjonesdev:
This PR removes a duplicate `TimeIdentifier` option which was added as a temporary solution to release ordering issues with academic years (AYNTP). Releases can now be ordered manually so this is no longer needed.

The work also includes a migration to swap any records which reference the duplicate to the proper one (AY).


src/GovUk.Education.ExploreEducationStatistics.Common/Model/TimeIdentifier.cs
```diff
@@ -285,14 +285,5 @@ public enum TimeIdentifier
...
         [TimeIdentifierMeta("December", "M12", Month)]
         December,
-
-        // EES-3959 Temporary time identifier added to give the correct order to this release
-        // The publication is made up of months and academic years, so by placing this last, the release
-        // is ordered after all the other monthly releases
-        // https://explore-education-statistics.service.gov.uk/find-statistics/national-tutoring-programme/2022-23,
-        // TODO remove this once we've got a solution for ordering releases which mix categories of time periods
-        [TimeIdentifierMeta("Academic year ", "AYNTP", NationalTutoringProgramme, Academic, NoLabel)]
-        AcademicYearNationalTutoringProgramme,
```
leeoades, line 295: And we are absolutely sure that there is nothing persisted, no messages or data or files, that are using this value - which will now fail to deserialise?

tomjonesdev: Possibly, one such place occurred to me right after raising the PR so just hold off for now 👍.

tomjonesdev: Cleared up in Slack, no additional migrations.

benoutram (approved):
Just wondering if we need to clear the cache of any releases that were using the time identifier?

We know there's at least one published publication in Prod which is currently using it.

We should make a note of at least one published release in the Dev environment which can be used for testing before and after this change is deployed.

I suppose there's nothing stopping us migrating any in Prod early and make sure they are still working in advance of deploying this, safe in the knowledge that no Prod releases will be changed when this is deployed, but we probably don't need to go this far if we've got releases set up to do testing with..

PR #5553, opened 2025-01-23
EES-5658 Swap usages of `Year`, `TimePeriodCoverage` and `Slug` from `ReleaseVersion` to `Release`

benoutram:
This PR completes swapping out usages of `ReleaseVersion`  properties `ReleaseName`, `TimePeriodCoverage` and `Slug` with their equivalents from `Release`.

`ReleaseVersion.Year` and `ReleaseVersion.YearTitle` are also removed.

A lot of this work was already covered by https://github.com/dfe-analytical-services/explore-education-statistics/pull/5522.

### UI Test report

image

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
@@ -140,9 +140,11 @@ public abstract class ReleaseServiceTests
                 Assert.Equal(TimeIdentifier.AcademicYear, actualRelease.TimePeriodCoverage);
                 Assert.Equal("2018-19-initial", actualRelease.Slug);
 
-                Assert.Equal(2018, actualReleaseVersion.Year);
+                // TODO EES-5659 - Remove the ReleaseVersion assertions on ReleaseName, TimePeriodCoverage, Slug
+                Assert.Equal("2018", actualReleaseVersion.ReleaseName);
                 Assert.Equal(TimeIdentifier.AcademicYear, actualReleaseVersion.TimePeriodCoverage);
                 Assert.Equal("2018-19-initial", actualReleaseVersion.Slug);
+
                 Assert.Equal(ReleaseType.OfficialStatistics, actualReleaseVersion.Type);
                 Assert.Equal(ReleaseApprovalStatus.Draft, actualReleaseVersion.ApprovalStatus);
                 Assert.Equal(0, actualReleaseVersion.Version);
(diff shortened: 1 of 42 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
```diff
@@ -2332,10 +2278,17 @@ public async Task ConvertsPublishedFromLocalToUniversalTimezone()
...
-                publication.ReleaseVersions[0].Slug
-            )).ReturnsAsync(new ReleaseCacheViewModel(releaseVersionId));
+                publication.Releases[0].Slug
+            )).ReturnsAsync(new ReleaseCacheViewModel(publication.Releases[0].Versions[0].Id));
+
+            releaseCacheService.Setup(s => s.UpdateRelease(
+                publication.Releases[0].Versions[0].Id,
+                publication.Slug,
+                null
+            )).ReturnsAsync(new ReleaseCacheViewModel(publication.Releases[0].Versions[0].Id));
```
jack-hive, line 2291: is this one meant to be here? Looks like we have a setup above

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
benoutram: Ah yeah, good spot! I think I can explain why this has been added.

This test originally setup data without test data generators like this:

```csharp
var publication = new Publication
{
    LatestPublishedReleaseVersionId = Guid.NewGuid(),
    ReleaseVersions =
    [
        new ReleaseVersion
        {
            Id = releaseVersionId,
            Slug = "release-slug",
            Published = DateTime.UtcNow
        }
    ],
    Slug = "publication-slug"
};
```

In this setup, the release version wasn't the publication's latest published release version - see `LatestPublishedReleaseVersionId = Guid.NewGuid(),`.

With the new changes, it's now set up like this:

```csharp
Publication publication = _dataFixture.DefaultPublication()
  .WithReleases(_ => [_dataFixture.DefaultRelease(publishedVersions: 1)]);
```

Behind the scenes in the generator the release version is automatically set as the publication's latest published release version.

To match the old setup we'd have to do something extra to avoid that, either by manipulating `LatestPublishedReleaseVersionId` or by doing something like I put into `Success_NotLatestReleaseInPublication` where two releases are set up and then it deliberately goes on to work with a release version from the non-latest release.

With the release version being the publication's latest release version, it means that in the method under test, as well as updating the cached release version view model (cached as blob `<release-slug.json`) in code:

```csharp
await _releaseCacheService.UpdateRelease(
    releaseVersionId,
    publicationSlug: releaseVersion.Release.Publication.Slug,
    releaseSlug: releaseVersion.Release.Slug);
```

there's also an additional call to update the publication's latest release version view model (cached as blob `latest-release.json`) in code:

```charp
if (releaseVersion.Release.Publication.LatestPublishedReleaseVersionId == releaseVersionId)
{
    // This is the latest published release version so also update the latest cached release version
    // for the publication which is a separate cache entry
    await _releaseCacheService.UpdateRelease(
        releaseVersionId,
        publicationSlug: releaseVersion.Release.Publication.Slug);
}
```

Note that `releaseSlug` is null in this second call because it's the latest release.

It's the mock behaviour for this second call being set up that you are seeing being added in the code here, which wasn't necessary before due to the release version not being the latest.

jack-hive (approved):
looks great!

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseServiceTests.cs
jack-hive: makes complete sense! Thank you for the detailed explanation :D

PR #5558, opened 2025-01-27
Ees 5814 add premium storage accounts

duncan-at-hiveit:
This PR:
- swaps the "Standard" storage account used for the Public API for a "Premium", File share-centric storage account.

## Config changes

Our Public API storage account is now set up as a Premium, `FileStorage` storage account, meaning that it is tailored towards providing file shares with extremely low latency.

Currently all environments are the same, each provisioning a file share of 100GBs (the minimum amount allowed in a Premium storage account.

All other existing storage accounts (e.g. the 3 used by the Data Processor) remain `StandardV2`.

In order to roll out these changes, existing Public API storage accounts firstly needed to be backed up, and then deleted completely, as these settings cannot be applied retrospectively.

### File share naming convention

As we had to rebuild any existing Public API storage accounts in Resource Groups prior to deploying this, it seemed like a good opportunity to update our file share naming convention, from `fs` to `share` which is the recommended one.

### Data Processor deploy process

I removed the Data Processor from Dev during the roll-out of this PR.  As part of that, I noticed that the deploy of the Data Processor would fail if there wasn't already a deployment in the staging slot (i.e. it would fail the first time you tried to deploy the infrastructure, or after removing the function app for whatever reason).

To fix this, I've now set it to allow 404s from an endpoint check while waiting for the function app to come back to life after appsetting updates, which only occurs the first time the deployment takes place.

### YAML template for checking existence of a function app

I broke this piece of logic out into a template as I though we'd need it multiple times.  Turns out we didn't but it's neater being extracted out so I have left this change in!

### DIFF
--- a/infrastructure/templates/public-api/abbreviations.bicep
+++ b/infrastructure/templates/public-api/abbreviations.bicep
@@ -5,8 +5,7 @@ var abbreviations = {
   appManagedEnvironments: 'cae'
   // TODO - remove the "-flexibleserver" suffix and change the suffix of our PSQL instance to "-01"
   dBforPostgreSQLServers: 'psql-flexibleserver'
-  // 'fs' is non-standard - it should be 'share'
-  fileShare: 'fs'
+  fileShare: 'share'
   // 'ai' is non-standard - it should be 'appi'
   insightsComponents: 'ai'
   managedIdentityUserAssignedIdentities: 'id'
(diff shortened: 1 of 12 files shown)


leeoades (approved):
In my limited ability, it looks reasonable enough to me. I trust that it has been applied already to the other environments so I'm sure it's fine.

PR #5559, opened 2025-01-27
OLD EES-5593 Public API wildcard data set versioning

mmoalin:
This PR adds datasets wildcard versioning support to the following API endpoints:

```
[get]
data-sets/{dataSetId}/meta
data-sets/{dataSetId}/query
data-sets/{dataSetId}/csv
data-sets/{dataSetId}/versions/{dataSetVersion}
data-sets/{dataSetId}/versions/{dataSetVersion}/changes

[post]
data-sets/{dataSetId}/query
```

For example:
- `dataSetVersion=2.*` should return the latest minor version in the v2 series
- `dataSetVersion=2.1` should return v2.1
- `dataSetVersion=2.0` should return v2.0
- `dataSetVersion=2` should return v2.0

Thank you in advance; also thank you @Mark, @Lee & @Duncan for your initial feedbacks over our calls!


tomjonesdev:
I'll leave the proper review to someone who's familiar with the public API, but just a general comment: before this gets merged in, could you tidy up the commits with a reset please?

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/VersionUtilsTests.cs
```diff
@@ -68,5 +69,32 @@ public void InvalidVersion_FailsToParse(string versionString)
...
+        [InlineData("1.*.*")]
+        [InlineData("  1.*.*")]
+        [InlineData("  1.*.*  ")]
+        [InlineData("1.*")]
+        [InlineData("1.2.*")]
+        [InlineData("v*")]
+        [InlineData("v1.*.*")]
+        [InlineData("v1.*")]
+        [InlineData("v1.2.*")]
+        public void ValidWildCardVersion_SuccessfullyParsed(
```
leeoades, line 83: I think we need to assert what the value is that was parsed. I realise that we're essentially testing the library here.

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/VersionUtils.cs
```diff
@@ -16,6 +16,18 @@ public static bool TryParse(string versionString, out SemVersion version)
 
         version = sv;
 
+        return successful;
+    }
+    public static bool TryParseWildcard(string versionString, out SemVersionRange version)
```
leeoades, line 21: version needs to be nullable here.
You can simplify this quite a lot by doing this:

`public static bool TryParseWildcard(string versionString, [NotNullWhen(true)] out SemVersionRange? version) =>
        SemVersionRange.TryParse(
            versionString,
            SemVersionRangeOptions.OptionalMinorPatch
            | SemVersionRangeOptions.AllowLowerV,
            out version);`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/UnitTests/WildCardDatasetVersionTests.cs
```diff
@@ -0,0 +1,201 @@
...
+
+        var release = new Release()
+        {
+            DataSetFileId = Guid.NewGuid(),
+            ReleaseFileId = Guid.NewGuid(),
+            Slug = "test",
+            Title = "Test"
+        };
+
+        var versions = new List<DataSetVersion>
```
leeoades, line 89: A suggestion
- it would be nice to have an array of just the version numbers here - so that that was easy to read. Then use a `select` statement to build the DataSetVersions from it. I think that would be clearer.
- another alternative would be to pass in the versions into this setup so that it was clearer from the tests what the versions are

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/UnitTests/WildCardDatasetVersionTests.cs
```diff
@@ -0,0 +1,201 @@
...
+    [InlineData("v0.2.*", 0, 2, 1)]
+    [InlineData("v0.1.*", 0, 1, 3)]
+    [InlineData("v2.*.*", 2, 1, 4)]
+    [InlineData("v2.*", 2, 1, 4)]
+    [InlineData("v2.1.*", 2, 1, 4)]
+    [InlineData("v1.*.*", 1, 3, 0)]
+    [InlineData("v1.2.*", 1, 2, 0)]
+    [InlineData("v1.1.*", 1, 1, 1)]
+    [InlineData("v1.*", 1, 3, 0)]
+    public async Task TestDataSetVersions_ReturnsExpectedVersion(string versionString,
```
leeoades, line 40: A couple of extra edge cases:
- Higher version than is available - what should that return? e,.g. v6.* Not found I'm assuming
- In-between version: remove 3.0.0 from your available and query for v3.0.0. Show that it returns not found
- Lower: I would remove 0.0.* from your available dataset versions and then query for 0.0.* to see what happens. Should be not found.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/UnitTests/WildCardDatasetVersionTests.cs
leeoades, line 1: I think a more appropriate location for these tests would be along side this file
`src\GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests\DataSetVersionTests.cs`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionQueryableExtensions.cs
```diff
@@ -12,6 +13,44 @@ public static async Task<Either<ActionResult, DataSetVersion>> FindByVersion(
         Guid dataSetId,
         string version,
         CancellationToken cancellationToken = default)
+    {
+        return await (version.Contains('*') ? queryable.FindByWildcardVersion(dataSetId, version, cancellationToken) : queryable.FindBySpecificVersion(dataSetId, version, cancellationToken));
```
leeoades, line 17: This would be more readable when the 2 parts of the ternary are on separate lines:

`    public static async Task<Either<ActionResult, DataSetVersion>> FindByVersion(
        this IQueryable<DataSetVersion> queryable,
        Guid dataSetId,
        string version,
        CancellationToken cancellationToken = default) =>
        await (version.Contains('*') 
            ? queryable.FindByWildcardVersion(dataSetId, version, cancellationToken) 
            : queryable.FindBySpecificVersion(dataSetId, version, cancellationToken));`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionQueryableExtensions.cs
```diff
@@ -12,6 +13,44 @@ public static async Task<Either<ActionResult, DataSetVersion>> FindByVersion(
...
+    public static async Task<Either<ActionResult, DataSetVersion>> FindByWildcardVersion(
+        this IQueryable<DataSetVersion> queryable,
+        Guid dataSetId,
+        string wildcardedVersion,
+        CancellationToken cancellationToken = default)
+    {
+        if (!VersionUtils.TryParseWildcard(wildcardedVersion, out var semVersionRange)) 
+            return new NotFoundResult();
+        
+        var parts = wildcardedVersion.Trim('v').Split('.');
```
leeoades, line 33: So you don't need to manually parse the string again here, you have the semVersionRange object which has done the work for you.

leeoades:
> I'll leave the proper review to someone who's familiar with the public API, but just a general comment: before this gets merged in, could you tidy up the commits with a reset please?

Hey @tomjonesdev  - is there an option to squash merge the PR? I'm assuming that's what you mean by the `reset`?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionQueryableExtensions.cs
mmoalin: Unfortunately because the semVersionRange is not a SemVersion (this can provide the single parsed version), it returns a range which has a start and end attribute. I've found the semVersionRange object doesn't provide a useful property for me to use. I'd be happy to discuss some more over a call or to provide a demo.

src/explore-education-statistics-api-docs/README.md
```diff
@@ -21,12 +21,45 @@ Ruby and its required gems:
...
+    ./configure --prefix=$HOME/.rbenv/versions/3.2.6
+    make
+    make install
+    ```
+
+3. Add Ruby Version to rbenv
+    ```shell
+    rbenv rehash
+    rbenv global 3.2.6
+    ```
```
ntsim, line 45: This doesn't look right we shouldn't need to set the global version of Ruby.

The correct approach would be to go into the API docs directory and just do `rbenv install`.

```shell
cd src/explore-education-statistics-api-docs
rbenv install
```

I believe `rbenv` automatically detects the correct version from `.ruby-version` when cd'ing into the directory and changes the shell to use it (if installed).

src/explore-education-statistics-api-docs/README.md
```diff
@@ -21,12 +21,45 @@ Ruby and its required gems:
...
+4. Confirm you have the right ruby version
+    ```shell
+    ruby -v
+    ```
+
+5. Install Bundler
+    ```shell
+    gem install bundler
+    rbenv rehash
+    ```
```
ntsim, line 56: Is this necessary? Fairly sure that Bundler is already installed with modern Ruby versions. From the Bundler docs:

> This guide assumes that you have [Ruby](https://www.ruby-lang.org/en/downloads/) installed. If you do not have Ruby installed, do that first and then check back here! Any modern distribution of Ruby comes with Bundler preinstalled by default.

https://bundler.io/guides/getting_started.html#getting-started

src/explore-education-statistics-api-docs/README.md
```diff
@@ -21,12 +21,45 @@ Ruby and its required gems:
...
 
 ## Getting started
 
 Once the pre-requisites have been installed, follow these steps:
 
-1. Install the project's Ruby dependencies:
+1. Navigate to the project's root: `/explore-education-statistics/src/explore-education-statistics-api-docs` & Install the project's Ruby dependencies:
```
ntsim, line 62: Don't need to include the `/explore-education-statistics` at the start of this path.

Someone may choose to clone the repo to a different path so we should work under the assumption that everything is just relative from the project root. The correct pathing would be `src/explore-education-statistics-api-docs` (note the lack of leading slash to avoid confusion with absolute pathing)

src/explore-education-statistics-api-docs/README.md
```diff
@@ -21,12 +21,45 @@ Ruby and its required gems:
...
+    tar -xzvf ruby-3.2.6.tar.gz
+    cd ruby-3.2.6
+    ```
+
+2. install ruby 3.2.6 version
+    ```shell
+    ./configure --prefix=$HOME/.rbenv/versions/3.2.6
+    make
+    make install
+    ```
```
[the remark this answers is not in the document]
mmoalin: Thanks Nick, I'll undo this change to the readme. 
Yeah I've had some issues and I was noting down my workarounds and thought it might be good to include here. I've clearly got the wrong end of the stick as I didn't realize they could be problematic. you're right, we should stick to the recommended installation guides.

src/explore-education-statistics-api-docs/README.md
mmoalin: Undoing as per top comment

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/VersionUtils.cs
mmoalin: Thanks Lee, I've modified my new change to be more simpler but not the existing code.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/UnitTests/WildCardDatasetVersionTests.cs
mmoalin: Thanks both

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/UnitTests/WildCardDatasetVersionTests.cs
mmoalin: Cheers Lee, I've added 3 extra edge cases. I've tested the last edge case locally and got the expected result but i havent removed the 0.0.1 because i want to use it in the other test.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Controllers/DataSetVersionsController.cs
```diff
@@ -56,7 +56,7 @@ public async Task<ActionResult<DataSetVersionPaginatedListViewModel>> ListDataSe
     [SwaggerResponse(404, type: typeof(ProblemDetailsViewModel))]
     public async Task<ActionResult<DataSetVersionViewModel>> GetDataSetVersion(
         [SwaggerParameter("The ID of the data set.")] Guid dataSetId,
-        [SwaggerParameter("The data set version e.g. 1.0, 1.1, 2.0, etc.")] string dataSetVersion,
+        [SwaggerParameter("The data set version e.g. 1.0, 1.1, 2.0, etc. \n\n This supports wildcard data set versions so that you can retreive the latest minor version automatically. \n\n For example: dataSetVersion=v2.* or dataSetVersion=2.* would return the latest minor version in the v2 series.")] string dataSetVersion,
```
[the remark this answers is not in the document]
mmoalin: Thanks Nick, please let me know if the new change in /overview/versioning/index.html is fit for purpose.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/VersionUtilsTests.cs
leeoades: OK. The reason we've wrapped SemVersionRange.TryParse is to pass in the parameter OptionalMinorPatch - This isn't being tested though. I can remove it and no tests fail.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/UnitTests/WildCardDatasetVersionTests.cs
leeoades: I'm sure you can test it with 1.0.1 instead for example.  By not including a test for a lower version, there's no knowing what it will do or if it changes. I doubt for example in reality that we have dataset versions 0.0.1 so if the lowest version is 1.0.0 or 0.6.0, what happens if we request 0.0.1?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionQueryableExtensions.cs
leeoades: In that case, have the TryParseWilcard return a different representation of a version that does do what you want. Create your own WildcardVersion if needs be if there isn't something appropriate in the Semver library and you can put the parse on that, and then it can expose the Major, Minor and Patch as `uint?` values for example.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/UnitTests/WildCardDatasetVersionTests.cs
```diff
@@ -0,0 +1,201 @@
...
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Database;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Extensions;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Utils;
+using Microsoft.EntityFrameworkCore;
+using Moq;
+using Moq.EntityFrameworkCore;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests.UnitTests;
+
+public class WildCardDatasetVersionTests
```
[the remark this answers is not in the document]
leeoades: Ah, interesting arrangement. OK, I'd not noticed that. 👍

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/VersionUtilsTests.cs
```diff
@@ -1,4 +1,5 @@
 #nullable enable
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
```
leeoades, line 2: This is superfluous.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/UnitTests/WildCardDatasetVersionTests.cs
mmoalin: No worries thats done

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionQueryableExtensions.cs
mmoalin: That's a great idea had a go at that, thanks again Lee

src/explore-education-statistics-api-docs/source/overview/versioning/index.html.md.erb
```diff
@@ -56,3 +56,14 @@ You can find lists of changes via:
...
+### Automatically receiving minor updates
+
+New data set versions can automatically be retreived from the API without having to specify the new data set version numbers explicitly. This can be used by
+supplying endpoints with the parameter `dataSetVersion` an astrick `*` wildcard character in the place of the minor or major version. 
+
+For example: 
+
+- `dataSetVersion=*` or `dataSetVersion=v*` will return the latest major and minor version for the given data set.
+- `dataSetVersion=2.*.*`, `dataSetVersion=2.*` or `dataSetVersion=v2.*` will return the latest minor version of the v2 series 
+- `dataSetVersion=2.3.*` or `dataSetVersion=v2.3.*` will return the latest patch version of the v2.3 series.
```
ntsim, line 69: Nice first draft of this :smile:

On reflection, we should probably go into a bit more detail about the `dataSetVersion` parameter generally.

Would suggest pasting in the following:

```md
### Requesting data set versions

There are several API endpoints that accept a `dataSetVersion` query parameter e.g. the 
[Get a data set version](/reference-v1/endpoints/GetDataSetVersion/index.html) endpoint.
The `dataSetVersion` parameter can be supplied with a version number (e.g. 1.0, 1.1, 2.0, 
etc) that controls the data set version that should be used in the request.

For example:

- `dataSetVersion=1.1` uses version 1.1
- `dataSetVersion=1.0` uses version 1.0
- `dataSetVersion=1` uses version 1.0

### Automatically receiving minor version updates

When using the `dataSetVersion` parameter, minor data set version updates can automatically
be retrieved from the API without having to change the version number explicitly. This can 
be done by including an asterisk `*` wildcard character in place of a major or minor version.

For example: 

- `dataSetVersion=*` uses the latest major and minor version
- `dataSetVersion=2.*` uses the latest minor version of the v2 series
```

Note I've removed any mention of **patch** versions as end users aren't supposed to directly know these exist.

Currently, we only have patch versions as an internal concept. They've not really been implemented yet but exist at the DB level as we may need them later for types of 'data replacements' (one to explain another day).

To an end user, they should only be concerned with specifying major and minor versions. They should automatically receive the latest patch version (if there is one).

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/VersionUtilsTests.cs
```diff
@@ -6,6 +6,52 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils;
...
+            int? major,
+            int? minor,
+            int? patch
+            )
+        {
+            Assert.True(WildcardVersion.TryParse(versionString, out var wildcardVersion));
+
+            Assert.NotNull(wildcardVersion);
+            Assert.Equal((uint?)major, wildcardVersion.VersionMajor);
+            Assert.Equal((uint?)major, wildcardVersion.VersionMajor);
```
leeoades, line 37: note: duplicated line

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/VersionUtils.cs
```diff
@@ -20,34 +22,33 @@ public static bool TryParse(string versionString, out SemVersion version)
...
-
-    public static bool TryParseWildcard(string versionString,
-        [NotNullWhen(true)] out WildcardVersion? version)
+}
+public record WildcardVersion(uint? VersionMajor, uint? VersionMinor, uint? VersionPatch)
+{
+    public static bool TryParse(string versionString, out WildcardVersion? version)
     {
-        versionString = versionString.Trim();
+        var parts = versionString.Trim('v').Trim().Split('.');
```
leeoades, line 30: Just a question: do we want to support leading blank space before the _v_? If we do, then I don't think this does.

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/VersionUtils.cs
```diff
@@ -20,34 +22,33 @@ public static bool TryParse(string versionString, out SemVersion version)
 
         return successful;
     }
-
-    public static bool TryParseWildcard(string versionString,
-        [NotNullWhen(true)] out WildcardVersion? version)
+}
+public record WildcardVersion(uint? VersionMajor, uint? VersionMinor, uint? VersionPatch)
```
leeoades, line 26: Forgot to mention yesterday - my bad. I think we can call these properties Major, Minor and Patch. I don't think we need a _Version_ prefix on the name.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/VersionUtilsTests.cs
```diff
@@ -6,6 +6,52 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Tests.Utils;
...
+        [InlineData("*", null, null, null)]
+        [InlineData("1.*.*", 1, null, null)]
+        [InlineData("  1.*.*", 1, null, null)]
+        [InlineData("  1.*.*  ", 1, null, null)]
+        [InlineData("1.*", 1, null, null)]
+        [InlineData("1.2.*", 1, 2, null)]
+        [InlineData("v*", null, null, null)]
+        [InlineData("v1.*.*", 1, null, null)]
+        [InlineData("v1.*", 1, null, null)]
+        [InlineData("v1.2.*", 1, 2, null)]
```
leeoades, line 25: As discussed - this currently supports whitespace around the numbers e.g. _v1  .  2   .  3_ which we probably don't want.

mmoalin: Good spot, covered this now in the tests and the logic.

src/explore-education-statistics-api-docs/source/overview/versioning/index.html.md.erb
mmoalin: That makes sense, that's very helpful. thank you Nick!

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/VersionUtils.cs
```diff
@@ -19,3 +23,32 @@ public static bool TryParse(string versionString, out SemVersion version)
...
+        var parts = versionString.Trim(' ', 'v').Split('.');
+
+        if (parts.Length == 1 && (parts[0].IsNullOrEmpty() || parts[0] == "*"))
+        {
+            version = new WildcardVersion(null, null, null);
+            return true;
+        }
+
+        version = null;
+        if (parts.Length > 3 || !parts.All(part => part.Is("*") || uint.TryParse(part, System.Globalization.NumberStyles.None, null, out _)))
```
mmoalin, line 39: Me and Lee had a slack conversation about not parsing the uint multiple times in this function, he has recommended I make more efficient use of this so just on this now.

PR #5560, opened 2025-01-28
EES-5819 Can specify chromedriver version in UI tests pipeline run sc…

mmyoungman:
…ript

This PR allows us to specify the chromedriver version we want to use when using the UI tests pipeline run script.

### DIFF
--- a/tests/robot-tests/scripts/run_tests_pipeline.py
+++ b/tests/robot-tests/scripts/run_tests_pipeline.py
@@ -35,6 +35,9 @@ def run_tests_pipeline():
     if args.rerun_attempts:
         run_tests_command += f" --rerun-attempts {str(args.rerun_attempts)}"
 
+    if args.chromedriver_version:
+        run_tests_command += f" --chromedriver {str(args.chromedriver_version)}"
+
     subprocess.check_call(run_tests_command, shell=True)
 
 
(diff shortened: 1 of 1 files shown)


tomjonesdev (approved):

tests/robot-tests/scripts/run_tests_pipeline.py
```diff
@@ -67,5 +70,13 @@ def run_tests_pipeline():
     parser.add_argument("--processes", dest="processes", help="number of processes to run", required=True)
 
     parser.add_argument("--rerun-attempts", dest="rerun_attempts", type=int, help="Number of rerun attempts")
+
+    parser.add_argument(
+        "--chromedriver",
+        dest="chromedriver_version",
+        metavar="{version}",
+        help="specify which version of chromedriver to use",
+    )
```
benoutram, line 79: Webdriver manager tries to find the closest matching Chromedriver for the browser version already. Could this lead to us forgetting that we've pinned the Chromedriver version in the pipeline when it should already be the closest match?

Maybe there is still some value in this change if a particular version of chromedriver has a problem, we could use this to try temporarily downgrading?

mmyoungman: Talked to Ben about this on Slack - don't think either of us are fussed either way, so we'll close it

mmyoungman:
Webdriver manager automatically chooses a chromedriver version appropriate for your google-chrome version, so this isn't necessary - arguable that it could be used for an older version to avoid an issue but we've not needed that before, so just closing this PR instead of merging it.

PR #5562, opened 2025-01-28
Ees 5802 add pre prod public api deploy stages

duncan-at-hiveit:
This PR:
- adds Pre-Prod as a deployable environment.
- Makes various syntactic improvements to the Bicep files.
- Corrects a typo in a deploy step.
- Removes manual deploy triggers in favour of leveraging DevOps Environments' approval steps.

This PR branch has been used to successfully deploy the Public API infrastructure to the Pre-Prod environment. 

## DevOps environment Approval steps

By default, a Public API deploy will now be triggered after a successful build of application code from the `master` branch.  It will not automatically continue the deployment though, because we now have an "Approval" step in place on the Pre-Prod environment.  Members of a new "Deploy Approvers" group can approve the deploy manually.

image

image

image

image


infrastructure/templates/public-api/ci/azure-pipelines.yml
```diff
@@ -3,27 +3,51 @@ trigger: none
 parameters:
   - name: deploySharedPrivateDnsZones
     displayName: Do the shared Private DNS Zones need creating or updating?
+    values:
```
leeoades, line 6: I think the verboseness of the options. I believe the responses would read more nicely if the options were _Yes_ and _No_?

infrastructure/templates/public-api/components/appGateway.bicep
```diff
@@ -99,8 +99,8 @@ resource publicIPAddresses 'Microsoft.Network/publicIPAddresses@2024-01-01' = [f
     publicIPAllocationMethod: 'Static'
     idleTimeoutInMinutes: 4
     dnsSettings: {
-      domainNameLabel: replace(site.fqdn, '.', '')
-      fqdn: '${replace(site.fqdn, '.', '')}.${location}.cloudapp.azure.com'
+      domainNameLabel: replace(replace(site.fqdn, '.', ''), '-', '')
```
leeoades, line 102: I thought this use of nested replace probably had a neater solution.
Consider introducing a bicep function to remove multiple specified character from a string.
This example shows a multiple replace - but obviously a multiple remove only needs to specify chars to remove.
https://alexdantico.com/2024/05/19/simplifying-string-replacement-in-azure-bicep-using-user-defined-functions/

infrastructure/templates/public-api/components/storageAccount.bicep
```diff
@@ -69,61 +69,53 @@ resource storageAccount 'Microsoft.Storage/storageAccounts@2023-05-01' = {
...
     serviceType: 'fileService'
-    subnetId: servicePrivateSubnetIds.file
+    subnetId: privateEndpointSubnetIds.?file ?? ''
     location: location
     tagValues: tagValues
   }
 }
 
-module blobStoragePrivateEndpointModule 'privateEndpoint.bicep' = if (servicePrivateSubnetIds.blob != '') {
+module blobStoragePrivateEndpointModule 'privateEndpoint.bicep' = if (privateEndpointSubnetIds.?blob != null) {
```
leeoades, line 85: Do we need both the null check here and the safe dereference below? I'm guessing that one of those is now redundant.

leeoades (approved):
Obviously, I don't really know what I'm talking about but a few comments if useful.

infrastructure/templates/public-api/ci/azure-pipelines.yml
duncan-at-hiveit: Good suggestion!  I've updated it now based on this suggestion ta.

infrastructure/templates/public-api/components/appGateway.bicep
duncan-at-hiveit: Nice one ta.  Thanks for the investigation and the link!  I have borrowed and used this function above, and also piggy-backed off it for a `removeMultiple` implementation too ta.  Note the `any()` that I was forced to use and lose about an hour on, but won the fight eventually.  You'll see I've updated any areas that were doing multiple replacements now as well.

PR #5571, opened 2025-01-30
Ees 5545

ntsim:
This PR adds a new 'Data file replacements' table to the 'Data and files' section. This table only contains data files that currently have a replacement in progress.

image

## Related changes

- Refactored out `DataFileSummaryList` component to contain data file details (instead of `DataFileDetailsTable`). Summary lists are more appropriate where there is only a single set of details to show.
- Removed max-width from the 'Status' column as this was mostly unnecessary due to the new 'Data file replacements' table. It was also causing the import progress bar to break out of its container for smaller screen sizes.


bennettstuart (approved):
Nice

PR #5584, opened 2025-02-03
EES-5835 fix to latest release link

rianski:
Fix to latest release link where the user would be redirected to the previously viewed release slug rather than the latest

### DIFF
--- a/src/explore-education-statistics-frontend/src/modules/find-statistics/PublicationReleasePage.tsx
+++ b/src/explore-education-statistics-frontend/src/modules/find-statistics/PublicationReleasePage.tsx
@@ -40,7 +40,7 @@ const PublicationReleasePage: NextPage<Props> = ({ release }) => {
     rsi => rsi.isLegacyLink || rsi.description !== release.title,
   );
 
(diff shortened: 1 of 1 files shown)


src/explore-education-statistics-frontend/src/modules/find-statistics/PublicationReleasePage.tsx
```diff
@@ -126,15 +130,11 @@ const PublicationReleasePage: NextPage<Props> = ({ release }) => {
                     <Link
                       className="govuk-!-display-none-print govuk-!-display-block govuk-!-margin-bottom-3"
                       unvisited
-                      to={`/find-statistics/${release.publication.slug}/${release.slug}`}
+                      to={`/find-statistics/${release.publication.slug}/${releaseSeriesNonLegacy?.releaseSlug}`}
```
benoutram, line 133: I think this works but naming it something like `latestReleaseSeriesItem` rather than `releaseSeriesNonLegacy` would help explain why it's being used here.

benoutram (commented):
I think `relea

benoutram (approved):

PR #5588, opened 2025-02-04
Ees 5802 add secret headers between fuapi and app gateway

duncan-at-hiveit:
# UI test runs

## Local

The only failing tests here are the 3 tests currently failing on Dev also.

image

## Dev

The only failing tests here are the 3 tests currently failing on Dev also.

image

This PR:
- secures the App Gateway so that only traffic coming via FUAPI is permitted.
- adds a private URL to the Public API Container App for Admin's use so that it does not need to go externally via the public internet when calling the Public API Container App.
- updates URLs for the Public API and the Public API docs so that they are directed to FUAPI rather than the App Gateway.
- adds support for adding Firewall Policies to specific HTTP listeners as well as globally for an App Gateway.
- updates the Robot test basic auth handling to prevent basic auth headers being sent via FUAPI requests.

# Securing the App Gateway

When people use the Public API, they don't interact directly with our Container App or even our App Gateway.  Instead, all traffic passes through "FUAPI" first, which is effectively a gateway of its own that is managed by DfE.  We want to make sure that all traffic passes only through FUAPI, and that no-one can interact with the Public API through our App Gateway directly.  The reason for this is that FUAPI gives us lots of protection, rate limiting, security etc etc, plus lets people subscribe to use our Public API.

We secure the App Gateway by enforcing that a special HTTP header `X-FUAPI-Secret` is present in every request coming to the App Gateway and that it has the correct value.  The correct value is stored in Key Vault and ALSO added to the FUAPI backends as an HTTP header that they will pass with every request that comes through the backend.  So effectively, when a request hits a FUAPI URL for, say, our Dev environment, the header will be added to the request that is forwarded to the App Gateway.

The App Gateway (or rather, the listener that listens to traffic on the App Gateway `dev.statistics...` domain) will then use its custom rules to reject any traffic that does not contain an `X-FUAPI-Secret` header at all, or if it does, will reject if its value does not equal the secret value that should match in FUAPI and our Key Vault.

### FUAPI side

We firstly create a new secret value within our FUAPI Workspace:

image

We then amend our Dev backend in FUAPI to always pass that secret value in a special HTTP header:

image

### App Gateway side

On the App Gateway side, we attach a WAF policy to the Public API's listener that guarantees that this header is present and is of the correct value.  Failure to include the correct header and value results in a 403:

image

## Allowing access for Admin

The Admin App Service talks to the Public API Container App in order to get changelog details between 2 different data set versions, as seen here:

image

In order to allow this without having to have Admin present the special FUAPI header to the App Gateway, we instead get Admin to bypass the public internet and talk to the Container App directly.  Hence Admin now takes a `PrivateUrl` and a `PublicUrl` for the Public API, one for talking securely over the VNet and bypassing the App Gateway, and one for presenting URLs for guidance in the front end (note the scribbles on these screenshots - these values are corrected in the final PR!):

image

image

image

## Config updates to use new FUAPI URL on public pages

As well as the updates to Admin above, the public site has also been updated to use the FUAPI URL.

Here we see the results of a public frontend API call grabbing a preview of the data set from the Public API:
 
image

And here we see the guidance on the frontend for using the API to query the data set:

image

## Updating Robot tests to stop basic auth headers being sent to FUAPI

Originally the Robot tests configured the browser to add basic auth headers to every request being sent, by use of the `Network.setExtraHttpHeaders` support.  This meant that the XHR requests being sent to the Public API via FUAPI were also sending those headers.  FUAPI in turn was rejecting those requests because it doesn't support the `Authentication` header being sent to it by default.

In order to fix this in the simplest way, I removed the behaviour to add headers entirely, and instead enforced that each browser being used in tests visit the public site homepage once with basic auth credentials in the URL in order to exchange them for appropriate headers for subsequent requests to the site alone.  I had considered just doing this for browsers that were being used for the frontend, but for the sake of a 1 second trip to the frontend for each browser, I opted for simplicity, so that the test writer doesn't even need to think about whether or not their particular browser instance needs to interact with the frontend or not.

A subsequent PR https://github.com/dfe-analytical-services/explore-education-statistics/pull/5595 clears up the now-redundant special-case frontend and backend navigation words.

## Additional Robot changes

To get a full successful UI test run from my local machine against the Dev environment (bar the 3 existing failing tests on Dev) I had to make some changes:
* `redirects.robot` was making incorrect assertions, so I've corrected those.
* Our scrolling behaviour was scrolling an element just into view, so that it could appear right at the top or right at the bottom of the page, which meant that it could be getting blocked with edit controls on certain admin pages (e.g. Release content and its edit / preview hover tools).  I've changed this so that now it scrolls target elements into the *center* of the viewpane where possible.
* There was an intermittent bug whereby sometimes secondary stats blocks don't render after their backing data block is changed.  I've raised a bug ticket to fix this https://dfedigital.atlassian.net/browse/EES-5856, but in the meantime I've forced it to always render by just scrolling the viewpane slightly.
* I added a `user waits` to `prerelease_and_amend.robot` to wait for the bottom portion of the page to load before continuing with the test, as this can be slower than the rest of the page load.
* I made some of our logging less verbose / only log on failure cases.


tests/robot-tests/tests/libs/common.robot
```diff
@@ -145,7 +164,7 @@ user scrolls up
 user scrolls to element
     [Arguments]    ${element}
     user waits until page contains element    ${element}
-    scroll element into view    ${element}
+    user scrolls element to center of view    ${element}
```
N-moh, line 167: Wow, this is a good one to use in several places :)

mmoalin (approved):

PR #5589, opened 2025-02-04
EES-5554 Fix accessibility issues in public API docs

ntsim:
This PR fixes various accessibility issues that have been highlighted by @N-moh

- Added unique `aria-label` / `aria-labelledby` attributes for parameter tables
- Fixed header appearing really broken at 200%+ zoom
- Fixed overflowing text from large word breaking the entire layout at 200%+ zoom (using `word-break`)
- Fixed mobile TOC styling being mostly broken when 200%+ zoom

Additionally, I've made some more tweaks whilst in the area:

- Fixed 'Endpoints' and 'Schemas' child items not being displayed in the mobile TOC
- Fixed parameter table styling being mostly broken at 200%+ zoom

## Screenshots

New header:

image

New TOC:

image

New parameter tables:

image

### DIFF
--- a/src/explore-education-statistics-api-docs/source/partials/_parameter_table.html.erb
+++ b/src/explore-education-statistics-api-docs/source/partials/_parameter_table.html.erb
@@ -3,7 +3,7 @@
   role="region"
   tabindex="0"
 >
-  <table>
+  <table class="app-parameter-table">
     <thead>
     <tr>
       <th scope="col">Parameter</th>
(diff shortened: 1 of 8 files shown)


amyb-hiveit (approved):
I can't get this running locally to fully check, but I think this all looks ok.

I did find another issue that hasn't been raised, though I don't know if it's something we can fix or if it's outside of our control - the heading level on search results skips from h1 to h3:
Screenshot 2025-02-06 111935

PR #5606, opened 2025-02-14
EES-5753 Add and use new email template to highlight breaking changes…

mmoalin:
This PR adds a new email template to be used by API data set notification subscribers. It also uses an existing email template but renames it to be more accurate for its purpose.

When a major data set version is published, this e-mail template will be sent:
image

When a minor data set version is published, the subscribers will receive a message that uses this template:
image

If a new published data set's version contains breaking changes (i.e., has a major version increment), then an email which highlights this is sent. Otherwise, if the new published data set contains only minor changes, a different e-mail template will be sent.

This uses a small utility class that can be unit tested to decide whether to send the major data set email or the minor data set email by checking the version number received by the function `send_email` is a new major version (and is not the very first major version published).

This also adds the associated template IDs as secrets in the infrastructure templates json. Please note this requires manual configurations in the infrastructure to add the new template id (major data set updates) and rename the existing other template id (minor data set updates).


src/GovUk.Education.ExploreEducationStatistics.Notifier/Services/ApiSubscriptionService.cs
```diff
@@ -363,13 +363,15 @@ private void SendNotificationEmail(
             { NotifierEmailTemplateFields.DataSetVersion, version },
             { NotifierEmailTemplateFields.UnsubscribeUrl, unsubscribeUrl }
         };
-
+        var govUkNotifyEmailsConfigs = govUkNotifyOptions.Value.EmailTemplates;
         emailService.SendEmail(
             email: subscription.RowKey,
-            templateId: govUkNotifyOptions.Value.EmailTemplates.ApiSubscriptionNotificationId,
+            templateId: new ApiSubscriptionTemplateIdHelper(
```
leeoades, line 369: Instead of passing in the ids like this, I would recommend one of the following:
- inject the options into ApiSubscriptionTemplateIdHelper directly using DI, and then you can inject IApiSubscriptionTemplateIdHelper into this class and call SelectEmailNotificationTemplateID on it.
- change ApiSubscriptionTemplateIdHelper.SelectEmailNotificationTemplateID to be an extension method on govUkNotifyEmailsConfigs so that you can call govUkNotifyEmailsConfigs.SelectEmailNotificationTemplateID
- or you could add that method directly to govUkNotifyEmailsConfigs

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/ApiSubscriptionFunctionsTests.cs
```diff
@@ -746,6 +747,43 @@ await functions.RemoveExpiredSubscriptions(
         }
     }
 
+    public class ApiSubscriptionEmailsHelperTests
+    {
+        [Theory]
```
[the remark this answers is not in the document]
mmoalin: Good question, I don't think it is possible to go from v.1.9 to v2.1 so I believe the test cases are fit for purpose as is unless I've misinterpreted your comment. 
I'll try to confirm this from the team but from my testing I have found no way of doing this within the EES system.

src/GovUk.Education.ExploreEducationStatistics.Notifier/Services/ApiSubscriptionService.cs
mmoalin: Thanks Lee, I've considered putting it in the govUkNotifyEmailsConfig but was put off as it's used for configs. I agree that is the simplest option tho so reverted to this.

src/GovUk.Education.ExploreEducationStatistics.Notifier/Options/EmailTemplateForNotificationSubscribers.cs
leeoades, line 1: Is this class required any more?

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/ApiSubscriptionFunctionsTests.cs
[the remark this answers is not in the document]
leeoades: It's worth checking. It does seem overly restrictive. For example, you have a test below that says that version 0.0.1 is a minor release, even though that is likely to be the first release.

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/ApiSubscriptionFunctionsTests.cs
```diff
@@ -560,7 +561,7 @@ public async Task Success()
             fixture.NotificationClient
                 .Setup(mock => mock.SendEmail(
                     Email,
-                    GetGovUkNotifyOptions().EmailTemplates.ApiSubscriptionNotificationId,
+                    GetGovUkNotifyOptions().EmailTemplates.ApiSubscriptionMinorDataSetVersionNotificationId,
```
leeoades, line 564: This setup only allows the minor template version to be used. If your intention is that this test doesn't care, then instead consider using A<string>._
Having said that, do you have a test that ensures SendEmail is being called with the expected template anywhere?

mmoalin: Just some background: this change is because I renamed the variable storing a template ID shared by this test and other bits of the codebase. This is a pre-existing test that isn't intended to change unless I have to and so this was one of those cases (as renaming the variable affected this test).

leeoades: So this test was from before, when the only expectation was that it would send an email with the single template id. There are now two different scenarios, based on the version number. Therefore, this test needs to be expanded on - with two different initial conditions (version major and version minor) and two different expectations (id major and id minor).
To put it another way, you can change the ApiSubscriptionService code to this:
```
emailService.SendEmail(
            email: subscription.RowKey,
            templateId: govUkNotifyOptions.Value.EmailTemplates.ApiSubscriptionMinorDataSetVersionNotificationId,
            values: personalisation);
```
and all of the tests pass.

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/ApiSubscriptionFunctionsTests.cs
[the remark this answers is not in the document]
mmoalin: Thanks Lee!
> For example, you have a test below that says that version 0.0.1 is a minor release, even though that is likely to be the first release.

This is the logic as is: the first release (regardless of if the data set version is a minor or a major release) sends the 'minor data set version ' email template ID because of me using the the template ID name `ApiSubscriptionMinorDataSetVersionNotificationId`. Maybe this is the source of the confusion and is worth a second look from my POV.

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/ApiSubscriptionFunctionsTests.cs
mmoalin: The test case `NotifySubscribersTests.Success` uses mocks in order to instantiate INotificationClient (which implements `SendEmail`). The new logic sits outside of what is being mocked which is why I was inclined to use a separate unit test to ensure it's coverage. 

There are many instances in the codebase that uses the `SendEmail` method in tests 
image
I'm put off from making any changes to any tests that are unrelated to this ticket unless I have to which was the case in this instance. Let's chat more next week mate.

mmoalin: I'll certainly look at this again on Monday and am grateful for your thoughts mate.

mmoalin: Please see comments under [ApiSubscriptionFunctionsTests.cs](https://github.com/dfe-analytical-services/explore-education-statistics/pull/5606/files#diff-06392558eb1dc9d4e9a2a0d0255541d64516e9b0597240a61f3d328d32907e3b)

mmoalin: Very grateful for your explanation on this Lee. I've updated the integration tests to test the correct template ID is called based on if the new version contains breaking changes or not.

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/ApiSubscriptionFunctionsTests.cs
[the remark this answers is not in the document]
mmoalin: I have added coverage of this new logic for the integration test thanks to Lee's help.

src/GovUk.Education.ExploreEducationStatistics.Notifier.Tests/Functions/ApiSubscriptionFunctionsTests.cs
```diff
@@ -553,20 +567,19 @@ public async Task Success()
                 Status = ApiSubscriptionStatus.Subscribed,
                 Expiry = null
             };
-
             await CreateApiSubscription(subscription);
-
+            var templates = GetGovUkNotifyOptions().EmailTemplates;
```
leeoades, line 571: Is this call needed anymore?

.gitignore
```diff
@@ -99,3 +99,4 @@ src/GovUk.Education.ExploreEducationStatistics.Admin/appsettings.IdpBootstrapUse
 
 ## IDE Files (e.g. local Visual Studio configuration)
 *.csproj.user
+/src/GovUk.Education.ExploreEducationStatistics.slnLaunch.user
```
leeoades, line 102: Might as well ignore all *.user files.

leeoades (approved):

PR #5607, opened 2025-02-14
EES-5860: Add handling for multiple filter items with the same label.

tomjonesdev:
This PR fixes an error when two or more filter items with the same label were found under different groups.

This is achieved by adding a condition to identify the filter's parent group, and then assigning it as the auto-selection filter item.

The PR also adds some basic validation to prevent a data set from being imported which defines a `filter_default` column as well as a filter group hierarchy, as this is not yet supported.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
@@ -1,10 +1,4 @@
 #nullable enable
-using System;
-using System.Collections.Generic;
-using System.Data;
-using System.IO;
-using System.Linq;
-using System.Threading.Tasks;
 using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Common.Utils;
@@ -18,6 +12,12 @@ using Microsoft.Data.SqlClient;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
+using System;
+using System.Collections.Generic;
+using System.Data;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
 using static System.StringComparison;
 
 namespace GovUk.Education.ExploreEducationStatistics.Data.Processor.Services
@@ -112,7 +112,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Processor.Services
 
             var filterAndIndicatorReader = new FilterAndIndicatorValuesReader(csvHeaders, subjectMeta);
             var fixedInformationReader = new FixedInformationDataFileReader(csvHeaders);
-            
+
             await CsvUtils.ForEachRow(dataFileStreamProvider, async (rowValues, index, _) =>
             {
                 if (index % Stage2RowCheck == 0)
@@ -129,7 +129,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Data.Processor.Services
 
                     await _dataImportService.UpdateStatus(dataImport.Id,
                         DataImportStatus.STAGE_2,
-                        (double) (index + 1) / dataImport.TotalRows!.Value * 100);
+                        (double)(index + 1) / dataImport.TotalRows!.Value * 100);
                 }
 
                 if (IsRowAllowed(soleGeographicLevel, rowValues, fixedInformationReader))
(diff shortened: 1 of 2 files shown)


src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
```diff
@@ -181,14 +181,27 @@ await _dataImportService.UpdateStatus(dataImport.Id,
...
-                            // If so, we set no autoSelectFilterItem.
-                            .SingleOrDefault();
+                                item.FilterGroup.FilterId == filterId &&
+                                item.Label.Equals(autoSelectFilterItemLabel, OrdinalIgnoreCase));
+
+                        // There might be two or more filter items with the same label under different groups
+                        if (filterItemsWithAutoSelectLabel.Count() > 1)
+                        {
+                            // If so, does one belong to a filter group with the same label?
+                            var filterItemWithMatchingGroupLabel = filterItemsWithAutoSelectLabel.SingleOrDefault(fi => fi.FilterGroup.Label == autoSelectFilterItemLabel);
```
ntsim, line 193: Not sure this is entirely correct. Think we're missing the following two cases with this logic:

1. The filter may only have one filter group (typically a 'Default' group but could be something else). If there is only a single filter group then we should be able to find the auto-select filter item as there's no ambiguity.

2. As we can now configure the auto-select via the `filter_default` column, the auto-select labels for the filter item and filter group may differ. For example, the filter item label could be 'All', but the filter group label might be 'Total'.

   For now, given we're preventing the filter hierarchies and filter groups being used together, we could just search for a filter group with a 'Total` label.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
```diff
@@ -181,14 +181,27 @@ await _dataImportService.UpdateStatus(dataImport.Id,
...
+                            if (filterItemWithMatchingGroupLabel is not null)
+                            {
+                                // If it does, this is the auto select filter item
+                                return new { filterItemWithMatchingGroupLabel.Id, filterItemWithMatchingGroupLabel.Label };
+                            }
+                        }
+
+                        return filterItemsWithAutoSelectLabel.Any()
+                            ? new { filterItemsWithAutoSelectLabel.First().Id, filterItemsWithAutoSelectLabel.First().Label }
+                            : null;
```
ntsim, line 204: Using `Any` like this actually has the following linter warning:

> CA1860: Prefer comparing 'Count' to 0 rather than using 'Any()', both for clarity and for performance

We're also doing a repeated `First` calls, which is a bit redundant too. Think we could tighten this up to the following:

```cs
return filterItemsWithAutoSelectLabel
    .Select(fi => new { fi.Id, fi.Label })
    .FirstOrDefault();
```

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
```diff
@@ -181,14 +181,27 @@ await _dataImportService.UpdateStatus(dataImport.Id,
...
+                        if (filterItemsWithAutoSelectLabel.Count() > 1)
+                        {
+                            // If so, does one belong to a filter group with the same label?
+                            var filterItemWithMatchingGroupLabel = filterItemsWithAutoSelectLabel.SingleOrDefault(fi => fi.FilterGroup.Label == autoSelectFilterItemLabel);
+
+                            if (filterItemWithMatchingGroupLabel is not null)
+                            {
+                                // If it does, this is the auto select filter item
+                                return new { filterItemWithMatchingGroupLabel.Id, filterItemWithMatchingGroupLabel.Label };
+                            }
```
ntsim, line 199: We need to return _something_ from the `if (filterItemsWithAutoSelectLabel.Count > 1)` block or we'll incorrectly fall through to the `filterItemsWithAutoSelectLabel == 0` default case which is almost certainly a bug.

This can be fixed with something like:

```cs
if (filterItemsWithAutoSelectLabel.Count > 1)
{
    return filterItemsWithAutoSelectLabel
        .Where(fi => fi.FilterGroup.Label == autoSelectFilterItemLabel)
        .Select(fi => new { fi.Id, fi.Label })
        .SingleOrDefault();
}
```

Note this above snippet doesn't take into account the two cases we also need to cover that I've mentioned in the earlier comment above.

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ImporterService.cs
```diff
@@ -181,14 +181,27 @@ await _dataImportService.UpdateStatus(dataImport.Id,
...
-                                item.FilterGroup.FilterId == filterId
-                                && item.Label.Equals(autoSelectFilterItemLabel, OrdinalIgnoreCase))
-                            .Select(item => new { item.Id , item.Label})
-                            // There might be two filter items with the same label under different groups.
-                            // If so, we set no autoSelectFilterItem.
-                            .SingleOrDefault();
+                                item.FilterGroup.FilterId == filterId &&
+                                item.Label.Equals(autoSelectFilterItemLabel, OrdinalIgnoreCase));
+
+                        // There might be two or more filter items with the same label under different groups
```
ntsim, line 189: Would mention that the current implementation only considers a filter groups and will not work for filter hierarchies for the time-being.

We should create a follow-up bug / task ticket to address this correctly and update the comment with a TODO referencing this ticket number e.g. `TODO EES-blah: Fix auto-select filter items for filter hierarchies`

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ValidatorService.cs
```diff
@@ -160,40 +160,53 @@ private async Task<Either<List<DataImportError>, Unit>> ValidateCsvFileType(
...
                 }
                 catch (Exception e)
                 {
                     errors.Add(new DataImportError($"error at data row {index + 1}: {MetaFileHasInvalidValues.GetEnumLabel()} : {e.Message}"));
                 }
             });
 
+            if (hasDefaultColumn && filterGroups > 1)
+            {
+                errors.Add(new DataImportError("The filter_default column cannot be used in conjunction with filter hierarchies"));
```
ntsim, line 193: Similar to earlier comment - would suggest leaving a comment here on why we're currently doing this and leave a TODO referencing the relevant ticket that should address this

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ValidatorService.cs
```diff
@@ -160,40 +160,53 @@ private async Task<Either<List<DataImportError>, Unit>> ValidateCsvFileType(
...
+                        filterGroups++;
+                    }
                 }
                 catch (Exception e)
                 {
                     errors.Add(new DataImportError($"error at data row {index + 1}: {MetaFileHasInvalidValues.GetEnumLabel()} : {e.Message}"));
                 }
             });
 
+            if (hasDefaultColumn && filterGroups > 1)
```
ntsim, line 191: Looks good to me at a surface level! Just want to double-check that we've tested this with a CSV that would be caught by this?

ntsim (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Services/ValidatorService.cs
tomjonesdev: Yep, I editing one of the test sets to meet this condition.

ntsim: Thinking about it a little bit more, not sure the current logic will be sufficient. 

We're currently just counting filter groups in a fairly simple way rather than figuring out if we actually have a hierarchy or not.

The quick and dirty way of doing that right now would be to simply gather all the non-empty values from `filter_grouping_column` into a set. We would then check this against all the column names for the filter meta rows.

If any of the filter group values have a matching filter meta row, we can infer that there is a filter hierarchy. The old filter grouping logic worked by *not* having explicit meta rows for the filter groups.

ntsim (approved):

PR #5626, opened 2025-02-25
EES-5882: Add Screener API response models.

tomjonesdev:
This PR adds some initial classes which will be implemented once the screener integration gets under way. These models are subject to change as the screener logic is still being developed, but give an idea of the structure as it stands.
image

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Models/DataScreenerTestResult.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Models/DataScreenerTestResult.cs
@@ -1,0 +1,37 @@
+#nullable enable
+using System.Collections.Generic;
+using System.Text.Json.Serialization;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Models;
+
+public class DataScreenerTestResult
+{
+    [JsonPropertyName("check")]
+    public required string TestFunctionName { get; set; }
+
+    [JsonPropertyName("result")]
+    public TestResult TestResult { get; set; }
+
+    // TODO (EES-5353): The screener currently returns a single string, so this list will fail to deserialise until the response has been updated
+    [JsonPropertyName("message")]
+    public List<string> Notes { get; set; } = [];
+
+    public Stage Stage { get; set; }
+}
+
+// TODO (EES-5353): Rename stages to be more descriptive
+public enum Stage
+{
+    InitialFileValidation = 1,
+    PreScreening1 = 2,
+    PreScreening2 = 3,
+    FullChecks = 4,
+    Passed = 5,
+}
+
+public enum TestResult
+{
+    PASS,
+    FAIL,
+    WARNING,
+}
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/DataScreenerRequest.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/DataScreenerRequest.cs
@@ -1,0 +1,16 @@
+#nullable enable
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Requests;
+
+public record DataScreenerRequest
+{
+    public required string StorageContainerName { get; set; }
+
+    public required string DataFileName { get; set; }
+
+    public required string MetaFileName { get; set; }
+
+    public required string DataFilePath { get; set; }
+
+    public required string MetaFilePath { get; set; }
+}
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Models/DataScreenerResponse.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Models/DataScreenerResponse.cs
@@ -1,0 +1,32 @@
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+using System.Text.Json.Serialization;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Models;
+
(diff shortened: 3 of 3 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Models/DataScreenerTestResult.cs
```diff
@@ -0,0 +1,37 @@
...
+public enum Stage
+{
+    InitialFileValidation = 1,
+    PreScreening1 = 2,
+    PreScreening2 = 3,
+    FullChecks = 4,
+    Passed = 5,
+}
+
+public enum TestResult
```
leeoades, line 32: It would be more usual for the c# code here to be pascal case, and if you want a different representation in the Json, to add an attribute on these with the alternative form.

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/DataScreenerRequest.cs
```diff
@@ -0,0 +1,16 @@
+#nullable enable
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Requests;
+
+public record DataScreenerRequest
+{
+    public required string StorageContainerName { get; set; }
+
+    public required string DataFileName { get; set; }
```
leeoades, line 9: I would suggest using the more conventional "Filename" casing?

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/DataScreenerRequest.cs
```diff
@@ -0,0 +1,16 @@
...
+
+    public required string MetaFileName { get; set; }
+
+    public required string DataFilePath { get; set; }
```
leeoades, line 13: Is there a reason that the path and the filename are separated? Perhaps just have a single full path to the file.

src/GovUk.Education.ExploreEducationStatistics.Admin/Models/DataFileInfo.cs
```diff
@@ -27,5 +27,8 @@ public record DataFileInfo : FileInfo
         public Guid? PublicApiDataSetId { get; set; }
 
         public string? PublicApiDataSetVersion { get; set; }
+
+        // TODO (EES-5353): Remove nullability once screener has been fully integrated
+        public DataScreenerResponse? ScreenerResponse { get; set; }
```
leeoades, line 32: I would recommend not making this nullable and instead introducing a static _DataScreenerResponse.None_ object. This will prevent the temporary nullability of this property from propagating through the consuming code.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ArchiveDataSetFileViewModel.cs
```diff
@@ -20,4 +21,7 @@ public record ArchiveDataSetFileViewModel
     public long MetaFileSize { get; set; }
 
     public Guid? ReplacingFileId { get; set; }
+
+    // TODO (EES-5353): Remove nullability once screener has been fully integrated
+    public DataScreenerResponse? ScreenerResponse { get; set; }
```
leeoades, line 26: Ditto the DataScreenerResponse.None concept here.

src/GovUk.Education.ExploreEducationStatistics.Admin/Models/DataFileInfo.cs
tomjonesdev: Tbh I was considering leaving the response classes alone to avoid complicating things/adding unused stuff - adding them now was mainly just to serve as a placeholder/reminder. I'm inclined to just remove the property.

src/GovUk.Education.ExploreEducationStatistics.Admin/Models/DataScreenerTestResult.cs
tomjonesdev: Agreed, and we use enum string converters elsewhere in the codebase, however this is the kind of thing I will ask to be changed in the screener's response so that we don't need to muck about with deserialisation logic - I decided to match with the current response for the time being as a reminder.

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/DataScreenerRequest.cs
tomjonesdev: There's a mixed treatment of this as two separate words vs. one across the codebase - I personally prefer the former so I went with that - but this class is the most likely to change during integration, so I'm not precious about it's code style at this stage.

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/DataScreenerRequest.cs
tomjonesdev: Partly because the screener endpoint isn't 100% defined yet, and partly because I know it uses the file names for messaging, and making it a separate property would avoid the need for string manipulation. As mentioned above though, this class is quite likely to change during development.

src/GovUk.Education.ExploreEducationStatistics.Admin/Models/DataFileInfo.cs
leeoades: Ah ok. I thought you were going to start consuming the response.

PR #5634, opened 2025-02-27
EES-5897 Create new parent_filter meta file column for filter hierarc…

mmyoungman:
…hies

This PR creates a separate column to meta file CSVs to indicate whether a particular filter has a parent filter for filter hierarchies (to be used in the table tool).

Previously we used the filter_grouping_column for the dual purpose of both creating filter groups for filter items *and* for indicating the parent filter. The problem is that even if you're using filter_grouping_column to indicate a parent filter, it is still used to group the columns as well.

To avoid any confusion over this, I decided it is probably best to separate them into two separate features - it wasn't simple to preventing the grouping behaviour if the filter_grouping_column was indicating a parent filter.

As far as I'm aware, the previous behaviour didn't break anything - this change is just to avoid confusion.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Models/MetaDataFileReader.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Models/MetaDataFileReader.cs
@@ -43,6 +43,7 @@ public class MetaDataFileReader
             ColumnType = Enum.Parse<ColumnType>(columnType!),
             Label = ReadMetaColumnValue(MetaColumns.label, rowValues),
             FilterGroupingColumn = ReadMetaColumnValue(MetaColumns.filter_grouping_column, rowValues),
+            ParentFilter = ReadMetaColumnValue(MetaColumns.parent_filter, rowValues),
             FilterHint = ReadMetaColumnValue(MetaColumns.filter_hint, rowValues),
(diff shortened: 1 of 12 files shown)


src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Models/MetaDataFileReader.cs
```diff
@@ -63,11 +64,16 @@ public MetaRow GetMetaRow(IReadOnlyList<string> rowValues)
...
+                    parentFilter: filterMetaRow.ParentFilter,
                     autoSelectFilterItemLabel: filterMetaRow.AutoSelectFilterItemLabel,
                     // NOTE: AutoSelectFilterItemId is set later, after filter items are created
                     subjectId: subject.Id),
                 column: filterMetaRow.ColumnName))
             .ToList();
+
+
+
+
```
benoutram, line 76: Minor one, looks like some extra blank lines got added in here.

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Data.Processor/Models/MetaDataFileReader.cs
mmyoungman: Removed

PR #5637, opened 2025-02-27
EES-5902 Update migration endpoint to set auto select for group Total…

mmyoungman:
… item Total filters

When the auto select work was initially deployed, we only set an auto select item for a filter if it only had one item labelled "Total". But this was incorrect - we also want to set the auto select item for filters where there are multiple totals, providing one of them belongs to the group "Total".

This change to the migration endpoint allows us to set these filter `AutoSelectFilterItemId`s to restore the previous behaviour from before the new auto select work was deployed.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/AutoSelectFilterItemMigrationController.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/AutoSelectFilterItemMigrationController.cs
@@ -16,7 +16,6 @@ public class AutoSelectFilterItemMigrationController(StatisticsDbContext statist
 {
     public class MigrationResult
     {
-        public bool IsDryRun;
         public int Processed;
     }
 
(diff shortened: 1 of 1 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Controllers/Api/AutoSelectFilterItemMigrationController.cs
```diff
@@ -28,11 +27,11 @@ public async Task<MigrationResult> AutoSelectFilterItemMigration(
...
-                && f.FilterGroups.Count(
-                    group => group.FilterItems.Count(
-                        item => item.Label == "Total" // this will be case agnostic (i.e. also find "total")
-                    ) == 1
-                ) == 1) // only one Total filter item across all filter items
+                // one group with label "Total" that contains one item with label "Total"
+                && f.FilterGroups.Count(group => group.Label == "Total"
+                                               && group.FilterItems.Count(item => item.Label == "Total")
+                                               == 1)
+                == 1)
```
benoutram, line 34: Don't we also want to set the auto select item for filters where there is a single filter item with label 'Total', regardless of what its group label is?

Or maybe this code assumes the previous version of the code has already done that, and that it's been successfully run in all environments and we don't need that.

mmyoungman: > Or maybe this code assumes the previous version of the code has already done that, and that it's been successfully run in all environments and we don't need that.

Yeah that's right. The previous migration already set the AutoSelectFilterItemId for Filters where there is only one "Total" FilterItem associated with a Filter. That was done in all environments in https://dfedigital.atlassian.net/browse/EES-5795

Just to be crystal clear about it, the original auto select work would only set an AutoSelectFilterItemId on a Filter if there was a sole "Total" FilterItem. This was changed in EES-5860 such that if there were mutliple "Total" FilterItems under a Filter, it would still set an AutoSelectFilterItem if a single one of those "Total" FilterItems had a "Total" FilterGroup. But that was only changed for newly imported data sets. This PR sets the auto select filter items of data sets with multiple "Total"s imported prior to these all changes.

benoutram (approved):

PR #5642, opened 2025-02-27
Ees 5802 add public api support to prod

duncan-at-hiveit:
This PR:
- enables the Public API in Prod, so that Admin, Publisher and the Public Site become aware of the Public API.
- adds an automated deploy stage to the Public API pipeline to allow an optional deploy stage to Prod following on from a successful deploy to Pre-Prod.  This is protected from automated rollout by manual approvals that have been set up in the Prod DevOps environment.
- adds a manual deploy stage to be able to deploy directly to Prod.  This'll most likely be removed shortly after getting Public API fully deployed and running on Prod.

### DIFF
--- a/infrastructure/parameters/prod.parameters.json
+++ b/infrastructure/parameters/prod.parameters.json
@@ -94,9 +94,6 @@
     },
     "maxStatsDbSizeBytes": {
       "value": 2199023255552
-    },
-    "publicDataDbExists": {
-      "value": false
     }
   }
 }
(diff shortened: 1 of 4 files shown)


N-moh (dismissed):

infrastructure/parameters/prod.parameters.json
```diff
@@ -94,9 +94,6 @@
     },
     "maxStatsDbSizeBytes": {
       "value": 2199023255552
-    },
-    "publicDataDbExists": {
-      "value": false
```
N-moh, line 99: Would it not work if we just changed the value to 'true' ?

duncan-at-hiveit: It would, but this also does the same thing and is consistent with the other environments :)  Next step will be to remove it entirely.

PR #5655, opened 2025-03-05
Ees 5826 analytics part 1

duncan-at-hiveit:
# Overview

This PR is part 1 in a series of PRs to capture analytics around the usage of the Public API.

This PR focuses on creating a storage account for analytics, and having the Public API Container App write queries to that analytics storage account.

Future PRs will handle the consumption of this raw data from the same analytics storage account and processing into a format that is useful to Cam, and which can then be fed into the Delta Lake.
 
## The design

When a query is issued against the Public API, the query is firstly executed and then captured for analytics.  The process involves taking the original query, "normalising" is as much as possible and then writing the query along with other metrics as a JSON file to the analytics file share.

The purpose of normalising queries is to take queries that are logically the same but potentially slightly different in their structuring, and identifying that they are in fact the same query under the hood.  Examples of this might be, for instance, nested criteria that are in a different order but the same in terms of what criteria they're attempting to apply, or lists of indicators that are the same values but just in different orders.

The reason that we want to ultimately normalise these queries is that it allows the analytics consuming code (in a future PR) to recognise that the same queries have been issued even when some fields are ordered slightly differently, for example.

To this end, pains have been taken to try to ensure that JSON is output in as consistent a way as possible as well, with the reported queries being output in camelCase and in alphabetical order of fields. 

A future PR will introduce a Function App that consumes these JSON files and presents reports in parquet format.

## Infrastructure changes

We are introducing a new storage account for services to use to write their raw analytics data to.  This is a file share-optimised account in order to provide the greatest performance for the reading / writing of small files.  Within this storage account is a file share which is being mounted onto the Public API Container App.  

## An example of what is generated for a query

Below is an example of what is captured for a successful query.   It includes details of the data set being queried, the access times, the results found etc.  It is also output in camelCase and with the JSON fields being ordered alphabetically.

```json
{
  "dataSetId": "59489501-541a-4576-afd8-a57aaf13d097",
  "dataSetTitle": "UI tests - Public API - restricted 20250227-165752-7x5qxm-0 - Subject 2",
  "dataSetVersion": "1.0.0",
  "dataSetVersionId": "59489501-711a-3875-90e2-2a94837b333f",
  "endTime": "2025-03-05T01:30:11.920Z",
  "query": {
    "criteria": {
      "filters": {
        "in": [
          "rQkNj"
        ]
      },
      "locations": {
        "in": [
          {
            "id": "IzBzg",
            "level": "NAT"
          }
        ]
      },
      "timePeriods": {
        "gt": {
          "code": "AY",
          "period": "2017/2018"
        }
      }
    },
    "debug": false,
    "indicators": [
      "eNuSW",
      "TrPJP"
    ],
    "page": 1,
    "pageSize": 1000
  },
  "resultsCount": 0,
  "startTime": "2025-03-05T01:30:11.901Z",
  "totalRowsCount": 0
}
```

## A screenshot of the results in Azure

image

## An example of raw data being output in Azure

```json
{
  "dataSetId": "2cad9401-c18d-5975-a0fb-0c35a00679b8",
  "dataSetTitle": "UI tests - Public API - preview token 20250128-134616-131d3s-0 - Subject 1",
  "dataSetVersion": "1.0.0",
  "dataSetVersionId": "2cad9401-ff8d-2a76-b2d4-ddaeb1f964ed",
  "endTime": "2025-03-05T14:39:51.948Z",
  "query": {
    "criteria": {
      "filters": {
        "eq": "jfu8A"
      },
      "locations": {
        "eq": {
          "id": "dP0Zw",
          "level": "NAT"
        }
      },
      "timePeriods": {}
    },
    "debug": false,
    "indicators": [
      "O722C"
    ],
    "page": 1,
    "pageSize": 1000
  },
  "resultsCount": 315,
  "startTime": "2025-03-05T14:39:51.622Z",
  "totalRowsCount": 315
}
```


leeoades (approved):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Services/AnalyticsServiceTests.cs
```diff
@@ -0,0 +1,66 @@
...
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests.Services;
+
+public abstract class AnalyticsServiceTests
+{
+    public class ReportDataSetVersionQuery : AnalyticsServiceTests
+    {
+        private const string SnapshotPrefix = $"{nameof(AnalyticsServiceTests)}.{nameof(ReportDataSetVersionQuery)}";
+        
+        [Fact]
+        public async Task Success()
```
leeoades, line 16: I would highly recommend that you use a mocked out filesystem. Something like: https://github.com/bkoelman/TestableFileSystem/tree/82219e7d2bf8967ee3176a5fa58c72809d489536

This has numerous advantages. There is an in-memory implementation of the filesystem which will not only see multiple factors of speed up in tests but also will not all foul of operating system differences etc.
Also generally, if you are writing a test that wants to ensure that the contents of a file are correct, you should not need to actually commit those contents to disk.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Services/AnalyticsServiceTests.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            var pathResolver = new TestAnalyticsPathResolver();
```
leeoades, line 18: Abstracting away the filesystem will also potentially remove the requirement for the IAnalyticsPathResolver. I've had a look at the implementation and it is using config (which is fine) but then overriding it by injecting IHostEnvironment. The config is already environment specific isn't it? I would not recommend this approach as it causes code to behave differently locally which makes troubleshooting difficult/confusing.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetQueryService.cs
```diff
@@ -120,17 +118,53 @@ private async Task<Either<ActionResult, DataSetVersion>> FindDataSetVersion(
...
+            RunQuery(
+                dataSetVersion: dataSetVersion,
+                query: query,
+                cancellationToken: cancellationToken,
+                baseCriteriaPath: baseCriteriaPath)
+            .OnSuccessDo(results =>
+            {
+                // Deliberately do not await this operation as we do not want to
+                // delay the return of the query to the end user.
+                _ = analyticsService.ReportDataSetVersionQuery(
```
leeoades, line 156: One should never not await a task. This is because exceptions thrown within a task are then only materialised by the garbage collector. This used to be a process ending event although I believe the .net runtime is more forgiving these days. However, it is not to be done. 

Instead, delegate the task to a thread-safe singleton service- have it added to some sort of queue - and have that deal with writing and handling errors etc. In addition, it could buffer them and perform bursts of writes if that is more efficient etc If this is going to be hammered a lot then also consider using a pool of objects that the singleton rents to this consumer, the consumer populates it and returns it. This will reduce the allocations.

leeoades (commented):
Some pointers for thought / further discussion.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Services/AnalyticsServiceTests.cs
duncan-at-hiveit: Hi @leeoades ,

I like the idea of this.  I've been mainly trying to stick to already-paved conventions that were introduced in the Public API workstream a while back, where we used the local filesystem and path resolvers to support test, dev and production environments.

I think that if we're considering making this change, it'd be good to get @benoutram 's opinion too as he was involved in the early days with Nick and will be familiar with the path resolvers and the strategy that was being applied here :+1:

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetQueryService.cs
```diff
@@ -120,17 +118,53 @@ private async Task<Either<ActionResult, DataSetVersion>> FindDataSetVersion(
...
+                    dataSetVersionId: dataSetVersion.Id,
+                    semVersion: dataSetVersion.SemVersion().ToString(),
+                    dataSetTitle: dataSetVersion.DataSet.Title,
+                    query: query,
+                    resultsCount: results.Results.Count,
+                    totalRowsCount: results.Paging.TotalResults,
+                    startTime: startTime,
+                    endTime: DateTime.UtcNow);
+            });
+    }
```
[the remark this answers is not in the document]
duncan-at-hiveit: Yeah I'd agree with you there.  I'd say that these changes are a bit beyond the scope of this particular piece of work as they're more Public API-specific, but I will raise a ticket to look at addressing this.

PR #5668, opened 2025-03-06
EES-5890 Add AI Search service and searchable document storage container

benoutram:
This PR adds new infrastructure for implementing site-wide search. It contains Bicep for the first resources to be deployed after #5628 which created an empty pipeline for deploying search infrastructure.

* Azure AI Search service - This will eventually be defined with a search data  source, index and indexers which will provide the search feature.
* Azure Storage account with Blob storage service - This will be used to store searchable documents to be indexed by the search service.


infrastructure/templates/search/application/searchService.bicep
```diff
@@ -20,26 +27,26 @@ resource vNet 'Microsoft.Network/virtualNetworks@2023-11-01' existing = {
...
 }
 
-module searchDocsStorageAccountModule '../../public-api/components/storageAccount.bicep' = {
-  name: 'searchDocsStorageAccountModuleDeploy'
+module searchStorageAccountModule '../../public-api/components/storageAccount.bicep' = {
+  name: 'searchStorageAccountModuleDeploy'
   params: {
     location: location
-    storageAccountName: resourceNames.search.searchDocsStorageAccount
+    storageAccountName: '${replace(resourcePrefix, '-', '')}${abbreviations.storageStorageAccounts}search'
```
leeoades, line 49: Will resourcePrefix really have a `-` in it? If so, do we not want to strip it out of line 38 too?

leeoades (approved):
Just the one query. Feel free to amend if needs be. Approved.

infrastructure/templates/search/application/searchService.bicep
leeoades: Also, resourcePrefix is what we're using as our "name" as such. When I read "prefix" I immediately thought it meant "we have a group of components and I want to prefix them all with something".

leeoades: Actually, if I want the prefix to be "lee-1-" then you'll strip out both the `-`s which I'm guessing you don't want to do.

benoutram: In general we want to use `-` as a separator. Most resources allow `-` in the name but storage accounts are an exception because they can only contain letters and numbers.

So we don't want to strip it out of line 38 but we want to strip it out of line 49.

https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftstorage

> When I read "prefix" I immediately thought it meant "we have a group of components and I want to prefix them all with something".

Yes that's the intention here, that we want to prefix our resource names with this, albeit with the exception for `-` in strorge account names.

PR #5671, opened 2025-03-06
EES-5899 Add HealthCheck function to Search Function App

benoutram:
This PR adds a basic `/api/HealthCheck` function to the Search Function App to be used with the Azure Function infrastructure deployment and for checking the resource health in the Azure Portal.

When healthy it returns HTTP response code 200 OK with content:

```json
{"healthy": true}
```

It currently makes no checks but we could improve on this, for example by checking the Azure Storage account blob container for searchable documents exists.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/HealthCheckFunction.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/HealthCheckFunction.cs
@@ -1,0 +1,31 @@
+﻿using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Azure.Functions.Worker;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Functions;
+
+public class HealthCheckFunction
+{
+    [Function(nameof(HealthCheck))]
+    [Produces("application/json")]
+    public IActionResult HealthCheck(
+#pragma warning disable IDE0060
+        [HttpTrigger(AuthorizationLevel.Anonymous, "get")]
+        HttpRequest request)
+#pragma warning restore IDE0060
+    {
+        var healthCheckResponse = new HealthCheckResponse();
+
+        if (healthCheckResponse.Healthy)
+        {
+            return new OkObjectResult(healthCheckResponse);
+        }
+
+        return new ObjectResult(healthCheckResponse) { StatusCode = StatusCodes.Status500InternalServerError };
+    }
+}
+
+public record HealthCheckResponse
+{
+    public bool Healthy => true;
+}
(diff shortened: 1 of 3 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/HealthCheckFunction.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Azure.Functions.Worker;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Functions;
```
benoutram, line 5: @leeoades Question: I wasn't sure if you think I should follow the same pattern as the other function and place this in `FunctionApp.Functions.HealthCheckFunction`?

leeoades: I suppose it would be neater and more consistent for it to be in its own directory. The response could be in its own file then too.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/HealthCheckFunction.cs
```diff
@@ -0,0 +1,31 @@
...
+#pragma warning restore IDE0060
+    {
+        var healthCheckResponse = new HealthCheckResponse();
+
+        if (healthCheckResponse.Healthy)
+        {
+            return new OkObjectResult(healthCheckResponse);
+        }
+
+        return new ObjectResult(healthCheckResponse) { StatusCode = StatusCodes.Status500InternalServerError };
```
leeoades, line 24: You're just checking whether returning a 500 is standard practice. Personally, I would want to return a code that was different but whatever the standard practice is.

leeoades (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/HealthCheckFunction.cs
benoutram: We discussed this over Slack. The docs say

>The Health check path should check critical components of your application. For example, if your application depends on a database and a messaging system, the Health check endpoint should connect to those components. If the application can't connect to a critical component, then the path should return a 500-level response code to indicate the app is unhealthy. Also, if the path doesn't return a response within one minute, the health check ping is considered unhealthy.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/HealthCheckFunction.cs
benoutram: I've done this now

So there's now

- `Search.FunctionApp.Functions.HealthChecks.HealthCheckFunction`
- `Search.FunctionApp.Functions.HealthChecks.Dtos.HealthCheckResponseDto`

PR #5674, opened 2025-03-10
EES-5734 copy download charts

rianski:
Changes made:

Added the ability to export chart as PNG or copy charts to clipboard

export chart new


src/explore-education-statistics-common/src/components/ExportButtonMenu.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  const [getDivJpeg] = useGenerateImage({
+    quality: 0.8,
+    type: 'image/jpeg',
+  });
+
+  const handleDivDownload = useCallback(async () => {
+    const jpeg = await getDivJpeg();
+    if (jpeg) {
+      downloadFile(jpeg, 'Chart.png');
```
mmyoungman, line 20: The file seem to be saving as a jpeg, not a png. Assuming there isn't any hard requirement on the image type, I'd rename the extension, variable names, and any link text to JPEG. Or maybe the mime type in `useGenerateImage` can be changed to `image/png`.

src/explore-education-statistics-common/src/modules/find-statistics/components/DataBlockTabs.tsx
```diff
@@ -81,11 +84,13 @@ const DataBlockTabs = ({
             {fullTable && (
               <ErrorBoundary fallback={errorMessage}>
                 {chart && (
-                  <ChartRenderer
-                    id="dataBlockTabs-chart"
-                    source={dataBlock.source}
-                    chart={chart}
-                  />
+                  <ExportButtonContext.Provider value={exportRef}>
+                    <ChartRenderer
```
mmyoungman, line 88: `ChartRenderer` is also used in `DataBlockPageReadOnlyTabs.tsx` - this is used in Admin to view a data block for a release that has been published.

This "View block" link:
image

mmyoungman (dismissed):
Thanks to Guy for help with the review

src/explore-education-statistics-common/src/components/ExportButtonMenu.tsx
```diff
@@ -0,0 +1,95 @@
...
+      document.addEventListener('mousedown', handleClickOutside);
+    }
+
+    return () => {
+      document.removeEventListener('mousedown', handleClickOutside);
+    };
+  }, [isOpen]);
+
+  return (
+    <div className={styles.exportDropdown}>
```
mmyoungman, line 63: Just checking that this includes everything we need for accessibility generally? Might be worth someone who knows these things looking it over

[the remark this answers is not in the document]
rianski: I've had a chat with Cam and we've agreed it's better where it is now although I'm going to move it to the right of the container to make it more visible. Worth mentioning the export stuff wouldn't work where the designs are as the tabs component is shared and used throughout the site for other tabs etc. You'd just keep losing the reference to the chart when you change tabs then the export buttons won't work

src/explore-education-statistics-common/src/components/ExportButtonMenu.tsx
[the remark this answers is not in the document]
rianski: Yer I was thinking that but just decided to work off the AC, will have a word with Cam on it as I'm not really sure what value we'd use for it

rianski: it actually always saves as a PNG so I'm not even sure why the library had it named that way. Will rename it so it's obvious

src/explore-education-statistics-common/src/modules/find-statistics/components/DataBlockTabs.tsx
rianski: Yeah though the AC hadn't specified if this was only supposed to be in the public site so I assumed it'd be both - another one for Cam

Guy-HiveIT:
I looked through this with Mark but since then a couple of things have jumped out:

1. The popover menu needs some accessibility improvements, at least:

- being able to close it with 'Esc'
- various `aria-` markup being added such as `aria-expanded` on the toggle button
- focus handling such as returning focus to the toggle when it is closed

Depending on the required browser support we could look to use the [popover api](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover)? This would at least [handle some of the above for us](https://hidde.blog/popover-accessibility/) such as light dismiss, focus handling.
This would also remove the need for the ref and 'mousedown' listeners for closing on outside click.


2. I think we need some loading and success / failure states adding to the UI and code. At the minute there is no indication the chart has been copied to clipboard, and downloading can take a short while, so could be made more obvious when it's happening and has happened or failed. Not sure if that needs more discussion with design.


Obviously I'm new to the project so if the above not relevant feel free to ignore!

src/explore-education-statistics-common/src/components/ExportButtonMenu.tsx
mmyoungman: I checked the type of the images I downloaded locally and they were jpegs

image

src/explore-education-statistics-common/src/modules/find-statistics/components/DataBlockTabs.tsx
rianski: Cleared this one with Cam he wants them on both ends so this is fine

rianski:
> I looked through this with Mark but since then a couple of things have jumped out:
> 
> 1. The popover menu needs some accessibility improvements, at least:
> 
> * being able to close it with 'Esc'
> * various `aria-` markup being added such as `aria-expanded` on the toggle button
> * focus handling such as returning focus to the toggle when it is closed
> 
> Depending on the required browser support we could look to use the [popover api](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover)? This would at least [handle some of the above for us](https://hidde.blog/popover-accessibility/) such as light dismiss, focus handling. This would also remove the need for the ref and 'mousedown' listeners for closing on outside click.
> 
> 2. I think we need some loading and success / failure states adding to the UI and code. At the minute there is no indication the chart has been copied to clipboard, and downloading can take a short while, so could be made more obvious when it's happening and has happened or failed. Not sure if that needs more discussion with design.
> 
> Obviously I'm new to the project so if the above not relevant feel free to ignore!

src/explore-education-statistics-common/src/components/ChartsToPNG.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  return [
+    generateImage,
+    {
+      isLoading,
+    },
+  ];
+}
+
+export type UseCurrentPng = [
```
amyb-hiveit, line 55: I don't think this is used anywhere else so doesn't need to be exported? Also, having a `UseCurrentPng` and `useCurrentPng` could be easily confused, so could call it something like `UseCurrentPngType`.

src/explore-education-statistics-common/src/components/ChartsToPNG.tsx
```diff
@@ -0,0 +1,106 @@
...
+  (callback?: BlobCallback) => Promise<string | undefined>,
+  {
+    isLoading: boolean;
+    ref: React.MutableRefObject<any>;
```
amyb-hiveit, line 59: I think this can be `React.MutableRefObject<HTMLDivElement | null>;` so we don't need to use `any`

src/explore-education-statistics-common/src/components/ChartsToPNG.tsx
```diff
@@ -0,0 +1,106 @@
+/* eslint-disable import/no-extraneous-dependencies */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { useExportButtonContext } from '@common/contexts/ExportButtonContext';
+import html2canvas, { Options as HTML2CanvasOptions } from 'html2canvas';
+import { useCallback, useState } from 'react';
+
+export type UseGenerateImage<T extends HTMLElement = HTMLDivElement> = [
```
amyb-hiveit, line 7: As with `UseCurrentPng` this doesn't need to be exported and might be better named `UseGenerateImageType`.

Is the `T` doing anything? My editor says it's declared but not read.

src/explore-education-statistics-common/src/components/ChartsToPNG.tsx
```diff
@@ -0,0 +1,106 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
amyb-hiveit, line 1: I don't think this is needed (seems fine if I remove it).

src/explore-education-statistics-common/src/components/ExportButtonMenu.module.scss
```diff
@@ -0,0 +1,23 @@
+@import '~govuk-frontend/dist/govuk/base';
+
+.exportDropdown {
```
amyb-hiveit, line 3: I think the linter will probably object to the css rules not being in alphabetical order.

src/explore-education-statistics-common/src/components/ChartsToPNG.tsx
```diff
@@ -0,0 +1,106 @@
...
+export function useGenerateImage(
+  args?: UseGenerateImageArgs,
+): UseGenerateImage {
+  const [isLoading, setIsLoading] = useState(false);
+  const ref = useExportButtonContext();
+
+  const generateImage = useCallback(
+    // eslint-disable-next-line consistent-return
+    async (callback?: BlobCallback) => {
+      if (ref && ref.current) {
```
amyb-hiveit, line 29: Is the check for `ref` needed here? Could it just be `ref?.current`?

src/explore-education-statistics-common/src/components/ChartsToPNG.tsx
```diff
@@ -0,0 +1,106 @@
...
+  options?: Partial<HTML2CanvasOptions>,
+): UseCurrentPng {
+  const ref = useExportButtonContext();
+
+  const [isLoading, setIsLoading] = useState(false);
+
+  const getPng = useCallback(
+    // eslint-disable-next-line consistent-return
+    async (callback?: BlobCallback) => {
+      if (ref !== null && ref?.current) {
```
amyb-hiveit, line 73: As above, is the check for `ref` needed here? Could it just be `ref?.current`?

src/explore-education-statistics-common/src/components/ChartsToPNG.tsx
```diff
@@ -0,0 +1,106 @@
...
+export type UseGenerateImageArgs = {
+  options?: HTML2CanvasOptions;
+  quality?: number;
+  type?: string;
+};
+
+export function useGenerateImage(
+  args?: UseGenerateImageArgs,
+): UseGenerateImage {
+  const [isLoading, setIsLoading] = useState(false);
```
amyb-hiveit, line 23: Does anything use this `isLoading` flag? Same question for `useCurrentPng`

src/explore-education-statistics-common/src/components/ChartsToPNG.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const ref = useExportButtonContext();
+
+  const generateImage = useCallback(
+    // eslint-disable-next-line consistent-return
+    async (callback?: BlobCallback) => {
+      if (ref && ref.current) {
+        setIsLoading(true);
+
+        return html2canvas(ref.current, {
```
amyb-hiveit, line 32: Could this be refactored to use `await` instead of promises? I think we generally prefer async/await over promises, though admittedly it's not in the style guide so up to you. Same question for `useCurrentPng`

src/explore-education-statistics-common/src/components/ChartsToPNG.tsx
```diff
@@ -0,0 +1,106 @@
...
+  },
+];
+
+export type UseGenerateImageArgs = {
+  options?: HTML2CanvasOptions;
+  quality?: number;
+  type?: string;
+};
+
+export function useGenerateImage(
```
amyb-hiveit, line 20: I'm a bit confused about why we need `useGenerateImage` and `useCurrentPng`, they seem to do the same thing but one has type and quality args and the other has those hardcoded? Also, if they are both needed could you improve the naming so it's clearer what they're for, something like `useGenerateImageForDownload` and `useGenerateImageForClipboard`.

src/explore-education-statistics-common/src/components/ExportButtonMenu.tsx
```diff
@@ -0,0 +1,95 @@
+import React, { useState, useCallback, useEffect, useRef } from 'react';
+import downloadFile from '@common/utils/file/downloadFile';
+import { useCurrentPng, useGenerateImage } from './ChartsToPNG';
+import ButtonText from './ButtonText';
+import styles from './ExportButtonMenu.module.scss';
+
+const ExportButtonMenu = () => {
```
amyb-hiveit, line 7: This is very specific to charts so should probably be in `charts/components` rather than the generic components folder. A more specific name would be better as well, something like `ChartExportButton`.

src/explore-education-statistics-common/src/components/ExportButtonMenu.tsx
```diff
@@ -0,0 +1,95 @@
...
+          aria-label="Export options"
+        >
+          Export
+        </ButtonText>
+        {isOpen && (
+          <div className={styles.exportMenu}>
+            <ButtonText onClick={handleDivDownload}>
+              Download chart as PNG image
+            </ButtonText>
+            <br />
```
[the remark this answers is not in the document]
amyb-hiveit: Could put them in a list instead?

src/explore-education-statistics-common/src/components/ExportButtonMenu.tsx
amyb-hiveit, line 63: Instead of using custom stuff for the dropdown you could use the `Details` component that we use on key stats for the guidance text, see https://github.com/dfe-analytical-services/explore-education-statistics/blob/dev/src/explore-education-statistics-common/src/modules/find-statistics/components/KeyStat.tsx. It's not the prettiest thing, but we'd get the build in accessibility and the arrow would make it clearer that it'll open a menu.

Screenshot 2025-03-14 095455

If a page has multiple charts then just having the button text as 'Export' will be confusing as it won't be clear what it refers to. You could add some visually hidden text so they'd read 'Export options for My chart title'. In the GDS design system it's generally preferred to use hidden text rather than `aria-label`, so that could be removed.

I'd also have the visible text be 'Export options' rather than just 'Export' as that also makes it more obvious that it'll open a menu.

Also I wonder if it might look better if it was under the chart title and subtitle, it looks a bit odd at the top?

src/explore-education-statistics-common/src/components/ExportButtonMenu.tsx
```diff
@@ -0,0 +1,95 @@
+import React, { useState, useCallback, useEffect, useRef } from 'react';
+import downloadFile from '@common/utils/file/downloadFile';
+import { useCurrentPng, useGenerateImage } from './ChartsToPNG';
+import ButtonText from './ButtonText';
+import styles from './ExportButtonMenu.module.scss';
+
+const ExportButtonMenu = () => {
+  const [isOpen, setIsOpen] = useState(false);
```
amyb-hiveit, line 8: We have a helpful little `useToggle` hook for toggling booleans that you could use here if you preferred.

src/explore-education-statistics-common/src/modules/find-statistics/components/KeyStat.tsx
```diff
@@ -47,7 +47,7 @@ const KeyStat = ({
   trend,
 }: KeyStatProps) => {
   const body = (
-    <>
+    <div>
```
amyb-hiveit, line 50: Why this change? There aren't any other changes to key stats here.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartBuilderTabSection.tsx
```diff
@@ -115,17 +116,21 @@ const ChartBuilderTabSection = ({
     [onTableUpdate, query, releaseVersionId],
   );
 
+  const exportRef = useRef(null);
```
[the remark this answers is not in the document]
amyb-hiveit: I agree it could be clearer, maybe `chartExportRef`? Same for on the frontend version.

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartBuilderTabSection.tsx
```diff
@@ -115,17 +116,21 @@ const ChartBuilderTabSection = ({
...
-      releaseVersionId={releaseVersionId}
-      data={table.results}
-      meta={meta}
-      initialChart={dataBlock.charts[0]}
-      tableTitle={dataBlock.heading}
-      onChartSave={handleChartSave}
-      onChartDelete={handleChartDelete}
-      onTableQueryUpdate={handleTableQueryUpdate}
-    />
+    <ExportButtonContext.Provider value={exportRef}>
```
amyb-hiveit, line 122: From what I remember of our conversation the context is needed for when the export key stats and table options are added, is that right?

src/explore-education-statistics-common/src/components/ExportButtonMenu.tsx
mmyoungman: I've no idea - probably one for Amy

src/explore-education-statistics-common/src/components/ChartsToPNG.tsx
rianski: Not currently no it was just an optional prop provided by the library we've taken the code from

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartBuilderTabSection.tsx
rianski: Yeah this is going to be done in a separate ticket now (straight after this one) so things may move slightly for that. Cam just wants the chart stuff through for now so it can be used

src/explore-education-statistics-common/src/components/ExportButtonMenu.tsx
rianski: I did consider using the Details component initially as it's the only thing even close to what we want in the style guide. Though I've just been going off the the Figma designs so not really sure what's preferred for this

src/explore-education-statistics-common/src/components/ExportButtonMenu.tsx
```diff
@@ -0,0 +1,95 @@
...
+    const handleClickOutside = (event: MouseEvent) => {
+      if (
+        dropdownRef.current &&
+        !dropdownRef.current.contains(event.target as Node)
+      ) {
+        setIsOpen(false);
+      }
+    };
+
+    if (isOpen) {
```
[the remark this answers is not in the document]
amyb-hiveit: Now it's using the details element escape closes it.

src/explore-education-statistics-common/src/charts/components/ChartExportButton.module.scss
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  :global(.govuk-details__summary) {
+    margin-bottom: 0;
+    max-width: calc(100% - 1rem);
+    overflow: hidden;
+    padding-left: 16px;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
```
amyb-hiveit, line 35: I'm not sure any of these styles are needed. We control what's in the list so it's unlikely we'll need to handle overflowing text, and generally better to let it wrap unless there's a good reason not to.

src/explore-education-statistics-common/src/charts/components/ChartExportButton.module.scss
```diff
@@ -0,0 +1,52 @@
...
+
+  :global(.govuk-details__text) {
```
amyb-hiveit, line 37: I don't think box-sizing, max-height, overflow and text align are really doing anything here.

src/explore-education-statistics-common/src/charts/components/ChartExportButton.module.scss
```diff
@@ -0,0 +1,52 @@
...
+
+.exportCloseButton {
+  margin-top: govuk-spacing(6);
+}
+
+.exportMenuList {
+  list-style-type: none;
+  margin: 0;
+  padding: 0;
+  width: max-content;
```
amyb-hiveit, line 11: Don't think this width is necessary.

src/explore-education-statistics-common/src/charts/components/ChartExportButton.module.scss
```diff
@@ -0,0 +1,52 @@
...
+    max-height: 40vh;
+    overflow: auto;
+    position: absolute;
+    text-align: left;
+    z-index: 2;
+  }
+}
+
+.exportMenu button {
+  padding: 2px 15px;
```
amyb-hiveit, line 51: Could use govuk-spacing here?

src/explore-education-statistics-common/src/components/ExportButtonMenu.tsx
amyb-hiveit: Using the details element should cover it accessibility-wise.

amyb-hiveit (changes requested):

src/explore-education-statistics-common/src/components/ChartsToPNG.tsx
rianski: If we ever have any sort of toast/alert wanting on it then it'd be useful. Happy to remove it as I feel toast messages aren't really a thing in the gov design

src/explore-education-statistics-common/src/charts/components/ChartExportButton.module.scss
rianski: without width

Not disagreeing but for reference this is how it looks without max-content width

bennettstuart (changes requested):
I've left a couple of comments around ExportButtonContext.
Also looks like there's some other unresolved change requests

src/explore-education-statistics-common/src/contexts/ExportButtonContext.tsx
```diff
@@ -0,0 +1,12 @@
+import { createContext, useContext } from 'react';
+
+export const ExportButtonContext =
+  createContext<React.RefObject<HTMLDivElement> | null>(null);
+
```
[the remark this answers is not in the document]
bennettstuart: on second thought, this is just a ref context isn't it, it's not really specific to the export button, and the name of it is confusing as it references a div instead of a button
`RefContext` might be a better name

src/explore-education-statistics-common/src/charts/components/ChartExportButton.module.scss
```diff
@@ -0,0 +1,52 @@
...
+
+.exportMenu {
+  background: govuk-colour('white');
+  display: grid;
+  font-size: 16px !important;
+  justify-content: flex-end;
+  margin: 0;
+  padding: 0.2rem 0 0.5rem 0.5rem;
+  position: relative;
+  text-align: left;
```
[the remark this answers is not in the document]
rianski: Just saw this comment it was hidden. So you're right the text-align isn't needed but the position is, as it keeps it within the chart container (solves the smaller screen issue now after a few changes).

position relative

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartBuilderTabSection.tsx
```diff
@@ -126,17 +127,21 @@ const ChartBuilderTabSection = ({
...
-      releaseVersionId={releaseVersionId}
-      data={table.results}
-      meta={meta}
-      initialChart={dataBlock.charts[0]}
-      tableTitle={dataBlock.heading}
-      onChartSave={handleChartSave}
-      onChartDelete={handleChartDelete}
-      onTableQueryUpdate={handleTableQueryUpdate}
-    />
+    <RefContext.Provider value={chartExportRef}>
```
bennettstuart, line 133: We can use `<RefContextProvider>` here instead, and then we don't need to instantiate a useRef ourselves. please can you do the same for the tests too, anywhere that we're using RefContext.Provider

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartBuilderTabSection.tsx
```diff
@@ -126,17 +127,21 @@ const ChartBuilderTabSection = ({
     [onTableUpdate, query, releaseVersionId, dataBlock],
   );
 
+  const chartExportRef = useRef(null);
```
bennettstuart, line 130: no need for this now with `RefContextProvider` creating a ref for us

bennettstuart (changes requested):

bennettstuart (approved):

PR #5675, opened 2025-03-10
EES-5828 - add analytics consumer

duncan-at-hiveit:
This PR introduces the 2nd part of the Analytics work for the Public API (and in future for other, non-Public API-related sources).

In Part 1 (https://github.com/dfe-analytical-services/explore-education-statistics/pull/5655), we added the writing out of Public API queries to an Analytics-specific file share from the Public API Container App.  Successful queries would be written out in a JSON file along with some other information about the query results, the time taken to issue the query etc.

Part 2 (this PR) involves consuming that raw data and converting it into a format that can be used conveniently in DfE's Delta Lake.  In this case, we decided on Parquet files, using the following schema:

Distinct query details (output in a `<timestamp>_public-api-queries.parquet` file):
```
When I talk about "logically distinct queries" or "logical structure" below, I refer to the fact that in Part 1 where we output the Public API queries to the file share, we attempted to "normalise" the queries as much as possible so that the query JSON for 2 different queries that were using the same criteria, filters, indicators, sorts, pagination and debug options (and therefore logically trying to query for the same set of results) would be output in identical format.  

QueryVersionHash - an MD5 hash of the query JSON concatenated with its DataSetVersionId.  This therefore represents a unique logical query along with the exact DataSetVersion that it was issued for.  This makes the distinction between a query that was exactly the same logical structure but was issued against 2 or more distinct DataSetVersions e.g. a query that was issued against a 1.0 version and then a 1.1 version of the same data set.

QueryHash - an MD5 hash of the query JSON.  This represents logically distinct queries but the same hash can apply to the same query but issued against different DataSetVersions.

Query - the query JSON itself.

DataSetId

DataSetVersionId

DataSetVersion - semver string.

DataSetTitle

ResultsCount - number of results returned in this particular query.

TotalRowsCount - the total number of (unpaginated) results available from this query.

QueryExecutions - the number of times this distinct query was executed against this particular DataSetVersion.
```

Distinct query access details (output in a `<timestamp>_public-api-query_access.parquet` file):
```
QueryVersionHash - as above.  This effectively acts as a foreign key to the rows in the file output above.  This lets us identify when each of the distinct queries captured in the first file were accessed.

StartTime - a UTC timestamp of then the start of the query execution occurred.

EndTime - a UTC timestamp of then the start of the query execution occurred.

DurationMillis - the time taken in milliseconds to execute the query.
```

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer.Tests/Functions/ConsumePublicApiQueriesFunctionTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer.Tests/Functions/ConsumePublicApiQueriesFunctionTests.cs
@@ -1,0 +1,407 @@
+using System.Reflection;
+using GovUk.Education.ExploreEducationStatistics.Analytics.Requests.Consumer.Functions;
+using GovUk.Education.ExploreEducationStatistics.Common.DuckDb.DuckDb;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
+using InterpolatedSql.Dapper;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Extensions.Logging;
+using Moq;
+using Xunit;
+
+namespace GovUk.Education.ExploreEducationStatistics.Analytics.Consumer.Tests.Functions;
+
+public abstract class ConsumePublicApiQueriesFunctionTests
+{
+    private readonly string _queryResourcesPath = Path.Combine(
+        Assembly.GetExecutingAssembly().GetDirectoryPath(),
+        "Resources",
+        "PublicApiQueries");
+
+    private readonly TestAnalyticsPathResolver _pathResolver = new();
+    
+    public class FunctionTests : ConsumePublicApiQueriesFunctionTests
+    {
+        [Fact]
+        public async Task NoSourceFolder()
+        {
+            var function = BuildFunction();
+            await function.Run(new TimerInfo());
+            
+            Assert.False(Directory.Exists(_pathResolver.PublicApiQueriesProcessingDirectoryPath()));
+            Assert.False(Directory.Exists(_pathResolver.PublicApiQueriesReportsDirectoryPath()));
+        }
+        
+        [Fact]
+        public async Task NoSourceQueriesToConsume()
+        {
+            Directory.CreateDirectory(_pathResolver.PublicApiQueriesDirectoryPath());
+
+            var function = BuildFunction();
+            await function.Run(new TimerInfo());
+            
+            // Check that as there were no files to process, no working directories were
+            // created as a result.
+            Assert.False(Directory.Exists(_pathResolver.PublicApiQueriesProcessingDirectoryPath()));
+            Assert.False(Directory.Exists(_pathResolver.PublicApiQueriesReportsDirectoryPath()));
+        }
+
+        [Fact]
+        public async Task SingleSourceQuery()
+        {
+            SetupQueryRequest("Query1Request.json");
+
+            var function = BuildFunction();
+            await function.Run(new TimerInfo());
+
+            Assert.False(Directory.Exists(_pathResolver.PublicApiQueriesProcessingDirectoryPath()));
+            Assert.True(Directory.Exists(_pathResolver.PublicApiQueriesReportsDirectoryPath()));
+
+            var reports = Directory.GetFiles(_pathResolver.PublicApiQueriesReportsDirectoryPath());
+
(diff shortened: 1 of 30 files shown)


useful-scripts/start.ts
benoutram, line 1: I don't think you mean to include the couple of blocks with references to `searchFunctionApp` here! We will also need to fix the port conflict as we've configured the search function app to use port 7075.

duncan-at-hiveit: Oops ta!

duncan-at-hiveit: I've sorted that out now thanks @benoutram

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Functions/ConsumePublicApiQueriesFunction.cs
```diff
@@ -0,0 +1,129 @@
...
+        
+        logger.LogInformation("Found {Count} data set queries to process", filesToProcess.Count);
+
+        var processingDirectory = pathResolver.PublicApiQueriesProcessingDirectoryPath();
+        var reportsDirectory = pathResolver.PublicApiQueriesReportsDirectoryPath();
+        
+        Directory.CreateDirectory(processingDirectory);
+        Directory.CreateDirectory(reportsDirectory);
+
+        Parallel.ForEach(filesToProcess, file =>
```
mmyoungman, line 48: I assume this is an async alternative to `filesToProcess.ForEach`? Surprised there is no await needed here. Good that it is self-contained.

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Functions/ConsumePublicApiQueriesFunction.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        duckDbConnection.ExecuteNonQuery($@"
+            COPY (SELECT * FROM QueryReport)
+            TO '{queryReportFilename}' (FORMAT 'parquet', CODEC 'zstd')");
+        
+        duckDbConnection.ExecuteNonQuery($@"
+            COPY (SELECT * FROM QueryAccessReport)
+            TO '{queryAccessReportFilename}' (FORMAT 'parquet', CODEC 'zstd')");
+        
+        Directory.Delete(processingDirectory, recursive: true);
```
mmyoungman, line 125: I think we may be deleting json query files that were created in between the `var filesToProcess =` call and this recursive delete, which wouldn't have been processed?

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Functions/ConsumePublicApiQueriesFunction.cs
duncan-at-hiveit: Yeah so it does all of this work in parallel so you'll have bulk moving of files concurrently, but the operation itself blocks, hence it has no await.

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Functions/ConsumePublicApiQueriesFunction.cs
duncan-at-hiveit: So when this function gets executed, it moves all files it's interested in working on into the `processing` folder.  When it's done processing them all into the parquet files, it deletes the `processing` folder.

This means that if any new queries were introduced in the time between this function executing and finishing, it will not remove any files that it's not processed - only ones that *have* been processed.

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer.Tests/Functions/ConsumePublicApiQueriesFunctionTests.cs
```diff
@@ -0,0 +1,407 @@
...
+        private static async Task<List<QueryReportLine>> ReadQueryReport(DuckDbConnection duckDbConnection, string queryReportFile)
+        {
+            var queryReportRows = (await duckDbConnection
+                    .SqlBuilder($"SELECT * FROM read_parquet('{queryReportFile:raw}')")
+                    .QueryAsync<QueryReportLine>())
+                .ToList();
+            return queryReportRows;
+        }
+
+        // TODO test for pagination and debug
```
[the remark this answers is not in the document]
duncan-at-hiveit: Thanks, removed!

mmyoungman (approved):

PR #5676, opened 2025-03-10
EES-5908 enhance healthchecks

leeoades:
New Search Function App
- Added healthchecks for Content API and Azure Blob Storage
- Added logging to Azure Blob Storage Client to log issues when uploading - in particular, metadata problems.


src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/HealthChecks/Strategies/AzureBlobStorageHealthCheckStrategy.cs
```diff
@@ -0,0 +1,37 @@
...
+    IOptions<AppOptions> appOptions) : IHeathCheckStrategy
+{
+    public async Task<HeathCheckResult> Run(CancellationToken cancellationToken)
+    {
+        var containerName = appOptions.Value.SearchableDocumentsContainerName;
+        if (string.IsNullOrWhiteSpace(containerName))
+        {
+            return new HeathCheckResult(false,
+                $"Azure blob storage container name is not specified in {AppOptions.Section}.{nameof(AppOptions.SearchableDocumentsContainerName)}");
+        }
```
benoutram, line 18: Just wondering what you think to removing the validation in the host startup now or do you think we should keep both?

In my experience it will be quite a lot more time consuming to find the right location to view the startup logs and diagnose a startup failure, than to query the healthcheck response.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Clients/ContentApi/IContentApiClient.cs
```diff
@@ -1,4 +1,5 @@
-﻿using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Domain;
+﻿using System.Diagnostics.CodeAnalysis;
```
benoutram, line 1: I think there's an unused import here!

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/HealthChecks/HealthCheckFunction.cs
```diff
@@ -1,27 +1,30 @@
...
     [Function(nameof(HealthCheck))]
     [Produces("application/json")]
-    public IActionResult HealthCheck(
-#pragma warning disable IDE0060
+    public async Task<IActionResult> HealthCheck(
         [HttpTrigger(AuthorizationLevel.Anonymous, "get")]
-        HttpRequest request)
-#pragma warning restore IDE0060
+        HttpRequest _,
+        CancellationToken cancellationToken)
```
benoutram, line 16: I'm pretty sure we've got this here for a reason, i.e. that `_` causes a binding error with the HTTP error, so I think you'll need to restore this.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Clients/ContentApi/ContentApiClient.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System.Net.Http.Json;
+﻿using System.Diagnostics.CodeAnalysis;
```
benoutram, line 1: There's an unused import here.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Builders/HeathCheckStrategyBuilder.cs
```diff
@@ -0,0 +1,41 @@
+using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Functions.HealthChecks.Strategies;
+using Moq;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests.Builders;
+
+public class HeathCheckStrategyBuilder
```
benoutram, line 6: Looks like 'L' key might be broken 😆.  Throughout this class 'health' is 'heath'.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Functions/HeathChecks/HealthCheckFunctionTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        var heathCheckResult = Assert.IsType<HealthCheckResponse>(okResult.Value);
+        Assert.True(heathCheckResult.IsHealthy);
+    }
+    
+    [Theory]
+    [InlineData(true, true, false)]
+    [InlineData(false, true, true)]
+    [InlineData(true, false, false)]
+    [InlineData(false, false, false)]
+    public async Task WhenAnyOrAllStrategiesAreUnheatlhy_ThenHeathCheckIsUnhealthy(
```
benoutram: Another unhealthy type here!

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Functions/HeathChecks/HealthCheckFunctionTests.cs
```diff
@@ -0,0 +1,98 @@
+using System.Net;
+using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Functions.HealthChecks;
+using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Functions.HealthChecks.Dtos;
+using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Functions.HealthChecks.Strategies;
+using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests.Builders;
+using Microsoft.AspNetCore.Mvc;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests.Functions.HeathChecks;
```
benoutram, line 8: The very last commit looks like it's reverted some of the fixes to the namespaces.

`HeathChecks` vs. `HealthChecks

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/ProgramTests.cs
```diff
@@ -152,47 +154,47 @@ public void ShouldBeConfigured()
...
-                    """);
+                IHost StartupHost() =>
+                    base.GetSutWithConfig(FullConfig, """
+                                                      {
+                                                         "App": {
+                                                              "SearchStorageConnectionString": "DefaultEndpointsProtocol=https;EndpointSuffix=core.windows.net;AccountName=mystorageaccount;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=https://mystorageaccount.blob.core.windows.net/;FileEndpoint=https://mystorageaccount.file.core.windows.net/;QueueEndpoint=https://mystorageaccount.queue.core.windows.net/;TableEndpoint=https://mystorageaccount.table.core.windows.net/",
+                                                              "SearchableDocumentsContainerName": null
+                                                         }
+                                                      }
+                                                      """);
```
benoutram, line 190: Just wondering if it might be better to use connection string `UseDevelopmentStorage=true`.

I know it's not an integration test and nothing is running but it's a shortcut to

```
DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;
AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;
BlobEndpoint=http://127.0.0.1:10000/devstoreaccount1;
QueueEndpoint=http://127.0.0.1:10001/devstoreaccount1;
TableEndpoint=http://127.0.0.1:10002/devstoreaccount1;
```

so it is a valid connection string and it avoids the long line here. Could also use this on L20, L139.

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
```diff
@@ -2445,7 +2446,7 @@ public async Task ListLatestReleaseVersions_SingleRelease()
                 Assert.Equal(releaseVersion.Release.TimePeriodCoverage, summaryViewModel.TimePeriodCoverage);
                 Assert.Equal(releaseVersion.Published, summaryViewModel.Published);
                 Assert.Equal(releaseVersion.Live, summaryViewModel.Live);
-                Assert.Equal(releaseVersion.PublishScheduled, summaryViewModel.PublishScheduled);
+                Assert.Equal(releaseVersion.PublishScheduled?.ConvertUtcToUkTimeZone(), summaryViewModel.PublishScheduled);
```
benoutram, line 2449: I think you need to make a similar change in `GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.ManageContent.ManageContentPageServiceTests.GetManageContentPageViewModel` because that test is also failing.

I was surprised that we don't also have the same issue in `GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.ReleaseVersionServiceTests.GetReleaseTests.Success` because we do exactly the same thing in the mapping to build the `ReleaseVersionViewModel` view model as we do here for the `ReleaseVersionSummaryViewModel`.

I investigated this and found that the difference is in how the data is being set up for the Published release version. In this test which isn't failing, the release version has no value of `PublishScheduled`. We could still make the same change though, with the null-forgiving operator like you've done here.

I also looked into why this started failing yesterday and noticed that today the date generated is 2024-04-01 and yesterday it was 2024-03-31. In 2024 BST began on 2024-03-31.

benoutram (changes requested):

benoutram (approved):

PR #5677, opened 2025-03-10
Rollback Aspect Injector to 2.8.1 to provide compatibility with M1 Ma…

leeoades:
Aspect Injector 2.8.2 broke everything for M1 Macbooks. 2.8.2 is the latest stable release so no fix has been forthcoming since 2023 when this was reported. Rollback to 2.8.1.


tomjonesdev (approved):

benoutram (approved):
Did a quick check and I think we've only ever used 2 versions, 2.7.3, and 2.8.2.

I think this rollback to 2.8.1 should be ok, with some testing!

This change will be a candidate to be reverted by Renovate in an 'Update minor dependencies' PR. We've not been very good at prioritising merging of any Renovate PR's  recently but just thinking that if you were to block this by adding some config to `renovate.json` that it might also be a good place in the code to leave a comment referring to a Github issue (presuming there already is one?) explaining why we are pinning the version to 2.8.1?

PR #5678, opened 2025-03-11
EES-5826 - use channels to separate Public API query response from the writing of queries for analytics

duncan-at-hiveit:
This PR:
* replaces the fire-and-forget implementation of the writing out of Public API queries for analytics with a channel implementation.

Rather than spawning off a separate Task from the main request-processing Thread and simply not awaiting it, instead we provide a Channel whereby the DataSetQueryService can write the desired query to be captured to the channel.

A BackgroundService continually monitors the Channel, and reads from the Channel when new messages are available.  It then forwards the message to the QueryAnalyticsWriter service, which then performs the actual work of writing out the query for analytics.


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/QueryAnalyticsChannel.cs
leeoades, line 1: Unless you think it provides added clarity, I wouldn't put _Channel_ in the name since that is an implementation detail. You can think if this more as the QueryAnalyticsReporter or Manager or Handler? ie it's just the place where all other components can go to send/report their analytics.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/QueryAnalyticsConsumer.cs
```diff
@@ -0,0 +1,19 @@
...
+    IQueryAnalyticsChannel channel,
+    IQueryAnalyticsWriter queryAnalyticsWriter
+    ) : BackgroundService
+{
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        while (!stoppingToken.IsCancellationRequested)  
+        {            
+            var message = await channel.ReadQuery(stoppingToken);  
+            await queryAnalyticsWriter.ReportDataSetVersionQuery(message);
```
leeoades, line 16: If this throws, I'm guessing this background service will stop writing the files? Probably want to guard against that - and log errors?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/QueryAnalyticsWriter.cs
```diff
@@ -47,15 +30,15 @@ public async Task ReportDataSetVersionQuery(
 
         Directory.CreateDirectory(directory);
```
leeoades, line 31: I would catch and log errors here - permissions, disk full, all sorts of things that could go wrong.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
```diff
@@ -373,13 +375,16 @@ private static void ApplyCustomMigrations(IApplicationBuilder app)
...
-            DataSetQueryRequest query, int resultsCount, int totalRowsCount, DateTime startTime, DateTime endTime)
+        public Task WriteQuery(CaptureDataSetVersionQueryRequest request, CancellationToken cancellationToken)
         {
-            return Task.CompletedTask;       
+            return Task.CompletedTask;
+        }
+
+        public ValueTask<CaptureDataSetVersionQueryRequest> ReadQuery(CancellationToken cancellationToken)
+        {
+            return default;
```
leeoades, line 387: Could this potentially lead to a spinning process? If something was consuming it, it would be pouring out a stream of values? Perhaps add :
await Task.Delay(Timeout.Infinite, cancellationToken);

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/Interfaces/IQueryAnalyticsChannel.cs
leeoades, line 1: Not something for you to change - but just whilst I have your attention :-)
It would be usual for a there to be two interfaces, one as the "Reporter" where components can send their data, and another as a "Consumer" that would take the data out and process it. This cleans up the intentions of components and prevents any possibility of anything else reading from the channel when they should only ever be writing. The container of course resolves to the same instance.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/QueryAnalyticsConsumer.cs
```diff
@@ -0,0 +1,19 @@
...
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Services.Interfaces;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Services;
+
+public class QueryAnalyticsConsumer(
+    IQueryAnalyticsChannel channel,
+    IQueryAnalyticsWriter queryAnalyticsWriter
+    ) : BackgroundService
+{
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
```
leeoades, line 11: Can this be unit tested?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/QueryAnalyticsChannel.cs
```diff
@@ -0,0 +1,20 @@
+using System.Threading.Channels;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Services.Interfaces;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Services;
+
+public class QueryAnalyticsChannel : IQueryAnalyticsChannel
```
leeoades, line 6: Unit tests for this class?

leeoades (approved):
This is very nice!
A few comments but will leave to your discretion. Let me know if you produce anything else for me to review.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/QueryAnalyticsChannel.cs
duncan-at-hiveit: Yup that's a good point ta.  I like `Manager`, so I'll go with that suggestion thanks.

duncan-at-hiveit: Also renamed "WriteQuery" to "AddQuery" to further move away from the Channel implementation details.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/QueryAnalyticsConsumer.cs
duncan-at-hiveit: I have guarded and added logging of errors, thanks!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/QueryAnalyticsWriter.cs
duncan-at-hiveit: Good idea - I've added try-catch and logging here.  Note that I've added the exception itself to the first argument of ILogger.LogError() here and in other places where I've taken your suggestion of logging.  This is to ensure that our exceptions are reported upon in Application Insights.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
duncan-at-hiveit: Haha yeah, good point!  Thanks for the spot!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/QueryAnalyticsConsumer.cs
duncan-at-hiveit: It is integration tested - covered by the successful capture of analytics in `DataSetsControllerGetQueryTests` and `DataSetsControllerPostQueryTests` and their various `QueryAnalytics*Tests`.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/QueryAnalyticsChannel.cs
duncan-at-hiveit: It is integration tested - covered by the successful capture of analytics in DataSetsControllerGetQueryTests and DataSetsControllerPostQueryTests and their various QueryAnalytics*Tests.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/QueryAnalyticsWriter.cs
```diff
@@ -28,13 +28,24 @@ public async Task ReportDataSetVersionQuery(CaptureDataSetVersionQueryRequest re
...
+                Path.Combine(directory, filename),
+                contents: serialisedRequest);
+        }
+        catch (Exception e)
+        {
+            logger.LogError(
+                exception: e,
+                message: "Error whilst writing {QueryRequest} to disk",
+                nameof(CaptureDataSetVersionQueryRequest));
+            throw;
```
leeoades, line 47: It might be worth considering whether you feel that this exception has been handled. Otherwise you end up with the classic situation of every layer logging the same exception. I'm not sure rethrowing this achieves much (other than excessive logging)?

leeoades (approved):
Just a minor comment added. Cool stuff though.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/QueryAnalyticsWriter.cs
duncan-at-hiveit: I actually meant to remove the throw after adding the Exception as the first argument to `LogError` so I'm glad you spotted it :)  I originally put it in so we'd have clear exception logging in App Insights, but we'll get this via the LogError and the first argument usage, so I will remove the throw ta.

PR #5682, opened 2025-03-13
EES-5874 Put Publication Summary in the metadata of the search blob instead of Release Version

leeoades:
Put Publication Summary in the metadata of the search blob instead of Release Version.
This involves bringing the summary into the cache so we'll need a faffy deploy ticket to invalidate the caches.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/ReleaseControllerTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/ReleaseControllerTests.cs
@@ -93,7 +93,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Api.Tests.Controlle
             {
                 Id = publicationId,
                 Title = "Publication Title",
-                Summary = "<p>This is the publication summary</p>",
+                Summary = "This is the publication summary",
                 Theme = new ThemeViewModelBuilder().WithTitle("the theme"),
                 Slug = "publication-slug",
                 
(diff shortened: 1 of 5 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/ReleaseControllerTests.cs
```diff
@@ -93,6 +93,7 @@ public async Task GetLatestReleaseAsSearchable()
             {
                 Id = publicationId,
                 Title = "Publication Title",
+                Summary = "<p>This is the publication summary</p>",
```
benoutram, line 96: Publication.Summary and PublicationCacheViewModel.Summary are plain text, not html.

src/GovUk.Education.ExploreEducationStatistics.Content.ViewModels/PublicationCacheViewModel.cs
```diff
@@ -8,6 +8,8 @@ public record PublicationCacheViewModel
 
     public string Title { get; init; } = string.Empty;
 
+    public string Summary { get; set; } = string.Empty;
```
benoutram, line 11: We need a faffy deploy ticket for this change, to populate this new value in the cache.

It will be a similar task to https://dfedigital.atlassian.net/browse/EES-5918 except that the request will need to be `DELETE /api/bau/public-cache/publications/publication-json`

src/GovUk.Education.ExploreEducationStatistics.Content.ViewModels/ReleaseSearchViewModel.cs
```diff
@@ -28,7 +28,7 @@ public ReleaseSearchViewModel(
         ReleaseVersionId = releaseVersion.Id;
         Published = releaseVersion.Published ?? throw new ArgumentException("Release must have a published date");
         PublicationTitle = publication.Title;
-        Summary = ExtractHtmlBlocks(releaseVersion.SummarySection.Content).StripHtml();
+        Summary = publication.Summary.StripHtml();
```
benoutram, line 31: PublicationCacheViewModel.Summary should be plain text, not html, so no need for `.StripHtml()`.

benoutram (approved):

PR #5683, opened 2025-03-13
EES-5911 Move focus to input after clicking 'Edit Section title'

Guy-HiveIT:
Fixes https://dfedigital.atlassian.net/browse/EES-5911

Autofocus attribute ensures that the input is focussed as soon as it is rendered (when the user clicks the 'Edit Section title' button.

n.b I have had to disable the eslint rule for the autofocus attribute as I don't think it is relevant for this - autofocus is expected behaviour and would be implemented by us manually if it wasn't handled by the browser and the `autofocus` attribute.


amyb-hiveit (approved):
Good call on using autoFocus 👍

PR #5686, opened 2025-03-14
EES 5941 fix focus on modal close data groupings

Guy-HiveIT:
This fixes the issue - https://dfedigital.atlassian.net/browse/EES-5909 - for the data groupings modal. Now if you close the modal, focus is returned to the 'Edit' button that triggered opening the modal.

The fix involved refactoring to make the modal use the 'Edit' button as the `triggerButton`. 

This required the modal to be included for each row, rather than being at the top level of the ChartDataGroupingsConfiguration.tsx file.


src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartDataGroupingsConfigurationRow.tsx
```diff
@@ -0,0 +1,83 @@
...
+  label,
+  map,
+  meta,
+  unit,
+  onChange,
+}: Props) {
+  const [editDataSetConfig, setEditDataSetConfig] = useState<{
+    dataSetConfig: MapDataSetConfig;
+    unit: string;
+  }>();
```
bennettstuart, line 34: I think we can just use a `const [showModal, toggleModal] = useToggle()` here instead, as the `editDataSetConfig` will only ever be which ever dataSetConfig is provided in the props.

in `ChartDataGroupingForm` below, we can simply pass those same props straight into the component

bennettstuart (changes requested):
A tidy approach, just a small change requested :)

src/explore-education-statistics-admin/src/pages/release/datablocks/components/chart/ChartDataGroupingsConfigurationRow.tsx
Guy-HiveIT: Good catch, I didn't see that. Have updated it now and it is much cleaner. Thanks :)

bennettstuart (approved):
I had to kick off the FE tests to run again, and they passed on the second time, weirdly...
Looks good to me!

PR #5687, opened 2025-03-14
EES-5913 - fix issues with using Preview Tokens on draft DataSets

duncan-at-hiveit:
This hotfix PR fixes various issues around the usage of Preview Tokens and the Preview Token page.

Specifically this PR fixes:

## Preview Token page examples

The Preview Tokens page was using the draft DataSetVersion Id in the code example tabs (e.g. the cURL example shown in the screenshot below).

image

This PR fixes this to use the DataSet Id instead, as per all of the other examples on the page.

This PR also adds Jest tests around the checking of the URLs and code example block on this page.

## Preview Token permissions to query the DataSet summary

Previously we had support for using Preview Tokens to call API endpoints when targeting specific DataSetVersions on the DataSet (by using the dataSetVersion query parameter that is included in all of the example endpoints on this page other than the Data set summary endpoint).

We would also get away with being able to call the Data set summary endpoint of a Data set that's *already published*, as the auth handler responsible for making this decision would see that the DataSet is already public and so allow access, valid Preview Token or no.

What we were failing on was using a Preview Token against a DataSet that's not yet had a published version gone out, and thus is still unavailable to the public.

This PR adds support for the use of Preview Tokens to view a Data set summary of a DataSet that's not yet publicly available.  It does this by checking that the DataSet being called has a *draft* DataSetVersion that has a matching valid token to the one being used.

## Missing test cases

This PR adds lots of additional tests around Preview Tokens for all of the endpoints listed on the Preview Tokens page that didn't yet have Preview Tokens tests.  Specifically, this was for the Data set summary, Download CSV, and Get Metadata endpoints.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Security/AuthorizationHandlers/ViewDataSetAuthorizationHandlerTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Security/AuthorizationHandlers/ViewDataSetAuthorizationHandlerTests.cs
@@ -1,8 +1,18 @@
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using GovUk.Education.ExploreEducationStatistics.Common.Tests.Fixtures;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Constants;
 using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Security.AuthorizationHandlers;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Services;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Services.Security;
 using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests.TheoryData;
 using GovUk.Education.ExploreEducationStatistics.Public.Data.Model;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Database;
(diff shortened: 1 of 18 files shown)


mmyoungman (dismissed):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Security/AuthorizationHandlers/ViewDataSetAuthorizationHandlerTests.cs
```diff
@@ -44,9 +54,215 @@ public void Failure(DataSetStatus status)
...
+
+        var handler = BuildHandler(
+            publicDataDbContext: publicDataDbContext.Object,
+            requestHeaders:
+            [
+                PreviewTokenRequestHeader(dataSetVersion.PreviewTokens[0])
+            ]);
+        var context = CreateAnonymousAuthContext<ViewDataSetRequirement, DataSet>(dataSet);
+
+        handler.HandleAsync(context);
```
leeoades, line 86: This is surely missing an await?

mmyoungman (approved):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Security/AuthorizationHandlers/ViewDataSetAuthorizationHandlerTests.cs
duncan-at-hiveit: Good spot!  I've made all test methods for this project's auth handlers async, and awaited these ta.

PR #5688, opened 2025-03-14
EES-5912 Prevent NVDA announcing 'unavailable' on button click in admin section

Guy-HiveIT:
This PR stops NVDA announcing 'unavailable' immediately after clicking two buttons in the admin section.
It removes 'double click prevention' on the buttons, which stops them from temporarily toggling to 'aria-disabled'.

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseContentAccordionSection.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseContentAccordionSection.tsx
@@ -226,7 +226,11 @@ const ReleaseContentAccordionSection = ({
               )}
 
               <ButtonGroup className="govuk-!-margin-bottom-8 dfe-justify-content--center">
-                <Button variant="secondary" onClick={addBlock}>
+                <Button
+                  variant="secondary"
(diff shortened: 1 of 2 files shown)


src/explore-education-statistics-admin/src/pages/release/content/components/ReleaseContentAccordionSection.tsx
```diff
@@ -226,7 +226,11 @@ const ReleaseContentAccordionSection = ({
               )}
 
               <ButtonGroup className="govuk-!-margin-bottom-8 dfe-justify-content--center">
-                <Button variant="secondary" onClick={addBlock}>
+                <Button
+                  variant="secondary"
+                  onClick={addBlock}
+                  preventDoubleClick={false}
```
bennettstuart, line 232: Is this a more general issue with asynchronous button onClicks, will **all** buttons potentially be announcing "unavailable" as they go disabled for a moment? @amyb-hiveit what do you think about this solution?

It seems the crux of the issue is that the user is left with their focus on a disabled input for a moment. We could manage their focus, or stop buttons with prevented double clicks from disabling the button, instead just turn off the functionality for a short duration.

PR #5695, opened 2025-03-18
EES-5959 Create a dedicated DevOps build pipeline for Search

benoutram:
This PR create a separate pipeline to do the build, test, and publishing steps of the Search solution which currently consists of the  Searchable Documents Azure Function application.

In future any other projects added to the Search solution can be built and published by this pipeline.

It it set up to trigger the existing Search infrastructure and application deployment pipeline created by https://github.com/dfe-analytical-services/explore-education-statistics/pull/5628.

Long build times and build failures in the main service build pipeline are currently delaying PR’s from being merged and delaying deployments of the infrastructure and Searchable Documents Azure Function application.

This new pipeline allows projects in the Search solution to be built, tested and deployed independently of the main service. PR and CI triggered builds of the Search solution will no longer have to wait for the main service build to complete. There will be no checks during the build on projects outside of the Search solution so they won’t fail due to flaky tests in other projects of the main service.

### Other changes

- Exclude changes solely under `/infrastructure` from triggering PR builds of the main EES build pipeline.

- Exclude changes solely under `/src/GovUk.Education.ExploreEducationStatistics.Content.Search*` from triggering PR builds of the main EES build pipeline.

- The main EES build pipeline CI trigger for code changes on the dev, test and master branches now excludes changes under `src/GovUk.Education.ExploreEducationStatistics.Content.Search*`.

- Disable unnecessary download of artifacts in the `DeploySearchInfrastructure` step of the Search infrastructure and application deployment pipeline. 

- Switches all projects to use the simplified _artifacts_ output directory, added by .NET 8. All build outputs from all projects are gathered into a common location, _artifacts_, separated by project. See [Artifacts output layout](https://learn.microsoft.com/en-us/dotnet/core/sdk/artifacts-output).

### DIFF
--- a/infrastructure/templates/search/ci/azure-pipelines-build.yml
+++ b/infrastructure/templates/search/ci/azure-pipelines-build.yml
@@ -1,0 +1,68 @@
+trigger:
+  branches:
+    include:
+    - master
+    - dev
+    - test
+  paths:
+    include:
+    - src/GovUk.Education.ExploreEducationStatistics.Content.Search*/**
+
+pr:
+  branches:
+    include:
+    - master
+    - dev
+    - test
+  drafts: false
+  paths:
+    include:
+    - src/GovUk.Education.ExploreEducationStatistics.Content.Search*/**
+
+pool:
+  vmImage: 'ubuntu-22.04'
+
+variables:
+  buildConfiguration: 'Release'
+  dotNetVersion: '8.0.x'
+  solution: 'src/GovUk.Education.ExploreEducationStatistics.Search.sln'
+
+jobs:
+- job: BuildAndDeploy
+  displayName: 'Build, test, and publish'
+  steps:
+  - checkout: self
+
+  - task: UseDotNet@2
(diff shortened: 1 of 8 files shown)


leeoades (approved):
I don't know much about this stuff so hard for me to judge. I know that with these things, the proof is in the pudding so need  to just run it really :-)

infrastructure/templates/search/ci/azure-pipelines-build.yml
```diff
@@ -0,0 +1,68 @@
...
+jobs:
+- job: BuildAndDeploy
+  displayName: 'Build, Test, and Publish Search Solution'
+  steps:
+  - checkout: self
+
+  - task: UseDotNet@2
+    displayName: 'Install .NET SDK'
+    inputs:
+      version: 8.0.x
```
duncan-at-hiveit, line 39: Should this be using the `dotNetVersion` variable that was defined above?

PR #5696, opened 2025-03-18
EES-5942 delete data files modal focus

Guy-HiveIT:
This PR fixes an issue where focus was not returned to a modal trigger button on the 'delete data files from release' section of the admin site, when the confirmation modal is closed.

It does this by moving the modal to be per file row instead of at the top level component, and using the `triggerButton` prop for the modal.

Mostly done by Amy B and tweaked/finished by me.

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/data/components/DeleteDataFileModal.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/data/components/DeleteDataFileModal.tsx
@@ -1,13 +1,14 @@
+import releaseDataFileQueries from '@admin/queries/releaseDataFileQueries';
 import releaseDataFileService, {
   DataFile,
-  DeleteDataFilePlan,
 } from '@admin/services/releaseDataFileService';
 import ButtonText from '@common/components/ButtonText';
 import LoadingSpinner from '@common/components/LoadingSpinner';
 import ModalConfirm from '@common/components/ModalConfirm';
 import useToggle from '@common/hooks/useToggle';
 import logger from '@common/services/logger';
-import React, { useCallback, useEffect, useState } from 'react';
+import { useQuery } from '@tanstack/react-query';
+import React, { useCallback } from 'react';
 
 interface Props {
   dataFile: DataFile;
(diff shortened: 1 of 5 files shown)


src/explore-education-statistics-admin/src/pages/release/data/components/DeleteDataFileModal.tsx
```diff
@@ -0,0 +1,131 @@
...
+    </ModalConfirm>
+  );
+}
+
+interface PlanProps {
+  dataFile: DataFile;
+  releaseVersionId: string;
+}
+
+function Plan({ dataFile, releaseVersionId }: PlanProps) {
```
amyb-hiveit, line 66: I'm not clear on why this needs to be separated out into it's own component, if it's so the request to fetch the delete plan isn't called until the modal is open then you could do that in `DeleteDataFileModal` by passing an `enabled` flag to `useQuery`, like:
```
  const { data: plan, isLoading } = useQuery({
    ...releaseDataFileQueries.getDeleteFilePlan(releaseVersionId, dataFile),
    enabled: open,
  });
  ```

File: src/explore-education-statistics-admin/src/queries/releaseDataFileQueries.ts [hunk 2]
```diff
@@ -8,6 +10,16 @@ const releaseDataFileQueries = createQueryKeys('releaseDataFile', {
       queryFn: () => releaseDataFileService.getDataFiles(releaseId),
     };
   },
+  getDeleteFilePlan(releaseVersionId: string, dataFile: DataFile) {
```
amyb-hiveit, line 13: A minor thing that doesn't really matter, but I noticed that the request in `releaseDataFileService` only needs the dataFile id, so we could just pass that id through to this and then on to the service instead of the whole dataFile.

amyb-hiveit (commented):
Looks good, just a query about the `Plan` component to discuss.

src/explore-education-statistics-admin/src/pages/release/data/components/DeleteDataFileModal.tsx
Guy-HiveIT: Yep it was to only fetch the plan when open. Didn't know about that `enabled` flag - a better approach, thanks, will change.

amyb-hiveit (approved):

PR #5697, opened 2025-03-18
EES-5943 Fix delete ancillary file modal focus restoration

Guy-HiveIT:
This PR refactors the "delete supporting file upload" modal to be rendered for each file rather than shared in the top level section. 
This lets us use the `triggerButton` for handling focus, so focus can be restored to the 'Delete file' button when the modal is closed.

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/data/components/AncillaryFileSummaryList.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/data/components/AncillaryFileSummaryList.tsx
@@ -8,6 +8,7 @@ import releaseAncillaryFileService, {
 } from '@admin/services/releaseAncillaryFileService';
 import ButtonText from '@common/components/ButtonText';
 import FormattedDate from '@common/components/FormattedDate';
+import ModalConfirm from '@common/components/ModalConfirm';
 import SummaryList from '@common/components/SummaryList';
 import SummaryListItem from '@common/components/SummaryListItem';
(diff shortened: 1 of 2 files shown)


src/explore-education-statistics-admin/src/pages/release/data/components/AncillaryFileSummaryList.tsx
```diff
@@ -80,7 +87,21 @@ export default function AncillaryFileSummaryList({
               >
                 Edit file
               </Link>
-              <ButtonText onClick={onDelete}>Delete file</ButtonText>
+              <ModalConfirm
+                open={deleteFile && deleteFile.id === file.id}
```
amyb-hiveit, line 91: `deleteFile` isn't being passed in from `ReleaseFileUploadsSection` so I don't think this is needed.

amyb-hiveit (changes requested):
Just one little thing to sort out, otherwise looks good 👍

src/explore-education-statistics-admin/src/pages/release/data/components/AncillaryFileSummaryList.tsx
Guy-HiveIT: Good catch, I thought I was passing that through and that was determining whether the modal was open! Apparently not.. Will remove here.

amyb-hiveit (approved):

PR #5699, opened 2025-03-18
EES-5960 Return focus to modal trigger - Edit publication details

Guy-HiveIT:
This PR fixes an issue where focus was not returned to a button that opens the publication update form modal, when the modal is closed.
To do this, the modal was refactored to use a triggerButton instead of handling state manually.


Guy-HiveIT:
I need to change this. As it stands this current solution seems to prevent validation before opening the modal.

PR #5700, opened 2025-03-18
EES-5934 Fixing Admin Public API Data Set Changelog page

jack-hive:
Previously, when an admin visited the Public API version history page for a particular API data set, and clicked to view the changelog for versions `1.1` and above, they were sometimes displayed with the incorrect version number and version 'type' (`Major`/`Minor`).

This was because the frontend had [some logic](https://github.com/dfe-analytical-services/explore-education-statistics/pull/5700/files#:~:text=const%20dataSetVersion%20%3D,%3F.latestLiveVersion%3B) which only ever showed the version number and type for either the **latest live** version of the data set, or the current **draft** version, depending on whether or not the version you selected is the **draft** one or not. It didn't display the details for the version you have specifically requested (**caveat** - it DID do this sometimes, but only by coincidence).

For example, previously, the changelog page for the following versions showed:
- `2.0` -> Displayed `2.0` `Major` -> CORRECT (Coincidentally)
- `1.2` -> Displayed `2.0` `Major` -> INCORRECT
- `1.1` -> Displayed `2.0` `Major` -> INCORRECT
- `1.0` -> Can't access the changelog page for the first version

## Implementation

Previously, the changelog page component in the FE was retrieving the data it needs from multiple endpoints. Namely, the following endpoints in the Admin API:
- `/public-data/data-sets/{dataSetId}`
- `/public-data/data-set-versions/{dataSetVersionId}/changes`  

It was also missing some data it needs specific to the Data Set Version we're interested in.

One solution, to provide the missing information, would have been to add yet another endpoint to the BE that this component can call - one that is used to `GET` details specific to a Data Set Version. The FE component would then retrieve the data it needs from 3 different endpoints.

The alternative solution, and the one we have went with, is to refactor the `/public-data/data-set-versions/{dataSetVersionId}/changes` endpoint to retrieve **ALL** of the data that this component needs; therefore removing the need for the component to call multiple endpoints and to stitch them together.

The endpoint now returns details of:
- The Data Set
- The requested Data Set Version
- And the changelog for the Data Set Version

### Other changes

One thing to note, is that we have refactored the `PublicDataApiClient` in the BE to deserialize the Changes object it receives from the Public API into a DTO. This has a few advantages:
- We no longer have to stream the data from the Public API to the FE via the Admin API
  - And we can do this **without** pulling in the project reference to the Public API's view-models (this is similar to how you would consume other public API's, such as Google's)
- Promotes loose coupling
- We only need to expose the **required** data (which, in this case, is coincidentally all of it)

We then map this DTO onto a corresponding view-model before returning the data to the FE.

It does, however, potentially introduce the risk of deserializing and mapping the data incorrectly; and therefore losing information. So we need to make sure that we test this thoroughly.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionsControllerTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Controllers/Api/Public.Data/DataSetVersionsControllerTests.cs
@@ -2,7 +2,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Net;
 using System.Net.Http;
 using System.Net.Http.Json;
 using System.Security.Claims;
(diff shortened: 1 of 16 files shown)


leeoades (commented):
Goodness, that ChangeViewModel is a beast isn't it...
I hadn't realised it was so massive. And we've introduced the dto to separate us from the admin api, and a viewmodel to send it back out - and they're all the same?
Is the UI using a component that is expecting the ChangeViewModel to always be the same as the admin api or does it never call the admin api?
I'm trying to remember now why the UI doesn't just call the admin api directly to get this change information. I'm sure you did tell me.

PR #5703, opened 2025-03-20
Implement reindexer function

leeoades:
The publisher raises an event signalling that a new release version has been published.
This event is subscribed to and queues for processing by a Function App.
That function creates the searchable document in Azure and then places its own message into a new queue signalling that a new document has been created in Azure storage.
This work is to implement the next part of the process, which is to consume that message and trigger a reindex.

Note that I needed to correct the dto of the message arriving from the publisher since I forgot that it would of course be a payload wrapped in an event envelope.

I also added another healthcheck strategy which reports back on whether the indexer is configured and can be reached.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Functions/HealthChecks/HealthCheckFunctionTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Functions/HealthChecks/HealthCheckFunctionTests.cs
@@ -75,7 +75,7 @@ public class HealthCheckFunctionTests
         [
             new HealthCheckStrategyBuilder().WhereResultIsHealthy("Result one").Build(),
             new HealthCheckStrategyBuilder().WhereResultIsUnhealthy("Result two").Build(),
-            new HealthCheckStrategyBuilder().WhereResultIsHealthy(null).Build()
+            new HealthCheckStrategyBuilder().WhereResultIsHealthy("Result three").Build()
         ];
         var sut = GetSut(strategies);
         
@@ -90,7 +90,7 @@ public class HealthCheckFunctionTests
         {
             new HealthCheckResult(true, "Result one"),
             new HealthCheckResult(false, "Result two"),
-            new HealthCheckResult(true, null)
+            new HealthCheckResult(true, "Result three")
         };
         Assert.Equal(expectedResults, healthCheckResult.Results);
     }
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Clients/AzureSearch/AzureSearchIndexerClientWrapper.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Clients/AzureSearch/AzureSearchIndexerClientWrapper.cs
@@ -1,0 +1,24 @@
+﻿namespace GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Clients.AzureSearch;
+
+/// <summary>
+/// A thin wrapper around Azure's SearchIndexerClient to allow it to be mocked
+/// </summary>
+/// <param name="azureSearchIndexerClient">A real SearchIndexerClient instance</param>
+public class AzureSearchIndexerClientWrapper(Azure.Search.Documents.Indexes.SearchIndexerClient azureSearchIndexerClient) : IAzureSearchIndexerClient
+{
+    public async Task ResetIndexerAsync(string indexName, CancellationToken cancellationToken) => await azureSearchIndexerClient.ResetIndexerAsync(indexName, cancellationToken);
+
+    public async Task RunIndexerAsync(string indexName, CancellationToken cancellationToken) => await azureSearchIndexerClient.RunIndexerAsync(indexName, cancellationToken);
+    public async Task<bool> IndexerExists(string indexName, CancellationToken cancellationToken)
+    {
+        try
+        {
+            var response = await azureSearchIndexerClient.GetIndexerAsync(indexName, cancellationToken);
+            return response.HasValue;
+        }
+        catch (Azure.RequestFailedException)
+        {
+            return false;
+        }
+    }
+}
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Clients/AzureSearch/AzureSearchIndexerClientWrapper.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Clients/AzureSearch/AzureSearchIndexerClientWrapper.cs
@@ -4,11 +4,15 @@
 /// A thin wrapper around Azure's SearchIndexerClient to allow it to be mocked
 /// </summary>
 /// <param name="azureSearchIndexerClient">A real SearchIndexerClient instance</param>
-public class AzureSearchIndexerClientWrapper(Azure.Search.Documents.Indexes.SearchIndexerClient azureSearchIndexerClient) : IAzureSearchIndexerClient
+public class AzureSearchIndexerClientWrapper(
+    Azure.Search.Documents.Indexes.SearchIndexerClient azureSearchIndexerClient) 
+    : IAzureSearchIndexerClient
 {
-    public async Task ResetIndexerAsync(string indexName, CancellationToken cancellationToken) => await azureSearchIndexerClient.ResetIndexerAsync(indexName, cancellationToken);
+    public async Task ResetIndexerAsync(string indexName, CancellationToken cancellationToken) =>
+        await azureSearchIndexerClient.ResetIndexerAsync(indexName, cancellationToken);
 
-    public async Task RunIndexerAsync(string indexName, CancellationToken cancellationToken) => await azureSearchIndexerClient.RunIndexerAsync(indexName, cancellationToken);
+    public async Task RunIndexerAsync(string indexName, CancellationToken cancellationToken) => 
+        await azureSearchIndexerClient.RunIndexerAsync(indexName, cancellationToken);
     
     public async Task<bool> IndexerExists(string indexName, CancellationToken cancellationToken)
     {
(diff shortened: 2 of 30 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Options/AzureSearchOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+    public const string Section = "AzureSearch";
+    
+    /// <summary>
+    /// Url of the search service endpoint. Likely to be similar to:
+    /// https://{search_service}.search.windows.net
+    /// </summary>
+    public string SearchServiceEndpoint { get; init; } = string.Empty;
+    public string? SearchServiceAccessKey { get; init; }
+    
+    public string IndexName { get; init; } = string.Empty;
```
benoutram, line 16: Is it possible to use the `required` keyword modifier here like we did in `EventTopicOptions` to avoid `= string.Empty` for options we know need to be set?

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Functions/HealthChecks/HealthCheckFunctionTests.cs
```diff
@@ -75,7 +75,7 @@ public async Task WhenStrategiesReturnResults_ThenHealthCheckResponseContainsRes
         [
             new HealthCheckStrategyBuilder().WhereResultIsHealthy("Result one").Build(),
             new HealthCheckStrategyBuilder().WhereResultIsUnhealthy("Result two").Build(),
-            new HealthCheckStrategyBuilder().WhereResultIsHealthy(null).Build()
+            new HealthCheckStrategyBuilder().WhereResultIsHealthy("Result three").Build()
```
benoutram, line 78: It doesn't really matter but with this change is there any advantage of having three strategies now? Could leave it as two.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Clients/AzureSearch/AzureSearchIndexerClientWrapper.cs
```diff
@@ -0,0 +1,24 @@
...
+/// <summary>
+/// A thin wrapper around Azure's SearchIndexerClient to allow it to be mocked
+/// </summary>
+/// <param name="azureSearchIndexerClient">A real SearchIndexerClient instance</param>
+public class AzureSearchIndexerClientWrapper(Azure.Search.Documents.Indexes.SearchIndexerClient azureSearchIndexerClient) : IAzureSearchIndexerClient
+{
+    public async Task ResetIndexerAsync(string indexName, CancellationToken cancellationToken) => await azureSearchIndexerClient.ResetIndexerAsync(indexName, cancellationToken);
+
+    public async Task RunIndexerAsync(string indexName, CancellationToken cancellationToken) => await azureSearchIndexerClient.RunIndexerAsync(indexName, cancellationToken);
+    public async Task<bool> IndexerExists(string indexName, CancellationToken cancellationToken)
```
benoutram, line 12: Minor one but I would add a line separator here, like you have done between `ResetIndexerAsync` and `RunIndexerAsync`.

Also there are some long lines here!

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/ReindexSearchDocuments/ReindexSearchDocumentsFunction.cs
```diff
@@ -0,0 +1,24 @@
+﻿using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Clients.AzureSearch;
+using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Functions.CreateSearchableReleaseDocuments.Dtos;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Extensions.Logging;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Functions.ReindexSearchDocuments;
+
+public class ReindexSearchDocumentsFunction(
```
benoutram, line 8: I think `ReindexSearchableDocumentsFunction` would be more in keeping with the naming elsewhere because we've used 'searchable documents' in quite a few places.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/ReindexSearchDocuments/ReindexSearchDocumentsFunction.cs
```diff
@@ -0,0 +1,24 @@
...
+    ILogger<ReindexSearchDocumentsFunction> logger,
+    ISearchIndexClient searchIndexClient)
+{
+    [Function("ReindexNewlyCreatedSearchableDocument")]
+    public async Task ReindexNewlyCreatedSearchableDocument(
```
benoutram, line 13: I know that this is being triggered with a `SearchDocumentCreatedMessageDto` queue message but just wondering if you think `ReindexSearchableDocumentsFunction` is the right name here?

It reindexes all the documents which you might not be expecting if you went off the function/method name alone.

Equally though I suppose if you go ahead with this rename, someone might be confused why the payload is a `SearchDocumentCreatedMessageDto`. 🙂

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Extensions/ObjectExtensions.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System.Diagnostics.CodeAnalysis;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Extensions;
+
+public static class ObjectExtensions
+{
+    public static bool IsDefault<T>([NotNullWhen(false)] this T obj) => EqualityComparer<T>.Default.Equals(obj, default);
```
benoutram, line 7: I feel like there's a reason for this, rather than just doing `obj.Equals(default(T))`  but I'm not sure what it is. I think adding a code comment here to explain this would be useful?

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/CreateSearchableReleaseDocuments/Dtos/ReleaseVersionPublishedEventDto.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Functions.CreateSearchableReleaseDocuments.Dtos;
+
+public record ReleaseVersionPublishedEventDto
+{
+    public required Guid ReleaseId {get;init;}
+    public required string ReleaseSlug { get; init; }
+    public required Guid PublicationId { get; init; }
+    public required string PublicationSlug { get; init; }
+    public required Guid PublicationLatestReleaseVersionId { get; init; }
```
benoutram, line 9: Shouldn't this be PublicationLatestPublishedReleaseVersionId given changes made elsewhere in https://github.com/dfe-analytical-services/explore-education-statistics/pull/5694 to ReleaseVersionPublishedEventDto?

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/CreateSearchableReleaseDocuments/Dtos/SearchDocumentCreatedMessageDto.cs
```diff
@@ -1,8 +1,15 @@
...
+    public required Guid PublicationId { get; init; }
     public required string PublicationSlug { get; init; }
+
+    public required Guid ReleaseId {get;init;}
+    public required string ReleaseSlug { get; init; }
+    
     public required Guid ReleaseVersionId { get; init; }
     public required string BlobName { get; init; }
+    
+    public required Guid PublicationLatestReleaseVersionId { get; init; }
```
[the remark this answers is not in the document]
leeoades: Oh yeah, great spot!

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Options/AzureSearchOptions.cs
leeoades: If I set it to required and there isn't any config defined then I'm wondering if it will throw when attempting to resolve the IOptions? I wonder if I can test that...

leeoades: OK, so required only enforces properties from the compiled code. Deserialising into the dto still sets the `required` property to `null` which is not helpful. I've added tests that demonstrate this.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Clients/AzureSearch/AzureSearchIndexerClientWrapper.cs
leeoades: Need to get you a wider monitor 🤣

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/ReindexSearchDocuments/ReindexSearchDocumentsFunction.cs
leeoades: I've renamed it to `ReindexSearchableDocuments` since you're right, it is reindexing everything new in there (at the moment).

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Extensions/ObjectExtensions.cs
leeoades: This is a reason but after attempting to write unit tests for it, it wasn't achieving what I wanted anyway. I'd forgotten actually and had meant to return to it. Anyway, I've removed it.
The original intention was to detect whether the deserialised value was actually empty.

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Functions/HealthChecks/HealthCheckFunctionTests.cs
benoutram: Ah ok, I hadn't spotted that the middle one is unhealthy. Thanks!

PR #5705, opened 2025-03-20
EES-5934 Fixing Admin Public API Data Set Changelog page

jack-hive:
Previously, when an admin visited the Public API version history page for a particular API data set, and clicked to view the changelog for versions `1.1` and above, they were sometimes displayed with the incorrect version number and version 'type' (`Major`/`Minor`).

This was because the frontend had [some logic](https://github.com/dfe-analytical-services/explore-education-statistics/pull/5700/files#:~:text=const%20dataSetVersion%20%3D,%3F.latestLiveVersion%3B) which only ever showed the version number and type for either the **latest live** version of the data set, or the current **draft** version, depending on whether or not the version you selected is the **draft** one or not. It didn't display the details for the version you have specifically requested (**caveat** - it DID do this sometimes, but only by coincidence).

For example, previously, the changelog page for the following versions showed:
- `2.0` -> Displayed `2.0` `Major` -> CORRECT (Coincidentally)
- `1.2` -> Displayed `2.0` `Major` -> INCORRECT
- `1.1` -> Displayed `2.0` `Major` -> INCORRECT
- `1.0` -> Can't access the changelog page for the first version

This PR makes the changes necessary for the page to retrieve the data it needs specific to the **requested** data set version. We have introduced a new Admin API `GET` endpoint for this, for the Data Set Version.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
+using AutoMapper;
 using GovUk.Education.ExploreEducationStatistics.Admin.Requests.Public.Data;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.Public.Data;
 using GovUk.Education.ExploreEducationStatistics.Admin.Services.Interfaces.Security;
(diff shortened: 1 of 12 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/Public.Data/DataSetVersionService.cs
```diff
@@ -267,7 +280,7 @@ private async Task<DataSetVersionSummaryViewModel> MapDraftVersionSummary(
         };
     }
 
-    private async Task<Either<ActionResult, DataSetVersion>> GetDataSetVersion(
+    private async Task<Either<ActionResult, DataSetVersion>> GetVersion(
```
leeoades, line 283: I don't think this naming makes sense as it isn't returning a version.

bennettstuart (approved):
LGTM!

PR #5713, opened 2025-03-21
EES-5910 Ensure tooltips visible when tabbing through map regions

Guy-HiveIT:
This PR fixes an issue where tooltips were sometimes not visible when tabbing through regions of a map block.
We detect tab presses and zoom the map out when this happens.

N.B I have also added handling for user pressing 'Enter' when focussing on a region, to 'select' the region and display the key stat block. If you think that is out of scope for this ticket or otherwise not wanted, let me know and I will remove it.

### DIFF
--- a/src/explore-education-statistics-common/src/modules/charts/components/MapGeoJSON.tsx
+++ b/src/explore-education-statistics-common/src/modules/charts/components/MapGeoJSON.tsx
@@ -88,8 +88,7 @@ export default function MapGeoJSON({
 
   useEffect(() => {
     if (!selectedFeature) {
-      // reset zoom
-      map.setZoom(5);
+      resetZoom();
       return;
     }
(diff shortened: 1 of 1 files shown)


src/explore-education-statistics-common/src/modules/charts/components/MapGeoJSON.tsx
```diff
@@ -107,6 +106,8 @@ export default function MapGeoJSON({
     }
   }, [map, selectedFeature]);
 
+  const resetZoom = () => map.setZoom(5);
```
bennettstuart, line 109: Do we need to reset the position also? the default position is available in the parent component so could be used if so

bennettstuart (approved):
Left one comment, if it's not required feel free to merge

PR #5720, opened 2025-03-25
EES-5984 'delete data block' modal focus restoration on close

Guy-HiveIT:
This PR fixes the focus restoration when the 'delete data block' modal is shown and then closed (via escape or cancel button).

This modal is used both on the _Release->Data Blocks List_ page, as well as the individual _Release->Data Block Edit_ page.

Focus handling is fixed by moving the modal to each instance of a data block, rather than being shared at a top level. This allows us to use a `triggerButton` that will be focussed when the modal is closed. 
Subsequently, all state related to showing the modal is moved into the modal itself.

There is still an `onConfirm` fn prop that can be passed through to the modal to handle what happens after a block is deleted.

It also has some misc improvements:
- try/catch for actually deleting a block, with logging on failure
- change `deletePlan` fetch to `useQuery`

I have also included an unrelated fix for a failing UI test - increasing the wait time on deleting data file modal as per Nusrath's request.


bennettstuart (changes requested):
Couple of minor comments

bennettstuart (approved):

PR #5721, opened 2025-03-26
EES-5640 Frontend changes to provide BAU users with the ability to edit the `Label` for any Release

jack-hive:
The purpose of this PR is to allow BAU users to edit the `Label` for **any** Release (published or not). It follows from the backend changes made in https://github.com/dfe-analytical-services/explore-education-statistics/pull/5591 which adds the corresponding endpoint for BAU users to edit the `Label`.

This PR consists predominantly of FE changes.

## Changes to the UI

All of the UI changes reside in the Admin's 'Manage Releases' page:

- On this page, individual Releases currently only appear in 1 of the two tables: **Draft**/**Published**. For instance, even if a specific Release has a version of it which has been published, and one version is an amendment and in draft, then currently the Release will only appear in the **Drafts** table. We have agreed to change this so that all published Releases now appear in the **Published** table, even if they have a draft amendment in progress - i.e. they will now appear in **BOTH** tables.
  - _**This is the reason why we have had to change the existing `api/publication/{publicationId:guid}/releases` endpoint to include a new `versionsType` query parameter. We had to make the endpoint more flexible in what results it returns.**_
   
- We have added a `Edit Details` button to the **Published** table. This is the button that will be used to edit the `Label` for the Release, even in a Published state. It does this via a Modal with a form. _The modal may be used to edit other fields in the future._
 
- We have **renamed** the `Edit` button in the **Drafts** table to `Edit Draft`.
 
- We have **removed** the `View existing version` button from the **Drafts** table; which currently appears for Releases which have a draft amendment. It is no longer necessary to have, due to these Releases now appearing in both tables - and the **Published** table entry will already have a `View` button.

- For any Releases appearing in both the **Drafts** and **Published** tables, we will hide the `Amend` button in the **Published** table.

### The new 'Manage Releases' page looks like:

image

### The new `Edit Details` modal looks like:

image

image


amyb-hiveit (commented):
I've only had time to had a brief look at this, so someone will need to give it a proper review. I've left a couple of comments and noticed a few issues:
- if submit by pressing enter while have the input focused it doesn't show the warning, it only shows it if use the submit button
- if edit I a label, then cancel, then edit again it has the edited value rather than the original one
- should the confirm button be red?
- I see a warning in the console when I open a modal for second time:
﻿`useToggle.ts:28 Warning: Cannot update a component (`FormModal`) while rendering a different component (FormProvider). To locate the bad setState() call inside FormProvider

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
```diff
@@ -653,5 +649,18 @@ private async Task<PublicationCreateViewModel> GeneratePublicationCreateViewMode
...
+
+        private async Task<List<ReleaseVersion>> ListReleaseVersions(Guid publicationId, ReleaseVersionsType versionsType)
+        {
+            return versionsType switch
+            {
+                ReleaseVersionsType.Latest => await releaseVersionRepository.ListLatestReleaseVersions(publicationId),
+                ReleaseVersionsType.LatestPublished => await releaseVersionRepository.ListLatestReleaseVersions(publicationId, publishedOnly: true),
+                ReleaseVersionsType.OnlyDraft => (await releaseVersionRepository.ListLatestReleaseVersions(publicationId))
+                    .Where(rv => rv.Live == false)
+                    .ToList(),
```
mmyoungman, line 661: This means that a release version in higher review will also be included in "OnlyDraft", so maybe it's better named "NotPublished". Maybe pedantic because I know on the admin dashboard higher review releases are included in the Drafts tab.

Not suggesting a change with this thought, but I did wonder if we should call it "LatestNotPublished". There should only be one non-published release version attached to a year at any time, so adding "latest" would be redundant here - although it's unclear whether someone unfamiliar with this code will work this out and/or what would be most helpful to them.

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataBlockServicePermissionTests.cs
```diff
@@ -38,7 +38,7 @@ public class DataBlockServicePermissionTests
         public async Task Get()
         {
             await PolicyCheckBuilder<ContentSecurityPolicies>()
-                .SetupResourceCheckToFail(ReleaseVersion, ContentSecurityPolicies.CanViewSpecificRelease)
```
mmyoungman, line 41: Not a big deal, but you could have pulled this renaming out into it's own PR to keep this PR more focused - although appreciate if you're already half way through making changes when you decide to rename stuff, it probably wastes more of your time than it causes inconvenience to the reviewer.

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/ListReleaseVersionsRequest.cs
mmyoungman, line 1: Rename this to ReleaseVersionsType - I'd also move this outside the Requests directory

mmyoungman (commented):
I won't approve as I've not reviewed the frontend code

src/explore-education-statistics-admin/src/pages/publication/components/PublicationPublishedReleasesTable.tsx
```diff
@@ -92,11 +102,32 @@ export default function PublicationPublishedReleasesTable({
...
+                {release.permissions.canUpdateRelease && (
+                  <ReleaseLabelEditModal
+                    currentReleaseSlug={release.slug}
+                    publicationSlug={publication.slug}
+                    initialValues={{ label: release.label }}
+                    triggerButton={
+                      <ButtonText
+                        className={`govuk-!-display-inline-block ${
+                          release.permissions.canMakeAmendmentOfReleaseVersion
+                            ? ' govuk-!-margin-right-4'
```
Guy-HiveIT, line 114: I think this needs some margin left on it too to handle `No permission` case.
image

Or alternatively, wrap all the actions in a flex container with some gap and remove the custom margins on each button

src/explore-education-statistics-admin/src/components/FormModal.tsx
```diff
@@ -0,0 +1,182 @@
...
+import Modal from '@common/components/Modal';
+import useMountedRef from '@common/hooks/useMountedRef';
+import useToggle from '@common/hooks/useToggle';
+import { FieldMessageMapper } from '@common/validation/serverValidations';
+import React, { ReactNode, useState } from 'react';
+import { DeepPartial, FieldValues } from 'react-hook-form';
+import { ObjectSchema, Schema } from 'yup';
+
+interface Props<TFormValues extends FieldValues> {
+  children?: ReactNode;
```
bennettstuart, line 15: NIT, would probably say children are required

src/explore-education-statistics-admin/src/components/FormModal.tsx
```diff
@@ -0,0 +1,182 @@
...
+    | FieldMessageMapper<TFormValues>[]
+    | ((values: TFormValues) => FieldMessageMapper<TFormValues>[]);
+  submitText?: string;
+  cancelText?: string;
+  hiddenSubmittingText?: string;
+  onSubmit: (formValues: TFormValues) => Promise<void>;
+  withConfirmationWarning?: boolean;
+  confirmationWarningText?:
+    | ReactNode
+    | ((formValues?: TFormValues) => ReactNode);
```
bennettstuart, line 33: I think formValues will be defined, can remove `?`

src/explore-education-statistics-admin/src/pages/publication/components/ReleaseLabelEditModal.tsx
```diff
@@ -0,0 +1,138 @@
...
+  currentReleaseSlug,
+  publicationSlug,
+  initialValues,
+  triggerButton,
+  onSubmit,
+}: Props) {
+  const { publicAppUrl } = useConfig();
+
+  const confirmationWarningText = (
+    formValues?: ReleaseLabelFormValues,
```
bennettstuart, line 74: associated with another comment, we can make formValues required , as we _should_ only ever get to this step after a valid form was move into a confirmation state

src/explore-education-statistics-admin/src/pages/publication/components/ReleaseLabelEditModal.tsx
```diff
@@ -0,0 +1,138 @@
...
+    },
+  }),
+];
+
+const replaceWhitespace = (str: string, replaceValue: string) => {
+  return str.replace(/\s+/g, replaceValue);
+};
+
+const releaseSlugLabelSuffix = (label?: string) => {
+  if (isUndefinedOrWhitespace(label)) {
```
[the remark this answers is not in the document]
bennettstuart: I agree, this is a bit verbose, we're just used to doing something like 
```js
if(!label?.trim()) { //....
```

src/explore-education-statistics-admin/src/pages/publication/components/ReleaseLabelEditModal.tsx
```diff
@@ -0,0 +1,138 @@
...
+    return '';
+  }
+
+  const trimmedLowercaseLabel = label!.trim().toLowerCase();
+  const whitespaceReplacedWithDashesLabel = replaceWhitespace(
+    trimmedLowercaseLabel,
+    '-',
+  );
```
bennettstuart, line 59: Amy mentioned elsewhere to use lodash's kebabCase, which covers everything we need here
```js
import kebabCase from "lodash/kebabCase"

// ...
return `-${kebabCase(label)}`;
```

src/explore-education-statistics-admin/src/pages/publication/components/ReleaseLabelEditModal.tsx
```diff
@@ -0,0 +1,138 @@
...
+    const newReleaseSlugLabelSuffix = releaseSlugLabelSuffix(formValues?.label);
+
+    const newReleaseSlug = currentReleaseSlug.replace(
+      new RegExp(`${currentReleaseSlugLabelSuffix}$`),
+      newReleaseSlugLabelSuffix,
+    );
+
+    const formattedLabel = isUndefinedOrWhitespace(formValues?.label)
+      ? undefined
+      : replaceWhitespace(formValues!.label!, ' ');
```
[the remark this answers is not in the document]
bennettstuart: If we ensure the form validation ensures appropriate strings, we shouldn't need to format the label

src/explore-education-statistics-common/src/utils/string/isUndefinedOrWhitespace.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Check if a {@param str} is null or whitespace.
+ */
+export default function isUndefinedOrWhitespace(str?: string) {
+  return !str || !str.trim();
```
bennettstuart, line 5: NIT, suggested elsewhere this is verbose

also NIT, if we keep this we can simplify this to `return !str?.trim()`

bennettstuart (changes requested):
I've left a few comments and NITs

src/explore-education-statistics-admin/src/pages/publication/components/ReleaseLabelEditModal.tsx
```diff
@@ -0,0 +1,138 @@
...
+  initialValues?: ReleaseLabelFormValues;
+  triggerButton: ReactNode;
+  onSubmit: (releaseDetailsFormValues: ReleaseLabelFormValues) => Promise<void>;
+}
+
+const validationSchema: ObjectSchema<ReleaseLabelFormValues> = Yup.object({
+  label: Yup.string().max(
+    20,
+    /* eslint-disable no-template-curly-in-string */
+    'Release label must be no longer than ${max} characters',
```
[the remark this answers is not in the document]
jack-hive: I don't think I can for this, unfortunately. `max` isn't a local variable, it's injected by the validation framework.

src/explore-education-statistics-admin/src/pages/publication/components/ReleaseLabelEditModal.tsx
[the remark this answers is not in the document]
jack-hive: Would we prefer the validation to error and make them re-type it if they had unnecessary whitespace, or let the code be smart and help the user out by stripping it away for them?

I'm not very knowledgeable on UX, but as a minimum it feels like we should at least allow them to have leading/trailing whitespace, not error, and strip it off for them. The whitespace mid-string, however, could be argued against!

Would be good to get both your thoughts on this, @bennettstuart @amyb-hiveit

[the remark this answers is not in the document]
jack-hive: I've used Lodash's kebabcase to simplify one of the methods a little :) 

I think I still need the private method though, as it does contain a guard clause for null/whitespace labels, and also prepends the suffix with a '-'

src/explore-education-statistics-admin/src/pages/publication/components/ReleaseLabelEditModal.tsx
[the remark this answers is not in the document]
jack-hive: In C# we have a built-in method called `IsNullOrWhitespace`. So I kind of copied the concept over.

It essentially does the same thing as `!label?.trim()` like you've mentioned, but I personally think it makes the intent more clear. `IsUndefinedOrWhitespace` reads super clearly to me, but `!label?.trim()` takes some mental overhead to work out what it's doing (albeit small). It also requires the knowledge that `!label?.trim()` will resolve to `false` when `label` is an **empty string** = `''` - which, I do not think is entirely obvious. However, extracting that away into a convenience method like `IsUndefinedOrWhitespace` feels a lot neater to me

Let me know your thoughts!

src/explore-education-statistics-admin/src/pages/publication/components/ReleaseLabelEditModal.tsx
jack-hive: Thanks both! :)

src/explore-education-statistics-common/src/utils/string/isUndefinedOrWhitespace.ts
jack-hive: You're completely right! That's a nice little tidy-up, thank you :)

src/explore-education-statistics-admin/src/pages/publication/components/ReleaseLabelEditModal.tsx
jack-hive: Awesome, I've removed it :)

src/explore-education-statistics-admin/src/components/FormModal.tsx
jack-hive: Good point! Changed :)

src/explore-education-statistics-admin/src/components/FormModal.tsx
```diff
@@ -0,0 +1,182 @@
...
+        {({ formState }) => {
+          if (!formState.isValid) {
+            toggleSubmitting.off();
+            toggleConfirmationWarning.off();
+          }
+
+          return (
+            <Form id={formId} onSubmit={handleSubmit} onChange={onChange}>
+              {!confirmationWarning && children}
+              {confirmationWarning && <>{confirmationWarningTextWrapper()}</>}
```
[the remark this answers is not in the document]
jack-hive: I don't fully understand how screen readers work, or accessibility, so I'm just going to trust that this works! :P 

I've changed it, thank you

src/explore-education-statistics-admin/src/components/FormModal.tsx
```diff
@@ -0,0 +1,182 @@
...
+  initialValues?: TFormValues;
+  validationSchema?: ObjectSchema<TFormValues> & Schema<TFormValues>;
+  errorMappings?:
+    | FieldMessageMapper<TFormValues>[]
+    | ((values: TFormValues) => FieldMessageMapper<TFormValues>[]);
+  submitText?: string;
+  cancelText?: string;
+  hiddenSubmittingText?: string;
+  onSubmit: (formValues: TFormValues) => Promise<void>;
+  withConfirmationWarning?: boolean;
```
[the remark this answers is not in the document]
jack-hive: Fair point! I agree. Changed :)

src/explore-education-statistics-admin/src/components/FormModal.tsx
```diff
@@ -0,0 +1,182 @@
...
+  confirmationWarningText?:
+    | ReactNode
+    | ((formValues?: TFormValues) => ReactNode);
+}
```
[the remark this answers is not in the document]
jack-hive: Would you recommend doing this?:
```typescript
interface Props<TFormValues extends FieldValues> extends ModalProps, FormProviderProps<TFormValues>
```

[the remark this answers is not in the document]
jack-hive: Would we then have lots of extra types that we don't want?

src/explore-education-statistics-admin/src/pages/publication/components/PublicationPublishedReleasesTable.tsx
jack-hive: I've opted for the flex container option!

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServicePermissionTests.cs
```diff
@@ -384,18 +384,26 @@ public async Task UpdatePublication_NoPermissionToChangeSupersededById()
         }
 
         [Fact]
-        public void GetPublication()
+        public async Task GetPublicationAsync()
```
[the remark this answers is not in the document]
jack-hive: Good spot! I think this was an intellisense change, and I just didn't spot it!

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/DataBlockServicePermissionTests.cs
jack-hive: Yeah I think this was one of those that I didn't realise needed changing until half-way through!

But I agree, in hindsight, this might have been better in another PR. I'll leave it here for now though in the interest of getting it over the line :)

src/GovUk.Education.ExploreEducationStatistics.Admin/Requests/ListReleaseVersionsRequest.cs
jack-hive: Agreed!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/PublicationService.cs
jack-hive: You're absolutely correct! Very good spot. I like `NotPublished`, so I've changed it to that :)

src/explore-education-statistics-admin/src/pages/publication/components/ReleaseLabelEditModal.tsx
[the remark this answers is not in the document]
bennettstuart: Happy keep as is, as you say it's easier to read

src/explore-education-statistics-admin/src/pages/publication/components/ReleaseLabelEditModal.tsx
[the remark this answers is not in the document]
bennettstuart: Ah so the provided `20` (ln:27) will be injected to the message then. LGTM

src/explore-education-statistics-admin/src/pages/publication/components/ReleaseLabelEditModal.tsx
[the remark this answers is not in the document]
bennettstuart: for this snippet I might suggest a few tweaks:
```tsx
const formatWhitespace = (str: string) => { // was "replaceAllWhitespaceWithSingleSpace"
  return str.replace(/\s+/g, ' ').trim();
};

// ...

const formattedLabel = formValues.label
? formatWhitespace(formValues.label)
 : '';
```

I've suggested alteration to `replaceAllWhitespaceWithSingleSpace` to accomodate removal of leading and trailing whitespace also and renamed accordingly.

where we get the formatted label, we simply check if we have a label value (so it's defined and it will be a string) if so we format the whitespace, if not then we return an empty string, I've opted for empty string over undefined to keep the type unified, also with the variable being called "label" it can suggest it will be a string, us front-enders are used to treating empty strings as false values (feel free to keep undefined if you prefer)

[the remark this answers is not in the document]
bennettstuart: with regards to validation, I had assumed the label is required and so could never be undefined, thought I think this is not the case, ignore me :)

bennettstuart:
Happy with my comment resolutions, there's some input form me on other points though

src/explore-education-statistics-admin/src/pages/publication/components/PublicationPublishedReleasesTable.tsx
Guy-HiveIT: Nice, more manageable and flexible :) 

Just a minor point, could you use a GDS variable for the spacing rather than hardcoding 20px gap?
e.g. `#{govuk-spacing(2)};` 
If you look at the `<ButtonGroup>` component, that spacing seems to be the default (works out as 10px). I don't think it's too important if it's `spacing(4)` or `spacing(2)` but just good to keep things in variables where possible.

jack-hive: I didn't even realise this was an option. Thanks for pointing it out!

bennettstuart (approved):

PR #5726, opened 2025-03-27
Move EventGridClient to Common

leeoades:
In preparation for publishing event grid events in a variety of places throughout the system, this PR moves most of the event grid publishing code into common.
The introduction of the Configured Event Grid Client Factory allows for the configuration to be consistent across all of the services and abstracted away. Obtaining an event grid client now requires only a config key.
Also, I renamed EventGridPublisherClient to EventGridClient just because Publisher is something else in our system, and it was confusing.


src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Services/EventGrid/ConfiguredEventGridClientFactoryBuilder.cs
```diff
@@ -0,0 +1,37 @@
+﻿#nullable enable
+using GovUk.Education.ExploreEducationStatistics.Common.Services.EventGrid;
+using Moq;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Tests.Services.EventGrid;
+
+public class ConfiguredEventGridClientFactoryBuilder
```
tomjonesdev, line 7: Minor DX point since `Builder` is a common suffix in codebases, any issue naming these classes something like `...MockBuilder` to help identify them easier?

mmoalin (approved):

PR #5729, opened 2025-03-27
Cleaning up unit tests

leeoades:
Renamed xxxBuilder to xxxMockBuilder - but removed MockBuilder suffix from usages - since all dependencies in the unit tests are mocks so seems superfluous to keep stating it.

Cleaned up PublishingCompletionServiceTests

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/Services/PublishingCompletionServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/Services/PublishingCompletionServiceTests.cs
@@ -73,7 +73,7 @@ public class PublishingCompletionServiceTests
                     .WhereFilesStatusIs(ReleasePublishingStatusFilesStage.NotStarted)
                     .Build();
                 
-                var readyKeys = new List<ReleasePublishingKey>()
+                var notReadyKeys = new List<ReleasePublishingKey>()
                 {
                     releasePublishingKey1, 
                     releasePublishingKey2
@@ -86,7 +86,7 @@ public class PublishingCompletionServiceTests
                 var sut = GetSut();
 
                 // ACT
-                await sut.CompletePublishingIfAllPriorStagesComplete(readyKeys);
+                await sut.CompletePublishingIfAllPriorStagesComplete(notReadyKeys);
 
                 // ASSERT
                 _releasePublishingStatusService.Assert.UpdatePublishingStageWasNotCalled();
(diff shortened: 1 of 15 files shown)


src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/Services/PublishingCompletionServiceTests.cs
```diff
@@ -91,129 +73,94 @@ public async Task WhenNoPublicationsReady_ThenNothingHappens()
...
-                }
-
-                _releaseVersion1 = new ReleaseVersionBuilder(ReleaseVersionId1)
+                // Create some sample release versions
+                _releaseVersion1 = new ReleaseVersionBuilder()
                     .WithPublicationId(PublicationId1)
                     .ForRelease(release => release
-                        .WithReleaseId(ReleaseId1)
-                        .WithReleaseSlug(_releaseSlug1))
+                        .WithReleaseId(Guid.NewGuid())
```
tomjonesdev, line 123: Could this extension be converted to generate `Guid.NewGuid()` when omitted from the caller? Including an argument implies you want the `releaseId` to be a specific value, but the usage here implies the actual ID doesn't matter (e.g. it's not using a pre-defined variable).

src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/Services/PublishingCompletionServiceTests.cs
```diff
@@ -91,129 +73,94 @@ public async Task WhenNoPublicationsReady_ThenNothingHappens()
                     .WhereFilesStatusIs(ReleasePublishingStatusFilesStage.NotStarted)
                     .Build();
                 
-                var releasePublishingKeys = new List<ReleasePublishingKey>()
+                var readyKeys = new List<ReleasePublishingKey>()
```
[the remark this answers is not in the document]
leeoades: Yes Sir! Well spotted!

src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/Services/PublishingCompletionServiceTests.cs
leeoades: Ha! Indeed. So it's because I was inlining the original values and changing them to be NewGuids instead of actual ones. You are correct and I've removed these lines.

PR #5731, opened 2025-03-28
When a theme is updated, raise a ThemeUpdated event

leeoades:
As part of the on-going work with the new Search functionality, we need to signal whenever anything changes that affects the data that is stored and indexed by the search indexer.

The Searchable Document representation of a Release Version contains some Theme information. Therefore, if a Theme changes, we need to update any Searchable Documents that are a part of that Theme.

In this PR, we are raising an event when a Theme is updated. A future piece of work will subscribe to that event, look up the publications and therefore release versions that fall under that theme, and have the corresponding Searchable Documents updated.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin/Events/ThemeChangedEventDto.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin/Events/ThemeChangedEventDto.cs
@@ -1,0 +1,46 @@
+﻿using Azure.Messaging.EventGrid;
+using GovUk.Education.ExploreEducationStatistics.Content.Model;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Events;
+
+public record ThemeChangedEventDto
+{
+    public ThemeChangedEventDto(Theme theme)
+    {
+        Subject = theme.Id.ToString();
+        Payload = new EventPayload
+        {
+            Title = theme.Title,
+            Summary = theme.Summary,
+            Slug = theme.Slug
+        };
+    }
+
+    // Changes to this event should also increment the version accordingly.
+    public const string DataVersion = "1.0";
+    public const string EventType = "theme-changed";
+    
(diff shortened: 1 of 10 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin/Services/AdminEventRaiserService.cs
```diff
@@ -0,0 +1,36 @@
...
+using GovUk.Education.ExploreEducationStatistics.Admin.Events;
+using GovUk.Education.ExploreEducationStatistics.Common.Services.EventGrid;
+using GovUk.Education.ExploreEducationStatistics.Content.Model;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Services;
+
+public interface IAdminEventRaiserService
+{
+    Task OnThemeUpdated(Theme theme);
+}
```
benoutram, line 11: We tend to put all interfaces into separate files, one interface per file, and put them in a separate namespace so they don't create clutter. Or at least I think that was our original intention for doing it this way! 

I didn't mention this before given that you were authoring new code in new projects but thought I'd point it out here given that it's in the Admin. In this case you should see lots of interfaces in `Admin.Services.Interfaces`.

src/GovUk.Education.ExploreEducationStatistics.Admin/Events/ThemeChangedEventDto.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    // Changes to this event should also increment the version accordingly.
+    public const string DataVersion = "1.0";
+    public const string EventType = "theme-changed";
+    
+    // Which Topic endpoint to use from the appsettings
+    public const string EventTopicOptionsKey = "ThemeChangesEvent";
+
+    /// <summary>
+    /// The ReleaseVersionId is the subject
```
benoutram, line 27: Should read 'The theme id is the subject' I think?

PR #5733, opened 2025-03-28
EES-5940 Add infrastructure to deploy an Azure AI Search data source, index and indexer

benoutram:
This PR adds the Bicep infrastructure to create or update a Azure AI Search data source, index and indexer. The result of this is that we now have an Azure AI Search API endpoint that we can run search queries against.

It builds on #5668 which added the infrastructure to create the Azure AI Search Service and Search Storage account/container but did not go as far as configuring the Storage container as a data source within the Azure AI Search, or define the index or indexer.

### Deployment Script resource

There's no Bicep template support for Azure AI Search data plane operations like creating an data source or an index.

Instead, we create an Azure Deployment Script resource to use data plane operations of the Azure AI Search REST API. The Deployment Script uses a Container Instance to execute a Powershell script from a temporary Storage Account created during the deploy. To allow for debugging the script output, the account has been configured with a retention period of 1 hour and will be deleted on expiration.

See [Use deployment scripts in Bicep](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/deployment-script-bicep) to learn more about using Deployment Script resources.

The deployment script has been set up with a managed identity and been assigned the Search Service Contributor role on the Azure AI Search service. This uses Microsoft Entra ID authentication and RBAC to provide a bearer token in the authorization header of API requests.

This script code was originally based off the Quickstart [Bicep deployment of Azure Cognitive Search](https://github.com/Azure-Samples/azure-search-deployment-template) template. I have adapted this to add additional naming and schedule parameters and make it so the indexer can be disabled by default if necessary.

The Powershell script has been written to support containers using multiple types of data source but for now only Azure Blob Storage is supported.

In order to make the 'searchServiceConfig' Bicep template a reusable component with a configurable index definition I replaced the hardcoded Powershell object definition of the index with a supporting script file. This uses the Deployment Script  `supportingScriptUris` property to supply the file in a relative directory to the primary script. See [Develop a deployment script in Bicep](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/deployment-script-develop) to learn more about the script properties and arguments.

### Prerequisites for creating Deployment Script resources

Creating an Azure Deployment Script resource requires the Azure Storage and Azure Container Instances resource providers to be registered.

In our case the `Microsoft.ContainerInstance` resource provider was not registered on the subscription.

image (3)

Registering the resource provider at subscription level is a manual step that needs to be done before attempting the deploy in other environments.

### Data source

image

The data source is configured for the 'searchable documents' container in the Search Storage Account blob service.

Making this accessible required ensuring the Azure AI Search service has a managed identity and assigning it the role 'Storage Blob Data Reader' on the Storage Account. It has the name `azureblob-${searchableDocumentsContainerName}-datasource`.

Since it uses RBAC with a managed identity it allows us to configure the data source with a credential-less connection string with no account key or password. See [Connect to Azure Storage using a managed identity](https://learn.microsoft.com/en-us/azure/search/search-howto-managed-identities-storage).

It is configured with the native soft deletion detection policy which should work since the Search Storage account has already been set up with soft delete for blobs enabled.

### Search index

image (1)

The index named `index-1` is created from the JSON definition provided in the base definitions directory. The index provided with this change is only an example with a single field. See `infrastructure/templates/search/application/indexes/index-1.json`.

This definition will be updated by future work.

### Indexer

image (2)

The indexer is configured for the data source, targeting the search index. It has a default name of `${indexName}-indexer`.

The [Change and delete detection using indexers for Azure Storage in Azure AI Search](https://learn.microsoft.com/en-us/azure/search/search-howto-index-changed-deleted-blobs) documentation says:

> For indexed content that originates from Azure Storage, change detection occurs automatically because indexers keep track of the last update using the built-in timestamps on objects and files in Azure Storage.

I wasn't sure if this happens automatically even if the indexer is only run once so for now it has been configured with a schedule of PT5M (every five minutes).

### DIFF
--- a/infrastructure/templates/search/ci/azure-pipelines.yml
+++ b/infrastructure/templates/search/ci/azure-pipelines.yml
@@ -13,7 +13,14 @@ pr: none
 
 parameters:
   - name: deployAlerts
-    displayName: Do the Azure Monitor alerts need creating or updating?
+    displayName: Do the Azure Monitor alerts need creating/updating?
+    type: string
+    values:
+      - Yes
+      - No
+    default: No
+  - name: deploySearchConfig
+    displayName: Does the Azure Search configuration need creating/updating?
     type: string
     values:
       - Yes
@@ -55,6 +62,8 @@ variables:
     value: ubuntu-latest
   - name: deployAlerts
     value: ${{ iif(eq(parameters.deployAlerts, 'Yes'), true, false) }}
+  - name: deploySearchConfig
+    value: ${{ iif(eq(parameters.deploySearchConfig, 'Yes'), true, false) }}
 
 pool:
   vmImage: $(vmImageName)
--- a/infrastructure/templates/public-api/components/storageAccount.bicep
+++ b/infrastructure/templates/public-api/components/storageAccount.bicep
@@ -181,6 +181,7 @@ module storeAccessKeyToKeyVault './keyVaultSecret.bicep' = {
   }
 }
 
+output storageAccountId string = storageAccount.id
 output storageAccountName string = storageAccount.name
 output connectionStringSecretName string = connectionStringSecretName
 output accessKeySecretName string = accessKeySecretName
--- a/infrastructure/templates/search/application/indexes/index-1.json
+++ b/infrastructure/templates/search/application/indexes/index-1.json
@@ -1,0 +1,16 @@
+{
+  "name": "index-1",
+  "fields": [
+    {
+      "name": "rid",
+      "type": "Edm.String",
+      "searchable": true,
+      "filterable": true,
+      "retrievable": true,
+      "stored": true,
+      "sortable": true,
+      "facetable": true,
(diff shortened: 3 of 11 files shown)


infrastructure/templates/search/ci/azure-pipelines.yml
```diff
@@ -19,6 +19,13 @@ parameters:
       - Yes
       - No
     default: No
+  - name: deploySearchConfig
+    displayName: Does the Azure Search configuration need creating or updating?
```
leeoades, line 23: Minor: This display name sounds like a choice - does it need creating or does it need updating? I'm assuming you're meaning "does it need creating/updating?"

leeoades (approved):
Oh my days. Well done with this. I'm assuming you've been running this over and over again in order to get it finely tuned.

PR #5734, opened 2025-03-31
EES-5969 SEO noindex for query params

Guy-HiveIT:
This PR adds middleware for the frontend site to prevent duplicate indexing of pages with different query params.
We check if there are query params present in the request, and add a `noindex` header to the response if there are, unless the only query param is `page`, which we still want to be indexed.

I considered adding a noindex meta tag to the document `head` as an alternative, but this would need to be done in multiple templates in `getServerSideProps` and didn't seem as clean or global as a solution.


bennettstuart (approved):
I love this solution

PR #5736, opened 2025-03-31
EES-5964 Change Search infrastructure pipeline to use common Deploy Function App template

benoutram:
This PR alters the  Search infrastructure pipeline to use the common Deploy Function App template.

Instead of using `deploy-search-docs-function-app.yml` it now uses `common/ci/tasks/deploy-function-app.yml` which was originally based from it.

This has made some minor changes to the common template to parameterise the download build pipeline so that it is not always 'MainBuild', filter the downloaded artifact folder by the deployment zip filename, and make the step display names more concise.

### Other changes

Format YAML in `search/ci/azure-pipelines.yml` and `search/ci/azure-pipelines-build.yml`.


infrastructure/templates/search/ci/stages/deploy.yml
```diff
@@ -47,8 +47,15 @@ stages:
...
           serviceConnection: ${{ parameters.serviceConnection }}
           environment: ${{ parameters.environment }}
-          dependsOn: DeploySearchInfrastructure
\ No newline at end of file
+          displayName: Deploy Search Docs Function App
+          dependsOn: DeploySearchInfrastructure
+          deploymentName: DeploySearchDocsFunction
+          downloadPipeline: SearchBuild
+          functionAppName: $(searchDocsFunctionAppName)
+          artifactName: drop
```
leeoades, line 59: What is this value? Seems a bit weird?

leeoades (approved):

infrastructure/templates/search/ci/stages/deploy.yml
benoutram: Do you mean `artifactName: drop`?

I'm not sure if it's DevOps terminology or a Microsoft term but the Azure DevOps documentation for publishing build artifacts to share by later stages or (as in our case) by a different pipeline uses the name 'drop'. It's used as the name of the folder for the set of build artifact outputs.

See the documentation with code samples in [Publish and download pipeline artifacts](https://learn.microsoft.com/en-us/azure/devops/pipelines/artifacts/pipeline-artifacts?view=azure-devops&tabs=yaml).

It's also mentioned as an example name to use in the documentation of the [Publish Pipeline Artifacts v1 task](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/publish-pipeline-artifact-v1?view=azure-pipelines) task

> Specifies the name of the artifact to publish. It can be any name you choose, for example drop. If not set, the default is a unique ID scoped to the job.

So I just decided to go with it. 🙂 

In our build pipeline configuration we do a `dotnet build` followed by a `dotnet test` on the solution.
Then we have individual tasks doing a `dotnet publish` for the specific projects that we want to publish. Currently this is only the Search Function App. These are output to the pipeline's artifact staging directory.

The Publish Pipeline Artifacts task then uploads all files/folders from the staging directory as an artifact named 'drop'.

The main (non-Search) build pipeline does this differently by calling the Publish Pipeline Artifacts task many times, each time for a different _project_ directory within the artifact staging directory , giving each upload a different project name as the artifact name. Imo this seems like it might be unnecessary.

PR #5747, opened 2025-04-02
EES-5974 infrastructure changes for search function app related to EES-5875

benoutram:
This PR makes infrastructure changes for the Search Azure Function App to support changes made to it in EES-5875 by #5703. These changes are possible now that the configuration for the Azure AI Search resource was added by #5733.

- Rename the `indexName` app setting and parameter names to `indexerName` to match Azure's SearchIndexerClient class since the operation to retrieve indexer information requires the indexer name rather than the index name.
- Configure the app with the Azure AI search REST API endpoint.
- Configure the app with the Azure AI search indexer name.
- Enable the app's search indexer health check by uncommenting `AzureSearchHealthCheckStrategy`.
- Assign the app's system managed identity the 'Search Service Contributor' role required to list indexers.
- Break out a new `searchServiceRoleAssignment` template and refactor `searchServiceConfig` to use it.

Updated health check response after deploying these changes:

image

### DIFF
--- a/infrastructure/templates/common/components/functionApp.bicep
+++ b/infrastructure/templates/common/components/functionApp.bicep
@@ -77,12 +77,8 @@ param allowedOrigins array = []
 @description('Specifies an optional URL for Azure to use to monitor the health of this resource')
 param healthCheckPath string?
 
-@description('An existing Managed Identity\'s Resource Id with which to associate this Function App.')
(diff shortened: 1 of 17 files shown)


leeoades (approved):
Absolutely amazing stuff, Ben! Ship it!!

PR #5752, opened 2025-04-03
Fix vulnerability in System.Text.Json

leeoades:
Lock all of the projects that had an implicit reference to System.Text.Json 7.0.0 to 8.0.5. 
The vulnerability was fixed in 8.0.4. See https://github.com/advisories/GHSA-hh2w-p6rv-4g7w


mmoalin:
Great spot and reaction! 💪

mmoalin (approved):

PR #5754, opened 2025-04-03
Ees 5954 On Publication Changed

leeoades:
Note: There are 4 steps to this PR which I have left as easy to understand separate commits

1. Renamed Mock Builder asserts from xxCalled to xxRaised to refer to the events instead of the methods

2. Very subtle change: I had originally named the Topics xxxChanges to signify that it was a stream of changes. But it's too easy to transpose xxxChanged which is the name of the events and event handler etc

3. Add new Event Raiser function - to raise a Publication Changed event

4. Utilise the new event in Publication Service. Update some tests to also check for the events (or lack of them where appropriate).

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
@@ -21,6 +21,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using GovUk.Education.ExploreEducationStatistics.Admin.Tests.MockBuilders;
 using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Security.SecurityPolicies;
 using static GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services.DbUtils;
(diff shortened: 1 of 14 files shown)


src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
```diff
@@ -32,3441 +33,3458 @@
 using IReleaseVersionRepository = GovUk.Education.ExploreEducationStatistics.Content.Model.Repository.Interfaces.IReleaseVersionRepository;
 using ReleaseVersionRepository = GovUk.Education.ExploreEducationStatistics.Content.Model.Repository.ReleaseVersionRepository;
 
-namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Tests.Services;
+
+public class PublicationServiceTests
 {
```
benoutram, line 36: This goes against one of our backend style guide rules because this change to use files scoped namespace is being made where there's something like a 28 line change here turning into a 5678 line change.

>  Change existing files to use file scoped namespace declarations only if there are significant line changes or if files are small, to minimise the impact on readability of pull requests.

We've discussed this as a team previously at Dev team meetings and  we do eventually want to switch all files to use file scoped namespace but for now we agreed to only use it on new files or switch where the majority of the existing code is changing.

As well as minimising the impact on readability of pull requests, it's also to reduce the chance of causing a conflict with anyone else with changes to the same code already in flight, and because we appreciate the history and being able to tell what issues certain lines refer to.

The same comment applies to PublicationService.cs where there's even less lines changing.

src/GovUk.Education.ExploreEducationStatistics.Admin/Events/PublicationChangedEventDto.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    // Changes to this event should also increment the version accordingly.
+    private const string DataVersion = "1.0";
+    private const string EventType = "publication-changed";
+    
+    // Which Topic endpoint to use from the appsettings
+    public const string EventTopicOptionsKey = "PublicationChangedEvent";
+
+    /// <summary>
+    /// The ThemeId is the subject
```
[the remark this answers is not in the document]
leeoades: Oh for crying out loud... 🤣

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/PublicationServiceTests.cs
leeoades: I'm guessing this was discussed at a meeting I wasn't at?

- It barely affects the PR if you turn off whitespace changes
- It won't affect other PRs if everyone is also converting the files as we go

At the moment, nothing is changing so we don't move towards our goal at all. We just kick the can down the road, no?

I'm happy to raise this in a meeting but it's difficult to if I'm not invited to those meetings.

benoutram: Yes it's been discussed prior to you starting at the Dev's team meeting, probably on multiple occasions I think. Those conversations led us to having that reference in the style guide about it.

So far we've either not reached an agreement or found a convenient time on if/when we should go ahead with the conversion of all files.

The Dev's team meeting would be the place to discuss it again and you should be invited to that?

I don't think you need to undo this change, although you could if you want, just to fit in with the style guide. I wouldn't make further changes like this until we've had chance to discuss it again though.

benoutram (approved):

PR #5760, opened 2025-04-04
EES-5977 Add analytics for zip downloads

mmyoungman:
This PR starts recording analytics for zip downloads.It essentially copies the pattern established by EES-5242 when we started recording analytics for Public API queries.

As part of this work we explored whether the the EES-5242 approach is suitable for both this and for future analytics work where things that might be requested more frequently like release page views) and gave it the OK. If a performance problem arises we could at the very worst turn analytics off and would be relatively straightforward to change approach - but we don't suspect this will be necessary.

The high-level overview is that every time a user downloads a zip file from either a release page or a data set page, a file is created to record the `releaseVersionId` and `fileIds` associated with that zip download. (A request with no `fileIds` indicates that the zip file contains all data files associated with that releaseVersion.) These files are then consumed by a new Azure function that processes these files once an hour and generates parquet file reports. These will be used by the DfE for their own purposes. This may be considered a strange approach, but the end users of these files are statisticians and used to working with data in this way.

Locally, the files and reports will be created in the `explore-education-statistics/data/analytics` directory - this can be edited in the Content API's `appsettings.Development.json`. On environments, they will be written to the analytics Azure Storage instance in a file share. This fileshare is mounted with a file path `\\mounts\\analytics` - this is `analyticsFileShareMountPath` in the main ARM template (infrastructure/template.json).

The writing of the original files to record the zip download are written using a `Channel`. Basically, we defer the creation of the files to a background task to avoid it interfering with the performance of more important tasks. `AnalyticsManager` creates the `Channel`. `AnalyticsWriter` writes request data into the channel, and then `AnalyticsConsumer` reads from the channel as a background task and creates the actual files (which will later be processed by the Azure Function and turned into a parquet file report).

The Azure Function is hooked up to an Azure Storage instance (`eessaanlyt`) and that is where the parquet files are saved.

### Other changes
- I've added Dispose methods to various analytic test classes to ensure temporary files get cleared up after running the unit tests.
- In the datetime in filenames of files to be consumed by the Azure Function, I've added extra fractions of a second, now seven digits instead of three. This is because `WriteAllTextAsync` overwrites files by default, so this is to try and avoid that. I've also added a check to throw an exception if any files with the same name already exists.
- I renamed some `Noop*` classes into `NoOp*`.


mmyoungman:
I will remake this later

PR #5764, opened 2025-04-07
EES-5954 Only validate date input field changes after form is submitted

Guy-HiveIT:
This PR changed the `FormFieldDateInput` component to only validate onChange input **after** the form has been submitted.
This prevents annoying inline validation when tabbing through the different date input text fields before the form has been submitted. 

It also amends the release status form, to separate out logic for actually submitting the release status form vs showing the confirm modal.

A side effect of this amend was to have to not use the form's submit button as the confirm modal trigger, as it was preventing form validation on click.
As a result I've had to amend how we handle returning focus when the modal is closed - this is by imperatively calling 'focus()' after a timeout (necessary as without a timeout the focus() call doesn't work).
**I couldn't use the `useEffect` method from other examples as it was focussing the submit button as soon as the form was mounted.**

I also had to amend the dateinput field tests, to not expect errors after simply tabbing through the different fields.

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/components/ReleaseStatusForm.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/components/ReleaseStatusForm.tsx
@@ -33,7 +33,7 @@ import Yup from '@common/validation/yup';
 import { endOfDay, format, isValid, parseISO } from 'date-fns';
 import keyBy from 'lodash/keyBy';
 import mapValues from 'lodash/mapValues';
-import React, { useMemo } from 'react';
+import React, { useMemo, useRef } from 'react';
 import { ObjectSchema } from 'yup';
 
 export interface ReleaseStatusFormValues {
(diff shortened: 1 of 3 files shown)


src/explore-education-statistics-admin/src/pages/release/components/ReleaseStatusForm.tsx
```diff
@@ -99,6 +99,38 @@ const ReleaseStatusForm = ({
...
   const [showScheduleErrorModal, toggleScheduleErrorModal] = useToggle(false);
 
+  const submitButtonRef = useRef<HTMLButtonElement>(null);
+
+  const closeConfirmScheduleModal = () => {
+    toggleConfirmScheduleModal.off();
+    setTimeout(() => {
+      submitButtonRef.current?.focus();
+    }, 1);
+  };
```
bennettstuart, line 109: Is there a reason we're avoiding the `useEffect` on `showScheduleErrorModal` like we've done elsewhere?
```tsx
useEffect(()=>{
  if(!showScheduleErrorModal) { submitButtonRef.current?.focus(); }
},[showScheduleErrorModal, submitButtonRef]);
```

bennettstuart (changes requested):

src/explore-education-statistics-admin/src/pages/release/components/ReleaseStatusForm.tsx
bennettstuart, line 109: We _could_ opt for the useEffect if we
1. change `useToggle(false)` to `useToggle()` this would set the initial state to `undefined`
2. check `showScheduleErrorModal` is `=== false` instead of `if(!showScheduleErrorModal)` in the useEffect

Personally I think it's better to avoid setTimeout where possible to keep within the react lifecycle

bennettstuart (changes requested):

src/explore-education-statistics-admin/src/pages/release/components/ReleaseStatusForm.tsx
Guy-HiveIT: I see what you mean but I don't think that approach is possible - `useToggle` needs a boolean value initially - it is always true or false.
I could switch it to not use `useToggle` at all and manage it with plain `useState` with `(boolean | undefined)` type but seems better to me to stick with useToggle.

Let me know what you think.

image

bennettstuart: My mistake!
I think let's keep this as is

bennettstuart (approved):

PR #5765, opened 2025-04-07
EES-5741 - Filter Hierarchies UI & Layout

bennettstuart:
Takes the Filter Hierarchy work further with UI layout bells & whistles.
Notable changes:
- includes feature flag search param `fh` appending `?subjectId=....&fh=true` will show filter hierarchies (when hierarchied data is used in the table tool)
- UI dropdowns & modal with associated styling
- Sub category tier level search for tiers with 7 or more options to filter long option lists
- Refactor / reverted option render method to go for a option tree generation first approach to cater for upcoming search (option tree filtering) functionality

image
image

### DIFF
--- a/src/explore-education-statistics-common/src/modules/table-tool/components/FilterHierarchy.module.scss
+++ b/src/explore-education-statistics-common/src/modules/table-tool/components/FilterHierarchy.module.scss
@@ -1,4 +1,23 @@
-.fhList {
-  margin: 0;
-  padding: 0;
+@import '~govuk-frontend/dist/govuk/base';
+
+.tier {
+  background: govuk-colour('light-grey');
+  padding: 0.5rem;
+  margin-bottom: 1.5rem;
+}
+
+.subTier {
+  position: relative;
+  margin-top: 3rem;
+
+  &::before {
+    content: '';
+    border-top: 1.2rem solid govuk-colour('blue');
+    border-left: 3rem solid transparent;
+    border-right: 3rem solid transparent;
+    left: 50%;
+    position: absolute;
+    top: -1.5rem;
+    transform: translate(-50%, -50%);
+  }
 }
--- a/src/explore-education-statistics-common/src/modules/table-tool/components/FilterHierarchy.module.scss
+++ b/src/explore-education-statistics-common/src/modules/table-tool/components/FilterHierarchy.module.scss
@@ -2,22 +2,22 @@
 
 .tier {
   background: govuk-colour('light-grey');
-  padding: 0.5rem;
-  margin-bottom: 1.5rem;
+  padding: govuk-spacing(2);
+  margin-bottom: govuk-spacing(5);
 }
 
 .subTier {
   position: relative;
-  margin-top: 3rem;
+  margin-top: govuk-spacing(8);
 
   &::before {
     content: '';
-    border-top: 1.2rem solid govuk-colour('blue');
-    border-left: 3rem solid transparent;
-    border-right: 3rem solid transparent;
+    border-top: govuk-spacing(4) solid govuk-colour('blue');
+    border-left: govuk-spacing(8) solid transparent;
+    border-right: govuk-spacing(8) solid transparent;
     left: 50%;
     position: absolute;
-    top: -1.5rem;
+    top: govuk-spacing(-5);
     transform: translate(-50%, -50%);
   }
 }
--- a/src/explore-education-statistics-common/src/modules/table-tool/components/FilterHierarchy.module.scss
+++ b/src/explore-education-statistics-common/src/modules/table-tool/components/FilterHierarchy.module.scss
@@ -1,0 +1,4 @@
+.fhList {
+  margin: 0;
+  padding: 0;
+}
--- a/src/explore-education-statistics-common/src/modules/table-tool/components/FilterHierarchy.tsx
+++ b/src/explore-education-statistics-common/src/modules/table-tool/components/FilterHierarchy.tsx
@@ -4,10 +4,11 @@ import { useFormIdContext } from '@common/components/form/contexts/FormIdContext
 import FormCheckboxSelectedCount from '@common/components/form/FormCheckboxSelectedCount';
 import { SubjectMetaFilterHierarchy } from '@common/services/tableBuilderService';
 import get from 'lodash/get';
-import React, { memo } from 'react';
+import React, { memo, useCallback, useState } from 'react';
 import { useFormContext, useWatch } from 'react-hook-form';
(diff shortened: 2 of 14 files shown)


src/explore-education-statistics-common/src/modules/table-tool/components/FilterHierarchy.module.scss
```diff
@@ -0,0 +1,23 @@
+@import '~govuk-frontend/dist/govuk/base';
+
+.tier {
+  background: govuk-colour('light-grey');
+  padding: 0.5rem;
```
amyb-hiveit, line 5: Could you use the `govuk-spacing` for the spacings in this file and in `FilterHierarchyOptions.module.scss`?

src/explore-education-statistics-common/src/modules/table-tool/components/FilterHierarchy.tsx
```diff
@@ -51,33 +80,148 @@ function FilterHierarchy({
...
+              expandedOptionsList={expandedOptionsList}
+            />
+          );
+        })}
+      </FormFieldset>
     </DetailsMenu>
   );
 }
 
+function sortAlphabeticalTotalsFirst(a: string, b: string) {
```
amyb-hiveit, line 166: Is this a duplicate of https://github.com/dfe-analytical-services/explore-education-statistics/blob/dev/src/explore-education-statistics-common/src/utils/sortAlphabeticalTotalsFirst.ts ?

src/explore-education-statistics-common/src/modules/table-tool/components/FilterHierarchyOptions.tsx
```diff
@@ -1,78 +1,155 @@
...
-            level={level + 1}
-          />
-        ))}
+      <div className={styles.checkbox}>
+        <FormField
+          hint={optionTree.filterLabel}
+          label={optionTree.label}
+          id={`${name}-${optionTree.value}`}
+          // @ts-expect-error  `value` required to make checkboxes unique
+          value={optionTree.value}
```
amyb-hiveit, line 92: I really think there's got to be a better way to do this. It's like react hook form isn't assigning the value as it does for other checkboxes so there's got to be something causing that. But, that said, I've had a quick go and can't figure it out.

src/explore-education-statistics-common/src/modules/table-tool/components/FilterHierarchyOptions.tsx
```diff
@@ -1,78 +1,155 @@
...
+          name={name}
+          as={FormCheckbox}
+          checked={selectedValues.includes(optionTree.value)}
+          disabled={disabled}
+        />
       </div>
+      {!!optionTree.options?.length && (
+        <DetailsMenu
+          summary={`${
```
amyb-hiveit, line 101: These need some screen reader text to differentiate them, otherwise there's multiple buttons with the same text.

src/explore-education-statistics-common/src/modules/table-tool/components/FilterHierarchyOptions.tsx
```diff
@@ -1,78 +1,155 @@
...
+                />
+              </div>
+            )}
+            <div>
+              {filteredOptions.length > 1 && (
+                <ButtonText
+                  className={styles.selectAllButton}
+                  onClick={toggleSelectAll}
+                >
+                  {hasAllSelected ? 'Unselect' : 'Select'} all{' '}
```
amyb-hiveit, line 133: These buttons also need some screen reader text to differentiate them.

amyb-hiveit (commented):
It'd be good to have some tests for the hierarchy components as there's a lot going on.

src/explore-education-statistics-common/src/modules/table-tool/components/FilterHierarchy.tsx
bennettstuart: Yes it was, I thought the original was removed in previous refactor, good spot!

src/explore-education-statistics-common/src/modules/table-tool/components/FilterHierarchyOptions.tsx
bennettstuart: There definitely was, it took a while for me to even get this working, I've now since improved it I'd say, this new implementation seems so obvious, not sure how I missed it.

src/explore-education-statistics-common/src/modules/table-tool/components/FilterHierarchyOptions.tsx
bennettstuart: I've added some visually hidden text which should read as follows 
```
"Show sector subject area [options for Entry level]"
```

src/explore-education-statistics-common/src/modules/table-tool/components/FilterHierarchyOptions.tsx
bennettstuart: I've added some visually hidden text which should read as follows
```tsx
"Unselect all 2 options" // (screen)
"Unselect all 2 [sector subject area options for entry level]" // (screen reader)
```

amyb-hiveit (commented):
Thanks for addressing those comments @bennettstuart, that all looks good. I have just noticed another thing though - the selected filters aren't shown when you're on the final step, ie. without feature flag:
Screenshot 2025-04-17 114201
With feature flag:
Screenshot 2025-04-17 114151

Also, it looks like the snapshot tests need to be updated because of the new class on checkboxes (if you've not come across them before you just run `pnpm test:snapshot` , check the results are what you'd expect and commit it).

bennettstuart:
Good spot @amyb-hiveit , I've [spoken to cam and gone with this fix for now regarding selected filters list](https://exploreeducationstats.slack.com/archives/CE3RSDYSV/p1746609797674569?thread_ts=1744902072.782649&cid=CE3RSDYSV) with the plan to further refine this in the following ticket 6019

benoutram (approved):

src/explore-education-statistics-common/src/modules/table-tool/components/FiltersForm.tsx
```diff
@@ -162,12 +161,32 @@ export default function FiltersForm({
...
+
+        return [
+          hierachyKey,
+          initialValues?.filters.filter(
+            optionId => !hierarchyOptionValues.includes(optionId),
+          ) ?? [],
+        ];
+      }),
+    );
+
```
bennettstuart, line 183: This is the fix for the filter hierarchies, when you'd go back to the previous step and it would select Total for you. A lack of initial value was causing the issue

PR #5769, opened 2025-04-08
EES-5948 Improve accessibility support for CKEditor

Guy-HiveIT:
Improve experience using CKEditor for keyboard and screen reader users by using CKEditor's Accessibility Help plugin - https://ckeditor.com/docs/ckeditor5/latest/features/accessibility.html#displaying-keyboard-shortcuts-in-the-editor

- Bump ckeditor deps from 41.0 to 41.2 (when the accessibility plugin was added). 
- Add accessibility help item to toolbar.


amyb-hiveit (commented):
Looks good, just one small thing - we have a very cut down version (just the add link button) of ck editor on the footnotes form so it'll need to be added there as well, it uses `toolbarConfigLinkOnly`.

Guy-HiveIT:
Good catch thanks, have added it there too.

amyb-hiveit (approved):

PR #5772, opened 2025-04-09
EES-5975 Infrastructure changes for Admin related to EES-5953, EES-5954, EES-5955, EES-5956, EES-5962

benoutram:
This PR makes an infrastructure change to the Admin app settings to support changes made to it in EES-5953, EES-5954, EES-5955, EES-5956, and EES-5962. These added the ability to publish events to Event Grid custom topics.

This PR adds the following app settings:
- `EventGrid__EventTopics__0__Key` - `PublicationChangedEvent`
- `EventGrid__EventTopics__0__TopicEndpoint` - `<Event Grid custom topic endpoint>`
- `EventGrid__EventTopics__1__Key` - `ReleaseChangedEvent`
- `EventGrid__EventTopics__1__TopicEndpoint` - `<Event Grid custom topic endpoint>`
- `EventGrid__EventTopics__2__Key` - `ThemeChangedEvent`
- `EventGrid__EventTopics__2__TopicEndpoint` - `<Event Grid custom topic endpoint>`

The endpoint values references the corresponding endpoints of the `s101[d|t|p][01|02]-ees-evgt-[publication-changed|release-changed|theme-changed]` topic resources. They are set conditionally depending on whether the topics exist, since they are not deployed to all environments yet.

## Other changes

- Add default settings configuration for `PublicationChangedEvent` and `ReleaseChangedEvent` to the Admin `appsettings.Development.json` file.


infrastructure/templates/template.json
```diff
@@ -1799,6 +1809,12 @@
         "App__EnableSwagger": "[parameters('enableSwagger')]",
         "App__EnableThemeDeletion": "[parameters('enableThemeDeletion')]",
         "Azure__SignalR__ConnectionString": "[concat('@Microsoft.KeyVault(SecretUri=', reference(variables('ees-signalr-admin-connectionstring'), '2018-02-14').secretUriWithVersion, ')')]",
+        "EventGrid__EventTopics__0__Key": "PublicationChangedEvent",
```
leeoades, line 1812: It would be nice if we could implement an appsettings.environment.json file instead that is mounted into the container so that we can have properly structured and nested config. This is pretty gnarly, eh?
Arguably, we're also adding potentially blank config into the array whereas those entries should really be omitted - but pfftt 🤣

leeoades (approved):
Looks ok to me, with the caveat as stated below :-)

infrastructure/templates/template.json
benoutram: Agree, it is pretty gnarly, in fact there's not many examples of how to do this nested configuration with arrays out there!

I think what's nice about this though is that we reference the resource, so it fails if the resource doesn't exist, and also we don't end up needing to hard code the region, domain and base url  like `westeurope-1.eventgrid.azure.net/api/events`. I suppose we'd only need to set these once somewhere else. Could be an improvement for the future?

benoutram: Ha ha yes, how would you get secrets in there or would you reference them? 😂 

We can definitely change this when we discover a better way to do things!

PR #5768, opened 2025-04-08
EES-5952 Add infrastructure for Event Grid custom topics

benoutram:
This PR adds the Bicep infrastructure to create Event Grid custom topics.

These topics will be used to publish events to Event Grid by Admin and Publisher when certain conditions are met.

Below is an explanation of when each topic will be used.

## Publication changed topic

- When a publication is changed e.g. a change to the publication title or summary.
- When the latest published release of a publication is changed after altering the publication's release order.

## Theme changed topic

- When a theme is changed e.g. a change to the theme title.

## Release changed topic

- When a release is changed e.g. a change to the release label.

## Release version changed topic

- When a new release version is published

Publishing events to the topics in each of the above scenarios has been the work of EES-5954, EES-5962, EES-5953, EES-5908, EES-5955 and EES-5956.

PR #5773, opened 2025-04-09
EES-5952 Add infrastructure for Event Grid topic subscriptions

benoutram:
This PR follows on from https://github.com/dfe-analytical-services/explore-education-statistics/pull/5768 which added Event Grid topics that the Admin and Publisher will use to publish events when certain conditions occur.

This PR subscribes the Search Function app to these topics.

Because the functions in the Search Function app have been configured to use queue trigger bindings rather than Event Grid trigger bindings, the Event Grid subscriptions are created using the storage queues as their destinations.

### DIFF
--- a/infrastructure/templates/common/components/event-grid/eventGridCustomTopicQueueSubscription.bicep
+++ b/infrastructure/templates/common/components/event-grid/eventGridCustomTopicQueueSubscription.bicep
@@ -3,7 +3,7 @@
 @maxLength(64)
 param name string
 
-@description('The name of the parent topic.')
+@description('The Event Grid topic name associated with the subscription.')
 param topicName string
 
 @description('The name of the storage account that contains the queue that is the destination of the subscription.')
--- a/infrastructure/templates/common/components/event-grid/eventGridCustomTopicQueueSubscription.bicep
+++ b/infrastructure/templates/common/components/event-grid/eventGridCustomTopicQueueSubscription.bicep
@@ -1,0 +1,49 @@
+@description('The resource name.')
+@minLength(3)
+@maxLength(64)
+param name string
+
+@description('The name of the parent topic.')
+param topicName string
+
+@description('The name of the storage account that contains the queue that is the destination of the subscription.')
+param storageAccountName string
+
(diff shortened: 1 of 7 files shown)


infrastructure/templates/common/components/event-grid/eventGridCustomTopicQueueSubscription.bicep
```diff
@@ -0,0 +1,49 @@
+@description('The resource name.')
+@minLength(3)
+@maxLength(64)
+param name string
+
+@description('The name of the parent topic.')
```
leeoades, line 6: Is the word "parent" necessary here? Perhaps "the name of the event grid topic to which this subscription is subscribed" or words to that effect?

leeoades (approved):
As always, please find some thoughts - but merge with the amendments (if any) you feel fit.

infrastructure/templates/common/components/event-grid/eventGridCustomTopicQueueSubscription.bicep
benoutram: No it's not necessary 🙂 . I only used the word "parent" because they refer to it as the parent in some of the descriptions of properties in the Bicep resource template documentation. I've renamed this now to "The Event Grid topic name associated with the subscription."

I also found that my formatter in VS code wasn't set up correctly so I've done a limited amount of reformatting at the same time.

infrastructure/templates/search/types.bicep
```diff
@@ -15,3 +15,14 @@ type ResourceNames = {
 @export()
 // The built in Search Service roles. See https://learn.microsoft.com/en-gb/azure/role-based-access-control/built-in-roles/ai-machine-learning
 type SearchServiceRole = 'Search Index Data Contributor' | 'Search Index Data Reader' | 'Search Service Contributor'
+
+@export()
+type StorageQueueNamesType = {
```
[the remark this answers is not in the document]
benoutram: No I don't think it's specific to use that as a suffix but a lot of the documentation on custom types has it in examples.

I've removed it for now to make it in keeping with our other custom types, and also renamed it `SearchStorageQueueNames` as suggested.

PR #5777, opened 2025-04-10
EES-6026 Remove Requests from Analytics.Consumer namespace

mmyoungman:
This PR remove "Requests" from the "Analytics.Consumer" namespace, so the namespace matches the directory name for the project.


leeoades (approved):
Well done for submitting this as a separate PR 👍

PR #5779, opened 2025-04-10
EES-6027 Change ConsumePublicApiQueries function to be general

mmyoungman:
This PR updates the `ConsumePublicApiQueriesFunction` to be more general so that we can more easily add different analytic request files to be processed in the future.

The logic that used to live in `ConsumePublicApiQueriesFunction` has basically been moved into a service, and we call the logic for that from the new `ConsumeAnalyticsRequestFilesFunction`.

As part of this change, I've also named the cron variable to indicate that it is more general.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer.Tests/Functions/ConsumeAnalyticsRequestFilesFunctionTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer.Tests/Functions/ConsumeAnalyticsRequestFilesFunctionTests.cs
@@ -12,23 +12,23 @@ public class ConsumeAnalyticsRequestFilesFunctionTests
     [Fact]
     public async Task ConsumeAnalyticsRequestFilesFunction_MultipleStrategies()
     {
-        var strategy1 = new Mock<IRequestFileProcessorService>(MockBehavior.Strict);
-        strategy1.Setup(s => s.Consume())
+        var strategy1 = new Mock<IRequestFileProcessor>(MockBehavior.Strict);
+        strategy1.Setup(s => s.Process())
             .Returns(Task.CompletedTask);
 
-        var strategy2 = new Mock<IRequestFileProcessorService>(MockBehavior.Strict);
-        strategy2.Setup(s => s.Consume())
+        var strategy2 = new Mock<IRequestFileProcessor>(MockBehavior.Strict);
+        strategy2.Setup(s => s.Process())
             .Returns(Task.CompletedTask);
 
         var function = BuildFunction([strategy1, strategy2]);
 
         await function.Run(new TimerInfo());
 
-        strategy1.Verify(s => s.Consume(), Times.Once);
-        strategy2.Verify(s => s.Consume(), Times.Once);
+        strategy1.Verify(s => s.Process(), Times.Once);
+        strategy2.Verify(s => s.Process(), Times.Once);
     }
 
-    private ConsumeAnalyticsRequestFilesFunction BuildFunction(List<Mock<IRequestFileProcessorService>> processorServices)
+    private ConsumeAnalyticsRequestFilesFunction BuildFunction(List<Mock<IRequestFileProcessor>> processorServices)
     {
         return new(
             processorServices.Select(s => s.Object),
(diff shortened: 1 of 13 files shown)


src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Services/Interfaces/IRequestFileProcessorService.cs
```diff
@@ -0,0 +1,6 @@
+namespace GovUk.Education.ExploreEducationStatistics.Analytics.Requests.Consumer.Services.Interfaces;
+
+public interface IRequestFileProcessorService
```
leeoades, line 3: Consider calling this `IRequestFileProcessor` ?

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Services/Interfaces/IRequestFileProcessorService.cs
```diff
@@ -0,0 +1,6 @@
...
+{
+    public Task Consume();
```
leeoades, line 5: If this is class is a `Processor` then should this method be `Process`?

leeoades (approved):
I've added some thoughts but looks good. I've approved so leave my comments in your hands to act on if you think them a good idea. Otherwise, no worries.

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer.Tests/Functions/ConsumeAnalyticsRequestFilesFunctionTests.cs
```diff
@@ -0,0 +1,37 @@
...
+using Xunit;
+
+namespace GovUk.Education.ExploreEducationStatistics.Analytics.Consumer.Tests.Functions;
+
+public class ConsumeAnalyticsRequestFilesFunctionTests
+{
+    [Fact]
+    public async Task ConsumeAnalyticsRequestFilesFunction_MultipleStrategies()
+    {
+        var strategy1 = new Mock<IRequestFileProcessorService>(MockBehavior.Strict);
```
leeoades: Surely a prime contender for a Mock Builder?! :-)

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer.Tests/Services/PublicApiQueriesProcessorServiceTests.cs
```diff
@@ -17,56 +16,62 @@ public abstract class ConsumePublicApiQueriesFunctionTests
...
-            
-            Assert.False(Directory.Exists(_pathResolver.PublicApiQueriesProcessingDirectoryPath()));
-            Assert.False(Directory.Exists(_pathResolver.PublicApiQueriesReportsDirectoryPath()));
+            var pathResolver = new TestAnalyticsPathResolver();
+
+            var service = BuildService(
+                pathResolver: pathResolver);
+            await service.Consume();
+
+            Assert.False(Directory.Exists(pathResolver.PublicApiQueriesProcessingDirectoryPath()));
```
leeoades, line 30: Isn't this asserting against your Test Analytics Path Resolver?

leeoades: Oh.. it's actually creating directories and checking for their existence... agghh...

leeoades: I think you've just ported these over or something. All good.

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer.Tests/Functions/ConsumeAnalyticsRequestFilesFunctionTests.cs
mmyoungman: For me, this is so simple that my inclination is to set it up inline t in these kinds of situation.

I was actually inclined to not right write this test because I feel like I'm testing whether dependency injection works than some actual functionality I've written.

PR #5786, opened 2025-04-11
EES-5791 Event Dto renames

leeoades:
More code tidying, separated from the refactor.
This time, I've renamed the Event Dtos by removing their Dto suffix.
Also, Ben and I discussed renaming the somewhat vague (and misleading)

OnPublicationLatestPublishedReleaseVersionChanged  
to
OnPublicationLatestPublishedReleaseReordered

since the event is not raised every time the latest published release version changes.


src/GovUk.Education.ExploreEducationStatistics.Admin/Events/PublicationLatestPublishedReleaseReorderedEvent.cs
```diff
@@ -4,9 +4,9 @@
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Events;
 
-public record PublicationLatestPublishedReleaseVersionChangedEventDto
+public record PublicationLatestPublishedReleaseReorderedEvent
```
benoutram, line 7: Do you think we should also change `Search.FunctionApp.Functions.OnPublicationLatestPublishedReleaseVersionChanged.Dtos.PublicationLatestPublishedReleaseVersionChangedEventDto` and change the queue name as well?

mmoalin:
Hey Lee, just regarding renaming 
OnPublicationLatestPublishedReleaseVersionChanged
to
OnPublicationLatestPublishedReleaseReordered
I notice there are bicep files that refer to `PublishedReleaseVersionChanged` and wondered if these also need changing? I have noticed the ticket number attached to this PR is the epic ticket number so just wondered if these changes are also planned for or required

src/GovUk.Education.ExploreEducationStatistics.Admin/Events/PublicationLatestPublishedReleaseReorderedEvent.cs
leeoades: Yes, good thinking. I've just pushed that change too.

benoutram:
> Hey Lee, just regarding renaming OnPublicationLatestPublishedReleaseVersionChanged to OnPublicationLatestPublishedReleaseReordered I notice there are bicep files that refer to `PublishedReleaseVersionChanged` and wondered if these also need changing? I have noticed the ticket number attached to this PR is the epic ticket number so just wondered if these changes are also planned for or required

We should have that covered by https://github.com/dfe-analytical-services/explore-education-statistics/pull/5788 @mmoalin @leeoades.

leeoades:
> Hey Lee, just regarding renaming OnPublicationLatestPublishedReleaseVersionChanged to OnPublicationLatestPublishedReleaseReordered I notice there are bicep files that refer to `PublishedReleaseVersionChanged` and wondered if these also need changing? I have noticed the ticket number attached to this PR is the epic ticket number so just wondered if these changes are also planned for or required

Hey Mo,
Good spot. Ben will be changing the Bicep files accordingly.
The refactors are part of some documentation work that I'm adding which can be viewed as part of the overall epic.

src/GovUk.Education.ExploreEducationStatistics.Admin/Events/PublicationLatestPublishedReleaseReorderedEvent.cs
mmyoungman, line 7: I see the event is being thrown in `PublicationService#UpdateReleaseSeries`, which I would expect given the name of the event, but then the actual event itself only records the old and new latest published release version.

I would expect this event to provide the new order (and perhaps the old order too), and then it would be on the consumer of the event to decide what they need to do with that. So if the consumer of the event needs to update the latest published release version, I think it should be on them to work that out?

Right now, if I were to look at this, see what information the event is storing, and not be aware of how it is being used, I'd assume it was named incorrectly and should be something like `UpdatedLatestPublishedReleaseVersionEvent`. 

I guess maybe the argument here is that we already know what the consumer of the event wants, so we're just including that in the event, and if future consumers want extra information that can be added to event?

mmyoungman, line 7: I'd be inclined to add `ReleaseSeries` to the name of the event, to be more explicit about what this is being raised against - maybe it could just be `UpdatedReleaseSeriesEvent`?

But not hugely fussed about this. So happy for you to leave it.

src/GovUk.Education.ExploreEducationStatistics.Admin/Events/ReleaseSlugChangedEvent.cs
```diff
@@ -3,9 +3,9 @@
 
 namespace GovUk.Education.ExploreEducationStatistics.Admin.Events;
 
-public record ReleaseSlugChangedEventDto
+public record ReleaseSlugChangedEvent
 {
-    public ReleaseSlugChangedEventDto(Guid releaseId, string newReleaseSlug, Guid publicationId, string publicationSlug)
+    public ReleaseSlugChangedEvent(Guid releaseId, string newReleaseSlug, Guid publicationId, string publicationSlug)
     {
         Subject = releaseId.ToString();
```
mmyoungman, line 10: I know this isn't part of this PR's changes - but there is ambiguity in "Subject" here as we often use "subject" to refer to a data set, where it seems this is saying subject as in subject of the event.

Happy for you to leave this, but just thought I'd raise it.

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Admin/Events/PublicationLatestPublishedReleaseReorderedEvent.cs
mmyoungman: Or maybe not because it seems this is mostly about the latest published release version updating

mmyoungman: Also not hugely fussed about this - I think I can see the rationale for doing it like this, but thought I'd raise it

leeoades: You're absolutely right and these are valid questions. It's a bit of an Open/Closed principle vs YAGNI. ie we could publish everything we could think of now to avoid having to make changes to the event in the future vs that's just a complete waste because not only do we not need that extra info now, but we may never need it - and so we're paying for lots of extra compute and storage for absolutely no reason.

I think because it is so simple to enrich the event with further information, my gut is to go with just this minimal set for now - safe in the knowledge that if we want to consume an event that doesn't yet have enough info, it's really straight forward to add to it.

One thing this approach avoids is the dreaded contract change where we start publishing something but later down the line realise we're doing it wrong and need to make a breaking change.

leeoades: So an UpdatedReleaseSeriesEvent / ReleaseSeriesChangedEvent would likely be raised whenever the series changed - including times when the latest release has not changed. The existing event we have is meant to be raised only when the latest published release has changed. And we renamed it in this PR because the event name implies that we raise it whenever the latest publish release changes - but actually, we only raise it when it is changed through a reorder.

src/GovUk.Education.ExploreEducationStatistics.Admin/Events/ReleaseSlugChangedEvent.cs
leeoades: Hopefully, in the context in this event object, Topic and Subject can be viewed as the common event properties. I guess if it were deemed to be confusing, this property could be renamed to EventSubject.

mmyoungman (approved):

PR #5789, opened 2025-04-14
EES-5973 EES-5975 Infrastructure for Admin and Publisher Event Grid role assignments

benoutram:
This PR assigns the system-assigned managed identities associated with the Admin App Service and the Publisher Function App with the built-in RBAC Event Grid roles, scoped to the Event Grid topics which both apps need to send events to. This sets up fine-grained permissions between the apps and the topics which they use.

The apps authenticate using Microsoft Entra access token authentication, using the token provided by their managed identity credentials. This avoids the need for enabling local authentication on the Event Grid topics and the configuration of the app settings with topic access keys.

## Publisher

Publisher is assigned the `EventGrid Data Sender` role, scoped to the `release-version-changed` topic.

## Admin

Admin is the assigned the `EventGrid Data Sender` role, scoped to the `publication-changed`, `release-changed`, and `theme-changed` topics.

## Other changes

* Change the `index` tokens in the `eventGridCustomTopicModule` and `eventGridQueueSubscriptionModuleDeploy` names to be suffixed rather than prefixed.


infrastructure/templates/search/application/eventMessaging.bicep
```diff
@@ -16,6 +20,27 @@ param tagValues object
...
+}
+
+resource publisherFunction 'Microsoft.Web/sites@2024-04-01' existing = {
+  name: resourceNames.existingResources.publisherFunction
+}
+
+// This is the built-in EventGrid Data Sender role. See https://learn.microsoft.com/azure/role-based-access-control/built-in-roles#integration
+var eventGridDataSenderRoleDefinitionId = 'd5a91429-5739-47e2-a06b-3470a27159e7'
+
+var topicsUsedByAdmin = [
```
leeoades, line 34: Is there anyway to give admin and publisher event sender role to all topics? Just wondering if this is tightening the permissions unnecessarily at the expense of extra administration overhead?
ie Every time we add a another topic, we'll need to remember to come in here and set this?

infrastructure/templates/search/application/eventMessaging.bicep
```diff
@@ -26,3 +51,27 @@ module eventGridMessagingModule '../../common/components/event-grid/eventGridMes
...
 }
+
+// Allow the Admin App Service to send events to Event Grid topics
+module adminTopicRoleAssignmentModuleDeploy '../../common/components/event-grid/eventGridTopicRoleAssignment.bicep' = [
+  for (topic, index) in topicsUsedByAdmin: {
+    name: 'adminTopicRoleAssignmentModuleDeploy-${index}'
+    params: {
+      principalId: adminAppService.identity.principalId
+      roleDefinitionId: eventGridDataSenderRoleDefinitionId
+      topicName: '${resourcePrefix}-${abbreviations.eventGridTopics}-${topic}'
```
leeoades, line 62: I think it would make sense to create a bicep function to generate the topic name so that the different places using it are consistent.

leeoades (approved):
Looks good to me. A couple of minor thoughts for your consideration.

infrastructure/templates/search/application/eventMessaging.bicep
benoutram: 👍 Added function `fullyQualifiedTopicName` and put it in common so it can be used outside of Search.

infrastructure/templates/search/application/eventMessaging.bicep
benoutram: Good point, I've made a change to remove the `topicsUsedByAdmin` and `topicsUsedByPublisher` arrays so now both modules loop over `topics` when creating the role assignments.

This is the same `topics` array that's used when creating the topics and now comes from the definition in `eventTopics.bicep`.

Alternatively I think we could have scoped the role assignments at the resource group level rather than an individual topic level meaning that Admin and Publisher would have these role assignments available to all resources.

There's also Event Grid namespaces, a method of grouping topics by application. It looks like it provides a single endpoint for all the namespace topics and handles all the networking and security in one place. It would be a bit of extra effort to switch over to them as they aren't the same type of topic resource that we've used currently, but they look like a good idea. The role assignments would be scoped to the namespace rather than individual topics.

infrastructure/templates/search/application/eventMessaging.bicep
leeoades: Perhaps need a verb in there? `buildFullyQualifiedTopicName` perhaps?

benoutram: Changed `fullyQualifiedTopicName` to `buildFullyQualifiedTopicName` 👍 .

PR #5797, opened 2025-04-15
EES-6035 Attempted fix of Dev env analytics bug

mmyoungman:
This PR is hopefully a fix for an analytics bug we've seen in the Dev environment.

This bug was revealed when running the performance tests with this command:
```
pnpm --environment=dev perftest dist/publicApiDataSetQuery.test.js -e QUERY_FILES=dev/api-tests-absence-query1.json,dev/api-tests-absence-query2.json -e PROFILE=load -e MAIN_TEST_STAGE_DURATION_MINS=10 -e RPS=10
```

This would in rare instances produce request files by the Public API's `QueryAnalyticWriter#ReportDataSetVersionQuery` that ended with an extra close curly bracket. We now believe this is due to the Public API container app scaling up, and then due to a clashing filename, `WriteAllTextAsync` running twice from two different containers writing to a single file at the same time. While I feel fairly confident this is the cause, it's possible it's not.

The fix was to add a random Guid to the end of the request file's name to ensure that all filenames are unique.

To make analytics more robust, we now import the request file json into a DuckDb table one file at a time, and any file that fails will be moved to a "failures" directory. This means we don't need to manually delete any existing request files with malformed json, as they'll automatically be dealt with.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Functions/ConsumeAnalyticsRequestFilesFunction.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Functions/ConsumeAnalyticsRequestFilesFunction.cs
@@ -16,7 +16,17 @@ public class ConsumeAnalyticsRequestFilesFunction(
 
         foreach (var requestFileProcessor in processors)
         {
-            await requestFileProcessor.Process();
+            try
+            {
+                await requestFileProcessor.Process();
+            }
+            catch (Exception e)
+            {
+                logger.LogError(
+                    e,
+                    "Failed to process request files with processor {RequestFileProcessorType}",
+                    requestFileProcessor.GetType());
+            }
         }
     }
 }
--- a/src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Services/PublicApiQueriesProcessor.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Services/PublicApiQueriesProcessor.cs
@@ -1,3 +1,4 @@
+using DuckDB.NET.Data;
 using GovUk.Education.ExploreEducationStatistics.Analytics.Consumer.Services.Interfaces;
 using GovUk.Education.ExploreEducationStatistics.Common.DuckDb.DuckDb;
 using Microsoft.Extensions.Logging;
@@ -37,7 +38,7 @@ public class PublicApiQueriesProcessor(
 
         var processingDirectory = pathResolver.PublicApiQueriesProcessingDirectoryPath();
         var reportsDirectory = pathResolver.PublicApiQueriesReportsDirectoryPath();
-        
+
         Directory.CreateDirectory(processingDirectory);
         Directory.CreateDirectory(reportsDirectory);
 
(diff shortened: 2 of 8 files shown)


src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Services/PublicApiQueriesProcessor.cs
```diff
@@ -47,30 +48,69 @@ public Task Process()
...
-                    totalRowsCount: INT,
-                    startTime: DATETIME,
-                    endTime: DATETIME,
-                    query: JSON
-                }})");
+        // We fetch the files again in case there are files leftover in the processing dir from a previous function run
+        var filesReadyForProcessing = Directory
+            .GetFiles(processingDirectory)
+            .Select(Path.GetFileName)
+            .Cast<string>()
```
leeoades, line 75: Isn't it already of type String?

leeoades: Ah. If you're trying to remove null values then use .OfType<string>() instead which will filter them out.

leeoades (approved):
I've left some thoughts which you can do with whatever you wish. If you think they're worth making changes for and want me to check them again, let me know.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/QueryAnalyticsWriter.cs
```diff
@@ -32,7 +32,7 @@ public async Task ReportDataSetVersionQuery(CaptureDataSetVersionQueryRequest re
         {
             Directory.CreateDirectory(directory);
 
-            var filename = $"{DateTime.UtcNow:yyyyMMdd-HHmmss-fff}_{request.DataSetVersionId}.json";
+            var filename = $"{DateTime.UtcNow:yyyyMMdd-HHmmss-fff}_{request.DataSetVersionId}_{Guid.NewGuid()}.json";
```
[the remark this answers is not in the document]
mmyoungman: I've removed the fractions of seconds.

I was thinking to leave the `DateTime` and the `DataSetVersionId` as that seems like it might be useful for debugging. So to keep the filename shorter, rather than add a random guid on the end, I've switched to generating a random string of 5 characters.

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Services/PublicApiQueriesProcessor.cs
```diff
@@ -47,30 +48,69 @@ public Task Process()
             var newPath = Path.Combine(processingDirectory, file);
             File.Move(originalPath, newPath);
         });
-        
+
         duckDbConnection.Open();
         
         duckDbConnection.ExecuteNonQuery("install json; load json");
+        duckDbConnection.ExecuteNonQuery(@"
```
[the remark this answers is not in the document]
mmyoungman: I think that should never happen. The DuckDb database tables are created fresh in memory each function run, and only used once to create new parquet report files. We don't carry the tables or any of their data over.

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Functions/ConsumeAnalyticsRequestFilesFunction.cs
```diff
@@ -16,7 +16,15 @@ public async Task Run(
...
         foreach (var requestFileProcessor in processors)
         {
-            await requestFileProcessor.Process();
+            try
+            {
+                await requestFileProcessor.Process();
+            }
+            catch (Exception e)
+            {
+                logger.LogError(
```
leeoades, line 25: Log Error takes the exception as the first argument - instead of adding it's ToString into the message. This allows the handling / formatting of exceptions to be controlled by the logger. Also, the string part of a log should be a constant. Instead of inserting the type string into the log format string, add a named placeholder into the log string, and provide the value in the next argument.
Therefore:
`logger.LogError(e, "Failed to process request files with processor {RequestFileProcessorType}", requestFileProcessor.GetType());`
In serilog, this would expose the type as a property of the log message that could then be filtered on - rather than being just a part of the string.
Lastly, if requestFileProcessor was a common complex type, then you could pass that in (instead of its Type) and then add a log destructor which would know the best way to represent that type in the logs.

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Services/PublicApiQueriesProcessor.cs
```diff
@@ -103,12 +104,12 @@ FROM read_json('{processingDirectory}/{filename}',
             }
             catch (DuckDBException e)
             {
-                logger.LogCritical($"Failed to process analytics request file {filename}\n Exception: {e}");
+                logger.LogError($"Failed to process analytics request file {filename}\n Exception: {e}");
```
leeoades, line 107: Please see my other comment about logging errors. :-)

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Services/PublicApiQueriesProcessor.cs
```diff
@@ -103,12 +104,12 @@ FROM read_json('{processingDirectory}/{filename}',
...
                 var badFilePath = Path.Combine(processingDirectory, filename);
```
leeoades, line 108: Performing these operations in the catch means that exceptions will be uncaught and will bubble up.
Consider creating a method called `MoveBadFileToFailuresDirectory` which can both house all of this code but also have its own exception handling.

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/RandomUtils.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Utils;
+
+public class RandomUtils
+{
+    private static Random random = new();
+
+    public static string RandomString(int length)
+    {
+        const string chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
```
leeoades, line 11: One thing that makes me slightly nervous about using this technique is that random is seeded off of the clock and we already know that the different processes can yield the same clock value (because that's what led to the clashing file names in the first place). Instead, consider creating a GUID and then substring it. That will have a higher probability of avoiding a clash.

leeoades (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Services/PublicApiQueriesProcessor.cs
mmyoungman: I considered creating a common function that could be used in other future processes, but for now I've just added a private method in this class.

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Functions/ConsumeAnalyticsRequestFilesFunction.cs
```diff
@@ -16,7 +16,17 @@ public async Task Run(
...
+                    e,
+                    "Failed to process request files with processor {RequestFileProcessorType}",
+                    requestFileProcessor.GetType().Name);
```
leeoades, line 28: Very minor (don't change it): you could pass the whole Type in here which would allow the logger to display it however it wanted. I'm not sure it's configured to do anything other than should the name and knowing my luck, it would probably try to output all of the properties of the Type which we definitely don't want!

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Services/PublicApiQueriesProcessor.cs
```diff
@@ -162,4 +160,18 @@ FROM queries
...
+    {
+        try
+        {
+            var failuresDirectoryPath = pathResolver.PublicApiQueriesFailuresDirectoryPath();
+            Directory.CreateDirectory(failuresDirectoryPath);
+            File.Move(filePath, failuresDirectoryPath);
+        }
+        catch (Exception e)
+        {
+            logger.LogError(e, "Failed to move bad file to failures directory {filePath}", filePath);
```
leeoades, line 174: Perhaps a good idea to output both the source and destination paths here to help troubleshoot why this didn't work? If the problem is the destination directory then we don't have a record here of what it thought that directory path was.

leeoades (approved):

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/RandomUtils.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace GovUk.Education.ExploreEducationStatistics.Common.Utils;
+
+public static class RandomUtils
+{
+    public static string RandomString()
+    {
+        var guidStr = Guid.NewGuid().ToString();
+        return guidStr[^5..];
```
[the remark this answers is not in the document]
mmyoungman: I had come across this syntax before, but in this instance I used Substring and then Rider suggested this :)

PR #5805, opened 2025-04-16
EES-XXXX - breaking out separate ConfigureServices method and supplying in-memory configuration with PublicDataDbExists flag set to false for simplicity

duncan-at-hiveit:
This PR:
- fixes an issue whereby the `PublishingCompletionServiceTests.EnsureSUTCanBeResolved` test was failing when the PublicDataDbExists flag was set to true in peoples' `appsettings.Local.json` files, as this resulted in no PublicDataDbContext being registered but subsequent dependencies needing one to be available.

We now provide explicit configuration to this test's HostBuilder and call `ConfigureServices()` directly without needing to rely on prior configuration from `ConfigureAppConfiguration()`.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/HostBuilderTestExtensions.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/HostBuilderTestExtensions.cs
@@ -1,0 +1,28 @@
+using System.Collections.Generic;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Hosting;
+
+namespace GovUk.Education.ExploreEducationStatistics.Publisher.Tests;
+
+public static class HostBuilderTestExtensions
+{
+    private const string StorageConnectionString = 
+        "DefaultEndpointsProtocol=http;AccountName=account;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://data-storage:10000/devstoreaccount1;";
+
+    public static IHostBuilder ConfigureTestAppConfiguration(this IHostBuilder hostBuilder)
+    {
+        return hostBuilder.ConfigureAppConfiguration(s =>
+        {
+            var configuration = new Dictionary<string, string>
+            {
+                { "PublicDataDbExists", "false" },
+                { "App:PrivateStorageConnectionString", StorageConnectionString },
+                { "App:NotifierStorageConnectionString", StorageConnectionString },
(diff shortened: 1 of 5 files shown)


src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/HostBuilderTestExtensions.cs
```diff
@@ -0,0 +1,28 @@
+using System.Collections.Generic;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Hosting;
+
+namespace GovUk.Education.ExploreEducationStatistics.Publisher.Tests;
+
+public static class HostBuilderTestExtensions
+{
+    private const string StorageConnectionString =
```
leeoades, line 9: Minor: I would have given this a descriptive name explaining that it is the well-known developer Azure connection string (or whatever it is!).

src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/HostBuilderTestExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+        "DefaultEndpointsProtocol=http;AccountName=account;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://data-storage:10000/devstoreaccount1;";
+
+    public static IHostBuilder ConfigureTestAppConfiguration(this IHostBuilder hostBuilder)
+    {
+        return hostBuilder.ConfigureAppConfiguration(s =>
+        {
+            var configuration = new Dictionary<string, string>
+            {
+                { "PublicDataDbExists", "false" },
+                { "App:PrivateStorageConnectionString", StorageConnectionString },
```
leeoades, line 19: Future: The reason that this connection string seems to be needed is because `BlobStorageService` instantiates a `BlobServiceClient` in its constructor, which is a bit premature. It could probably defer the creation of the client until it actually needs one. Anyway.

leeoades (approved):
Well done and thanks for sorting that!

src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/HostBuilderTestExtensions.cs
duncan-at-hiveit: No problem!  I've updated to `DevelopmentStorageConnectionString`

src/GovUk.Education.ExploreEducationStatistics.Publisher.Tests/HostBuilderTestExtensions.cs
duncan-at-hiveit: :+1:

PR #5809, opened 2025-04-17
EES-6049 - Public API wildcard data set versions try to access unpublished data bug fix

mmoalin:
### Summary
This Pull Request fixes functionality that handles requests to the Public API with wildcard versions (e.g., 2.*) that was previously returning unpublished data set versions.

It tweaks logic of Public API endpoints that accept data set version that user specifies, that if the user a wildcard dataset version, that only published dataset versions are returned. 

### Main Changes
Ensure that for wildcard data set version - only published data set versions are returned if present. 
Ensures 404 response if a non-published version exists and has been requested using wildcard parameter.

### Updates to tests:
Added or updated test coverage for controllers handling wildcard dataset version queries.
Extended existing tests to check the behavior for both published and non-published versions.
Introduced a new method WherePublishedStatus in dataset version query extension to filter only published versions.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/DataSetVersionNumberTest.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/DataSetVersionNumberTest.cs
@@ -125,7 +125,7 @@ public static class DataSetVersionNumberTest
         [InlineData("*")]
         public void PassedWildcard_ExpectsIsWildcardToReturnTrue(string versionString)
         {
-            DataSetVersionNumber.TryParse(versionString, out var version);
+            Assert.True(DataSetVersionNumber.TryParse(versionString, out var version));
             Assert.True(version.IsWildcard);
         }
         
@@ -135,7 +135,7 @@ public static class DataSetVersionNumberTest
         [InlineData("1")]
         public void PassedNonWildcard_ExpectsIsWildcardToReturnFalse(string versionString)
         {
-            DataSetVersionNumber.TryParse(versionString, out var version);
+            Assert.True(DataSetVersionNumber.TryParse(versionString, out var version));
             Assert.False(version.IsWildcard);
         }
     }
--- a/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/DataSetVersionNumberTest.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/DataSetVersionNumberTest.cs
@@ -118,5 +118,25 @@ public static class DataSetVersionNumberTest
         {
             Assert.False(DataSetVersionNumber.TryParse(versionString, out _));
         }
+
+        [Theory]
+        [InlineData("1.*")]
+        [InlineData("1.1.*")]
+        [InlineData("*")]
+        public void PassedWildcard_ExpectsIsWildcardToReturnTrue(string versionString)
+        {
+            DataSetVersionNumber.TryParse(versionString, out var version);
+            Assert.True(version.IsWildcard);
+        }
+        
+        [Theory]
+        [InlineData("1.1")]
+        [InlineData("1.1.1")]
+        [InlineData("1")]
+        public void PassedNonWildcard_ExpectsIsWildcardToReturnFalse(string versionString)
+        {
+            DataSetVersionNumber.TryParse(versionString, out var version);
+            Assert.False(version.IsWildcard);
+        }
     }
 }
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Utils/DataSetVersionNumber.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Utils/DataSetVersionNumber.cs
@@ -8,7 +8,7 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Utils;
 
 public record DataSetVersionNumber(int? Major, int? Minor, int? Patch)
 {
-    public bool IsWildcard { get; set; }
+    public bool IsWildcard { get; init; }
 
     public static bool TryParse(string versionString, [NotNullWhen(true)] out DataSetVersionNumber? version)
     {
--- a/src/GovUk.Education.ExploreEducationStatistics.Common/Utils/DataSetVersionNumber.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Common/Utils/DataSetVersionNumber.cs
@@ -8,7 +8,9 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Utils;
 
 public record DataSetVersionNumber(int? Major, int? Minor, int? Patch)
 {
-    public static bool TryParse(string versionString, out DataSetVersionNumber? version)
+    public bool IsWildcard { get; set; }
+
+    public static bool TryParse(string versionString, [NotNullWhen(true)] out DataSetVersionNumber? version)
     {
         version = null;
         if (versionString.Contains('*'))
@@ -27,8 +29,12 @@ public record DataSetVersionNumber(int? Major, int? Minor, int? Patch)
         {
             return false;
         }
-
-        version = new DataSetVersionNumber(sv.Major, sv.Minor, sv.Patch);
+        
+        version = new DataSetVersionNumber(sv.Major, sv.Minor, sv.Patch)
+        {
+            IsWildcard = false
+        };
+        
         return successful;
     }
 
@@ -59,7 +65,11 @@ public record DataSetVersionNumber(int? Major, int? Minor, int? Patch)
 
         version = new DataSetVersionNumber(parts[0],
             parts.Length > 1 ? parts[1] : null,
-            parts.Length > 2 ? parts[2] : null);
+            parts.Length > 2 ? parts[2] : null)
+        {
+            IsWildcard = true
+        };
+
         return true;
     }
 }
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetVersionsControllerTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetVersionsControllerTests.cs
@@ -1869,8 +1869,10 @@ public abstract class DataSetVersionsControllerTests(TestApplicationFactory test
                     dsv.SetVersionNumber(2, 1);
                 })
                 .GenerateList();
+        
+        var version = versionStatus == DataSetVersionStatus.Published ? "2.1" : "1.2";
 
-        dataSet.LatestLiveVersion = dataSetVersion.FirstOrDefault(dsv => dsv.PublicVersion == "1.2");
+        dataSet.LatestLiveVersion = dataSetVersion.FirstOrDefault(dsv => dsv.PublicVersion == version);
         dataSet.Versions = dataSetVersion;
 
         await TestApp.AddTestData<PublicDataDbContext>(
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetVersionsControllerTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetVersionsControllerTests.cs
@@ -1721,6 +1721,35 @@ public abstract class DataSetVersionsControllerTests(TestApplicationFactory test
             response.AssertOk();
         }
 
+        [Fact]
+        public async Task WildCardSpecified_RequestedPublished_Returns200()
+        {
+            var (dataSet, dataSetVersion) = await SetupDataSetWithSpecifiedVersionStatuses(DataSetVersionStatus.Published);
+            
+            var response = await GetDataSetVersionChanges(
+                dataSetId: dataSet.Id,
+                dataSetVersion: "2.*",
+                previewTokenId: dataSetVersion.Last().PreviewTokens[0].Id);
+
+            response.AssertOk<DataSetVersionChangesViewModel>(useSystemJson: true);
+        }
+
+
+        [Theory]
+        [MemberData(nameof(DataSetVersionStatusViewTheoryData.NonPublishedStatus),
+            MemberType = typeof(DataSetVersionStatusViewTheoryData))]
+        public async Task WildCardSpecified_RequestedNonPublished_Returns404(DataSetVersionStatus dataSetVersionStatus)
+        {
+            var (dataSet, dataSetVersion) = await SetupDataSetWithSpecifiedVersionStatuses(dataSetVersionStatus);
+            
+            var response = await GetDataSetVersionChanges(
+                dataSetId: dataSet.Id,
+                dataSetVersion: "2.*",
+                previewTokenId: dataSetVersion.Last().PreviewTokens[0].Id);
+
+            response.AssertNotFound();
+        }
+        
         private async Task<HttpResponseMessage> GetDataSetVersionChanges(
             Guid dataSetId,
             string dataSetVersion,
@@ -1740,6 +1769,59 @@ public abstract class DataSetVersionsControllerTests(TestApplicationFactory test
         }
     }
 
+    private async Task<(DataSet, List<DataSetVersion>)> SetupDataSetWithSpecifiedVersionStatuses(
+        DataSetVersionStatus versionStatus)
+    {
+        DataSet dataSet = DataFixture
+            .DefaultDataSet()
+            .WithStatusPublished();
+
+        await TestApp.AddTestData<PublicDataDbContext>(context => context.DataSets.Add(dataSet));
+
+        var dataSetVersion = DataFixture
+                .DefaultDataSetVersion(filters: 1, indicators: 1, locations: 1, timePeriods: 3)
+                .WithStatusPublished()
+                .WithDataSetId(dataSet.Id)
+                .WithMetaSummary(
+                    DataFixture.DefaultDataSetVersionMetaSummary()
+                        .WithGeographicLevels(
+                            [
+                                GeographicLevel.Country,
+                                GeographicLevel.LocalAuthority,
+                                GeographicLevel.Region,
+                                GeographicLevel.School
+                            ]
+                        )
+                )
+                .WithPreviewTokens(() => [DataFixture.DefaultPreviewToken()])
+                .ForIndex(1, dsv => dsv.SetVersionNumber(1, 1))
+                .ForIndex(2, dsv => dsv.SetVersionNumber(1, 2))
+                .ForIndex(3, dsv =>
+                {
+                    dsv.SetStatus(versionStatus);
+                    dsv.SetVersionNumber(2, 0);
+                    
+                })
+                .ForIndex(4, dsv =>
+                {
+                    dsv.SetStatus(versionStatus);
+                    dsv.SetVersionNumber(2, 1);
+                })
+                .GenerateList();
+
+        dataSet.LatestLiveVersion = dataSetVersion.FirstOrDefault(dsv => dsv.PublicVersion == "1.2");
+        dataSet.Versions = dataSetVersion;
+
+        await TestApp.AddTestData<PublicDataDbContext>(
+            context =>
+            {
+                context.DataSetVersions.AddRange(dataSetVersion);
+                context.DataSets.Update(dataSet);
+            }
+        );
+
+        return (dataSet, dataSetVersion);
+    }
     private WebApplicationFactory<Startup> BuildApp(
         IContentApiClient? contentApiClient = null,
         ClaimsPrincipal? user = null)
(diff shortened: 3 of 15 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionQueryableExtensions.cs
```diff
@@ -8,23 +8,33 @@ namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Extension
 public static class DataSetVersionQueryableExtensions
 {
     /// <summary>
-    /// Returns specified data set version or the latest version of a data set based on a major/minor/patch level that's being wildcarded (i.e., substituted with '*').
+    /// Returns specified data set version or the latest version of a data set based on a major/minor/patch level that can be wildcarded (i.e., substituted with '*').
     /// </summary>
     /// <param name="version">Data set version which can contain a wildcard</param>
+    /// <param name="publicOnly">Specifies whether this method should return versions that are not
+    /// "Published", "Depreciated" or "Withdrawn".
```
leeoades, line 15: Depreciated or Deprecated?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetService.cs
```diff
@@ -101,6 +101,7 @@ public async Task<Either<ActionResult, DataSetVersionViewModel>> GetVersion(
             .FindByVersion(
                 dataSetId: dataSetId,
                 version: dataSetVersion,
+                publicOnly: true,
```
leeoades, line 104: As you say in your PR summary, this is changing the behaviour of GetDataSetVersion. There don't seem to be any associated unit tests or integration tests that have therefore also needed to change? This tends to suggest that this end point is untested. Please could you add some.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetService.cs
```diff
@@ -258,6 +259,7 @@ private async Task<Either<ActionResult, DataSetVersion>> FindVersion(
             .AsNoTracking()
             .FindByVersion(
                 dataSetId: dataSetId,
+                publicOnly: true,
```
leeoades, line 262: As you say in your PR summary, this is changing the behaviour of GetDataSetVersion. There don't seem to be any associated unit tests or integration tests that have therefore also needed to change? This tends to suggest that this end point is untested. Please could you add some.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetVersionChangeService.cs
```diff
@@ -27,6 +27,7 @@ public Task<Either<ActionResult, DataSetVersionChangesViewModel>> GetChanges(
             .FindByVersion(
                 dataSetId: dataSetId,
                 version: dataSetVersion,
+                publicOnly: true,
```
leeoades, line 30: As you say in your PR summary, this is changing the behaviour of GetDataSetVersion. There don't seem to be any associated unit tests or integration tests that have therefore also needed to change? This tends to suggest that this end point is untested. Please could you add some.

leeoades (changes requested):
Good to see your unit test in there. Looks like the control endpoints that are being affected by this change have no tests around them?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetQueryService.cs
```diff
@@ -133,7 +133,11 @@ private async Task<Either<ActionResult, DataSetVersion>> FindDataSetVersion(
             .DataSetVersions
             .AsNoTracking()
             .Include(dsv => dsv.DataSet)
-            .FindByVersion(dataSetId, dataSetVersion, cancellationToken);
+            .FindByVersion(
+                dataSetId: dataSetId, 
+                version: dataSetVersion,
+                publicOnly: true, 
+                cancellationToken);
     }
```
benoutram, line 141: Just wondering if above on L120 where we already had `if (dataSetVersion is null)`, if this should be altered to cover null **OR** '*'?

That would mean if the version is a wildcard then it would fall into the block above this one which uses `.Select(ds => ds.LatestLiveVersion!)` which could be an optimisation over retrieving the top public version from all the versions after sorting.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionQueryableExtensions.cs
```diff
@@ -8,23 +8,33 @@ namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Extension
...
-        return await queryable
-            .Where(dsv => dsv.DataSetId == dataSetId)
-            .Where(v =>
+        var query = queryable
+            .Where(dsv => dsv.DataSetId == dataSetId);
+
+        if (publicOnly)
+        {
+            query = query.Where(dsv => dsv.Status == DataSetVersionStatus.Published);
+        }
```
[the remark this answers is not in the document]
jack-hive: I'd imagine we want to include `Withdrawn` and `Deprecated` like Ben suggested.

There is an extension method that exists for this called `WherePublicStatus`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionQueryableExtensions.cs
```diff
@@ -8,23 +8,33 @@ namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Extension
 public static class DataSetVersionQueryableExtensions
 {
     /// <summary>
-    /// Returns specified data set version or the latest version of a data set based on a major/minor/patch level that's being wildcarded (i.e., substituted with '*').
+    /// Returns specified data set version or the latest version of a data set based on a major/minor/patch level that can be wildcarded (i.e., substituted with '*').
```
jack-hive, line 11: I personally think the previous summary was more accurate. But I do agree it might be possible to word it better...

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetVersionChangeService.cs
jack-hive: I think this might break Admin.

The Admin app also calls into the endpoint that triggers this. See the `GetVersionChanges` method in the `DataSetVersionsController` in Admin.

That endpoint should, I think, be able to grab the changeset details for an unpublished data set version. I believe this would stop that from happening.

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetService.cs
mmoalin: Yes, I had some unit tests but no integration tests so had added integration tests to DataSetsControllerTests (GetDataSetMetaTests.DataSetVersionQueryParameterTests, DownloadTests), DataSetsControllerGetQueryTests.AccessTests, DataSetsControllerPostQueryTests.AccessTests, DataSetVersionsControllerTests.GetDataSetVersionChangesTests

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetService.cs
mmoalin: Please see above comment

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetVersionChangeService.cs
mmoalin: Totally makes sense, thanks Jack, I modified the code so that the logic of this ticket only runs when the version specified is being wildcarded (which I should've been more strict and done in the first place).

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionQueryableExtensions.cs
```diff
@@ -8,23 +8,33 @@ namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Extension
...
     /// </summary>
     /// <param name="version">Data set version which can contain a wildcard</param>
+    /// <param name="publicOnly">Specifies whether this method should return versions that are not
+    /// "Published".
+    /// </param>
```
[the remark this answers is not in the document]
mmoalin: Thanks Ben

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionQueryableExtensions.cs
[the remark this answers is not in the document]
mmoalin: Thansk Both. 
So I had a chat with Cam on Thursday to inquire about this and he specified that for the Wildcard they would not expect withdrawn or deprecated versions. Had modified the code to only return published versions when there is a wildcard (as specced).

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetQueryService.cs
```diff
@@ -133,7 +133,11 @@ private async Task<Either<ActionResult, DataSetVersion>> FindDataSetVersion(
             .DataSetVersions
             .AsNoTracking()
             .Include(dsv => dsv.DataSet)
-            .FindByVersion(dataSetId, dataSetVersion, cancellationToken);
+            .FindByVersion(
+                dataSetId: dataSetId, 
+                version: dataSetVersion,
+                publicOnly: true, 
+                cancellationToken);
```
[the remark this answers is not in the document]
mmoalin: Thanks Ben! no I have missed this extension! I've made my changes so that it uses a similar extension for the wildcards (published only statuses)

mmoalin:
> Good to see your unit test in there. Looks like the control endpoints that are being affected by this change have no tests around them?

This should be in place now, thanks!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionQueryableExtensions.cs
mmoalin: Had retweaked this. Thanks!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionQueryableExtensions.cs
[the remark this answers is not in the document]
mmoalin: Had confirmed this morning again that deprecated or withdrawn are not required.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionQueryableExtensions.cs
```diff
@@ -8,9 +8,12 @@ namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Extension
 public static class DataSetVersionQueryableExtensions
 {
     /// <summary>
-    /// Returns specified data set version or the latest version of a data set based on a major/minor/patch level that's being wildcarded (i.e., substituted with '*').
+    /// Returns specified data set version, if a wildcard is specified, the latest version of a data set based on a major/minor/patch level that's being wildcarded (i.e., substituted with '*') is returned.
```
jack-hive, line 11: NITPIC: But I'd put a full stop after 'Returns specified data set version'

Other than that, looks good to me now :)

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetService.cs
```diff
@@ -96,7 +96,17 @@ public async Task<Either<ActionResult, DataSetVersionViewModel>> GetVersion(
         string dataSetVersion,
         CancellationToken cancellationToken = default)
     {
-        return await publicDataDbContext.DataSetVersions
+        return dataSetVersion.Contains('*')
```
jack-hive, line 99: now we're repeating this logic, I'm wondering if it should go inside the `.FindByVersion` method?
@benoutram?

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetService.cs
jack-hive: i.e. we're having to put a conditional of `dataSetVersion.Contains('*') ? ... : ...` everywhere now. Might be better to let the `.FindByVersion` method figure that out for you

benoutram: We've had a call now to resolve this comment, and @mmoalin is going to make this change.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionQueryableExtensions.cs
mmoalin: This is sorted, cheers!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetService.cs
mmoalin: Had sorted this, thanks both!

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetVersionsControllerTests.cs
```diff
@@ -1869,8 +1869,10 @@ private async Task<HttpResponseMessage> GetDataSetVersionChanges(
                     dsv.SetVersionNumber(2, 1);
                 })
                 .GenerateList();
+        
+        var version = versionStatus == DataSetVersionStatus.Published ? "2.1" : "1.2";
```
leeoades, line 1873: Please could you add a task somewhere (a jira even?) to dedupe all of these `SetupDataSetWithSpecifiedVersionStatuses` and other cut n pasted code.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/DataSetVersionNumberTest.cs
```diff
@@ -118,5 +118,25 @@ public void InvalidVersion_FailsToParse(string versionString)
...
             Assert.False(DataSetVersionNumber.TryParse(versionString, out _));
         }
+
+        [Theory]
+        [InlineData("1.*")]
+        [InlineData("1.1.*")]
+        [InlineData("*")]
+        public void PassedWildcard_ExpectsIsWildcardToReturnTrue(string versionString)
+        {
+            DataSetVersionNumber.TryParse(versionString, out var version);
```
leeoades, line 128: Assert.True this line too.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/DataSetVersionNumberTest.cs
```diff
@@ -118,5 +118,25 @@ public void InvalidVersion_FailsToParse(string versionString)
...
+            Assert.True(version.IsWildcard);
+        }
+        
+        [Theory]
+        [InlineData("1.1")]
+        [InlineData("1.1.1")]
+        [InlineData("1")]
+        public void PassedNonWildcard_ExpectsIsWildcardToReturnFalse(string versionString)
+        {
+            DataSetVersionNumber.TryParse(versionString, out var version);
```
leeoades, line 138: Assert.True this line too.

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/DataSetVersionNumber.cs
```diff
@@ -8,7 +8,9 @@ namespace GovUk.Education.ExploreEducationStatistics.Common.Utils;
 
 public record DataSetVersionNumber(int? Major, int? Minor, int? Patch)
 {
-    public static bool TryParse(string versionString, out DataSetVersionNumber? version)
+    public bool IsWildcard { get; set; }
```
leeoades, line 11: Please make this `init` (or just `get;`). At the moment, it could be modified.
e.g. `myDataSetVersionNumber.IsWildcard = true`

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Extensions/DataSetVersionQueryableExtensionsTests.cs
```diff
@@ -135,17 +130,17 @@ public async Task TestPublicOnlyDataSetVersions_SpecifyWildCard_DoesNotReturnDra
                 .ForIndex(2, dsv => dsv.SetVersionNumber(1, 2))
                 .ForIndex(3, dsv =>
                 {
-                    dsv.SetStatusDraft();
+                    dsv.SetStatus(status); //The status is not published status and so these will not be returned
```
leeoades, line 133: It would be clearer then to rename the variable `status` to `nonPublishedStatus`.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model/Extensions/DataSetVersionQueryableExtensions.cs
```diff
@@ -19,13 +19,21 @@ public static async Task<Either<ActionResult, DataSetVersion>> FindByVersion(
...
+        var query = queryable
+            .Where(dsv => dsv.DataSetId == dataSetId);
+
+        if (parsedVersion.IsWildcard)
+        {
+            query = query.WherePublishedStatus();
+        }
+
+        return await query
             .Where(dsv => dsv.DataSetId == dataSetId)
```
leeoades, line 37: You're filtering for the dataSetId twice.

leeoades (changes requested):
A few minor issues.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Controllers/DataSetVersionsControllerTests.cs
mmoalin: Great point, I've got rid of the duplication by putting the shared set up code in a new class, please let me know what you think

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests/Extensions/DataSetVersionQueryableExtensionsTests.cs
```diff
@@ -154,4 +249,12 @@ private IQueryable<DataSetVersion> SetupDataSetVersions(out Guid dataSetGuid)
...
         };
     }
+
+    private static class DataSetVersionStatusQueryTheoryData
+    {
+        public static IEnumerable<object[]> NonPublishedStatusWithVersionString =>
+            from status in EnumUtil.GetEnums<DataSetVersionStatus>().Except(new[] { DataSetVersionStatus.Published })
+            from versionString in new[] { "1.*", "*", "v1.*", "v*" }
+            select new object[] { status, versionString };
+    }
```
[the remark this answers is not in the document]
mmoalin: Thanks Ben, you're right, should've taken more precaution with the typing and using a separate class. I've now adopted your suggested approach to use `MemberData` and have the `TheoryData` get generated from a method within the test class.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests/Fixture/IntegrationTestFixtureWithCommonTestDataSetup.cs
```diff
@@ -0,0 +1,65 @@
+﻿using GovUk.Education.ExploreEducationStatistics.Common.Model.Data;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Database;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Model.Tests.Fixtures;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Tests.Fixture;
+
+public abstract class IntegrationTestFixtureWithCommonTestDataSetup(TestApplicationFactory testApp) : IntegrationTestFixture(testApp)
```
leeoades, line 8: Wow, that cleared up a load of copied code! I didn't mean for you to do this now though, but thanks. I expect there is a lot more duplicated code in there that can be cleaned up.

benoutram (approved):

jack-hive (approved):

leeoades (approved):

PR #5811, opened 2025-04-17
Add "Is Indexer Running" check.

leeoades:
When multiple searchable documents are created, or other changes occur that cause the indexer to be requested to reindex multiple times in quick succession, we were experiencing errors whether the indexer was already running.

This PR introduces a check to see if the indexer is already running. Also, because of the async nature of things, we also need to catch the exception that occurs when we attempt to start an indexer that is already running.

This simple first step should remove the errors. It does not address the much harder problem of ensuring a reindex occurs once all of the documents have been refreshed. That will come in a future PR.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Clients/AzureSearch/SearchIndexClient.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Clients/AzureSearch/SearchIndexClient.cs
@@ -1,4 +1,5 @@
 ﻿using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Options;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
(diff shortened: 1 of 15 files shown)


benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Clients/AzureSearch/SearchIndexClient.cs
```diff
@@ -8,17 +9,41 @@ namespace GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.
 /// </summary>
 public class SearchIndexClient(
```
benoutram, line 10: Not really part of this PR but I think this should be SearchIndexerClient rather than SearchIndexClient?

leeoades: Good spot. I included the rename too. Harmless enough.

PR #5822, opened 2025-04-25
EES-5977 Add analytics for zip downloads

mmyoungman:
This PR adds analytics for zip downloads.It essentially copies the pattern established by EES-5242 when we started recording analytics for Public API queries, with a couple of tweaks.

As part of this work we explored whether the the EES-5242 approach is suitable for both this and for future analytics work where things that might be requested more frequently like release page views) and gave it the OK. If a performance problem arises we could at the very worst turn analytics off and would be relatively straightforward to change approach. 

When testing this work I did see that the Azure Function would take several minutes to process a couple thousands request files, so I do think that could become a problem in the future if we have a flood of requests in a short space of time and the Azure Function times out before they're all processed and/or another execution of the function overlaps the previous. But there are ways to mitigate these issues.

The high-level overview is that every time a user downloads a zip file from either a release page or a data set page, a file is created to record data associated with that zip download. (A request without a `SubjectId` is to download all data sets for a release version. A request with a `SubjectId` is only for that specific data set.) These files are then consumed by the analytics Azure function that processes all analytics request files once an hour and generates parquet file reports. These will be used by the DfE for their own purposes. The end users of the parquet files are statisticians and are used to working with data in this way.

Locally, the files and reports will be created in the `explore-education-statistics/data/analytics` directory - this can be edited in the Content API's `appsettings.Development.json`. On environments, the files will be written to a file share in the analytics Azure Storage instance. This fileshare is mounted with a file path `\\mounts\\analytics` - file paths must be provided in this "/mounts/{name}" format for app services. In unit tests, files are written to "ExploreEducationStatistics" in your OS's temp directory.

The writing of the original files to record the zip download are written using a `Channel`. We defer the creation of request files to a background task to avoid it interfering with the performance of more important tasks. `AnalyticsManager` creates the `Channel`. `AnalyticsWriter` writes request data into the channel, and then `AnalyticsConsumer` reads from the channel as a background task and creates the actual files (which will later be processed by the Azure Function and turned into a parquet file report).

The Channel is set up differently to `QueryAnalyticsManager`, where we have one channel for all recorded requests, rather than a separate channel for each type of recorded request. `QueryAnalyticsManager` will be aligned with the Content DB approach in EES-6066. 

The Azure Function is hooked up to an Azure Storage instance (`eessaanlyt`) and that is where the parquet files are saved.

### Other changes/notes
- `ReleaseFileService#ZipFilesToStream` allows users to specific an array of files that they want included in a zip. The public frontend used to allow users to specify a number of data set to be included in the zip, but this is no longer the case: users either download a zip with all data sets related to a release version or a sole data set. I've created EES-6034 to update `ZipFilesToStream` to take a single optional FileId as an argument.

- I've added Dispose methods to various analytic test classes to ensure temporary files get cleared up after running the unit tests.
- When INSERTing into the DuckDb tables, I've changed the format to `unstructured`. This was to prevent an issue where DuckDb wanted JSON in request files to be in an array. The DuckDB docs are here: https://duckdb.org/docs/stable/data/json/format_settings.html I believe the `PublicApiQueriesProcessor` INSERT format, which was set to `auto` would have been using `unstructured`, but for some reason the same didn't happen with the `PublicZipDownloadsProcessor` INSERT, so I needed to set `unstructured` explicitly.
- Renamed `ConsumePublicApiQueriesProcessorTests` to `PublicApiQueriesProcessorTests`.
- Removed an unnecessary `IF NOT EXISTS` from a DuckDb queries - the tables are created fresh each Azure Function run so no need for it.
- I've updated `Analytics.Consumer`'s `TestAnalyticsPathResolver`'s directories to match the structure from the real `AnalyticsPathResolver`.

### DIFF
--- a/infrastructure/templates/template.json
+++ b/infrastructure/templates/template.json
@@ -750,6 +750,20 @@
       "metadata": {
         "description": "The Client ID of a manually-created App Registration that represents the Public API Container App in Entra ID"
       }
+    },
+    "analyticsEnabled": {
+      "type": "bool",
+      "defaultValue": true,
+      "metadata": {
+        "description": "Whether analytics is enabled"
+      }
+    },
+    "analyticsFileShareMountPath": {
+      "type": "string",
+      "defaultValue": "\\mounts\\analytics",
+      "metadata": {
+        "description": "Where analytics request files are saved before being processed"
+      }
     }
   },
   "variables": {
@@ -1242,6 +1256,8 @@
     "publicDataFileShareMountPathWindows": "\\mounts\\public-api-data",
     "publicDataFileShareName": "[concat(parameters('subscription'), '-ees-papi-share-data')]",
     "publicDataStorageAccountName": "[concat(parameters('subscription'), 'eespapisa')]",
+    "analyticsFileShareName": "[concat(parameters('subscription'), '-ees-share-anlyt')]",
+    "analyticsStorageAccountName": "[concat(parameters('subscription'), 'eessaanlyt')]",
     "eventGridTopicNamePublicationChanged": "[concat(parameters('subscription'), '-', parameters('environment'), '-evgt-publication-changed')]",
     "eventGridTopicNameReleaseChanged": "[concat(parameters('subscription'), '-', parameters('environment'), '-evgt-release-changed')]",
     "eventGridTopicNameReleaseVersionChanged": "[concat(parameters('subscription'), '-', parameters('environment'), '-evgt-release-version-changed')]",
@@ -1576,7 +1592,29 @@
         "WEBSITE_NODE_DEFAULT_VERSION": "20.16.0",
         "WEBSITE_RUN_FROM_PACKAGE": "1",
         "PublicStorage": "[concat('@Microsoft.KeyVault(SecretUri=', reference(variables('ees-storage-public'), '2018-02-14').secretUriWithVersion, ')')]",
-        "enableSwagger": "[parameters('enableSwagger')]"
+        "enableSwagger": "[parameters('enableSwagger')]",
+        "Analytics__Enabled": "[parameters('analyticsEnabled')]",
+        "Analytics__BasePath": "[parameters('analyticsFileShareMountPath')]"
+      }
+    },
+    {
+      "condition": "[parameters('analyticsEnabled')]",
+      "name": "[concat(variables('contentAppName'), '/azurestorageaccounts')]",
+      "type": "Microsoft.Web/sites/config",
+      "location": "[resourceGroup().location]",
+      "apiVersion": "2019-08-01",
+      "dependsOn": [
+        "[resourceId('Microsoft.Web/sites', variables('contentAppName'))]"
+      ],
+      "properties": {
+        "[variables('analyticsFileShareName')]": {
+          "type": "AzureFiles",
+          "accountName": "[variables('analyticsStorageAccountName')]",
+          "accessKey": "[listKeys(resourceId('Microsoft.Storage/storageAccounts', variables('analyticsStorageAccountName')), '2018-02-01').keys[0].value]",
+          "shareName": "[variables('analyticsFileShareName')]",
+          "mountPath": "[variables('analyticsFileShareMountPath')]",
+          "protocol": "Smb"
+        }
       }
     },
     {
--- a/src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer.Tests/Services/PublicZipDownloadsProcessorTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer.Tests/Services/PublicZipDownloadsProcessorTests.cs
@@ -1,10 +1,14 @@
 using System.Reflection;
+using System.Security.Cryptography;
+using System.Text;
+using AngleSharp.Io;
 using GovUk.Education.ExploreEducationStatistics.Analytics.Consumer.Services;
 using GovUk.Education.ExploreEducationStatistics.Common.DuckDb.DuckDb;
 using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
 using InterpolatedSql.Dapper;
 using Microsoft.Extensions.Logging;
 using Moq;
+using Newtonsoft.Json;
 using Xunit;
 
 namespace GovUk.Education.ExploreEducationStatistics.Analytics.Consumer.Tests.Services;
@@ -21,7 +25,7 @@ public abstract class PublicZipDownloadsProcessorTests
         [Fact]
         public async Task NoSourceFolder_NoReportProduced()
         {
-            var pathResolver = new TestAnalyticsPathResolver();
+            using var pathResolver = new TestAnalyticsPathResolver();
 
             var service = BuildService(
                 pathResolver: pathResolver);
@@ -34,7 +38,7 @@ public abstract class PublicZipDownloadsProcessorTests
         [Fact]
         public async Task NoRequestFilesToConsume_NoReportProduced()
         {
-            var pathResolver = new TestAnalyticsPathResolver();
+            using var pathResolver = new TestAnalyticsPathResolver();
 
             Directory.CreateDirectory(pathResolver.PublicZipDownloadsDirectoryPath());
 
@@ -51,7 +55,7 @@ public abstract class PublicZipDownloadsProcessorTests
         [Fact]
         public async Task SingleRequestFileNoSubjectId_ProducesOneReportRow()
         {
-            var pathResolver = new TestAnalyticsPathResolver();
+            using var pathResolver = new TestAnalyticsPathResolver();
             SetupZipDownloadRequest(pathResolver, "ZipDownloadRequestFile_NoSubjectId.json");
 
             var service = BuildService(
@@ -77,20 +81,16 @@ public abstract class PublicZipDownloadsProcessorTests
             // match the expected values also.
             var zipDownloadReportRow = Assert.Single(zipDownloadReportRows);
 
-            Assert.Equal("2e1ff1faca8870e00a9ec1fab7e58409", zipDownloadReportRow.ZipDownloadHash);
-            Assert.Equal("publication name", zipDownloadReportRow.PublicationName);
-            Assert.Equal(Guid.Parse("319750f6-4c33-476c-9e6d-3da7a403201d"), zipDownloadReportRow.ReleaseVersionId);
-            Assert.Equal("release name", zipDownloadReportRow.ReleaseName);
-            Assert.Equal("release label", zipDownloadReportRow.ReleaseLabel);
-            Assert.Null(zipDownloadReportRow.SubjectId);
-            Assert.Null(zipDownloadReportRow.DataSetName);
-            Assert.Equal(1, zipDownloadReportRow.Downloads);
+            await AssertZipDownloadReportRow(
+                zipDownloadReportRow,
+                "ZipDownloadRequestFile_NoSubjectId.json",
+                1);
         }
 
         [Fact]
         public async Task TwoDifferentSourceQueries_ProduceTwoDistinctReportRows()
         {
-            var pathResolver = new TestAnalyticsPathResolver();
+            using var pathResolver = new TestAnalyticsPathResolver();
 
             SetupZipDownloadRequest(pathResolver, "ZipDownloadRequestFile_NoSubjectId.json");
             SetupZipDownloadRequest(pathResolver, "ZipDownloadRequestFile_WithSubjectId.json");
@@ -111,29 +111,21 @@ public abstract class PublicZipDownloadsProcessorTests
 
             Assert.Equal(2, zipDownloadReportRows.Count);
 
-            Assert.Equal("2e1ff1faca8870e00a9ec1fab7e58409", zipDownloadReportRows[0].ZipDownloadHash);
-            Assert.Equal("publication name", zipDownloadReportRows[0].PublicationName);
-            Assert.Equal(Guid.Parse("319750f6-4c33-476c-9e6d-3da7a403201d"), zipDownloadReportRows[0].ReleaseVersionId);
-            Assert.Equal("release name", zipDownloadReportRows[0].ReleaseName);
-            Assert.Equal("release label", zipDownloadReportRows[0].ReleaseLabel);
-            Assert.Null(zipDownloadReportRows[0].SubjectId);
-            Assert.Null(zipDownloadReportRows[0].DataSetName);
-            Assert.Equal(1, zipDownloadReportRows[0].Downloads);
+            await AssertZipDownloadReportRow(
+                zipDownloadReportRows[0],
+                "ZipDownloadRequestFile_NoSubjectId.json",
+                1);
 
-            Assert.Equal("670871b1327feb682dd3516374f35928", zipDownloadReportRows[1].ZipDownloadHash);
-            Assert.Equal("publication name 2", zipDownloadReportRows[1].PublicationName);
-            Assert.Equal(Guid.Parse("72a6856c-8d7b-4ad9-b533-2066d171d146"), zipDownloadReportRows[1].ReleaseVersionId);
-            Assert.Equal("release name 2", zipDownloadReportRows[1].ReleaseName);
-            Assert.Equal("release label 2", zipDownloadReportRows[1].ReleaseLabel);
-            Assert.Equal(Guid.Parse("66c5eb5f-a85b-4586-bae8-dc3504d3042f"), zipDownloadReportRows[1].SubjectId);
-            Assert.Equal("data set name 2", zipDownloadReportRows[1].DataSetName);
-            Assert.Equal(1, zipDownloadReportRows[1].Downloads);
+            await AssertZipDownloadReportRow(
+                zipDownloadReportRows[1],
+                "ZipDownloadRequestFile_WithSubjectId.json",
+                1);
         }
 
         [Fact]
         public async Task MultipleRequestFilesForSameZipFile_ProduceSingleReportRow()
         {
-            var pathResolver = new TestAnalyticsPathResolver();
+            using var pathResolver = new TestAnalyticsPathResolver();
 
             SetupZipDownloadRequest(pathResolver, "ZipDownloadRequestFile_WithSubjectId.json");
             SetupZipDownloadRequest(pathResolver, "ZipDownloadRequestFile_WithSubjectId_Copy.json");
@@ -154,14 +146,10 @@ public abstract class PublicZipDownloadsProcessorTests
 
             var zipDownloadReportRow = Assert.Single(zipDownloadReportRows);
 
-            Assert.Equal("670871b1327feb682dd3516374f35928", zipDownloadReportRow.ZipDownloadHash);
-            Assert.Equal("publication name 2", zipDownloadReportRow.PublicationName);
-            Assert.Equal(Guid.Parse("72a6856c-8d7b-4ad9-b533-2066d171d146"), zipDownloadReportRow.ReleaseVersionId);
-            Assert.Equal("release name 2", zipDownloadReportRow.ReleaseName);
-            Assert.Equal("release label 2", zipDownloadReportRow.ReleaseLabel);
-            Assert.Equal(Guid.Parse("66c5eb5f-a85b-4586-bae8-dc3504d3042f"), zipDownloadReportRow.SubjectId);
-            Assert.Equal("data set name 2", zipDownloadReportRow.DataSetName);
-            Assert.Equal(2, zipDownloadReportRow.Downloads);
+            await AssertZipDownloadReportRow(
+                zipDownloadReportRow,
+                "ZipDownloadRequestFile_WithSubjectId.json",
+                2);
         }
 
         private static async Task<List<ZipDownloadReportLine>> ReadZipDownloadReport(DuckDbConnection duckDbConnection, string reportFile)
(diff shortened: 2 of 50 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/StrategiesTests/AnalyticsWritePublicZipDownloadStrategyTests.cs
```diff
@@ -0,0 +1,106 @@
...
+                .ToList();
+
+            Assert.Equal(2, files.Count);
+
+            var requestFile1 = Assert.Single(
+                files
+                    .Where(file => file.Contains(releaseVersionId1.ToString()))
+                    .ToList());
+            var requestFile1Contents = await File.ReadAllTextAsync(requestFile1);
+            Snapshot.Match(
```
leeoades, line 83: This isn't the way that I've seen Snapshot being used. It's usually used to run a test and capture a state (in a snapshot, no less). The point being that if the results of that test changes down the line then the snapshot diff will change and the test will fail.
I've not seen it used as a json comparison tool like this. Seems a bit like it's being abused...

src/GovUk.Education.ExploreEducationStatistics.Content.Services/AnalyticsConsumer.cs
```diff
@@ -0,0 +1,21 @@
...
+
+public class AnalyticsConsumer(
+    IAnalyticsManager analyticsManager,
+    IAnalyticsWriter analyticsWriter) : BackgroundService
+{
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        while (!stoppingToken.IsCancellationRequested)
+        {
+            var request = await analyticsManager.Read(stoppingToken);
```
leeoades, line 16: If an exception occurs in here then the whole background service will stop. So you should guard against that.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/AnalyticsPathResolver.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    private readonly string _basePath;
+
+    public AnalyticsPathResolver(IOptions<AnalyticsOptions> options, IWebHostEnvironment environment)
+    {
+
+        if (options.Value.BasePath.IsNullOrWhitespace())
+        {
+            throw new ArgumentException(
+                message: $"'{nameof(AnalyticsOptions.BasePath)}' must not be blank",
```
leeoades, line 22: It would be more useful here to explain that you mean config and output the full section as well as the property name so that if it occurs, it's easier to track down which setting is missing.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/AnalyticsPathResolver.cs
```diff
@@ -0,0 +1,42 @@
...
+                paramName: nameof(options)
+            );
+        }
+
+        var originalPath = options.Value.BasePath;
+        if (environment.IsDevelopment())
```
leeoades, line 28: It isn't advisable to have this kind of dev only branching code in the actual production class. Instead, create a `LocalDevelopmentAnalyticsPathResolver` and register that into the container instead during service registration.
What's more, `LocalDevelopmentAnalyticsPathResolver`  could derive from `AnalyticsPathResolver`, and you could make BasePath virtual, allowing it to be overridden. Then in AnalyticsPathResolver, make sure you refer to BasePath(), not _basePath and Bob is your uncle.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/AnalyticsWriter.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Services;
+
+public class AnalyticsWriter(
+    IEnumerable<IAnalyticsWriteStrategy> writeStrategies) : IAnalyticsWriter
+{
+    public async Task Report(BaseCaptureRequest request, CancellationToken cancellationToken)
+    {
+        var strategy = writeStrategies
+            .Single(s => s.CanHandle(request));
```
leeoades, line 17: If you use Single, that means it will check every strategy on every call to make sure there is exactly one - which could be a bit wasteful. Are you envisaging other strategies for different analytics messages? If the differentiator is the Type of message then you could instead of CanHandle have a property that returns `Type CaptureRequestType` and put them into a dictionary. That way you can go straight to the strategy you want.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Requests/CaptureRequests.cs
```diff
@@ -0,0 +1,14 @@
+#nullable enable
+using System;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Services.Requests;
+
+public abstract record BaseCaptureRequest;
```
leeoades, line 6: The naming convention is that `Base` is the suffix for an abstract class. It also doesn't seem to reference analytics at all - so it's quite hard to ascertain its purpose. This is to represent an event and its data that we want to send to the analytics pool of data, right? Is it only for requests? I'm not sure what a "capture request" is?

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Requests/CaptureRequests.cs
```diff
@@ -0,0 +1,14 @@
...
+
+public record CaptureZipDownloadRequest(
```
leeoades, line 8: Please could you add a comment to this class to explain its purpose?

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Strategies/AnalyticsWritePublicZipDownloadStrategy.cs
```diff
@@ -0,0 +1,56 @@
...
+using Newtonsoft.Json;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Services.Strategies;
+
+public class AnalyticsWritePublicZipDownloadStrategy(
+    IAnalyticsPathResolver analyticsPathResolver,
+    ILogger<AnalyticsWritePublicZipDownloadStrategy> logger
+    ) : IAnalyticsWriteStrategy
+{
+    public bool CanHandle(BaseCaptureRequest request)
```
leeoades, line 20: I've mentioned above that unless there are likely to be multiple strategies handling the same Type, you're probably better off exposing the request Type so that the AnalyticsWriter can go straight for the strategy to use.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Utils/AnalyticsUtils.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Services.Utils;
+
+public static class AnalyticsUtils
+{
+    public static async Task WriteToFileShare(
```
leeoades, line 10: Is there a particular reason to extract this method like this? If all of the AnalyticsWriteStrategies would benefit from it then perhaps it would be better placed in a AnalyticsWriteStrategyBase ?

leeoades (commented):
Just some thoughts and observations. Feel free to call me to discuss.

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer.Tests/Services/PublicZipDownloadsProcessorTests.cs
```diff
@@ -0,0 +1,203 @@
...
+            Assert.EndsWith("public-zip-downloads.parquet", zipDownloadsReport);
+
+            var duckDbConnection = new DuckDbConnection();
+            duckDbConnection.Open();
+
+            var zipDownloadReportRows = await ReadZipDownloadReport(duckDbConnection, zipDownloadsReport);
+
+            Assert.Equal(2, zipDownloadReportRows.Count);
+
+            Assert.Equal("2e1ff1faca8870e00a9ec1fab7e58409", zipDownloadReportRows[0].ZipDownloadHash);
```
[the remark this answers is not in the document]
mmyoungman: I've added an `.OrderBy(...)` to `PublicZipDownloadsProcesorTests#ReadZipDownloadReport` and `PublicApiQueriesProcessorTests#ReadQueryReport`.

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Services/PublicZipDownloadsProcessor.cs
```diff
@@ -0,0 +1,151 @@
...
+            reportsDirectory,
+            $"{reportFilenamePrefix}_public-zip-downloads.parquet");
+
+        duckDbConnection.ExecuteNonQuery($@"
+            COPY (SELECT * FROM zipDownloadsReport)
+            TO '{zipDownloadReportFilename}' (FORMAT 'parquet', CODEC 'zstd')");
+
+        Directory.Delete(processingDirectory, recursive: true);
+
+        return Task.CompletedTask;
```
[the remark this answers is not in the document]
mmyoungman: As far as I'm aware. The same is also true of `PublicApiQueriesProcessor#Process`

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests.csproj
```diff
@@ -27,4 +27,8 @@
         <ProjectReference Include="..\GovUk.Education.ExploreEducationStatistics.Content.Model.Tests\GovUk.Education.ExploreEducationStatistics.Content.Model.Tests.csproj" />
         <ProjectReference Include="..\GovUk.Education.ExploreEducationStatistics.Data.Model.Tests\GovUk.Education.ExploreEducationStatistics.Data.Model.Tests.csproj" />
     </ItemGroup>
+
+    <ItemGroup>
+      <Folder Include="Services\" />
```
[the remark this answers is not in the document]
mmyoungman: Removed

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Startup.cs
```diff
@@ -246,5 +268,19 @@ public void Configure(IApplicationBuilder app,
                 ?.Addresses
                 .ForEach(address => Console.WriteLine($"Server listening on address: {address}"));
         }
+
+        private class NoOpAnalyticsManager : IAnalyticsManager
```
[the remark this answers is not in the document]
mmyoungman: I've also moved `NoOpQueryAnalyticsManager` into `QueryAnalyticsManger.cs`

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/StrategiesTests/AnalyticsWritePublicZipDownloadStrategyTests.cs
mmyoungman: I am copying the pattern from `QueryAnalyticsWriterTests.cs`. I'm personally okay with it - saves us having to deserialize the json from the created files in the tests and makes updating the tests easy. So I'm going to leave it, but if you feel strongly about it happy to discuss.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Utils/AnalyticsUtils.cs
mmyoungman: It's possible that a write strategy in the future won't use a fileshare, and with the util method we avoid inheritance, which we generally haven't used much around the codebase and I like the idea of keeping language features use to a minimum - I think it's helps general readability.

All that said, either approach is fine with me and I was interested in what the change would look like, so I moved `WriteToFileShare` into `AnalyticsWriteStrategyBase`.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Requests/CaptureRequests.cs
mmyoungman: I've renamed the base record to `AnalyticsCaptureRequestBase` and added a comment to `CaptureZipDownloadRequest` to explain it's use.

`CaptureZipDownloadRequest` is following the naming pattern set by `CaptureDataSetVersionQueryRequest`.

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Services/PublicZipDownloadsProcessor.cs
```diff
@@ -0,0 +1,151 @@
...
+        ");
+
+        // We fetch the files again in case there are files leftover in the processing dir from a previous function run
+        var filesReadyForProcessing = Directory
+            .GetFiles(processingDirectory)
+            .Select(Path.GetFileName)
+            .OfType<string>()
+            .ToList();
+
+        foreach (var filename in filesReadyForProcessing)
```
duncan-at-hiveit, line 75: I'm guessing that the `foreach` processing here and in the Public API analytics might be the cause for the Function App execution being slow (although I can't prove this 100%!).

If that's the case and the performance becomes a problem, I might suggest that we switch to either all-in-one as we'd done before, or in batches off 100, say.  If we hit a bad batch, we move them into `failures` and continue to the next.

All this'd need proving of course, but certainly on my local environment, it took 10 seconds to handle 40,000 JSON files.

src/GovUk.Education.ExploreEducationStatistics.Common.Tests/Utils/MockUtils.cs
```diff
@@ -189,5 +192,16 @@ private static Mock<TConfiguration> PopulateMockConfiguration<TConfiguration>(Tu
 
             return configuration;
         }
+
+        public static void LoggerSetup<T>(Mock<ILogger<T>> logger, LogLevel logLevel, string logMessage)
```
duncan-at-hiveit, line 196: I would personally rename this to something like `SetupLoggerExpectation` or `ExpectLogMessage`.  `LoggerSetup` is quite a generic name that doesn't broadcast what its purpose is to me.

duncan-at-hiveit: Or you could make it an extension method of Mock<ILogger<T>> and then you could rewrite:

`MockUtils.LoggerSetup(logger, LogLevel.Warning, expectedWarning);`

as:

`logger.SetupExpectedMessage(LogLevel.Warning, expectedWarning)` or `logger.SetupExpectedWarning(expectedWarning)`

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ReleaseFileServiceTests.cs
```diff
@@ -340,13 +337,29 @@ public async Task ZipFilesToStream_DataGuidanceForMultipleDataFiles()
...
+                PublicationName: releaseVersion.Release.Publication.Title,
+                ReleaseVersionId: releaseVersion.Id,
+                ReleaseName: releaseVersion.Release.Title,
+                ReleaseLabel: releaseVersion.Release.Label,
+                SubjectId: releaseFile.File.SubjectId,
+                DataSetName: releaseFile.Name
+            );
+            var analyticsManager = new Mock<IAnalyticsManager>(MockBehavior.Strict);
+            analyticsManager.Setup(m => m.Add(
+                    It.Is<CaptureZipDownloadRequest>(r =>
```
duncan-at-hiveit, line 358: We have an `ItIs.DeepEqualTo` method too!

duncan-at-hiveit: Applicable here and a few other places too.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ReleaseFileServiceTests.cs
```diff
@@ -548,11 +459,26 @@ public async Task ZipFilesToStream_FiltersInvalidFileTypes()
 
             var publicBlobStorageService = new Mock<IPublicBlobStorageService>(MockBehavior.Strict);
 
+            var captureRequest = new CaptureZipDownloadRequest
```
duncan-at-hiveit, line 462: The other option here (and one that I used in Public API) is to set analytics enabled to false by default, so you don't have to add additional analytics-related expectations everywhere that might otherwise make non-related tests noisy.

Then I had a set of tests specifically for testing when it is on.

Don't know if this is any use to you or if you're happy with where you've added new expectations in here?

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Requests/CaptureRequests.cs
```diff
@@ -0,0 +1,14 @@
...
+    string PublicationName,
+    Guid ReleaseVersionId,
+    string ReleaseName,
+    string? ReleaseLabel,
+    Guid? SubjectId = null,
```
duncan-at-hiveit, line 16: Under what circumstances would SubjectId be null out of interest?

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Strategies/AnalyticsWritePublicZipDownloadStrategy.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        var serialisedRequest = JsonSerializationUtils.Serialize(
+            obj: zipDownloadRequest,
+            formatting: Formatting.Indented,
+            orderedProperties: true,
+            camelCase: true);
+
+        var directory = analyticsPathResolver.PublicZipDownloadsDirectoryPath();
+
+        var filename =
```
duncan-at-hiveit: I'm assuming that adding the random string to the end of the filename successfully stopped the invalid JSON being produced in that other bug?

infrastructure/templates/template.json
```diff
@@ -1576,7 +1592,29 @@
         "WEBSITE_NODE_DEFAULT_VERSION": "20.16.0",
         "WEBSITE_RUN_FROM_PACKAGE": "1",
         "PublicStorage": "[concat('@Microsoft.KeyVault(SecretUri=', reference(variables('ees-storage-public'), '2018-02-14').secretUriWithVersion, ')')]",
-        "enableSwagger": "[parameters('enableSwagger')]"
+        "enableSwagger": "[parameters('enableSwagger')]",
+        "Analytics__Enabled": "[parameters('analyticsEnabled')]",
+        "Analytics__BasePath": "[parameters('analyticsFileShareMountPath')]"
+      }
+    },
+    {
```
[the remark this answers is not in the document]
mmyoungman: That assumes this works :)

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Requests/CaptureRequests.cs
mmyoungman: It is null when someone is downloading a zip with all data sets for a particular release version.

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Services/PublicZipDownloadsProcessor.cs
mmyoungman: I wrote some code to process files in batches, but have a few concerns about how easy it will be to spot which file is malformed if it fails to process a batch of say 100 files - I cannot for the life of me remember if it tells you the offending file when you ask DuckDb to process multiple files at once...

Worst case, maybe we can do something like process files in batches and then process the failed files one by one after the fact.

All that said, I doubt we'll get malformed files often, so maybe we don't need to care about this too much.

I'm inclined to leave this for this PR and do it in a separate ticket/PR after this, if you're cool with that. And I do mean do it rather than let it rot in the backlog - I think we should solve this issue.

duncan-at-hiveit: I remember that it *does* tell you the name of the offending file, if that helps!

duncan-at-hiveit: Happy to leave for a separate task! :+1:

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/ReleaseFileServiceTests.cs
duncan-at-hiveit: Yup that's correct.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Requests/CaptureRequests.cs
duncan-at-hiveit: Gotcha thanks.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Strategies/AnalyticsWritePublicZipDownloadStrategy.cs
duncan-at-hiveit: Nice one - fingers crossed then, we should never *actually* end up with invalid JSON!

duncan-at-hiveit (approved):

PR #5833, opened 2025-04-30
Invalidate Publication Cache on Theme change

leeoades:
When a Theme is updated, it wasn't invalidating the publication cache and therefore getting publications still contained the old Theme details. Now, we invalidate the cache if a theme changes.


src/GovUk.Education.ExploreEducationStatistics.Content.Services/Cache/PublicationCacheService.cs
```diff
@@ -77,6 +78,32 @@ public Task<IList<PublicationTreeThemeViewModel>> UpdatePublicationTree()
         return _publicationService.GetPublicationTree();
     }
 
+    /// <summary>
+    /// Invalidate all Publications in the cache for the specified Theme
+    /// </summary>
+    public Task<Either<ActionResult, Unit>> InvalidatePublicationsByTheme(Guid themeId)
+    {
+        return _publicationService
+            .ListPublications(themeId: themeId, pageSize: int.MaxValue)
```
benoutram, line 87: I don't think `ListPublications` is really designed to be used like this, it's intended to be the endpoint for the public Find Stats search.

It's paginated, sorted, it builds a complex view model that's not used at all here, and internally it produces a SQL query joined with a free text view even if no search term is provided.

It doesn't return superseded/archived publications so invalidating these would be missed.

These publications are still accessible in the public site even though they don't appear in search results.

I would expect that 'invalidating publications by theme' would invalidate all cached publications for the specified theme.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Cache/PublicationCacheService.cs
```diff
@@ -77,6 +78,32 @@ public Task<IList<PublicationTreeThemeViewModel>> UpdatePublicationTree()
...
+    /// Invalidates the Publication in the cache. Errors are logged and not returned.
+    /// </summary>
+    private Task<Either<Unit, Unit>> InvalidatePublicationSafe(string publicationSlug) =>
+        UpdatePublication(publicationSlug)
+            .OnFailureVoid(
+                result => _logger.LogWarning(
+                    "Failed to invalidate cache for Publication {PublicationSlug}. Reason: {Result}",
+                    publicationSlug,
+                    result))
+            .OnSuccessVoid();
```
benoutram, line 105: I feel like something like this would be more readable than these two methods put together, and it returns `Task` as well.

Your version returns `Task<Either<ActionResult, Unit>>` but looking at the implementation of it I don't think any failure can occur except for one in `ListPublications`? See my other comment about why I don't think that using `ListPublications` here is a good idea.

Your consuming code in `ThemeService` ignores any failure returned here too, so if there was one returned here it would continue anyway.

```csharp
    public async Task InvalidatePublicationsByTheme(Guid themeId)
    {
        var publicationSlugs = await _contentDbContext.Publications
            .Where(p => p.ThemeId == themeId)
            .Select(p => p.Slug)
            .ToListAsync();

        await publicationSlugs
            .ToAsyncEnumerable()
            .ForEachAwaitAsync(async publicationSlug => await UpdatePublication(publicationSlug)
                .OnFailureDo(result => _logger.LogWarning(
                    "Failed to invalidate cache for Publication {PublicationSlug}. Reason: {Result}",
                    publicationSlug,
                    result)));
    }
```

Here we only select a single column. We don't build any search view models which are thrown away. We don't rely on something else to tell us what publications may exist in the public cache but attempt it for all publications regardless which I think is safer. It ignores failures invalidating the cache entries in the same way as before.

leeoades: Oh that's so much better than what I was fumbling about with. Thank you!

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/Cache/PublicationCacheServiceTests.cs
```diff
@@ -11,7 +11,11 @@
 using Moq;
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
+using GovUk.Education.ExploreEducationStatistics.Common.Cache.Interfaces;
+using GovUk.Education.ExploreEducationStatistics.Content.Services.Tests.Builders;
+using Org.BouncyCastle.Bcpg;
```
benoutram, line 18: I think these changes are left over and should be reverted.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Cache/PublicationCacheService.cs
```diff
@@ -14,6 +14,7 @@
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
+using GovUk.Education.ExploreEducationStatistics.Common.Extensions;
```
benoutram, line 17: Looks like an unused import

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Interfaces/Cache/IPublicationCacheService.cs
```diff
@@ -1,4 +1,5 @@
 #nullable enable
+using System;
```
benoutram, line 2: Looks like this could be another unused import?

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/Builders/PublicationServiceMockBuilder.cs
```diff
@@ -0,0 +1,96 @@
...
+using GovUk.Education.ExploreEducationStatistics.Common.ViewModels;
+using GovUk.Education.ExploreEducationStatistics.Content.Model;
+using GovUk.Education.ExploreEducationStatistics.Content.Requests;
+using GovUk.Education.ExploreEducationStatistics.Content.Services.Interfaces;
+using GovUk.Education.ExploreEducationStatistics.Content.ViewModels;
+using Moq;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Services.Tests.Builders;
+
+public class PublicationServiceMockBuilder
```
benoutram, line 14: I guess this whole `PublicationServiceMockBuilder` is not used now, so could be removed, but worth leaving in?

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/Builders/PublicationServiceMockBuilder.cs
```diff
@@ -0,0 +1,96 @@
...
+                        It.IsAny<ReleaseType?>(),
+                        It.Is<Guid?>(t => themeId == null || themeId.Equals(t)),
+                        It.IsAny<string?>(),
+                        It.IsAny<PublicationsSortBy?>(),
+                        It.IsAny<SortDirection?>(),
+                        It.IsAny<int>(),
+                        It.IsAny<int>(),
+                        It.IsAny<IEnumerable<Guid>?>()),
+                    Times.Once);
+        }
```
benoutram, line 94: `SearchWasFor(Guid? themeId)` on its own looks a bit strange now, if you decide to leave this in.

benoutram (commented):
Cool, this looks a lot better now! I've left you some comments about some unused imports and an unused mock builder to have a read through.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/Builders/PublicationServiceMockBuilder.cs
leeoades: Ah yes, of course. Good point.
I was going to leave it in but since you said search might be changing anyway, I'll just get rid of it.

benoutram (approved):

PR #5808, opened 2025-04-17
EES-5956 Add event handling to Search function app for publication archiving events

benoutram:
This PR adds event handling to the Search Function App for publication archiving events.

When subscribed to publication archiving events, on archiving of a publication the Search Function App will remove searchable documents from the Search blob storage account for any published releases of the publication.

It does this by making a call to the Content API to retrieve all published releases for the publication. Using the release id's it removes searchable documents from the Search blob storage account for any blobs matching the release id, if they exist.

PR #5834, opened 2025-04-30
EES-6103 Subscribe Search Function App to publication archived events

benoutram:
This PR adds in the configuration to subscribe the Search Function App to publication archived events. This was missed from https://github.com/dfe-analytical-services/explore-education-statistics/pull/5808 which added the function and the storage queue but did not subscribe the function via the queue.

### Other changes

- Rename `Slug` to `PublicationSlug` in `PublicationArchivedEventDto` for clarity. Add `PreviousSupersededByPublicationId` and `NewSupersededByPublicationId` as extra information which will be in the event payload.
- Correct unit test `GivenEvent_WhenPayloadContainsSlug_ReturnsExpectedDto` to use `PublicationArchivedEventDto` instead of `PublicationChangedEventDto`.

### DIFF
--- a/infrastructure/templates/search/application/searchDocsFunctionEventSubscriptions.bicep
+++ b/infrastructure/templates/search/application/searchDocsFunctionEventSubscriptions.bicep
@@ -18,6 +18,11 @@ var eventGridCustomTopicSubscriptions = [
     topicName: eventTopics.publicationChanged
     subscriptions: [
       {
+        name: 'publication-archived'
+        includedEventTypes: ['publication-archived']
+        queueName: storageQueueNames.publicationArchived
+      }
+      {
         name: 'publication-changed'
         includedEventTypes: ['publication-changed']
         queueName: storageQueueNames.publicationChanged
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/OnPublicationArchived/Dtos/PublicationArchivedEventDto.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/OnPublicationArchived/Dtos/PublicationArchivedEventDto.cs
@@ -2,5 +2,7 @@
 
 public class PublicationArchivedEventDto
 {
-    public string? Slug { get; init; }
+    public Guid? SupersededByPublicationId { get; init; }
+
+    public string? PublicationSlug { get; init; }
 }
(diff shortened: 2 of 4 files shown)


infrastructure/templates/search/application/searchDocsFunctionEventSubscriptions.bicep
```diff
@@ -17,6 +17,11 @@ var eventGridCustomTopicSubscriptions = [
   {
     topicName: eventTopics.publicationChanged
     subscriptions: [
+      {
+        name: 'publication-archived'
+        includedEventTypes: ['publication-archived']
+        queueName: storageQueueNames.publicationArchived
+      }
```
benoutram, line 24: Note for @leeoades: I wasn't sure here whether to split out a new topic for publication archiving events, given that there will be event types for archiving and unarchiving. Do you have any thoughts?

I've grouped this under the publication changed topic for now.

leeoades: I think having the publication archiving and unarchiving events under the publication topic seems very sensible. The subject for those events is the publication too so looks good to me.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/OnPublicationArchived/Dtos/PublicationArchivedEventDto.cs
```diff
@@ -2,5 +2,9 @@
 
 public class PublicationArchivedEventDto
 {
-    public string? Slug { get; init; }
+    public Guid? PreviousSupersededByPublicationId { get; init; }
```
leeoades, line 5: This might be my lack of understanding and might be clarified by a comment.
If a publication is superseded, then it becomes archived, right?
So if this is an event to indicate that the publication has become archived, how could it have had a SupersededByPublicationId value already? Because if it had one, wouldn't it already be archived?

leeoades (approved):
My questions may be misunderstanding in which case, I've approved this PR.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/OnPublicationArchived/Dtos/PublicationArchivedEventDto.cs
benoutram: We've resolved this on a call now.

I'm going to replace

```csharp
    public Guid? PreviousSupersededByPublicationId { get; init; }

    public Guid? NewSupersededByPublicationId { get; init; }
```

with

```csharp
    public Guid? SupersededByPublicationId { get; init; }
```

The reason for having two was because I was also intending on triggering this event when a publication is already archived but the publication is updated and the publication it is archived by is changed to a different publication. The consumer could do something with this if they were interested. On reflection, this should be a different event if it's ever needed.

PR #5845, opened 2025-05-01
EES-6066 Align Public API analytics code with Content API analytics code

mmyoungman:
This PR aligns the Public API analytics code so it matches the analytics code in the Content API.

Essentially the `QueryAnalytics*` classes have been made more general so (in theory) it's easier for us to add analytics for other requests in the future.

### Other changes
- I've added `#nullable enable` to files in Content that I missed in the other PR
- In some `*AnalyticPathResolver` classes I've removed the injected `IWebHostEnvironment` as it wasn't being used.


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/AnalyticsConsumer.cs
```diff
@@ -1,28 +1,26 @@
...
     protected override async Task ExecuteAsync(CancellationToken stoppingToken)
     {
         while (!stoppingToken.IsCancellationRequested)  
         {
             try
             {
-                var message = await manager.ReadQuery(stoppingToken);
-                await queryAnalyticsWriter.ReportDataSetVersionQuery(message);
+                var message = await manager.Read(stoppingToken);
+                await analyticsWriter.Report(message, stoppingToken);
```
leeoades, line 17: When this cancellationToken signals, it will throw a TaskCanceledException so usually a good idea to catch that first and simply exit - perhaps log an info to say that the task was cancelled.. It avoids seeing errors in the logs when actually, it's just that the task cancelled.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Strategies/AnalyticsWriteStrategyBase.cs
```diff
@@ -0,0 +1,30 @@
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Strategies;
+
+public class AnalyticsWriteStrategyBase(
+    ILogger<AnalyticsWriteStrategyBase> logger)
+{
+    protected async Task WriteFileToShare(
+        string requestTypeName,
+        string directory,
+        string filename,
+        string serialisedRequest)
```
leeoades, line 10: An observation that you don't have to change if you don't want to.
Instead of passing in the serialisedRequest, I would pass in the actual TRequest.
Then move the searilisation into here, because how a request is represented in the file is the concern of this writer code, not the consuming code. There's then a central point for consistency - and one less thing for a consumer to have to know about doing.
Consider if you needed to configure/amend the serialisation - you'd have to find and modify all consumers.
Also, the consumers aren't to be trusted - is it ok for me to base64 encode an image and pass that in here?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/Interfaces/IAnalyticsManager.cs
```diff
@@ -0,0 +1,10 @@
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Requests;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Services.Interfaces;
+
+public interface IAnalyticsManager
```
leeoades, line 5: Funnily enough, I mentioned to Duncan that ideally, this should be two different interfaces:
- one for analytics producers to use to record analytics: IAnalyticsRecorder.Record(...)
- and one for consumers to use to read: IAnalyticsReader.Read
The Manager would implement both.
Why?
Because an interface signals intent - and no component would want to both record and consume the analytics.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
```diff
@@ -253,9 +255,9 @@ public void ConfigureServices(IServiceCollection services)
 
         if (_analyticsOptions.Enabled)
         {
-            services.AddSingleton<IQueryAnalyticsManager, QueryAnalyticsManager>();
-            services.AddHostedService<QueryAnalyticsConsumer>();
-            services.AddSingleton<IQueryAnalyticsWriter, QueryAnalyticsWriter>();
+            services.AddSingleton<IAnalyticsManager, AnalyticsManager>();
```
leeoades, line 258: Similar to my previous comment, unless you really need an instance of an object to be a singleton, and build it to handle multithreaded calls and handle the memory concerns of dependencies, then what you probably want here is transient.

leeoades (commented):
I did spot a few things actually. Have a think about them and let me know what you think.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/AnalyticsConsumer.cs
mmyoungman: Also made the change to Content's `AnalyticsConsumer`.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Strategies/AnalyticsWriteStrategyBase.cs
```diff
@@ -0,0 +1,30 @@
...
+                filePath,
+                contents: serialisedRequest);
+        }
+        catch (Exception e)
+        {
+            logger.LogError(
+                e,
+                "Error whilst writing {RequestTypeName} to disk",
+                requestTypeName);
+        }
```
benoutram, line 28: Not really one for this PR but I don't quite understand our approach of using files as a way for these client services to communicate with the analytics function to transmit analytics.

I would have thought that out of the possible options - file share, blob storage, table storage, queue service, etc, that a queue triggered function would have been more suitable for this. The analytics function can process events at its own rate, scaling out automatically if necessary. Automatic retry is built-in too. In the analytics function it looks like we move failed files that cannot be processed to a 'failed' directory on the fileshare without retrying.

I think you've mentioned there are already some concerns around performance in the writeup of #5822 such as the possibility of there being thousands of files when the function next wakes up, and scheduled runs overlapping. What happens if this writer is writing at the same time that the function is trying to read?

The file approach seems a bit prone to error with us having to handle a lot of things that could go wrong reading and writing files, locking, access permissions etc. This increases the complexity of the solution.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Strategies/AnalyticsWritePublicApiQueryStrategy.cs
```diff
@@ -0,0 +1,40 @@
+using GovUk.Education.ExploreEducationStatistics.Common.Utils;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Requests;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Services.Utils;
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Strategies.Interfaces;
+using Newtonsoft.Json;
```
benoutram, line 5: The Public API uses System.Text.Json rather than Newtonsoft.Json and I thought we want to migrate away from Newtonsoft.Json in favour of System.Text.Json in the long run? If it's not too late is it possible to change this to System.Text.Json?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Strategies/AnalyticsWritePublicApiQueryStrategy.cs
```diff
@@ -0,0 +1,40 @@
...
+            },
+            formatting: Formatting.Indented,
+            orderedProperties: true,
+            camelCase: true);
+
+        await this.WriteFileToShare(
+            requestTypeName: request.GetType().ToString(),
+            directory: analyticsPathResolver.PublicApiQueriesDirectoryPath(),
+            filename:
+            $"{DateTime.UtcNow:yyyyMMdd-HHmmss}_{queryRequest.DataSetVersionId}_{RandomUtils.RandomString()}.json",
```
benoutram, line 37: Injecting the .NET8 TimeProvider into this to get the Utc time now, rather than rely on `DateTime.UtcNow` would help unit test the format of the filename being generated here.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Strategies/AnalyticsWriteStrategyBase.cs
mmyoungman: I've made the change. I was concerned that a particular strategy might need bespoke serialisation, but I think you're right and it can be done inside `WriteToFileShare`.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
```diff
@@ -265,10 +267,22 @@ public void ConfigureServices(IServiceCollection services)
             {
                 services.AddSingleton<IAnalyticsPathResolver, AnalyticsPathResolver>();
             }
+
+            services.AddSingleton<AnalyticsWritePublicApiQueryStrategy>();
```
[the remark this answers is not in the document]
mmyoungman: Done. Made the same change to the `AnalyticsWritePublicZipDownloadStrategy`.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Strategies/AnalyticsWriteStrategyBase.cs
```diff
@@ -0,0 +1,30 @@
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Strategies;
+
+public class AnalyticsWriteStrategyBase(
```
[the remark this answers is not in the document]
mmyoungman: I've decided to inline `WriteToFileShare` back into `AnalyticsWritePublicZipDownloadStrategy`, and remove `AnalyticsWriteStrategyBase`.

My thought is that pulling this out makes it difficult to do another approach. For example, if we decided to save request files as binary files rather than as JSON - and that might be something we actually want to do. That would be tricky if we've already abstracted a "serialise to JSON" step and a "write to file" step, and we'd have to undo the abstraction. It seems conceivable that we could change `Report` in the future to not write to a file at all and send a message to a queue or something like that.

I now think that trying to create an abstraction here was premature as I don't think it's clear what we want here or what the final code will look like, so I'm inclined to write the code plainly for now and that leaves our options open for later. When we have more code and/or understanding of what we're doing here then I think the right abstraction will become more obvious and we can do it at that point.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Strategies/Interfaces/IAnalyticsWriteStrategy.cs
```diff
@@ -0,0 +1,8 @@
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Requests;
+
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Strategies.Interfaces;
+
+public interface IAnalyticsWriteStrategy
```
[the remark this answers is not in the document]
mmyoungman: I'm inclined to leave it. I think it makes sense that AnalyticsConsumer calls AnalyticsWriter.Report which in turn calls the correct AnalyticsWriteStrategy.Report. 

Although looking over, I am inclined to inline the AnalyticsWriter code into AnalyticsConsumer (i.e. inject strategies into AnalyticsConsumer instead), which allows us to remove AnalyticsWriter and then the AnalyticsWriteStrategy -> AnalyticsConsumerStrategy rename does make sense. But I'll leave it for now.

Still happy to make a change if you feel strongly about it.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/Interfaces/IAnalyticsManager.cs
mmyoungman: I'll leave this as it is - I don't see any confusion here by having a single interface - but happy to make a change if you feel strongly about it.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
mmyoungman: I figure that AnalyticsManager needs to be a singleton as we want one permanent channel for all analytics - without that we potentially have multiple channels or could something we add to a channel potentially get lost?

src/GovUk.Education.ExploreEducationStatistics.Content.Services/AnalyticsConsumer.cs
```diff
@@ -1,3 +1,4 @@
+#nullable enable
```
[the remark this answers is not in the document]
mmyoungman: Created EES-6115 to do this

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/AnalyticsCaptureRequests.cs
```diff
@@ -0,0 +1,14 @@
+namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Requests;
+
+public record AnalyticsCaptureRequestBase();
```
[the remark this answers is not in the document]
mmyoungman: Also noticed that in Content the file was called CaptureRequests.cs, so I've renamed that to match the PAPI filename

mmyoungman:
Something about my PR changes broke a load of tests and I'm not sure what. It seems that AnalyticsOptions.BasePath wasn't being loaded and failing in the AnalyticsPathResolver constructor. I fixed this by adding `   .ReplaceService(new Mock<IAnalyticsPathResolver>());` to `TestApplicationFactory`.

Highlighting this cause I'm not sure this is right.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/AnalyticsCaptureRequests.cs
```diff
@@ -0,0 +1,14 @@
...
+
+public record CaptureDataSetVersionQueryRequest(
+    Guid DataSetId,
+    Guid DataSetVersionId,
+    string DataSetVersion,
+    string DataSetTitle,
+    int ResultsCount,
+    int TotalRowsCount,
+    DateTime StartTime,
+    DateTime EndTime,
```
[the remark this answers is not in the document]
mmyoungman: I'm not sure - these will get serialised into JSON so I'm not sure there are any concerns there.

This wasn't introduced with my PR, and I don't think it's that big of a deal, so I'm going to leave this. Changing them might require a bit of coordination as I'm not sure if existing request files with `DateTime` would successfully get deserialised as a `DateTimeOffset`.

I can leave a comment on it or make a comment to switch these, although I'm not sure it's worthwhile?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/AnalyticsManager.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        return Task.CompletedTask;
+    }
+
+    public async ValueTask<AnalyticsCaptureRequestBase> Read(CancellationToken cancellationToken)
+    {
+        await Task.Delay(Timeout.Infinite, cancellationToken);
+        return default!;
+    }
+}
```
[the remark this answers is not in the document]
mmyoungman: Lee has highlighted a similar thing, and I've already agreed with Cam that I will be pulling common code out soon. I'm just taking it step by step. When I first did the Content analytics work I did things slightly differently to the Public API work, and I didn't want to pull out common code in that PR amongst all my other changes. This PR is to just align them. I will then introduce batching in processing request files, and then look to pull out common code.

I know this particular code is an obvious candidate to be pulled out - especially now I've aligned them -  and you could argue to just do it immediately, but there is other code where it isn't so obvious what can or cannot have some common package/util, and my plan was to just do all that in one PR.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/AnalyticsWriter.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    public async Task Report(AnalyticsCaptureRequestBase request, CancellationToken cancellationToken)
+    {
+        var success = strategyByRequestType.TryGetValue(request.GetType(), out var strategy);
+        if (!success)
+        {
+            throw new Exception($"No write strategy for request type {request.GetType()}");
+        }
+
+        await strategy!.Report(request, cancellationToken);
```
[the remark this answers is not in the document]
mmyoungman: See above comment.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Strategies/AnalyticsWritePublicApiQueryStrategy.cs
```diff
@@ -0,0 +1,40 @@
...
+public class AnalyticsWritePublicApiQueryStrategy(
+    IAnalyticsPathResolver analyticsPathResolver,
+    ILogger<AnalyticsWritePublicApiQueryStrategy> logger
+    ) : AnalyticsWriteStrategyBase(logger), IAnalyticsWriteStrategy
+{
+    public async Task Report(AnalyticsCaptureRequestBase request, CancellationToken cancellationToken)
+    {
+        var queryRequest = request as CaptureDataSetVersionQueryRequest
+                           ?? throw new ArgumentException(
+                               $"request isn't a {nameof(CaptureDataSetVersionQueryRequest)}");
```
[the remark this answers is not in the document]
mmyoungman: I've switched it here and in AnalyticsWritePublicZipDownloadStrategy.cs.

I don't really have much of a preference here - either seems fine to me.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Strategies/AnalyticsWritePublicApiQueryStrategy.cs
```diff
@@ -0,0 +1,40 @@
...
+            queryRequest.DataSetTitle);
+
+        var serialisedRequest = JsonSerializationUtils.Serialize(
+            obj: queryRequest with
+            {
+                Query = DataSetQueryNormalisationUtil.NormaliseQuery(queryRequest.Query)
+            },
+            formatting: Formatting.Indented,
+            orderedProperties: true,
+            camelCase: true);
```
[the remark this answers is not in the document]
mmyoungman: I will make a note to explore this when I look at pulling common code out. I think this may be okay because it doesn't really matter if we format the json differently. 

My general approach right now is to try and keep our options open so that future strategies have a lot of flexibility - they might not serialise a particular request to json at all and instead push it to a queue or whatever else.

Even so, it might still be worth to have a `serialiseAnalyticsRequest` util...

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Strategies/AnalyticsWriteStrategyBase.cs
mmyoungman: Happy to have a discussion about it. 

I think my preference would be to use a azure storage queue if we change approach, but that would mean reworking what we have significantly, which would take a fair bit of time and what we have does work.

We could consider new strategies using queues instead and then if that works move old strategies to queues at a later point - as I've said in other comments I've tried to make what abstractions we have flexible enough that we could use different approaches and don't have to solely stick to this fileshare approach.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Strategies/AnalyticsWritePublicApiQueryStrategy.cs
mmyoungman: I wasn't the one who did this originally. I think System.Text.Json might not have had the formatting options we wanted, but I'm not sure.

Not sure how much of a pain switching is. As this wasn't introduced by me and wasn't introduced in this PR, I am going to leave this. But I'll make a note to explore this later.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/AnalyticsCaptureRequests.cs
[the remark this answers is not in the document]
duncan-at-hiveit: I'm happy to pick this up in a separate PR @benoutram @mmyoungman when this work is in.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Strategies/AnalyticsWriteStrategyBase.cs
duncan-at-hiveit: Hey @benoutram @mmyoungman so this approach all stems from a research piece that we carried out a while ago.  The main reason that we're not processing requests with individual events and having them consumed individually is that we are wanting to *batch them up* and generate parquet files that capture a number of requests over a certain time.  The reason we want parquet, other than that is what Cam likes to work with, is that the Databricks team support parquet as a transfer file format into the Delta Lake.  So what we end up doing is quickly capturing requests and leaving them for a decent period of time, before we then generate a parquet file containing a big batch of them together.  The Databricks pipeline will then at some point consume these files from their file share using their File Loader pipeline type (possibly daily or some similar frequency) and transfer them to the Delta Lake. 

I guess that the important thing to note here is that we want to capture analytics information at the point that it happens, but we don't have to rush to consume it as we are happy to let them batch up for a decent time before reporting on them, so an hourly function that checks to see what needs reporting on is all that's needed rather than anything that requires immediate and individual attention.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/AnalyticsManager.cs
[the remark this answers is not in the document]
benoutram: We agreed today that this abstracting of common code will be coming soon! 🙂  cc: @mmyoungman  @duncan-at-hiveit

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Strategies/AnalyticsWritePublicApiQueryStrategy.cs
benoutram: @duncan-at-hiveit not sure if you have an opinion on this too? Do you think it is possible for us to pick this up sooner rather than later, if we are trying and standardise on one JSON serializer across the service, and preferring `System.Text.Json` over `Newtonsoft.Json`?

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/Interfaces/IAnalyticsManager.cs
leeoades: If I had felt strongly about it, I would have talked to Duncan more about it :-).
But it is a good thing to bear in mind going forward - that different consumers of what may be a common class can have different requirements and that the implementation doesn't need to dictate the interface.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Strategies/AnalyticsWriteStrategyBase.cs
mmyoungman: Discussed this with Dunc on Slack. The main problem with using Azure Storage Queue is that the max message size might not be large enough for what we need, if there are requests with thousands of schools, for example.

I'm not going to arrange a meeting for this unless other poke me to do so.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Startup.cs
leeoades: My bad. This one does, indeed, need to be a singleton.

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Fixtures/TestApplicationFactory.cs
```diff
@@ -1,33 +1,61 @@
...
 {
+    private readonly HashSet<string> _additionalAppsettingsFiles = new();
+
     public async Task ClearAllTestData()
     {
         await EnsureDatabaseDeleted<ContentDbContext>();
         await EnsureDatabaseDeleted<StatisticsDbContext>();
     }
 
+    public TestApplicationFactory AddAppSettings(string filename)
```
leeoades, line 29: I couldn't see where any of this new AppSettings file stuff is used?

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/StrategiesTests/AnalyticsWritePublicZipDownloadStrategyTests.cs
```diff
@@ -72,10 +73,12 @@ await strategy.Report(new CaptureZipDownloadRequest(
 
     private AnalyticsWritePublicZipDownloadStrategy BuildStrategy(
         IAnalyticsPathResolver pathResolver,
+        DateTimeProvider? dateTimeProvider = null,
```
leeoades, line 76: I had a look at how the date is being used to create the filename. Is this important? ie if I modified the code so that it didn't have the date and time as the prefix, would that matter? If it does matter, then I would add a unit test to protect this - and so you'll need a mock DateTimeProvider.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/AnalyticsCaptureRequests.cs
```diff
@@ -1,6 +1,6 @@
 namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Requests;
 
-public record AnalyticsCaptureRequestBase();
+public abstract record AnalyticsCaptureRequestBase;
```
leeoades, line 3: I've been having a think about this and a look through the code. I believe that `AnalyticsCaptureRequestBase` is simply acting as a marker for a request now, right? I don't think it serves any purpose? In which case, I would recommend that you change it to an interface: `IAnalyticsCaptureRequest`. Why you ask? Because it is possible to use a marker interface against anything we wish to use as a request. However, if we insist that all request must derive from this base class, it could prevent us from using another more logical class in future - which the interface would not.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Strategies/AnalyticsWriteStrategyBase.cs
benoutram: Ok @duncan-at-hiveit thank you! I did already understand the gist of the _batch_ and _parquet file_ requirements for the destination output, but thanks for the detailed explanation.

I didn't explain this in my comment but I was thinking something along the lines of two functions:

One to consume individual analytic events from a queue and write them to some kind of database temporarily - Azure Table Storage was the first thing that came to mind because it's scalable and low cost but there's probably alternatives. The total cost per function invocation should be tiny/insignificant.

Then another _batch_ function, which is the timer triggered function we've already got which does the export to parquet, except instead of working with files it would query _x_ number of rows or query by date window to skim off a batch of events from the database and export as parquet.

From the producer's point of view I'm not sure how serializing data to the queue endpoint would fair vs. writing to the file share, but I think the fileshare is mounted over the network so both involve network requests and might be fairly similar.

This is an alternative approach that avoids file handling and two services potentially accessing files at the same time.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/AnalyticsCaptureRequests.cs
benoutram: Ah yes it was me who suggested making this `abstract` but I agree if it is only being used as a marker interface then an interface would make more sense.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/StrategiesTests/AnalyticsWritePublicZipDownloadStrategyTests.cs
leeoades: The reason I asked the question was in case the filename was used later in order to group up files or process them in order or whatever.

leeoades:
This looks good to me.
@benoutram  - once you're happy and have resolved any conversations etc then please approve.

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/StrategiesTests/AnalyticsWritePublicZipDownloadStrategyTests.cs
benoutram: It's worth noting that I've left a separate comment suggesting that if different strategies should be writing files with consistent file naming (which seems fairly sensible even if these filenames don't impact the end deliverable), that a common method should be pulled out to provide the name.

We can already rely on DateTimeProvider doing what it says it does - It would be our makeup of the filename that would be tested here.

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Services.Tests/StrategiesTests/AnalyticsWritePublicZipDownloadStrategyTests.cs
mmyoungman: I have a note recorded for that to look into pulling out the filename creation into a common util/method or whatever.

For me, even if we were relying on the filename, I don't think there the creation of the filename is doing anything particularly complex or that could fail, and so don't think it warrants a test - but I am more relaxed than most others about these things.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Strategies/AnalyticsWriteStrategyBase.cs
benoutram: Ok thanks @mmyoungman, I hadn't realised the 64Kb limit of storage queue messages. There might be alternatives like service bus queues.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Strategies/AnalyticsWritePublicApiQueryStrategy.cs
duncan-at-hiveit: Hey @benoutram so if you remember, a while back I did quiz you about this and you confirmed the general approach to move to System.Text where possible, so I was considering this back then as well.  The reason I stuck with Newtonsoft here rather than switch immediately was due to the fact that I needed to introduce some IContractResolvers (OrderedFieldsCamelCaseContractResolver / OrderedFieldsContractResolver / etc) to get more fine-grained control over the JSON serialisation (in particular about keeping the order of JSON fields consistent in the Public API queries as they're serialised, for consistent hashing).

I'm sure this is possible to do with System.Text somehow too, so I'll look to see what the equivalent support looks like there.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Strategies/AnalyticsWriteStrategyBase.cs
mmyoungman: I did see service bus queues - or maybe we could use blob storage only when required for larger requests.

I also saw that Azure Storage Queues allow you to pull messages off them in batches of 32, which could be an advantage over files.

Anyway, all that said, while I still like the idea of a queue - primarily because it requires us to write less bespoke code that may have issues - since our current method is already written and tested and working, I'd be inclined to leave it as is. If we hit an issue later then I don't think switching is a huge pain - switching now would potentially paying a cost to avoid a problem we'll never have.

PR #5849, opened 2025-05-06
EES-6106 Raise an event in Admin when archived publications are restored

benoutram:
This PR is a follow up to https://github.com/dfe-analytical-services/explore-education-statistics/pull/5848 and raises an event in Admin when archived publications are restored.

The Admin service allows updating publications including setting and un-setting a 'superseded by' publication for a publication. A publication is considered to be archived if it has a superseding publication which is live, i.e. the superseding publication has one or more published releases.

This PR changes the Admin service to raise an Event Grid event when archived publications are restored. The event is only raised if a publication was previously archived but now isn't.

To know whether to raise the event or not, we work out the transition of the publication's archived status and only raise it if the status is changing from archived to not archived.

The Search Function App is subscribed to these events and will manage searchable documents accordingly. This change was made to the Search Function App in https://github.com/dfe-analytical-services/explore-education-statistics/pull/5846.


leeoades (changes requested):
A small point or two.

leeoades (approved):
Excellent stuff.

PR #5852, opened 2025-05-08
EES-6070 Remove Searchable Document Function

leeoades:
We introduced a new Azure function that removes specified searchable documents in preparation for removing searchable documents when there has been a reorder.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Functions/RemoveSearchableDocument/RemoveSearchableDocumentTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Functions/RemoveSearchableDocument/RemoveSearchableDocumentTests.cs
@@ -1,0 +1,41 @@
+﻿using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Functions.RemoveSearchableDocument;
+using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Functions.RemoveSearchableDocument.Dto;
+using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests.Builders;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests.Functions.RemoveSearchableDocument;
+
+public class RemoveSearchableDocumentTests
+{
+    private readonly SearchableDocumentRemoverMockBuilder _searchableDocumentRemoverMockBuilder = new();
+    private RemoveSearchableDocumentFunction GetSut() => new(
+        new NullLogger<RemoveSearchableDocumentFunction>(),
+        _searchableDocumentRemoverMockBuilder.Build());
+
+    [Fact]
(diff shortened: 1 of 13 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/RemoveSearchableDocument/RemoveSearchableDocumentFunction.cs
```diff
@@ -0,0 +1,30 @@
...
+    ILogger<RemoveSearchableDocumentFunction> logger,
+    ISearchableDocumentRemover searchableDocumentRemover)
+{
+    [Function(nameof(RemoveSearchableDocument))]
+    public async Task RemoveSearchableDocument(
+        [QueueTrigger("%RemoveSearchableDocumentQueueName%")] RemoveSearchableDocumentDto message,
+        FunctionContext context)
+    {
+        var releaseId = message.ReleaseId;
+        if (releaseId is null || releaseId == Guid.Empty)
```
benoutram, line 18: I don't think you should mix `is` with `||` and `==` here and that it would be better to use `releaseId == null` to be consistent.

Anyway, I think this could also be replaced by

```csharp
if (!releaseId.HasValue)
{
    return;
}
```

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Functions/RemoveSearchableDocument/RemoveSearchableDocumentTests.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        var command = new RemoveSearchableDocumentDto { ReleaseId = Guid.NewGuid() };
+
+        await GetSut().RemoveSearchableDocument(command, new FunctionContextMockBuilder().Build());
+
+        _searchableDocumentRemoverMockBuilder.Assert
+            .RemoveSearchableDocumentCalledFor(command.ReleaseId.Value);
+    }
+
+    public static TheoryData<Guid?> MissingGuids = [Guid.Empty, null];
```
benoutram, line 29: Looks like this would be pretty handy to be shared by some other tests too like `GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests.Functions.RemovePublicationSearchableDocuments.RemovePublicationSearchableDocumentsTests.WhenMessageDoesNotContainSlug_ThenDocumentRemoverNotCalled`

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/RemoveSearchableDocument/RemoveSearchableDocumentFunction.cs
leeoades: Interesting that you are worried about combining pattern matching and the equality check. I'm assuming that's simply a stylistic / aesthetic thought? 

Note that using "is null" and "is not null" is considered better practice.

Also, Guid.Empty has a value so fails the suggested HasValue check.

PR #5854, opened 2025-05-09
EES-6124 Abstract EventGrid and Event types from Common, Admin and Publisher to new Events project

benoutram:
In order to make the `PublicationArchivedEvent` reusable across both Admin and Publisher services for [EES-6107](https://dfedigital.atlassian.net/browse/EES-6107), this PR creates a new Events package that provides all of the Event types.

While doing this it also made sense to abstract all of the Event Grid types from Common so that projects using Common but not using Events don’t have a dependency on Event Grid.


leeoades (approved):
Looks spot on to me.

PR #5856, opened 2025-05-09
EES-6123 Remove `theme` and add `releaseId`, `publicationId` and `themeId` in Searchable documents and Search index configuration

benoutram:
This PR removes `theme` and adds `releaseId`, `publicationId` and `themeId` as metadata keys in searchable document blobs and in fields in the Azure AI Search index configuration.

- `releaseid` already existed but only in the searchable documents storage container as the blob name which was also mapped to  field `metadata_storage_name` in the search index. It's being added as a blob metadata attribute and search index field explicitly.
- `themeId` is replacing `theme` to support the frontend development in [EES-6001](https://dfedigital.atlassian.net/browse/EES-6001). Using title meant that it ended up being a query parameter value in search requests. This wasn’t ideal because the title isn’t guaranteed to be unique, there’s possible encoding issues, it’s not as efficient as using an id, and the title could change in-between initial presentation of results and an updated query filter being submitted.
- `publicationId` is being added along with the other id's for completeness.

### Other changes

- Change `ThemeViewModel` to not use a primary constructor to make the properties being set more explicit, improving readability.
- Set the `releaseVersionId` field to be filterable in the Azure AI Search index configuration to be consistent with `publicationId` and `themeId`.


leeoades (approved):
This all looks good to me.

PR #5859, opened 2025-05-12
Ees 6084 extract common analytics processor workflow

duncan-at-hiveit:
## Overview

This PR:
* Takes the 3 Analytics IRequestFileProcessor implementations, and extracts a common workflow from them.
* Supports scenario-specific actions by abstracting out an "actor" that can be taken through the workflow and used to perform scenario-specific steps at the right time.

## New common workflow

The workflow common to all of the 3 implementations is now represented by a class and takes a provided "actor" through the workflow within its `Process()` method.

Note that it isn't necessary for future IRequestFileProcessor implementations to use this workflow if for whatever reason they need to do something more specialised - it's only that our current 3 can all make use of it.

### PathResolver simplifications

Now that the workflow is made common between all processors, it didn't seem necessary to keep all of the PathResolver folders, as these can be worked out by convention, so I've removed a lot of them.  E.g. all "processing" and "failures" folders are always created underneath a scenario's main folder for finding source files, so we just use that convention rather than having to declare the locations each time.

The only downside is that the tests now need to know where to find the "processing" folder!

### DuckDB calls are now async

I swapped all of our processors' various `ExecuteNonQuery` for the `ExecuteNonQueryAsync` equivalents.  The methods were already using Task as the return type, so this was straightforward.

### Test the common workflow in isolation

Now that the workflow has been broken out, we don't *have to* keep testing common patterns of behaviour in our Processor tests  e.g. "no source files available".  Instead, the workflow could be tested with a mock IWorkflowActor against various scenarios.  We would also have to restructure things a bit to test that the Processors were actually using the workflows correctly too!  At the moment we know they are, simply by our tests still working without needing any changes.


src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Services/Workflow/ProcessRequestFilesWorkflow.cs
```diff
@@ -0,0 +1,177 @@
...
+            .GetFiles(_processingDirectory)
+            .Select(Path.GetFileName)
+            .OfType<string>()
+            .ToList();
+
+        foreach (var filename in filesReadyForProcessing)
+        {
+            try
+            {
+                await actor.ProcessSourceFile(
```
mmyoungman, line 102: I have been intending to switch this to process files in batches - can't remember if I discussed that with you or someone else. Doesn't seem to me like this PR prevents that change later - in fact I think it probably makes it easier - but raising this in case you see something I don't.

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Services/Workflow/ProcessRequestFilesWorkflow.cs
```diff
@@ -0,0 +1,177 @@
...
+/// <param name="sourceDirectory">
+///     Folder in which to find the source files to read in this workflow.
+/// </param>
+/// <param name="reportsDirectory">
+///     Folder in which to generate the Parquet report files in this workflow.
+/// </param>
+/// <param name="actor">
+///     The <see cref="IWorkflowActor" /> implementation being guided through the workflow.
+/// </param>
+public class ProcessRequestFilesWorkflow(
```
mmyoungman, line 37: Before seeing this change, I was thinking that maybe we should not tightly coupled together with the fileshare / consume azure function approach we have - in case want a different approach for some reason. But with reflection, I think it wouldn't be hard to create a new type of workflow if that's we wanted so no reason to worry about that. And doesn't seem too likely that would happen anyway unless we hit a big unforeseen issue.

mmyoungman (approved):

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Services/Workflow/ProcessRequestFilesWorkflow.cs
duncan-at-hiveit: Yup it was me who had this discussion with you, and I agree that it'll be easy enough to fit in in the future here!

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer/Services/Workflow/ProcessRequestFilesWorkflow.cs
duncan-at-hiveit: Yeah exactly.   That's why I didn't tie this behaviour into the IFileRequestProcessors themselves but rather let all of our existing ones "opt in" to using it.  I'm sure as well that this workflow could be more flexible in the future too if we need it to be!

PR #5863, opened 2025-05-13
EES-6121 Full Searchable Documents Reset

leeoades:
There have been several situations where we have altered the index which has meant that the searchable documents needed to be rebuilt. This has been performed manually.

Whilst expected to be a rare occurrence, especially once the new search is live, it is still useful to have a programmatic way to perform this reset.

This PR contains a new Azure function that for the time being will be invoked from the Azure portal. It removes all searchable documents and then queues all publications to have their searchable document refreshed.

It is relying on the scheduled indexer to rebuild the index.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Builders/ContentApiClientMockBuilder.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Builders/ContentApiClientMockBuilder.cs
@@ -11,6 +11,7 @@ internal class ContentApiClientMockBuilder
     private ReleaseSearchableDocument? _releaseSearchableDocument;
     private PublicationInfo[]? _publicationsForTheme;
     private ReleaseInfo[]? _releasesForPublication;
+    private PublicationInfo[]? _eesPublications;
 
     public ContentApiClientMockBuilder()
     {
@@ -33,6 +34,11 @@ internal class ContentApiClientMockBuilder
                 It.IsAny<string>(),
                 It.IsAny<CancellationToken>()))
             .ReturnsAsync(() => _releasesForPublication ?? []);
+
+        _mock
+            .Setup(m => m.GetAllLivePublicationInfos(
+                It.IsAny<CancellationToken>()))
+            .ReturnsAsync(() => _eesPublications ?? []);
     }
 
     public IContentApiClient Build()
@@ -63,6 +69,12 @@ internal class ContentApiClientMockBuilder
     public ContentApiClientMockBuilder WherePublicationHasReleases(params ReleaseInfo[] releases)
     {
         _releasesForPublication = releases;
+        return this;
+    }
+
+    public ContentApiClientMockBuilder WhereEESHasPublications(params PublicationInfo[] publications)
+    {
+        _eesPublications = publications;
         return this;
     }
 
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Builders/ContentApiClientMockBuilder.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Builders/ContentApiClientMockBuilder.cs
@@ -11,7 +11,7 @@ internal class ContentApiClientMockBuilder
     private ReleaseSearchableDocument? _releaseSearchableDocument;
     private PublicationInfo[]? _publicationsForTheme;
     private ReleaseInfo[]? _releasesForPublication;
-    private PublicationInfo[]? _eesPublications;
+    private PublicationInfo[]? _publications;
 
     public ContentApiClientMockBuilder()
     {
@@ -38,7 +38,7 @@ internal class ContentApiClientMockBuilder
         _mock
             .Setup(m => m.GetAllLivePublicationInfos(
                 It.IsAny<CancellationToken>()))
-            .ReturnsAsync(() => _eesPublications ?? []);
+            .ReturnsAsync(() => _publications ?? []);
     }
 
     public IContentApiClient Build()
@@ -72,9 +72,9 @@ internal class ContentApiClientMockBuilder
         return this;
     }
 
-    public ContentApiClientMockBuilder WhereEESHasPublications(params PublicationInfo[] publications)
+    public ContentApiClientMockBuilder WhereHasPublications(params PublicationInfo[] publications)
     {
-        _eesPublications = publications;
+        _publications = publications;
         return this;
     }
 
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Services/ResetSearchableDocuments/FullSearchableDocumentResetterTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Services/ResetSearchableDocuments/FullSearchableDocumentResetterTests.cs
@@ -20,7 +20,7 @@ public class FullSearchableDocumentResetterTests
     public void Can_instantiate_SUT() => Assert.NotNull(GetSut());
 
     [Fact]
-    public async Task WhenPerformFullReset_ThenAllSearchableDocumentsAreRemoved()
+    public async Task WhenPerformReset_ThenAllSearchableDocumentsAreRemoved()
     {
         // Arrange
         var sut = GetSut();
@@ -47,7 +47,7 @@ public class FullSearchableDocumentResetterTests
             })
             .ToArray(); 
 
-        _contentApi.WhereEESHasPublications(publications);
+        _contentApi.WhereHasPublications(publications);
             
         var sut = GetSut();
         
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Services/ResetSearchableDocuments/FullSearchableDocumentResetterTests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Services/ResetSearchableDocuments/FullSearchableDocumentResetterTests.cs
@@ -1,0 +1,6 @@
+﻿namespace GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests.Services.ResetSearchableDocuments;
+
+public class FullSearchableDocumentResetterTests
+{
+    
+}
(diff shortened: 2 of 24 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Builders/ContentApiClientMockBuilder.cs
```diff
@@ -11,6 +11,7 @@ internal class ContentApiClientMockBuilder
     private ReleaseSearchableDocument? _releaseSearchableDocument;
     private PublicationInfo[]? _publicationsForTheme;
     private ReleaseInfo[]? _releasesForPublication;
+    private PublicationInfo[]? _eesPublications;
```
benoutram, line 14: Just wondering if we can think of a name that doesn't include 'ees' as that looks a bit strange to me - the other local `PublicationInfo` properties here can be considered to be 'ees' publications too.

Something like one of the following?

`_publications` / `WhereHasPublications`
`_allPublications` / `WhereHasAllPublications`
`_livePublications` / `WhereHasLivePublications`

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Services/ResetSearchableDocuments/FullSearchableDocumentResetterTests.cs
```diff
@@ -0,0 +1,63 @@
...
+        new FullSearchableDocumentResetter(
+            _contentApi.Build(),
+            _searchableDocumentRemover.Build(),
+            new NullLogger<FullSearchableDocumentResetter>());
+
+    [Fact]
+    public void Can_instantiate_SUT() => Assert.NotNull(GetSut());
+
+    [Fact]
+    public async Task WhenPerformFullReset_ThenAllSearchableDocumentsAreRemoved()
```
benoutram, line 23: This test method and the one below could be named `WhenPerformReset_`... because the method name  is `PerformReset`.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/CommandHandlers/FullReset/FullResetFunction.cs
```diff
@@ -0,0 +1,35 @@
...
+using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Services.ResetSearchableDocuments;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Azure.Functions.Worker;
+
+namespace GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Functions.CommandHandlers.FullReset;
+
+public class FullResetFunction(IFullSearchableDocumentResetter fullSearchableDocumentResetter)
+{
+    [Function(nameof(FullSearchableDocumentsReset))]
+    public async Task<FullSearchableDocumentsResetOutput> FullSearchableDocumentsReset(
```
benoutram, line 12: I can't see any test class for this function. I know it would be pretty trivial but I think we've got unit tests for all the other functions.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/CommandHandlers/FullReset/FullResetFunction.cs
```diff
@@ -0,0 +1,35 @@
...
+        [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = nameof(FullSearchableDocumentsReset))]
```
benoutram, line 13: With this set to level `Anonymous` it's not secured by any of the function app's own access keys. That's fine - we don't really want to secure it using this method as we don't want to be sharing any access keys and there's no proof of identity by using access keys.

However, `Anonymous` does mean security of this method is then left to the function app itself which can authenticate/authorise different clients, but the Search function app has authentication disabled.

That leaves it down to network security of the function app, which is set to be publicly accessible from a limited range of IP addresses. This means any client in that IP address range could call this anonymously which is not ideal.

I think that enabling built in Entra ID authentication on the function app and locking this down so that only authorised applications like the Admin service can use this is the way to go. We could add a button to the Admin's Platform Administrative area allowing this to be called.

If you agree, we should add a ticket on this. It will require infrastructure changes to configure the authentication so it's more than just a change to the Admin service. The underlying bicep template for our generic function app component doesn't have support for this yet.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Clients/AzureBlobStorage/IAzureBlobStorageClient.cs
```diff
@@ -29,4 +29,9 @@ public interface IAzureBlobStorageClient
     /// <param name="cancellationToken"></param>
     /// <returns></returns>
     Task<bool> ContainerExists(string containerName, CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Delete all blobs from the specified container
+    /// </summary>
```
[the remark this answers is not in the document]
leeoades: I've added them for completeness, although I'm not sure they add a lot of value.  👍

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/CommandHandlers/FullReset/FullResetFunction.cs
```diff
@@ -0,0 +1,35 @@
...
+        
+        return new FullSearchableDocumentsResetOutput
+        {
+            RefreshSearchableDocuments = response.AllPublications
+                .Select(publication => new RefreshSearchableDocumentMessageDto
+                                                    {
+                                                        PublicationSlug = publication.PublicationSlug
+                                                    })
+                .ToArray()
+        };
```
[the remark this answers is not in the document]
leeoades: I try to keep the brackets connected to the object type these are initialising - so hopefully this is acceptable?
```
return new FullSearchableDocumentsResetOutput
        {
            RefreshSearchableDocuments = response.AllPublications
                .Select(publication =>
                            new RefreshSearchableDocumentMessageDto
                            {
                                PublicationSlug = publication.PublicationSlug
                            })
                .ToArray()
        };
```

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/CommandHandlers/FullReset/FullResetFunction.cs
```diff
@@ -0,0 +1,35 @@
...
+    }
+}
+
+public class FullSearchableDocumentsResetOutput
+{
+    [QueueOutput("%RefreshSearchableDocumentQueueName%")]
+    public RefreshSearchableDocumentMessageDto[] RefreshSearchableDocuments { get; init; }
+}
```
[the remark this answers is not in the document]
leeoades: Indeed. I started with it like this and then I wasn't sure whether we'd want to return anything else. Since I keep preaching yagni, I've modified it to match what you suggested.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Clients/AzureBlobStorage/AzureBlobStorageClient.cs
```diff
@@ -59,4 +59,19 @@ public async Task<bool> ContainerExists(string containerName, CancellationToken
...
+        await foreach (var blob in asyncPageable)
+        {
+            blobNames.Add(blob.Name);
+        }
+
+        await Task.WhenAll(
+            blobNames.Select(blobName => 
+                    blobContainerClient.DeleteBlobIfExistsAsync(blobName, cancellationToken: cancellationToken)));
+    }
 }
```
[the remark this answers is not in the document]
leeoades: Great detective work, Mr Holmes. OK, I created 10,000 blobs in our integration test container in the Hive account. The loop successfully retrieved all 10,000. 
For our information, I changed it to retrieve pages of 10 and used .AsPages. It never left the loop by its own accord - so basically, if you break out of the asyn foreach loop manually, keep hold of the continuation token, then you can come back in and continue where you left off. Also, by specifying a page size of 10, predicatably, it kept going back to Azure to retrieve the next page. So the code above is actually much faster than using pages.

Once I was satisfied, I let the code carry on and delete the blobs. It took a while to delete 10,000 - although probably less than a minute. I didn't think to time it.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Clients/ContentApi/ContentApiClient.cs
```diff
@@ -122,6 +122,18 @@ public async Task<ReleaseInfo[]> GetReleasesForPublication(
...
 
+    public async Task<PublicationInfo[]> GetAllLivePublicationInfos(CancellationToken cancellationToken)
+    {
+        var publicationDtos =
+            await GetAllPaginatedItems<PublicationDto>(page => BuildGetPublicationsPageEndpoint(page, numberOfItems:30),
+                cancellationToken);
+        
+        return publicationDtos
+            .Where(dto => !string.IsNullOrEmpty(dto.Slug))
+            .Select(dto => new PublicationInfo { PublicationSlug = dto.Slug! })
```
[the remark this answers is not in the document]
leeoades: I would imagine it is because it is a deserialised dto and we can not guarantee anything about the shape of that object going forward. If the slug property is renamed to PublicationSlug for example.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Tests/Builders/ContentApiClientMockBuilder.cs
leeoades: Yeah, I didn't like that name either :-)

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/CommandHandlers/FullReset/FullResetFunction.cs
leeoades: OK, I suggest therefore we raise a ticket to combine all of these into a single piece of work - and we leave it as anonymous in my PR. In the meantime, it seems unlikely that anyone would attempt to call the function or even know how to.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Clients/AzureBlobStorage/AzureBlobStorageClient.cs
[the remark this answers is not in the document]
benoutram: Ok that's good to know. Brilliant work on the investigation! 💯

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/CommandHandlers/FullReset/FullResetFunction.cs
benoutram: I've raised EES-6151 for this.

benoutram (approved):

PR #5864, opened 2025-05-13
EES-6132 azure search frontend general

Guy-HiveIT:
# Summary

This PR adds an alternative search to the find stats page, accessible by adding the query param `?azsearch=true` to the URL.

The new search uses Azure and has already been configured.
- It has a different layout to the existing page reflecting a new design.
- It shows results with highlighted content snippets when a search term is entered.
- It should sort results based on the user's choice in the 'order by' select.
- It should allow filtering results based on release type and/or theme.
  - Additionally, it should show the number of results next to each filter option in the select, and sort the options from most results to least.
- It should maintain the same pagination as existing search.
- It should show currently applied filters as little 'pills' that can clear that filter, as per existing search, albeit with updated designs.

## Code considerations

To keep the existing functionality and design alongside the azure variant, I've created new modules/components/files as necessary, adding `Azure` somewhere in the name to make it clear.
When this work goes live, we can remove the non-azure versions (hopefully easily) to clean up.

We still need to fetch the list of `theme`s from the API as we currently do: Azure returns facets only of the `themeId` - so we need the full list of themes from the API to be able to match facet result IDs to titles. 

## IMPORTANT

**The Azure search client needs credentials which can't be tracked in git, so there are new entries in the frontend `.env` which will need entering for this to work, locally and on dev.**
Locally, you can create a `.env.local` with the required values - this file is gitignored but will be read when running the app locally. You can also point the frontend content api env values to the dev environment content API to get the same list of themes from dev as are returned from Azure (we are currently working from the search index of dev environment content)

So in a `.env.local` in the frontend project, add in values to point to dev:
```
CONTENT_API_BASE_URL=
DATA_API_BASE_URL=
AZURE_SEARCH_QUERY_KEY=
AZURE_SEARCH_ENDPOINT=
AZURE_SEARCH_INDEX=
```

## Tests

All tests should be passing currently, as they are not opting in to the new feature flagged functionality. I have added in an empty mock for `@azure/search-documents` to get the existing unit tests working.

Tests still need to be written for the new azure search functionality which will be a separate PR. 

### Demo video
Viewable in slack in the improved-search channel


src/explore-education-statistics-frontend/src/modules/find-statistics/FindStatisticsPageAzure.tsx
```diff
@@ -0,0 +1,403 @@
...
+            Search and browse statistical summaries and download associated data
+            to help you understand and analyse our range of statistics.
+          </p>
+          <SearchForm
+            label="Search publications"
+            value={search}
+            onSubmit={nextValue =>
+              handleChangeFilter({ filterType: 'search', nextValue })
+            }
+          />
```
bennettstuart, line 235: The validation feels quite aggressive on this, I think the clear search button will help but we might want to alter the validation settings to onSubmit instead of onChange to avoid UI jumping around as you type

bennettstuart (changes requested):

src/explore-education-statistics-frontend/src/modules/find-statistics/utils/createAzurePublicationListRequest.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+  if (sortBy === 'oldest') {
+    return {
+      sort: 'published asc',
+    };
+  }
+  return {
+    sort: 'published desc',
+  };
+}
```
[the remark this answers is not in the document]
Guy-HiveIT: Thanks, good spot, much cleaner

bennettstuart (approved):
LGTM!

src/explore-education-statistics-frontend/src/services/azurePublicationService.ts
```diff
@@ -0,0 +1,139 @@
...
+  search?: string;
+  orderBy?: AzurePublicationOrderByParam;
+  sortDirection?: SortDirection;
+}
+
+const { AZURE_SEARCH_ENDPOINT, AZURE_SEARCH_INDEX, AZURE_SEARCH_QUERY_KEY } =
+  env;
+
+const azureSearchClient = new SearchClient<AzurePublicationSearchResult>(
+  AZURE_SEARCH_ENDPOINT || '',
```
benoutram, line 57: Looks like we also need to add this endpoint URL to `cspConnectSrc` in `server.js` to avoid the browser blocking the request because the URL is not in the site's content security policy.

benoutram (approved):

PR #5865, opened 2025-05-13
EES-6119 - (STEP 1) Migrate all Release `Lead` Roles to Release `Contributor` Roles and remove the `Lead` role

jack-hive:


src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20250513134323_EES6119-MigrateReleaseLeadRolesToContributors.cs
```diff
@@ -0,0 +1,51 @@
...
+                      FROM UserReleaseRoles urr2
+                      WHERE urr2.UserId = urr.UserId
+                        AND urr2.ReleaseVersionId = urr.ReleaseVersionId
+                        AND urr2.Role = 'Contributor'
+                  );
+            ");
+
+            // Delete all `Lead` roles where the same user DOES already have a `Contributor` role assigned for the same
+            // release version
+            migrationBuilder.Sql(@"
```
duncan-at-hiveit, line 31: Can't you just simply delete ALL "Lead" roles at this point?  I think that after this, there is no need for them anymore.

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/AuthorizationHandlerService.cs
```diff
@@ -18,16 +18,14 @@ public class AuthorizationHandlerService
 {
     private static readonly ReleaseRole[] ReleaseEditorRoles =
```
duncan-at-hiveit, line 19: Maybe we should be making this public and using it around the codebase...

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20250513134323_EES6119-MigrateReleaseLeadRolesToContributors.cs
jack-hive: Yeah you're right. Running these two queries should, in theory, result in no `Lead` roles existing afterwards. Because the only ones left, after running the 1st query, should be the ones that satisfy the 2nd query.

So you could change the 2nd query to just delete all the remaining ones, with no condition. In theory, it would result in the same outcome but the query would be simpler!

I don't mind either way, but 1 thought I have, is that I wonder if the current queries are slightly safer? I.e. if any `Leads` **are** remaining after this migration has ran, then something has gone wrong or the query logic is wrong. Whereas, if I change the 2nd query to do a blanket delete, it may mask a mistake I've made?

I think the chances I've made a mistake here are pretty slim, but just a thought! What do you reckon?

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandlerTests.cs
```diff
@@ -125,7 +125,7 @@ public class ReleaseRoleTests
             public async Task EditorsOrApproversOnAnyOwningPublicationReleaseCanUpdateMethodology()
             {
                 var expectedReleaseRolesToPass =
-                    ListOf(ReleaseRole.Approver, ReleaseRole.Contributor, ReleaseRole.Lead);
+                    ListOf(ReleaseRole.Approver, ReleaseRole.Contributor);
```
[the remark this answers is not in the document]
jack-hive: We do actually have it on `AuthorizationHandlerService.ReleaseEditorAndApproverRoles` (`Approver` is technically not classed as an 'Editor' in our nomenclature for some reason), so we could use that!

Although, I'm wondering if we leave it since this will be getting changed later, once the new publication roles come into play, and the old ones go out of the window?

This code will have to be touched at that point, at which point the roles will be simpler as there will only be 2. And it probably then makes sense to distinguish between them, and there probably won't be an equivalent `public static` field stored for the pair in the `AuthorizationHandlerService`

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/AuthorizationHandlerService.cs
jack-hive: Same as my point above, I wonder if it might be wasted effort?

Happy for you to disagree though!

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Security/AuthorizationHandlers/UpdateSpecificMethodologyAuthorizationHandlerTests.cs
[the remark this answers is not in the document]
duncan-at-hiveit: :+1:

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/UserReleaseRoleRepositoryTests.cs
```diff
@@ -463,7 +456,7 @@ await contentDbContext.AddRangeAsync(userReleaseRole1, userReleaseRole2, userRel
             {
                 var service = SetupUserReleaseRoleRepository(contentDbContext);
                 await service.RemoveMany(
-                    ListOf(userReleaseRole1, userReleaseRole2, userReleaseRole3),
+                    ListOf(userReleaseRole1, userReleaseRole2),
```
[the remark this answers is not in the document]
duncan-at-hiveit: Nice one!

src/GovUk.Education.ExploreEducationStatistics.Admin/Security/AuthorizationHandlers/AuthorizationHandlerService.cs
duncan-at-hiveit: Nope that's fine ta - I was thinking more long-term anyhoo so that's cool!

src/GovUk.Education.ExploreEducationStatistics.Admin/Migrations/ContentMigrations/20250513134323_EES6119-MigrateReleaseLeadRolesToContributors.cs
duncan-at-hiveit: I'm fine either way - should be the same outcome.

PR #5867, opened 2025-05-14
EES-6107 Refactor `PublishingCompletionService` to improve handling when publications are archived

benoutram:
This PR refactors the Publisher's `PublishingCompletionService.CompletePublishingIfAllPriorStagesComplete` method to improve handling when publications are archived.

It adds a new `HandleArchivedPublications` method which is called with a list of publications that have been affected by publishing (i.e. they are related to releases just published).

Within `HandleArchivedPublications`, it only performs actions for publications which it finds are newly archived. These are publications which are replaced by any of the publications affected by publishing that were not already published (i.e. they had no published releases prior to publishing). If the publications they are replaced by were already published, their archive status is not changing, and no action is necessary.

Currently the only action performed is cache clearance, but in https://github.com/dfe-analytical-services/explore-education-statistics/pull/5868 this will be updated to also include raising a `PublicationArchivedEvent` event.

This change is being made to support adding the event and also because prior to it, this transition in archive status was not being determined, so cache clearance was happening for publications regardless.

### Other changes

- Rename Events `ReleaseVersionPublishedEvent.PublishedReleaseVersionInfo` to `ReleaseVersionPublishedEventInfo` and reinstate Publisher's own `PublishedReleaseVersionInfo` which now only has properties related directly to the published release version. A separate `PublishedPublicationInfo` is added to Publisher which holds useful info about publications affected by publishing, built after publications are updated.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/PublishedPublicationInfo.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/PublishedPublicationInfo.cs
@@ -1,10 +1,11 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
+using Generator.Equals;
 
 namespace GovUk.Education.ExploreEducationStatistics.Publisher.Services;
 
-public record PublishedPublicationInfo
+[Equatable]
+public partial record PublishedPublicationInfo
 {
     public required Guid PublicationId { get; init; }
 
@@ -35,6 +36,7 @@ public record PublishedPublicationInfo
     /// <summary>
     /// The list of information about published release versions for a publication.
     /// </summary>
+    [UnorderedEquality]
     public required IReadOnlyList<PublishedReleaseVersionInfo> PublishedReleaseVersions { get; init; }
 
     /// <summary>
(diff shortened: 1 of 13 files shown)


src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/PublishingCompletionService.cs
```diff
@@ -122,34 +94,91 @@ await publicationSlugsToUpdate
...
+
+    private async ValueTask<List<PublishedPublicationInfo>> PublishPublications(
+        IReadOnlyList<PublishedReleaseVersionInfo> releaseVersions)
+    {
+        return await releaseVersions
+            .GroupBy(info => info.PublicationId)
+            .ToAsyncEnumerable()
+            .SelectAwait(group => PublishPublication(
+                publicationId: group.Key,
+                publishedReleaseVersions: group.ToImmutableList()))
```
leeoades, line 130: `IReadOnlyList<T>` is sufficient, no? I just wondered why you chose the immutable list?

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/PublishingCompletionService.cs
```diff
@@ -122,34 +94,91 @@ await publicationSlugsToUpdate
...
+            PreviousLatestPublishedReleaseVersionId = previousLatestPublishedReleaseVersion?.Id,
+            PublishedReleaseVersions = publishedReleaseVersions
         };
     }
+
+    private async Task HandleArchivedPublications(IReadOnlyList<PublishedPublicationInfo> publications)
+    {
+        // Archived publications are those replaced by another publication with at least one published release.
+        // Filter publications from this run, excluding those already published.
+        var publicationIds = publications
```
leeoades, line 168: I suppose it would be slightly clearer to call this variable `nowArchivedPublicationIds` or something like that?

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/PublishedPublicationInfo.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// </summary>
+    public required IImmutableList<PublishedReleaseVersionInfo> PublishedReleaseVersions { get; init; }
+
+    /// <summary>
+    /// Indicates whether the publication was already published before the publishing run.
+    /// </summary>
+    ///
+    public bool WasAlreadyPublished => PreviousLatestPublishedReleaseVersionId != null;
+
+    public virtual bool Equals(PublishedPublicationInfo? other)
```
leeoades, line 46: It isn't necessary to build your own equals and hashcode any more. There is a package called Equatable that does it for you. You simply mark the class with an `[Equatable]` attribute and make the class `partial`. I hope you won't mind that I made the change for you and pushed it.

leeoades (approved):
There are some minimal optional suggestions for your enjoyment, plus my additional commit which I help is helpful. I've approved.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/PublishingCompletionService.cs
benoutram: These aren't the archived ones, they are ones having published releases for the first time, which can then affect other ones causing them to be archived because that status is now enforced.

But yes good point, I've gone with `newlyPublishedPublicationIds`.

src/GovUk.Education.ExploreEducationStatistics.Publisher/Services/PublishedPublicationInfo.cs
benoutram: I forgot about the `Equatable` package, good call suggesting we add this.

I needed to override equals for this

```
PublishedReleaseVersions.SequenceEqual(other.PublishedReleaseVersions)
```

otherwise there was a unit test failure comparing two `PublishedPublicationInfo`'s are equal, comparing the equality of  `PublishedReleaseVersions`.

Will `Equatable` solve this?

PR #5873, opened 2025-05-15
EES-6071: Add previous release id to release version published event

leeoades:
By including the previous release id, if the latest release has changed then we can use this information to delete the previous latest searchable document.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Events/ReleaseVersionPublishedEvent.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Events/ReleaseVersionPublishedEvent.cs
@@ -15,6 +15,7 @@ public record ReleaseVersionPublishedEvent : IEvent
             PublicationId = publishedReleaseVersion.PublicationId,
             PublicationSlug = publishedReleaseVersion.PublicationSlug,
             PublicationLatestPublishedReleaseVersionId = publishedReleaseVersion.PublicationLatestPublishedReleaseVersionId,
+            PreviousLatestReleaseId = publishedReleaseVersion.PreviousLatestReleaseId,
         };
     }
 
@@ -40,6 +41,7 @@ public record ReleaseVersionPublishedEvent : IEvent
         public required Guid PublicationId { get; init; }
         public required string PublicationSlug { get; init; }
         public required Guid PublicationLatestPublishedReleaseVersionId { get; init; }    
+        public Guid? PreviousLatestReleaseId { get; init; }
     }
     public EventPayload Payload { get; }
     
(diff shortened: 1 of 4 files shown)


src/GovUk.Education.ExploreEducationStatistics.Events/ReleaseVersionPublishedEvent.cs
```diff
@@ -15,6 +15,7 @@ public ReleaseVersionPublishedEvent(PublishedReleaseVersionInfo publishedRelease
             PublicationId = publishedReleaseVersion.PublicationId,
             PublicationSlug = publishedReleaseVersion.PublicationSlug,
             PublicationLatestPublishedReleaseVersionId = publishedReleaseVersion.PublicationLatestPublishedReleaseVersionId,
+            PreviousLatestReleaseId = publishedReleaseVersion.PreviousLatestReleaseId,
```
benoutram, line 18: I think we could name this `PreviousLatestPublishedReleaseId` to remove any ambiguity over whether its the latest release or the latest published release.

You don't make this change but if you agree I can change it when the next PR comes along.

src/GovUk.Education.ExploreEducationStatistics.Events/ReleaseVersionPublishedEvent.cs
```diff
@@ -40,18 +41,48 @@ public record EventPayload
...
+        /// <summary>
+        /// The release version that has been published may not be the latest.
+        /// This property contains the latest published release version id.
+        /// </summary>
         public Guid PublicationLatestPublishedReleaseVersionId { get; init; }
+        
+        /// <summary>
+        /// The Release Id of the previous "latest release version"
+        /// </summary>
+        public Guid? PreviousLatestReleaseId { get; init; }
```
benoutram, line 86: I think we could name this PreviousLatestPublishedReleaseId to remove any ambiguity over whether its the latest release or the latest published release.

It's then consistent with the naming of `PublicationLatestPublishedReleaseVersionId`. Alternatively we could go the other way and shorten that one. I think it depends if you think the fact it's in a type that already contains the name 'Published' gives it away.

You don't make this change but if you agree I can make the change when the next PR comes along.

benoutram (approved):

src/GovUk.Education.ExploreEducationStatistics.Events/ReleaseVersionPublishedEvent.cs
leeoades: Absolutely, I'm all for better names. I'll let you do that then. 👍

PR #5878, opened 2025-05-19
EES-5977 Track all-data-set zip download on Data Catalogue page

mmyoungman:
In the previous PR for EES-5977, we missed that it's possible to download a zip file with all data sets related to a release version. This PR fixes this by tracking this place as well.

This added a complication in that the same backend endpoint is used to download the all-data-set zip from both the Data Catalogue and Release pages. So we've added an extra query param to allow the backend to tell which page has triggered the request so that can be recorded in analytics.

### Other changes
- In the frontend, `@common`'s `downloadService.ts` was being used by both the public and admin frontend. Since we only need to record analytics on the public frontend, I've created a `downloadService.ts` for the `@frontend`.
- Also changed various links/buttonlinks into buttons on the frontend so they directly call the new `downloadService#downloadZip`.


Guy-HiveIT (approved):

mmyoungman:
> Can't speak to the backend but the frontend bits look good to me.
> 
> I don't understand your point about creating a new downloadService in the `@frontend` because of analytics reasons - from what I can see, the analytics calls are all separate to the downloadService functions anyway? If it's purely for code organisation and the zip download won't be used in admin, makes sense.

Yeah, it's just for code organisation. `@common`'s `downloadService` previously called two different endpoints in the backend - one in Admin and the other in the Content API - both with the same signature. I only wanted to change the Content API endpoint, so I had to create the new `@frontend` `downloadService`.

Writing this out did make me think maybe `@common`'s `downloadService` maybe should live in `@admin` now... I'll check

mmyoungman:
> Writing this out did make me think maybe @common's downloadService maybe should live in @admin now... I'll check

Not possible - used by common table tool code, amongst other things.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Requests/AnalyticsCaptureRequests.cs
```diff
@@ -14,4 +14,6 @@ public record CaptureZipDownloadRequest(
     string ReleaseName,
     string? ReleaseLabel,
     Guid? SubjectId = null,
-    string? DataSetTitle = null) : IAnalyticsCaptureRequestBase;
+    string? DataSetTitle = null,
+    // TODO: FromPage can be made non-optional later
```
mmoalin, line 18: Is this TODO comment still needed?

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Requests/AnalyticsCaptureRequests.cs
```diff
@@ -14,4 +14,6 @@ public record CaptureZipDownloadRequest(
...
+    string? FromPage = null) : IAnalyticsCaptureRequestBase;
```
duncan-at-hiveit, line 19: Would it make sense to keep this as an enum at this point?  I did so on https://github.com/dfe-analytical-services/explore-education-statistics/pull/5883 - you have to just use a `JsonConvert` attribute targeted at `property` level.

src/explore-education-statistics-frontend/src/modules/data-catalogue/DataSetFilePage.tsx
```diff
@@ -200,7 +190,19 @@ export default function DataSetFilePage({
                 'govuk-!-margin-bottom-0',
                 styles.infoDownloadButton,
               )}
-              onClick={handleDownload}
+              onClick={async () => {
```
duncan-at-hiveit, line 193: These chunks of code look the same.  Any reason why it was changed from a single shared `handleDownload` function?

src/explore-education-statistics-frontend/src/modules/data-catalogue/DataSetFilePage.tsx
```diff
@@ -251,7 +253,19 @@ export default function DataSetFilePage({
                 dataSetFileId={dataSetFile.id}
                 hasApiDataSet={!!apiDataSet}
                 tableToolLink={`/data-tables/${release.publication.slug}/${release.slug}?subjectId=${file.subjectId}`}
-                onDownload={handleDownload}
+                onDownload={async () => {
```
duncan-at-hiveit, line 256: These chunks of code look the same. Any reason why it was changed from a single shared handleDownload function?

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/FromPageEnum.cs
```diff
@@ -0,0 +1,8 @@
+namespace GovUk.Education.ExploreEducationStatistics.Common.Utils;
+
+public enum FromPage
```
duncan-at-hiveit, line 3: Would this potentially be reusable for other analytics too?  Do we think it is only going to be used for analytics, or could it be repurposed for other uses?  Main reason I ask is that perhaps we should signal that this is for the purposes ofg analytics by maybe including "Analytics" in the name somehow...?  Or could this potentially make its way into the common Analytics project?

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer.Tests/Services/PublicZipDownloadsProcessorTests.cs
```diff
@@ -228,5 +230,6 @@ private record ZipDownloadReportLine(
         string ReleaseLabel,
         Guid? SubjectId,
         string? DataSetTitle,
+        string? FromPage,
```
duncan-at-hiveit, line 233: This looks optional here but mandatory in the Controller.  Is it one or the other, or maybe I've gotten the wrong end of the stick with its nullability in the Controller.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Requests/AnalyticsCaptureRequests.cs
mmyoungman: I have intentionally left it. It is a change we could make in the future but it didn't seem worth creating a ticket about - and I don't see a big advantage to changing it beyond removing any confusion if someone wonders why it is optional.

I was debating over whether to just keep `FromPage` as an optional query param and keep it optional throughout, but in the end I left the query param as a required.

src/GovUk.Education.ExploreEducationStatistics.Content.Api.Tests/Controllers/ReleaseFileControllerTests.cs
```diff
@@ -260,20 +261,21 @@ await TestApp.AddTestData<ContentDbContext>(context =>
...
+                        FromPage.ReleaseUsefulInfo,
                         It.Is<IEnumerable<Guid>>(
                             ids => ids.SequenceEqual(ListOf(fileId))),
                         It.IsAny<CancellationToken>()
                     )
                 )
                 .ReturnsAsync(Unit.Instance)
-                .Callback<Guid, Stream, IEnumerable<Guid>, CancellationToken?>(
-                    (_, stream, _, _) => stream.WriteText("Test zip"));
+                .Callback<Guid, Stream, FromPage, IEnumerable<Guid>, CancellationToken?>(
```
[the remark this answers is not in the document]
mmyoungman: No idea - I lost at least an hour not realising I needed to update this.

src/GovUk.Education.ExploreEducationStatistics.Content.Api/Controllers/ReleaseFileController.cs
```diff
@@ -62,6 +62,7 @@ public async Task<ActionResult> StreamFilesToZip(
             // zip file, hence why this endpoint takes an array of fileIds, but this is no longer the case. Via the
             // public frontend, users only download all the releaseVersion's data (by not providing fileIds) or provide
             // a single fileId for a specific data set.
+            [FromQuery] FromPage fromPage,
```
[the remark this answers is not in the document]
mmyoungman: I did debate whether to make the query param optional, but in the end I left it as required. 

I changed a few links into buttons on the frontend, and did ask Cam if we intended them to be links so users could copy and paste the download link, but Cam said nope and that they're fine as buttons. I don't think there is an easy way currently for the user to fetch the link itself, and so I don't see an issue with the query param being required as it isn't really visible to the user.

All that said, I'm happy to change it if you think it's better to be optional - I wasn't confident about my decision here, to be honest.

src/explore-education-statistics-frontend/src/modules/data-catalogue/DataSetFilePage.tsx
mmyoungman: I think I inlined it for some reason which later disappeared, and didn't bother to change it back. I figure a single repetition is fine, but happy to recreate `handleDownload` if you think that's better.

src/GovUk.Education.ExploreEducationStatistics.Content.Services/Requests/AnalyticsCaptureRequests.cs
mmyoungman: Actually thinking about this more, I'm not sure this does need to be optional. It only needs to be optional the the request file consumer side...

src/GovUk.Education.ExploreEducationStatistics.Common/Utils/FromPageEnum.cs
mmyoungman: I've renamed this `AnalyticsFromPage` and moved the enum into the Analytics.Common project.

src/GovUk.Education.ExploreEducationStatistics.Analytics.Consumer.Tests/Services/PublicZipDownloadsProcessorTests.cs
mmyoungman: I decided to make this optional as after the deploy, it's possible the Analytics.Consumer function could process request files that don't include an `AnalyticsFromPage`, and I assume this would break things if it is required. So I made this optional to prevent this.

As soon as this is deployed and all older request files have been processed, this could be made mandatory.

All that said, I had also made this optional on the Content API side, which isn't necessary, so I'm going to make it mandatory there. I will also add a comment to Analytics.Consumer to explain why it is currently optional.

mmyoungman: Actually looking at the code again, I think I don't need to worry about this, so I've made fromPage mandatory everywhere.

There is still a question of whether it should be mandatory, but that can be addressed in the other PR comment.

duncan-at-hiveit: Nice one!

src/explore-education-statistics-frontend/src/modules/data-catalogue/DataSetFilePage.tsx
duncan-at-hiveit: I'd be tempted to, just to avoid a scenario where someone updates one not realising there's another identical one that also needs updating?

PR #5881, opened 2025-05-19
Ees 6107 changes to release version published event

benoutram:
Following on from https://github.com/dfe-analytical-services/explore-education-statistics/pull/5867 this PR makes changes to `ReleaseVersionPublishedEvent`, `ReleaseVersionPublishedEventInfo`, and `ReleaseVersionPublishedEventDto` to tidy up some of the properties related to the published release version's publication's previous and latest published release.

### Consistency

To make the naming of properties consistent with each other:

- Rename `PublicationLatestPublishedReleaseVersionId` to `LatestPublishedReleaseVersionId`
- Rename `PreviousLatestReleaseId` to `PreviousLatestPublishedReleaseId`

### Add useful values to the Event

To provide a full compliment of both 'latest' and 'previous', 'release id' and 'release version id' properties:

- Add `LatestPublishedReleaseId`
- Add `PreviousLatestPublishedReleaseVersionId`

### Result

`ReleaseVersionPublishedEvent`, `ReleaseVersionPublishedEventInfo`, and `ReleaseVersionPublishedEventDto` now have the following latest/previous properties consistently:

```
LatestPublishedReleaseId
LatestPublishedReleaseVersionId
PreviousLatestPublishedReleaseId
PreviousLatestPublishedReleaseVersionId
```

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/EventHandlers/OnReleaseVersionPublished/Dtos/ReleaseVersionPublishedEventDto.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/EventHandlers/OnReleaseVersionPublished/Dtos/ReleaseVersionPublishedEventDto.cs
@@ -6,42 +6,50 @@ public record ReleaseVersionPublishedEventDto
     /// Newly published release version id
     /// </summary>
     public Guid ReleaseVersionId { get; init; }
-    
+
     /// <summary>
     /// The Release Id for the newly published release version
     /// </summary>
-    public Guid? ReleaseId {get;init;}
-    
+    public Guid? ReleaseId { get; init; }
+
     /// <summary>
     /// The release slug for the newly published release version
     /// </summary>
     public string? ReleaseSlug { get; init; }
-    
+
     /// <summary>
     /// The publication id for the newly published release version
     /// </summary>
     public Guid? PublicationId { get; init; }
-    
+
     /// <summary>
     /// The publication slug for the newly published release version
     /// </summary>
     public string? PublicationSlug { get; init; }
-    
+
+    /// <summary>
+    /// </summary>
+    public Guid? LatestPublishedReleaseId { get; init; }
+
     /// <summary>
     /// The Release Version Id of the current "latest release version".
     /// </summary>
     public Guid? LatestPublishedReleaseVersionId { get; init; }
-    
+
     /// <summary>
     /// The Release Id of the previous "latest release version"
     /// </summary>
     public Guid? PreviousLatestPublishedReleaseId { get; init; }
-    
+
+    /// <summary>
+    /// </summary>
+    public Guid? PreviousLatestPublishedReleaseVersionId { get; init; }
+
     /// <summary>
     /// Is this newly published release version the new latest version?
     /// </summary>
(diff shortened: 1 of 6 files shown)


src/GovUk.Education.ExploreEducationStatistics.Events/ReleaseVersionPublishedEvent.cs
```diff
@@ -52,37 +52,37 @@ public record ReleaseVersionPublishedEventInfo
...
         /// </summary>
-        public Guid PublicationLatestPublishedReleaseVersionId { get; init; }
-        
+        public required Guid LatestPublishedReleaseVersionId { get; init; }
+
         /// <summary>
         /// The Release Id of the previous "latest release version"
         /// </summary>
-        public Guid? PreviousLatestReleaseId { get; init; }
+        public required Guid? PreviousLatestPublishedReleaseId { get; init; }
```
leeoades, line 96: Minor: I don't think you need to mark a nullable field as required.

src/GovUk.Education.ExploreEducationStatistics.Events/ReleaseVersionPublishedEvent.cs
```diff
@@ -44,8 +44,8 @@ public record EventPayload
         public required string PublicationSlug { get; init; }
         public required Guid LatestPublishedReleaseId { get; init; }
         public required Guid LatestPublishedReleaseVersionId { get; init; }
-        public Guid? PreviousLatestPublishedReleaseId { get; init; }
-        public Guid? PreviousLatestPublishedReleaseVersionId { get; init; }
+        public required Guid? PreviousLatestPublishedReleaseId { get; init; }
```
leeoades, line 47: Minor: same comment really - do these really need to be required since they are nullable?

leeoades (approved):
Cool stuff.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Functions/EventHandlers/OnReleaseVersionPublished/Dtos/ReleaseVersionPublishedEventDto.cs
```diff
@@ -6,44 +6,56 @@ public record ReleaseVersionPublishedEventDto
...
+    /// The latest published release version id of the publication's latest published release before the release version was published.
+    /// </summary>
+    public Guid? PreviousLatestPublishedReleaseVersionId { get; init; }
+
     /// <summary>
     /// Is this newly published release version the new latest version?
     /// </summary>
-    public bool NewlyPublishedReleaseVersionIsLatest => PublicationLatestPublishedReleaseVersionId == ReleaseVersionId;
-    
+    public bool NewlyPublishedReleaseVersionIsLatest => LatestPublishedReleaseVersionId == ReleaseVersionId;
```
leeoades, line 55: I suppose there is a bit of an edge case here where for some reason the event does not contain either of these values, so both are null, and therefore `NewlyPublishedReleaseVersionIsLatest` is true.
We ought to really introduce some validators that ensures all of the events are full formed and not missing vital information? (Not in this PR obvs)

benoutram: Yeah it does seem like an edge case. That might happen if the contract changed before this type was updated? Validators do seem like a good idea. 👍

PR #5883, opened 2025-05-20
EES-6129 - added the capture of the "Download CSV" Public API endpoint

duncan-at-hiveit:
This PR:
- adds the capture of "Download CSV" endpoint calls to the Public API.

## Details being captured

This is the first of many "capture" tasks for DataSetVersion-level calls to the Public API.  In general they all contain a common set of information, notably:

* DataSetId - the Id of the overarching data set that owns the version being downloaded
* DataSetTitle - the title of the overarching data set that owns the version being downloaded
* DataSetVersionId - the Id of the actual version that is being downloaded
* DataSetVersion - the Semver of the actual version that is being downloaded
* PreviewToken - an optional preview token that is taken from the Preview Token HTTP header
* RequestedDataSetVersion - an optional HTTP parameter that allows specific or wildcarded data set versions to be targeted 
* StartTime - the datetime when the request was made

In addition to the common details above, some calls support additional parameters, captured in a generic "Parameters" field.

And finally, all DataSetVersion-level calls support a "Type" field which is used to identify the type of call being made.  In this PR, the call is "DownloadCsv".

### PreviewTokenService

I needed easy access to the optional Preview Token that is provided in an HTTP header.  We already had a way to pull this string value (its Id) from the header and then pass the value to a `PreviewTokenService` for validation, but that didn't help me too much, other than potentially create some duplication.

Therefore, I moved the grabbing of the token id from the HTTP header *into* PreviewTokenService itself, plus added a method for getting the optional PreviewToken entity from the database if the HTTP header was present.  This way, our analytics code could get to the Preview Token much more easily.

### AnalyticsService

Because the capturing of the Download CSV endpoint was a bit more involved than the existing Public API Query-capturing code, I felt it best to move it into a separate `AnalyticsService` and keep `DataSetService` a little cleaner.

My plan is to then do the same thing for the existing Public API Query-capturing code, thus making `DataSetQueryService` a little cleaner too, but this would have bloated this PR.

After that is in place, I was also pondering creating a NoOpAnalyticsService within the Public API project, and have the selection of that no-op implementation dependent on the "AnalyticsEnabled" flag, rather than using a `NoOpAnalyticsManager`, but it's not a massive advantage unless we're actually planning to disable analytics anywhere to gain a performance edge.

## Updates to Analytics.Tests tests

Taking the JSON output from `AnalyticsWriteDataSetVersionLevelCallsStrategyTests.cs`, I transferred them to the equivalent Analytics.Consumer report creation test (PublicApiDataSetVersionCallsProcessorTests.cs) Resources folder and updated expectations there to match any small variations in behaviour.

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/AnalyticsCaptureRequests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/AnalyticsCaptureRequests.cs
@@ -1,3 +1,5 @@
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
 using GovUk.Education.ExploreEducationStatistics.Analytics.Common.Interfaces;
 
 namespace GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Requests;
@@ -21,6 +23,7 @@ public record CaptureDataSetVersionCallRequest(
     PreviewTokenRequest? PreviewToken,
     string? RequestedDataSetVersion,
     DateTimeOffset StartTime,
+    [property:JsonConverter(typeof(StringEnumConverter))]
     DataSetVersionCallType Type,
     object? Parameters = null
 ) : IAnalyticsCaptureRequestBase;
@@ -32,7 +35,6 @@ public enum DataSetVersionCallType
     DownloadCsv,
     GetChanges
 }
-
 
 public record PreviewTokenRequest(
     string Label,
--- a/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/AnalyticsCaptureRequests.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/AnalyticsCaptureRequests.cs
@@ -1,3 +1,4 @@
+using GovUk.Education.ExploreEducationStatistics.Public.Data.Api.Model;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
 using GovUk.Education.ExploreEducationStatistics.Analytics.Common.Interfaces;
@@ -35,6 +36,10 @@ public enum DataSetVersionCallType
     DownloadCsv,
     GetChanges
 }
+
+public record GetMetadataAnalyticsParameters(
+    [property:JsonProperty(ItemConverterType=typeof(StringEnumConverter))]
+    IEnumerable<DataSetMetaType> Types);
 
 public record PreviewTokenRequest(
     string Label,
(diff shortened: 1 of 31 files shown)


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/DataSetService.cs
```diff
@@ -45,14 +48,31 @@ public async Task<Either<ActionResult, FileStreamResult>> DownloadDataSet(
                 dataSetVersion: dataSetVersion,
                 cancellationToken: cancellationToken)
             .OnSuccessDo(userService.CheckCanViewDataSetVersion)
+            .OnSuccessDo(async dsv =>
+            {
+                try
+                {
+                    await analyticsService.CaptureDataSetVersionCall(
```
mmyoungman, line 55: In `ReleaseFileService`, we handle a similar situation by creating a private `RecordZipDownloadAnalytics` method inside the `ReleaseFileService` class.

I don't have much of a preference as to which way we do it, but would be good to keep it consistent.

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/PreviewTokenService.cs
```diff
@@ -59,11 +79,31 @@ private async Task<bool> HasActivePreviewToken(
...
+        return await publicDataDbContext
+            .PreviewTokens
             .Where(pt => pt.Id == previewTokenId)
             .Where(pt => pt.DataSetVersionId == dataSetVersionId)
             .Where(pt => pt.DataSetVersion.Status == DataSetVersionStatus.Draft)
             .Where(pt => pt.Expiry > DateTimeOffset.UtcNow)
             .AnyAsync(cancellationToken: cancellationToken);
     }
+    
+    private Guid? GetPreviewTokenIdFromRequest()
```
mmyoungman, line 91: Nice bit of clean up :)

src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Requests/AnalyticsCaptureRequests.cs
```diff
@@ -12,3 +15,30 @@ public record CaptureDataSetVersionQueryRequest(
...
+
+public record CaptureDataSetVersionCallRequest(
+    Guid DataSetId,
+    Guid DataSetVersionId,
+    string DataSetVersion,
+    string DataSetTitle,
+    PreviewTokenRequest? PreviewToken,
+    string? RequestedDataSetVersion,
+    DateTimeOffset StartTime, 
+    [property:JsonConverter(typeof(StringEnumConverter))]
```
mmyoungman, line 27: Didn't know you could do this - cool

mmyoungman (approved):

mmyoungman:
I rebased off dev to resolve a merge conflict

mmyoungman (approved):

PR #5894, opened 2025-05-22
EES-6071 Delegate logging to Serilog

leeoades:
This has been an ordeal!
It is recommended that the 3 commits are reviewed one at a time in order for added clarity of intention.

### Goal
Replace the default logging to the console and Application Insights with Serilog.
Retain metrics measurements sent to App Insights.

### Why
The primary reason is that Serilog adds the ability to destructure objects into their constituent properties within the logs.

### Remarks
Logging settings in appsettings.json, appsettings.development.json and host.json are unused and therefore removed.

### Additional
Remove unused using declarations across the solution

### DIFF
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Extensions/ServiceCollectionExtensions.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Extensions/ServiceCollectionExtensions.cs
@@ -5,6 +5,7 @@ using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Func
 using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.Azure.Functions.Worker;
 using Microsoft.Extensions.Azure;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Serilog;
@@ -33,25 +34,31 @@ public static class ServiceCollectionExtensions
             .AddTransient<Func<ISearchIndexerClient>>(sp => sp.GetRequiredService<ISearchIndexerClient>);
 
     public static IServiceCollection ConfigureLogging(
-        this IServiceCollection serviceCollection) =>
+        this IServiceCollection serviceCollection,
+        IConfiguration configuration) =>
         serviceCollection
+            .SetupAppInsights()
             .AddLogging(lb =>
             {
                 // Prevent the default Azure Function logging provider from logging.
                 // Instead, let Serilog handle that.
                 lb.SetMinimumLevel(LogLevel.None);
-                
+
+                // Resolve the telemetry configuration that was registered by SetupAppInsights
+                var telemetryConfiguration = serviceCollection.BuildServiceProvider().GetRequiredService<TelemetryConfiguration>();
+
                 // Setup Serilog to log to the Console and to App Insights
                 lb.AddSerilog(
                     new LoggerConfiguration()
+                        .ReadFrom.Configuration(configuration)
                         .Enrich.WithExceptionDetails()
+                        .Enrich.FromLogContext()
                         .WriteTo.Console()
-                        .WriteTo.ApplicationInsights(TelemetryConfiguration.CreateDefault(), TelemetryConverter.Traces)
+                        .WriteTo.ApplicationInsights(telemetryConfiguration, TelemetryConverter.Traces)
                         .CreateLogger(),
                     dispose: true
                 );
             })
-            .SetupAppInsights()
         ;
 
     private static IServiceCollection SetupAppInsights(this IServiceCollection serviceCollection) =>
@@ -66,6 +73,9 @@ public static class ServiceCollectionExtensions
             .Configure<LoggerFilterOptions>(
                 options =>
                 {
+                    // The Application Insights SDK adds a default logging filter that instructs ILogger to capture
+                    // only Warning and more severe logs. Application Insights requires an explicit override.
+                    
                     // Find the built-in App Insights logging rule
                     var defaultRule = options.Rules.FirstOrDefault(
                         rule => rule.ProviderName ==
--- a/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Extensions/ServiceCollectionExtensions.cs
+++ b/src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Extensions/ServiceCollectionExtensions.cs
@@ -1,9 +1,12 @@
-using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Clients.AzureBlobStorage;
+﻿using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Clients.AzureBlobStorage;
 using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Clients.AzureSearch;
 using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Clients.ContentApi;
 using GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Functions.HealthChecks.Strategies;
+using Microsoft.Azure.Functions.Worker;
 using Microsoft.Extensions.Azure;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 
 namespace GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp.Extensions;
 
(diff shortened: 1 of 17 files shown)


src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -31,26 +33,59 @@ public static IServiceCollection AddHealthChecks(this IServiceCollection service
...
+                // Prevent the default Azure Function logging provider from logging.
+                // Instead, let Serilog handle that.
+                lb.SetMinimumLevel(LogLevel.None);
+                
+                // Setup Serilog to log to the Console and to App Insights
+                lb.AddSerilog(
+                    new LoggerConfiguration()
+                        .Enrich.WithExceptionDetails()
+                        .WriteTo.Console()
+                        .WriteTo.ApplicationInsights(TelemetryConfiguration.CreateDefault(), TelemetryConverter.Traces)
```
benoutram, line 49: Reading the docs it says we can pass in the `TelemetryConfiguration` instance that was added by `services.AddApplicationInsightsTelemetryWorkerService()` using DI like

```csharp
serviceProvider.GetRequiredService<TelemetryConfiguration>()
```

Did you already give this a try instead of using `TelemetryConfiguration.CreateDefault()`? I've seen it in a few other examples too.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/appsettings.json
```diff
@@ -1,11 +1,2 @@
 {
-  "Logging": {
-    "LogLevel": {
-      "Default": "Error",
-      "GovUk.Education.ExploreEducationStatistics": "Information",
-      "Microsoft": "Warning"
-    }
-  },
```
benoutram, line 8: Do we not still need an equivalent of this now, either in code or in a config file, to add in overrides like we have here with `"Microsoft": "Warning"`?

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -26,4 +31,61 @@ public static IServiceCollection AddHealthChecks(this IServiceCollection service
...
+                {
+                    // Find the built-in App Insights logging rule
+                    var defaultRule = options.Rules.FirstOrDefault(
+                        rule => rule.ProviderName ==
+                                "Microsoft.Extensions.Logging.ApplicationInsights.ApplicationInsightsLoggerProvider");
+
+                    if (defaultRule is not null)
+                    {
+                        // Remove the default rule
+                        options.Rules.Remove(defaultRule);
```
benoutram, line 77: I think the comment we had here before explaining why the default rule is removed, i.e. to include logs of severity < Warning, would still be good to keep here, assuming we still need to do this?

```
// The Application Insights SDK adds a default logging filter that instructs ILogger to capture
// only Warning and more severe logs. Application Insights requires an explicit override.
```

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -26,4 +31,61 @@ public static IServiceCollection AddHealthChecks(this IServiceCollection service
...
             .AddTransient<Func<ISearchIndexerClient>>(sp => sp.GetRequiredService<ISearchIndexerClient>);
+
+    public static IServiceCollection ConfigureLogging(
+        this IServiceCollection serviceCollection) =>
+        serviceCollection
+            .AddLogging(lb =>
+            {
+                // Prevent the default Azure Function logging provider from logging.
+                // Instead, let Serilog handle that.
+                lb.SetMinimumLevel(LogLevel.None);
```
benoutram, line 42: Is this to prevent duplicate logging? I didn't see this in any examples that I came across when I did a quick search.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -26,4 +31,61 @@ public static IServiceCollection AddHealthChecks(this IServiceCollection service
...
+                        // Create a new rule, manually setting the log level to None
+                        // to prevent the function from logging to App Insights.
+                        // Let Serilog handle the logging.
+                        options.Rules.Add(
+                            new LoggerFilterRule(
+                                defaultRule.ProviderName,
+                                defaultRule.CategoryName,
+                                LogLevel.None,
+                                defaultRule.Filter
+                            ));
```
benoutram, line 88: It seems like there are a lot of cases out there having issues with duplicate logs and a lot of different of different solutions! 😬 Nice work on trying to fix this in advance of us deploying it! 💯

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -31,26 +33,59 @@ public static IServiceCollection AddHealthChecks(this IServiceCollection service
...
+                // Instead, let Serilog handle that.
+                lb.SetMinimumLevel(LogLevel.None);
+                
+                // Setup Serilog to log to the Console and to App Insights
+                lb.AddSerilog(
+                    new LoggerConfiguration()
+                        .Enrich.WithExceptionDetails()
+                        .WriteTo.Console()
+                        .WriteTo.ApplicationInsights(TelemetryConfiguration.CreateDefault(), TelemetryConverter.Traces)
+                        .CreateLogger(),
```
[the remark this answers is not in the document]
leeoades: Yes, good idea. I've added the config into appsettings which we can now easily adjust.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -31,26 +33,59 @@ public static IServiceCollection AddHealthChecks(this IServiceCollection service
...
+            .AddLogging(lb =>
+            {
+                // Prevent the default Azure Function logging provider from logging.
+                // Instead, let Serilog handle that.
+                lb.SetMinimumLevel(LogLevel.None);
+                
+                // Setup Serilog to log to the Console and to App Insights
+                lb.AddSerilog(
+                    new LoggerConfiguration()
+                        .Enrich.WithExceptionDetails()
```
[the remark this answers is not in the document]
leeoades: Added.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/appsettings.json
leeoades: I've added this into the logging overrides which we now read from config.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Extensions/ServiceCollectionExtensions.cs
leeoades: Yes, I wasn't sure whether to keep that or not. I've added it back in.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Extensions/ServiceCollectionExtensions.cs
leeoades: Are you saying that the comment isn't clear enough? 🤣
So, the log messages are send to every logging provider. The serilog one does its magic. The default Azure Function one is more basic. However, if we remove the default one (which I was doing) then the Serilog Console sink doesn't work - nothing is output to the console. And if I add back in the default Console logger, we start seeing all of the in built logging again. Only way to prevent it is to do the above.
Did I mention that I've been wrestling with this for a while?

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Extensions/ServiceCollectionExtensions.cs
leeoades: I did have a go at organising the logger like that. It involved a bit of a faff but I've come up with a little hack to resolve it. I think you're right that it's more likely to work if we utilise the config that is there already rather than just hoping the default one will magically work.

src/GovUk.Education.ExploreEducationStatistics.Content.Search.FunctionApp/Extensions/ServiceCollectionExtensions.cs
leeoades: Ben, your thoroughness of review puts us all to shame! Thank you for double-checking all of these things for me and for making my code submissions much better!

benoutram (approved):

PR #5895, opened 2025-05-22
EES 5779 / EES-6177 cancel data replacements

mmoalin:
# Summary
The main change here is that when user clicks the 'Cancel' button on an ongoing replacement, if that replacement has an API data set version attached to it, that api data set version also gets deleted.

Tweaks the UI & backend of the 'data file details' tab support 
1. Displaying when a major version increment has occurred to the user in the data replacement details page
2. Displaying cancel button text changes conditionally if the user is replacing a data file with api version. when user clicks, modal shows asking for user to confirm cancelling replacement and deleting api version. (i.e., when there is a replacement with an API data set, it displays text that informs the attached api data set will be deleted) 
3. Cancelling an ongoing replacement that has a draft API data set attached to it deletes the associated draft api data set.

Changes to `DataSetVersionMappingService` include: to move out logic that was fetching mapping related data from database into a new repository class (`IMappingTypesRepository`), to move in logic for getting mapping status (from `DataSetService`) and to add method `IsMajorVersionUpdate` to use within the data and files data file details page to inform the user there has been a major version detected from an automap or manual map of api data set.

## UI Changes include:
1: Major change selected found
image

2: Manual mapping required
image

3: Manual mapping completed but not finalized (please note the backend changes to make this work are delivered in the next PR
image

4: Manual mapping complete and finalized
image

5: Cancel button confirm modal
image


jack-hive (changes requested):
Please check everywhere in your changes for where new lines would be needed. I've probably missed some places

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseVersionService.cs
```diff
@@ -758,7 +768,7 @@ private async Task<Either<ActionResult, Unit>> CheckCanDeleteDataFiles(
                 return ValidationActionResult(CannotRemoveDataFilesOnceReleaseApproved);
             }
 
-            if (releaseFile.PublicApiDataSetId is not null)
+            if (!featureFlags.Value.EnableReplacementOfPublicApiDataSets && releaseFile.PublicApiDataSetId is not null)
```
jack-hive, line 771: Don't we also want to check the case where the feature flag is **turned on** but the attached Public API data set is **Published**, and therefore can no longer be deleted?

I can't 100% remember if we should block deleting published ones, but I'm pretty sure we wouldn't want to as that would break people's APIs

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseVersionService.cs
```diff
@@ -589,14 +592,14 @@ await footnoteRepository.GetFootnotes(releaseVersionId: releaseVersionId,
...
-                        : new DeleteApiDataSetVersionPlanViewModel
+                        : new ReplacementApiDataSetVersionPlanViewModel
                         {
                             DataSetId = tuple.apiDataSetVersion.DataSetId,
                             DataSetTitle = tuple.apiDataSetVersion.DataSet.Title,
                             Id = tuple.apiDataSetVersion.Id,
                             Version = tuple.apiDataSetVersion.PublicVersion,
                             Status = tuple.apiDataSetVersion.Status,
-                            Valid = false
+                            Valid = featureFlags.Value.EnableReplacementOfPublicApiDataSets
```
jack-hive, line 602: Same as comment below. I think this also depends if the Public API data set is **Published** or not?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReplacementService.cs
```diff
@@ -138,28 +105,28 @@ public async Task<Either<ActionResult, DataReplacementPlanViewModel>> GetReplace
...
-                apiDataSetVersionPlan.MappingStatus = await _dataSetService.GetMappingStatus(replacementApiDataSetVersion.Id, cancellationToken);
-                apiDataSetVersionPlan.Valid = apiDataSetVersionPlan.MappingStatus is
-                    { FiltersComplete: true, LocationsComplete: true };
+                return apiDataSetVersionPlan;
             }
-
+            //TODO: Please note EES-5779 PR Note this will be updated in an upcomming PR - Valid's value will take into account
+            //more things than just the values { FiltersComplete: true, LocationsComplete: true }. It will
+            //also include whether the user has 'finalized' the data set version mapping & whether the version has
+            //a breaking change and results in a major version increment.
```
jack-hive, line 126: Can this comment be removed now?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReplacementService.cs
```diff
@@ -138,28 +105,28 @@ public async Task<Either<ActionResult, DataReplacementPlanViewModel>> GetReplace
...
+            var apiDataSetVersionPlan = new ReplacementApiDataSetVersionPlanViewModel
+            {
+                DataSetId = replacementApiDataSetVersion.DataSetId,
+                DataSetTitle = replacementApiDataSetVersion.DataSet.Title,
+                Id = replacementApiDataSetVersion.Id,
+                Version = replacementApiDataSetVersion.PublicVersion,
+                Status = replacementApiDataSetVersion.Status,
+                Valid = false,
+            };
+            if (!featureFlags.Value.EnableReplacementOfPublicApiDataSets || replacementApiDataSetVersion.VersionPatch <= 0)
```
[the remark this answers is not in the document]
jack-hive: Also, I can't quite remember why we have this check here... I may need to jump on a call with you again

src/explore-education-statistics-admin/src/pages/release/data/ReleaseDataFileReplacePage.tsx
```diff
@@ -125,13 +125,33 @@ const ReleaseDataFileReplacePage = ({
       },
     });
   };
-
+  const cancelText =
+    dataFile?.publicApiDataSetId === undefined
```
jack-hive, line 129: I'm not 100% sure but I think you might be able to remove the `=== undefined`. Worth trying out!

src/explore-education-statistics-admin/src/pages/release/data/ReleaseDataFileReplacePage.tsx
```diff
@@ -125,13 +125,33 @@ const ReleaseDataFileReplacePage = ({
...
+      ? 'Cancel data replacement'
+      : 'Cancel data replacement & delete attached draft API version';
+  const cancelBodyText =
+    dataFile?.publicApiDataSetId === undefined ? (
```
jack-hive, line 133: same again here

jack-hive (changes requested):

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseVersionService.cs
```diff
@@ -605,7 +608,7 @@ await footnoteRepository.GetFootnotes(releaseVersionId: releaseVersionId,
                         SubjectId = tuple.subject.Id,
                         DeleteDataBlockPlan = await dataBlockService.GetDeletePlan(releaseVersionId, tuple.subject),
                         FootnoteIds = footnotes.Select(footnote => footnote.Id).ToList(),
-                        DeleteApiDataSetVersionPlan = linkedApiDataSetVersionDeletionPlan
+                        ApiDataSetVersionPlan = linkedApiDataSetVersionDeletionPlan
```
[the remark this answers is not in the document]
mmoalin: Yeah, this is used in the data file replacement tab when the replacement is on-going for a data file that has an api version attached to. Lines 104 @ 
\src\pages\release\data\components\DataFileReplacementPlan.tsx

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseVersionService.cs
mmoalin: Cheers Jack

src/explore-education-statistics-admin/src/pages/release/data/ReleaseDataFileReplacePage.tsx
mmoalin: Cheers Jack, you're right, that works.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReplacementService.cs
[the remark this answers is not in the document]
mmoalin: Sure thing, let's jump on a call whenever best! Just on why the check is here, when the feature flag is turned on, then we want to use the new api mapping status to determine what to display on the data file replacement details tab. Whereas when the feature flag is turned off, we want to set Valid to false and return there.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReplacementService.cs
```diff
@@ -138,28 +105,28 @@ public async Task<Either<ActionResult, DataReplacementPlanViewModel>> GetReplace
...
-                    { FiltersComplete: true, LocationsComplete: true };
+                return apiDataSetVersionPlan;
             }
-
+            //TODO: Please note EES-5779 PR Note this will be updated in an upcomming PR - Valid's value will take into account
+            //more things than just the values { FiltersComplete: true, LocationsComplete: true }. It will
+            //also include whether the user has 'finalized' the data set version mapping & whether the version has
+            //a breaking change and results in a major version increment. 
+            apiDataSetVersionPlan.MappingStatus =  await dataSetVersionMappingService.GetMappingStatus(replacementApiDataSetVersion.Id, cancellationToken);
+            apiDataSetVersionPlan.MappingStatus.Complete = replacementApiDataSetVersion.Status == DataSetVersionStatus.Draft;
```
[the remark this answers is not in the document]
mmoalin: HI Jack, I have renamed the attribute from `Complete` to `finishedMapping` as that describes it better. 

So this property is used in the frontend on the data file replacement details page in order to indicate to the user whether they need to press the 'Finalise this data set version' button on the API data set details page after they have successfully manually mapped the locations and filters. 

I've noticed the data set version status changes from `Mapping` to `Draft` once the user presses this button and so I thought it would be useful here to inform the user when they have finished uploading a replacement file to finalize the data set mapping on the API details tab.  
`apiDataSetVersionPlan.MappingStatus.LocationsComplete` and `apiDataSetVersionPlan.MappingStatus.FiltersComplete` are independent from when the user presses this 'Finalise this data set version' button and so I've had to add this new property `finishedMapping` to use for informing the user to press that button after mapping.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReplacementService.cs
```diff
@@ -138,28 +105,28 @@ public async Task<Either<ActionResult, DataReplacementPlanViewModel>> GetReplace
...
+            apiDataSetVersionPlan.Valid = apiDataSetVersionPlan.ValidDefinition;
```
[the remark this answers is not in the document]
mmoalin: Thanks Jack! That's a fair comment, had removed ValidDefinition and assigned the properties. 

Although I agree using the suggested syntax looks nicer, it might do more than we need in this context in my opinion. 

I think using the `with` expression "Nondestructive mutation" creates a new object with modified properties which to me feels overkill for we need here (as opposed to using the same objects and setting it's properties). 
I don't think there's any need to copy data into a new object as this object is a local object that is about to be returned and it isn't shared or needs to be immutable.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ApiDataSetVersionPlanViewModel.cs
```diff
@@ -16,12 +16,17 @@ public record DeleteApiDataSetVersionPlanViewModel
     public string Version { get; init; } = null!;
 
     public DataSetVersionStatus Status { get; init; }
-
+    
     public bool Valid { get; set; }
```
jack-hive, line 20: Could be `init` to make this record immutable

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ApiDataSetVersionPlanViewModel.cs
```diff
@@ -16,12 +16,17 @@ public record DeleteApiDataSetVersionPlanViewModel
...
 }
 
-public record ApiDataSetVersionPlanViewModel : DeleteApiDataSetVersionPlanViewModel
+public record DeleteApiDataSetVersionPlanViewModel : ApiDataSetVersionPlanViewModel;
+
+public record ReplaceApiDataSetVersionPlanViewModel : ApiDataSetVersionPlanViewModel
 {
     public MappingStatusViewModel? MappingStatus { get; set; }
```
jack-hive, line 27: Could be `init` to make this record immutable

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ApiDataSetVersionPlanViewModel.cs
jack-hive, line 20: might be worth just adding a comment here to say this property will also be made abstract in the future (same as your other comment) - just so we don't forget (which I most definitely will!)

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseVersionService.cs
jack-hive: Awesome, I think that's the correct call. Might just be worth double checking this with Cam or someone though just in case I'm wrong? I.e. should we allow them to delete a data file as long as the feature flag is turned on (even though the feature flag is technically for replacements, not deletions) **AND** the Public API data set version attached **IS NOT** Published

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseVersionService.cs
```diff
@@ -631,17 +636,24 @@ public async Task<Either<ActionResult, Unit>> RemoveDataFiles(Guid releaseVersio
...
                 .OnSuccessVoid(() => releaseDataFileService.Delete(releaseVersionId, fileId));
         }
 
+        private async Task<Either<ActionResult, Unit>> DeleteApiDataSetVersionIfAttached(DeleteDataFilePlanViewModel deletePlan)
+        {
+            return !featureFlags.Value.EnableReplacementOfPublicApiDataSets || deletePlan.ApiDataSetVersionPlan == null
+                ? Unit.Instance
+                : await dataSetVersionService.DeleteVersion(dataSetVersionId: deletePlan.ApiDataSetVersionPlan.Id);
+        }
         public async Task<Either<ActionResult, DataImportStatusViewModel>> GetDataFileImportStatus(
```
jack-hive, line 657: still need a new line here I think

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseVersionService.cs
```diff
@@ -758,17 +770,32 @@ private async Task<Either<ActionResult, Unit>> CheckCanDeleteDataFiles(
...
 
-            return Unit.Instance;
+            async Task<bool> ShouldPreventApiDataSetDeletion(
+                bool isReplacementEnabled,
+                Guid apiDataSetId,
+                SemVersion? version)
+            {
+                return !isReplacementEnabled && (await dataSetVersionRepository
+                    .GetDataSetVersion(apiDataSetId, version)).Status != DataSetVersionStatus.Published;
+            }
```
jack-hive, line 798: I'd pull this into a private method and re-use it on line `604`

I'd also not pass the value of `featureFlags.Value.EnableReplacementOfPublicApiDataSets` into it as you can access it at the class level

Also, should the condition be:

```csharp
private async Task<bool> ShouldPreventApiDataSetDeletion(
                Guid apiDataSetId,
                SemVersion? version)
            {
                var dataSetVersion = await dataSetVersionRepository.GetDataSetVersion(apiDataSetId, version);

                return !isReplacementEnabled || dataSetVersion.Status == DataSetVersionStatus.Published;
            }
```

i.e. should it be a `||` operator and a `==` operator, as opposed to `&&` and `!=`?

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReplacementService.cs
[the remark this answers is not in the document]
jack-hive: Discussed on a call, and have concluded that the following replacement plan scenarios need to be tested for:

**REPLACEMENT PLAN ANALYSIS**

> flag is off

1. original is linked to PUBLISHED + replacement is linked
  - NOW --> SHOULD BE INVALID
3. original is linked to DRAFT + replacement is linked
  - NOW --> SHOULD BE INVALID. 
  - AFTER EES-6172 --> SHOULD BE VALID
6. original is NOT linked but PUBLISHED + replacement is NOT linked
  - NOW --> SHOULD HAVE A NULL API PLAN
8. original is NOT linked but DRAFT + replacement is NOT linked
  - NOW --> SHOULD HAVE NULL API PLAN

> flag is on

1. original is linked to PUBLISHED + replacement is linked
  - NOW --> SHOULD BE VALID AS LONG AS YOU'RE PATCHING IT
3. original is linked to DRAFT + replacement is linked
  - NOW --> SHOULD BE INVALID. 
  - AFTER EES-6172 --> SHOULD BE VALID BUT THE PLAN SHOULD BE RETRIEVABLE BEFORE ANYTHING IS ACTUALLY EXECUTED (NOTHING DELETED)
6. original is NOT linked but PUBLISHED + replacement is NOT linked
  - NOW --> SHOULD HAVE NULL API PLAN
8. original is NOT linked but DRAFT + replacement is NOT linked
  - NOW --> SHOULD HAVE NULL API PLAN

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReplacementService.cs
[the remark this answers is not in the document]
mmoalin: Cheers Jack, how about I've gone for `finishedImporting` after our discussion.

src/GovUk.Education.ExploreEducationStatistics.Admin/Repositories/DataSetVersionRepository.cs
```diff
@@ -27,4 +29,20 @@ public async Task<List<DataSetVersion>> GetDataSetVersions(Guid releaseVersionId
...
+    {
+        return await publicDataDbContext.DataSetVersions
+            .AsNoTracking()
+            .Include(dsv => dsv.DataSet)
+            .Where(dsv => dsv.DataSetId == dataSetId)
+            .Where(dsv => dsv.VersionMajor == version.Major)
+            .Where(dsv => dsv.VersionMinor == version.Minor)
+            .Where(dsv => dsv.VersionPatch == version.Patch)
+            .FirstOrDefaultAsync(cancellationToken: cancellationToken);
+    }
```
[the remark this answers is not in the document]
mmoalin: That's a fair comment, thanks for the great discussion afterwards @jack-hive & @benoutram . I've gotten rid of the repository method and I've used the `DataSetVersionService` method and the `onSuccess` approach.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseVersionService.cs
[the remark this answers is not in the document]
mmoalin: ah true, yeah I think you're right. Had a look around and I dont think this is being used in the FE. It is being used in the BE though. I've started to use the Valid property to validate against deleting a data set version.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseVersionService.cs
mmoalin: Thanks for your time discussing this earlier Jack, I had tidied this up and implemented how you suggested.

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ApiDataSetVersionPlanViewModel.cs
```diff
@@ -16,12 +16,17 @@ public record DeleteApiDataSetVersionPlanViewModel
...
 
-public record ApiDataSetVersionPlanViewModel : DeleteApiDataSetVersionPlanViewModel
+public record DeleteApiDataSetVersionPlanViewModel : ApiDataSetVersionPlanViewModel;
+
+public record ReplaceApiDataSetVersionPlanViewModel : ApiDataSetVersionPlanViewModel
 {
-    public MappingStatusViewModel? MappingStatus { get; set; }
+    public MappingStatusViewModel? MappingStatus { get; init; }
+    
+    public bool FinishedImporting => Status == DataSetVersionStatus.Draft;
```
jack-hive, line 29: just to double check, is there ever a scenario where we will be using this view-model (i.e. retrieving the replacement plan) when the status has moved from `Draft` to `Published`? 

Hopefully not?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseVersionService.cs
```diff
@@ -758,17 +782,25 @@ private async Task<Either<ActionResult, Unit>> CheckCanDeleteDataFiles(
...
                     Detail = new ApiDataSetErrorDetail(releaseFile.PublicApiDataSetId.Value)
                 });
-            }
+        }
 
-            return Unit.Instance;
+        private bool ShouldAllowApiDataSetDeletion(DataSetVersionStatus? dataSetVersionStatus)
+        {
+            return featureFlags.Value.EnableReplacementOfPublicApiDataSets
+                   && dataSetVersionStatus is not DataSetVersionStatus.Published;
```
jack-hive, line 803: sorry, just thought... We might be best checking for **all** non-public statuses

See some examples of extension methods in `DataSetVersionAuthExtensions.cs`. We might want to add another extension method for the `DataSetVersionStatus` type, and overload the method `IsPrivateStatus`

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataSetFileStorage.cs
```diff
@@ -86,26 +86,19 @@ public async Task<DataFileInfo> UploadDataSet(
...
-                replacedReleaseDataFile.PublicApiDataSetId.Value,
-                dataSetVersion.Id,
-                cancellationToken
-            );
+            cancellationToken);
+
+        if (featureFlags.Value.EnableReplacementOfPublicApiDataSets 
+            && dataSet.ReplacingFile is not null 
+            && releaseFileHasApiDataSet
+            && replacedReleaseDataFile is not null)
```
jack-hive, line 98: this seems like a lot of conditions, and I haven't had time to fully digest them yet. But it makes me wonder if we need them all?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataSetFileStorage.cs
```diff
@@ -131,6 +124,29 @@ await dataSetVersionService.CreateNextVersion(
...
 
+    private async Task CreateNextDraftDataSetVersion(
+        CancellationToken cancellationToken,
+        ReleaseFile replacedReleaseDataFile,
+        ReleaseFile dataReleaseFile)
+    {
+        var dataSetId = replacedReleaseDataFile.PublicApiDataSetId;
+        var errorDetails =
+            $"Data set id: {dataSetId?.ToString()}\n" +
+            $"Data set version: {replacedReleaseDataFile.PublicApiDataSetVersion?.Major}.{replacedReleaseDataFile.PublicApiDataSetVersion?.Minor}.{replacedReleaseDataFile.PublicApiDataSetVersion?.Patch}";
```
jack-hive, line 134: I'd imagine at this point we can assume `PublicApiDataSetVersion` is non-null? In which case we could stick a `!` on the end of one of them, and get rid of the `?` marks

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataSetFileStorage.cs
```diff
@@ -131,6 +124,29 @@ await dataSetVersionService.CreateNextVersion(
...
+
+        await dataSetVersionService.GetDataSetVersion(dataSetId!.Value,
+                replacedReleaseDataFile?.PublicApiDataSetVersion!,
+                cancellationToken)
+            .OnFailureDo(_ => throw new Exception("Failed to find the data set version attached to the release file that is being replaced.\n" + errorDetails))
```
jack-hive, line 139: I wonder if we have a standard way of throwing these exceptions and formatting the errors. We possible need to inject a logger? Also worth having a think about the exact type of Exception we want to throw.

I've ran out of time but will come back to this on Monday. Have a think!

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseVersionService.cs
```diff
@@ -620,16 +619,28 @@ public async Task<Either<ActionResult, Unit>> RemoveDataFiles(Guid releaseVersio
...
+                    : await dataSetVersionService
+                    .GetDataSetVersion(
+                        releaseFile.PublicApiDataSetId.Value, 
+                        releaseFile.PublicApiDataSetVersion!))
+                    .OnSuccess(dataSetVersion => (releaseFile, dataSetVersion)))
+                .OnSuccessDo(releaseFileAndVersion => CheckCanDeleteDataFiles(
+                    releaseVersionId, 
+                    releaseFileAndVersion.releaseFile, 
+                    releaseFileAndVersion.dataSetVersion?.Status ?? null))
+                .OnSuccessDo(async releaseFileAndVersion =>
```
jack-hive, line 636: I think it would be better to combine these. So fetch the data set version, if it should exist, inside of the `CheckCanDeleteDataFiles` method.

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/ViewModels/ApiDataSetVersionPlanViewModel.cs
mmoalin: No I don't believe so. This is because when an amendment is made, new release versions and release files get created that represent a draft amendment. So when users are on the replacement page, the linked api data set version will always be a new draft data set version that is not yet published. When users cancel or confirm the replacement, the will be working with the draft api version and not the published version.

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataSetFileStorage.cs
mmoalin: I think I've gotten rid off ` && replacedReleaseDataFile is not null` as i think that seems redundant

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseVersionService.cs
```diff
@@ -762,12 +756,11 @@ private async Task<bool> CanUpdateDataFiles(Guid releaseVersionId)
                     releaseFile.PublicApiDataSetVersion!,
                     cancellationToken)
                 .OnSuccess(dsv => (DataSetVersion?)dsv)
-                .OnFailureDo(_ => throw new ApplicationException(
+                .OnFailureDo(_ => throw new DataSetVersionNotFoundException(
```
jack-hive, line 759: I'd use a built-in exception for this. Perhaps `InvalidOperationException` again?

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/ReleaseVersionService.cs
```diff
@@ -786,7 +779,21 @@ private async Task<Either<ActionResult, Unit>> CheckCanDeleteDataFiles(
...
+            await dataSetVersionService
+                .GetDataSetVersion(
+                    releaseFile.PublicApiDataSetId.Value,
+                    releaseFile.PublicApiDataSetVersion!)
+                .OnFailureDo(_ =>
+                {
+                    var errorMessage =
+                        "Failed to find the data set version expected to be linked to the release file that is being deleted for the " +
+                        $"data set id: {releaseFile.PublicApiDataSetId.Value} and the data set version number: {releaseFile.PublicApiDataSetVersionString}. This has occured when creating the next draft version.";
+                    logger.LogError(errorMessage);
```
jack-hive: again I think a different message for the logger VS the exception might be better

jack-hive (changes requested):

jack-hive (changes requested):

src/GovUk.Education.ExploreEducationStatistics.Admin/Exceptions/DataSetVersionNotFoundException.cs
```diff
@@ -0,0 +1,5 @@
+﻿using System;
+
+namespace GovUk.Education.ExploreEducationStatistics.Admin.Exceptions; 
+
+public class DataSetVersionNotFoundException(string message) : InvalidOperationException(message);
```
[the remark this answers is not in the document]
mmoalin: Cheers Jack, this is reverted and changed to InvalidOperationException

src/GovUk.Education.ExploreEducationStatistics.Admin/Services/DataSetFileStorage.cs
```diff
@@ -129,21 +131,30 @@ private async Task CreateNextDraftDataSetVersion(
...
+                replacedReleaseDataFile.PublicApiDataSetVersion!,
                 cancellationToken)
-            .OnFailureDo(_ => throw new Exception("Failed to find the data set version attached to the release file that is being replaced.\n" + errorDetails))
+            .OnFailureDo(_ =>
+            {
+                var errorMessage =
+                    "Failed to find the data set version expected to be linked to the release file that is being replaced for the " +
+                    $"data set id: {dataSetId} and the data set version number: {replacedReleaseDataFile.PublicApiDataSetVersionString}. This has occured when creating the next draft version.";
+                logger.LogError(errorMessage);
+                throw new DataSetVersionNotFoundException(errorMessage);
```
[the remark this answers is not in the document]
mmoalin: All done!

src/GovUk.Education.ExploreEducationStatistics.Admin.Tests/Services/ReleaseVersionServiceTests.cs
```diff
@@ -128,19 +131,29 @@ public async Task Success()
...
 
-        [Fact]
-        public async Task FileIsLinkedToPublicApiDataSet_InvalidPlan()
+        [Theory] 
+        [InlineData(DataSetVersionStatus.Mapping, true, true)] 
+        [InlineData(DataSetVersionStatus.Draft, true, true)] 
+        [InlineData(DataSetVersionStatus.Mapping, false, false)]
+        [InlineData(DataSetVersionStatus.Published, true, false)] 
+        [InlineData(DataSetVersionStatus.Published, false, false)]
+        [InlineData(DataSetVersionStatus.Draft, false, false)]
```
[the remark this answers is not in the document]
mmoalin: This is done, ta mate.

PR #5896, opened 2025-05-22
EES-6161 fix comment highlighting

amyb-hiveit:
Fixes a bug where comment highlighting didn't work when the selected text was not a whole element - if you selected a whole paragraph and added a comment the highlighting would work, if you only selected part of a paragraph the highlighting would not work.


Guy-HiveIT (approved):
Looks good to me and I don't get the intermittent UI test failures you mentioned

PR #5908, opened 2025-05-27
EES-5708: Consolidate remaining data set upload methods to use a shared approach

tomjonesdev:
This work applies a staged approach to data set uploads for the remaining upload methods (CSV file pair, and zipped CSV file pair), to match the process created for bulk zip uploads in preparation for the screener integration.

The process is; the files are uploaded to the server, validated, then uploaded into a temp container with a result returned to the user. When the screener is integrated, the result will be included in this response. When validation succeeds, the user can then manually trigger the import - this process used to be performed automatically as part of the upload.

### Notes
- I renamed the bulk zip-specific view model to `DataSetUploadResultViewModel` to represent consolidation. The name might stay the same when the screener is integrated, but the structure will change to include the screener result data
- I’ve left the original FE service methods in because they are still used by the replacement page. This area will be consolidated to use the same functions as part of [EES-6176](https://dfedigital.atlassian.net/browse/EES-6176)
- The two upload methods changed here will only ever return a single `DataSetUploadResultViewModel`, but these are returning a `List` for the time being for simplicity and consistency with the bulk upload method. Along with keeping these changes small, this will also avoid unnecessary work, as the response structure will be changed as part of the screener integration
- _New_ duplicate data sets can be uploaded into temp storage before the import is triggered, as the duplication validation only applies to _imported_ data sets. The validation acts against records stored in the database, which don't exist until the import has been confirmed by the user. This is because a file is given a unique path when it is uploaded, consisting of a newly-generated GUID - uploading the same file again will give it a new unique path as the process has no knowledge of previous upload paths. The only issues here are:
  - users may waste their time uploading data sets multiple times if they cancel the upload results modal
  - more storage use in the temporary container, however files are automatically removed after 24h of inactivity (see [EES-5709](https://dfedigital.atlassian.net/browse/EES-5709))
  - Another point to consider is the removal of duplicate data set validation on uploads, and allow auto replacement for all upload methods (i.e. matching bulk upload). This may also lead to the question of whether a separate screen for manual replacements is still needed, see [EES-6176](https://dfedigital.atlassian.net/browse/EES-6176).

### DIFF
--- a/src/explore-education-statistics-admin/src/pages/release/data/components/ReleaseDataUploadsSection.tsx
+++ b/src/explore-education-statistics-admin/src/pages/release/data/components/ReleaseDataUploadsSection.tsx
@@ -5,7 +5,7 @@ import DataFileUploadForm, {
 import releaseDataFileQueries from '@admin/queries/releaseDataFileQueries';
 import permissionService from '@admin/services/permissionService';
 import releaseDataFileService, {
-  ArchiveDataSetFile,
+  DataSetUploadResult,
   DataFile,
   DataFileImportStatus,
 } from '@admin/services/releaseDataFileService';
@@ -16,7 +16,7 @@ import WarningMessage from '@common/components/WarningMessage';
 import useToggle from '@common/hooks/useToggle';
 import { useQuery } from '@tanstack/react-query';
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
-import BulkZipUploadModalConfirm from './BulkZipUploadModalConfirm';
+import DataSetUploadModalConfirm from './DataSetUploadModalConfirm';
 import DataFilesTable from './DataFilesTable';
 
 interface Props {
@@ -33,7 +33,7 @@ export default function ReleaseDataUploadsSection({
   onDataFilesChange,
 }: Props) {
   const [allDataFiles, setAllDataFiles] = useState<DataFile[]>([]);
-  const [bulkUploadPlan, setBulkUploadPlan] = useState<ArchiveDataSetFile[]>();
+  const [uploadResults, setUploadResults] = useState<DataSetUploadResult[]>();
   const [isReordering, toggleReordering] = useToggle(false);
 
   const {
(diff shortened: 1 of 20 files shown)


src/explore-education-statistics-admin/src/pages/release/data/components/ReleaseDataUploadsSection.tsx
```diff
@@ -33,7 +33,7 @@ export default function ReleaseDataUploadsSection({
   onDataFilesChange,
 }: Props) {
   const [allDataFiles, setAllDataFiles] = useState<DataFile[]>([]);
-  const [bulkUploadPlan, setBulkUploadPlan] = useState<ArchiveDataSetFile[]>();
+  const [uploadResults, setUploadResult] = useState<DataSetUploadResult[]>();
```
duncan-at-hiveit, line 36: Rename to `setUploadResults` plural?

src/explore-education-statistics-admin/src/pages/release/data/components/ReleaseDataUploadsSection.tsx
```diff
@@ -208,12 +218,11 @@ export default function ReleaseDataUploadsSection({
             dataFiles={allDataFiles}
             onSubmit={handleSubmit}
           />
-          {bulkUploadPlan === undefined ||
-          bulkUploadPlan.length === 0 ? null : (
-            <BulkZipUploadModalConfirm
-              bulkUploadPlan={bulkUploadPlan}
-              onConfirm={confirmBulkUploadPlan}
-              onCancel={() => setBulkUploadPlan(undefined)}
+          {uploadResults === undefined || uploadResults.length === 0 ? null : (
```
duncan-at-hiveit, line 221: I think this can be simplified to:

```
{uploadResults?.length > 0 && (
        <DataSetUploadModalConfirm
              uploadResults={uploadResults}
              onConfirm={confirmDataSetImport}
              onCancel={() => setUploadResult(undefined)}
         />
)}` 
```

but I could be wrong!

src/explore-education-statistics-admin/src/services/releaseDataFileService.ts
```diff
@@ -132,11 +132,14 @@ const releaseDataFileService = {
     data.append('metaFile', metadataFile);
     data.append('replacingFileId', replacingFileId ?? '');
 
-    const file = await client.post<DataFileInfo>(`/releaseVersions/data`, data);
+    const file = await client.post<DataFileInfo>(
+      `/releaseVersions/replacement-data`,
```
duncan-at-hiveit, line 136: This can use simple hyphens rather than backticks if there's no string tokenisation going on.  I realise it was like this when you found it!

duncan-at-hiveit (approved):

PR #5923, opened 2025-05-29
EES-6091 - adding existing Public API queries analytics code into Ana…

duncan-at-hiveit:
This PR:
- moves the analytics-capturing Public API query code into AnalyticsService.
- adds Preview Token and requested data set version details to Public API query analytics, to bring in line with all other DataSetVersion-level calls in the Public API.
- simplifies the disabling of analytics in the Public API by using a NoOpAnalyticsService rather than an NoOpAnalyticsManager.


src/GovUk.Education.ExploreEducationStatistics.Public.Data.Api/Services/AnalyticsService.cs
```diff
@@ -66,6 +68,48 @@ public async Task CaptureDataSetVersionCall(
...
+        DataSetVersion dataSetVersion,
+        string? requestedDataSetVersion,
+        DataSetQueryRequest query,
+        DataSetQueryPaginatedResultsViewModel results,
+        DateTime startTime,
+        CancellationToken cancellationToken)
+    {
+        try
+        {
+            if (!IncludeAnalyticsCall())
```
mmyoungman, line 82: I'd add a comment to this matching the other PR

mmyoungman (approved):

